PR #21423, opened 2022-05-24
connect: not doing delay close for CONNECT responses

alyssawilk:
Not doing delay close for CONNECT responses as they frame by connection close.

Risk Level: low
Testing: improved integration tests
Docs Changes: n/a
Release Notes: inline
runtime guard: envoy.reloadable_features.no_delay_close_for_upgrades


source/common/http/http1/codec_impl.cc
```diff
@@ -295,8 +295,8 @@ void StreamEncoderImpl::endEncode() {
   flushOutput(true);
   connection_.onEncodeComplete();
   // With CONNECT or TCP tunneling, half-closing the connection is used to signal end stream.
-  if (connect_request_ || is_tcp_tunneling_) {
-    connection_.connection().close(Network::ConnectionCloseType::FlushWriteAndDelay);
+  if (connect_request_ || is_tcp_tunneling_ || is_response_to_connect_request_) {
+    connection_.connection().close(Network::ConnectionCloseType::FlushWrite);
```
mattklein123, line 299: Can you add some comment trail here on why we are using flush write without delay and we used to do the delay? It's not immediately clear to me.

source/common/http/conn_manager_impl.cc
```diff
@@ -251,7 +251,8 @@ void ConnectionManagerImpl::doEndStream(ActiveStream& stream) {
   bool connection_close = stream.state_.saw_connection_close_;
   bool request_complete = stream.filter_manager_.remoteDecodeComplete();
 
-  checkForDeferredClose(connection_close && (request_complete || http_10_sans_cl));
+  checkForDeferredClose((connection_close && (request_complete || http_10_sans_cl)) ||
+                        stream.state_.is_tunneling_);
```
mattklein123, line 255: Can you add some comments here on the various conditions? It's kind of complicated now and hard to quickly parse.

mattklein123 (changes requested):
Thanks for fixing. Mostly a request for more comments which I can then read to make sure I actually understand. :)

/wait

source/common/http/conn_manager_impl.cc
mattklein123: OK I think I'm roughly following you here. I could go either way, but I would request:
1) More comments here for those that aren't as well versed in all things CONNECT
2) Release note
3) It seems like this should probably be guarded, but I will defer to your opinion on how risky this is.

mattklein123 (approved):
Thanks!

mattklein123:
Sorry looks like this needs a main merge.

/wait

PR #21426, opened 2022-05-24
http_cache: added route config in http cache filter

sergkir85:
Signed-off-by: Sergey Kirnos <s.kirnos85@gmail.com>

Commit Message: http_cache: Support per-route Http Cache filter configuration
Additional Description:

This pull request allows to specify Http Cache filter config on per-route basis. It means that for specific route can be set specific http cache type (simple, fs etc) and also can be overridden global config parameters (at present only "vary allow headers").

Risk Level: low
Testing: New unit and integration tests added.
Release Notes:
Platform Specific Features:


source/extensions/filters/http/cache/cache_filter.cc
```diff
@@ -30,11 +31,10 @@ struct CacheResponseCodeDetailValues {
...
 using CacheResponseCodeDetails = ConstSingleton<CacheResponseCodeDetailValues>;
 
-CacheFilter::CacheFilter(const envoy::extensions::filters::http::cache::v3::CacheConfig& config,
-                         const std::string&, Stats::Scope&, TimeSource& time_source,
-                         HttpCache& http_cache)
-    : time_source_(time_source), cache_(http_cache),
-      vary_allow_list_(config.allowed_vary_headers()) {}
+CacheFilter::CacheFilter(const std::string&, Stats::Scope&, TimeSource& time_source,
+                         const CacheFilterConfig& config)
+    : config_(config), cache_(nullptr), time_source_(time_source),
```
ravenblackx, line 36: I think it's more envoy-style to do like `=nullptr` in the header file for things to be initialized that don't require parameters.

source/extensions/filters/http/cache/cache_filter_config.h
```diff
@@ -0,0 +1,40 @@
...
+#include "source/extensions/filters/http/cache/http_cache.h"
+#include "source/extensions/filters/http/common/pass_through_filter.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+
+using CacheFilterConfigPb = envoy::extensions::filters::http::cache::v3::CacheConfig;
+using CacheFilterConfigPbRef = std::reference_wrapper<const CacheFilterConfigPb>;
```
ravenblackx, line 20: Why do you need a reference_wrapper rather than just a plain reference here? It's only set at init-time.

source/extensions/filters/http/cache/http_cache.h
```diff
@@ -268,9 +269,17 @@ class HttpCacheFactory : public Config::TypedFactory {
   //
   // Pass factory context to allow HttpCache to use async client, stats scope
   // etc.
-  virtual std::shared_ptr<HttpCache>
+
+  // Get cache for global config
+  virtual HttpCachePtr
   getCache(const envoy::extensions::filters::http::cache::v3::CacheConfig& config,
            Server::Configuration::FactoryContext& context) PURE;
```
ravenblackx, line 276: Rather than splitting this into two identical functions, I think it would work to have this function take a [Server::Configuration::CommonFactoryContext&](https://github.com/envoyproxy/envoy/blob/main/envoy/server/factory_context.h#L108).

(Then you could also propagate this change up to where you have a templated getCache to simplify that too.)

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
```diff
@@ -305,7 +305,7 @@ CacheInfo SimpleHttpCache::cacheInfo() const {
   return cache_info;
 }
 
-SINGLETON_MANAGER_REGISTRATION(simple_http_cache_singleton);
+SINGLETON_MANAGER_REGISTRATION(simple_http_cache);
```
ravenblackx, line 308: I'd avoid renaming preexisting variables unnecessarily, when the code belongs to someone else.

test/extensions/filters/http/cache/cache_filter_integration_test.cc
```diff
@@ -597,6 +598,54 @@ TEST_P(CacheIntegrationTest, ServeGetFollowedByHead200WithValidation) {
...
+        auto* config = cfg.mutable_route_config()
+                           ->mutable_virtual_hosts()
+                           ->Mutable(0)
+                           ->mutable_typed_per_filter_config();
+
+        (*config)["envoy.filters.http.cache"].PackFrom(cache_per_route);
+      };
+}
+
+TEST_P(CacheIntegrationTest, RouteOverride) {
```
ravenblackx, line 625: How does this test verify that the override in vhost worked?

It might be easier to usefully test that route_config works in a unit test. I think it's config_test in file_system_buffer_filter that does some of that.

ravenblackx (changes requested):

source/extensions/filters/http/cache/cache_filter_config.h
sergkir85: If leave just reference then in `cache_filter` following line `config_ = std::move(*cfg);` raises error (like copy constructor is deleted ...).
For this reason I have added `reference_wrapper`

sergkir85: I think that you will not have nullptr but sort of empty config you will). Removed std::move.

ravenblackx: The HttpCachePtr of the original would become a nullptr after being std::moved from. Looks like a std::reference_wrapper's move is the same as a copy so it wouldn't hurt that one.

source/extensions/filters/http/cache/cache_filter.h
```diff
@@ -86,8 +88,9 @@ class CacheFilter : public Http::PassThroughFilter,
   // Updates filter_state_ and continues the encoding stream if necessary.
   void finalizeEncodingCachedResponse();
 
+  CacheFilterConfig config_;
+  HttpCachePtr cache_ = nullptr;
```
ravenblackx, line 92: Sorry, this is my fault - I said rather than explicit initialization in the constructor it could be done here, but now that it's here I realize HttpCachePtr is a shared_ptr so it already default-initializes to nullptr. (If anything =nullptr here harms readability because it makes it look like it's a regular pointer rather than the shared pointer that it is.)

source/extensions/filters/http/cache/http_cache.h
```diff
@@ -256,6 +256,7 @@ class HttpCache {
 
   virtual ~HttpCache() = default;
 };
+using HttpCachePtr = std::shared_ptr<HttpCache>;
```
ravenblackx, line 259: Envoy style has been moving towards naming this HttpCacheSharedPtr to avoid confusion as to whether it's shared/unique/bare at all the use-sites.

source/extensions/filters/http/cache/config.cc
```diff
@@ -1,32 +1,56 @@
...
 namespace Envoy {
 namespace Extensions {
 namespace HttpFilters {
 namespace Cache {
 
-Http::FilterFactoryCb CacheFilterFactory::createFilterFactoryFromProtoTyped(
-    const envoy::extensions::filters::http::cache::v3::CacheConfig& config,
-    const std::string& stats_prefix, Server::Configuration::FactoryContext& context) {
+namespace {
+template <typename FactoryCtx>
```
ravenblackx, line 12: This no longer needs to be a template - you can use CommonFactoryContext&, because it's a superclass of both the other types.

source/extensions/filters/http/cache/config.cc
```diff
@@ -1,32 +1,56 @@
...
+Http::FilterFactoryCb CacheFilterFactory::createFilterFactoryFromProtoTyped(
+    const envoy::extensions::filters::http::cache::v3::CacheConfig& config,
+    const std::string& stats_prefix, Server::Configuration::FactoryContext& context) {
+
+  auto cache = getCache(config, context);
+
+  return [config, stats_prefix, cache,
+          &context](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+    callbacks.addStreamFilter(std::make_shared<CacheFilter>(
+        stats_prefix, context.scope(), context.timeSource(), CacheFilterConfig(cache, config)));
```
ravenblackx, line 41: CacheFilter's constructor is taking a reference to CacheFilterConfig, and here you're passing in a temporary instance. As soon as this call goes out of scope that reference will be invalid, and the Filter outlives it.

You can (and should) still construct with a reference though - it just needs to be captured and owned by this lambda, rather than a temporary, i.e. `auto cache_filter_config = CacheFilterConfig(getCache(config, context))` instead of `auto cache =...` before the lambda.

OR you can pass it in as a temporary but make sure it's fully consumed during the constructor, i.e. initialize cache_ and vary_allow_headers_ and don't store config_.

test/extensions/filters/http/cache/cache_filter_test.cc
```diff
@@ -15,17 +15,18 @@ namespace Envoy {
...
 protected:
   // The filter has to be created as a shared_ptr to enable shared_from_this() which is used in the
   // cache callbacks.
-  CacheFilterSharedPtr makeFilter(HttpCache& cache) {
-    auto filter = std::make_shared<CacheFilter>(config_, /*stats_prefix=*/"", context_.scope(),
-                                                context_.timeSource(), cache);
+  CacheFilterSharedPtr makeFilter(HttpCachePtr cache) {
+    auto filter =
+        std::make_shared<CacheFilter>(/*stats_prefix=*/"", context_.scope(), context_.timeSource(),
+                                      CacheFilterConfig(simple_cache_, config_));
```
ravenblackx, line 26: If you fix the other reference-to-a-temporary-object at source rather than by consuming it in the constructor then you'll have the same problem here, and would need to make this CacheFilterConfig persistent for the life of the test.

source/extensions/filters/http/cache/cache_filter.cc
```diff
@@ -63,10 +78,12 @@ Http::FilterHeadersStatus CacheFilter::decodeHeaders(Http::RequestHeaderMap& hea
   }
   ASSERT(decoder_callbacks_);
 
+  initConfig();
```
ravenblackx, line 81: No need to initConfig in encodeHeaders because you can't all of
1. get here
2. still need a cache
3. not have gone through the initConfig in decodeHeaders

(Some filters do need to do it in both paths because an early-out in decodeHeaders can mean they still need the config for the reply, but the only early-outs in encodeHeaders here involve either already having the cache or not needing it.)

source/extensions/filters/http/cache/cache_filter.cc
```diff
@@ -46,6 +46,21 @@ void CacheFilter::onDestroy() {
...
+    return;
+  }
+
+  const auto* cfg = Http::Utility::resolveMostSpecificPerFilterConfig<CacheFilterConfig>(
+      "envoy.filters.http.cache", decoder_callbacks_->route());
+  if (cfg != nullptr) {
+    config_ = *cfg;
+  }
+
+  cache_ = config_.getCache();
```
ravenblackx, line 60: I think here you need to be updating `vary_allow_list_` too.

ravenblackx (changes requested):

test/extensions/filters/http/cache/cache_filter_integration_test.cc
sergkir85: added unittest

source/extensions/filters/http/cache/cache_filter.cc
sergkir85: It is not only about cache but also other config parameters. If situation is possible: 'init_config has not been called in decode_headers' then I would prefer to leave this config initialization here.

test/extensions/filters/http/cache/cache_filter_test.cc
```diff
@@ -575,6 +582,39 @@ TEST_F(CacheFilterTest, GetRequestWithBodyAndTrailers) {
...
+
+  // Filter cache should be null (due to we create filter with null cache - 'makeFilter(nullptr)')
+  ASSERT_TRUE(getFilterCache(filter) == nullptr);
+  EXPECT_EQ(Http::FilterHeadersStatus::StopAllIterationAndWatermark,
+            filter->decodeHeaders(request_headers_, true));
+
+  // During header decoding filter should reinit config using per route config.
+  // It means that cache should be initialized with 'simple_cache_' and allowed_vary_headers should
+  // be taken from per route config as well
+  ASSERT_TRUE(getFilterCache(filter) == simple_cache_);
```
ravenblackx, line 614: `ASSERT_EQ(getFilterCache(filter), simple_cache_);` would potentially have more useful output in case of failure. Also, probably makes more sense as EXPECT_EQ since the test isn't going to crash if the line fails, which would allow for *also* getting more useful output from the next line.

test/extensions/filters/http/cache/cache_filter_test.cc
```diff
@@ -575,6 +582,39 @@ TEST_F(CacheFilterTest, GetRequestWithBodyAndTrailers) {
...
+  ASSERT_TRUE(getFilterCfg(filter).proto().allowed_vary_headers_size() > 0);
```
ravenblackx, line 615: Similarly, you could get more readable output from this on failure, given that the configs should be identical, with
EXPECT_EQ(getFilterCfg(filter).proto().DebugString(), cache_per_route.DebugString());

In general, EXPECT_TRUE and ASSERT_TRUE shouldn't be used for conditions, they should only be used for testing boolean values - writing the condition in actual test terms will nearly always give more useful output on failure.

Ideally this one would be EXPECT_THAT(getFilterCfg(filter).proto(), ProtoEquals(cache_per_route)), but it seems ProtoEquals isn't as widely available as it should be.

ravenblackx (changes requested):

source/extensions/filters/http/cache/cache_filter_config.h
```diff
@@ -0,0 +1,40 @@
...
+
+// Config contains raw proto config and cache reference which is used in current filter.
+// Initially 'CacheRouteFilterConfig' is created from global config and can be
+// overriden with route config (see config.cc)
```
ravenblackx, line 24: This comment isn't true any more (not a cache reference, and CacheRouteFilterConfig is not created from global config). Please correct it.

ravenblackx (changes requested):

source/extensions/filters/http/cache/http_cache.h
ravenblackx, line 259: Guess I'll say it again since there's no way to unmark a conversation as resolved:
Envoy style has been moving towards naming this HttpCacheSharedPtr to avoid confusion as to whether it's shared/unique/bare at all the use-sites.

ravenblackx (changes requested):

ravenblackx (approved):

source/extensions/filters/http/cache/cache_filter_config.h
```diff
@@ -0,0 +1,35 @@
...
+
+using CacheFilterConfigPb = envoy::extensions::filters::http::cache::v3::CacheConfig;
+using CacheFilterConfigPbRef = std::reference_wrapper<const CacheFilterConfigPb>;
+
+/**
+ * Route configuration for the HTTP cache filter.
+ */
+class CacheRouteFilterConfig : public Router::RouteSpecificFilterConfig {
+public:
+  explicit CacheRouteFilterConfig(HttpCacheSharedPtr cache, const CacheFilterConfigPb& config)
```
jmarantz, line 21: nit: explicit not needed for 2-arg ctor

source/extensions/filters/http/cache/cache_filter.cc
```diff
@@ -30,11 +31,12 @@ struct CacheResponseCodeDetailValues {
 
 using CacheResponseCodeDetails = ConstSingleton<CacheResponseCodeDetailValues>;
 
-CacheFilter::CacheFilter(const envoy::extensions::filters::http::cache::v3::CacheConfig& config,
-                         const std::string&, Stats::Scope&, TimeSource& time_source,
-                         HttpCache& http_cache)
-    : time_source_(time_source), cache_(http_cache),
-      vary_allow_list_(config.allowed_vary_headers()) {}
+bool is_route_config_init_ = false;
```
jmarantz, line 34: This does not look thread-safe at first glance, and that might be the tip of the iceberg of issues with this. Can you add code comments detailing the lifetime and scope of the route config init and how it relates to (say) having multiple tenants using the cache in the same process?


Update: I see this is also the name of a non-static member variable. Maybe you meant to just use the member variable and not a global? Can you just delete the above line?

source/extensions/filters/http/cache/cache_filter.cc
```diff
@@ -56,17 +75,21 @@ Http::FilterHeadersStatus CacheFilter::decodeHeaders(Http::RequestHeaderMap& hea
...
+
   if (!CacheabilityUtils::canServeRequestFromCache(headers)) {
     ENVOY_STREAM_LOG(debug, "CacheFilter::decodeHeaders ignoring uncacheable request: {}",
                      *decoder_callbacks_, headers);
     return Http::FilterHeadersStatus::Continue;
   }
+
   ASSERT(decoder_callbacks_);
 
+  initRouteConfig();
```
jmarantz, line 87: I think if you make initRouteConfig thread-safe (ignoring other potential issues of sharing) you may be tempted to add a mutex, which could be contended in this data-plane context.

Consider initializing config related stuff in the filter's Config object instead.

jmarantz (commented):
I started this review but I think I want to pause and let @capoferro or @toddmgreer review first and then I'll do a final pass.

Thanks!

source/extensions/filters/http/cache/cache_filter.cc
ravenblackx: Route config can't be initialized in the Config object, because it requires the route() that isn't even available until decodeHeaders. It has to be initialized per-filter, usually during decodeHeaders. (See, e.g. the call to [initConfig](https://github.com/envoyproxy/envoy/blob/0ab9440ec3082800f72ef7feeec393644f1fb89a/source/extensions/filters/http/buffer/buffer_filter.cc#L57) in buffer_filter.)

source/extensions/filters/http/cache/cache_filter_config.h
toddmgreer, line 21: Please document that CacheRouteFilterConfig retains a reference to `config`.

source/extensions/filters/http/cache/config.cc
```diff
@@ -2,31 +2,56 @@
...
+  return [config, stats_prefix, cache,
+          &context](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+    callbacks.addStreamFilter(std::make_shared<CacheFilter>(stats_prefix, context.scope(),
+                                                            context.timeSource(), config, cache));
   };
 }
 
+Router::RouteSpecificFilterConfigConstSharedPtr
+CacheFilterFactory::createRouteSpecificFilterConfigTyped(
+    const envoy::extensions::filters::http::cache::v3::CacheConfig& config,
```
toddmgreer, line 48: Is it safe for createFilterFactoryFromProtoTyped to store a reference to `config` into its return value? (I couldn't find any documented guarantees.)

toddmgreer (changes requested):

source/extensions/filters/http/cache/cache_filter_config.h
```diff
@@ -0,0 +1,35 @@
...
+      : cache_(cache), pb_config_(config) {}
+
+  const CacheFilterConfigPb& proto() const { return pb_config_; }
+  HttpCacheSharedPtr getCache() const { return cache_; }
+
+private:
+  HttpCacheSharedPtr cache_;
+  CacheFilterConfigPbRef pb_config_;
```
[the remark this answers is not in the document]
jmarantz: It looks to me like the original reason for reference_wrapper was to support a std::move, but that std::move has been dropped. So I think it would be cleanest to make this a plain reference.

source/extensions/filters/http/cache/config.cc
jmarantz: This overall pattern of holding onto refs in constructed objects is common in Envoy.

In this case I am not sure if it's safe though...configs get updated via xDS during operation and I'm not clear on the lifetime of a config.

Looking at the usage i think we are just copying a couple of simple fields though. Instead of storing the config by reference can we just copy it? Is it very large? If so we could just copy explicitly the fields that matter for the lookup.

jmarantz (commented):
/wait

source/extensions/filters/http/cache/config.cc
ravenblackx: The config is captured by copy in the lambda (line 39). I believe the FilterFactory is guaranteed to persist as long as any filter using it. ["This happens after all workers have discarded the previous config so it can be safely deleted on the main thread by an update with the new config."](https://sourcegraph.com/github.com/envoyproxy/envoy@b88641b/-/blob/source/common/filter/config_discovery_impl.h?L134) (though it's not entirely unambiguous.)

Copying the proto (of unknown size because it has a type-extension field in it) would be pretty expensive as a thing to do per filter instance. Just taking the specific fields would be dubious because the whole point is that fields from the type-extension field might be needed by the cache implementation (though potentially those fields could be bound up through getCache into the cache instance shared_ptr).

source/extensions/filters/http/cache/config.cc
```diff
@@ -2,31 +2,56 @@
...
+    Server::Configuration::ServerFactoryContext& context, ProtobufMessage::ValidationVisitor&) {
+
+  auto cache = getCache(config, context);
```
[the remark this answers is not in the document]
toddmgreer: It should be possible for a cache implementation to have a heavy, global non-route-specific configuration (e.g. RPC server, disk partition, global timeouts), and a lighter, local route-specific configuration (e.g. admission rules, per-route rate limits, default max-age). In other implementations, non-route-specific configuration may just be default values for the route-specific config.

Regardless, the final step of config acceptance has to be per-filter object, because config changes. Regarding the cost of repeatedly parsing cache implementation protobufs, perhaps we need an HttpCacheFactory method that consumes a protobuf and returns an interface object for a parsed form, that we would then pass to the cache implementation?

I definitely don't have the answers figured out, but I'm confident we'll find them.

source/extensions/filters/http/cache/config.cc
ravenblackx: Why was this marked resolved without doing anything?

PR #21431, opened 2022-05-25
fuzz test: Allow quiche to continue after error

vehre-x41:
Commit Message: fuzz test: Prevent aborts in quiche, that could continue.
Additional Description:
Quiche logs and aborts on occasions were it could continue. Prevent this
by setting the disable abort flag.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


vehre-x41:
CC @adisuissa @KBaichoo @pradeepcrao

vehre-x41:
Er??? What is that precheck format_pre about? All I see is a missing changelog entry for `envoy_reloadable_features_lua_respond_with_send_local_reply` which is not referenced in the patch. What can I do about it?

adisuissa (commented):
LGTM, thanks!
Pre-check problem will probably be solved by merging main.

Just to keep a reference for why it makes sense to disable Oghttp2 aborts:
If when disabling the FATAL log abort the fuzzer will create a test that reaches an inconsistent state (and crashes/accesses inconsistent memory), then it will be a bug we'll need to further evaluate. However, in fuzzing we allow OgHttp2 reaching FATAL states that may happen in production environments.

adisuissa:
> Do want me to add this:
> 
> ```
> Just to keep a reference for why it makes sense to disable Oghttp2 aborts:
> If when disabling the FATAL log abort the fuzzer will create a test that reaches an inconsistent state (and crashes/accesses inconsistent memory), then it will be a bug we'll need to further evaluate. However, in fuzzing we allow OgHttp2 reaching FATAL states that may happen in production environments.
> ```
> 
> to somewhere in the code. Or will it be sufficient to stay in the PR?

Yeah, good idea. Feel free to change it if you think it will help clarify the intention.

vehre-x41:
Done

adisuissa (approved):
LGTM, thanks!

Issue #20378, opened 2022-03-16
TcpProxy improperly sets `:protocol=bytestream` for HTTP2


stevenctl:
**TcpProxy improperly sets `:protocol=bytestream` for HTTP2**:  When using `TunnelingConfig` for an HTTP/2 tunnel, it send an out-of-spec psuedo header. 

*Description*:

https://github.com/envoyproxy/envoy/blob/main/source/common/tcp_proxy/upstream.cc#L282-L285 

Envoy will always set `:protocol`, but it's not part of the HTTP/2 spec:

* https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3

An extended CONNECT spec does mention it however:

* https://datatracker.ietf.org/doc/html/rfc8441#section-3

This becomes a problem for picky implementations, like Go/Golang's net/http:

* https://github.com/golang/go/issues/32763
* https://github.com/golang/go/blob/81d3c25c6cf39a76b17ab4eda97e8ad7b92a21e9/src/net/http/h2_bundle.go#L2808-L2833

*Repro steps*:

1. Run. a Go HTTP/2 server with `GODEBUG="http2debug=2"`
2. Create a listener with `TunnelingConfig` (UsePost=false)
3. Send traffic from the listener to a cluster with `Http2ProtocolOptions` `AllowConnect` and a `static` endpoint which is a Golang HTTP2 server. 
4. You will see `http2: invalid pseudo headers: invalid pseudo-header ":protocol"`

stevenctl:
cc @lambdai

ggreenway:
cc @alyssawilk

lambdai:
Thank @stevenctl for drafting this.

@alyssawilk 
I would propose tcp_proxy tunnel doesn't set `:protocol` by default because tcp_proxy unlikely to proxy websocket 

Also, if any protocol override is needed we can relax `headers_to_add` [here](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto#L84).

This relax aligns with other requirements of setting host headers in tunnel. I have some code implementing the relaxation without impacting the HCM router

stevenctl:
Additional issue:

For HTTP **2** CONNECT [rfc7540 8.3](https://datatracker.ietf.org/doc/html/rfc7540#section-8.3):

   o  The ":scheme" and ":path" pseudo-header fields MUST be omitted.

[...]

PR #21440, opened 2022-05-25
http: fixing up CONNECT to be RFC compliant

alyssawilk:
Risk Level: Medium (data plane changes)
Testing: updated tests
Docs Changes: n/a
Release Notes: inline
Runtime guard: envoy.reloadable_features.use_rfc_connect
Fixes https://github.com/envoyproxy/envoy/issues/20378


alyssawilk:
cc @howardjohn

test/common/tcp_proxy/upstream_test.cc
```diff
@@ -243,6 +247,18 @@ TYPED_TEST(HttpUpstreamRequestEncoderTest, RequestEncoder) {
   this->upstream_->setRequestEncoder(this->encoder_, false);
 }
 
+TYPED_TEST(HttpUpstreamRequestEncoderTest, RequestEncoder) {
+  this->setupUpstream();
+  std::unique_ptr<Http::RequestHeaderMapImpl> expected_headers;
+  expected_headers = Http::createHeaderMap<Http::RequestHeaderMapImpl>({
+      {Http::Headers::get().Method, "CONNECT"},
+      {Http::Headers::get().Host, this->config_->hostname()},
```
howardjohn, line 255: This is great, thanks for the fix!

One possibly concern is that in the current code we are using a header to replace Host, which allows dynamic metadata (
```go
TunnelingConfig: &tcp.TcpProxy_TunnelingConfig{
    Hostname: "ignored",
    HeadersToAdd: []*core.HeaderValueOption{
        {Header: &core.HeaderValue{Key: "x-destination", Value: "%DYNAMIC_METADATA([\"tunnel\", \"address\"])%"}},
    },
},
```
).

Now we would want to use Hostname to follow the spec, but that is only a static string.

Seems like something that could be expanded in a followup though @lambdai

source/common/http/http2/codec_impl.cc
```diff
@@ -300,9 +300,15 @@ Status ConnectionImpl::ClientStreamImpl::encodeHeaders(const RequestHeaderMap& h
     // configurable if necessary.
     // https://tools.ietf.org/html/draft-kinnear-httpbis-http2-transport-02
     modified_headers = createHeaderMap<RequestHeaderMapImpl>(headers);
-    modified_headers->setProtocol(Headers::get().ProtocolValues.Bytestream);
-    if (!headers.Path()) {
-      modified_headers->setPath("/");
+    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.use_rfc_connect")) {
```
RyanTheOptimist, line 303: Should we consider moving the comment from line 296 down into the `if` blocks and editing it accordingly?

source/common/quic/envoy_quic_client_stream.cc
```diff
@@ -43,7 +43,8 @@ Http::Status EnvoyQuicClientStream::encodeHeaders(const Http::RequestHeaderMap&
   SendBufferMonitor::ScopedWatermarkBufferUpdater updater(this, this);
   auto spdy_headers = envoyHeadersToSpdyHeaderBlock(headers);
   if (headers.Method()) {
-    if (headers.Method()->value() == "CONNECT") {
+    if (headers.Method()->value() == "CONNECT" &&
+        !Runtime::runtimeFeatureEnabled("envoy.reloadable_features.use_rfc_connect")) {
```
RyanTheOptimist, line 47: I'm slightly surprised not to see an `else` for the case where the runtime flag is enabled to be parallel with the logic in the h2 code. Is that not required?

test/common/http/http2/codec_impl_test.cc
```diff
@@ -3639,6 +3640,32 @@ TEST_P(Http2CodecImplTest, ConnectTest) {
   driveToCompletion();
 }
 
+// CONNECT without upgrade type gets tagged with "bytestream"
```
RyanTheOptimist, line 3643: Given that the codec strips the `:protocol = bytestream` header, I'm not sure I understand this comment.

test/extensions/filters/http/jwt_authn/filter_integration_test.cc
```diff
@@ -303,16 +303,11 @@ TEST_P(LocalJwksIntegrationTest, ConnectRequestWithRegExMatch) {
...
-    // to find a route match and return a 404.
-    ASSERT_TRUE(response->waitForEndStream());
-    ASSERT_TRUE(response->complete());
-    EXPECT_EQ("404", response->headers().getStatusValue());
-  } else {
-    ASSERT_TRUE(response->waitForReset());
-    ASSERT_TRUE(codec_client_->waitForDisconnect());
-  }
+  // Because CONNECT requests do not include a path, they will fail
+  // to find a route match and return a 404.
```
RyanTheOptimist, line 307: It looks like this is a change in behavior that is visible to this filter. Presumably that's ok (and expected). Is there any need to call this out in release notes, or anything? Probably not, since they already mention the flag?

RyanTheOptimist (commented):
Looks great. A couple of minor comments.

RyanTheOptimist (dismissed):

alyssawilk:
also @moderation if you have thoughts please feel encouraged to do a pass as well

alyssawilk:
@snowp ping?

source/common/tcp_proxy/upstream.cc
```diff
@@ -278,13 +278,18 @@ void Http2Upstream::setRequestEncoder(Http::RequestEncoder& request_encoder, boo
...
       {Http::Headers::get().Method, config_.usePost() ? "POST" : "CONNECT"},
       {Http::Headers::get().Host, config_.hostname()},
-      {Http::Headers::get().Path, "/"},
-      {Http::Headers::get().Scheme, scheme},
   });
 
-  if (!config_.usePost()) {
-    headers->addReference(Http::Headers::get().Protocol,
-                          Http::Headers::get().ProtocolValues.Bytestream);
+  if (!Runtime::runtimeFeatureEnabled("envoy.reloadable_features.use_rfc_connect")) {
```
snowp, line 283: super nit but it might be nice to have all these checks have the feature enabled path first, just to make it easier to follow along the two branches when reading

snowp (changes requested):
One small nit but otherwise this LGTM

PR #21443, opened 2022-05-25
thrift: ensure an RPC isn't removed more than once

rgs1:
Signed-off-by: Raul Gutierrez Segales <rgs@pinterest.com>


JuniorHsu (approved):
Thanks.

zuercher (approved):

PR #21445, opened 2022-05-25
Simplify Parser HTTP version calls.

bencebeky:
Replace Parser::httpMajor() and httpMinor() with a single isHttp11()
method, as that is the only thing ConnectionImpl cares about.  Note that
Protocol struct does not define HTTP/0.9 (in envoy/http/protocol.h)
anyway.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/http/http1/codec_impl.cc
+++ b/source/common/http/http1/codec_impl.cc
@@ -769,7 +769,7 @@ StatusOr<CallbackResult> ConnectionImpl::onHeadersCompleteImpl() {
   ENVOY_CONN_LOG(trace, "onHeadersCompleteBase", connection_);
   RETURN_IF_ERROR(completeLastHeader());
 
-  if (!(parser_->httpMajor() == 1 && parser_->httpMinor() == 1)) {
+  if (!parser_->isHttp11()) {
     // This is not necessarily true, but it's good enough since higher layers only care if this is
     // HTTP/1.1 or not.
     protocol_ = Protocol::Http10;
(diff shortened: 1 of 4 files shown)


bencebeky:
/assign @KBaichoo 

Kevin: PTAL.  Thanks!

KBaichoo (commented):
Hmm, might be more messy as though that's not in that enum from https://github.com/envoyproxy/envoy/blob/575ab6ab48b22ca0b6e9b6acae51aba3f6b7fc8a/api/envoy/config/core/v3/protocol.proto#L288 I think if we accept HTTP 1.0 we do support 0.9.

cc @alyssawilk

alyssawilk:
Haha, was about to say I didn't remember and we should write a test but looks like I did :-P

https://github.com/envoyproxy/envoy/blob/main/test/integration/integration_test.cc#L1008

bencebeky:
I'm looking at Kevin's and Alyssa's last comment.  Thank y'all for the code pointers.  If I understand correctly, you are stating that HTTP/0.9 support is tied to HTTP/1.0 support.  As a consequence, there is indeed no need to distinguish HTTP/1.0 from HTTP/0.9 in the parser, so this change is okay.  Would you like me to update the PR description with this explanation?  Is there any other action for me here?

My motivation is that Balsa does not currently parse HTTP version, so I would need to write that code from scratch, and there is no need to implement something more complicated than absolutely necessary.

Thanks.

KBaichoo (approved):
looking at how those were used by the codec layer, I think this works. Thanks!

PR #16946, opened 2021-06-11
TSAN: fix data race when reading a histogram's sample count

rgs1:
Commit Message: Fix TSAN issue caused from tests when reading histograms.

Additional Description:
This was introduced by #15908. We need to read a histogram's from
the main thread to avoid data races between writes (from the main
thread) and reads (from a worker thread).

Risk Level: Low.
Testing: Unit tests + TSAN should pass.
Docs Changes: NA
Release Notes: NA
Platform Specific Features: NA

Signed-off-by: Raul Gutierrez Segales <rgs@pinterest.com>

PR #21446, opened 2022-05-25
Use absl::Notification when it makes sense

rgs1:
I had to run some code in the main thread for a test recently
and the first example I bumped into wasn't using absl::Notification,
so let's port them over.

Signed-off-by: Raul Gutierrez Segales <rgs@pinterest.com>


rgs1:
cc: @JuniorHsu

rgs1:
This is a follow-up to #16946.

cc: @antoniovicente

KBaichoo:
CI failures seem legit:
`//test/common/event:dispatcher_impl_test                        FAILED TO BUILD`

PR #21451, opened 2022-05-26
skywalking: fix to avoid tracing all requests

wbpcode:
Commit Message: skywalking: fix to avoid tracing all requests
Additional Description:

Current skywalking tracer will tracing all the requests by default and igore the sampling decision of Envoy. 
But in fact, if there is no downstream tracing context (propagation header) is provided and samping decision of Envoy is false, the tracer should do nothing.

Risk Level: Low.
Testing: Updated.
Docs Changes: n/a.
Release Notes: n/a.


KBaichoo:
friendly ping @lizan PTAL

wbpcode:
friendly ping @lizan PTAL

lizan (approved):

PR #21459, opened 2022-05-26
bazel: add config_setting for android

keith:
This is usable in selects for any android target.


alyssawilk (approved):
Most excellent, thanks both for the quick response and knowing exactly the 6 magic lines I wanted :-)

alyssawilk:
(though fwiw I only need it in envoy mobile - it may be useful upstream so fine to land it here)

PR #21460, opened 2022-05-26
quiche: Switch QUICHE flag handling to atomics instead of Mutexes

RyanTheOptimist:
quiche: Switch QUICHE flag handling to atomics instead of Mutexes
Resolves some lock contention issues in HTTP/3 code

Risk Level: Low
Testing: Existing tests
Docs Changes: N/A
Release Notes: N/A


RyanTheOptimist:
@alyssawilk this had a 3x improvement on the internal testing max QPS. I'll also investigate switching over to absl flags though I think it would be good to get this in first.

source/common/quic/platform/quiche_flags_impl.h
```diff
@@ -76,55 +77,47 @@ class Flag {
...
-    absl::MutexLock lock(&mutex_);
-    has_reloaded_value_ = false;
+    value_.store(explicit_value_, std::memory_order_relaxed);
   }
 
 private:
-  mutable absl::Mutex mutex_;
-  T value_ ABSL_GUARDED_BY(mutex_);
+  std::atomic<T> value_;
+  std::atomic<T> explicit_value_;
```
alyssawilk, line 112: Can we comment these up?  Alternately if using absl flags works, we might just want to migrate straight over :-)
/wait

RyanTheOptimist: Added comments. I'm working on the absl flag approach now but it's going to take a bit of work to wrap the absl::Flags in the QUICHE flag api, and refactor things accordingly. Once that's working I'll work with the internal folks to get a max QPS load test and validate the performance properties. There is a comment about `GetFlag()` which says not to use it in tight loops, so I'm very curious to understand the perf impact.

https://github.com/abseil/abseil-cpp/blob/master/absl/flags/flag.h#L84

RyanTheOptimist: I've converted this to a draft and will revisit after we address the various other flag cleanup/refactor PRs get sorted out.

RyanTheOptimist:
@danzh2010 As discussed offline, I think this is ready for review. We'll go ahead and land this now. Once the full stack of flag refactors are pulled into our internal repo we can perf test against an absl::Flag based implementation, as per Alyssa's request.

danzh2010 (approved):
Nice!

KBaichoo (approved):
lgtm thanks. Perhaps we could get away with a weaker memory order on the load but I'm not an expert there

RyanTheOptimist:
> lgtm thanks. Perhaps we could get away with a weaker memory order on the load but I'm not an expert there

Thanks for the review! I think relaxed is the weakest order but I'm also not an expert :)

KBaichoo:
the load is in sequential consistency

PR #21417, opened 2022-05-24
tls: support async cert validation

danzh2010:
Commit Message: change TLS transport socket to use SSL_CTX_set_custom_verify() instead of SSL_CTX_set_verify() and change CertValidator interface to support async cert validation. Also change EnvoyQuicCertVerifier to use the new async interfaces.

This change is needed for https://github.com/envoyproxy/envoy-mobile/issues/1575. Envoy Mobile allows certificates to be verified by the OS-provided certificate verifier. And these verification can be very slow and so blocking the network thread while the verification happens is not an option. Instead, the verification should be performed asynchronously on a different thread. (This is how the cert verification works in Chrome, which is what we're modeling this implementation on).

Risk Level: high, change boring SSL interface used
Testing: added new unit tests and integration tests
Docs Changes: release note
Release Notes: documented tls transport changes.
Runtime guard: envoy.reloadable_features.tls_async_cert_validation

PR #21462, opened 2022-05-26
quic: add Envoy specific quic::ProofVerifyContext implementation

danzh2010:
Commit Message: add EnvoyQuicProofVerifyContext to wrap interesting properties of the connection to be used for cert verification. And instantiate it during client crypto stream creation.

This PR is a preparation for https://github.com/envoyproxy/envoy/pull/21417

Risk Level: low, the context is not used yet.
Testing: existing tests pass
Docs Changes: N/A
Release Notes: N/A


source/common/quic/envoy_quic_proof_verifier.cc
```diff
@@ -15,6 +15,14 @@ quic::QuicAsyncStatus EnvoyQuicProofVerifier::VerifyCertChain(
     std::unique_ptr<quic::ProofVerifyDetails>* details, uint8_t* out_alert,
     std::unique_ptr<quic::ProofVerifierCallback> callback) {
   ASSERT(details != nullptr);
+  ASSERT(!certs.empty());
+  auto* verify_context = dynamic_cast<const EnvoyQuicProofVerifyContext*>(context);
+  if (verify_context == nullptr) {
+    IS_ENVOY_BUG("QUIC proof verify context was not setup correctly.");
+    return quic::QUIC_FAILURE;
+  }
+  ENVOY_BUG(!verify_context->isServer(), "QUIC client cert is not supported yet");
```
RyanTheOptimist, line 24: nit: "Client certificates are not supported in QUIC yet."

source/common/quic/envoy_quic_proof_verifier.h
```diff
@@ -18,6 +18,36 @@ class CertVerifyResult : public quic::ProofVerifyDetails {
   bool is_valid_{false};
 };
 
+// An interface for the Envoy specific QUICHE verify context.
```
RyanTheOptimist, line 21: nit: QUICHE => QUIC, I think, to match the class name and since this is QUIC specific.

test/common/quic/envoy_quic_proof_verifier_test.cc
```diff
@@ -90,6 +95,9 @@ class EnvoyQuicProofVerifierTest : public testing::Test {
   Ssl::MockCertificateValidationContextConfig cert_validation_ctx_config_;
   std::unique_ptr<EnvoyQuicProofVerifier> verifier_;
   NiceMock<Ssl::MockContextManager> tls_context_manager_;
+  Event::MockDispatcher dispatcher_;
+  Network::TransportSocketOptionsConstSharedPtr transport_socket_options_;
+  NiceMock<MockProofVerifyContext> verify_context_;
```
RyanTheOptimist, line 100: Can we also add simple unit tests in this file which verify that the getters for EnvoyQuicProofVerifyContextImpl return the correct values? (If for no other reason than to keep coverage high :>)

RyanTheOptimist (commented):
Nice

test/common/quic/envoy_quic_proof_verifier_test.cc
danzh2010: This test uses mocked verify context, so I wasn't able to test the real impl here. But I added a unit test in client session.

RyanTheOptimist (dismissed):
Looks great! Any idea what's up with CI?

RyanTheOptimist (approved):

danzh2010:
Ping for merge?

Issue #12274, opened 2020-07-24
allow typed_per_filter_config to refer to "custom" HTTP filter names


mpuncel:
Use case: We'd like to use multiple ext_authz filters configured with different gRPC servers, and would like to be able to toggle them separately via route config. This means we can't just use the canonical filter name, but need to be able to refer to custom names.

https://github.com/envoyproxy/envoy/pull/9618 made Envoy able to infer most extension types from the type URL of the config, which frees up users to use the name field to make filter names more descriptive (particularly useful for distinguishing multiple instances of the same filter).

It looks like typed_per_filter_config doesn't yet have this capability and requires that the canonical filter name is used.

I'm not super familiar with the code or filter registration system, but it seems like this could be done by either:
1) keeping a map of filter name to type, and looking up the supplied name in that map to figure out the filter config type
2) use the type URL of the config to infer the filter type
3) both (with a precedence order for applying 1 and 2).

Probably just 1) makes sense? that would ensure that there can't be multiple configs that map to the same filter.

It also might require a config option somewhere backwards compatible? There could be a working config today where a filter has a custom name but the typed_per_filter_config is using the old name.

lizan:
To be consistent with #9618, I think 2 is the way to go, if the config isn't typed (i.e. google.protobuf.Struct), then it should be inferred from name.

mpuncel:
Looking at this more closely, I think I misunderstood what needs to happen in my original comment.

It looks like the issue is that the individual HTTP filters (ext_authz used here as an example) get their per route config like this;

```      Http::Utility::resolveMostSpecificPerFilterConfig<FilterConfigPerRoute>(
          HttpFilterNames::get().ExtAuthorization, route);
```

It's using the well-known name there. Knowing this, I think the solution should be to make each filter know its user-specified name from the `http_filters` list in HCM. Then the filter can try to resolve config using this name instead.

I'm not sure how to handle backwards compatibility for this, since users may have started using user supplied filter names but continue to rely on per route config using the well known filter names, I'm open to suggestions. I'd love to do something like "try to resolve with the user supplied name, and if that is a miss then use the canonical name", but I think that could have strange behavior because for some routes you might hit on one name, and on other routes you might hit on the other.

htuch:
Agree with the above, we should use the opaque filter name in the HTTP filter chain as the key to join with the per-route overrides. I think this makes most sense. @fuqianggao @markdroth

fuqianggao:
[...]

PR #21466, opened 2022-05-27
api: new policy field for per filter config

wbpcode:
Commit Message: api: new policy field for per filter config
Additional Description:

API only PR. This API is used to configure a policy that determine which name or value will be used as searching key of `typed_per_filter_config`.

**It's just a initial idea. So it's open for suggestions and discussions.**

Ref #12274 for more detailed context.

Risk Level: n/a.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.

### DIFF
--- a/api/envoy/config/route/v3/route.proto
+++ b/api/envoy/config/route/v3/route.proto
@@ -135,6 +135,10 @@ message RouteConfiguration {
   // NOTE: this option will not strip the port number (if any) contained in route config
   // :ref:`envoy_v3_api_msg_config.route.v3.VirtualHost`.domains field.
   bool ignore_port_in_host_matching = 14;
+
+  // The policy that used to determine which name or value will be used as the searching key to get per filter config.
+  // [#not-implemented-hide:]
+  FilterConfigKeyPolicy filter_config_key_policy = 15;
 }
 
 message Vhds {
--- a/api/envoy/config/route/v3/route_components.proto
+++ b/api/envoy/config/route/v3/route_components.proto
@@ -15,6 +15,7 @@ import "envoy/type/v3/range.proto";
 
 import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
+import "google/protobuf/empty.proto";
 import "google/protobuf/wrappers.proto";
 
 import "xds/annotations/v3/status.proto";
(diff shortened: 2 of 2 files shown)


wbpcode:
I personally still think that explicit configuration would be more clear. Give users the ability to control this searching behaviour. And this also keep the possibility to create more flexible searching policy in the future (e.g., get route level filter config by authentication/user info header).

api/envoy/config/route/v3/route_components.proto
```diff
@@ -2228,3 +2229,31 @@ message FilterConfig {
...
+    google.protobuf.Empty filter_name_key = 2;
+    // :ref:`custom filter config names
+    // <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpFilter.name>` will be used
+    // as the searching key of per filter config.
+    google.protobuf.Empty custom_name_key = 3;
+
+    // Extendable custom provider that will used to generate a searching key from custom filter config names,
+    // canonical filter names, or even request headers.
+    // [#not-implemented-hide:]
+    core.v3.TypedExtensionConfig custom_provider = 4;
```
soulxu, line 2257: Do we have a use-case for this?

Also If both `filter_name_key ` and `fallback_policy`'s primary goal is backward compatibility, then I didn't see the value of them except for adding more complexity for the user. Then I would prefer to use the runtime key or only add the fallback logic, then wait for the end-user to move to the final target in the future.

wbpcode: > Do we have a use-case for this?

A simple scene is per user/tenant route level config. We can get user/tenant info  after authorization/authentication, then select a route level fitler config based on the filter name and user/tenant info.

soulxu: another solution is we use the fallback logic, and disable searching by name in the future. And for `custom_provider`, it can be a separate story. Just not sure `filter_name_key` and `custom_name_key` is worth or not.

wbpcode: I think it would be very hard to disable serching by filter name completely. Too much users will be effected.

wbpcode: What I think is that this field can be ignored by most of normal users. It will not bring any new complexity to these users.

And for some proficient users, this field can provides complete control to the searching behaviour.

api/envoy/config/route/v3/route.proto
```diff
@@ -135,6 +135,10 @@ message RouteConfiguration {
   // NOTE: this option will not strip the port number (if any) contained in route config
   // :ref:`envoy_v3_api_msg_config.route.v3.VirtualHost`.domains field.
   bool ignore_port_in_host_matching = 14;
+
+  // The policy that used to determine which name or value will be used as the searching key to get per filter config.
+  // [#not-implemented-hide:]
+  FilterConfigKeyPolicy filter_config_key_policy = 15;
```
mattklein123, line 141: I didn't fully track the conversation that you were having with @markdroth, but I would really rather not add new configuration if we don't need to. Can we sketch out what a fallback policy would look like and see if that will cover known use cases? We can add config later if we really need to. Thank you!

/wait

wbpcode:
@mattklein123 

If the fallback policy is applied, as the comments of the API, when we try to get route level filter config of a filter at runtime, the custom config name (HttpFilter.name) will be used first to search config from the typed_per_filter_config. And if we get nothing by the custom config name, then we will use the canonical filter name try again.

For most of current users, one time additional hash searching for every filter is necessary.
But it will not break most of current users except they use one filter's canonical filter name as another filter's custom config name. This exception is highly unlikely to happen.


So, yeah, fallback policy can covers most of known cases. But I personally prefer the explict configuration rather than implicit fallback. 😂 I always want everything can be controlled by the configuration.

wbpcode:
Of course, #8659 also a reason of that why I  add this field.

mattklein123:
Can we add explicit config later if it's really needed? Any config we add adds cost and it's not clear to me that there is enough benefit here? WDYT?

/wait

wbpcode:
> I hear that you are saying that you prefer this behavior to be explicitly configurable, but I'm not really hearing you give a compelling reason for that preference.

Part of the reason for this is my personal preference. Part of the reason is the duplication of searches. The last part is because I fear we may never be able to remove the legacy behaviour.

But since we all agree that the current configuration is too heavy, I will respect the community's opinion.

markdroth:
> Part of the reason is the duplication of searches. The last part is because I fear we may never be able to remove the legacy behaviour.

Wouldn't the runtime knob I suggested address those two concerns?

wbpcode:
> Wouldn't the runtime knob I suggested address those two concerns?

🤔 If we can never remove this legacy behaviour, then we need keep the runtime knob forever. I usually use r

wbpcode:
untime

wbpcode:
convert this to draft but keep it open for #8659.

PR #21477, opened 2022-05-27
fix: add envoy demo missing connect_timeout

mkungla:
I assume that it would be nice that very first example in docs would work :)

Docs Changes: `envoy-demo.yaml`


phlax:
@mkungla please make sure that the version of Envoy that you are using matches the documentation that you are following 

`connect_timeout` is no longer required

PR #21488, opened 2022-05-27
quiche: Make QUICHE reloadable flag behavior match runtime guard behavior

RyanTheOptimist:
quiche: Make QUICHE reloadable flag behavior match runtime guard behavior

Currently QUICHE reloadable flags have curious behavior. When reloading a new snapshot, any reloadable flags which are not present in the snapshot have their values reverted to their default values. However, runtime guards which are not present in the snapshot do not have their values changed. This PR changes QUICHE reloadable flag behavior to match 

Risk Level: Medium
Testing: New unit tests
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/quic/platform/quiche_flags_impl.cc
+++ b/source/common/quic/platform/quiche_flags_impl.cc
@@ -56,8 +56,8 @@ FlagRegistry& FlagRegistry::getInstance() {
 FlagRegistry::FlagRegistry() : flags_(makeFlagMap()) {}
 
 void FlagRegistry::resetFlags() const {
-  for (auto& kv : flags_) {
-    kv.second->resetValue();
+  for (auto& [flag_name, flag] : flags_) {
+    flag->resetValue();
   }
 }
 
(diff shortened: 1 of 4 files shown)


RyanTheOptimist:
@alyssawilk: @danzh2010 and I reviewed the quiche reloadable flag logic today and both were surprised by the "revert to previous default if not in the snapshot" logic, since this is different from other runtime guards. I'm not sure if the current QUICHE behavior is "correct" or just "complicated", but hopefully you have some insight here. If it's not necessary, it would be a helpful simplification of the QUICHE flags and would make it that much easier to get absl::Flag wired up.

yanavlasov (approved):

yanavlasov:
The change looks good to me, but I have no idea what consequences this behavioral change would make. I'll leave it to @alyssawilk  for final review.

RyanTheOptimist:
Discussed this offline with @alyssawilk (who is out sick, but popped into chat). She concluded: "awesome, make it so 🙂" So I think we're good to go.

danzh2010 (approved):
Nice! Thanks for confirming with Alyssa and making the fix! LGTM other than that the PR description seems to be cut?

PR #21489, opened 2022-05-27
quiche: Remove FlagRegistry::resetFlags()

RyanTheOptimist:
quiche: Remove FlagRegistry::resetFlags() and instead implement a test-only QuicheFlagSaverImpl
Also change the FlagRegistry to only contain a map of reloadable flags.

Risk Level: Low
Testing: New unit tests
Docs Changes: N/A
Release Notes: N/A


danzh2010 (approved):
Nice! only one nit

source/common/quic/platform/quiche_flags_impl.h
```diff
@@ -104,6 +70,24 @@ template <typename T> class TypedFlag : public Flag {
...
+  void updateReloadableFlags(const absl::flat_hash_map<std::string, bool>& quiche_flags_override);
+
+private:
+  FlagRegistry();
+
+  const absl::flat_hash_map<absl::string_view, ReloadableFlag*> reloadable_flags_;
+};
+
 // Flag declarations
 #define QUIC_FLAG(flag, ...) extern TypedFlag<bool>* flag;
```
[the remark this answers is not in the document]
RyanTheOptimist: Oh, good point! Done.

danzh2010 (approved):

test/common/quic/platform/quic_platform_test.cc
```diff
@@ -474,33 +475,34 @@ TEST_F(QuicPlatformTest, MonotonicityWithFakeEpollClock) {
...
+    quiche::test::QuicheFlagSaver saver;
+    EXPECT_FALSE(GetQuicReloadableFlag(quic_testonly_default_false));
+    EXPECT_TRUE(GetQuicReloadableFlag(quic_testonly_default_true));
+    SetQuicReloadableFlag(quic_testonly_default_false, true);
+    EXPECT_TRUE(GetQuicReloadableFlag(quic_testonly_default_false));
+
+    EXPECT_FALSE(GetQuicRestartFlag(quic_testonly_default_false));
+    EXPECT_TRUE(GetQuicRestartFlag(quic_testonly_default_true));
+    SetQuicRestartFlag(quic_testonly_default_false, true);
+    EXPECT_TRUE(GetQuicRestartFlag(quic_testonly_default_false));
```
KBaichoo, line 491: seems like this code is just a copy of above, from surrounding context I would have assumed this would have set the `quic_testonly_default_true` to false s.t. it gets restored to true when we leave the scope of flag saver.

KBaichoo: Good catch.

I think we're not checking the `FLAGS_quic_reloadable_flag_quic_testonly_default_false` flag then afterwards as we set that true in the second block. but we check if 

`EXPECT_TRUE(GetQuicRestartFlag(quic_testonly_default_true));` is true which we didn't change in this process

RyanTheOptimist: Oh, I see! The EXPECT on line 500 should be changed. Thanks, nice catch!

Done.

RyanTheOptimist:
> lgtm pending CI, thanks!

Thanks! And thanks for the detailed review!

PR #16344, opened 2021-05-05
Matching engine: add a cookie matching input object

aguinet:
Commit Message:  Matching: add HttpRequestCookieMatchInput data input
Additional Description:

This allows matching data extracted from cookies. It also adds a new parseCookieValues API in Http::Utility, which allows to return the potential multiple values that could be set in malicious HTTP requests, like in this case:

Cookie: mykey=a;mykey=b

To be in pair with what is already done with the HTTP header input matcher, this input matcher returns, in this case, `a,b`.

I've done only the request side, but if these patches are fine for you, I can also add the response side, and factorize the code in a similar way that what's done for headers!

Risk Level: Low
Testing: adapting current unit tests
Docs Changes: New API in Http::Testing documented
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #21413, opened 2022-05-23
registry: populate type_urls for extensions

kyessenov:
Commit Message: Implement type descriptor info in the node info. Extensions can have multiple type URLs (due to route-specific protos). Increase in raw JSON for server_info is 44k->57k. This is needed to document a mapping from protos to extensions which is absent.
Additional Description:
Risk Level: low (more bytes per first xDS request)
Testing: in progress
Docs Changes: none
Release Notes: none

PR #21496, opened 2022-05-29
 Matching engine: add a cookie matching input object (revival)

aguinet:
(Revival of https://github.com/envoyproxy/envoy/pull/16344)

Commit Message: Matching: add HttpRequestCookieMatchInput data input
Additional Description:

This allows matching data extracted from cookies. It also adds a new parseCookieValues API in Http::Utility, which allows to return the potential multiple values that could be set in malicious HTTP requests, like in this case:

Cookie: mykey=a;mykey=b

To be in pair with what is already done with the HTTP header input matcher, this input matcher returns, in this case, a,b.

I've done only the request side, but if these patches are fine for you, I can also add the response side, and factorize the code in a similar way that what's done for headers!

Risk Level: Low
Testing: adapting current unit tests
Docs Changes: New API in Http::Testing documented
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/http/utility.cc
```diff
@@ -329,6 +329,29 @@ Utility::parseCookies(const RequestHeaderMap& headers,
...
+                                                                     const size_t max_vals) {
+  absl::InlinedVector<absl::string_view, 2> ret;
+
+  forEachCookie(headers, Http::Headers::get().Cookie,
+                [&ret, &key, &max_vals](absl::string_view k, absl::string_view v) -> bool {
+                  if (k == key) {
+                    ret.push_back(v);
+                    // max_vals == 0 => infinity, so the condition above will never be true
+                    // in this case.
+                    if (ret.size() == max_vals) {
```
snowp, line 343: If we default to std::numeric_limits::max instead of treating 0 as special I think this logic would seem a bit cleaner

source/common/http/utility.h
```diff
@@ -300,9 +301,25 @@ absl::flat_hash_map<std::string, std::string> parseCookies(const RequestHeaderMa
  * Parse a particular value out of a set-cookie
  * @param headers supplies the headers to get the set-cookie from.
  * @param key the key for the particular set-cookie value to return
- * @return std::string the parsed set-cookie value, or "" if none exists
+ * @return absl::string_view the parsed set-cookie value, or "" if none exists
```
snowp, line 304: Here and below please document the lifetime of the returned string views 

More generally I'd see if there are any concerns about this interface switch, given that a std::string ends up being a bit safer than a string_view, cc @alyssawilk @mattklein123

snowp (changes requested):
Overall this seems good to me, main question I have is around the common code interface changes

source/common/http/utility.h
aguinet: Maybe the safest option for now is indeed to keep `std::string`. I will revert back to that API.

source/common/http/utility.cc
aguinet: Done in https://github.com/envoyproxy/envoy/pull/21496/commits/b7f34796cf5bbe216a159745a0a86cc49a0d2268

source/common/http/utility.h
aguinet: Done in https://github.com/envoyproxy/envoy/pull/21496/commits/b7f34796cf5bbe216a159745a0a86cc49a0d2268

api/envoy/type/matcher/v3/http_inputs.proto
```diff
@@ -60,3 +60,17 @@ message HttpResponseTrailerMatchInput {
   string header_name = 1
       [(validate.rules).string = {well_known_regex: HTTP_HEADER_NAME strict: false}];
 }
+
+// Match input indicates that matching should be done on a specific cookie key.
```
mattklein123, line 64: Please add a release note for this new feature. @kyessenov what do we need to do to get this matcher to show up in the docs? I forget where we landed on this in terms of what is an extension category, etc.

mattklein123 (changes requested):
API LGTM with small comment, thanks.

/wait

api/envoy/type/matcher/v3/http_inputs.proto
kyessenov: 1. Add the new extension to the list of extensions in `extension_manifest.yaml`. Tag it as implementing the HTTP matcher category.
2. Add a comment to the proto as in `// [#extension: envoy.matching.inputs.cookies]`
3. Reference the extension here: https://github.com/envoyproxy/envoy/blob/main/docs/root/intro/arch_overview/advanced/matching/matching_api.rst#http-input-functions.

I'm working to get (1)-(2) validated once we merge https://github.com/envoyproxy/envoy/pull/21413.

aguinet: Thanks @kyessenov for this! Done in https://github.com/envoyproxy/envoy/pull/21496/commits/3930b3ab6fa75cfc523995bffd031f4f3a93e70f

tzuryby:
@aguinet can you take a quick look what else is missing to get it in?

Issue #21497, opened 2022-05-30
contrib: std::exception can't be catched correctly with libc++


durd07:
*Title*: *contrib: std::exception can't be catched correctly with libc++*

*Description*:
When compile envoy-contrib with clang and libc++ with following command:
```
./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.dev.contrib'
```
which actually build like with this command:
```
./bazel/setup_clang.sh /opt/llvm
bazel build --config=libc++ //contrib/exe:envoy
```
or even in official envoy-contrib image `envoyproxy/envoy-contrib:v1.22.0`

The following code doesn't work, if `some_code_with_throw` throw some error, it can't be catched by `catch` and `std::terminate()` will be invoked with following function backtrace
``` 
try {
    some_code_with_throw
    ...
} catch (const std::exception& e) {
    ...
} 
```

function backtrace:
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737087567424) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737087567424) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737087567424, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7ab4476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7a9a7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x000055556066a218 in Envoy::TerminateHandler::logOnTerminate() const::$_0::operator()() const (this=0x55556066a010 <Envoy::TerminateHandler::logOnTerminate() const::$_0::__invoke()>) at source/exe/terminate_handler.cc:14
#6  0x000055556066a019 in Envoy::TerminateHandler::logOnTerminate() const::$_0::__invoke() () at source/exe/terminate_handler.cc:11
#7  0x0000555563054b63 in std::__terminate(void (*)()) ()
#8  0x0000555563054246 in __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) ()
#9  0x00005555630541e0 in __cxa_throw ()
#10 0x000055555b371286 in Envoy::Extensions::NetworkFilters::Common::Redis::DecoderImpl::parseSlice (this=0x7917f46c480, slice=...) at source/extensions/filters/network/common/redis/codec_impl.cc:385
[...]

PR #20143, opened 2022-02-28
filter: add language http filter "accept-language"

diceride:
Commit Message:
filter: add a new anguage filter "accept-language"

Additional Description:
The language detection filter (i18n) picks the best match between a client's desired locales and an application's supported locales.

Parses a list of locales from an ``Accept-Language`` header [RFC 2616 Section 14.4](https://tools.ietf.org/html/rfc2616#section-14.4)
and adds a new `x-language` header to the request.

The filter uses [Unicode ICU](https://github.com/unicode-org) for locale parsing.

```txt
    // Multiple types, weighted with the quality value syntax:
    Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, *;q=0.5
```

* The client is from the Romandy region and prefers Swiss French, the variety of French spoken in the French-speaking area of Switzerland
* The client sends an ``Accept-Language`` header
* An ``Accept-Language`` header indicates the natural language and locale that the client prefers
* The application is only configured to match French fr
* The filter sets the value of the ``x-language`` header to fr

The language detection filter outputs statistics in the *http.<stat_prefix>.language.* namespace.

Risk Level:
Low

Testing:
Adds integration tests

Docs Changes:
Adds a new documentation page for the language filter

Release Notes:

Platform Specific Features:

PR #19473, opened 2022-01-11
contrib: add Hyperscan matcher

zhxie:
Signed-off-by: Xie Zhihao zhihao.xie@intel.com

Commit Message: contrib: add Hyperscan matcher
Additional Description:

Hyperscan is a high-performance multiple regex matching library. Hyperscan uses hybrid automata techniques to allow simultaneous matching of large numbers of regular expressions, as well as matching of regular expressions across streams of data.

This commit provides a Hyperscan matcher which can be plugged into the matcher API and replace the use of the default regex matcher in string matcher. Hyperscan supports most of the pattern syntax used by the PCRE and has a performance improvement compared with Google RE2.

The benchmark result on i9-9980HK shows that Hyperscan matcher reaches nearly a 50% performance improvement compared with the default regex matcher.
```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
BM_CompiledGoogleReMatcher        369 ns          368 ns      1963625
BM_HyperscanMatcher               244 ns          244 ns      2846760
```

Risk Level: Low
Testing: Unit and benchmark
Docs Changes: API interface
Release Notes: WIP
Platform Specific Features: Requires processor with SSSE3 support (nearly any modern x86 processor)
Fixes #19472

PR #21499, opened 2022-05-30
contrib: std::exception can't be catched correctly with libc++

durd07:
Signed-off-by: Felix Du <felix.du@nokia-sbell.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] https://github.com/envoyproxy/envoy/issues/21497
[Optional Deprecated:]

### DIFF
--- a/bazel/foreign_cc/BUILD
+++ b/bazel/foreign_cc/BUILD
@@ -150,7 +150,6 @@ configure_make(
         "ICU_DATA_FILTER_FILE": "$(execpath //bazel/foreign_cc:icu_data_filter.json)",
     },
     lib_source = "@com_github_unicode_org_icu//:all",
-    linkopts = ["-lstdc++"],
     out_static_libs = [
         "libicuuc.a",
         "libicudata.a",


daixiang0:
The PR title confuse me: is it only related to contrib part? And what is the problem now, remove it works for all cases?

durd07:
> The PR title confuse me: is it only related to contrib part? And what is the problem now, remove it works for all cases?

is it only related to contrib part?
--> yes
And what is the problem now,
--> coredump occured if there is some code throw exceptions
remove it works for all cases
--> I don't know how to cover all cases and what does the all cases mean? But all checks have passed.

Details please refer to this issue: https://github.com/envoyproxy/envoy/issues/21497

bazel/foreign_cc/BUILD
```diff
@@ -150,7 +150,6 @@ configure_make(
         "ICU_DATA_FILTER_FILE": "$(execpath //bazel/foreign_cc:icu_data_filter.json)",
     },
     lib_source = "@com_github_unicode_org_icu//:all",
-    linkopts = ["-lstdc++"],
```
KBaichoo, line 153: wonder if this should have this in the CXXFLAGs as in:

`CXXFLAGS": "-lstdc++` used for ragel. That seems to have worked in this case: https://github.com/envoyproxy/envoy/pull/19473#discussion_r790427535 using the `ENVOY_STDLIB` override.

Will defer to the expert (Lizan) in this area.

lizan:
Is there anyway to test/verify this so we don't get regression?

durd07:
@diceride Could you please help to verify whether this PR will block your commit https://github.com/envoyproxy/envoy/pull/20143?

@lizan Even though I don't know how to verify this PR will affect features involved from PR https://github.com/envoyproxy/envoy/pull/20143(need owner's help to verify), There is obvious error that both link libc++ and libstdc++ at the same time. Actually PR20143 has already block all contrib extensions. 

At least compile is OK and all CI pipelines are passed.

durd07:
> What do you mean by "block all contrib extensions"? Their tests passes in main branch and here.

Yes, As I described in the issue https://github.com/envoyproxy/envoy/issues/21497, **All try/catch** in the contrib image doesn't work.

> Their tests passes in main branch and here.
That's right, but there is no cases to touch the `try/catch` exceptions. 

So I think the solution is merge my PR ASAP or revert the origianl PR https://github.com/envoyproxy/envoy/pull/20143

lizan:
@durd07 Can you port this test for contrib? https://github.com/envoyproxy/envoy/blob/main/test/exe/envoy_static_test.sh

lizan (approved):

durd07:
@KBaichoo @lizan Could you help to merge this PR?

durd07:
> @durd07 Can you port this test for contrib? https://github.com/envoyproxy/envoy/blob/main/test/exe/envoy_static_test.sh

I can try to do it later, but seems there is no such framework for contrib, we can discuss how to port some test of the official binary into contrib. There is less of quality check of contrib.

bazel/foreign_cc/BUILD
zhxie: I think if the `configure_make` does not produce any binaries like ragel, then it is safe to remove `-lstdc++`.

durd07: @zhxie It is true that there won't be any other binaries generated besides envoy-static, so it is safe to remove `-lstdc++`.

PR #21507, opened 2022-05-30
customized_envoy_docker_build_dir_support

xu1zhou:
Commit Message: support pass default_envoy_docker_build_dir as enviroment variable when build envoy with docker
Additional Description: customized docker build directory is useful for a shared host witl multiple developer
Platform Specific Features: Linux


xu1zhou:
@phlax Is that ok to review this PR

ci/run_envoy_docker.sh
```diff
@@ -37,7 +37,7 @@ else
   ENVOY_DOCKER_OPTIONS+=(-u root:root)
   ENVOY_DOCKER_OPTIONS+=(-v /var/run/docker.sock:/var/run/docker.sock)
   ENVOY_DOCKER_OPTIONS+=(--cap-add SYS_PTRACE --cap-add NET_RAW --cap-add NET_ADMIN)
-  DEFAULT_ENVOY_DOCKER_BUILD_DIR=/tmp/envoy-docker-build
+  DEFAULT_ENVOY_DOCKER_BUILD_DIR=${defualt_envoy_docker_build_dir:=/tmp/envoy-docker-build}
```
daixiang0, line 40: `${DEFAULT_ENVOY_DOCKER_BUILD_DIR:=xxx}`

xu1zhou: Thanks for your suggestion. Using lowercase to align with other env defined in this script like 
`export HTTP_PROXY="${http_proxy:-}"`

phlax: please use UPPER_CASE and fix the spelling (s/defualt/default/)

`http_proxy` is lower case for a reason - and it is exceptional - global vars in bash files are usually upper case

the other thing im wondering about this change - whether it is the best way to achieve what you are trying to do

surely you just need to set `ENVOY_DOCKER_BUILD_DIR` - if this env var is set it is used to construct the container, whereas `DEFAULT_ENVOY_DOCKER_BUILD_DIR` is just the default, is overridden by the former var, and is not used or passed through anywhere itself

xu1zhou: > please use UPPER_CASE and fix the spelling (s/defualt/default/)
> 
> `http_proxy` is lower case for a reason - and it is exceptional - global vars in bash files are usually upper case
> 
> the other thing im wondering about this change - whether it is the best way to achieve what you are trying to do
> 
> surely you just need to set `ENVOY_DOCKER_BUILD_DIR` - if this env var is set it is used to construct the container, whereas `DEFAULT_ENVOY_DOCKER_BUILD_DIR` is just the default, is overridden by the former var, and is not used or passed through anywhere itself
ENVOY_DOCKER_BUILD_DIR works fine for me, thanks.  
In fact, the problem I want to solve is:  
The permisstion of files within $DEFAULT_ENVOY_DOCKER_BUILD_DIR will stay the same with last build user. In that case, build from other user will be blocked by "permission denied". Maybe I can add a command to change owner of last build remnant or just remove them.

phlax: @xu1zhou can we close this - i think the change is not necessary/desirable

 not sure re perms - my understanding is that using `./ci/run_envoy_docker sh 'ci/do_ci.sh somecommand'`will preserve permissions correctly

i dont use this codepath myself too much - so not sure, but, either way i think we dont need to change this env var

xu1zhou:
Will create any PR to resolve problem

PR #21515, opened 2022-05-31
tls/handshaker: extend HandshakerFactoryContext with singletonManager

belyalov:
Commit Message: tls/handshaker: extend HandshakerFactoryContext with singletonManager
Additional Description: Allow Handshaker extension to access singleton manager
Risk Level: No Risk
Testing: No tests added, existing mock updated
Docs Changes: No
Release Notes: No

Sometimes it is required to have access to some sort of "global cache" from TLS handshaker extension. So extending `HandshakerFactoryContext` with singletonManager property.

### DIFF
--- a/envoy/ssl/handshaker.h
+++ b/envoy/ssl/handshaker.h
@@ -6,6 +6,7 @@
 #include "envoy/network/post_io_action.h"
 #include "envoy/protobuf/message_validator.h"
 #include "envoy/server/options.h"
+#include "envoy/singleton/manager.h"
 
 #include "openssl/ssl.h"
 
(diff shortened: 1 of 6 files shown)


belyalov:
Please let me know if there is any other place to add / update tests.

ggreenway:
What's the intended usage of this? Is it for something out-of-tree, or something that will be added in the future?

belyalov:
> What's the intended usage of this? Is it for something out-of-tree, or something that will be added in the future?

The intended usage is, for example, HTTPS forward proxy (man in the middle) - in this case you usually need:
- generate SSL certificates on the fly (with domain matched SNI, so client will trust)
- cache/re-sign upstream server original certificate (so for next connections present it to the client)

In order to do that, you may want to write custom TLS handshaker which will set proper certificate for client to present / extract certificate from upstream connection and cache it.

While generating certificates on the fly is just doable with existing implementation, making global cache certificates requires having `singletoneManager` in the `handshakerFactory`.

There may be other use cases when you have "something global" and you want to use it during handshake.

belyalov:
As for the future - improved https forward proxy support (e.g. handling certificates part)

ggreenway (approved):
LGTM

alyssawilk:
FWIW without tests and a comment this may end up getting removed at some point
(found it doing dead code clean up but checked git blame and saw it may be used downstream)

PR #21517, opened 2022-05-31
xds spec: add section about PGV annotations

markdroth:
Signed-off-by: Mark D. Roth <roth@google.com>

Commit Message: xds spec: add section about PGV annotations
Additional Description: Describe use of PGV annotations in xDS.
Risk Level: Low
Testing: N/A
Docs Changes: Included in PR
Release Notes: N/A
Platform Specific Features:

CC @htuch

### DIFF
--- a/docs/root/api-docs/xds_protocol.rst
+++ b/docs/root/api-docs/xds_protocol.rst
@@ -51,13 +51,19 @@ PGV annotations are not intended to be an exhaustive list of validation checks
 to be performed by the client; clients may reject a resource for reasons
 unrelated to the PGV annotations.
 
-In general, it is not recommended that control planes or xDS proxies directly
-do any validation using the PGV annotations. There may be some rare cases where
-a control plane knows a priori that all of its clients have been compiled with
-the same version of the xDS proto files and are therefore using the same PGV
-annotations, but in general it is not possible for a control plane to
-know that, so any validation that it performs may not wind up matching
-the validation that will actually be done by its clients.
+In general, the PGV annotations are not intended to be used by control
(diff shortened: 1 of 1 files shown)


docs/root/api-docs/xds_protocol.rst
```diff
@@ -37,6 +37,28 @@ appears below, and takes the form ``type.googleapis.com/<resource type>`` -- e.g
...
+(PGV), which indicate semantic constraints to be used to validate the contents
+of a resource when it is received by a client.
+
+Clients are not required to use these PGV annotations to validate the
+resources (e.g., Envoy does this validation, but gRPC does not). Also, the
+PGV annotations are not intended to be an exhaustive list of validation checks
+to be performed by the client; clients may reject a resource for reasons
+unrelated to the PGV annotations.
+
+In general, it is not recommended that control planes or xDS proxies directly
```
htuch, line 54: I'm not sure I'd put it exactly this way. Control planes and xDS proxies shouldn't blindly perform PGV validation at some random version of the API. However, if they have a range of supported clients, they are free to perform PGV validation on this range and use this as a means to sanity check config. In the same way they might actually test out the configuration on the range of clients in a heavier weight config validation.

What is definitely true is that the PGV validations are not intended for consumption by the control plane - the only value is as a method to catch potential rejection from clients.

markdroth: I've updated the wording to attempt to better capture the intent here.  PTAL.

htuch (approved):
Great, makes sense, thanks!

Issue #20536, opened 2022-03-26
Linked packet capture feature provided by hardware Loadbalancer(Netscaler/BigIP)


zhangbo1882:
Current the hardware load balancer (HLB) such as Netscaler/BigIP provide the linked packet capture functionality to make the packet capture more convenient. It mainly solves two pain points when we do packet capture.

If you set the filter with client IP and VIP, it can capture all the packets from client to vip and also from LB to all upstream servers. So we can get a full picture about the whole stream.

Several clients-to-vip connections map to one connection from LB to upstream server. So there should be a way to show the relationship between the packet from client to vip and the packet from LB to backend server.

I have a solution for envoy as a software load balancer (SLB). There are two parts of work. One is for envoy, one is for linux kernel.

For envoy side, the main works is as follows:

Pass the client connection information to the server side connection
Add the client port information as the socket ancillary data when envoy sends packet out.
Add configuration to filter which client/VIP should be captured. ()
For kernel side, need recognize the ancillary data added by envoy and add the client port information as TCP option. So when we capture the packet from LB to upstream server, we can know which client connection the packet is from.

I have done the work for both envoy and kernel side and it works as expected. I just want to confirm with community to see whether this kind of envoy PR can be accepted since the feature need some kernel help. Are there any suggestions for this kind of case which need other software/kernel's co-working.

zhangbo1882:
https://github.com/envoyproxy/envoy/pull/20525

mattklein123:
Sorry I don't think I have enough information to have an opinion on how this might work in Envoy. In general though, we fully decouple upstream and downstream, so we are never going to directly connect them.

The way we typically handle such issues is to pass metadata between downstream and upstream. This might be the way to go for your use case. For example take a look at https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/original_src/original_src.cc and other uses of `addUpstreamSocketOptions`. Similar things could be done with network filters, dynamic metadata, etc.

mattklein123:
> The original_src feature will change the source IP for the traffic from envoy to upstream server. This is NOT what we want.

[...]

PR #21524, opened 2022-06-01
Copy downstream metadata to upstream

zhangbo1882:
https://github.com/envoyproxy/envoy/issues/20536
Add a new dynamic metadata namespace "envoy.common.down_up",  you can add your own dynamic metadata into this namespace if you want to propagate your downstream dynamic metadata to the upstream stream info.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/router/upstream_request.cc
+++ b/source/common/router/upstream_request.cc
@@ -229,11 +229,17 @@ void UpstreamRequest::copyDynamicMetaDataFromL4Downstream(StreamInfo::StreamInfo
     const StreamInfo::StreamInfo& downstream_streamInfo =
         parent_.callbacks()->connection()->streamInfo();
     const auto& dynamic_metadata = downstream_streamInfo.dynamicMetadata();
-    auto name = Envoy::Extensions::NetworkFilters::NetworkFilterNames::get().DownToUp;
+    auto name = Envoy::Extensions::NetworkFilters::NetworkFilterNames::get().SharedWithUpstream;
     auto filter_it = dynamic_metadata.filter_metadata().find(name);
+
     if (filter_it != dynamic_metadata.filter_metadata().end()) {
-      ProtobufWkt::Struct metadata(filter_it->second);
-      l4_stream_info.setDynamicMetadata(name, metadata);
+      auto downstream_connection_id = parent_.callbacks()->connection()->id();
+      ProtobufWkt::Struct upstream_metadata1;
+      auto& fields = *upstream_metadata1.mutable_fields();
+      ProtobufWkt::Value val;
+      val = ValueUtil::structValue(filter_it->second);
+      fields.insert({std::to_string(downstream_connection_id), val});
+      l4_stream_info.setDynamicMetadata(name, upstream_metadata1);
     }
   }
 }
--- a/source/common/router/upstream_request.cc
+++ b/source/common/router/upstream_request.cc
@@ -37,6 +37,7 @@
 #include "source/common/stream_info/uint32_accessor_impl.h"
 #include "source/common/tracing/http_tracer_impl.h"
 #include "source/extensions/common/proxy_protocol/proxy_protocol_header.h"
+#include "source/extensions/filters/network/well_known_names.h"
 
 namespace Envoy {
 namespace Router {
@@ -221,6 +222,20 @@ const RouteEntry& UpstreamRequest::routeEntry() const { return *parent_.routeEnt
 
 const Network::Connection& UpstreamRequest::connection() const {
   return *parent_.callbacks()->connection();
+}
+
+void UpstreamRequest::copyDynamicMetaDataFromDownstream(StreamInfo::StreamInfo& l4_stream_info) {
+  if (parent_.callbacks()->connection() != nullptr) {
+    const StreamInfo::StreamInfo& downstream_streamInfo =
+        parent_.callbacks()->connection()->streamInfo();
+    const auto& dynamic_metadata = downstream_streamInfo.dynamicMetadata();
+    auto name = Envoy::Extensions::NetworkFilters::NetworkFilterNames::get().DownToUp;
+    auto filter_it = dynamic_metadata.filter_metadata().find(name);
+    if (filter_it != dynamic_metadata.filter_metadata().end()) {
+      ProtobufWkt::Struct metadata(filter_it->second);
+      l4_stream_info.setDynamicMetadata(name, metadata);
+    }
+  }
 }
 
 void UpstreamRequest::decodeMetadata(Http::MetadataMapPtr&& metadata_map) {
@@ -526,6 +541,7 @@ void UpstreamRequest::onPoolReady(
         [this]() -> void { onStreamMaxDurationReached(); });
     max_stream_duration_timer_->enableTimer(*max_stream_duration);
   }
+  copyDynamicMetaDataFromDownstream(info);
 
   const Http::Status status =
       upstream_->encodeHeaders(*parent_.downstreamHeaders(), shouldSendEndStream());
--- a/source/common/router/upstream_request.cc
+++ b/source/common/router/upstream_request.cc
@@ -224,7 +224,7 @@ const Network::Connection& UpstreamRequest::connection() const {
   return *parent_.callbacks()->connection();
 }
 
-void UpstreamRequest::copyDynamicMetaDataFromDownstream(StreamInfo::StreamInfo& l4_stream_info) {
+void UpstreamRequest::copyDynamicMetaDataFromL4Downstream(StreamInfo::StreamInfo& l4_stream_info) {
   if (parent_.callbacks()->connection() != nullptr) {
     const StreamInfo::StreamInfo& downstream_streamInfo =
         parent_.callbacks()->connection()->streamInfo();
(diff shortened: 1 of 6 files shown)


yanavlasov:
@zhangbo1882  Please also add unit tests

source/extensions/filters/network/well_known_names.h
```diff
@@ -59,6 +59,8 @@ class NetworkFilterNameValues {
   const std::string ZooKeeperProxy = "envoy.filters.network.zookeeper_proxy";
   // WebAssembly filter
   const std::string Wasm = "envoy.filters.network.wasm";
+  // Copy dynamic metadata from downstream to upstream
+  const std::string DownToUp = "envoy.filters.network.down_up";
```
kyessenov, line 63: This doesn't seem like the right place to specify a name as there is no filter. I think you want more something like shared metadata documented in https://www.envoyproxy.io/docs/envoy/latest/configuration/advanced/well_known_dynamic_metadata.

source/common/router/upstream_request.cc
```diff
@@ -223,6 +224,20 @@ const Network::Connection& UpstreamRequest::connection() const {
   return *parent_.callbacks()->connection();
 }
 
+void UpstreamRequest::copyDynamicMetaDataFromDownstream(StreamInfo::StreamInfo& l4_stream_info) {
```
kyessenov, line 227: I think it'd be OK to have a "convention" over "configuration", but the way it's implemented is too specific to L4->L4 metadata transfer, so that leaves no room to generalize it later. So I see a bunch of options:
* reserve name for specific transfer, be very precise that it's L4 to L4;
* add API to upstream clusters, similar to internal upstream;
* add API to listeners to somehow access upstream metadata from filters;

Thoughts from others?

source/extensions/filters/network/well_known_names.h
zhangbo1882: So where should I put the metadata namespace definition ?

source/common/router/upstream_request.cc
zhangbo1882: I prefer to the first option. Do you have any suggestion about the name ?

kyessenov: Shouldn't this also be done for tcp_proxy and other network routing filters?

source/extensions/filters/network/well_known_names.h
kyessenov: I think something like "envoy.common.shared_network_metadata" might be appropriate? I am not good at naming, but most names are associated with filter names, and the PR is adding a common namespace. I think it's fine to keep it in this file.

source/common/router/upstream_request.cc
zhangbo1882: I think we can focus on the basic case for this PR.

zhangbo1882:
@kyessenov could you help review? Thanks.

source/common/router/upstream_request.cc
```diff
@@ -223,6 +224,20 @@ const Network::Connection& UpstreamRequest::connection() const {
...
 
+void UpstreamRequest::copyDynamicMetaDataFromL4Downstream(StreamInfo::StreamInfo& l4_stream_info) {
+  if (parent_.callbacks()->connection() != nullptr) {
+    const StreamInfo::StreamInfo& downstream_streamInfo =
+        parent_.callbacks()->connection()->streamInfo();
+    const auto& dynamic_metadata = downstream_streamInfo.dynamicMetadata();
+    auto name = Envoy::Extensions::NetworkFilters::NetworkFilterNames::get().DownToUp;
+    auto filter_it = dynamic_metadata.filter_metadata().find(name);
+    if (filter_it != dynamic_metadata.filter_metadata().end()) {
+      ProtobufWkt::Struct metadata(filter_it->second);
```
kyessenov, line 235: This feels like an unnecessary local copy? Why not pass by reference?

source/extensions/filters/network/well_known_names.h
```diff
@@ -59,6 +59,9 @@ class NetworkFilterNameValues {
   const std::string ZooKeeperProxy = "envoy.filters.network.zookeeper_proxy";
   // WebAssembly filter
   const std::string Wasm = "envoy.filters.network.wasm";
+  // A common namespace which will propagated the dynamic metadata from
+  // downstream to upstream
+  const std::string DownToUp = "envoy.common.down_up";
```
kyessenov, line 64: Naming suggestion: `shared_with_upstream`? Would be good to align with the naming choice for the filter state (https://github.com/envoyproxy/envoy/pull/21700/files).

test/common/router/upstream_request_test.cc
```diff
@@ -32,6 +32,29 @@ class UpstreamRequestTest : public testing::Test {
                                     false, true};
 };
 
+TEST_F(UpstreamRequestTest, CopyDynamicMetaData) {
```
kyessenov, line 35: I think we also need some integration test here? Maybe find some integration test in `test/integration` to build from?

source/common/router/upstream_request.cc
```diff
@@ -223,6 +224,20 @@ const Network::Connection& UpstreamRequest::connection() const {
   return *parent_.callbacks()->connection();
 }
 
+void UpstreamRequest::copyDynamicMetaDataFromL4Downstream(StreamInfo::StreamInfo& l4_stream_info) {
```
kyessenov, line 227: Can you explain why this is only implemented for HTTP connection manager and not other network filters? I think it'd make sense for `tcp_proxy` to apply the same logic consistently.

source/common/router/upstream_request.cc
zhangbo1882: downstream and upstream have different lifetime cycle, it is safer to do a local copy.

source/extensions/filters/network/well_known_names.h
zhangbo1882: will change

source/common/router/upstream_request.cc
```diff
@@ -526,6 +541,7 @@ void UpstreamRequest::onPoolReady(
         [this]() -> void { onStreamMaxDurationReached(); });
     max_stream_duration_timer_->enableTimer(*max_stream_duration);
   }
+  copyDynamicMetaDataFromL4Downstream(info);
```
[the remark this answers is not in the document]
zhangbo1882: Will add the downstream connection id as the key when copy the metadata to upstream so that upstream can include all metadata from different downstream.

source/common/router/upstream_request.cc
zhangbo1882: I am not sure whether other filters have the similar use case. I think we can add more in future if necessary. In this PR, only focus on the HCM.

test/common/router/upstream_request_test.cc
zhangbo1882: Now there is no filter which will really add the metadata in new namespace. It seems that we cannot add integration test here unless I change the existing network filter's code or add a new network filter which is not in the scope of this PR.

adisuissa:
@zhangbo1882 please [avoid force-pushing](https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#submitting-a-pr)

@kyessenov is this PR ready for a senior-maintainers round?

source/common/router/upstream_request.cc
```diff
@@ -223,6 +224,26 @@ const Network::Connection& UpstreamRequest::connection() const {
...
+    auto name = Envoy::Extensions::NetworkFilters::NetworkFilterNames::get().SharedWithUpstream;
+    auto filter_it = dynamic_metadata.filter_metadata().find(name);
+
+    if (filter_it != dynamic_metadata.filter_metadata().end()) {
+      auto downstream_connection_id = parent_.callbacks()->connection()->id();
+      ProtobufWkt::Struct upstream_metadata1;
+      auto& fields = *upstream_metadata1.mutable_fields();
+      ProtobufWkt::Value val;
+      val = ValueUtil::structValue(filter_it->second);
+      fields.insert({std::to_string(downstream_connection_id), val});
```
kyessenov, line 241: Does this leak memory when downstream creates many connections through a single upstream connection? Is the upstream filter expected to delete a field?

kyessenov:
@adisuissa @zhangbo1882 Sorry I was on and off the last few days. I think that we still need to resolve sharing of metadata between different downstreams for the same upstream, but we could benefit from senior maintainer on how to design that (e.g. should we use a hash again or just aggregate in a bag like this PR proposes).

phlax:
/wait-any

cc @mattklein123 re design question above

source/common/router/upstream_request.cc
mattklein123: +1 to at minimum doing this for tcp_proxy

source/common/router/upstream_request.cc
zhangbo1882: I will consider more. Thanks for the input.

source/common/router/upstream_request.cc
zhangbo1882: I will add.

zhangbo1882:
in progress

PR #21525, opened 2022-06-01
router: new methods for stream filter callback to get route per filter config

wbpcode:
Commit Message: router: new methods for stream filter callback to get route per filter config
Additional Description:

The final goal is to support searching for route level filter config from `typed_per_filter_config` by the custom filter config name or even dynamic value from request.

Check #12274 and #21466 for complete context.

Regardless of which policy is uesed to select the route config searching key, there is some pre-work need to be addresed for the implementation.

**This initial commit is used as a quick example for discussions.**

Risk Level: Low.
Testing: Waiting.
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
It would be hard that let every filter handle the `filter name`/`custom name` key searching by itself.

IMO, it would be better that using `Http::StreamFilterCallbacks` as a proxy and provides filters a method by the  `Http::StreamFilterCallbacks` to get route config.
The filters need to use the `Http::StreamFilterCallbacks`'s method rather than the method provided by the `Route`.

It means that the HTTP filter should get route config by the way as following code.

```
  // Http::StreamDecoderFilter
  Http::FilterHeadersStatus decodeHeaders(Http::RequestHeaderMap&, bool) override {
    auto route_config = decode_callback_->mostSpecificPerFilterConfig();
  }

```

## How?

First, we need make the `custom name`/`filter name` available for the `Http::StreamFilterCallbacks`. We can add new method to `Http::StreamFilterBase` to provide these name.

```
class StreamFilterBase {
public:
  /**
   * Get custom filter name of current filter instance. This name typically should be the
   * name of HttpFilter proto configuration.
   */
  virtual absl::string_view customName() const { return {}; }

  /**
   * Get filter name of current filter instance.
   */
  virtual absl::string_view filterName() const { return {}; }
};
```

Second, we can add method `mostSpecificPerFilterConfig` and `traversePerFilterConfig` to the  `Http::StreamFilterCallbacks`.

```
class StreamFilterCallbacks {
  /**
   * This is a helper to get the route's per-filter config if it exists, otherwise the virtual
   * host's. Or nullptr if none of them exist.
   */
  virtual const RouteSpecificFilterConfig* mostSpecificPerFilterConfig() const PURE;

  /**
   * Fold all the available per route filter configs, invoking the callback with each config (if
   * it is present). Iteration of the configs is in order of specificity. That means that the
   * callback will be called first for a config on a Virtual host, then a route, and finally a route
   * entry (weighted cluster). If a config is not present, the callback will not be invoked.
   */
  virtual void traversePerFilterConfig(std::function<void(const Router::RouteSpecificFilterConfig&)> cb) const PURE;
};
```

Third, update `FilterFactoryCb` to provide custom filter name. Then when we create the filter instance, we can set the custom name to the filter.

```
/**
 * This function is used to wrap the creation of an HTTP filter chain for new streams as they
 * come in. Filter factories create the function at configuration initialization time, and then
 * they are used at runtime.
 * @param custom_name custom name of the filter that created by this factory.
 * @param callbacks supplies the callbacks for the stream to install filters to. Typically the
 * function will install a single filter, but it's technically possibly to install more than one
 * if desired.
 */
using FilterFactoryCb = std::function<void(absl::string_view custom_name, FilterChainFactoryCallbacks& callbacks)>;

Http::FilterFactoryCb FilterFactory::createFilterFactoryFromProtoTyped(const FakePb&, const std::string& stats_prefix,
    Server::Configuration::FactoryContext& context) {
  return [config](absl::string_view custom_name, Http::FilterChainFactoryCallbacks& callbacks) -> void {
    auto fitler = std::make_shared<CorsFilter>(config);
    filter->setNames(custom_name, name());
    callbacks.addStreamFilter(std::move(filter));
  };
}

```

Fouth, upate every filter to use new `mostSpecificPerFilterConfig` and `traversePerFilterConfig` provided by the `Http::StreamFilterCallbacks`.

wbpcode:
cc @mattklein123 Could you take a quick look to this simple proposal?

kyessenov:
Looks reasonable to me. One question I have is whether we have to expose the custom name method. The filter manager should have that info, and only share it with the callbacks implementation, without adding another public method.

wbpcode:
> Looks reasonable to me. One question I have is whether we have to expose the custom name method. The filter manager should have that info, and only share it with the callbacks implementation, without adding another public method.

In the current implementation, the filter instance is added to filter chain by the callback as following. So when we create the filter chain, it's hard to make the custom name available directly without change the callback or add new another public method.

```
class FilterChainFactoryCallbacks {
  virtual void addStreamDecoderFilter(Http::StreamDecoderFilterSharedPtr filter) PURE;
};
```

mattklein123:
> So when we create the filter chain, it's hard to make the custom name available directly without change the callback or add new another public method.

As long as you are working in this area, can you please fix all debug lines like this one:

https://github.com/envoyproxy/envoy/blob/dbfdab2f6d7613d174b42460e7fbbfb3f814bace/source/common/http/filter_manager.cc#L507-L509

To print out the name instead of the pointer address of the filter? This will make debugging multiple filters *so much easier*. I actually started to do this at one point but gave up because it would require a new public method which I think this solves.

wbpcode:
> But since we all agree that the current configuration is too heavy, I will respect the community's opinion.

👌 I will update all these log line.

wbpcode:
Inherited relationships before this PR of HTTP stream filter.

```mermaid

graph TD
  BA(StreamFilterBase)
  SD(StreamDecoderFilter)
  SE(StreamEncoderFilter)
  SF(StreamFilter)
  PD(PassthroughDecoderFilter)
  PE(PassthroughEncoderFilter)
  PF(PassthroughFilter)

  BA --> SD
  BA --> SE
  SD --virtual--> PD
  SD --virtual--> SF
  SE --virtual--> SF
  SE --virtual--> PE
  PD --> PF
  SF --> PF
  PE --> PF

```

Inherited relationships after this PR of HTTP stream filter:

```mermaid

graph TD
  BA(StreamFilterBase)
  SD(StreamDecoderFilter)
  SE(StreamEncoderFilter)
  SF(StreamFilter)
  PD(PassthroughDecoderFilter)
  PE(PassthroughEncoderFilter)
  PF(PassthroughFilter)

  BA --virtual--> SD
  BA --virtual--> SE
  SD --virtual--> PD
  SD --virtual--> SF
  SE --virtual--> SF
  SE --virtual--> PE
  PD --> PF
  SF --> PF
  PE --> PF

```

envoy/http/filter.h
```diff
@@ -302,6 +302,21 @@ class StreamFilterCallbacks {
    * Called when filter activity indicates that the stream idle timeout should be reset.
    */
   virtual void resetIdleTimer() PURE;
+
+  /**
+   * This is a helper to get the route's per-filter config if it exists, otherwise the virtual
+   * host's. Or nullptr if none of them exist.
+   */
+  virtual const Router::RouteSpecificFilterConfig* mostSpecificPerFilterConfig() const PURE;
```
yanavlasov, line 310: Isn't there already API for that https://github.com/envoyproxy/envoy/blob/c92c0326628fc75a6bcc15ae8e0d5448ca2a9c45/source/common/http/utility.h#L543 ?

yanavlasov (commented):
/wait-any

envoy/http/filter.h
wbpcode: Yeah. But it just is a help function that used to get per filter config from route and is keeped for some historical reson.

The new APIs will be used as a proxy to implement the fallback config searching policy. Finally, all the filters need to use these new APIs.

wbpcode:
Hi, cc @mattklein123 @kyessenov , this PR is ready for formal review.

This PR complete almost all the pre-work of #12274. The main changes of this PR are as follow:

1. New public method `(set)customName` and `(set)filterName` of stream filter to get and set config name and filter name.
2. A `FilterChainFactoryCallbacksNameHelper` to update inject config name and filter name to the filter instances automatically.
3. New API of `Http::StreamFilterCallbacks` to get per filter config by the config/filter name with a fallback policy. And it would be easily to support more flexible policy in the future.

Next, we only need a PR to update filters to used the new API to get per filter route config. After that, we can close #12274.

source/common/http/filter_manager.cc
```diff
@@ -277,6 +277,48 @@ void ActiveStreamFilterBase::resetIdleTimer() {
...
+  ASSERT(base_filter_ != nullptr);
+  auto route = parent_.filter_manager_callbacks_.route(nullptr);
+  if (route == nullptr) {
+    return nullptr;
+  }
+
+  std::string custom_name(base_filter_->customName());
+  auto* result = route->mostSpecificPerFilterConfig(custom_name);
+
+  if (result == nullptr) {
```
soulxu, line 291: We can skip the fallback if the filter doesn't use the custom name, so the check can be:
`if (result == nullptr && base_filter_->customName() != base_filter_->fileName())`

envoy/http/filter.h
```diff
@@ -302,6 +302,21 @@ class StreamFilterCallbacks {
    * Called when filter activity indicates that the stream idle timeout should be reset.
    */
   virtual void resetIdleTimer() PURE;
+
+  /**
+   * This is a helper to get the route's per-filter config if it exists, otherwise the virtual
+   * host's. Or nullptr if none of them exist.
```
soulxu, line 308: probably document the fallback logic here.

wbpcode: I am not sure. A key point to create this API is to make the searching policy transparent to the stream filters.

The filters needn't to know which policy is used for config searching.

wbpcode: I think it would be better to add some more comments in the proto API of typed_per_filter_config.

envoy/http/filter.h
```diff
@@ -626,6 +641,33 @@ class StreamDecoderFilterCallbacks : public virtual StreamFilterCallbacks {
   virtual absl::optional<absl::string_view> upstreamOverrideHost() const PURE;
 };
 
+/**
+ * Hepler used to set and get stream filter names. This is to avoid having to implement
+ * these method for every filter.
+ */
+class StreamFilterNameHepler {
```
mattklein123, line 648:
```suggestion
class StreamFilterNameHelper {
```

s/Hepler/Helper here and elsewhere.

envoy/http/filter.h
```diff
@@ -639,7 +681,7 @@ class StreamDecoderFilterCallbacks : public virtual StreamFilterCallbacks {
  * - onDestroy is used to cleanup all pending filter resources like pending http requests and
  * timers.
  */
-class StreamFilterBase {
+class StreamFilterBase : public StreamFilterNameHepler {
```
mattklein123, line 684: The inheritance here is complicated enough as it is. Are all these changes required? Couldn't we just expose name as a function on StreamFilterBase, why does it need to be a base class?

envoy/http/filter.h
```diff
@@ -626,6 +641,33 @@ class StreamDecoderFilterCallbacks : public virtual StreamFilterCallbacks {
   virtual absl::optional<absl::string_view> upstreamOverrideHost() const PURE;
 };
 
+/**
```
mattklein123, line 644: Per below does this need to be part of the public interface, especially the setting part? Can't we just expose the name?

mattklein123 (changes requested):
Some high level questions/comments to get started. Thank you for working on this!

/wait

envoy/http/filter.h
wbpcode: It's ok. ​My intention is to have a single helper to implement all name related methods.

envoy/http/filter.h
wbpcode: > can't this just be internal to the filter manager?

Yes, we can. But it wouldn't make things easier.

As you can see in the current implementation code snippet. The filter manager will create a filter chain for every request by the `FilterChainFactory`. And we can only add filters to the chain ( or filter manager) by the interface of `FilterChainFactoryCallbacks` for now.

```c++
class FilterChainFactory {
public:

  virtual void createFilterChain(FilterChainFactoryCallbacks& callbacks) PURE;

};
```

---

So, yeah we do have some other ways to make the name available for the filter manager. But there will also be some API changes.

First, we can update this `Http::FilterChainFactoryCallbacks` to add some new methods as following snippet. And we can also make the name available by a `FilterChainFactoryCallbacks` wrapper like `FilterChainFactoryCallbacksNameHelper` automatically.

```c++
class FilterChainFactoryCallbacks {
  /**
   * Add a decoder/encoder filter that is used both when reading and writing stream data.
   * @param custom_name supplies the custom configuration name.
   * @param filter_name supplies the canonical filter name.
   * @param filter supplies the filter to add.
   */
  virtual void addStreamFilter(absl::string_view custom_name, absl::string_view filter_name,
                               Http::StreamFilterSharedPtr filter) PURE;

};
```

---

Or we can update the `FilterChainFactory` to expose the factories and names to the filter manager rather than current way. For example, we can create new `FilterManager` API that will handle every named filter factory, then we can handle the name internally in the filter manager.

```c++
// New interface for the filter manager.
class FilterManager {
public:
     void evaluateFilterFactory(absl::string_view custom_name, absl::string_view filter_name,
                               Http::FilterFactoryCb& cb) PURE;
};

class FilterChainFactory {
public:
  // Add filter to the filter manager by the new interface rather than FilterChainFactoryCallbacks directly. Then
  // we can handle the name of filter internally in the filter manager.
  virtual void createFilterChain(FilterManager& filter_manager) PURE;
};
```


Except current implemantion in this PR, I prefer the second way. How do you think about it? cc @mattklein123

soulxu: > So, yeah we do have some other ways to make the name available for the filter manager. But there will also be some API changes.
> 
> First, we can update this `Http::FilterChainFactoryCallbacks` to add some new methods as following snippet. And we can also make the name available by a `FilterChainFactoryCallbacks` wrapper like `FilterChainFactoryCallbacksNameHelper` automatically.
> 
> ```c++
> class FilterChainFactoryCallbacks {
>   /**
>    * Add a decoder/encoder filter that is used both when reading and writing stream data.
>    * @param custom_name supplies the custom configuration name.
>    * @param filter_name supplies the canonical filter name.
>    * @param filter supplies the filter to add.
>    */
>   virtual void addStreamFilter(absl::string_view custom_name, absl::string_view filter_name,
>                                Http::StreamFilterSharedPtr filter) PURE;
> 
> };
> ```

not sure I understand that your second way right. If I understand right, I feel like that is a better choice.

Then we can enable the `ActiveStreamDecoderFilter` or `ActiveStreamFilterBase` to hold the custom name and filter name by this new `addStreamFilter` interface.

envoy/http/filter.h
```diff
@@ -626,6 +641,33 @@ class StreamDecoderFilterCallbacks : public virtual StreamFilterCallbacks {
...
+
+  void setCustomName(absl::string_view view) { custom_name_ = std::string(view); }
+  void setFilterName(absl::string_view view) { filter_name_ = std::string(view); }
+
+protected:
+  // TODO(wbpcode): typically, the custom name and filter name have longer lifetime than the
+  // stream filter instances. We can simply keep string view here to avoid unnecessary copy
+  // and memory allocation. Currently, new strings are still always created for safety.
+  std::string custom_name_;
+  std::string filter_name_;
```
soulxu, line 668: not sure it is worth or not, in case of the filter doesn't have custom name, there will be a duplicated of the same name, maybe using the std::shared_ptr to able to get rid of this duplicated name.

envoy/http/filter.h
wbpcode: > not sure I understand that your second way right. If I understand right, I feel like that is a better choice.

> For example, we can create new FilterManager API that will handle every named filter factory, then we can handle the name internally in the filter manager.
```c++
// New interface for the filter manager.
class FilterManager {
public:
     void evaluateFilterFactory(absl::string_view custom_name, absl::string_view filter_name,
                               Http::FilterFactoryCb& cb) PURE;
};

class FilterChainFactory {
public:
  // Add filter to the filter manager by the new interface rather than FilterChainFactoryCallbacks directly. Then
  // we can handle the name of filter internally in the filter manager.
  virtual void createFilterChain(FilterManager& filter_manager) PURE;
};
```

Do you mean this one? I also like this one. If we choose this way finally, I think I would create another seperate PR to update `FilterChainFactory`.

envoy/http/filter.h
wbpcode: I will try to make these names to be string_view finally. And I think it's unnecessary to introduce a shared_ptr for single string. 🤔

envoy/http/filter.h
soulxu: haha, sorry, that isn't the one I said. Looks like what I choose is your first way.

It is almost the same as the current version, just change the `addStreamFilter` interface to accept the custom name and filter name, then pass those two names into the `ActiveStreamDecoderFilter` or `ActiveStreamFilterBase`. Then we needn't the filter itself to hold those two names and expose those interfaces.

wbpcode: @soulxu In that way, we need to update three different APIs `addStreamFilter`, `addDecoderStreamFilter` and `addEncoderStreamFilter`.

And we need keep two different version (old APIs and new APIs with names) in the `Http::FilterChainFactoryCallbacks` to avoid breaking current filters. (This is the main reason I think this way is not so good.)

mattklein123: I think we need to step back a second. I don't really understand why anyone would specify a custom name in code. shouldn't this come from config or the statically known names, and if so, can't that just be done directly by the filter manager? That seems like a much better solution to me. I still don't quite understand why any of this needs to be exposed publicly. It seems it should be all internal to the filter manager and the tracked filters there.

/wait

wbpcode: In the current implementation of `FilterChainFactoryCallbacks` and `FilterChainFactory`, there is no any easy way to make the custom/filter names available to the filter manager. There is no position for additional filter info (custom/filter names) in current interfaces.


**We must update stream filter APIs to expose these names**

**or**

**update APIs of `FilterChainFactoryCallbacks` and `FilterChainFactory`.**


In the current PR, I choose to update stream filter APIs to expose these names. And the filter itself can also easily access these name.


The setting part is compromise. By these public setter, we can set the names from configuration or static name uniformly at the outrage of filters. Then we needn't implement the getter one by one. But I admit that public setters really aren't a good design.

Of course, we can step back one second. 
We can choose **update APIs of `FilterChainFactoryCallbacks` and `FilterChainFactory`.**

mattklein123: Here is an example filter creation:
```
return [&context, config](Http::FilterChainFactoryCallbacks& callbacks) -> void {
    callbacks.addStreamDecoderFilter(
        std::make_shared<SquashFilter>(config, context.clusterManager()));
  };
```

What if we construct a synthetic FilterChainFactoryCallbacks that we pass to each filter factory. This synthetic callbacks would have the name info. Then, it can trap the add filter calls, add the custom names, and create them internally in the filter manager with the names. Wouldn't that be transparent and pretty clean?

wbpcode: Yeah, it is transparent and clean. It's also what I want.

But note that the custom names and filter factories are managed by the HCM config (derived from FilterChainFactory). And when filter manager needs create a filter chain, it will call the creatFilterChain method of FilterChainFactory. And filter manager self will be used as FilterChainFacotryCallbacks parameter of creatFilterChain method.

We can created a wrapper to trap the add filter calls. But the wrapper can only call the APIs exposed by the FilterChainFacotryCallbacks. It cannot make the names available to the filter manager directly.

If we don't want add new APIs in the stream filter, we still need update FilterChainFacotryCallbacks or FilterChainFactory to complete this work.

In fact, I has implemented similar wrapper in this PR.

mattklein123: @wbpcode and I discussed this offline and I think came up with a good solution which will involve changing the interface that the filter manager uses but not any of the filters, so there shouldn't be any public changes. @wbpcode do you mind summarizing to make sure everyone likes the plan? Thank you!

wbpcode: Get it.

wbpcode:
This PR is part of work to close #12274. We want the per filter configs can be indexed by the `HttpFilter.name`. 
And one of the key points is to make the configured HTTP filter name available to the filter manager.

However, in the current implementation, the filter names and filter factories is managed by the HCM config (derived from `FilterChainFactory`). And the `FilterChainFactoyCallbacks` interface are the only way used by HCM config (derived from `FilterChainFactory`) to inject the filter instances to the filter manager. There's no position for the filter names in the current `FilterChainFactoyCallbacks`.

So, to make the filter names available to the filter manager, we must update `StreamFilter`, `FilterChainFactoyCallbacks`, or `FilterChainFactory`. 
**And we must ensure these updates won't break current filters.** 

Our initial design is add new interfaces to the StreamFilter to expose the filter names. But this need updates all HTTP filters or some unbefitting setting methods. 

Our second way is that updating `FilterChainFactoyCallbacks`. We can add some new interfaces to support adding names to filter manager and keep the old version methods to avoid breaking current filters. But the `FilterChainFactoyCallbacks` is widely used by the current filter developers. So we don't want to change that. 

So, we finally decide to update `FilterChainFactory`.  We will create a new interface `FilterChainManager`. And the HCM config (derived from FilterChainFactory) will post filter names and factories to the filter manager by the method of `FilterChainManager`. 

```
/**
 * The filter chain manager is provided by the connection manager to the filter chain factory.
 * The filter chain factory will post the filter factory context and filter factory to the
 * filter chain manager to create filter and construct HTTP stream filter chain.
 */
class FilterChainManager {
public:
  /**
   * Post filter factory context and filter factory to the filter chain manager. The filter
   * chain manager will create filter instance based on the context and factory internally.
   * @param context supplies additional contextual information of filter factory.
   * @param factory
   */
  virtual void postFilterFactory(FilterFacotryContext context, FilterFactoryCb& factory) PURE;
};

// Update FilterChainFactory to use the new interface FilterChainManager instead of FilterChainFactoyCallbacks.
class FilterChainFactory {
public:
  /**
   * Called when a new HTTP stream is created on the connection.
   * @param manager supplies the "sink" that is used for actually creating the filter chain. @see
   *                FilterChainManager.
   */
  virtual void createFilterChain(FilterChainManager& manager) PURE;
  // ......
};
```

And the filter manager will create `FilterChainFactoyCallbacks`  implementation with filter names internally. When the filter factories creating filters, this FilterChainFactoyCallbacks will trap the filter instances and adding instances and names to filter manager. 

```
class FilterChainFactoryCallbacksNameHelper : public Http::FilterChainFactoryCallbacks {
public:
  FilterChainFactoryCallbacksNameHelper(absl::string_view custom_name,
                                        absl::string_view filter_name,
                                        Http::FilterManager& filter_manager)
      : custom_name_(custom_name), filter_name_(filter_name), filter_manager_(callback) {}

  void addStreamDecoderFilter(Http::StreamDecoderFilterSharedPtr filter) override {
    filter_manager_.addStreamDecoderFilterWorker(custom_name_, filter_name_, std::move(filter), false);
  }
  // ......
};
```

wbpcode:
Thanks for all the great suggesions. 🌹  cc @mattklein123 @soulxu

source/common/http/filter_manager.h
```diff
@@ -840,6 +864,45 @@ class FilterManager : public ScopeTrackedObject,
...
+    }
+
+    void addStreamFilter(Http::StreamFilterSharedPtr filter) override {
+      StreamDecoderFilter* decoder_filter = filter.get();
+      manager_.addStreamFilterBase(decoder_filter);
+
+      manager_.addStreamDecoderFilter(std::make_unique<ActiveStreamDecoderFilter>(
+          manager_, filter, true, context_.custom_name, context_.filter_name));
+      manager_.addStreamEncoderFilter(std::make_unique<ActiveStreamEncoderFilter>(
+          manager_, std::move(filter), true, context_.custom_name, context_.filter_name));
```
soulxu, line 892: nit, since it is shared ptr, so actually, we needn't the std::move here, right?

source/common/filter/config_discovery_impl.h
```diff
@@ -164,9 +164,17 @@ class DynamicFilterConfigProviderImpl : public DynamicFilterConfigProviderImplBa
   ThreadLocal::TypedSlot<ThreadLocalConfig> tls_;
 };
 
+// Struct of canonical filter name and HTTP stream filter factory callback.
+struct NamedHttpFilterFactoryCb {
+  // Canonical filter name.
+  std::string name;
```
soulxu, line 170: Is it possible to put this name to the FilterConfigProvider also?

soulxu (commented):
I like this way. Just one more comment about whether we can put the all the name to the FilterConfigProvider. Then we needn't that `NamedHttpFilterFactoryCb` wrapper.

source/common/http/filter_manager.h
wbpcode: This move is used to avoid unnecessry atomic counter updates.

wbpcode:
> I like this way. Just one more comment about whether we can put the all the name to the FilterConfigProvider. Then we needn't that `NamedHttpFilterFactoryCb` wrapper.

It's ok for me. But I am just not sure if we need to expose this canonical name for every type extension? Until for now, the  canonical name is only used by the HTTP filters to get route config. And we keep it for the backwards compatibility. So I am not sure if it's worth it to update the APIs of `ExtensionConfigProvider`.

But I'm not against this either. I can update it if the community wishes to do so.

cc @mattklein123 cc @soulxu

soulxu:
> > I like this way. Just one more comment about whether we can put the all the name to the FilterConfigProvider. Then we needn't that `NamedHttpFilterFactoryCb` wrapper.
> 
> It's ok for me. But I am just not sure if we need to expose this canonical name for every type extension. Until for now, the canonical name is only used by the HTTP filters to get route config. And we keep it for the backwards compatibility. So I am not sure if it's worth it to update the APIs of `ExtensionConfigProvider`.
> 

Since I saw the FilterConfigProvider already holds the custom name, and the canonical name is in the subscription object, so I thought the FilterConfigProvider can expose easily just return the name from the subscription object. But yea, I can see others' opinions. Thanks

wbpcode:
> > > I like this way. Just one more comment about whether we can put the all the name to the FilterConfigProvider. Then we needn't that `NamedHttpFilterFactoryCb` wrapper.
> > 
> > 
> > It's ok for me. But I am just not sure if we need to expose this canonical name for every type extension. Until for now, the canonical name is only used by the HTTP filters to get route config. And we keep it for the backwards compatibility. So I am not sure if it's worth it to update the APIs of `ExtensionConfigProvider`.
> 
> Since I saw the FilterConfigProvider already holds the custom name,https://github.com/envoyproxy/envoy/pull/21525/files#diff-9ffd7a166e0bd6afee8cbe95c75544a26060e9d2bc814ba3309fa4c04d8860ecR738 so I thought the FilterConfigProvider can hold the canonical name also

custom name is used by the ECDS, but canonical name is unnecessary for most of extention config providers.

Any way, it ok for me to udpate it or not.  So waiting for some more suggessions from matt.

envoy/http/filter.h
wbpcode: New comments should be added to the typed_per_filter_config. But it should be done in the new PR which will update the filters to use new mostSpecificPerFilterConfig.

envoy/http/filter.h
```diff
@@ -1063,6 +1078,35 @@ class FilterChainFactoryCallbacks {
  */
 using FilterFactoryCb = std::function<void(FilterChainFactoryCallbacks& callbacks)>;
 
+/**
+ * Simple struct of additional contextual information of filter factory, e.g. custom filter name
+ * from configuration, canonical filter name, etc.
+ */
+struct FilterFacotryContext {
```
mattklein123, line 1085:
```suggestion
struct FilterFactoryContext {
```

envoy/http/filter.h
```diff
@@ -1063,6 +1078,35 @@ class FilterChainFactoryCallbacks {
...
+  // Custom filter name from configuration.
+  absl::string_view custom_name;
+  // Canonical filter name.
+  absl::string_view filter_name;
```
mattklein123, line 1089: I don't know what a custom filter name and canonical name is. Can you provide some examples?

source/common/filter/config_discovery_impl.h
mattklein123: +1 to merge this in some way. I'm not familiar with this code but it seems like there is a bunch of duplication with the `FilterFactoryContext` also? Can this be shared somehow?

source/common/http/filter_manager.h
```diff
@@ -138,6 +143,9 @@ struct ActiveStreamFilterBase : public virtual StreamFilterCallbacks,
   FilterManager& parent_;
   IterationState iteration_state_{};
 
+  const std::string custom_name_;
+  const std::string filter_name_;
```
mattklein123, line 147: Probably better to just store the FactoryContext here? This would avoid the same general questions about custom  name vs. filter name which would be resolved by the comments in that struct.

mattklein123 (changes requested):
Thanks in general this LGTM. Flushing some small comments. Thank you!

/wait

source/common/http/filter_manager.h
wbpcode: get it.

source/common/filter/config_discovery_impl.h
wbpcode: I have just tried to add a method to expose this name in the `FilterConfigProvider`. But it just make the code dirty.

Because it's supported to used different type filters by the ECDS, so the canonical filter name may be changed when the filter config is updateing by the ECDS.
The canonical filter name has different lifetime with the custom config name and may be updated frequently, so it's hard to used `FilterFactoryContext` to  store these name in the `FilterConfigProvider`.

wbpcode:
```
message ExtensionConfigSource {
  ConfigSource config_source = 1 [(validate.rules).any = {required: true}];

  // Optional default configuration to use as the initial configuration if
  // there is a failure to receive the initial extension configuration or if
  // `apply_default_config_without_warming` flag is set.
  google.protobuf.Any default_config = 2;

  // Use the default config as the initial configuration without warming and
  // waiting for the first discovery response. Requires the default configuration
  // to be supplied.
  bool apply_default_config_without_warming = 3;
  
  // **Here the repeated type_urls make the filter type can be changed by the ECDS. A ratelimit filter can be** 
  // **updated to ext_authz.**
  // A set of permitted extension type URLs. Extension configuration updates are rejected
  // if they do not match any type URL in the set.
  repeated string type_urls = 4 [(validate.rules).repeated = {min_items: 1}];
}
```

IMO, it would be better to add some restriction here. For example, only multiple version APIs of same filters can be used here.

soulxu: > IMO, it would be better to add some restrictions here. For example, only multiple version APIs of same type filter can be used here.

I guess even though there can be multiple extension type URLs, they should pointed to the same factory, or they should point to the same canonical filter name?
https://github.com/envoyproxy/envoy/blob/154a7a865c064870f500a11d2772b2ca106337ac/source/extensions/filters/network/http_connection_manager/config.cc#L674-L682

Then you can get `factory->name()`, and pass that to the dynamic filter config provider.
https://github.com/envoyproxy/envoy/blob/154a7a865c064870f500a11d2772b2ca106337ac/source/extensions/filters/network/http_connection_manager/config.cc#L684-L686

@adisuissa probably knows the usecase of multiple extension type urls.

wbpcode: > I guess even though there can be multiple extension type URLs, they should pointed to the same factory, or they should point to the same canonical filter name?

It seems that there is no any restriction for now. It would be great if we can add this restriction. But it would be a break change.
So, I think we can keep this PR's implemantation and after we this restriction is shipped someday, we can update the implemantation with a new minor patch.

mattklein123 (commented):
Thanks LGTM with one question.

/wait-any

envoy/http/filter.h
```diff
@@ -1079,14 +1079,16 @@ class FilterChainFactoryCallbacks {
...
-  absl::string_view custom_name;
-  // Canonical filter name.
-  absl::string_view filter_name;
+struct FilterContext {
+  // The name of the filter configuration that used to create related filter factory function.
+  // This could be any legitimate non-empty string.
+  std::string config_name;
+  // Canonical filter name. This is used to indicate type of filter. For example,
+  // "envoy.filters.http.buffer" for the HTTP buffer filter.
+  std::string filter_name;
```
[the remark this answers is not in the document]
wbpcode: It's still necessary for now. Because this name will be used to search the per filter config, as a fallback of config name. 🤔 

If we deprecate the fallback policy and used the config name only someday, we can remove this name.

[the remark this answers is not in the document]
kyessenov: I think the phrasing needs to be better. 
```
DEPRECATED: Filter extension qualified name. Used as a fallback for `config_name`.  E.g. ...
```

This is needed to support extension names in typed_config I think, so we can't change it to anything but what it was (which was extension qualified name).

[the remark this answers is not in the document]
wbpcode: get it.

soulxu (approved):
LGTM, thanks

mattklein123 (approved):
Thank you!

PR #21529, opened 2022-06-01
fuzz test: Close only open connections.

vehre-x41:
Commit Message: fuzz test: Close only open connections.
Additional Description:
- In tls_inspector_fuzz_test an event-loop could block due to getting no
signal.
- Use a flag to indicate, that a connection has been opened and close it
only, when the flag has been set ensuring a close event can be received.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


adisuissa (approved):
LGTM, thanks!

Issue #21538, opened 2022-06-02
Make HappyEyeballsConnectionImpl more general to accommodate broader use cases


wanlill:
*Description*:

According to the comment: `HappyEyeballsConnectionImpl` transparently attempts connections to multiple different IP addresses, and uses the first connection that succeeds. After a connection is established, all methods simply delegate to the underlying connection.

I would like to see whether we want to make it more generic: instead of limiting the multiple connections to different IP addresses only, we can introduce a `ConnectionProvider` policy interface to it and abstract the `HappyEyeballsConnectionImpl` into a `MultiConnectionBaseImpl` class, while the logic of trying different IP addresses can be implemented by a `HappyEyeballsConnectionProvider`.

With `MultiConnectionBaseImpl`, people can write extensions to provide different `ConnectionProvider` policies for use cases like [Opportunistic encryption for downstream envoy](https://github.com/envoyproxy/envoy/issues/7135#)  (I understand we have security concern with it but this can be a real business requirement to ease migration efforts and this is just one example that shows what can be done with the generalized `HappyEyeballsConnectionImpl`)

I'm going to start a POC PR to make sure the idea is expressed clearly, I'm more than happy to refine it later if we think this proposal is reasonable.

PR #21539, opened 2022-06-02
Add a connection provider to happy eyeball connection implementation and split it into a new class for broader use cases

wanlill:
Commit Message:
Additional Description: this is a POC for https://github.com/envoyproxy/envoy/issues/21538 which proposed to make HappyEyeballsConnectionImpl more generic to accommodate broader use cases.
Risk Level: low
Testing: refactored existing tests to cover this change
Docs Changes: none
Release Notes: none
Platform Specific Features: none


KBaichoo:
/assign @RyanTheOptimist 

I think you added happy eyeball

source/common/network/happy_eyeballs_connection_impl.h
```diff
@@ -32,198 +57,20 @@ namespace Network {
...
+                              const ConnectionSocket::OptionsSharedPtr options)
+      : MultiConnectionBaseImpl(dispatcher,
+                                std::make_unique<Network::HappyEyeballsConnectionProvider>(
+                                    dispatcher, address_list, source_address, socket_factory,
+                                    transport_socket_options, options)) {}
 };
 
 } // namespace Network
-} // namespace Envoy
+} // namespace Envoy
```
RyanTheOptimist, line 76: nit: I think this is missing a newline.

source/common/network/multi_connection_base_impl.cc
```diff
@@ -0,0 +1,591 @@
+#include "source/common/network/multi_connection_base_impl.h"
+
+#include <vector>
+
+namespace Envoy {
+namespace Network {
+
+MultiConnectionBaseImpl::MultiConnectionBaseImpl(Event::Dispatcher& dispatcher,
```
RyanTheOptimist, line 8: This file was basically renamed from happy_eyeballs_connection_impl.cc, right? It's really unfortunate that git is failing to notice the copy and so the diff is showing up as complete new code. This makes it hard to see what's changed about it in this PR (other than the names :/). I have two thoughts for how to improve this though I'm not sure if either will work. Both will take some work.

First, I think we might be able to help git figure this out by having the first commit in this PR simply copy happy_eyeballs_connection_impl.{h,cc} => multi_connection_base_impl.{h,cc}. Then the second commit can make changes on top of that. You'd end up needing to do git push -f I think.

If that doesn't work, we could add a PR which simply does the copy but doesn't edit or use the new files. I'm sure that would work, though it would be slightly annoying. (Though better than losing the history)

test/common/network/happy_eyeballs_connection_impl_test.cc
```diff
@@ -172,9 +172,9 @@ TEST_F(HappyEyeballsConnectionImplTest, DisallowedFunctions) {
   startConnect();
 
   EXPECT_ENVOY_BUG(connection_callbacks_[0]->onAboveWriteBufferHighWatermark(),
```
RyanTheOptimist, line 174: In addition to these happy eyeballs tests, we should add new unit tests for the new MultiConnectionBaseImpl class itself.

RyanTheOptimist (commented):
Overall the code in this PR seems reasonable. I added a few suggestions, the biggest of which is a request to see if we can make git do a better job of showing the actual diffs, since it's hard to review otherwise. I'm curious, though, if we have an actual use case for this at the moment. The issue mentions Opportunistic Encryption, but that was abandoned for good security reasons and I don't think we're likely to restart that effort. Is there another use case you might foresee?

wanlill:
I apologize for the late response (on a trip this week), I'll come back to this early next week.

source/common/network/happy_eyeballs_connection_impl.h
wanlill: Ack, newline added.

source/common/network/multi_connection_base_impl.cc
wanlill: I've split the commit into two, it's still not very clear but we can now at least check the change introduced by the second commit with Github UI, thanks for the guide!

wanlill:
> Overall the code in this PR seems reasonable. I added a few suggestions, the biggest of which is a request to see if we can make git do a better job of showing the actual diffs, since it's hard to review otherwise. I'm curious, though, if we have an actual use case for this at the moment. The issue mentions Opportunistic Encryption, but that was abandoned for good security reasons and I don't think we're likely to restart that effort. Is there another use case you might foresee?

#21640 referenced this PR and I would +1 on that this PR can be helpful for that use case.

Another use case I have in my mind is like below:

In a mutual TLS environment, sometimes the client side needs to update its certificate but it may take sometime for the server side to be updated to accept the new certificate. In this scenario, we can create use a custom ConnectionProvider and associate two transport socket contexts (one for each client certificate). This can unblock the rollout of the client side change before the server sides update are all done and can be important for an environment owned by different teams. That being said, we can definitely change the TLS transport socket to allow [multiple client certificates](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#extensions-transport-sockets-tls-v3-commontlscontext) for this use case but the ConnectionProvider way is more extensible and flexible to me.

test/common/network/happy_eyeballs_connection_impl_test.cc
wanlill: I've split happy_eyeballs_connection_impl_test.cc` into `multi_connection_base_impl_test.cc` and `happy_eyeballs_connection_provider_test.cc`, IMO most of the test cases should be moved to `multi_connection_base_impl_test`

source/common/network/multi_connection_base_impl.cc
RyanTheOptimist: I'm sure I'm just doing it wrong, but how I can I see the change to source/common/network/multi_connection_base_impl.cc which were made after copying it from source/common/network/happy_eyeballs_connection_impl.cc? (Sometimes the GH UI is really confusing!)

wanlill: I was able to check the changes for a single commit with link like [this](https://github.com/envoyproxy/envoy/pull/21539/commits/e10035dc214a852b6abe218bf8c44ec402d65a66), and different commit can be selected with "Files Changed" tab and then -> "Changes from all commits" dropdown list to select a particular commit.

RyanTheOptimist: Ah! Thanks, I was having a hard time finding the right commit, I guess.!

RyanTheOptimist (dismissed):

RyanTheOptimist:
Needs a main merge, then I can hand off to a non-google sr maintainer for final review.

RyanTheOptimist (approved):
/assign @mattklein123

soulxu:
I imagine I can reuse this for the `item #3` in https://github.com/envoyproxy/envoy/issues/21640, but it is still under discussion.

After look into this more, I found maybe I needn't add new connection provider implement here. Just need to add more policies to the current happy eyeball implement. Those policies can be `default(the current behavior)`, `v4_first`, `v6_first`, `strict_match_downstream_version`. So I wasn't very sure my case need the connection provider yet.

But, currently, the happy eyeball connection will be enabled implicitly when you use logic DNS and the DNS returns multiple addresses https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/connection_pooling#arch-overview-happy-eyeballs

So I need to extend the API to accept those policies, if we are going this way, Will this expose an extension point to the API? (I guess that extension point should be at the ClientConnection level, not the connection provider level), then I can add policies to the happy eyeball connection extension point.

mattklein123 (approved):
Thanks!

PR #21540, opened 2022-06-02
Check owner of envoy docker build dir

xu1zhou:
Commit Message: Check if `ENVOY_DOCKER_BUILD_DIR` is created and owned by other user. Build will exit to make sure permission problem is resolved
Additional Description: The permisson of files within `$ENVOY_DOCKER_BUILD_DIR` will stay the same with last build user. In that case, build from other user will be blocked by "permission denied". 
Platform Specific Features: Linux

### DIFF
--- a/ci/run_envoy_docker.sh
+++ b/ci/run_envoy_docker.sh
@@ -54,6 +54,12 @@ fi
 [[ -z "${ENVOY_DOCKER_BUILD_DIR}" ]] && ENVOY_DOCKER_BUILD_DIR="${DEFAULT_ENVOY_DOCKER_BUILD_DIR}"
 # Replace backslash with forward slash for Windows style paths
 ENVOY_DOCKER_BUILD_DIR="${ENVOY_DOCKER_BUILD_DIR//\\//}"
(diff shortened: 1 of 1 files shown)


xu1zhou:
@phlax

xu1zhou:
@phlax This PR is to help developer who works on shared host with other guys. Beginner user build envoy with minimal options which uses `/tmp/envoy-docker-build`  as `ENVOY_DOCKER_BUILD_DIR`. But if there is other guy has built before and owned this directory, subsequent build of other user will fail due to in lack of permission to `/tmp/envoy-docker-build`. Worse still, user may be confused since the permission error message only implys mounted path in container (like /build) have insufficient permission.

xu1zhou:
@phlax any suggestion on that?

> @phlax This PR is to help developer who works on shared host with other guys. Beginner user build envoy with minimal options which uses `/tmp/envoy-docker-build` as `ENVOY_DOCKER_BUILD_DIR`. But if there is other guy has built before and owned this directory, subsequent build of other user will fail due to in lack of permission to `/tmp/envoy-docker-build`. Worse still, user may be confused since the permission error message only implys mounted path in container (like /build) have insufficient permission.

PR #21544, opened 2022-06-02
Listener ECDS: reject the connection directly when missing config

### LINKED ISSUE #21427: listener ecds: reject the connection directly instead of using a fake listener filter
see the discussion here https://github.com/envoyproxy/envoy/pull/21133#discussion_r867488644

current implementation is using a `MissingConfigTcpListenerFilter` to reject the connection https://github.com/envoyproxy/envoy/pull/21133/files/9f2b442e8c90cf1e426917ed4de8d3f303eff90f#diff-998fda72180293e9c2eaf07d3ead906aba6998220cbcd9f10f63bc283f5c6119R41

but it seems overcomplex, we should be able to reject the connection directly.

soulxu:

Commit Message: Listener ECDS: reject the connection directly when missing config
Additional Description:
Risk Level: high
Testing: the existing integration test covered
Docs Changes: n/a
Release Notes: n/a
Fixes #21427

### DIFF
--- a/source/server/active_stream_listener_base.h
+++ b/source/server/active_stream_listener_base.h
@@ -90,6 +90,7 @@ public:
       // If create listener filter chain failed, it means the listener is missing
       // config due to the ECDS. Then close the connection directly.
       active_socket->socket_->close();
+      ASSERT(active_socket->iter_ == active_socket->accept_filters_.end());
     }
 
     // Move active_socket to the sockets_ list if filter iteration needs to continue later.
--- a/source/server/active_stream_listener_base.h
+++ b/source/server/active_stream_listener_base.h
@@ -84,8 +84,13 @@ public:
 
   void onSocketAccepted(std::unique_ptr<ActiveTcpSocket> active_socket) {
     // Create and run the filters
-    config_->filterChainFactory().createListenerFilterChain(*active_socket);
-    active_socket->continueFilterChain(true);
+    if (config_->filterChainFactory().createListenerFilterChain(*active_socket)) {
+      active_socket->continueFilterChain(true);
+    } else {
+      // If create listener filter chain failed, it means the listener is missing
+      // config due to the ECDS. Then close the connection directly.
+      active_socket->socket_->close();
+    }
 
     // Move active_socket to the sockets_ list if filter iteration needs to continue later.
     // Otherwise we let active_socket be destructed when it goes out of scope.
--- a/source/server/configuration_impl.cc
+++ b/source/server/configuration_impl.cc
@@ -38,60 +38,15 @@ bool FilterChainUtility::buildFilterChain(Network::FilterManager& filter_manager
   return filter_manager.initializeReadFilters();
 }
 
-/**
- * All missing listener config stats. @see stats_macros.h
- */
-#define ALL_MISSING_LISTENER_CONFIG_STATS(COUNTER) COUNTER(extension_config_missing)
-/**
- * Struct definition for all missing listener config stats. @see stats_macros.h
- */
-struct MissingListenerConfigStats {
-  ALL_MISSING_LISTENER_CONFIG_STATS(GENERATE_COUNTER_STRUCT)
-};
-
-class MissingConfigTcpListenerFilter : public Network::ListenerFilter,
-                                       public Logger::Loggable<Logger::Id::filter> {
-public:
-  MissingConfigTcpListenerFilter(Stats::Scope& stats_scope)
-      : scope_(stats_scope), stats_({ALL_MISSING_LISTENER_CONFIG_STATS(POOL_COUNTER(scope_))}) {}
-
-  // Network::ListenerFilter
-  Network::FilterStatus onAccept(Network::ListenerFilterCallbacks& cb) override {
-    ENVOY_LOG(debug, "Listener filter: new connection accepted while missing configuration. "
-                     "Close socket and stop the iteration onAccept.");
-    cb.socket().ioHandle().close();
-    stats_.extension_config_missing_.inc();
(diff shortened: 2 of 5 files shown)


soulxu:
cc @yanjunxiang-google @adisuissa @kyessenov

source/server/configuration_impl.cc
```diff
@@ -38,60 +38,15 @@ bool FilterChainUtility::buildFilterChain(Network::FilterManager& filter_manager
...
-      config_value(filter_manager);
-      continue;
-    }
-
-    // If a filter config is missing after warming, stop iteration.
-    if (!added_missing_config_filter) {
-      filter_manager.addAcceptFilter(nullptr,
-                                     std::make_unique<MissingConfigTcpListenerFilter>(stats_scope));
-      added_missing_config_filter = true;
+    if (!config.has_value()) {
```
yanjunxiang-google, line 45: Thanks for taking care of this!

We are working on the UDP listener ECDS now. IIUC, we need to keep this logic consistent there, i.e, directly return in this function if the dynamic config is missing: https://github.com/envoyproxy/envoy/blob/97d599f0bda7ca2a4402f7ef0902484a602492ab/source/server/configuration_impl.cc#L100

Currently buildUdpFilterChain() return void.  Should we change it into return bool, and install same logic here to keep them consistent?
https://github.com/envoyproxy/envoy/blob/97d599f0bda7ca2a4402f7ef0902484a602492ab/source/server/active_udp_listener.cc#L102

soulxu: Checking the code, the UDP is different from TCP. The TCP will create new filter factories for the new connection, the UDP is different.

https://github.com/envoyproxy/envoy/blob/75aa3fbac7a72d27e6a7a003ddfeac15883e36e2/source/server/active_udp_listener.cc#L102

I'm curious how do you implement ecds for udp, since the filter factories is created in the beginning of listener created, if the filter factories begin with a invalid config, then you have no chance update the filter factories.

soulxu: Saw @yanjunxiang-google submitted a draft for UDP ecds https://github.com/envoyproxy/envoy/pull/21606

If we enable to trigger the udp listener filter factory to recreate when config updated, then we should be able to do the same thing as the TCP doing, just reject the connection, when an empty config was updated.

source/server/active_stream_listener_base.h
```diff
@@ -84,8 +84,13 @@ class ActiveStreamListenerBase : public ActiveListenerImplBase,
...
   void onSocketAccepted(std::unique_ptr<ActiveTcpSocket> active_socket) {
     // Create and run the filters
-    config_->filterChainFactory().createListenerFilterChain(*active_socket);
-    active_socket->continueFilterChain(true);
+    if (config_->filterChainFactory().createListenerFilterChain(*active_socket)) {
+      active_socket->continueFilterChain(true);
+    } else {
+      // If create listener filter chain failed, it means the listener is missing
+      // config due to the ECDS. Then close the connection directly.
+      active_socket->socket_->close();
```
adisuissa, line 92: Is there a test that validates what happens when the connection is closed, but `active_socket->iter_` isn't at the end of the filters?
Will those filters work as expected if the actual socket is closed?

adisuissa (commented):
Thanks, this does simplify the code.
Left a high-level question.

source/server/active_stream_listener_base.h
soulxu: This is going to line 101, which is the same code path when the listener filter rejected the connection. So it should be tested well.

Here is ActiveTcpSocket knew that the listener filter rejected the connection
https://github.com/envoyproxy/envoy/blob/c0bf266187924c780d5cf2d2ffd08dd44f755b5a/source/server/active_tcp_socket.cc#L126-L129

Then it sets the `active_socket->iter_` to the `active_socket->accept_filters_.end()`
https://github.com/envoyproxy/envoy/blob/c0bf266187924c780d5cf2d2ffd08dd44f755b5a/source/server/active_tcp_socket.cc#L159-L164

The initial value of `active_socket->iter_` is `active_socket->accept_filters_.end()`
https://github.com/envoyproxy/envoy/blob/c0bf266187924c780d5cf2d2ffd08dd44f755b5a/source/server/active_tcp_socket.cc#L16

So we close the socket and have `active_socket->iter_` as the end, then it is going to the same code path.

We have integration test cover this 
https://github.com/envoyproxy/envoy/blob/c0bf266187924c780d5cf2d2ffd08dd44f755b5a/test/integration/listener_lds_integration_test.cc#L917

adisuissa: I guess that essentially after the call for `close()` one can add an `ASSERT(active_socket->iter_ == active_socket->accept_filters_.end()`.
What confused me was the structure of this function.

adisuissa (approved):
LGTM.
@lambdai can you PTAL.

source/server/active_stream_listener_base.h
soulxu: got it, that is good idea, thanks!

yanjunxiang-google: LGTM for the changes outside of this function.  For the changes inside this function, I am a little bit not clear what the code does from line 97 to the end and how this change impact them.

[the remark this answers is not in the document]
soulxu: thanks! that is good question. If we return here, then we will lose the access log at line 103. I guess it would be good to keep the access log, WDYT?

adisuissa (approved):
LGTM.
@mattklein123 can you PTAL?

mattklein123 (approved):

PR #21551, opened 2022-06-02
tooling: Add python data macro

phlax:
For optimized preloading of data

Commit Message:
Additional Description:

This macro takes a yaml or json data source, creates a python data pickle from it and exposes a python lib containing the `data`

If we load all of protodoc's data this way (and do any other validation/cleanup/mangling beforehand) we should be able to reduce overheads

in this initial pr the data artefacts required by protodoc are imported individually - in a followup i will merge the data to a single artefact

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/tools/base/envoy_python.bzl
+++ b/tools/base/envoy_python.bzl
@@ -1,5 +1,5 @@
 load("@rules_python//python:defs.bzl", "py_binary", "py_library")
-load("@base_pip3//:requirements.bzl", base_entry_point = "entry_point")
+load("@base_pip3//:requirements.bzl", "requirement", base_entry_point = "entry_point")
 load("@aspect_bazel_lib//lib:jq.bzl", "jq")
 load("@aspect_bazel_lib//lib:yq.bzl", "yq")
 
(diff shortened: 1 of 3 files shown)


tools/base/envoy_python.bzl
```diff
@@ -54,4 +57,61 @@ def envoy_entry_point(
...
         data = (data or []),
+        deps = (deps or []),
+    )
+
+def envoy_py_data(name, src, format = None, entry_point = base_entry_point):
+    """Preload JSON/YAML data as a python lib.
+
+    Example:
+
+    ```starlark
```
phlax, line 68: ...example to follow...

tools/base/envoy_python.bzl
```diff
@@ -244,3 +244,90 @@ def envoy_genjson(name, srcs = [], yaml_srcs = [], filter = None, args = None):
         args = args,
         filter = filter,
     )
+
+def envoy_py_data(name, src, format = None, entry_point = base_entry_point):
+    """Preload JSON/YAML data as a python lib.
+
+    Data is loaded to python and then dumped to a pickle file.
+
+    A python lib is provided which exposes the pickled data.
```
htuch, line 253: This is very cool. Do we want to pickle into an in-memory proto object as well, e.g. factor out things like parsing YAML into JSON into protobuf and validating?

htuch (approved):

tools/base/envoy_python.bzl
phlax: not sure if pickling a proto would work - json works because it can be marshalled as python primitives - im guessing that is not the case with a proto object

in the case where we currently parse yaml -> json -> protobuf -> validate -> json - ie the protodoc manifest - i have created a pipeline such that the resulting json is made available via this macro

phlax: not entirely sure if i understood the suggestion here

re pickling - anything can theoretically be pickled - eg the ZODB is an object db entirely built on pickles (iirc), but you have to then provide marshals to load/dump types

re protobuf - i see that you can de/serialize descriptors, not sure but there might be a way the python objects could be pickled - from my reading of how protobuf works the python libs load from the descriptors dynamically - so _if_ you could pickle protobuf objects there may be some speed up by preloading this part. At the cost of considerable complexity.

Issue #21151, opened 2022-05-04
gRPC-JSON Transcoder doesn't url decode query parameter names


jeffsawatzky:
*Title*: gRPC-JSON Transcoder doesn't url decode query parameter names

*Description*:
When we have a query string param like `?search[field]=value` I was expecting that in the protobuf file we would do `json_name="search[field]"`, however we actually need to do `json_name="search%5Bfield%5D"`

*Repro steps*:
Define a gRPC service like so
```
message ListRequest {
  string search_field = 1 [json_name = "search[field]"];
}

message ListCallbacksResponse {
}

service Service {
  rpc List(ListRequest) returns (ListResponse) {
    option (google.api.http) = {
      get: "/v1/list"
    };
  }
}
```
make requests like
```
http://localhost/v1/list?search%5Bfield%5D=fieldvalue
```
or
```
http://localhost/v1/list?search[field]=fieldvalue
```
and note how it doesn't work.

Change the protobuf to look like:
```
message ListRequest {
  string search_field = 1 [json_name = "search%5Bfield%5D"];
}
```
and note how it starts to work.

This doesn't seem right to me, and the protobuf file shouldn't need to know to url encode the square brackets.

See this [Stack Overflow](https://stackoverflow.com/q/72118020/2233608) question.

lizan:
Yeah that's the current behavior. Shall we add url decode for that @qiwzhang?

jeffsawatzky:
For comparison, the [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/blob/2ce32afe5735d0bfa28ee3369e5975dfb8192313/runtime/query.go#L31) project uses the [`Values`](https://pkg.go.dev/net/url#Values) type from the `net/url` golang package and it [unescapes the key names](https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/url/url.go;l=933).

I'm not sure what Google's cloud endpoints do. But if it also unescapes the key name then it might make sense for envoy to do it too.

lizan:
OK thanks for the info. It makes sense for the transcoder to decode URL.

jeffsawatzky:
@lizan / @qiwzhang any update on this? I'd attempt to fix it myself however I haven't touched C++ is many many years...and I'm scared of the `malloc`/`free`. :)

lizan:
@qiwzhang it isn't that deep, we can urldecode before passing to protobuf?

qiwzhang:
[...]

PR #21563, opened 2022-06-02
grpc_json_transcoder: to support escaped json_name in query parameters

qiwzhang:

For https://github.com/envoyproxy/envoy/issues/21151

* Updated `grpc-httpjson-transcoding` repo to have this [change](https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/pull/69).  
* Add some unit tests for different json_name case:

a proto message field 
```
  // To test json name is percentage decoded
  string search_decoded = 3 [json_name = "search[decoded]"];

  // To test json name is percentage encoded
  string search_encoded = 4 [json_name = "search%5Bencoded%5D"];
```

with query parameter names in following
```
* search[decode]:   work
* search%5Bdecode%5D:  work

* search[encode] :     fail
* search%5Bencode%5D:    work
```


test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
```diff
@@ -264,6 +264,22 @@ TEST_P(GrpcJsonTranscoderIntegrationTest, TestParamUnescapePlus) {
       R"({"id":"20","theme":"Children"})");
 }
 
+TEST_P(GrpcJsonTranscoderIntegrationTest, QueryParams111) {
+  HttpIntegrationTest::initialize();
+  testTranscoding<bookstore::CreateShelfRequest, bookstore::Shelf>(
+      Http::TestRequestHeaderMapImpl{{":method", "POST"},
+                                     {":path", "/shelf?shelf.search[field]=Google"},
```
jeffsawatzky, line 271: As @lizan mentions in #21151 [here](https://github.com/envoyproxy/envoy/issues/21151#issuecomment-1145216299) `[` and `]` aren't valid characters in query params so they will need to be escaped (many browsers will do this for you automatically, so even if you type `[` and `]` in the query param the browser actually sends  `%5B` and `%5D`)

This means that the path envoy gets is actually "/shelf?shelf.search%5Bfield%5D=Google"

Does the test still pass if you use that as the path?

jeffsawatzky: [Here is the RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) and you will see that `[` and `]` are reserved characters, and if you want to use them in the query param you need to [percent encode them](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).

qiwzhang:
I think the test failure is flaky.  The error is 
```
Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/build/tmp/_bazel_envoybuild/b570b5ccd0454dc9af9f65ab1833764d/server/jvm.out')
```
Could someone help to retest it?  Thanks

test/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter_test.cc
```diff
@@ -274,6 +274,31 @@ TEST_F(GrpcJsonTranscoderConfigTest, InvalidQueryParameter) {
...
+      getProtoConfig(TestEnvironment::runfilesPath("test/proto/bookstore.descriptor"),
+                     "bookstore.Bookstore"),
+      *api_);
+
+  // When "json_name" is percent encoded, but the field name in query parameter
+  // is percent decoded, it will not match, transcoding fails.
+  // * json_name = "search%5Bencoded%5D", defined in "test/proto/bookstore.proto".
+  // " the query parameter is "search[encode]".
+  Http::TestRequestHeaderMapImpl headers{{":method", "POST"},
+                                         {":path", "/shelf?shelf.search[encode]=Google"}};
```
lizan, line 288:
```suggestion
                                         {":path", "/shelf?shelf.search[encoded]=Google"}};
```
otherwise it doesn't match test proto

qiwzhang: Done

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
qiwzhang: Fixed it

wrowe:
/retest
to let the windows artifacts finish compiling

lizan (approved):

jeffsawatzky:
@lizan / @qiwzhang when will this be released? I see there have been some releases since this was merged, but this change wasn't included in them. Thanks!

qiwzhang:
Probably will be in next Envoy release from top of three, v1.23.0.  Not sure when it will be.  Recent envoy releases are all security patches.

PR #21572, opened 2022-06-03
quic: early return encoding on a closed stream

danzh2010:
Commit Message: if encodeHeaders(), encodeData(), encodeTrailers() is called on a closed QUIC stream, which could only happen in problematic HCM and filter implementation, the current code would mess up QUIC flow control in a very unnoticeable way. This PR makes them to early return with ENVOY_BUG, so that the connection flow control can continue function and the error can be noted.

Additional Message: also change a bunch of ASSERT into ENVOY_BUG.

Risk Level: low
Testing: new unit tests
Docs Changes: N/A
Release Notes: N/A


RyanTheOptimist (approved):
Nice!

danzh2010:
Ping for merge?

PR #21573, opened 2022-06-03
logger: add method for verbosity log level update in fancy logger

botengyao:

This PR adds a method `updateVerbositySetting` to allow developers to update fine-grained loggers' level in runtime. 
* Stores the update info in `FancyContex` as a vector.
* For a slow path, it will check the stored verbosity updates during logger initialization. 
* The check will be returned when finding a match from left to right.
* Adds glob support for the verbosity log level updates in runtime.
  *  Supports file basename update, eg.  ("foo", 1), any files with basename `foo` will be updated to debug level.
  * The update should not include file extension.
  *  Supports glob `* `and `?` pattern, eg., ("foo/b*", 3), ("f??, 1).
* Hasn't added the hooks for the admin interface.
* For now it supports Linux/Unix file path pattern.

Commit Message: logger: add method for verbosity log level update in fancy logger
Additional Description:
Risk Level: low
Testing: unit tests
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/common/fancy_logger.h
```diff
@@ -62,28 +78,68 @@ class FancyContext {
...
+  spdlog::logger* createLogger(std::string key) ABSL_EXCLUSIVE_LOCKS_REQUIRED(fancy_log_lock_);
+
+  /**
+   * Append verbosity level updates to the VerbosityLogUpdateInfo vector.
+   */
+  void appendVerbosityLogUpdate(absl::string_view update_pattern, level_enum log_level)
+      ABSL_EXCLUSIVE_LOCKS_REQUIRED(fancy_log_lock_);
+
+  /**
+   * Returns the current log level of `file`. default log level is used if there is no
```
KBaichoo, line 123: default -> Default

source/common/common/fancy_logger.cc
```diff
@@ -109,24 +141,87 @@ void FancyContext::initSink() {
...
   new_logger->flush_on(level_enum::critical);
   fancy_log_map_->insert(std::make_pair(key, new_logger));
   return new_logger.get();
 }
 
+void FancyContext::updateVerbositySetting(std::vector<std::pair<absl::string_view, int>> updates) {
+  absl::WriterMutexLock ul(&fancy_log_lock_);
+  log_update_info_.clear();
+  for (const auto& [glob, level] : updates) {
+    if (level < 0 || level > 6) {
```
KBaichoo, line 163: these special numbers aren't that readable.

Also should these be saturated or simply ignored?

source/common/common/fancy_logger.cc
```diff
@@ -109,24 +141,87 @@ void FancyContext::initSink() {
...
+      continue;
+    }
+    appendVerbosityLogUpdate(glob, static_cast<level_enum>(level));
+  }
+
+  for (auto& [key, logger] : *fancy_log_map_) {
```
KBaichoo, line 169: This will be O(Loggers * verbosity log updates)

It will also re-do some work; perhaps we can avoid getting / setting log level for something that wasn't updated.

source/common/common/fancy_logger.h
```diff
@@ -14,6 +14,21 @@ using SpdLoggerSharedPtr = std::shared_ptr<spdlog::logger>;
 using FancyMap = absl::flat_hash_map<std::string, SpdLoggerSharedPtr>;
 using FancyMapPtr = std::shared_ptr<FancyMap>;
 using FancyLogLevelMap = absl::flat_hash_map<std::string, spdlog::level::level_enum>;
+using spdlog::level::level_enum;
+
+/**
+ * Data struct that stores the necessary verbosity log update info
+ */
+struct VerbosityLogUpdateInfo final {
```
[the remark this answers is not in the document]
botengyao: Yes, made them const.

source/common/common/fancy_logger.h
```diff
@@ -62,28 +78,68 @@ class FancyContext {
...
+   *
+   * It will store the current verbosity updates and clear all previous modifications for
+   * future check when initializing a new logger.
+   *
+   * Files are matched against globs in updates in order, and the first match determines
+   * the verbosity level.
+   *
+   * Files which do not match any pattern use the value of default log level form Context.
+   */
+  void updateVerbositySetting(std::vector<std::pair<absl::string_view, int>> updates)
```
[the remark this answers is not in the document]
botengyao: Done.

source/common/common/fancy_logger.h
```diff
@@ -14,6 +14,21 @@ using SpdLoggerSharedPtr = std::shared_ptr<spdlog::logger>;
 using FancyMap = absl::flat_hash_map<std::string, SpdLoggerSharedPtr>;
 using FancyMapPtr = std::shared_ptr<FancyMap>;
 using FancyLogLevelMap = absl::flat_hash_map<std::string, spdlog::level::level_enum>;
+using spdlog::level::level_enum;
```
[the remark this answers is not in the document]
botengyao: Gotcha, removed it.

source/common/common/fancy_logger.cc
botengyao: I am also looking into if there is a way to optimize it. 

This initial design is corresponding to the Google vmoudle behavior: it will override all previous updates whenever we get a new verbosity update, and support the file basename setting. Therefore at lease we need to check all the loggers during the verbosity update.

Let me know if you have any thoughts!

test/common/common/log_verbosity_update_test.cc
```diff
@@ -0,0 +1,229 @@
...
+  std::pair<absl::string_view, int> update_2 = std::make_pair("envoy/*/f??", 2);
+  getFancyContext().updateVerbositySetting({update_1, update_2});
+
+  SpdLoggerSharedPtr p = getFancyContext().getFancyLogEntry(file_1);
+  EXPECT_NE(p, nullptr);
+  EXPECT_EQ(p->level(), spdlog::level::warn);
+}
+
+TEST(Fancy, invalidLevelUpdate) {
+  const char* file_1 = "envoy/src/foo.cc";
```
KBaichoo, line 217: any reason using c-style string vs  std::string ?

source/common/common/fancy_logger.h
```diff
@@ -62,28 +77,69 @@ class FancyContext {
...
+   *
+   * It will store the current verbosity updates and clear all previous modifications for
+   * future check when initializing a new logger.
+   *
+   * Files are matched against globs in updates in order, and the first match determines
+   * the verbosity level.
+   *
+   * Files which do not match any pattern use the value of default log level from Context.
+   */
+  void updateVerbositySetting(const std::vector<std::pair<absl::string_view, int>> updates)
```
KBaichoo, line 94: pass the updates vector by reference

source/common/common/fancy_logger.cc
```diff
@@ -109,24 +141,86 @@ void FancyContext::initSink() {
...
   fancy_log_map_->insert(std::make_pair(key, new_logger));
   return new_logger.get();
 }
 
+void FancyContext::updateVerbositySetting(std::vector<std::pair<absl::string_view, int>> updates) {
+  absl::WriterMutexLock ul(&fancy_log_lock_);
+  log_update_info_.clear();
+  for (const auto& [glob, level] : updates) {
+    if (level < 0 || level >= static_cast<int>(ARRAY_SIZE(spdlog::level::level_string_views))) {
+      printf("This log level: %d is out of scope, and it should be in [0, 6]. Skipping.", level);
```
KBaichoo, line 164: Would probably be helpful to have the glob expression as well here.

source/common/common/fancy_logger.cc
```diff
@@ -109,24 +141,86 @@ void FancyContext::initSink() {
...
   new_logger->flush_on(level_enum::critical);
   fancy_log_map_->insert(std::make_pair(key, new_logger));
   return new_logger.get();
 }
 
+void FancyContext::updateVerbositySetting(std::vector<std::pair<absl::string_view, int>> updates) {
+  absl::WriterMutexLock ul(&fancy_log_lock_);
+  log_update_info_.clear();
+  for (const auto& [glob, level] : updates) {
+    if (level < 0 || level >= static_cast<int>(ARRAY_SIZE(spdlog::level::level_string_views))) {
```
KBaichoo, line 163: I think having constexpr constants defined in an an that describe the what would be more readable.

source/common/common/fancy_logger.h
```diff
@@ -62,28 +77,69 @@ class FancyContext {
...
    */
   void initSink();
 
   /**
-   * Creates a logger given key and log level, and add it to map.
+   * Creates a logger given key, and add it to map. Log level is from getLogLevel.
    * Key is the log component name, e.g. file name now.
    */
-  spdlog::logger* createLogger(std::string key, int level = -1)
+  spdlog::logger* createLogger(std::string key) ABSL_EXCLUSIVE_LOCKS_REQUIRED(fancy_log_lock_);
```
KBaichoo, line 113: while we're at it, make this a const ref string

source/common/common/fancy_logger.cc
KBaichoo: We can go with it then for now since this is expected to rarely occur and we can revisit if needed.

test/common/common/log_verbosity_update_test.cc
```diff
@@ -0,0 +1,229 @@
...
+#include "source/common/common/logger.h"
+
+#include "test/test_common/logging.h"
+
+#include "absl/strings/string_view.h"
+#include "gtest/gtest.h"
+
+namespace Envoy {
+
+TEST(Fancy, safeFileNameMatch) {
```
KBaichoo, line 13: Seems like you have 3 different sections can you comment over them what each section is testing

test/common/common/log_verbosity_update_test.cc
```diff
@@ -0,0 +1,229 @@
...
+  EXPECT_NE(p, nullptr);
+  EXPECT_EQ(p->level(), spdlog::level::warn);
+}
+
+TEST(Fancy, invalidLevelUpdate) {
+  const char* file_1 = "envoy/src/foo.cc";
+  std::atomic<spdlog::logger*> flogger{nullptr};
+  getFancyContext().initFancyLogger(file_1, flogger);
+
+  std::pair<absl::string_view, int> update_1 = std::make_pair("f??", 9);
```
KBaichoo, line 221: const (here and elsewhere if not mutating the value)

test/common/common/log_verbosity_update_test.cc
```diff
@@ -0,0 +1,229 @@
...
+}
+
+TEST(Fancy, globQuestionUpdate) {
+  const char* file_1 = "envoy/src/foo.cc";
+  const char* file_2 = "envoy/f__.cc";
+  const char* file_3 = "/bar.cc";
+  std::atomic<spdlog::logger*> flogger{nullptr};
+  getFancyContext().initFancyLogger(file_1, flogger);
+  getFancyContext().initFancyLogger(file_2, flogger);
+  getFancyContext().initFancyLogger(file_3, flogger);
```
KBaichoo, line 177: are these not clobbering the same flogger?

test/common/common/log_verbosity_update_test.cc
```diff
@@ -0,0 +1,229 @@
...
+}
+
+TEST(Fancy, globStarUpdate) {
+  const char* file_1 = "envoy/src/foo.cc";
+  const char* file_2 = "envoy/src/bar.cc";
+  const char* file_3 = "envoy/baz.cc";
+  std::atomic<spdlog::logger*> flogger{nullptr};
+  getFancyContext().initFancyLogger(file_1, flogger);
+  getFancyContext().initFancyLogger(file_2, flogger);
+  getFancyContext().initFancyLogger(file_3, flogger);
```
KBaichoo, line 114: likewise as below

KBaichoo (commented):
looks great, here another pass

source/common/common/fancy_logger.h
botengyao: got it & good catch

source/common/common/fancy_logger.cc
botengyao: Done

source/common/common/fancy_logger.cc
botengyao: Good point. Done

source/common/common/fancy_logger.h
botengyao: Done

test/common/common/log_verbosity_update_test.cc
botengyao: Added comments and re-arranged them.

test/common/common/log_verbosity_update_test.cc
botengyao: gotcha

test/common/common/log_verbosity_update_test.cc
botengyao: Nope. actually this can be static I am thinking, and we still stored the underlying loggers.

botengyao:
Thanks @KBaichoo for reviewing it!

test/common/common/log_verbosity_update_test.cc
```diff
@@ -0,0 +1,234 @@
...
+  EXPECT_EQ(p->level(), Logger::Context::getFancyDefaultLevel());
+}
+
+TEST(Fancy, inOrderUpdate) {
+  const std::string file_1 = "envoy/src/foo.cc";
+  std::atomic<spdlog::logger*> flogger{nullptr};
+  getFancyContext().initFancyLogger(file_1, flogger);
+
+  const std::pair<absl::string_view, int> update_1 = std::make_pair("f??", 3);
+  const std::pair<absl::string_view, int> update_2 = std::make_pair("envoy/*/f??", 2);
```
KBaichoo, line 213: would be more useful to to have these be `SPDLOG_TO_INT(spdlog::level::warn)` as the int <-> enum conversion to levels isn't obvious. Here and elsewhere

test/common/common/log_verbosity_update_test.cc
KBaichoo: ah, it does clobber the flogger value, but we end up logger pointer in the map for the others. So it's fine.

KBaichoo (commented):
lgtm otherwise, nice work

source/common/common/fancy_logger.h
```diff
@@ -15,8 +15,11 @@ using FancyMap = absl::flat_hash_map<std::string, SpdLoggerSharedPtr>;
 using FancyMapPtr = std::shared_ptr<FancyMap>;
 using FancyLogLevelMap = absl::flat_hash_map<std::string, spdlog::level::level_enum>;
 
+constexpr static int LOGLEVEL_MAX = 6; // spdlog level is in [0, 6].
+constexpr static int LOGLEVEL_MIN = 0;
```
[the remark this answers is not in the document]
botengyao: Got it. Thank you for pointing it out. Done Looks like there are multiple code styles in Enovy ;-).

test/common/common/log_verbosity_update_test.cc
botengyao: Done

mattklein123 (approved):
Thanks.

If we are going to do more work on this can we bulk rename "fancy" to something else? We discussed this when the code was originally written but the intern ran out of time IIRC. This can be done in a follow up PR.

I'm open to any name that is a bit more descriptive/professional. DetailedLogger or something else? Thank you!

botengyao:
Good point. If we decide to replace the name, we'd better to stop some further work here, and bulk change `FANCY` at first.  My proposal is to replace it with `FINE_GRAIN`, eg.

`FancyContext` -> `FineGrainLogContext`
`FANCY_LOG(debug, ...)` -> `FINE_GRAIN_LOG(debug, ...)`
`FANCY_CONN_LOG` -> `FINE_GRAIN_CONN_LOG`
`enableFancyLogger()` -> `enableFineGrainLogger()`
 -> `getFineGrainLogContext().getFineGrainLogEntry()`
`...`

This is also corresponding to the option: --enable-fine-grain-logging. 

Does that sound good if we decide to replace them all? @mattklein123 @KBaichoo

KBaichoo:
That naming scheme sounds good to me... much better than what it currently is.

mattklein123:
> @mattklein123 @KBaichoo Does that sound good if we decide to replace them all?

+1 SGTM. Please do the replace now and then we can keep going. Please also check to see if "fancy" shows up in the docs anywhere. Ultimately we should document how to enable this.

Issue #21553, opened 2022-06-02
CORS Access-Control-Allow-Private-Network


t0rr3sp3dr0:
*Title*: CORS Access-Control-Allow-Private-Network

*Description*:

Add a new `allow_private_network` option to `config.route.v3.CorsPolicy` that will make Envoy respond requests with the header `Access-Control-Allow-Private-Network: true` when the header `Access-Control-Request-Private-Network: true` is present on OPTIONS requests.

> Chrome is deprecating direct access to private network endpoints from public websites as part of the [Private Network Access](https://wicg.github.io/private-network-access/) (PNA) specification.
> 
> Chrome will start sending a CORS preflight request ahead of any [private network request](https://developer.chrome.com/blog/private-network-access-preflight/#what-is-private-network-access) for a subresource, which asks for explicit permission from the target server. This preflight request will carry a new header, Access-Control-Request-Private-Network: true, and the response to it must carry a corresponding header, Access-Control-Allow-Private-Network: true.

*Relevant Links*:
> https://developer.chrome.com/blog/private-network-access-preflight/

KBaichoo:
cc @daixiang0 @wbpcode

daixiang0:
I will take a look.

daixiang0:
@envoyproxy/envoy-maintainers please close as it is fixed.

bruce-c-liu:
Hi when will this be released?

adisuissa:
> Hi when will this be released?

IIUC this will be part of 1.24.0, and according to the [releases schedule](https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#major-release-schedule) the version should be cut in a couple of weeks.

PR #21581, opened 2022-06-06
Cors: add PNA support

daixiang0:
Add a PNA header check refer to https://developer.chrome.com/blog/private-network-access-preflight/.

Commit Message: 
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA] #21553
[Optional Deprecated:]


daixiang0:
Since changes are related to v2 API, switch to runtime guard for now.

daixiang0:
kindly ping @adisuissa

source/extensions/filters/http/cors/cors_filter.h
```diff
@@ -88,6 +88,7 @@ class CorsFilter : public Http::StreamFilter {
   const std::string& exposeHeaders();
   const std::string& maxAge();
   bool allowCredentials();
+  bool allowPrivateNetworkAccess();
```
wbpcode, line 91: Where is the implementation of this method?

source/extensions/filters/http/cors/cors_filter.cc
```diff
@@ -125,6 +131,12 @@ Http::FilterHeadersStatus CorsFilter::decodeHeaders(Http::RequestHeaderMap& head
     response_headers->setInline(access_control_max_age_handle.handle(), maxAge());
   }
 
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.cors_private_network_access")) {
+    if (headers.getInlineValue(access_control_request_private_network_handle.handle()) == "true") {
+      response_headers->setInline(access_control_response_private_network_handle.handle(), "true");
+    }
+  }
```
wbpcode, line 138: I didn't get why this runtime guard is necessary 🤔 If it's necessary indeed, can you add some more comments here?

wbpcode (changes requested):
Thanks for you contributions. Some comments are added. And I think may be we should add a new API in the `envoy.config.route.v3.CorsPolicy`, right?

source/extensions/filters/http/cors/cors_filter.h
daixiang0: In fact I delete it but not clean all :(

source/extensions/filters/http/cors/cors_filter.cc
daixiang0: Since I am not sure to add it to API, so add runtime guard here.

wbpcode: IMO, we should add a new field in the `envoy.config.route.v3.CorsPolicy` for this new feature. cc @envoyproxy/api-shepherds for some more inputs.

changelogs/current.yaml
```diff
@@ -186,6 +186,9 @@ new_features:
 - area: dubbo_proxy
   change: |
     added :ref:`metadata_match <envoy_v3_api_field_extensions.filters.network.dubbo_proxy.v3.RouteAction.metadata_match>` support to the dubbo proxy.
+- area: cors
+  change: |
+    added support for cors PNA. This behavioral change can be temporarily reverted by setting runtime guard ``envoy_reloadable_features_cors_private_network_access`` to false.
```
adisuissa, line 191: Please add a link to `https://developer.chrome.com/blog/private-network-access-preflight/`

adisuissa (commented):
Not sure I grok the intention behind this feature, but it seems to address what's described in: https://developer.chrome.com/blog/private-network-access-preflight/

Probably no need for guarding it behind a runtime-flag.

source/extensions/filters/http/cors/cors_filter.cc
daixiang0: Sure.

source/extensions/filters/http/cors/cors_filter.cc
```diff
@@ -125,6 +131,12 @@ Http::FilterHeadersStatus CorsFilter::decodeHeaders(Http::RequestHeaderMap& head
     response_headers->setInline(access_control_max_age_handle.handle(), maxAge());
   }
 
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.cors_private_network_access")) {
+    if (headers.getInlineValue(access_control_request_private_network_handle.handle()) == "true") {
+      response_headers->setInline(access_control_response_private_network_handle.handle(), "true");
```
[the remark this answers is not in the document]
daixiang0: According to https://developer.chrome.com/blog/private-network-access-preflight/#server-side-requests:
> When your server receives a preflight request (an OPTIONS request with CORS headers), the server should check for the presence of an Access-Control-Request-Private-Network: true header. If this header is present on the request, the server should examine the Origin header and the request path along with any other relevant information (such as Access-Control-Request-Headers) to ensure the request is safe to allow. Typically, you should allow access to a single origin under your control.

I think we do not need to check the destination is more private than the origin since:
1. if it has this header, which means it knows it is requesting one more private network, and we have checked the Origin header and the request path along with any other relevant information (such as Access-Control-Request-Headers), we can just ensure it is safe to allow and send the response header
2. if it has not, the response header added or not become meaningless


> Chrome will start sending a CORS preflight request ahead of any [private network request](https://developer.chrome.com/blog/private-network-access-preflight/#what-is-private-network-access) for a subresource, which asks for explicit permission from the target server. This preflight request will carry a new header, Access-Control-Request-Private-Network: true, and the response to it must carry a corresponding header, Access-Control-Allow-Private-Network: true.

Now obviously this CORS header is added by **Chrome**  only for verification. I think this pair of headers is an extra check. I am not the CORS expert, if any misunderstandings, feel free to point them out, thanks.

changelogs/current.yaml
daixiang0: Added.

source/extensions/filters/http/cors/cors_filter.cc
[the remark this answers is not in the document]
adisuissa: Not an expert on CORS as well.
If I read correctly the [spec proposal](https://wicg.github.io/private-network-access/#private-network-request-heading), I think that by sending back the `Access-Control-Allow-Private-Network: true` header, it implies that the server allows the resource to be shared with external networks, and that might not be the right thing to do.

@mattklein123 can you recommend a reviewer that can take a look at CORS from the server perspective?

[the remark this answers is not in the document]
mattklein123: > @mattklein123 can you recommend a reviewer that can take a look at CORS from the server perspective?

Not really. @RyanTheOptimist is probably our best bet as someone that is more familiar with web standards?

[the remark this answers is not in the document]
RyanTheOptimist: It looks like the implementation has changed since this discussion started. Initially there was a runtime feature to control sending `Access-Control-Allow-Private-Network: true` in response to `Access-Control-Request-Private-Network: true`. I agree that this does not seem correct as it would not allow a server to prevent public access to resources that should otherwise be private access only. However, the implementation now appears to check a CorsPolicy allow_private_network_access value, and only if that is true will it send `Access-Control-Allow-Private-Network: true`. This behavior sounds correct to me.

source/common/runtime/runtime_features.cc
```diff
@@ -40,6 +40,7 @@ RUNTIME_GUARD(envoy_reloadable_features_cares_accept_nodata);
 RUNTIME_GUARD(envoy_reloadable_features_conn_pool_delete_when_idle);
 RUNTIME_GUARD(envoy_reloadable_features_conn_pool_new_stream_with_early_data_and_http3);
 RUNTIME_GUARD(envoy_reloadable_features_correctly_validate_alpn);
+RUNTIME_GUARD(envoy_reloadable_features_cors_private_network_access);
```
adisuissa, line 43: Is this still needed?

source/common/router/config_impl.cc
```diff
@@ -378,6 +378,10 @@ CorsPolicyImpl::CorsPolicyImpl(const envoy::config::route::v3::CorsPolicy& confi
   if (config.has_allow_credentials()) {
     allow_credentials_ = PROTOBUF_GET_WRAPPED_REQUIRED(config, allow_credentials);
   }
+  if (config.has_allow_private_network_access()) {
+    allow_private_network_access_ =
+        PROTOBUF_GET_WRAPPED_REQUIRED(config, allow_private_network_access);
+  }
```
adisuissa, line 384: What's the benefit of using this over `PROTOBUF_GET_WRAPPED_OR_DEFAULT` without an optional value?

api/envoy/config/route/v3/route_components.proto
```diff
@@ -655,6 +655,12 @@ message CorsPolicy {
   // Envoy will lookup the runtime key to get the percentage of requests for which it will evaluate
   // and track the request's *Origin* to determine if it's valid but will not enforce any policies.
   core.v3.RuntimeFractionalPercent shadow_enabled = 10;
+
+  // Specify whether allow requests whose target server's IP address is more private than that from
+  // which the request initiator was fetched.
+  //
+  // More details refer to https://developer.chrome.com/blog/private-network-access-preflight.
+  google.protobuf.BoolValue allow_private_network_access = 12;
```
adisuissa, line 663: What is the default value?

source/extensions/filters/http/cors/cors_filter.cc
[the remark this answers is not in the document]
adisuissa: IIRC the previous implementation which I commented on was essentially checking if the `Access-Control-Request-Private-Network: true` is sent as part of the request, and then automatically added `Access-Control-Allow-Private-Network: true` in the response. The question was whether there should be any other validation, because it seems to me that just adding a response to a request doesn't really help with anything.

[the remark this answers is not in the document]
RyanTheOptimist: Does the current implementation seem correct to you? I would expect that by default, allow_private_network_access would be false and hence `Access-Control-Allow-Private-Network: true` would not be sent, and hence Chrome would block requests to this origin. Does that accurately capture the current implementation? Are you thinking that perhaps we should have a config which specifies a list of origins for which private network access it allowed and check the Origin: header against that list?

source/common/runtime/runtime_features.cc
daixiang0: Not, I will remove it.

api/envoy/config/route/v3/route_components.proto
daixiang0: `google.protobuf.BoolValue` is false as default, see https://developers.google.com/protocol-buffers/docs/proto#optional

source/common/router/config_impl.cc
daixiang0: Thanks for the suggestion, use `PROTOBUF_GET_WRAPPED_OR_DEFAULT` is better here.

source/extensions/filters/http/cors/cors_filter.cc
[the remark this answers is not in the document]
daixiang0: @RyanTheOptimist yep, it is what exactly implementation did.

@adisuissa wait for your input :)

daixiang0:
Also please remove `v2-freeze` label, now it is unrelated.

daixiang0:
@adisuissa friendly ping then I can update.

daixiang0:
@adisuissa friendly ping

wrowe:
@adisuissa and @wbpcode, ping for when you are able (with all consideration for US holiday and pto, final approval may still be a bit delayed).

phlax:
@daixiang0 looks like this needs a main merge

@adisuissa @wbpcode would you mind reviewing

/wait-any

daixiang0:
@phlax thanks for the tip, I merge the latest main branch.

changelogs/current.yaml
```diff
@@ -13,5 +13,107 @@ removed_config_or_runtime:
...
 
 new_features:
+- area: access_log
+  change: |
+    added new access_log command operators to retrieve upstream connection information change: ``%UPSTREAM_PROTOCOL%``, ``%UPSTREAM_PEER_SUBJECT%``, ``%UPSTREAM_PEER_ISSUER%``, ``%UPSTREAM_TLS_SESSION_ID%``, ``%UPSTREAM_TLS_CIPHER%``, ``%UPSTREAM_TLS_VERSION%``, ``%UPSTREAM_PEER_CERT_V_START%``, ``%UPSTREAM_PEER_CERT_V_END%``, ``%UPSTREAM_PEER_CERT%` and ``%UPSTREAM_FILTER_STATE%``.
+  change: |
+    added configuration for OpenTelemetry :ref:`resource_attributes <envoy_v3_api_field_extensions.access_loggers.open_telemetry.v3.OpenTelemetryAccessLogConfig.resource_attributes>`.
+- area: dns_resolver
+  change: |
+    added :ref:`include_unroutable_families<envoy_v3_api_field_extensions.network.dns_resolver.apple.v3.AppleDnsResolverConfig.include_unroutable_families>` to the Apple DNS resolver.
```
wbpcode, line 23: May be this is a minor mistake?

test/common/router/config_impl_test.cc
```diff
@@ -6692,6 +6693,7 @@ TEST_F(RoutePropertyTest, TestVHostCorsConfig) {
   EXPECT_EQ(cors_policy->exposeHeaders(), "test-expose-headers");
   EXPECT_EQ(cors_policy->maxAge(), "test-max-age");
   EXPECT_EQ(cors_policy->allowCredentials(), true);
+  EXPECT_EQ(cors_policy->allowPrivateNetworkAccess(), true);
```
wbpcode, line 6696: May a simple test to ensure `cors_policy->allowPrivateNetworkAccess()` is `false` when we have no `allow_private_network_access` or `allow_private_network_access` is set to `false`?

test/extensions/filters/http/cors/cors_filter_test.cc
```diff
@@ -758,6 +759,72 @@ TEST_F(CorsFilterTest, OptionsRequestWithWildcardAllowHeaders) {
   EXPECT_EQ(Http::FilterTrailersStatus::Continue, filter_.encodeTrailers(response_trailers_));
 }
 
+TEST_F(CorsFilterTest, OptionsRequestMatchingOriginByWildcardWithPNA) {
+  Http::TestRequestHeaderMapImpl request_headers{
+      {":method", "OPTIONS"},
+      {"origin", "test-host"},
+      {"access-control-request-method", "GET"},
+      {"access-control-request-private-network", "true"}};
```
wbpcode, line 767: May we test a additional case that the request has no `access-control-request-private-network` header or the `access-control-request-private-network` header with invalid value?

test/extensions/filters/http/cors/cors_filter_test.cc
```diff
@@ -758,6 +759,72 @@ TEST_F(CorsFilterTest, OptionsRequestWithWildcardAllowHeaders) {
...
+  EXPECT_EQ(Http::FilterTrailersStatus::Continue, filter_.decodeTrailers(request_trailers_));
+
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue, filter_.encodeHeaders(response_headers_, false));
+  EXPECT_EQ(Http::FilterDataStatus::Continue, filter_.encodeData(data_, false));
+  EXPECT_EQ(Http::FilterTrailersStatus::Continue, filter_.encodeTrailers(response_trailers_));
+  ASSERT_TRUE(decoder_callbacks_.stream_info_.responseCodeDetails().has_value());
+  EXPECT_EQ(decoder_callbacks_.stream_info_.responseCodeDetails().value(), "cors_response");
+}
+
+TEST_F(CorsFilterTest, OptionsRequestMatchingOriginByWildcardWithPNAWithRuntimeGuard) {
```
wbpcode, line 794: Do we have a runtime guard for this?

wbpcode (changes requested):
Thanks for your contribution. And some comments are added.

changelogs/current.yaml
daixiang0: Yes, deleted.

test/extensions/filters/http/cors/cors_filter_test.cc
daixiang0: No, delete.

test/common/router/config_impl_test.cc
daixiang0: Add it in cors tests.

test/extensions/filters/http/cors/cors_filter_test.cc
```diff
@@ -758,6 +759,104 @@ TEST_F(CorsFilterTest, OptionsRequestWithWildcardAllowHeaders) {
...
+  EXPECT_EQ(Http::FilterTrailersStatus::Continue, filter_.decodeTrailers(request_trailers_));
+
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue, filter_.encodeHeaders(response_headers_, false));
+  EXPECT_EQ(Http::FilterDataStatus::Continue, filter_.encodeData(data_, false));
+  EXPECT_EQ(Http::FilterTrailersStatus::Continue, filter_.encodeTrailers(response_trailers_));
+  ASSERT_TRUE(decoder_callbacks_.stream_info_.responseCodeDetails().has_value());
+  EXPECT_EQ(decoder_callbacks_.stream_info_.responseCodeDetails().value(), "cors_response");
+}
+
+TEST_F(CorsFilterTest, OptionsRequestMatchingOriginByWildcardWithInvalidPNA) {
```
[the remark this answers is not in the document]
wbpcode: > May a simple test to ensure cors_policy->allowPrivateNetworkAccess() is false when we have no allow_private_network_access or allow_private_network_access is set to false?

I mean a simple test case to ensure the `allowPrivateNetworkAccess` method can return `false` correctly when we have no allow_private_network_access or allow_private_network_access is set to false in the proto config.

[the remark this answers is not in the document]
daixiang0: Thanks for the detailed explanation, I have updated it.

test/common/router/config_impl_test.cc
```diff
@@ -6692,62 +6693,62 @@ TEST_F(RoutePropertyTest, TestVHostCorsConfig) {
...
 }
 
-TEST_F(RoutePropertyTest, TestRouteCorsConfig) {
+TEST_F(RoutePropertyTest, TestVHostCorsConfigDisablePNA) {
   const std::string yaml = R"EOF(
 virtual_hosts:
   - name: "default"
     domains: ["*"]
+    cors:
+      allow_private_network_access: false
```
wbpcode, line 6705: Could we keep the original test case here and just add a new simple case for `DisablePNA`?

wbpcode (commented):
Thanks for all your contributions and patience. There all still some minor comments.

And why you remove the case that test the invalid PNA header value? 🤣 

```
 Http::TestResponseHeaderMapImpl response_headers{
      {":status", "200"},
      {"access-control-allow-origin", "test-host"},
      {"access-control-allow-methods", "GET"},
      {"access-control-allow-headers", "content-type"},
      {"access-control-max-age", "0"},
      {"access-control-allow-private-network", "invalid"},
  };
```

test/extensions/filters/http/cors/cors_filter_test.cc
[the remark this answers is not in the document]
wbpcode: @daixiang0 Sorry for my unclear expression. The `I think this isn't what I want` is mean the case is not used to ensure the  `allowPrivateNetworkAccess` method can return false correctly. 

But this case is still necessary to ensure only correct PNA value can be accepted.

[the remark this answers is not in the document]
daixiang0: I re-add it.

test/common/router/config_impl_test.cc
```diff
@@ -6750,6 +6750,69 @@ TEST_F(RoutePropertyTest, TestRouteCorsConfig) {
...
+          cors:
+            allow_private_network_access: invalid
+)EOF";
+
+  TestConfigImpl config(parseRouteConfigurationFromYaml(yaml), factory_context_, false);
+
+  const Router::CorsPolicy* cors_policy =
+      config.route(genHeaders("api.lyft.com", "/api", "GET"), 0)->routeEntry()->corsPolicy();
+
+  EXPECT_EQ(cors_policy->allowCredentials(), false);
```
wbpcode, line 6792: Is the `allow_private_network_access: invalid` accepted for the proto validator?

test/common/router/config_impl_test.cc
```diff
@@ -6750,6 +6750,69 @@ TEST_F(RoutePropertyTest, TestRouteCorsConfig) {
   EXPECT_EQ(cors_policy->allowCredentials(), true);
```
wbpcode, line 6750: I think you can just add a `EXPECT_EQ(cors_policy->allowPrivateNetworkAccess(), false);` after here, then that is enough.

test/common/router/config_impl_test.cc
daixiang0: Yeah, you are totally right, I will update it.

daixiang0:
@wbpcode in `TestVHostCorsConfig`, I add PNA enable case, in `TestRouteCorsConfig`, I add PNA empty case.

Also add `TestCorsConfigWithInvalidPNA` for invalid case.

Do I need to add PNA disable case?

wbpcode:
> @wbpcode in `TestVHostCorsConfig`, I add PNA enable case, in `TestRouteCorsConfig`, I add PNA empty case.
> 
> Also add `TestCorsConfigWithInvalidPNA` for invalid case.
> 
> Do I need to add PNA disable case?

IMO, that's enough.

test/common/router/config_impl_test.cc
```diff
@@ -6748,6 +6750,24 @@ TEST_F(RoutePropertyTest, TestRouteCorsConfig) {
...
+    routes:
+      - match:
+          prefix: "/api"
+        route:
+          cors:
+            allow_private_network_access: invalid
+)EOF";
+
+  EXPECT_THROW_WITH_REGEX(parseRouteConfigurationFromYaml(yaml), EnvoyException,
+                          "Unable to parse JSON as proto");
```
wbpcode, line 6770: nit: I think this case is unnecessary.

wbpcode (dismissed):
LGTM with nit comment. Thanks.

test/common/router/config_impl_test.cc
daixiang0: Deleted.

daixiang0:
@phlax can we enable debug in verify step? The [log](https://dev.azure.com/cncf/envoy/_build/results?buildId=113577&view=logs&j=ccaf6c92-47bd-5471-a9ce-c38c6e7e1730&t=a591d59f-9a2b-5865-b00a-820411986fd0&l=269) is not helpful enough, I am not sure it is a network issue.

daixiang0:
@envoyproxy/api-shepherd for anther look. Thank you!

wbpcode:
cc @envoyproxy/api-shepherds

adisuissa (commented):
Apologies for dropping the ball on this.
/lgtm api
Will defer to @wbpcode for feature usage correctness.

daixiang0:
@wbpcode @adisuissa please review again, and remove "v2-freeze" label which is misleading.

source/common/router/config_impl.cc
```diff
@@ -384,6 +384,8 @@ CorsPolicyImpl::CorsPolicyImpl(const envoy::config::route::v3::CorsPolicy& confi
   if (config.has_allow_credentials()) {
     allow_credentials_ = PROTOBUF_GET_WRAPPED_REQUIRED(config, allow_credentials);
   }
+  allow_private_network_access_ =
+      PROTOBUF_GET_WRAPPED_OR_DEFAULT(config, allow_private_network_access, false);
```
wbpcode, line 388: The `allow_private_network_access_` always has a value, may we shouldn't use `absl::optional` for it?

daixiang0: Sure.

daixiang0:
kindly ping @wbpcode

wbpcode (approved):
LGTM. Thanks.

PR #18265, opened 2021-09-26
Assert the case of both read and closed event registered

soulxu:
Signed-off-by: He Jie Xu <hejie.xu@intel.com>

Commit Message: Assert the case of both read and closed event registered
Additional Description:
The Read event may be removed when both the read and closed event registered. That will lead the thread is waiting for next read event forever.
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

PR #21585, opened 2022-06-06
[WIP] io: enable closed event for listener filter

soulxu:

Commit Message: io: enable the closed event for listener filter
Additional Description:
Previously the PR https://github.com/envoyproxy/envoy/pull/18265 removed the `Closed` event for listener filter, since the Windows doesn't work well.

@ggreenway was found there is still have a corner case that can't be handled without the `Closed` event https://github.com/envoyproxy/envoy/pull/18265#issuecomment-1033180816

This PR tries to introduce the `Closed` event back to the listener filter. As the comments on Windows code said, both `Read` and `Closed` can't be registered at same time due to the `Connection` doesn't support that yet. That means it shouldn't be a problem for the listener filter. Also currently, there is no Envoy code using both `Read` and `Closed` event, so it should be ok to remove those workaround code to make the both `Read` and `Closed` events registered to work with listener filter.

Risk Level: high
Testing: unittest should be added
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Windows platform code changed.

### DIFF
--- a/source/common/event/file_event_impl.cc
+++ b/source/common/event/file_event_impl.cc
@@ -55,9 +55,7 @@ void FileEventImpl::activate(uint32_t events) {
 void FileEventImpl::assignEvents(uint32_t events, event_base* base) {
   ASSERT(dispatcher_.isThreadSafe());
   ASSERT(base != nullptr);
-  // TODO(antoniovicente) remove this once ConnectionImpl can
-  // handle Read and Close events delivered together.
-  ASSERT(!((events & FileReadyType::Read) && (events & FileReadyType::Closed)));
+
   enabled_events_ = events;
   event_assign(
       &raw_event_, base, fd_,
@@ -137,10 +135,6 @@ void FileEventImpl::registerEventIfEmulatedEdge(uint32_t event) {
     ASSERT((event & (FileReadyType::Read | FileReadyType::Write)) == event);
     if (trigger_ == FileTriggerType::EmulatedEdge) {
       auto new_event_mask = enabled_events_ | event;
-      if (event & FileReadyType::Read && (enabled_events_ & FileReadyType::Closed)) {
-        // We never ask for both early close and read at the same time.
-        new_event_mask = new_event_mask & ~FileReadyType::Read;
-      }
       updateEvents(new_event_mask);
     }
   }
@@ -149,15 +143,6 @@ void FileEventImpl::registerEventIfEmulatedEdge(uint32_t event) {
 void FileEventImpl::mergeInjectedEventsAndRunCb(uint32_t events) {
   ASSERT(dispatcher_.isThreadSafe());
   if (injected_activation_events_ != 0) {
-    // TODO(antoniovicente) remove this adjustment to activation events once ConnectionImpl can
-    // handle Read and Close events delivered together.
-    if constexpr (PlatformDefaultTriggerType == FileTriggerType::EmulatedEdge) {
-      if (events & FileReadyType::Closed && injected_activation_events_ & FileReadyType::Read) {
-        // We never ask for both early close and read at the same time. If close is requested
-        // keep that instead.
-        injected_activation_events_ = injected_activation_events_ & ~FileReadyType::Read;
-      }
-    }
     events |= injected_activation_events_;
     injected_activation_events_ = 0;
     activation_cb_->cancel();
--- a/source/common/network/listener_filter_buffer_impl.cc
+++ b/source/common/network/listener_filter_buffer_impl.cc
@@ -18,7 +18,7 @@ ListenerFilterBufferImpl::ListenerFilterBufferImpl(IoHandle& io_handle,
 
   io_handle_.initializeFileEvent(
       dispatcher_, [this](uint32_t events) { onFileEvent(events); },
-      Event::PlatformDefaultTriggerType, Event::FileReadyType::Read);
+      Event::PlatformDefaultTriggerType, Event::FileReadyType::Read | Event::FileReadyType::Closed);
 }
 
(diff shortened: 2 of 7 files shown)


soulxu:
/wait

Running some tests on the CI first.

source/common/event/file_event_impl.cc
```diff
@@ -149,15 +143,6 @@ void FileEventImpl::registerEventIfEmulatedEdge(uint32_t event) {
 void FileEventImpl::mergeInjectedEventsAndRunCb(uint32_t events) {
   ASSERT(dispatcher_.isThreadSafe());
   if (injected_activation_events_ != 0) {
-    // TODO(antoniovicente) remove this adjustment to activation events once ConnectionImpl can
-    // handle Read and Close events delivered together.
-    if constexpr (PlatformDefaultTriggerType == FileTriggerType::EmulatedEdge) {
-      if (events & FileReadyType::Closed && injected_activation_events_ & FileReadyType::Read) {
-        // We never ask for both early close and read at the same time. If close is requested
-        // keep that instead.
-        injected_activation_events_ = injected_activation_events_ & ~FileReadyType::Read;
```
soulxu, line 158: I wasn't sure why the windows remove the read event, not sure there is the reason why windows can't handle both read and closed events. at least the CI works.

If nobody knows the history or background, a safer way is only register both read and closed events for non-Windows platforms, just silent ignore closed events when both read and closed events registered

source/common/event/file_event_impl.cc
```diff
@@ -149,15 +143,6 @@ void FileEventImpl::registerEventIfEmulatedEdge(uint32_t event) {
 void FileEventImpl::mergeInjectedEventsAndRunCb(uint32_t events) {
   ASSERT(dispatcher_.isThreadSafe());
   if (injected_activation_events_ != 0) {
-    // TODO(antoniovicente) remove this adjustment to activation events once ConnectionImpl can
-    // handle Read and Close events delivered together.
```
soulxu, line 153: As this comment said, this code is for `ConnectionImpl` can't works with read and close events. So it doesn't say it won't works with listener filter.

soulxu:
Also, note we don't have to fix this if we think this corner case doesn't worth taking the risk.

yanavlasov:
My understanding is that this issue is not deterministic and is just specific to Windows. If we do not have data that the Windows issue was addressed, then it seems to be unwise to revert this fix.

source/common/event/file_event_impl.cc
ggreenway: @davinci26 do you remember the history of this?

soulxu:
> Please add a test for the case that this fixes (client closes connection before listener filter has read enough data, but after it has written some bytes). I want to see if that test passes on windows in CI.
> 

Thanks, done. I thought I have the test and close event handle in the code, but I'm wrong. Add them in the new commit.

source/common/event/file_event_impl.cc
davinci26: I think it was an implementation detail based on how event registration worked at that point. @rectified95 works now at Microsoft so he should be able to test event registrations and which events get raised.

soulxu:
Emm...it is a real failure on macos. @ggreenway do you know if this is a different behavior on macos?

ggreenway:
@soulxu I looked through the libevent source. The feature we're wanting is `EV_CLOSE`, which can be checked with `EV_FEATURE_EARLY_CLOSE`. It appears that this is only supported with `epoll`, and MacOS uses `kqueue`.

soulxu:
> EV_FEATURE_EARLY_CLOSE

Thanks for the help! Let me add a check for EV_FEATURE_EARLY_CLOSE

soulxu:
> I'm not sure if we want to start sprinkling libevent `#define`s in our codebase. It might be cleaner to just add `#if` for macos, with a comment of why it's not supported on this platform.
> 
> /wait

got it, we have the unittest to check it is support or not when an new platform added (although it is rare case), so it should be fine.

Also note, the `EV_FEATURE_EARLY_CLOSE` is checked by runtime https://libevent.org/doc/event_8h.html#aced890aa77dac1669325f74a27584f03, but yes, it is still sprinkling libevent.

source/common/network/listener_filter_buffer_impl.cc
```diff
@@ -16,9 +16,18 @@ ListenerFilterBufferImpl::ListenerFilterBufferImpl(IoHandle& io_handle,
   // If the buffer_size not greater than 0, it means that doesn't expect any data.
   ASSERT(buffer_size > 0);
 
+// `FileReadyType::Closed` is required `EV_FEATURE_EARLY_CLOSE` feature for libevent, and this feature is
+// only supported with `epoll`. But `MacOS` uses the `kqueue`.
+// https://libevent.org/doc/event_8h.html#a98f643f9c9063a4cbf410f519eb61e55
+#if defined(__APPLE__)
```
soulxu, line 22: @ggreenway I'm a little hesitant to add this check, even if we add `Closed` event for macOS, it won't break anything. maybe without this check will make the code clear? We can just skip the test for macOS. but let me know your opinion.

ggreenway: I think the check can be moved to the test only. Setting the Closed event doesn't cause any harm on macos I don't think; it just doesn't do anything.

ggreenway (changes requested):
/wait

source/common/network/listener_filter_buffer_impl.cc
soulxu: got it

ggreenway (approved):

PR #21593, opened 2022-06-06
http1: allowing for upstream absolutele URIs

alyssawilk:
Risk Level: low (and config guarded)
Testing: new unit, integration tests
Docs Changes: in with APIs
Release Notes: inline

part of https://github.com/envoyproxy/envoy-mobile/issues/1622

### DIFF
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -247,7 +247,7 @@ message HttpProtocolOptions {
   google.protobuf.UInt32Value max_requests_per_connection = 6;
 }
 
-// [#next-free-field: 8]
+// [#next-free-field: 9]
 message Http1ProtocolOptions {
   option (udpa.annotations.versioning).previous_message_type =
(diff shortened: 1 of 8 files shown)


api/envoy/config/core/v3/protocol.proto
```diff
@@ -331,6 +331,10 @@ message Http1ProtocolOptions {
   // If set, this overrides any HCM :ref:`stream_error_on_invalid_http_messaging
   // <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.stream_error_on_invalid_http_message>`.
   google.protobuf.BoolValue override_stream_error_on_invalid_http_message = 7;
+
+  // Allows sending fully qualified URLs when proxying the first line of the
+  // response. By default, Envoy will only send the path components in the first line.
```
htuch, line 336: Can you elaborate on where the host comes from? Presumably this is the request host (subject to rewriting), not the logical DNS entry etc.

mattklein123 (approved):
Nice!

PR #21599, opened 2022-06-07
contrib-sip-proxy: fix only first record-route add ep

durd07:
all sip message in single connection share the same connection manager,
so share the same decoder, the bug is that only the first record-route
will be setted ep, the first-record-route flag should be reset at decode
complete.

Signed-off-by: Felix Du <felix.du@nokia-sbell.com>

Commit Message: sip-proxy: fix only first record-route add ep
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


yanavlasov:
@nearbyfly @dorisd0102 for the first pass review

dorisd0102:
LGTM

yanavlasov (approved):

Issue #20049, opened 2022-02-18
ecds: add support for listener filters 


yanjunxiang-google:
*Title*: ecds: add support for listener filters 

*Description*:
ECDS support for HTTP filters are added by PR: https://github.com/envoyproxy/envoy/pull/11826.
ECDS support for network filters are tracked by issue : https://github.com/envoyproxy/envoy/issues/14696

This issue is to track the ECDS support for listener filters.

Task list:

1) API change is here: https://github.com/envoyproxy/envoy/pull/20050
2) Further abstraction ECDS config discovery header file to make it extendable for other filters like lisenter ECDS filter: https://github.com/envoyproxy/envoy/pull/20445

The reason for this change is that prior to #20445, the two methods in class DynamicFilterConfigProviderImpl is tied into HTTP listener only:
validateMessage()
instantiateFilterFactory().

Thus #20445 is to abstract them out thus the logic can be implemented in filter type specific class. 

3) Create a class ListenerFilterConfigProviderImpl that implements FilterConfigProviderManagerImpl
The PR for this work is: https://github.com/envoyproxy/envoy/pull/20560

4) Adding a testing listener filter for unit and integration test(initially it's using tls_inspector ).  The test listener filter added by https://github.com/envoyproxy/envoy/pull/20571 can be leveraged once it is committed. Or writing something similar. (this work is done within in #20560 )

TCP listener ECDS functionalities support :
5) Add A field with type ListenerFilterConfigProviderImpl in ListenerManagerImpl class for process listener filter.
6) Listener filter processing for the ECDS configuration.
7) Call listener filter factories when building the listener filter chain in createListenerFilterChain()

These are done by: https://github.com/envoyproxy/envoy/pull/21133 Listener filter ECDS Support

8) UDP listener ECDS functionalities support:
This will be just changing the UDP listener counter part code as in above 5) 6) 7).
i)also cleanup the parameter list in createDynamicFilterConfigProvider() (separate HTTP filter and Listener filter paramameters)
ii) add a stats counter in the MissingConfigFilter installation case.

This is the PR:
https://github.com/envoyproxy/envoy/pull/21606

9) Adding release notes for listener ECDS.
https://github.com/envoyproxy/envoy/pull/21646

PR #21606, opened 2022-06-07
UDP listener filter ECDS experiments---draft only

yanjunxiang-google:
UDP listener filter ECDS experiments---draft only

This is continue for issue: https://github.com/envoyproxy/envoy/issues/20049

UDP listener filter chain is created during listener creation. Thus ECDS may not work in UDP listener.

Created this draft PR so folks can discuss on it.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/envoy/filter/config_provider_manager.h
+++ b/envoy/filter/config_provider_manager.h
@@ -22,6 +22,8 @@ using DynamicFilterConfigProviderPtr = std::unique_ptr<DynamicFilterConfigProvid
 // Listener filter config provider aliases
 using ListenerFilterFactoriesList =
     std::vector<FilterConfigProviderPtr<Network::ListenerFilterFactoryCb>>;
+using UdpListenerFilterFactoriesList =
+    std::vector<FilterConfigProviderPtr<Network::UdpListenerFilterFactoryCb>>;
 
 /**
  * The FilterConfigProviderManager exposes the ability to get an FilterConfigProvider
--- a/envoy/server/listener_manager.h
+++ b/envoy/server/listener_manager.h
@@ -103,9 +103,9 @@ public:
    * Creates a list of UDP listener filter factories.
    * @param filters supplies the configuration.
    * @param context supplies the factory creation context.
-   * @return std::vector<Network::UdpListenerFilterFactoryCb> the list of filter factories.
+   * @return Filter::UdpListenerFilterFactoriesList the list of filter factories.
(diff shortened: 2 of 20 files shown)


yanjunxiang-google:
This PR is for folks involved to discuss ECDS on UDP listener.
Some experiments shows the config update to UDP ECDS listener filter is ignored due to UDP listener filter chain is created during listener creation. Thus config update is not taken in actual traffic handling.  This may mean ECDS may not applicable to UDP listener filter.

Any way, throw this PR out so folks can refer this when discussing.

soulxu:
I think the major issue is that the ActiveUDPListener only creates the UDP listener filter factory in the beginning
https://github.com/envoyproxy/envoy/blob/607b2987565aba652c2f2fdaaa1f6085c0aa59ef/source/server/active_udp_listener.cc#L94-L102

And never recreate those factories again, which means those factories will keep using the old config.

This is different from TCP, ActiveTcpListener will create a new listener filter factory for each new connection, which means when the ECDS updates the config, then the new connection can use the new config by creating the new factory.

https://github.com/envoyproxy/envoy/blob/607b2987565aba652c2f2fdaaa1f6085c0aa59ef/source/server/active_stream_listener_base.h#L85-L87

We probably need some ways to trigger a recreation of the UDP listener filter when the config is updated.

yanjunxiang-google:
> > This PR is for folks involved to discuss ECDS on UDP listener. Some experiments shows the config update to UDP ECDS listener filter is ignored due to UDP listener filter chain is created during listener creation. Thus config update is not taken in actual traffic handling. This may mean ECDS may not applicable to UDP listener filter.
> > Any way, throw this PR out so folks can refer this when discussing.
> 
> How about to recreate the udp listener filter when the ECDS updates the config?

Currently UDP listener filter chain is created during Envoy instance creation:
https://github.com/envoyproxy/envoy/blob/64a45c4c8261b94a7ba57b87ed631643ae8fac51/source/server/active_udp_listener.cc#L90

If we recreated the UDP listener filter chian whenever ECDS updates received, which could be frequently,  could this cause data loss? For example, when the UDP listener filters are processing live traffic, and ECDS updates received, what should Envoy do with that traffic?

PR #21609, opened 2022-06-07
Set clang-format to sort includes

ravenblackx:
Commit Message: Set clang-format to sort includes
Additional Description: format-checking presubmit requires includes to be sorted; vscode does most formatting on save, but because of the .clang-format file does not perform header sorting. This change makes headers get sorted by default, making the autoformat behavior match the CI requirement.
Risk Level: n/a
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


ravenblackx:
Ugh, that's got horribly mangled. This is probably not something for me to try to do anyway. :)

ravenblackx:
Yeah, the ci/* scripts auto-sort headers but take a really long time to run (and have to be run manually) compared to the clangd built-in format-on-save method. But maybe it makes sense to just put up with that, versus trying to fight the system on this. :)

jmarantz:
I bet the sort order script would run very fast if you ran it on is own, on
only the files in your PR, fwiw.

PR #21614, opened 2022-06-07
Router filter pattern rewrite

silverstar194:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/common/matchers.h
```diff
@@ -191,18 +220,25 @@ class MetadataMatcher {
...
+  static PathMatcherConstSharedPtr createPattern(const std::string& pattern, bool ignore_case);
   static PathMatcherConstSharedPtr
   createSafeRegex(const envoy::type::matcher::v3::RegexMatcher& regex_matcher);
 
   bool match(const absl::string_view path) const override;
 
 private:
   const StringMatcherImpl<envoy::type::matcher::v3::StringMatcher> matcher_;
+  const PatternMatcherImpl<envoy::type::matcher::v3::PatternMatcher> pattern_matcher_;
+  const bool pattern_based_match_;
```
adisuissa, line 241: nit: should probably be an enum instead of a bool.

Another alternative is to separate the static creation methods from the classes that implement the `match()` function (similar to what is done in the Factory patterns).

adisuissa (commented):
Thanks for working on this. I've left a few high level comments.

PR #21632, opened 2022-06-08
Fix `--log-format-escaped` to only log one newline per log line

nareddyt:
Commit Message: Fix `--log-format-escaped` to only log one newline per log line

Additional Description:

Previously, when a log line contained multiple trailing newline characters, it would result in creating multiple empty lines in the log output. 

For example, `ENVOY_LOG(debug, "message\n message \n\n")` is translated to log line `message\n message \n\n\n`. With `--log-format-escaped` enabled, this is logged as:

```
message\n message



```

Because **all trailing whitespace characters** would not be escaped under the previous implementation. The correct behavior is for all `\n` to be escaped **except** the last one:

```
message\n message\n\n

```

Risk Level: Low

Testing: Unit tests, including platform specific ones

Docs Changes: Updated wording in flag documentation

Release Notes: `--log-format-escaped` now only logs one newline per log line

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -74,6 +74,9 @@ minor_behavior_changes:
     This means that encoder filters will be correctly invoked, including adding configured response
     headers, etc. This behavioral change can be reverted by setting runtime guard
     ``envoy.reloadable_features.lua_respond_with_send_local_reply`` to false.
+- area: logging
+  change: |
+    changed flag ``--log-format-escaped`` to only log one trailing newline per log line.
 
 bug_fixes:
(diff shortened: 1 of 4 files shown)


nareddyt:
cc @qiwzhang @TAOXUY

nareddyt:
PTAL @snowp

nareddyt:
> lgtm, can you add a note to the release notes at https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml ?

Done

snowp (approved):
Thanks!

PR #11187, opened 2020-05-14
regex: don't enforce max_program_size when field isn't present in config.

LisaLudique:
Commit Message: regex: don't enforce max_program_size when field isn't present in config.
Additional Description: max_program_size was deprecated, but when the field isn't specified, a default max program size of 100 is still enforced. This change makes it such that when the field isn't present, no max program size check is enforced.
Risk Level: Low
Testing: Unit testing
Docs Changes: Included
Release Notes: Included
Fixes #11185 
Fixes https://github.com/envoyproxy/envoy/issues/9174

PR #21633, opened 2022-06-09
regex: add regex engine factory and config option

### LINKED ISSUE #21618: proposal: add regex engine factory and config option
*Title*: Add regex engine factory and config option

*Description*:

Nowadays, we use RE2 as the only regex engine in Envoy. In scenario with a large number of regex matcher like a gateway, users may face high CPU usage and performance bottleneck.

We can make regex engine extensible, to introduce a bootstrap config option to configure regex engine and its option like `SocketInterface`. We can benefit both a simpler regex matcher configuration since engine-specific options will be removed (see API change below) and avoiding the runtime value like `re2.max_program_size.error_level` since they can be configured in `bootstrap_extensions`.

*API change*:

```yaml
safe_regex:
  # All the fields in google_re2 have been deprecated, but google_re2 is still required by safe_regex.
  # We will deprecate it this time.
  # google_re2: {}
  regex: "/safe.*"
```

```yaml
# If the default_regex_engine is unset, RE2 will be used by default.
default_regex_engine: envoy.regex_engine.google_re2
bootstrap_extensions:
- name: envoy.regex_engine.google_re2
  typed_config:
    '@type': type.googleapis.com/envoy.type.matcher.v3.RegexMatcher.GoogleRE2
    # Options like max_program_size can be configured here.
```

zhxie:

Commit Message: regex: add regex engine factory and config option
Additional Description:

The patch adds support for injecting regex engine with its options on startup in a bootstrap option. These are some API, runtime and implementation changes.

1. `google_re2` in `safe_regex` will be deprecated and no longer be required. `Regex::parseRegex` will choose the registered regex engine to parse expressions into matchers.
2. A new bootstrap option `default_regex_engine` will be introduced for regex engine selection.
3. Integer runtime values will no longer take effect. `re2_max_program_size_error_level` and `re2_max_program_size_warn_level` have been deprecated 3 months ago, and the patch will add them into a new `GoogleRE2` proto. The `GoogleRE2` message will be used for initialize regex engine in startup.
4. For compatibility, `GoogleRE2` will be chosen as default regex engine if no regex engine is designated. Also, `max_program_size` and `warn_program_size` in `GoogleRE2` have the same value with runtime values described.
5. For compatibility, if any regex matcher message includes the deprecated `google_re2` fields, `Regex::parseRegex` will take the fallback path, to generate regex matcher in the old form, using the deprecated `google_re2` field, deprecated runtime values, and forced Google RE2 engine.

Risk Level: Medium
Testing: Unit and integration
Docs Changes: N/A
Release Notes: WIP
Fixes #21618


api/envoy/extensions/regex_engine/v3/google_re2.proto
```diff
@@ -0,0 +1,34 @@
...
+//
+// Envoy emits two stats for tracking the program size of regexes: the histogram `re2.program_size`,
+// which records the program size, and the counter `re2.exceeded_warn_level`, which is incremented
+// each time the program size exceeds the warn level threshold.
+message GoogleRE2 {
+  // This field controls the RE2 "program size" which is a rough estimate of how complex a
+  // compiled regex is to evaluate. A regex that has a program size greater than the configured
+  // value will fail to compile. In this case, the configured max program size can be increased
+  // or the regex can be simplified. If not specified, the default is 100.
+  google.protobuf.UInt32Value max_program_size = 1;
```
lizan, line 29: We already deprecated `max_program_size` in GoogleRE2 and states that should be performed on the management server. So no need to undeprecate these two field?

api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -330,6 +330,10 @@ message Bootstrap {
   // Optional path to a file with performance tracing data created by "Perfetto" SDK in binary
   // ProtoBuf format. The default value is "envoy.pftrace".
   string perf_tracing_file_path = 33;
+
+  // Optional overriding of default regex engine. The value must be the name of one of the
+  // regex engine factories initialized through a bootstrap extension.
+  string default_regex_engine = 34;
```
lizan, line 336: Is there a way to override regex engine other than default in each regex matcher? We don't use name to identify extensions anymore, so shouldn't this just be an `TypedExtensionConfig`?

lizan (commented):
start with API comments.

api/envoy/extensions/regex_engine/v3/google_re2.proto
zhxie: In fact, I add an new `GoogleRE2` message (api/envoy/extensions/regex_engine/v3/google_re2.proto), since I think it should be located outside `RegexMatcher`. And it is also a compatibility concern of xDS API, since we must change both Envoy's API and xDS's API to modify the `GoogleRE2` in `RegexMatcher`.

api/envoy/config/bootstrap/v3/bootstrap.proto
zhxie: > Is there a way to override regex engine other than default in each regex matcher?

The bootstrap option provides the way configuring regex engine in one single place other than write them in each `RegexMatcher`.

> We don't use name to identify extensions anymore, so shouldn't this just be an TypedExtensionConfig?

Ok, I may take some time to change it from `string` to `TypedExtensionConfig`.

zhxie: Done.

api/envoy/extensions/regex_engine/v3/google_re2.proto
lizan: I'm not asking you to modify `GoogleRE2` in `RegexMatcher`, since those fields are already deprecated, so you don't need them in this new `GoogleRE2` message.

api/envoy/config/bootstrap/v3/bootstrap.proto
lizan: > The bootstrap option provides the way configuring regex engine in one single place other than write them in each `RegexMatcher`.

My question is, how to override in each `RegexMatcher`? It doesn't seems to have such a way in this PR. Just not over complicate this that's fine though.

api/envoy/extensions/regex_engine/v3/google_re2.proto
zhxie: In fact, we deprecate these fields but we still bring them back in #11187 as runtime values. Now we have decided to deprecated these runtime values in #20168, I think it will be proper to set them in a bootstrap option. https://github.com/envoyproxy/envoy/blob/c9c46bb41130eff0c84dcde0d609dd5c145ac34a/source/common/runtime/runtime_features.cc#L97-L100

api/envoy/config/bootstrap/v3/bootstrap.proto
zhxie: Yes, the patch does not provide the way to override in each `RegexMatcher`. The original `GoogleRE2` message has deprecated all its fields, so I think we do not need overriding methods currently.

api/envoy/extensions/regex_engine/v3/google_re2.proto
lizan: We are not intended to bring them back, it's there and deprecated just for compatibility. The document here clearly states that's not Envoy's responsibility but control-planes'. So we should just add them in a new message.

cc @alyssawilk

zhxie: Done.

source/common/common/regex.cc
```diff
@@ -56,6 +59,28 @@ CompiledGoogleReMatcher::CompiledGoogleReMatcher(
...
+CompiledMatcherPtr GoogleReEngine::matcher(const std::string& regex) const {
+  return std::make_unique<CompiledGoogleReMatcher>(regex, true);
+}
+
+EnginePtr
+GoogleReEngineFactory::createEngine(const Protobuf::Message& config,
+                                    ProtobufMessage::ValidationVisitor& validation_visitor) {
+  const auto google_re2 =
+      MessageUtil::downcastAndValidate<const envoy::extensions::regex_engines::v3::GoogleRE2&>(
+          config, validation_visitor);
```
lizan, line 71: not used?

source/server/server.cc
```diff
@@ -592,6 +592,20 @@ void InstanceImpl::initialize(Network::Address::InstanceConstSharedPtr local_add
     }
   }
 
+  // Initialize the regex engine and inject to singleton.
+  const auto& default_regex_engine = bootstrap_.default_regex_engine();
+  if (default_regex_engine.has_typed_config()) {
```
lizan, line 597: bootstrap_.has_default_regex_engine?

where is default engine initialized?

source/common/common/regex.cc
zhxie: Removed, thanks.

source/server/server.cc
zhxie: > bootstrap_.has_default_regex_engine?

Changed, thanks.

> where is default engine initialized?

It is initialized in a singleton.

https://github.com/envoyproxy/envoy/blob/54e6ecf106e2ee1d8e9078f4299543ca7ba38a2a/source/common/common/regex.cc#L77

lizan: OK move that initialization up here, so we don't always initialize a GoogleReEngine in global scope?

zhxie: No, since `engine_` is a static variable, a `GoogleReEngine` will be initialized at start up. If user designates an other regex engine or regex engine options, we will re-initialize a new one in bootstrap.

zhxie: Usually, `Regex::Utility::parseRegex` takes the responsibility to parse the regex, and it is a static method. If we do not initialize it with a static variable, we may potentially confront a segfault.
If we do not want a static variable, then I can add a fallback path in `parseRegex` to handle situation without regex engine specific.

lizan: I'm just asking an `else` block here, you don't need to change everything else and it should still work?

zhxie: Done, with about 10 tests affected and updated.

lizan (approved):

api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -330,6 +330,10 @@ message Bootstrap {
   // Optional path to a file with performance tracing data created by "Perfetto" SDK in binary
   // ProtoBuf format. The default value is "envoy.pftrace".
   string perf_tracing_file_path = 33;
+
+  // Optional overriding of default regex engine.
+  // If the value is not specified, Google RE2 will be used by default.
+  core.v3.TypedExtensionConfig default_regex_engine = 34;
```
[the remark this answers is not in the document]
zhxie: Yes, I have added a category `envoy.regex_engines` and the default Google RE2 is registered as `envoy.regex_engines.google_re2`. I am sorry but could you please explain more about "extension category docs"? Maybe I can add one in the following up PR #21846.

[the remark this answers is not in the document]
kyessenov: @zhxie, we'd need to document the extension with `[#extension: ...]` for Google RE2, add it to manifest, and that will be automatically linked against this field if you put `[#extension-category: envoy.regex_enginers]` here.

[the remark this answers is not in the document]
mattklein123: Yeah that is what I was referring to. Can we do this as a follow up please?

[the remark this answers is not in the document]
zhxie: Of course, I will try to fix that on #21846.

PR #21638, opened 2022-06-09
fuzz test: Improve on no match validation

vehre-x41:
Commit Message: fuzz test: Improve on_no_match validation. 
Additional Description:
When fuzzing a routeaction with on_no_match, reject a configuration, if some fields are unset.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/common/router/route_fuzz_test.cc
+++ b/test/common/router/route_fuzz_test.cc
@@ -18,14 +18,8 @@ namespace {
 // Limit the size of the input.
 static const size_t MaxInputSize = 64 * 1024;
 
-bool isUnsupportedRouteConfig(const envoy::config::route::v3::Route& route,
-                              bool extended_check = false) {
+bool isUnsupportedRouteConfig(const envoy::config::route::v3::Route& route) {
   if (route.has_filter_action() || route.has_non_forwarding_action()) {
-    return true;
-  }
-  if (extended_check && route.has_route() &&
-      !(route.route().has_cluster_specifier_plugin() || route.route().has_cluster() ||
-        route.route().has_cluster_header() || route.route().has_weighted_clusters())) {
     return true;
   }
   return false;
@@ -57,15 +51,17 @@ bool validateOnMatchConfig(const xds::type::matcher::v3::Matcher::OnMatch& on_ma
   if (on_match.has_matcher() && !validateMatcherConfig(on_match.matcher())) {
     return false;
   }
-  if (on_match.action().typed_config().type_url().find("envoy.config.route.v3.Route") !=
+  // Only envoy...Route is allowed as typed_config here.
+  if (on_match.action().typed_config().type_url().find("envoy.config.route.v3.Route") ==
       std::string::npos) {
-    envoy::config::route::v3::Route on_match_route_action_config;
-    MessageUtil::unpackTo(on_match.action().typed_config(), on_match_route_action_config);
-    ENVOY_LOG_MISC(trace, "typed_config of on_match.action is: {}",
-                   on_match_route_action_config.DebugString());
-    if (isUnsupportedRouteConfig(on_match_route_action_config, true)) {
-      return false;
-    }
+    return false;
+  }
+  envoy::config::route::v3::Route on_match_route_action_config;
+  MessageUtil::unpackTo(on_match.action().typed_config(), on_match_route_action_config);
+  ENVOY_LOG_MISC(trace, "typed_config of on_match.action is: {}",
+                 on_match_route_action_config.DebugString());
+  if (isUnsupportedRouteConfig(on_match_route_action_config)) {
+    return false;
   }
   return true;
 }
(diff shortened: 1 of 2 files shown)


vehre-x41:
CC @adisuissa @KBaichoo  @pradeepcrao

test/common/router/route_fuzz_test.cc
```diff
@@ -16,6 +18,19 @@ namespace {
...
+bool isUnsupportedRouteConfig(const envoy::config::route::v3::Route& route,
+                              bool extended_check = false) {
+  if (route.has_filter_action() || route.has_non_forwarding_action()) {
+    return true;
+  }
+  if (extended_check && route.has_route() &&
+      !(route.route().has_cluster_specifier_plugin() || route.route().has_cluster() ||
+        route.route().has_cluster_header() || route.route().has_weighted_clusters())) {
+    return true;
+  }
```
adisuissa, line 30: Is this part needed to exclude [`inline_cluster_specifier_plugin `](https://github.com/envoyproxy/envoy/blob/469f7bbd5951a29dfc24a3a05ca01459b835da2c/api/envoy/config/route/v3/route_components.proto#L967)?

IIUC the `cluster_specifier` is a `oneof` that is required, so one of the fields must appear there, no?

test/common/router/route_fuzz_test.cc
```diff
@@ -50,6 +64,19 @@ bool validateConfig(const test::common::router::RouteTestCase& input) {
       ENVOY_LOG_MISC(debug, "retry_policy_typed_config: not implemented");
       return false;
     }
+    if (virtual_host.has_matcher() && virtual_host.matcher().on_no_match().has_action() &&
+        virtual_host.matcher().on_no_match().action().name() ==
```
adisuissa, line 68: Moving a discussion we had in another repo here:

This validates the `on_no_match` field of the matcher. However, there could be other fields (e.g., [FieldMatcher](https://github.com/envoyproxy/envoy/blob/f8f8628f4ec7cb2dcec92f094ffd4715dcc31dab/api/envoy/config/common/matcher/v3/matcher.proto#L99)) that will have the same issue.

I don't think there is a generic way to traverse this recursive data-structure, as it is very specific to how `Matcher` is defined. One can probably use protobuf reflection for this, but it may be an overkill.
I see 2 options here:
1. Handle the cases one-by-one as the fuzzer finds them.
2. Create a recursive function that traverses the Matcher object, and whenever it reaches an [`OnMatch` action](https://github.com/envoyproxy/envoy/blob/f8f8628f4ec7cb2dcec92f094ffd4715dcc31dab/api/envoy/config/common/matcher/v3/matcher.proto#L43) validates that it is not `filter_action` or `non_forwarding_action`.

adisuissa (commented):
Thanks for fixing this!

test/common/router/route_fuzz_test.cc
vehre-x41: The `inline_cluster_specifier_plugin` is omitted from the check, because in [`clusterEntry()`](https://github.com/envoyproxy/envoy/blob/911a87020b00e77e7a43fd6924b9579c741261c2/source/common/router/config_impl.cc#L1172-L1189) it is not referenced. Its presence will therefore not suffice to prevent the assert.

vehre-x41: During research to answer this question, I found another issue, that is not completely covered by the current state. Commit https://github.com/envoyproxy/envoy/pull/21638/commits/dff70d23b4508ab3b2870ae88c987a2cdef1a0df now also validates route action config where the type_url is not exactly given as expected, but only contains `"envoy.config.route.v3.Route"`.

adisuissa, line 30: I think these 4 lines are not needed.

test/common/router/route_fuzz_test.cc
```diff
@@ -37,7 +51,68 @@ cleanRouteConfig(envoy::config::route::v3::RouteConfiguration route_config) {
...
+bool validateMatcherConfig(const xds::type::matcher::v3::Matcher& matcher);
+
+bool validateOnMatchConfig(const xds::type::matcher::v3::Matcher::OnMatch& on_match) {
+  if (on_match.has_matcher() && !validateMatcherConfig(on_match.matcher())) {
+    return false;
+  }
+  if (on_match.action().typed_config().type_url().find("envoy.config.route.v3.Route") !=
+      std::string::npos) {
+    envoy::config::route::v3::Route on_match_route_action_config;
+    MessageUtil::unpackTo(on_match.action().typed_config(), on_match_route_action_config);
```
adisuissa, line 63: This call may throw an EnvoyException. The call to `validateConfig(input)` in line 137 should move into the try-catch scope.

test/common/router/route_fuzz_test.cc
```diff
@@ -37,7 +51,68 @@ cleanRouteConfig(envoy::config::route::v3::RouteConfiguration route_config) {
...
 }
 
-// Check configuration for size and unimplemented options.
+bool validateMatcherConfig(const xds::type::matcher::v3::Matcher& matcher);
+
+bool validateOnMatchConfig(const xds::type::matcher::v3::Matcher::OnMatch& on_match) {
+  if (on_match.has_matcher() && !validateMatcherConfig(on_match.matcher())) {
+    return false;
+  }
+  if (on_match.action().typed_config().type_url().find("envoy.config.route.v3.Route") !=
```
adisuissa, line 60: Just to make sure I understand this correctly, if the type isn't "envoy....Route" will the config be rejected later (either in the call to `TestUtility::validate(input)` or when creating the `ConfigImpl` object)?
If not, then the config isn't valid for this fuzz test, and I think this validation should return false.

adisuissa (commented):
Looks very good, thanks!
Left a couple of small comments.

test/common/router/route_fuzz_test.cc
vehre-x41: With the corpi present, these lines are indeed not needed. Removed.

test/common/router/route_fuzz_test.cc
vehre-x41: done

adisuissa (approved):
LGTM, thanks!

PR #21641, opened 2022-06-09
config-plane: Refactoring OpaqueResourceDecoder to be a shared pointer

adisuissa:
Commit Message: Refactoring OpaqueResourceDecoder to be a shared pointer
Additional Description:
Prior to this PR the OpaqueResourceDecoder is owned by each of the Subscriptions, and a reference to it is used by the GrpcMux's. This can cause an issue if the Mux outlives the subscription, as can happen in the newly added test "ValidResourceDecoderAfterRemoval", and in the unified Mux implementation.
This PR converts the OpaqueResourceDecoder to a shared pointer, that can be used by both.

Risk Level: Low/Medium - modifying many places in the config-plane codebase, but the refactor moves from reference to shared pointer.
Testing: Added a unit test that passes after the refactor.
Docs Changes: N/A.
Release Notes: N/A.

### DIFF
--- a/envoy/config/grpc_mux.h
+++ b/envoy/config/grpc_mux.h
@@ -100,7 +100,7 @@ public:
   virtual GrpcMuxWatchPtr addWatch(const std::string& type_url,
                                    const absl::flat_hash_set<std::string>& resources,
                                    SubscriptionCallbacks& callbacks,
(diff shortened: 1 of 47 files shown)


adisuissa:
Looking at changing this to unique pointer instead of shared.
/wait

PR #21648, opened 2022-06-09
transport sockets: split downstream and upstream transport socket factory

kyessenov:

Description: Addresses the unfortunate pattern of using `createTransportSocket(nullptr)` to indicate a downstream transport socket, by making the two interfaces explicit. In theory, this should permit adding more context to the calls (such as host for upstream).
Description: 
Risk Level: low, refactor
Testing: WIP
Docs Changes: none
Release Notes: none

/assign @lizan


envoy/network/transport_socket.h
```diff
@@ -269,6 +262,21 @@ class TransportSocketFactory {
    * not client-side TLS sockets.
    */
   virtual absl::string_view defaultServerNameIndication() const PURE;
+};
+
+/**
+ * A factory for creating upstream transport sockets. It will be associated to clusters.
+ */
+class TransportSocketFactory : public virtual TransportSocketFactoryBase {
```
lizan, line 270: name this UpstreamTransportSocketFactory?

source/extensions/transport_sockets/alts/tsi_socket.h
```diff
@@ -130,7 +130,8 @@ class TsiSocket : public Network::TransportSocket,
 /**
  * An implementation of Network::TransportSocketFactory for TsiSocket
  */
-class TsiSocketFactory : public Network::CommonTransportSocketFactory {
+class TsiSocketFactory : public Network::DownstreamTransportSocketFactory,
+                         public Network::CommonTransportSocketFactory {
```
lizan, line 134: CommonTransportSocketFactory is actually upstream as it only handles transport socket options.

kyessenov: Yeap, renaming.

envoy/network/transport_socket.h
kyessenov: Ack.

source/common/network/transport_socket_options_impl.h
```diff
@@ -92,7 +92,7 @@ class TransportSocketOptionsUtility {
   fromFilterState(const StreamInfo::FilterStateSharedPtr& stream_info);
 };
 
-class CommonTransportSocketFactory : public TransportSocketFactory {
+class CommonUpstreamTransportSocketFactory : public UpstreamTransportSocketFactory {
```
lizan, line 95: Just merge this interface with UpstreamTransportSocketFactory? It existed because we didn't have upstream/downstream transport socket factory split.

kyessenov: This one has an implementation of `hashKey` so it will make `UpstreamTransportSocketFactory` not a pure interface. Is that OK with core interface carrying hashing implementation?

lizan: Oh ok, nvm.

source/common/quic/quic_transport_socket_factory.h
```diff
@@ -46,7 +46,7 @@ class QuicTransportSocketFactoryBase : public Network::CommonTransportSocketFact
   // To be called right after construction.
   virtual void initialize() PURE;
 
-  // Network::TransportSocketFactory
+  // Network::UpstreamTransportSocketFactory
```
lizan, line 49: move this to QuicClientTransportSocketFactory?

kyessenov: Done.

lizan (approved):

source/server/active_stream_listener_base.cc
```diff
@@ -40,7 +40,7 @@ void ActiveStreamListenerBase::newConnection(Network::ConnectionSocketPtr&& sock
     return;
   }
   stream_info->setFilterChainName(filter_chain->name());
-  auto transport_socket = filter_chain->transportSocketFactory().createTransportSocket(nullptr);
+  auto transport_socket = filter_chain->transportSocketFactory().createDownstreamTransportSocket();
```
danzh2010, line 43: The intention of this PR is to split createTransportSocket() into client method and server method. But is it intentional that the transport socket option param is removed from the server transport socket creation interface? We don't have config knob for server transport socket options right now, but if we want to do mutual TLS and server transport socket will need to verify cert, would server transport socket also take some transport socket options?

kyessenov: Socket options mostly serve as a way to create multiple upstream transport sockets for distinct groups of downstream requests, one per each variation. It's not about configuring the transport socket itself, that's done through the factory.

danzh2010: TransportSocketOptions::verifySubjectAltNameListOverride() is used in cert verification: https://github.com/envoyproxy/envoy/blob/469f7bbd5951a29dfc24a3a05ca01459b835da2c/source/extensions/transport_sockets/tls/cert_validator/default_validator.cc#L215

Could this be used by the server to match SAN values in a client cert? Or allowing per-connection override is non-sense? @RyanTheOptimist

kyessenov: I think the listener filter chain match is used to match SAN values. I can imagine a more flexible transport socket that can adapt during negotiation, but Envoy doesn't have that yet.

Issue #21590, opened 2022-06-06
add compatibility path-parameter 


stevenzzzz:
*Title*: *One line description*

*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

stevenzzzz:
@yanavlasov @alyssawilk

yanavlasov:
@mattklein123 @alyssawilk @snowp for more comments

mattklein123:
Sure sounds reasonable to me.

PR #21664, opened 2022-06-10
[router] Ignore path parameters when RouteConfiguration is configured

stevenzzzz:
Commit Message: Add an option in RouteConfiguration, if enabled, ignore path-parameters (RFC1808) in path matching.

Additional Description:
Risk Level: LOW(default false).
Testing: unit test
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #21590
[Optional Deprecated:]

### DIFF
--- a/api/envoy/config/route/v3/route.proto
+++ b/api/envoy/config/route/v3/route.proto
@@ -22,7 +22,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // * Routing :ref:`architecture overview <arch_overview_http_routing>`
 // * HTTP :ref:`router filter <config_http_filters_router>`
 
-// [#next-free-field: 15]
+// [#next-free-field: 16]
 message RouteConfiguration {
   option (udpa.annotations.versioning).previous_message_type = "envoy.api.v2.RouteConfiguration";
 
@@ -134,6 +134,12 @@ message RouteConfiguration {
   // NOTE: this option will not strip the port number (if any) contained in route config
   // :ref:`envoy_v3_api_msg_config.route.v3.VirtualHost`.domains field.
   bool ignore_port_in_host_matching = 14;
+
+  // Ignore path-parameters in path-matching.
+  // Before RFC3986, URI were like(RFC1808): <scheme>://<net_loc>/<path>;<params>?<query>#<fragment>
+  // Envoy by default takes ":path" as "<path>;<params>".
+  // For users who want to only match path on the "<path>" portion, this option should be true.
+  bool ignore_path_parameters_in_path_matching = 15;
 }
 
 message Vhds {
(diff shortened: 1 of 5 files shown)


api/envoy/config/route/v3/route.proto
```diff
@@ -134,6 +134,12 @@ message RouteConfiguration {
   // NOTE: this option will not strip the port number (if any) contained in route config
   // :ref:`envoy_v3_api_msg_config.route.v3.VirtualHost`.domains field.
   bool ignore_port_in_host_matching = 14;
+
+  // Ignore path-parameters in path-matching.
```
soulxu, line 138: Should we add new field to RouteMatch to enable the user match the path parameters? But that should be separated PR and requirement.

source/common/router/config_impl.cc
```diff
@@ -1442,8 +1454,7 @@ RouteConstSharedPtr RegexRouteEntryImpl::matches(const Http::RequestHeaderMap& h
                                                  const StreamInfo::StreamInfo& stream_info,
                                                  uint64_t random_value) const {
   if (RouteEntryImplBase::matchRoute(headers, stream_info, random_value)) {
-    const absl::string_view path = Http::PathUtil::removeQueryAndFragment(headers.getPathValue());
-    if (path_matcher_->match(path)) {
+    if (path_matcher_->match(sanitizePathBeforePathMatching(headers.getPathValue()))) {
```
soulxu, line 1457: This won't work with the case without path parameter and with query parameter.

source/common/router/config_impl.cc
```diff
@@ -1505,9 +1516,9 @@ PathSeparatedPrefixRouteEntryImpl::matches(const Http::RequestHeaderMap& headers
   if (!RouteEntryImplBase::matchRoute(headers, stream_info, random_value)) {
     return nullptr;
   }
-  absl::string_view path = Http::PathUtil::removeQueryAndFragment(headers.getPathValue());
-  if (path.size() >= prefix_.size() && path_matcher_->match(path) &&
-      (path.size() == prefix_.size() || path[prefix_.size()] == '/')) {
+  absl::string_view sanitized_path = sanitizePathBeforePathMatching(headers.getPathValue());
```
soulxu, line 1519: ditto, this probably why the test failed.

stevenzzzz: readd back removeQueryAndFragment.

stevenzzzz (commented):
thanks!

yanavlasov (changes requested):
/wait

api/envoy/config/route/v3/route.proto
soulxu: Sorry, probably I didn't describe it clearly. I mean a new field for people matching the `path parameters`, since we have a field to match the path, and a field to match the query parameter. There is no field for `path parameters`, you only can match that with path currently, but I think that won't work with multiple path parameters with different orders.

stevenzzzz: ahh, I agree that it should be in another PR. 
I am not sure path-parameters is a widely used legacy feature, (same as this PR, not many users are using it), but would be nice if Envoy can support it.

yanavlasov (approved):

PR #21669, opened 2022-06-11
tooling: Use pre-compiled `protoc`

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
This gives a pretty big speedup for anything that requires `protoc`

In my local testing the `protoc` cache seems to be blown very easily and currently the docs build compiles this twice (i havent been able to track down why) - this resolves that and makes building docs/api a lot less painful

bazel/protobuf.patch
```diff
@@ -11,6 +11,113 @@ index 1690d4219..8a7f1bf14 100644
...
+-    name = "protoc",
++    name = "compiled_protoc",
+     srcs = ["src/google/protobuf/compiler/main.cc"],
+     linkopts = LINK_OPTS,
+     visibility = ["//visibility:public"],
+     deps = [":protoc_lib"],
+ )
+ 
++
++# Lifted from `rules_proto`
```
lizan, line 29: Any reason not use rules_proto directly?

phlax: we dont install their deps, and in the many attempts at trying one way or another it didnt work

also, our current rules_proto's protobuf version is incompatible with ours, so this seemed the easiest way all round

bazel/repository_locations.bzl
```diff
@@ -1,4 +1,16 @@
 # This should match the schema defined in external_deps.bzl.
+
+PROTOC_VERSIONS = dict(
```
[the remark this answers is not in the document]
phlax: yep

bazel/repository_locations.bzl
```diff
@@ -1174,3 +1186,19 @@ REPOSITORY_LOCATIONS_SPEC = dict(
         cpe = "N/A",
     ),
 )
+
+def _compiled_protoc_deps(locations, versions):
+    for platform, sha in versions.items():
+        locations["com_google_protobuf_protoc_%s" % platform] = dict(
+            project_name = "Protocol Buffers (protoc) %s" % platform,
+            project_desc = "Language-neutral, platform-neutral extensible mechanism for serializing structured data",
```
phlax, line 1194: add something distinguishing

bazel/repositories.bzl
```diff
@@ -744,6 +744,12 @@ def _com_google_protobuf():
         name = "rules_python",
     )
 
+    for platform in PROTOC_VERSIONS:
+        external_http_archive(
+            "com_google_protobuf_protoc_%s" % platform,
+            build_file = "@rules_proto//proto/private:BUILD.protoc",
```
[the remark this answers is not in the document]
phlax: or perhaps copy this file out so it matches other versions

bazel/repository_locations.bzl
phlax, line 3: @moderation im guessing its ok for `repository_locations.bzl` to share `locals()` namepsace with `REPOSITORY_LOCATIONS_SPEC` 

it was kinda necessary to do it this way - rather than post-processing for a few reasons, but essentially becaue it does introduce these new "deps" albeit from an existing release

we might also want to consider adding a `toolchain_downloads` or similar to the repo locations spec to automate this for other deps

phlax:
now we have check/stats breakdowns in azp analytics (https://dev.azure.com/cncf/envoy/_pipeline/analytics/duration?definitionId=11&contextType=build) its slightly easier to see current issues

in postsubmit nothing is cached - coverage currently takes > 1 hour to run

running this PR to get an uncached run:

- https://github.com/envoyproxy/envoy/pull/21695
- https://dev.azure.com/cncf/envoy/_build/results?buildId=110547&view=logs&s=4f7d954b-a765-565f-91a2-c04870dab43f&j=bbe4b42d-86e6-5e9c-8a0b-fea01d818a24

this seems to reliably reduce coverage down to < 20 mins

i thought it also resolved compile_time_options as this doesnt seem to bad in some runs - but it can still take long

the reason i believe that compiling protoc is having impact here is because i believe the cache is blown for invokation with different build options

either way this PR has a pretty massive impact across the board, and from a local-without-rbe perspective makes a load of tasks a lot more doable

phlax:
> does seem like it would be nice to try and move to rules_proto eventually

to clarify on this - apart from the protobuf version mismatch they dont currently expose toolchains - you have to use their deps or you dont get the precompiled versions

see https://github.com/bazelbuild/rules_proto/blob/dcf9e47b0df2218ca33e02a1a51803ab3134f42d/proto/repositories.bzl#L34-L38

bazel/repository_locations.bzl
phlax: re this - this also raises a load of extra dependency issues in the dep checker - there are a bunch of ways to fix/workaround but im inclined not to as it provides (~correct) checking on the deps added here

yanavlasov (approved):

PR #21678, opened 2022-06-13
[socket] PassthroughSocket: Add implementation for `connect`.

conqerAtapple:
Socket extensions like `proxy_protocol` socket extends PassthroughSocket which currently does not
delegate the `connect` call to the wrapped socket. This renders the wrapped sockets unable to
intercept `connect`.

This change fixes this problem by adding an override for `connect` method in PassthroughSocket.

Signed-off-by: Jojy George Varghese <jojy_varghese@apple.com>

Risk Level: Low
Testing: Unit tests, Manual
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/extensions/transport_sockets/common/passthrough.cc
+++ b/source/extensions/transport_sockets/common/passthrough.cc
@@ -24,6 +24,10 @@ absl::string_view PassthroughSocket::failureReason() const {
 
 bool PassthroughSocket::canFlushClose() { return transport_socket_->canFlushClose(); }
 
+Api::SysCallIntResult PassthroughSocket::connect(Network::ConnectionSocket& socket) {
+  return transport_socket_->connect(socket);
+}
+
 void PassthroughSocket::closeSocket(Network::ConnectionEvent event) {
   transport_socket_->closeSocket(event);
 }
--- a/source/extensions/transport_sockets/common/passthrough.h
+++ b/source/extensions/transport_sockets/common/passthrough.h
@@ -38,6 +38,7 @@ public:
   std::string protocol() const override;
   absl::string_view failureReason() const override;
   bool canFlushClose() override;
+  Api::SysCallIntResult connect(Network::ConnectionSocket& socket) override;
   void closeSocket(Network::ConnectionEvent event) override;
   Network::IoResult doRead(Buffer::Instance& buffer) override;
   Network::IoResult doWrite(Buffer::Instance& buffer, bool end_stream) override;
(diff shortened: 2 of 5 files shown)


RyanTheOptimist (approved):
This LGTM, but I wonder if it should have a runtime flag to protect the behavior change? it's pretty minor, so maybe not. @alyssawilk, what do you think?

conqerAtapple:
> This LGTM, but I wonder if it should have a runtime flag to protect the behavior change? it's pretty minor, so maybe not. @alyssawilk, what do you think?

Thanks for review @RyanTheOptimist. I agree about the behavior change. Although  IMHO, this is the right behavior, given that it is a `passthrough`. Maybe a release note?

conqerAtapple:
> Is there a behavior change? sorry if I'm missing something but I don't see where this new function is used? /wait-any

@alyssawilk it is  not a new function but more of a `fix`. Earlier Passthrough socket was not delegating `connect`. This meant that socket implementations like `proxy` which wraps another socket did not delegate the `connect` to the wrapped socket. This effected for example -  custom socket extensions which expected to intercept `connect`.

alyssawilk:
oh, my bad!  I didn't realize the passthrough was an override of the base - usually passthroughs in Envoy are pure interface impls to automatically catch new functions.  I didn't realize in this case we'd implemented in the interface, so didn't catch it.  Just one more question - do you think along with this we should move the implementation out of the base class to discourage things like this that accidentally bypass the passthrough pattern?

conqerAtapple:
> oh, my bad! I didn't realize the passthrough was an override of the base - usually passthroughs in Envoy are pure interface impls to automatically catch new functions. I didn't realize in this case we'd implemented in the interface, so didn't catch it. Just one more question - do you think along with this we should move the implementation out of the base class to discourage things like this that accidentally bypass the passthrough pattern?

Good observation. Trying to understand - 
We had moved from pure TransportSocket interface to an abstract when we added implementation for `connect`.  This was done so that we can let any underlying transport socket implementation (not just PassthroughSocket) to be able to intercept `connect` at the same time not change the existing behavior. But when this transport socket is wrapped by `PassthroughSocket`, it didn't delegate `connect`.
In the case of proxy socket (`UpstreamProxyProtocolSocket`), it wraps the wrapper (`PassthroughSocket`). This implies 2 layers of wrapping. The issue here was that the 2nd layer `PassthroughSocket`  did not delegate `connect` but the rest of the calls. 

Open to suggestions.

alyssawilk:
yeah I think given the complexity as described better to land this as-is.  Thanks for your patience with all the questions!

Issue #9867, opened 2020-01-29
SSL Connection: combine SSL_set_ex_data and SSL_set_app_data


jimini-lumox:
*Title*: *Combine SSL Connection's SSL_set_ex_data and SSL_set_app_data*

*Description*:
PR #9172 introduces additional SSL_set_ex_data to be able to update SslExtendedSocketInfo during SSL validation.
This issue is to refactor the new SSL_set_ex_data and existing SSL_set_app_data to combine to a single SSL_set_app_data.

Currently:
SSL_set_app_data: The ContextImpl's `const Network::TransportSocketOptions*` options set by the ContextImpl but only if (options && !options->verifySubjectAltNameListOverride().empty()) when the SslSocket is created
SSL_set_ex_data: The SslSocketInfo::SslExtendedSocketInfo member set when the SslSocket is subsequently moved to be owned by the constructed SslSocketInfo

PR #21679, opened 2022-06-13
tls: store SslSocket::transport_socket_options_ in SSL object

danzh2010:
Commit Message: Store a pointer to shared pointer `SslSocket::transport_socket_options_` in SSL object via SSL_set_app_data() for both Client and Server ContextImpl. The old behavior is storing the pointer to the object pointed by shared pointer `SslSocket::transport_socket_options_`, and only for ClientContextImpl. 

This  change make it possible the caller of SSL_get_app_data() can extend the life time of the object pointed by `SslSocket::transport_socket_options_`.

Additional Description: interface change to ContextImpl::newSsl()
Risk Level: high, behavioral change in client cert validation, but no external behavioral change.
Testing: existing ssl socket tests pass
Docs Changes: N/A
Release Notes: N/A
Runtime guard: N/A


source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -418,23 +418,38 @@ std::vector<uint8_t> ContextImpl::parseAlpnProtocols(const std::string& alpn_pro
...
   ContextImpl* impl = reinterpret_cast<ContextImpl*>(arg);
   SSL* ssl = reinterpret_cast<SSL*>(
       X509_STORE_CTX_get_ex_data(store_ctx, SSL_get_ex_data_X509_STORE_CTX_idx()));
   auto cert = bssl::UniquePtr<X509>(SSL_get_peer_certificate(ssl));
+  const Network::TransportSocketOptions* transport_socket_options = nullptr;
+  if (impl->tls_async_cert_validation_) {
+    auto transport_socket_options_shared_ptr_ptr =
+        static_cast<const Network::TransportSocketOptionsConstSharedPtr*>(SSL_get_app_data(ssl));
+    ASSERT(transport_socket_options_shared_ptr_ptr);
+    transport_socket_options = (*transport_socket_options_shared_ptr_ptr).get();
```
RyanTheOptimist, line 443: Is the only difference between these two branches the ASSERT()? I can't quite seem to convince myself of it.

RyanTheOptimist (commented):
This looks like a nice small, PR. WOO HOO! :) How hard would it be to verify the new (server-side) behavior in unit tests?

danzh2010:
> xref: #9867
> 
> We might want to create a struct for SSL_set/get_app_data.

I'm not sure about the direction of xref: #9867. I agree we should use only one interface for the same purpose. But SSL_set_ex_data() allows the SSL object to points to multiple places with different indexes, while SSL_set_app_data() only allows one. We could combine extended socket info and transport socket option into a struct, but we also could use SSL_set_ex_data() with different indexes for them. To me SSL_set_ex_data() seems to provide more flexibility and we don't need to refactor ssl socket interfaces.

Anyway, can I keep this PR as it before we clearly know which direction to go?

source/extensions/transport_sockets/tls/context_impl.cc
RyanTheOptimist: Hm. So line 441 calls `SSL_get_app_data(ssl)` and casts it to TransportSocketOptionsConstSharedPtr*. but link 446 calls the same method and casts it to TransportSocketOptions*. This is confusing to me. What is the actual object stored in SSL_get_app_data? Maybe the type of the object is different in the old/new codepath?

danzh2010: Right, in the old path SSL_set_app_data() takes the TransportSocketOptions* which is from TransportSocketOptionsConstSharedPtr, but the in the new path SSL_set_app_data() takes a pointer to TransportSocketOptionsConstSharedPtr itself.

RyanTheOptimist:
> > /wait
> 
> What's the action item?

I asked a question last week about the casting on line 441/446 and I don't see a reply to it. https://github.com/envoyproxy/envoy/pull/21679#discussion_r899180170

Oh, bizarre! Man, github is just the worst UI in history. I don't see a reply on the main page, but if click to the file I see a response there. I'm so sorry about that.  I'll take a look now.

source/extensions/transport_sockets/tls/context_impl.cc
RyanTheOptimist: Ah, I see. Thanks for the explanation!

Storing a pointer to a shared pointer sets off my spidey sense a bit. A shared pointer implies the that object is owned by whoever has the last shared pointer to it, whereas a pointer implies that the object is owned elsewhere. But I guess SSL_set_app_data() requires a raw pointer and we can't put a smart pointer in there, right? And I guess this is safe because we require the caller to own Options and make it outlive the context, right?

RyanTheOptimist (approved):
I'm *really* sorry for the delay! I think this LGTM, but I'd like your confirmation on a question before merging.

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -418,23 +418,38 @@ std::vector<uint8_t> ContextImpl::parseAlpnProtocols(const std::string& alpn_pro
...
+  } else {
+    transport_socket_options =
+        static_cast<const Network::TransportSocketOptions*>(SSL_get_app_data(ssl));
```
ggreenway, line 446: I'm not a fan of storing different types here depending on options. It makes it harder to reason about (given lack of type safety), and much harder to sort it out in gdb if it crashes. I think it would be better to just have 2 indexes, and store the appropriate data in each index. Then if the wrong one is read, it will be a nullptr instead of a valid pointer of the wrong type.

ggreenway (changes requested):
/wait

source/extensions/transport_sockets/tls/context_impl.cc
ggreenway: Oh, I see from the PR description that you want the shared_ptr so that you can extend the lifetime of the object via what's stored in the `SSL`. Can you elaborate on the use case for this?

Given that this PR doesn't yet have any external behavior change, it might simplify to remove the runtime guard.

danzh2010: Okay, reverted runtime guard. PTAL!

danzh2010:
Ping for a review after merge!

ggreenway (approved):

PR #21682, opened 2022-06-13
Shrink Heap: Remove assert in try shrink heap as it can fail if an allocation occurs

KBaichoo:
between when the measurements are taken.

Commit Message: Remove assert in try shrink heap as it can fail if an allocation occurs between when the measurements are taken.
Additional Description: 
Risk Level: low
Testing: 
Docs Changes:
Release Notes:
Platform Specific Features:


KBaichoo:
thanks I've updated the description

PR #19809, opened 2022-02-03
transport socket: refactor to capture the filter state

kyessenov:
Signed-off-by: Kuat Yessenov <kuat@google.com>

Commit Message: Capture the filter state in the transport socket options to be exposed to the internal listener transport socket. The goal is to enable propagation the filter state from the downstream request/connection to the upstream internal listener. The assumption is that the propagation is consistent with the hash key: if multiple downstream HTTP requests share the internal upstream connection, then their filter state objects are "equivalent". 

Risk Level: low, refactor (shared pointer does not seem to prolong the lifetime)
Testing: unit
Docs Changes: none
Release Notes: none
Issue: https://github.com/envoyproxy/envoy/issues/19274

PR #21700, opened 2022-06-14
filter state: add a mark for sharing with the upstream

kyessenov:
Commit Message: Adds a new flag for filter state objects that indicates the intent to share with the upstream.
Additional Description: Follow-up to https://github.com/envoyproxy/envoy/pull/19809. There have been multiple reports of unexpected lifecycle changes for the filter state objects because they are stored in the transport socket options. This PR addresses this issue by introducing a new mark for filter state that explicitly changes the usage of filter state objects:
* marked objects always participate in the connection pool hashing (generalizing and simplifying https://github.com/envoyproxy/envoy/pull/19435);
* marked objects are copied by reference to the upstream info - this allows sharing state between downstream and upstream (and further down the chain, the internal listeners). 

Risk Level: medium, revert to the original behavior prior to https://github.com/envoyproxy/envoy/pull/19809
Testing: yes
Docs Changes: yes
Release Notes: yes

### DIFF
--- a/source/common/network/happy_eyeballs_connection_impl.cc
+++ b/source/common/network/happy_eyeballs_connection_impl.cc
@@ -25,7 +25,8 @@ ClientConnectionPtr HappyEyeballsConnectionProvider::createNextConnection(const 
                   address_list_[next_address_]);
   return dispatcher_.createClientConnection(
       address_list_[next_address_++], source_address_,
-      socket_factory_.createTransportSocket(transport_socket_options_, host_), options_);
+      socket_factory_.createTransportSocket(transport_socket_options_, host_), options_,
+      transport_socket_options_);
 }
 
 size_t HappyEyeballsConnectionProvider::nextConnection() { return next_address_; }


envoy/stream_info/filter_state.h
```diff
@@ -26,7 +26,24 @@ using FilterStateSharedPtr = std::shared_ptr<FilterState>;
  */
 class FilterState {
 public:
-  enum class StateType { ReadOnly, Mutable };
+  enum StateType {
```
ggreenway, line 29: I think it's simpler to just leave this as the previous enum, and add another bool for `SharedWithUpstreamConnection`.

ggreenway (changes requested):
This approach seems good to me.

/wait

envoy/stream_info/filter_state.h
kyessenov: When the upstream L7 filters get done, we'll also need to share with the upstream request I think. I could add another enum for sharing configuration I suppose?

kyessenov:
@ggreenway This should be ready now. I could simplify the recently merged extension, too, which was the only client in-tree for filter state within transport socket options.

envoy/event/dispatcher.h
```diff
@@ -210,13 +210,15 @@ class Dispatcher : public DispatcherBase, public ScopeTracker {
...
-  createClientConnection(Network::Address::InstanceConstSharedPtr address,
-                         Network::Address::InstanceConstSharedPtr source_address,
-                         Network::TransportSocketPtr&& transport_socket,
-                         const Network::ConnectionSocket::OptionsSharedPtr& options) PURE;
+  virtual Network::ClientConnectionPtr createClientConnection(
+      Network::Address::InstanceConstSharedPtr address,
+      Network::Address::InstanceConstSharedPtr source_address,
+      Network::TransportSocketPtr&& transport_socket,
+      const Network::ConnectionSocket::OptionsSharedPtr& options,
+      const Network::TransportSocketOptionsConstSharedPtr& transport_options = nullptr) PURE;
```
lizan, line 221: no default argument for virtual methods per style guide.

lizan (commented):
/lgtm api

breaking API changes approved per exception:
Changes made within 14 days of the introduction of a new API field or message, provided the new field or message has not been included in an Envoy release.

envoy/event/dispatcher.h
kyessenov: Done.

changelogs/current.yaml
```diff
@@ -116,6 +116,10 @@ minor_behavior_changes:
 - area: logging
   change: |
     changed category name for access log filter extensions to ``envoy.access_loggers.extension_filters``.
+- area: filter state
+  change: |
```
ggreenway, line 120: Because the original change never made it into a release, please make the changelog reflect the net-change for this envoy release.

envoy/network/transport_socket.h
```diff
@@ -235,9 +235,10 @@ class TransportSocketOptions {
   virtual absl::optional<Network::ProxyProtocolData> proxyProtocolOptions() const PURE;
 
   /**
-   * @return filter state from the downstream request or connection.
+   * @return filter state objects from the downstream request or connection
+   * that are marked as shared with the upstream connection.
    */
-  virtual const StreamInfo::FilterStateSharedPtr& filterState() const PURE;
+  virtual const StreamInfo::FilterState::Objects& filterStateObjects() const PURE;
```
ggreenway, line 241: I think this function name should be more descriptive, maybe something like `downstreamSharedFilterStateObjects()`.

envoy/stream_info/filter_state.h
```diff
@@ -45,8 +46,34 @@ class FilterState {
   //
   // Note that order matters in this enum because it's assumed that life span grows as enum number
   // grows.
+  //
+  // Note that for more accurate book-keeping it is recommended to subscribe to
+  // the stream callbacks instead of relying on the destruction of the filter
+  // state.
+  //
+  // As a special case, objects that are marked as shared with the upstream
+  // become bound to the upstream connection life span, regardless of the
```
ggreenway, line 55: clarify: the object will live as long as either the downstream or upstream connection is open. The current wording implies that the lifetime is strictly tied to the upstream connection.

envoy/stream_info/filter_state.h
```diff
@@ -66,6 +93,16 @@ class FilterState {
     virtual absl::optional<std::string> serializeAsString() const { return absl::nullopt; }
   };
 
+  struct FilterObject {
+    std::shared_ptr<Object> data_;
+    StateType state_type_;
+    StreamSharing stream_sharing_{StreamSharing::None};
+    std::string name_{""};
```
ggreenway, line 100: The initialization can just be `{}` instead of `{""}`, but empty is the default so it can be omitted entirely.

source/common/network/transport_socket_options_impl.cc
```diff
@@ -40,46 +42,64 @@ void CommonUpstreamTransportSocketFactory::hashKey(
...
     pushScalarToByteVector(StringUtil::CaseInsensitiveHash()(protocol), key);
   }
-}
 
-TransportSocketOptionsConstSharedPtr TransportSocketOptionsUtility::fromFilterState(
-    const StreamInfo::FilterStateSharedPtr& filter_state) {
-  if (!filter_state) {
-    return nullptr;
+  for (const auto& object : options->filterStateObjects()) {
+    if (auto hashable = dynamic_cast<const Hashable*>(object.data_.get()); hashable != nullptr) {
```
ggreenway, line 47: What happens if an object isn't `Hashable`? Can we end up with bugs due to matching transport socket options hash even though it contains different filter state? Should `FilterState::Object` always inherit from `Hashable`? Or maybe make it required when the object opts in to sharing, via an accessor like `getHashable()`? We mostly try to avoid `dynamic_cast` in envoy for this type of feature detection.

source/extensions/io_socket/user_space/io_handle_impl.cc
```diff
@@ -391,18 +392,16 @@ void PassthroughStateImpl::mergeInto(envoy::config::core::v3::Metadata& metadata
...
-                             StreamInfo::FilterState::LifeSpan::Connection);
-      } catch (const EnvoyException& e) {
-        ENVOY_LOG(trace, "Failed to set data for '{}': {}", name, e.what());
-      }
+  for (const auto& object : filter_state_objects_) {
+    try {
+      filter_state.setData(object.name_, object.data_, object.state_type_,
+                           StreamInfo::FilterState::LifeSpan::Connection, object.stream_sharing_);
+    } catch (const EnvoyException& e) {
+      ENVOY_LOG(trace, "Failed to set data for '{}': {}", object.name_, e.what());
```
ggreenway, line 400: Ignoring this error seems unsafe; something important may have been discarded. Are there cases where this is expected to happen? Should this have an `ENVOY_BUG()`?

ggreenway (changes requested):
/wait

changelogs/current.yaml
kyessenov: The original PR did make it into 1.22 AFAICT. It's from 2/23.

envoy/network/transport_socket.h
kyessenov: Renamed.

envoy/stream_info/filter_state.h
kyessenov: Clarified, LMK if it can be improved further.

envoy/stream_info/filter_state.h
kyessenov: Done.

envoy/stream_info/filter_state.h
```diff
@@ -66,6 +96,16 @@ class FilterState {
     virtual absl::optional<std::string> serializeAsString() const { return absl::nullopt; }
   };
 
+  struct FilterObject {
+    std::shared_ptr<Object> data_;
+    StateType state_type_;
```
lizan, line 101:
```suggestion
    StateType state_type_{};
```
default initialize this?

kyessenov: Done.

lizan (commented):
LGTM, can you resolve conflicts?

ggreenway (approved):
LGTM

Issue #21410, opened 2022-05-23
New, non-idempotent DNS-related 503 behavior


ravenblackx:
In an envoy version from mid-December, a request for a URL whose DNS isn't present would still go to the filters and thereby present a 403 error in one of our tests.

In an envoy version from head, the same request pre-empts the filters and sends a 503 localReply.

```
D0523 17:24:16.700779 346 forward_proxy dns_cache_impl.cc:88] thread local lookup for host 'wdsafjeaasjkerhajkf.com'
D0523 17:24:16.700793 346 forward_proxy dns_cache_impl.cc:109] cache miss for host 'wdsafjeaasjkerhajkf.com', posting to main thread
D0523 17:24:16.700832 346 forward_proxy proxy_filter.cc:161] [C305][S2554531192191987142] waiting to load DNS cache entry
D0523 17:24:16.700901 307 forward_proxy dns_cache_impl.cc:284] starting main thread resolve for host='wdsafjeaasjkerhajkf.com' dns='wdsafjeaasjkerhajkf.com' port='443'
D0523 17:24:16.700942 307 dns dns_impl.cc:338] dns resolution for wdsafjeaasjkerhajkf.com started
D0523 17:24:16.714733 307 dns dns_impl.cc:140] dns resolution for wdsafjeaasjkerhajkf.com failed with c-ares status 4
D0523 17:24:16.714783 307 dns dns_impl.cc:259] dns resolution for wdsafjeaasjkerhajkf.com completed with status 1
D0523 17:24:16.714796 307 forward_proxy dns_cache_impl.cc:307] main thread resolve complete for host 'wdsafjeaasjkerhajkf.com': []
D0523 17:24:16.714930 346 forward_proxy proxy_filter.cc:206] [C305][S2554531192191987142] load DNS cache complete, continuing after adding resolved host: wdsafjeaasjkerhajkf.com
D0523 17:24:16.714973 307 forward_proxy dns_cache_impl.cc:402] DNS refresh rate reset for host 'wdsafjeaasjkerhajkf.com', (failure) refresh rate 60000 ms
D0523 17:24:16.714996 346 http filter_manager.cc:952] [C305][S2554531192191987142] Sending local reply with details dns_resolution_failure
```

Doing the exact same request a moment later *does not* inject the same 503, resulting in extra surprising non-idempotent behavior - the second identical request receives the expected 403, in this case.

```
[...]

PR #21701, opened 2022-06-14
dns: adding double test with no cache

alyssawilk:
More trying to repro #21410


test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
```diff
@@ -176,6 +171,8 @@ name: envoy.clusters.dynamic_forward_proxy
   envoy::config::cluster::v3::Cluster cluster_;
   std::string cache_file_value_contents_;
   bool use_cache_file_{};
+  std::string filename_{};
+  std::string key_value_config_{};
```
RyanTheOptimist, line 175: nit: no need for {} on these as the default constructor will do the needful.

RyanTheOptimist (approved):
LGTM though it looks like there is a format issue?

RyanTheOptimist (approved):

PR #21703, opened 2022-06-14
[cares dns] Use AF_UNSPEC for dual family resolution

Reflejo:
Commit Message: [cares dns] Use AF_UNSPEC for dual family resolution
Additional Description: Right now we are calling two parallel DNS resolution with stateful branching which seems pretty brittle on reused channels. This will make the lookup all case perform one `getaddrinfo` call.
Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Android/Linux


test/extensions/network/dns_resolver/cares/dns_impl_test.cc
```diff
@@ -1576,7 +1576,9 @@ TEST_P(DnsImplTest, FallbackToNodataWithErrorOnAAAA) {
   EXPECT_NE(nullptr,
             resolveWithNoRecordsExpectation("some.good.domain", DnsLookupFamily::V4Preferred));
   dispatcher_->run(Event::Dispatcher::RunType::Block);
-  EXPECT_NE(nullptr, resolveWithNoRecordsExpectation("some.good.domain", DnsLookupFamily::All));
+  EXPECT_NE(nullptr,
+            resolveWithExpectations("some.good.domain", DnsLookupFamily::All,
+                                    DnsResolver::ResolutionStatus::Failure, {}, {}, absl::nullopt));
```
mattklein123, line 1581: Why this test change?

mattklein123 (changes requested):
Looks like there is still some problem here with CI? I can take a look today as I'm going to work on a getaddrinfo() resolver.

/wait

test/extensions/network/dns_resolver/cares/dns_impl_test.cc
mattklein123: OK I understand now. I think technically this is a behavior change then? I think it's probably fine, especially for ALL, but can you add a comment or maybe move this case to a different test since the title is FallbackToNodataWithErrorOnAAAA? Thank you.

/wait

mattklein123 (approved):
Thank you!

PR #21708, opened 2022-06-15
bazel: fix useless info warning

daixiang0:
Signed-off-by: Loong Dai <loong.dai@intel.com>

The warning itself is not useful anyway:
```
WARNING: info command does not support starlark options. Ignoring options: [--@com_googlesource_googleurl//build_config:system_icu=0]
```

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -414,3 +414,5 @@ build:windows --dynamic_mode=off
 try-import %workspace%/clang.bazelrc
 try-import %workspace%/user.bazelrc
 try-import %workspace%/local_tsan.bazelrc
+
+info --ui_event_filters=-WARNING


.bazelrc
```diff
@@ -414,3 +414,5 @@ build:windows --dynamic_mode=off
 try-import %workspace%/clang.bazelrc
 try-import %workspace%/user.bazelrc
 try-import %workspace%/local_tsan.bazelrc
+
+info --ui_event_filters=-WARNING
```
phlax, line 418: i dont think we want to filter all warnings (for info)

im wondering if instead of suppressing the warnings we could prevent them

daixiang0: I think we do not miss anything import:
```
root:[envoy]$ bazel info &>old
root:[envoy]$ bazel info --ui_event_filters=-WARNING &>new
root:[envoy]$ diff old new
1d0
< WARNING: info command does not support starlark options. Ignoring options: [--@com_googlesource_googleurl//build_config:system_icu=0]
9,10c8,9
< gc-count: 15
< gc-time: 272ms
---
> gc-count: 16
> gc-time: 328ms
23c22
< used-heap-size: 35MB
---
> used-heap-size: 36MB
```

ggreenway:
@phlax please review

.bazelrc
phlax: when i run the command with this flag i lose deprecation warnings - i dont think warnings should be supressed - how about trying to resolve the problem itself

/wait

phlax:
as per above comment i think we need to address the thing that is triggering a warning, not supress all warnings

/wait

daixiang0:
Will be fixed in bazel 6.0, see https://github.com/bazelbuild/bazel/issues/13473

moderation:
According to https://github.com/bazelbuild/bazel/issues/13473#issuecomment-1088014392= we should already have this fix with https://github.com/envoyproxy/envoy/blob/main/.bazelversion

daixiang0:
@moderation I miss it,  has commented in that issue.

daixiang0:
From https://github.com/bazelbuild/bazel/issues/13473#issuecomment-1164496845, now we can not make it hide.

PR #21713, opened 2022-06-15
http: removing throw

alyssawilk:
Removing the http status code check which throws an exception. By default status will now ENVOY_BUG and return 0 if the status header is not present.
May break downstream unit tests which relied on the throw (as I had to refactor envoy unit tests accordingly)

Risk Level: low
Testing: updated tests
Docs Changes: n/a
Release Notes: n/a


yanavlasov (approved):

yanavlasov:
It looks like ti should really be a clean-up project to get rid of the problematic function altogether. I see its usage is very high, including proprietary filters, so it will be quite a slog.

PR #21714, opened 2022-06-15
file system buffer: Prevent 0 byte memory buffers

vehre-x41:
Commit Message: file system buffer: Prevent 0 byte memory buffers
Additional Description:
Throw an exception when the configured memory buffer size 0 to prevent
an endless loop eating memory for the overhead of allocating management
structures for zero byte buffers.

Prevent that the fuzzer creates a configuration for the
file_system_buffer where either response's or request's
memory_buffer_bytes_limit is zero.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/api/envoy/extensions/filters/http/file_system_buffer/v3/file_system_buffer.proto
+++ b/api/envoy/extensions/filters/http/file_system_buffer/v3/file_system_buffer.proto
@@ -71,7 +71,7 @@ message StreamConfig {
 
   // The amount stored in the memory buffer before buffering to disk.
   // If unset in route, vhost and listener config, defaults to a hardcoded value of 1MiB
-  google.protobuf.UInt64Value memory_buffer_bytes_limit = 2;
+  google.protobuf.UInt64Value memory_buffer_bytes_limit = 2 [(validate.rules).uint64 = {gt: 0}];
 
   // The maximum storage (excluding memory) to be buffered in this filter.
   // If unset in route, vhost and listener config, defaults to a hardcoded value of 32MiB
--- a/source/extensions/filters/http/file_system_buffer/filter_config.cc
+++ b/source/extensions/filters/http/file_system_buffer/filter_config.cc
@@ -80,11 +80,7 @@ namespace {
 size_t getMemoryBufferBytesLimit(const StreamConfigVector& configs) {
   for (const ProtoStreamConfig& cfg : configs) {
     if (cfg.has_memory_buffer_bytes_limit()) {
-      if (cfg.memory_buffer_bytes_limit().value() > 0) {
-        return cfg.memory_buffer_bytes_limit().value();
-      } else {
-        throw EnvoyException("memory_buffer_bytes_limit must not be zero");
-      }
+      return cfg.memory_buffer_bytes_limit().value();
     }
   }
   return FileSystemBufferFilterMergedConfig::default_memory_buffer_bytes_limit;
--- a/source/extensions/filters/http/file_system_buffer/filter_config.cc
+++ b/source/extensions/filters/http/file_system_buffer/filter_config.cc
@@ -80,7 +80,11 @@ namespace {
 size_t getMemoryBufferBytesLimit(const StreamConfigVector& configs) {
   for (const ProtoStreamConfig& cfg : configs) {
     if (cfg.has_memory_buffer_bytes_limit()) {
-      return cfg.memory_buffer_bytes_limit().value();
+      if (cfg.memory_buffer_bytes_limit().value() > 0) {
+        return cfg.memory_buffer_bytes_limit().value();
+      } else {
+        throw EnvoyException("memory_buffer_bytes_limit must not be zero");
+      }
     }
   }
   return FileSystemBufferFilterMergedConfig::default_memory_buffer_bytes_limit;
--- a/test/extensions/filters/http/common/fuzz/filter_corpus/oss-fuzz-47869-filter_fuzz_test-5617673892921344
+++ b/test/extensions/filters/http/common/fuzz/filter_corpus/oss-fuzz-47869-filter_fuzz_test-5617673892921344
@@ -1,0 +1,16 @@
+config {
+  name: "envoy.filters.http.file_system_buffer"
+  typed_config {
+    type_url: "type.googleapis.com/envoy.extensions.filters.http.file_system_buffer.v3.FileSystemBufferFilterConfig"
+    value: "\"\002\022\000"
+  }
+}
+data {
+  trailers {
+  }
+}
+upstream_data {
+  http_body {
+    data: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000"
+  }
+}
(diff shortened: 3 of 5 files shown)


botengyao (commented):
Thanks for working on it!

We probably still need to update the [API](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/file_system_buffer/v3/file_system_buffer.proto#envoy-v3-api-msg-extensions-filters-http-file-system-buffer-v3-filesystembufferfilterconfig) doc. PGV & comments could be an option.

vehre-x41:
I tried to use PGV `[(validate.rules).uint64 = { gt: 0 }]` (or very similar) for `memory_buffer_bytes_limit`, but the PGV was not triggered. At least not by the corpus supplied, because in the corpus `memory_buffer_bytes_limit {}` is used, which of course sets no (numeric) `value` but only default initializes. This somehow does not trigger the PGV rule.

As for the other members `storage_buffer_bytes_limit` and `storage_buffer_queue_high_watermark_bytes` I was not sure, if there could be some usecases where one might want to explicit set those to zero to prevent running into a bug or the like. I have something in mind, like using file storage in a certain environment triggers a bug and a first work-around can be to prevent using file storage by setting the `storage_buffer_bytes_limit` to zero, which then deactivates the feature. (Just how I think about this. It might not work at all. But as a user I might me tempted to try something like this as an emergency mitigation).

adisuissa:
> I tried to use PGV `[(validate.rules).uint64 = { gt: 0 }]` (or very similar) for `memory_buffer_bytes_limit`, but the PGV was not triggered. At least not by the corpus supplied, because in the corpus `memory_buffer_bytes_limit {}` is used, which of course sets no (numeric) `value` but only default initializes. This somehow does not trigger the PGV rule.
> 
Interesting.
I've tried changing it to the following:
```
google.protobuf.UInt64Value memory_buffer_bytes_limit = 2 [(validate.rules).uint64 = {lte: 512 gte: 1}];
```
and it seems that the config was rejected.

> As for the other members `storage_buffer_bytes_limit` and `storage_buffer_queue_high_watermark_bytes` I was not sure, if there could be some usecases where one might want to explicit set those to zero to prevent running into a bug or the like. I have something in mind, like using file storage in a certain environment triggers a bug and a first work-around can be to prevent using file storage by setting the `storage_buffer_bytes_limit` to zero, which then deactivates the feature. (Just how I think about this. It might not work at all. But as a user I might me tempted to try something like this as an emergency mitigation).

I agree. I think @ravenblackx may be able to help with assessing this.

ravenblackx:
I agree that a memory_buffer_bytes_limit of 0 should be forbidden, thanks for proposing the change.

I don't think the same is true for storage_buffer_bytes_limit - there is even [a case](https://github.com/envoyproxy/envoy/blob/4a2835afcd36dd6077768aac6235b40c9db68882/test/extensions/filters/http/file_system_buffer/filter_test.cc#L427) in the unit tests that intentionally uses storage_buffer_bytes_limit=0, both to ensure correct behavior for that configuration, and to simplify testing of the watermark behavior. (It does deactivate the file system part of the buffer, while continuing to perform the general buffering behavior.)

ravenblackx:
If making a change to the API with respect to validation of memory_buffer_bytes_limit, please ensure that unset is still a valid option, as the value might be inherited in a route config, or overridden by a route config.

Also, adding a death test like [this one](https://github.com/envoyproxy/envoy/blob/4a2835afcd36dd6077768aac6235b40c9db68882/test/extensions/filters/http/file_system_buffer/config_test.cc#L83) for the new exception would be nice for coverage!

vehre-x41:
> Interesting.
I've tried changing it to the following:
`google.protobuf.UInt64Value memory_buffer_bytes_limit = 2 [(validate.rules).uint64 = {lte: 512 gte: 1}];`
and it seems that the config was rejected.

That's odd. I tried the expression given and it does not reject the input from the corpus. Only the latter exception prevents the program from failing. So what do we do? Add the PGV or not?

vehre-x41:
@ravenblackx @adisuissa Any thoughts on the latest changes?

adisuissa:
> That's odd. I tried the expression given and it does not reject the input from the corpus. Only the latter exception prevents the program from failing. So what do we do? Add the PGV or not?

I've tested it in https://github.com/adisuissa/envoy/commit/add8358fd322243841eed5b697e1722f9e3b18be.
Can you check if this works on your end (trying to understand where the difference is coming from).

Generally speaking we would like to have more PGV constraints, as this is part of the contract (API) that we support.

adisuissa:
@ravenblackx can you please review the fields that should be bounded and suggest the correct limits (lower and upper bound, if applicable).

ravenblackx:
Change looks good to me.

@adisuissa I don't know if there should be hard bounds other than the memory buffer must not be explicitly zero. Like a memory buffer of a terabyte is *probably* silly [at least for a few years], but it's not invalid. Zero disk buffer is explicitly valid. A memory buffer of 1 byte is silly but valid (and setting a "sensible" minimum on that makes its behavior harder to test.) So the only thing I would endorse limiting in the API is memory buffer must not be zero (but must also be allowed to be unset - I don't know if the validator can make that distinction.)

ravenblackx (dismissed):

vehre-x41:
> > That's odd. I tried the expression given and it does not reject the input from the corpus. Only the latter exception prevents the program from failing. So what do we do? Add the PGV or not?
> 
> I've tested it in [adisuissa@add8358](https://github.com/adisuissa/envoy/commit/add8358fd322243841eed5b697e1722f9e3b18be). Can you check if this works on your end (trying to understand where the difference is coming from).
> 
> Generally speaking we would like to have more PGV constraints, as this is part of the contract (API) that we support.

I can't figure what I did wrong, but now PGV works. Therefore, did I revert the custom check and exception as well as insert the PGV and adapted the tests for `memory_buffer_bytes_limit > 0`.

ravenblackx (approved):

ravenblackx:
Thanks! I appreciate the fix and also the lesson in validation rules. :)

adisuissa:
> I agree that a memory_buffer_bytes_limit of 0 should be forbidden, thanks for proposing the change.
> 
> I don't think the same is true for storage_buffer_bytes_limit - there is even [a case](https://github.com/envoyproxy/envoy/blob/4a2835afcd36dd6077768aac6235b40c9db68882/test/extensions/filters/http/file_system_buffer/filter_test.cc#L427) in the unit tests that intentionally uses storage_buffer_bytes_limit=0, both to ensure correct behavior for that configuration, and to simplify testing of the watermark behavior. (It does deactivate the file system part of the buffer, while continuing to perform the general buffering behavior.)

@ravenblackx will it make sense to add upper bounds to any of the config knobs, or do you think it's unnecessary?

adisuissa:
Gotcha... Although I think that 640KB should suffice :)

Just to present a counterpoint, there are cases where one would like to have a more strict bounds to avoid misconfiguration by the users, and not only for fuzzing.
This is done in [`max_request_headers_kb `](https://github.com/envoyproxy/envoy/blob/3e3721fd6cb2fb70dec1e082c5351991f64432f0/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L439-L440) for example.
Also note that setting this value doesn't mean the threshold is set in stone. In the future, the range can be extended (but not narrowed). This is because it won't break any existing user configs.

Not saying we should change these value, just giving another perspective.

PR #21716, opened 2022-06-15
codeowners: calling attention to newly underowned extensions

alyssawilk:
Commit Message: n/a
Risk Level: n/a
Testing: n/a
Docs Changes: yes


alyssawilk:
@yanavlasov can you suggest anyone to take a look at the newly unowned extensions and find google owners for ones we care about? 
I've pinged pinterest to see if we can find another owner for Thrift.  Unsure about the others

alyssawilk:
cc @krajshiva alternately perhaps you could find someone on EPT to check on these?

PR #21722, opened 2022-06-15
proto: fixing a few more warnings

alyssawilk:
cc @phlax these showed up when I ran one of the check_format scripts - I wonder if we could make it fail


phlax:
i think we should, altho i havent figured out a way to do it yet - i think `buf` might provide buf linting so i was going to look at that

iiuc, altho  warnings are being issued somewhere i think bazel sees them as info messages which makes it harder to catch

phlax:
re above when we figure it out i think we also want to add in xds repo

mattklein123 (approved):

Issue #7763, opened 2019-07-30
support URI Template based route match


lizan:
*Description*:
This was a part of discussion for https://github.com/envoyproxy/envoy/issues/7728 but a separate issue, to add full or subset of URI Template (RFC6570) based route matching. The URI Template is already used in OpenAPI spec and gRPC JSON transcoding binding.

It will reduce the use case for regex as I see many use of regex is to support this.

justin-mp:
RFC 6570 is great for creating URIs, but seems mum on actual matching patterns.  What would be the plan for turning URI Templates into matchers?

I ask because we're in need of a more flexible route-matching, but are trying to avoid exposing the complexity (both in user-friendliness and in runtime speed) of regexes.  We're thinking of building something based on "glob" operators:

> The "glob" operators here follow common UNIX command-line conventions in order to be as familiar and flexible as possible. 
 For non-URL pattern matching, such as in header/cookie values and/or in CORS policies, the `*` operator matches any number of (zero or more) characters.
>
> * The "*" operator, which matches zero or more characters, up to the next path separator ("/")
>   * Example: "/api/v1/*" would match "/api/v1/get" but not "/api/v1/resourceName/post"
>   * Similarly, “/videos/*.ts” would match “/videos/foo.ts” but not “/videos/id/1234/foo.ts”
> * The "**" operator, which matches zero or more characters, ignoring any path separators.
>   * Example: "/videos/**.ts" would match "/videos/hd/18313/output/hevc/segment-000001.ts", "/videos/sd/90574/output/dash/segment-000555.ts" and "/videos/test/foo.ts"
> * The "?" operator, which matches exactly one character.
>    * This can only be specified once (use `*` or `**` if you need to match multiple characters)
>   * Example: "/api/v?/**" would match "/api/v1/hello/get" but not "/api/v1beta1/hello/get"

The gRPC transcoder extends the RFC 6570 syntax so that expansion expressions can take matcher expressions. The matcher expressions are of the form `{name=TEMPLATE}` where `TEMPLATE` can expand to a combination of literals and glob operators (including `*` and `**`!).  See [http.proto](https://github.com/googleapis/googleapis/blob/f829b1334cce86aa3738f3c0698d814b56664445/google/api/http.proto#L208) for more details.

gRPC's `TEMPLATE` doesn't allow single-character `?`, but presumably that wouldn't be hard to add especially since `?` is already a reserved character for gRPC's `TEMPLATE` syntax.

[...]

PR #21725, opened 2022-06-15
Adding proto changes for new pattern based url matching and rewriting

silverstar194:
Commit Message: Adding proto changes for new pattern based url matching and rewriting

Additional Description:
The overarching goal of this PR (and the additional PRs to follow) is to provide support for matching pattern based paths and provide support for rewriting pattern based paths.

Taken from https://cloud.google.com/media-cdn/docs/routing#pattern-matching :
Pattern matching lets you match multiple parts of a URL, including partial URLs and suffixes (file extensions), using a simple wildcard syntax.

You can also associate one or more path components with named variables, and then refer to those variables when rewriting the URL. This lets you reorder and remove URL components before the request is sent to your origin.

This is one PR in a series breaking apart changes in https://github.com/envoyproxy/envoy/pull/21614 for easy review. End result will be pattern based matching for routes and pattern based rewriting for routes.

See above link for more detailed docs and examples. Parity between linked Media CDN product is expected.

Risk Level: Low
More functionality is being added but no existing functionality should change.

Testing: Unit tests for proto config load

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


api/envoy/config/route/v3/route_components.proto
```diff
@@ -542,6 +542,8 @@ message RouteMatch {
     //
     // Expect the value to not contain ``?`` or ``#`` and not to end in ``/``
     string path_separated_prefix = 14 [(validate.rules).string = {pattern: "^[^?#]+[^?#/]$"}];
+
+    string path_template_match = 15 [(validate.rules).string = {min_len: 1}];
```
htuch, line 546: Please add docs and link to the issue in the PR.

silverstar194: Added docstrings for both new proto fields

api/envoy/config/route/v3/route_components.proto
```diff
@@ -542,6 +542,19 @@ message RouteMatch {
     //
     // Expect the value to not contain ``?`` or ``#`` and not to end in ``/``
     string path_separated_prefix = 14 [(validate.rules).string = {pattern: "^[^?#]+[^?#/]$"}];
+
+    // If specified, the route is a template match rule meaning that the
+    // ``:path`` header (without the query string) must match the given
+    // ``path_template_match`` pattern.
+    //
+    // For example:
```
edin44, line 550: These examples are good, but I think it would be worth explicitly stating what * means, what ** means, and what {foo} is, instead of only relying on the examples.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -542,6 +542,19 @@ message RouteMatch {
...
+    //
+    // * ``/videos/*/*/*.m4s`` would match `videos/123414/hls/1080p5000_00001.m4s``
+    //
+    // * ``/videos/{file}`` would match ``/videos/1080p5000_00001.m4s"
+    //
+    // * ``/**.mpd`` would match ``/content/123/india/dash/55/manifest.mpd``
+    string path_template_match = 15 [(validate.rules).string = {min_len: 1}];
```
edin44, line 557: ISTM path_pattern_match would be a more consistent name than path_template_match given the other code changed in this PR. Are you using "template" to be consistent with terminology that already exists somewhere?

source/common/common/matchers.cc
```diff
@@ -110,6 +110,14 @@ PathMatcherConstSharedPtr PathMatcher::createPrefix(const std::string& prefix, b
   return std::make_shared<const PathMatcher>(matcher);
 }
 
+PathMatcherConstSharedPtr PathMatcher::createPattern(const std::string& pattern, bool ignore_case) {
+  // TODO: implement pattern specific matcher
```
edin44, line 114: Nit: TODO(silverstar194)

source/common/common/matchers.cc
```diff
@@ -110,6 +110,14 @@ PathMatcherConstSharedPtr PathMatcher::createPrefix(const std::string& prefix, b
   return std::make_shared<const PathMatcher>(matcher);
 }
 
+PathMatcherConstSharedPtr PathMatcher::createPattern(const std::string& pattern, bool ignore_case) {
```
edin44, line 113: This should be tested in matchers_test.cc.

test/common/router/config_impl_test.cc
```diff
@@ -5050,6 +5050,24 @@ TEST_F(RouteMatcherTest, TestPrefixAndRegexRewrites) {
       "Cannot specify both prefix_rewrite and regex_rewrite");
 }
 
+TEST_F(RouteMatcherTest, TestPatternRewriteConfigLoad) {
```
edin44, line 5053: Does this actually test the functions you added in config_impl.cc?

api/envoy/config/route/v3/route_components.proto
silverstar194: Yes it matches with the Cloud Media config 
(https://cloud.google.com/media-cdn/docs/routing#path-matching)

source/common/common/matchers.cc
edin44: Since it currently behaves like PathMatcher and there's a TODO then maybe it's not too meaningful to add tests at this point. Optional: Maybe make it clear in the TODO that unit tests need to be added for the whole function.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -1042,6 +1056,32 @@ message RouteAction {
...
+  // :ref:`regex_rewrite <envoy_v3_api_field_config.route.v3.RouteAction.regex_rewrite>`,
+  // or *pattern_rewrite* may be specified.
+  //
+  // Examples using pattern_rewrite:
+  //
+  // * The pattern ``/{one}/{two}`` paired with a substitution string of ``/{two}/{one}`` would
+  //   transform ``/cat/dog`` into ``/dog/cat``.
+  //
+  // * The pattern ``/videos/{language=lang/*}/*`` paired with a substitution string of
+  //   ``\{language}`` transforming ``/videos/lang/en/video.m4s`` into ``lang/en``.
```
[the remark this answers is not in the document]
silverstar194: done

source/common/common/matchers.cc
silverstar194: Changed

source/common/common/matchers.cc
silverstar194: It does not perform anything useful yet. Implementation and tests will come in text PR

test/common/router/config_impl_test.cc
silverstar194: For now I am just testing config is loaded and fields are populated

api/envoy/config/route/v3/route_components.proto
lizan: Is this also the path template matching also used in gRPC transcoding? https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#httprule

silverstar194: Yes very similar markup and pattern rules it looks like

edin44 (approved):

api/envoy/config/route/v3/route_components.proto
```diff
@@ -542,6 +542,33 @@ message RouteMatch {
...
+    //
+    //
+    // For example:
+    //
+    // * ``/videos/*/*/*.m4s`` would match ``videos/123414/hls/1080p5000_00001.m4s``
+    //
+    // * ``/videos/{file}`` would match ``/videos/1080p5000_00001.m4s``
+    //
+    // * ``/**.mpd`` would match ``/content/123/india/dash/55/manifest.mpd``
+    string path_template_match = 15 [(validate.rules).string = {min_len: 1}];
```
stevenzzzz, line 571: pls add max_len validation as well, the g-doc says max length 256, I think that's very reasonable.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -1042,6 +1070,48 @@ message RouteAction {
...
+  // * The pattern ``/{one}/{two}`` paired with a substitution string of ``/{two}/{one}`` would
+  //   transform ``/cat/dog`` into ``/dog/cat``.
+  //
+  // * The pattern ``/videos/{language=lang/*}/*`` paired with a substitution string of
+  //   ``\{language}`` would transform ``/videos/lang/en/video.m4s`` into ``lang/en``.
+  //
+  // * The path pattern ``/content/{format}/{lang}/{id}/{file}.vtt`` paired with a substitution
+  //   string of ``/{lang}/{format}/{file}.vtt`` would transform ``/content/hls/en-us/12345/en_193913.vtt``
+  //   into ``/en-us/hls/en_193913.vtt``.
+  string path_template_rewrite = 41;
```
stevenzzzz, line 1113: add len validation rules on max_len, min_len?

silverstar194: Added validations for length based on spec doc

api/envoy/config/route/v3/route_components.proto
silverstar194: Added

api/envoy/config/route/v3/route_components.proto
```diff
@@ -542,6 +542,34 @@ message RouteMatch {
...
+
+    // If specified, the route is a template match rule meaning that the
+    // ``:path`` header (without the query string) must match the given
+    // ``path_template_match`` pattern.
+    //
+    // Path template matching types:
+    //
+    // * ``*`` : Matches a single path component, up to the next path separator: /
+    //
+    // * ``**`` : Matches zero or more path segments. If present, must be the last operator.
```
htuch, line 554: Is there a formal grammar we can point people to as well? It doesn't have to be in the docs, but unlike somehting like RE2, this is not an open standard as such. @mattklein123 I'd like to confirm here that we want this as a 1st class API - I think this is a good idea as it's not regex and for the various reasons discussed historically. OTOH, it could theoretically be a pluggable regex engine (if you excuse the misnomer). The main challenge would be to include the substitution language as well I think.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -1042,6 +1071,49 @@ message RouteAction {
...
+  // Only named matches can be used to perform rewrites.
+  //
+  //
+  // Examples using path_template_rewrite"
+  //
+  // * The pattern ``/{one}/{two}`` paired with a substitution string of ``/{two}/{one}`` would
+  //   transform ``/cat/dog`` into ``/dog/cat``.
+  //
+  // * The pattern ``/videos/{language=lang/*}/*`` paired with a substitution string of
+  //   ``\{language}`` would transform ``/videos/lang/en/video.m4s`` into ``lang/en``.
```
htuch, line 1109: ``/{language}''?

silverstar194: done

api/envoy/config/route/v3/route_components.proto
htuch: Friendly ping @mattklein123

mattklein123: Sorry I missed this originally.

What are the reasons this is not regex? I assume performance and simplicity of implementation? Can we potentially spell out the reasons why someone would use this over regex because as you say regex with pattern matching could work also?

Also I agree that if this is some standard that is going to be shared across products it should be documented somewhere canonical?

htuch: This is basically https://github.com/envoyproxy/envoy/issues/7763. The main thing I wanted to clarify is that, given the difference in syntax between regex and URL templates/globs, it would make more sense to treat it as first class rather than a pseudo-regex "engine.

markdroth: FWIW, I don't think it makes sense to treat this as a regex engine, since it's not regex syntax.  But I do wonder if it would make sense to add an extension point for a generic path-matching plugin here, and this could then be an extension that plugs into that extension point rather than a first-class mechanism baked directly into the API.  I'm not sure how many people will actually use this, and adding an extension point would also pave the way for future custom path-matching extensions when others want to do this differently.

mattklein123: Yeah I'm fine either way between extension or not, but I think we should more clearly document the rationale behind why this exists and why one would use this over regex. Also, if this really is shared between other systems as I said before I think it would be better if there was some canonical documentation on the syntax.

silverstar194: @mattklein123 and @htuch It sounds like the only action item here is additional docs. Are we ok to approve these set of changes?

htuch: From my side, correct. As a tiny nit I might rename this field to `path_template`, but that's all API wise.

silverstar194: I wanted to follow the same conversation as Media CDN where it's path_template_match. https://cloud.google.com/media-cdn/docs/routing#path-matching

If we're all set LGTM

htuch: Prefer to preference Envoy API consistency here rather than a specific vendor product. LGTM with that change and the not-impelemnted-hide stuff we discussed.

envoy/router/router.h
```diff
@@ -782,6 +782,7 @@ enum class PathMatchType {
   Exact,
   Regex,
   PathSeparatedPrefix,
+  Pattern,
```
stevenzzzz, line 785: is pathTemplate cleaner than Pattern? 
no strong feeling tho.

stevenzzzz:
LGTM
/lgtm

api/envoy/config/route/v3/route_components.proto
silverstar194: Changed name and added not-impelemnted-hide stuff. Should be all set

htuch (commented):
/lgtm api

htuch:
For docs, @silverstar194 is planning on adding them in a distinct PR before the `not-implemented-hide` is removed, so seems reasonable from an API perspective to merge. @silverstar194 you might want to reject the field in this PR IMHO until we implement.

We need a non-Google reviewer to also take a look here. @mattklein123 can you do this?

silverstar194:
@mattklein123 Feel free to merge once approved. Looks like I don't have permissions to complete

mattklein123 (approved):
LGTM, thanks. Please remember to add a release note when you unhide the documentation. Also, per previous discussion, there still are not any of the documentation modifications that we previously discussed such as rationale, why someone would use this over regex, etc. so this will also need to be provided in the next PR during implementation and unhiding.

silverstar194:
Noted. Will be sure to add before release.

PR #21727, opened 2022-06-15
bazel: add -no_deduplicate for dbg builds on darwin

keith:
This is a linker optimization that trades link time for binary size.
This change enables -no_deduplicate for dbg and fastbuild, which on my
M1 Max mbp takes link time of the main envoy binary from 60s to 12s on
macOS.

### DIFF
--- a/bazel/envoy_internal.bzl
+++ b/bazel/envoy_internal.bzl
@@ -154,6 +154,13 @@ def envoy_stdlib_deps():
         "//conditions:default": ["@envoy//bazel:static_stdlib"],
     })
 
+def envoy_dbg_linkopts():
+    return select({
+        # TODO: Remove once we have https://github.com/bazelbuild/bazel/pull/15635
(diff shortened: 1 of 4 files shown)


bazel/envoy_internal.bzl
```diff
@@ -154,6 +154,13 @@ def envoy_stdlib_deps():
         "//conditions:default": ["@envoy//bazel:static_stdlib"],
     })
 
+def envoy_dbg_linkopts():
+    return select({
+        # TODO: Remove once we have https://github.com/bazelbuild/bazel/pull/15635
+        "@envoy//bazel:apple_non_opt": ["-Wl,-no_deduplicate"],
```
keith, line 160: naming is hard, dbg, and non_opt, i don't like either, up for suggestions here. Really I want a single word that conveys both "fastbuild and debug"

lizan (approved):

PR #21728, opened 2022-06-16
fine-grain logger: add verbosity default log level support in fine-grain logger

botengyao:

Adds the support method for verbosity default log level update.
* Verbosity default log level is stateful.
* Glob update vector has higher priority than verbosity default log level.
* The verbosity default log level could be different from the default log level in Context.
* Hasn't update the hooks & docs.
* Hasn't cleared some unnecessary methods in FineGrainLogContext

Commit Message: fine-grain logger: add verbosity default log level support in fine-grain logger
Additional Description:
Risk Level: low
Testing: unit tests
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


test/common/http/header_map_impl_test.cc
```diff
@@ -424,8 +424,9 @@ TEST_P(HeaderMapImplTest, AllInlineHeaders) {
     INLINE_REQ_RESP_STRING_HEADERS(TEST_INLINE_STRING_HEADER_FUNCS)
   }
   {
-      // No request trailer O(1) headers.
-  } {
+    // No request trailer O(1) headers.
```
botengyao, line 427: This is an auto format fix.

test/common/common/log_verbosity_update_test.cc
```diff
@@ -232,7 +274,23 @@ TEST(FineGrainLog, inOrderUpdate) {
...
+  getFineGrainLogContext().initFineGrainLogger(file_1, flogger);
+
+  const std::pair<absl::string_view, int> update_1 =
+      std::make_pair("f??", static_cast<int>(spdlog::level::warn));
+  const std::pair<absl::string_view, int> update_2 =
+      std::make_pair("ff?", static_cast<int>(spdlog::level::info));
+  getFineGrainLogContext().updateVerbositySetting({update_1, update_2});
+
+  SpdLoggerSharedPtr p = getFineGrainLogContext().getFineGrainLogEntry(file_1);
+  EXPECT_NE(p, nullptr);
```
KBaichoo, line 289: this (and likely some other places in this) should be an `ASSERT_NE` otherwise we'll go and execute the next line and segfault. See: http://google.github.io/googletest/primer.html#assertions

test/common/common/log_verbosity_update_test.cc
```diff
@@ -232,7 +274,23 @@ TEST(FineGrainLog, inOrderUpdate) {
   EXPECT_EQ(p->level(), spdlog::level::warn);
 }
 
-TEST(FineGrainLog, invalidLevelUpdate) {
+TEST(FineGrainLog, earlyExitGlobUpdate) {
```
KBaichoo, line 277: can you clarify what this is testing?

KBaichoo (commented):
thanks for working on this!

source/common/common/fine_grain_logger.h
```diff
@@ -44,6 +44,11 @@ class FineGrainLogContext {
   SpdLoggerSharedPtr getFineGrainLogEntry(absl::string_view key)
       ABSL_LOCKS_EXCLUDED(fine_grain_log_lock_);
 
+  /**
+   * Gets the default verbosity log level.
+   */
+  spdlog::level::level_enum getVerbosityDefaultLevel() ABSL_LOCKS_EXCLUDED(fine_grain_log_lock_);
```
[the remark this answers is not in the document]
botengyao: For now it is but it could be not in future. The verbosity default level can be different with the default ENVOY log level.

source/common/common/fine_grain_logger.h
```diff
@@ -58,9 +63,9 @@ class FineGrainLogContext {
       ABSL_LOCKS_EXCLUDED(fine_grain_log_lock_);
 
   /**
-   * Sets the default logger level and format when updating context. It should only be used in
-   * Context, otherwise the fine_grain_default_level will possibly be inconsistent with the actual
-   * logger level.
+   * Sets the default verbosity log level and format when updating context. It is used in Context
+   * during initialization or the enable process. Basically, it will reset FineGrainLogContext and
+   * clear any previous verbosity updates.
```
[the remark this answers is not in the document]
botengyao: Good point. My initial thinking is that this method will reset `FineGrainLogContext`, but it could lose some very beginning verbosity update info. Changed it back to keep the verbosity info.

source/common/common/fine_grain_logger.h
```diff
@@ -94,17 +99,34 @@ class FineGrainLogContext {
...
    *
-   * Files which do not match any pattern use the value of default log level from Context.
+   * Files which do not match any pattern use the default verbosity log level.
    */
   void updateVerbositySetting(const std::vector<std::pair<absl::string_view, int>>& updates)
       ABSL_LOCKS_EXCLUDED(fine_grain_log_lock_);
 
+  /**
+   * Updates all the loggers based on the new default verbosity log level.
+   * The priority of verbosity_update_info_ is higher than verbosity_default_level_. It will change
```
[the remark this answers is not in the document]
botengyao: Changed the comments. We don't compare the value in the verbosity info vector with the default verbosity log level. The default verbosity log level is overridable whenever we find a match in the update info vector.

source/common/common/fine_grain_logger.h
```diff
@@ -146,7 +168,13 @@ class FineGrainLogContext {
...
    * Vector that stores <update, level> pairs, key can be the file basename or glob expressions.
    */
-  std::vector<VerbosityLogUpdateInfo> log_update_info_ ABSL_GUARDED_BY(fine_grain_log_lock_);
+  std::vector<VerbosityLogUpdateInfo> verbosity_update_info_ ABSL_GUARDED_BY(fine_grain_log_lock_);
+
+  /**
+   * Default verbosity log level. It could be different from fine_grain_default_level_ in Context.
+   */
+  spdlog::level::level_enum
+      verbosity_default_level_ ABSL_GUARDED_BY(fine_grain_log_lock_) = spdlog::level::info;
```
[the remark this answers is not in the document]
botengyao: Yes. During the initialization, it will be changed to the default `ENVOY_LOG` log level in the `Context` for now.

test/common/common/log_verbosity_update_test.cc
botengyao: Thank you! Done!

test/common/common/log_verbosity_update_test.cc
botengyao: Added comments. Done!

source/common/common/fine_grain_logger.cc
```diff
@@ -73,10 +71,9 @@ void FineGrainLogContext::setDefaultFineGrainLogLevelFormat(spdlog::level::level
     ABSL_LOCKS_EXCLUDED(fine_grain_log_lock_) {
   absl::WriterMutexLock wl(&fine_grain_log_lock_);
   verbosity_default_level_ = level;
-  verbosity_update_info_.clear();
-  for (const auto& it : *fine_grain_log_map_) {
-    it.second->set_level(level);
-    it.second->set_pattern(format);
+  for (auto& [key, logger] : *fine_grain_log_map_) {
+    logger->set_level(getLogLevel(key));
+    logger->set_pattern(format);
```
KBaichoo, line 76: if verbosity_update_info is empty, we don't need to query any of these levels, and can simply set them.

source/common/common/fine_grain_logger.cc
```diff
@@ -162,10 +161,22 @@ spdlog::logger* FineGrainLogContext::createLogger(const std::string& key)
...
 
+void FineGrainLogContext::updateVerbosityDefaultLevel(level_enum level) {
+  absl::WriterMutexLock ul(&fine_grain_log_lock_);
+  if (level == verbosity_default_level_) {
+    return;
+  }
+
+  verbosity_default_level_ = level;
+  for (auto& [key, logger] : *fine_grain_log_map_) {
+    logger->set_level(getLogLevel(key));
```
KBaichoo, line 172: likewise, if verbosity_update_info is empty, we don't need to query any of these levels, and can simply set them. There's an existing `setAllFineGrainLoggers` that does this

source/common/common/fine_grain_logger.h
```diff
@@ -127,15 +146,15 @@ class FineGrainLogContext {
...
    * Returns the current log level of `file`. Default log level is used if there is no
-   * match in log_update_info_.
+   * match in verbosity_update_info_.
    */
   spdlog::level::level_enum getLogLevel(absl::string_view file) const
       ABSL_EXCLUSIVE_LOCKS_REQUIRED(fine_grain_log_lock_);
 
   /**
    * Lock for the following global map and update vector (not for the corresponding loggers).
    */
```
KBaichoo, line 156: seems like this needs to be updated since the fine_grain_log_lock also guards the verbosity_default_level.

source/common/common/fine_grain_logger.cc
```diff
@@ -64,17 +69,11 @@ bool FineGrainLogContext::setFineGrainLogger(absl::string_view key, level_enum l
...
-    return;
-  }
-  absl::ReaderMutexLock l(&fine_grain_log_lock_);
-  for (const auto& it : *fine_grain_log_map_) {
-    if (it.second->level() == Logger::Context::getFineGrainDefaultLevel()) {
-      // if logger is default level now
-      it.second->set_level(level);
-    }
-    it.second->set_pattern(format);
+  absl::WriterMutexLock wl(&fine_grain_log_lock_);
```
KBaichoo, line 72: this is only a writer lock because we modify the verbosity_default_level, perhaps we can transition it to a reader lock when doing the loop to avoid blocking logging from occuring during this call

source/common/common/fine_grain_logger.cc
```diff
@@ -162,10 +161,22 @@ spdlog::logger* FineGrainLogContext::createLogger(const std::string& key)
   return new_logger.get();
 }
 
+void FineGrainLogContext::updateVerbosityDefaultLevel(level_enum level) {
+  absl::WriterMutexLock ul(&fine_grain_log_lock_);
```
KBaichoo, line 165: this is only a writer lock because we modify the verbosity_default_level, perhaps we can transition it to a reader lock when doing the loop to avoid blocking logging from occuring during this call

KBaichoo (commented):
looks good otherwise, thanks!

source/common/common/fine_grain_logger.h
botengyao: Good catch, done!

source/common/common/fine_grain_logger.cc
botengyao: Good point! Done.

source/common/common/fine_grain_logger.cc
botengyao: The current admin server interface is using the `setAllFineGrainLoggers()`, which is stateless and may make `FinGrainLogContext` inconsistent. It needs to be migrated to the verbosity default level update. We can refactor and generalize the update part if the above doesn't sound good, eg.
```
void FineGrainLogContext::updateAllFineGrainLoggers()
    ABSL_LOCKS_EXCLUDED(fine_grain_log_lock_) {
  absl::ReaderMutexLock l(&fine_grain_log_lock_);
  if (verbosity_update_info_.empty()) {
    for (const auto& it : *fine_grain_log_map_) {
      it.second->set_level(verbosity_default_level_);
    }
  } else {
    for (const auto& [key, logger] : *fine_grain_log_map_) {
      logger->set_level(getLogLevel(key));
    }
  }
}
```

`setDefaultFineGrainLogLevelFormat` includes format setting, which will only be called during initialization, and we can leave as it is, I think.

botengyao: Well, as a post request in the admin hook, it makes sense to be stateless as well... We can discuss more for the admin hook. These are now enough for the google3 hook.

KBaichoo: Let's go with the second approach as it's slightly better as we'd then have more consistent behavior with the other system. As the admin doesn't yet have a hook to set the verbosity_update_info it'd fall back to behaving as is now -- this will just help clients who can use verbosity_update_info to not run into inconsistencies.

botengyao: Done.

KBaichoo (approved):
lgtm thanks!

KBaichoo:
/assign @mattklein123 

for final pass

PR #21730, opened 2022-06-16
Trailer support fuzz

KBaichoo:
Commit Message: H1 trailer fuzz support
Additional Description: Coauthored with Andre Vehreschild <vehre@x41-dsec.de>
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/memory/utils.cc
+++ b/source/common/memory/utils.cc
@@ -38,9 +38,8 @@ void Utils::tryShrinkHeap() {
   auto total_physical_bytes = Stats::totalPhysicalBytes();
   auto allocated_size_by_app = Stats::totalCurrentlyAllocated();
 
-  ASSERT(total_physical_bytes >= allocated_size_by_app);
-
-  if ((total_physical_bytes - allocated_size_by_app) >= MAX_UNFREED_MEMORY_BYTE) {
+  if (total_physical_bytes >= allocated_size_by_app &&
+      (total_physical_bytes - allocated_size_by_app) >= MAX_UNFREED_MEMORY_BYTE) {
     Utils::releaseFreeMemory();
   }
 #endif
--- a/test/common/http/codec_impl_corpus/clusterfuzz-testcase-minimized-codec_impl_fuzz_test-5074389318107136-h1-with-trailers
+++ b/test/common/http/codec_impl_corpus/clusterfuzz-testcase-minimized-codec_impl_fuzz_test-5074389318107136-h1-with-trailers
@@ -1,0 +1,235 @@
+h1_settings {
+  server {
+    enable_trailers: true
+  }
+  client {
+    enable_trailers: true
+  }
+}
+actions {
+  new_stream {
+    request_headers {
+      headers {
+        key: ":method"
+        value: "GET"
+      }
+      headers {
+        key: ":path"
+        value: "/"
+      }
+      headers {
+        key: ":scheme"
+        value: "http"
+      }
+      headers {
+        key: ":authority"
+        value: "foo.com"
+      }
+      headers {
+        key: "blah"
+        value: "nosniff"
+      }
+      headers {
+        key: "cookie"
+        value: "foo=bar"
+      }
+      headers {
+        key: "cookie"
+        value: "foo2=bar2"
+      }
+    }
+  }
+}
+actions {
+  stream_action {
+    request {
+      data: 3000000
+    }
+  }
+}
+actions {
+  stream_action {
+    request {
+      data: 54
+    }
+  }
+}
+actions {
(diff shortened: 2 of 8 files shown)


source/common/memory/utils.cc
```diff
@@ -38,9 +38,8 @@ void Utils::tryShrinkHeap() {
   auto total_physical_bytes = Stats::totalPhysicalBytes();
   auto allocated_size_by_app = Stats::totalCurrentlyAllocated();
 
-  ASSERT(total_physical_bytes >= allocated_size_by_app);
-
-  if ((total_physical_bytes - allocated_size_by_app) >= MAX_UNFREED_MEMORY_BYTE) {
+  if (total_physical_bytes >= allocated_size_by_app &&
+      (total_physical_bytes - allocated_size_by_app) >= MAX_UNFREED_MEMORY_BYTE) {
```
KBaichoo, line 42: will revert this.

test/common/http/codec_impl_corpus/clusterfuzz-testcase-minimized-codec_impl_fuzz_test-5074389318107136-h1-with-trailers
```diff
@@ -0,0 +1,235 @@
...
+  stream_action {
+    response {
+      headers {
+        headers {
+          key: ":status"
+          value: "200"
+        }
+        headers {
+          key: "content-length"
+          value: "5"
```
KBaichoo, line 78: note this doesn't yet quite work as we don't actually encode the trailers due to content-length for H1, will fix in follow up. See the test/common/http/codec_impl_corpus/clusterfuzz-testcase-minimized-codec_impl_fuzz_test-5074389318107136-h1-with-trailers-te below which is just with this header removed

test/common/http/codec_impl_fuzz_test.cc
```diff
@@ -297,14 +300,41 @@ class HttpStream : public LinkedObject<HttpStream> {
...
-          state.request_encoder_->encodeTrailers(
-              fromSanitizedHeaders<TestRequestTrailerMapImpl>(directional_action.trailers()));
+          // For HTTP1 we need to see if the given direction supports encoding
+          // trailers.
+          if (response && http1_allow_server_trailers_) {
+            // TODO(kbaichoo): guard to see if can encode trailers by seeing if
+            // content length missing from hdrs so we support trailers.
+            state.response_encoder_->encodeTrailers(
+                fromSanitizedHeaders<TestResponseTrailerMapImpl>(directional_action.trailers()));
+            encoded_trailers = true;
```
vehre-x41, line 323: Just for my understanding: We are encoding trailers on the server side here, but settings this bool to close the client side. Why does this make sense? Wouldn't it be "more natural" to close the server side (remote) here?

test/common/http/codec_impl_fuzz_test.cc
```diff
@@ -398,7 +428,11 @@ class HttpStream : public LinkedObject<HttpStream> {
         ENVOY_LOG_MISC(debug, "Setting dispatching action  on {} in state {} {}", stream_index_,
                        static_cast<int>(request_.stream_state_),
                        static_cast<int>(response_.stream_state_));
-        auto request_action = stream_action.dispatching_action().directional_action_selector_case();
+        // auto request_action =
+        // stream_action.dispatching_action().directional_action_selector_case();
+        //  I think we should have it be on the reuqest action taken, not on
+        //  what the dispatching_action e.g. what to do in response is...
+        auto request_action = stream_action.request().directional_action_selector_case();
```
vehre-x41, line 435: I disaggree with using the `request()` here. See comment in [codec_impl_fuzz.proto](https://github.com/envoyproxy/envoy/blob/44634cc608f11bed6852c67f40902cea0d5c9554/test/common/http/codec_impl_fuzz.proto#L43-L47). The "directional action" is meant to be some kind of hybrid and with your change, it is not taken into account anymore.

vehre-x41 (commented):
I like your changes. Let me add some feedback/ideas...

vehre-x41:
The OSS fuzz issue has been closed as fixed. So this is not necessary anymore?

adisuissa:
> The OSS fuzz issue has been closed as fixed. So this is not necessary anymore?

This is probably due to #21751, so the issue is still there,

KBaichoo:
I'll continue working on this; with the example corpuses given here (and likely those for the other issues) we can reproduce the fuzz issue

Issue #17321, opened 2021-07-14
wasm: make vm_config.runtime default to V8 (or Wasmtime/WAVM/WAMR).


mathetake:
Given that almost all Wasm plugin users would put `runtime: envoy.wasm.runtime.v8`, it would make sense to me that make `vm_config.runtime` optional and default to V8 or any runtime which Envoy is compiled with. @PiotrSikora any thoughts?

PiotrSikora:
Yes, we should definitely make `runtime` field optional, and if there is only one Wasm runtime built-in (which is currently the only option), then it should be used without any extra configuration.

It gets a little tricky because of NullVM, but we can always try to load NullVM plugin first before falling back to loading from file/remote.

vikaschoudhary16:
this is done and can be closed.

PR #21749, opened 2022-06-16
wasm: make vm_config.runtime optional

bryanwux:
Commit Message: This PR makes `vm_config.runtime` optional. Since envoy only supports only a single wasm runtime, the default value is set to the wasm runtime compiled-in.
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #17321
[Optional Deprecated:]

Signed-off-by: Jiayu Wu [jiayu1.wu@intel.com](jiayu1.wu@intel.com)


source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -74,9 +74,9 @@ bool EnvoyWasmVmIntegration::getNullVmFunction(std::string_view function_name, b
 
 WasmVmPtr createWasmVm(absl::string_view runtime) {
   if (runtime.empty()) {
-    ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::wasm), warn,
-                        "Failed to create Wasm VM with unspecified runtime");
-    return nullptr;
+    ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::wasm), info,
+                        "Wasm VM runtime is unspecified, set to envoy.wasm.runtime.v8");
+    runtime = "envoy.wasm.runtime.v8";
```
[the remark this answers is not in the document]
bryanwux: Fixed.

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -50,7 +49,7 @@ message VmConfig {
   // [#comment: TODO: add ref for details.]
   string vm_id = 1;
 
-  // The Wasm runtime type.
+  // The Wasm runtime type, default is "envoy.wasm.runtime.v8".
```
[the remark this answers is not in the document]
bryanwux: Fixed.

phlax (commented):
this looks like a really helpful simplification of the config

we should update any config examples to remove the defaults also

bryanwux:
Not sure why windows test fails.

phlax:
https://dev.azure.com/cncf/envoy/_build/results?buildId=111278&view=logs&j=b840a642-5ff3-5357-2e4b-e06e40b0cffd&t=67965174-5100-5631-9dc0-68b9f0aacb53&l=1324

source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -72,11 +72,27 @@ bool EnvoyWasmVmIntegration::getNullVmFunction(std::string_view function_name, b
...
 }
 
+absl::string_view getWasmEngine() {
+  absl::string_view wasm_ext_runtime;
+  for (const auto& ext : Envoy::Registry::FactoryCategoryRegistry::registeredFactories()) {
+    if (ext.first == "envoy.wasm.runtime") {
+      for (auto engine_name : ext.second->registeredNames()) {
+        if (engine_name != "envoy.wasm.runtime.null") {
+          wasm_ext_runtime = engine_name;
+        }
```
PiotrSikora, line 82: To future-proof this code, you might want to add `break` after finding the first non-null Wasm engine.

source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -72,11 +72,27 @@ bool EnvoyWasmVmIntegration::getNullVmFunction(std::string_view function_name, b
...
+
 WasmVmPtr createWasmVm(absl::string_view runtime) {
+  // Set wasm runtime to build-in wasm engine if it is not specified
   if (runtime.empty()) {
-    ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::wasm), warn,
-                        "Failed to create Wasm VM with unspecified runtime");
-    return nullptr;
+    ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::wasm), info,
+                        "Wasm VM runtime is unspecified, set to envoy built-in wasm runtime: {}",
+                        getWasmEngine());
```
PiotrSikora, line 94: We're changing this field to optional with a clearly defined default value, so I'd completely remove this log entry.

test/extensions/bootstrap/wasm/config_test.cc
```diff
@@ -104,12 +104,13 @@ TEST_P(WasmFactoryTest, MissingImport) {
                             "Unable to create Wasm service test");
 }
 
-TEST_P(WasmFactoryTest, UnspecifiedRuntime) {
+TEST_P(WasmFactoryTest, UnSpecifiedRuntime) {
```
PiotrSikora, line 107: Unnecessary/wrong change (`s` -> `S`).

source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -72,11 +72,27 @@ bool EnvoyWasmVmIntegration::getNullVmFunction(std::string_view function_name, b
   return false;
 }
 
+absl::string_view getWasmEngine() {
```
PiotrSikora, line 75: Nit: `getDefaultWasmEngineName` or similar might be more appropriate, since we're not getting an instance here.

source/extensions/common/wasm/wasm_vm.h
```diff
@@ -41,6 +41,7 @@ using WasmVmPtr = std::unique_ptr<proxy_wasm::WasmVm>;
 // Create a new low-level Wasm VM using runtime of the given type (e.g. "envoy.wasm.runtime.wavm").
 WasmVmPtr createWasmVm(absl::string_view runtime);
 
+absl::string_view getWasmEngine();
 } // namespace Wasm
```
PiotrSikora, line 45: Nit: missing docs and empty line.

source/extensions/common/wasm/wasm_vm.cc
bryanwux: Done.

source/extensions/common/wasm/wasm_vm.cc
bryanwux: Log entry is removed.

test/extensions/bootstrap/wasm/config_test.cc
bryanwux: Done.

source/extensions/common/wasm/wasm_vm.h
bryanwux: Done.

source/extensions/common/wasm/wasm_vm.cc
PiotrSikora: Actually, you should probably `return engine_name;` here, and `return "";` at the end of the function, otherwise you don't cover builds with Wasm disabled.

test/extensions/common/wasm/wasm_vm_test.cc
```diff
@@ -49,7 +49,11 @@ class BaseVmTest : public testing::Test {
   Stats::ScopeSharedPtr scope_;
 };
 
-TEST_F(BaseVmTest, NoRuntime) { EXPECT_EQ(createWasmVm(""), nullptr); }
+TEST_F(BaseVmTest, UnSpecifiedRuntime) {
+  auto wasm_vm = createWasmVm("");
+  EXPECT_TRUE(wasm_vm.get() != nullptr);
```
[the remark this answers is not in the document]
PiotrSikora: Guarding this test on Windows is not a good idea, since people can use `--define wasm=disabled` in their builds, and it will be eventually enabled on Windows, in which case this test would result in false positive failures.

Maybe guard it on `getDefaultWasmEngineName()`?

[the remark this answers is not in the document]
bryanwux: True. Now it is guarded on `getDefaultWasmEngineName()`.

test/extensions/common/wasm/wasm_vm_test.cc
```diff
@@ -49,7 +49,16 @@ class BaseVmTest : public testing::Test {
   Stats::ScopeSharedPtr scope_;
 };
 
-TEST_F(BaseVmTest, NoRuntime) { EXPECT_EQ(createWasmVm(""), nullptr); }
+TEST_F(BaseVmTest, UnSpecifiedRuntime) {
```
PiotrSikora, line 52: Nit: `UnSpecifiedRuntime` -> `UnspecifiedRuntime` (`S` -> `s`).

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -50,7 +49,7 @@ message VmConfig {
   // [#comment: TODO: add ref for details.]
   string vm_id = 1;
 
-  // The Wasm runtime type.
+  // The Wasm runtime type, default to wasm engine used at envoy build-time.
```
PiotrSikora, line 52: Nit: `default` -> `defaults` and `envoy` -> `Envoy`.

PiotrSikora (approved):
LGTM, small nits.

test/extensions/common/wasm/wasm_vm_test.cc
bryanwux: Fixed.

api/envoy/extensions/wasm/v3/wasm.proto
bryanwux: Fixed.

PiotrSikora (approved):
Thanks!

bryanwux:
ping @lizan

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -50,7 +49,7 @@ message VmConfig {
   // [#comment: TODO: add ref for details.]
   string vm_id = 1;
 
-  // The Wasm runtime type.
+  // The Wasm runtime type, defaults to wasm engine used at Envoy build-time.
```
lizan, line 52: Technically it is possible to build multiple engines in Envoy (e.g. wasmtime and V8 doesn't conflict), what will be default then?

PiotrSikora: Can you define multiple `wasm` values? Defining `wasm` twice results in only the latter value taking effect:
```
$ bazel build --config=libc++ --define wasm=v8 --define wasm=wasmtime //source/exe:envoy-static
```
```
$ bazel-bin/source/exe/envoy-static 2>&1 | grep wasm.runtime
[...]   envoy.wasm.runtime: envoy.wasm.runtime.null, envoy.wasm.runtime.wasmtime
```

lizan: Current bazel configuration doesn't allow this, but it doesn't mean any future build or a downstream build cannot build multiple wasm engines in same binaries as they are registered as extensions. So we still need to define a behavior for that case.

PiotrSikora: Right, but the documentation can be updated once this capability becomes available. I find it a bit misleading to say that "the first configured Wasm engine will be used" when we support only one.

wrowe:
Ping @lizan it seems all open questions have been satisfied, can you bring this full circle?

api/envoy/extensions/wasm/v3/wasm.proto
lizan: My point here is the Envoy API isn't fully bound to Envoy implementation, there's other implementations and there are downstream builds of Envoy as well. There is no deterministic way to find "the first configured Wasm engine will be used" when we build more than one engines. If we want to make a default engine selection here, either hardcode one or hardcode a priority list from ones that is supported today, so there's no ambiguity?

PiotrSikora: @bryanwux could you make the change requested by @lizan?

bryanwux: I am not sure if I understand correctly. What @lizan says about "hardcode one" here means we still need to define the wasm runtime in `vm_config.runtime` since bazel may support multiple wasm engines in one envoy build. So we should not make `vm_config.runtime` here?

PiotrSikora: No, he means that you could hardcode a list and use first one that's available, so that the order is deterministic, e.g.:

- Check if `envoy.wasm.runtime.v8` is available. If it is, then use it.
- Otherwise, check if `envoy.wasm.runtime.wasmtime` is available. If it is, then use it.
- Otherwise, check if `envoy.wasm.runtime.wamr` is available. If it is, then use it.
- Otherwise, check if `envoy.wasm.runtime.wavm` is available. If it is, then use it.

This doesn't make a difference (vs querying which engine was used at the build time), since there is only one available in Envoy anyway (at least, right now).

source/extensions/common/wasm/wasm_vm.h
PiotrSikora: Nit: still missing an empty line before closing brackets.

source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -72,11 +72,28 @@ bool EnvoyWasmVmIntegration::getNullVmFunction(std::string_view function_name, b
...
+  } else if (isWasmEngineAvailable("envoy.wasm.runtime.wamr")) {
+    return "envoy.wasm.runtime.wamr";
+  } else if (isWasmEngineAvailable("envoy.wasm.runtime.wavm")) {
+    return "envoy.wasm.runtime.wavm";
+  }
+  return "";
+}
+
 WasmVmPtr createWasmVm(absl::string_view runtime) {
+  // Set wasm runtime to build-in wasm engine if it is not specified
```
PiotrSikora, line 94: Nit: `build-in` -> `built-in` and `wasm` -> `Wasm`.

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -50,7 +49,8 @@ message VmConfig {
   // [#comment: TODO: add ref for details.]
   string vm_id = 1;
 
-  // The Wasm runtime type.
+  // The Wasm runtime type, defaults to the first available wasm engine used at Envoy build-time.
```
PiotrSikora, line 52: Nit: `wasm` -> `Wasm`.

PiotrSikora (approved):
LGTM, with a few small nits in comments.

lizan (dismissed):

source/extensions/common/wasm/wasm_vm.cc
bryanwux: Fixed.

PiotrSikora (approved):
Thanks!

bryanwux:
The problem is envoy is built with only one wasm runtime which is v8 now, so these lines will never get hit.

bryanwux:
The problem is CI is only built with one wasm runtime which is v8 now, so these new lines will never get hit.

jmarantz:
If the code can't be reached can you just compile it out until it becomes reachable?

mattklein123:
Sorry what is the resolution on @jmarantz comment? Can we just remove this code if it's not used? Otherwise if it's used some times can you refactor it in a way that it can be tested?

/wait

bryanwux:
> Sorry what is the resolution on @jmarantz comment? Can we just remove this code if it's not used? Otherwise if it's used some times can you refactor it in a way that it can be tested?
> 
> /wait

Hi @mattklein123 , the previous failure for code coverage was caused by current CI only supports v8 for default wasm runtime. If I want to make it testable, the only way I can do is to add support for testing other wasm runtime in CI script? On the other hand, the code is needed because users may want to specify other wasm runtime when building envoy. My solution for now is to reduce the coverage limit for `source/extensions/common/wasm`. WDYT?

mattklein123:
If the one that we actually build in CI uses v8 perhaps that should just be the default and then it can fail if someone wants something different? It's an advanced operation for them to compile something different anyway? Or is the idea that you want to be able to have a single config and test builds with different runtimes?

source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -72,11 +72,28 @@ bool EnvoyWasmVmIntegration::getNullVmFunction(std::string_view function_name, b
   return false;
 }
 
+bool isWasmEngineAvailable(absl::string_view runtime) {
+  auto runtime_factory = Registry::FactoryRegistry<WasmRuntimeFactory>::getFactory(runtime);
+  return runtime_factory != nullptr;
+}
+
+absl::string_view getFirstAvailableWasmEngineName() {
+  if (isWasmEngineAvailable("envoy.wasm.runtime.v8")) {
```
lizan, line 81: Or you can do a loop:

```
static const std::vector<absl::string_view> wasm_engines{"envoy.wasm.runtime.v8", ...};
for (const auto wasm_engine : wasm_engines) {
  if (isWasmEngineAvailable(wasm_engine)) {
    return wasm_engine;
  }
}
```

source/extensions/common/wasm/wasm_vm.cc
```diff
@@ -78,7 +78,9 @@ bool isWasmEngineAvailable(absl::string_view runtime) {
 }
 
 absl::string_view getFirstAvailableWasmEngineName() {
-  static const std::vector<absl::string_view> wasm_engines{"envoy.wasm.runtime.v8", "envoy.wasm.runtime.wasmtime", "envoy.wasm.runtime.wamr", "envoy.wasm.runtime.wavm"};
+  static const std::vector<absl::string_view> wasm_engines{
```
jmarantz, line 81: constexpr. statically constructed non-pods are not allowed, but if you can declare it constexpr it's OK.

or you could make it non-static.

api/envoy/extensions/wasm/v3/wasm.proto
```diff
@@ -79,7 +79,7 @@ message VmConfig {
   // This runtime is not enabled in the official build.
   //
   // [#extension-category: envoy.wasm.runtime]
-  string runtime = 2 [(validate.rules).string = {min_len: 1}];
+  string runtime = 2;
```
jmarantz, line 82: should this be an enum?

test/extensions/common/wasm/wasm_vm_test.cc
```diff
@@ -49,7 +49,16 @@ class BaseVmTest : public testing::Test {
...
 
-TEST_F(BaseVmTest, NoRuntime) { EXPECT_EQ(createWasmVm(""), nullptr); }
+TEST_F(BaseVmTest, UnspecifiedRuntime) {
+  auto wasm_vm = createWasmVm("");
+  // Envoy is built with "--define wasm=disabled", so no Wasm engine is available
+  if (getFirstAvailableWasmEngineName().empty()) {
+    EXPECT_TRUE(wasm_vm.get() == nullptr);
+  } else {
+    EXPECT_TRUE(wasm_vm.get() != nullptr);
+    EXPECT_TRUE(std::string(getFirstAvailableWasmEngineName()).find(wasm_vm->getEngineName()));
```
jmarantz, line 59: find() does not return true/false.

Suggest using `EXPECT_THAT(std::string(getFirstAvailableWasmEngineName()), HasSubstr(wasm_vm->getEngineName()));` if that is what you mean.

test/extensions/common/wasm/wasm_vm_test.cc
```diff
@@ -49,7 +49,16 @@ class BaseVmTest : public testing::Test {
...
 };
 
-TEST_F(BaseVmTest, NoRuntime) { EXPECT_EQ(createWasmVm(""), nullptr); }
+TEST_F(BaseVmTest, UnspecifiedRuntime) {
+  auto wasm_vm = createWasmVm("");
+  // Envoy is built with "--define wasm=disabled", so no Wasm engine is available
+  if (getFirstAvailableWasmEngineName().empty()) {
+    EXPECT_TRUE(wasm_vm.get() == nullptr);
+  } else {
+    EXPECT_TRUE(wasm_vm.get() != nullptr);
```
jmarantz, line 58: ASSERT_TRUE -- that way if this line fails the test fails, but doesn't crash on the next line.

test/extensions/common/wasm/wasm_vm_test.cc
bryanwux: Thanks for pointing out, will fix.

source/extensions/common/wasm/wasm_vm.cc
lizan: static constructed non-prod are allowed in methods as the initialization order isn't an issue. this cannot be constexpr.

jmarantz: Sorry; I meant this *can* be constexpr.

jmarantz: Hm. I thought you could make constexpr vector but could not find any in the envoy codebase. You don't need a vector though; you can use `constexpr std::array` or simply `constexpr absl::string_view[] ...`.

This is marginally better than `const` IMO as it can be constructed at compile-time.

jmarantz: Also I see why Lizan says non-pod static const doesn't have an init-order issue, but my reading of https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables is that they are still recommended against, and they are not needed here.

api/envoy/extensions/wasm/v3/wasm.proto
bryanwux: I think we can make this as an enum. But the idea of this PR is to make wasm runtime optional because currently envoy only supports one wasm engine (which is the engine used at envoy build-time) at a time. So making this enum or not makes not much difference? WDYT?

bryanwux: I think we can make this as an enum. But the idea of this PR is to make wasm runtime optional because currently envoy only supports one wasm engine (which is the engine used at envoy build-time) at a time. So making this enum or not makes not much difference? WDYT?

lizan: No we shouldn't change existing API and this is registered as extension so should be as is.

source/extensions/common/wasm/wasm_vm.cc
bryanwux: Do we have a consensus on this? Personally I think since the value is known at compile time, static const won't be an issue here.

jmarantz: It violates the style guide, although in this case I agree it's safe.

But it forces the readers to study and convince themselves it's safe. So the style guide is partially to help us write code that's more obviously correct, and thus easier to read.

So I vote to use a C array or std::array of absl::string_view which you can construct as a constexpr which is obviously safe. I believe it will be faster and smaller as well.

bryanwux: Yes, I agree using a constexpr will be faster in run time and follows the style guide as well. Not really understand why lizan said can not use constexpr here.

jmarantz: You can't make a constexpr std::vector because it heap-allocates its data. But you can do that with a std::array or C array since the data is inline and can be organized by the compiler and placed into the object file directly.

bryanwux: oh I see what he was saying now. I thought he just opposed to `constexpr` in general here, and that makes me confused.

source/extensions/common/wasm/wasm_vm.h
```diff
@@ -41,6 +41,12 @@ using WasmVmPtr = std::unique_ptr<proxy_wasm::WasmVm>;
 // Create a new low-level Wasm VM using runtime of the given type (e.g. "envoy.wasm.runtime.wavm").
 WasmVmPtr createWasmVm(absl::string_view runtime);
 
+// Check if the provided Wasm Engine is compiled with Envoy
```
jmarantz, line 44: s/Check/@return true/

and use 
```
/**
 * ...
 */
```
for doxygen format.

source/extensions/common/wasm/wasm_vm.h
```diff
@@ -41,6 +41,12 @@ using WasmVmPtr = std::unique_ptr<proxy_wasm::WasmVm>;
...
+bool isWasmEngineAvailable(absl::string_view runtime);
+
+// Get the name of the first available Wasm Engine compiled with Envoy
```
jmarantz, line 47: s/Get/@return/ and use doxygen format.

test/extensions/common/wasm/wasm_vm_test.cc
```diff
@@ -49,7 +49,17 @@ class BaseVmTest : public testing::Test {
   Stats::ScopeSharedPtr scope_;
 };
 
-TEST_F(BaseVmTest, NoRuntime) { EXPECT_EQ(createWasmVm(""), nullptr); }
+TEST_F(BaseVmTest, UnspecifiedRuntime) {
+  auto wasm_vm = createWasmVm("");
+  // Envoy is built with "--define wasm=disabled", so no Wasm engine is available
+  if (getFirstAvailableWasmEngineName().empty()) {
```
jmarantz, line 55: nit: I'd capture getFirstAvailableWasmEngineName() in a string_view temp rather than calling it twice.

But I'm not hung up over it.

jmarantz (approved):
Looks great generally with minor nits.

mattklein123:
LGTM with remaining nits.

/wait

source/extensions/common/wasm/wasm_vm.h
bryanwux: Done

bryanwux:
nits done. Kindly ping @mattklein123

mattklein123 (approved):

PR #21751, opened 2022-06-16
Split codec_impl_fuzz_test targets out by protocol

KBaichoo:
This allows the fuzzer to focus on the individual protocols better and allows us to categorize failures by protocol.

Commit Message: Split codec_impl_fuzz_test targets out by protocol.
Additional Description:
Risk Level: low
Testing: built and ran target
Docs Changes: na
Release Notes: na
Platform Specific Features: na
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/common/http/codec_impl_fuzz_test.cc
+++ b/test/common/http/codec_impl_fuzz_test.cc
@@ -786,9 +786,16 @@ DEFINE_PROTO_FUZZER(const test::common::http::CodecImplFuzzTestCase& input) {
   try {
     // Validate input early.
     TestUtility::validate(input);
+#ifdef FUZZ_CODEC_http1
     codecFuzz(input, HttpVersion::Http1);
+#endif
+#ifdef FUZZ_CODEC_nghttp2
     codecFuzz(input, HttpVersion::Http2Nghttp2);
+#endif
+#ifdef FUZZ_CODEC_wrapped_nghttp2
     codecFuzz(input, HttpVersion::Http2WrappedNghttp2);
+#endif
+#ifdef FUZZ_CODEC_oghttp2
     // Prevent oghttp2 from aborting the program.
     // If when disabling the FATAL log abort the fuzzer will create a test that reaches an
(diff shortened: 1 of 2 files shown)


adisuissa:
The downside of this approach is that now we have 4 different fuzzers that may benefit from running the same test across different instances (well... at least for the H/2 case).
I guess one can look at the stack trace and see the line number of the failing call to `codecFuzz()`, but this may change over-time.
What do you think about explicit functions that call the `codecFuzz()` functions, for example:
```
void codecFuzzHttp1() {
  codecFuzz(input, HttpVersion::Http1);
}
```

KBaichoo:
Split out build target by protocol instead and added wrapper functions for the individual codecs as suggested.

test/common/http/codec_impl_fuzz_test.cc
```diff
@@ -779,23 +779,47 @@ void codecFuzz(const test::common::http::CodecImplFuzzTestCase& input, HttpVersi
...
   try {
     // Validate input early.
     TestUtility::validate(input);
-    codecFuzz(input, HttpVersion::Http1);
-    codecFuzz(input, HttpVersion::Http2Nghttp2);
-    codecFuzz(input, HttpVersion::Http2WrappedNghttp2);
+#ifdef FUZZ_PROTOCOL_http1
+    codecFuzzHttp1(input);
+#endif
+#ifdef FUZZ_PROTOCOL_http2
```
adisuissa, line 812: nit: add a small comment explaining that we have different function names so they will appear explicitly in the stacktrace.

adisuissa (commented):
Small nit, but otherwise LGTM.

test/common/http/codec_impl_fuzz_test.cc
KBaichoo: done

adisuissa (approved):
LGTM, thanks!

PR #21752, opened 2022-06-17
thrift: more logging

JuniorHsu:

Commit Message:
We have way less logging than necessary. 
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/network/thrift_proxy/router/upstream_request.cc
+++ b/source/extensions/filters/network/thrift_proxy/router/upstream_request.cc
@@ -300,6 +300,7 @@ bool UpstreamRequest::onResetStream(ConnectionPool::PoolFailureReason reason) {
 
     // TODO(zuercher): distinguish between these cases where appropriate (particularly timeout)
     if (response_started_) {
+      ENVOY_LOG(warn, "reset downstream connection for a partial response");
       // Error occurred after a partial response, propagate the reset to the downstream.
       parent_.resetDownstreamConnection();
     } else {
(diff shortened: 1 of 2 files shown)


source/extensions/filters/network/thrift_proxy/router/upstream_request.cc
```diff
@@ -300,6 +300,7 @@ bool UpstreamRequest::onResetStream(ConnectionPool::PoolFailureReason reason) {
 
     // TODO(zuercher): distinguish between these cases where appropriate (particularly timeout)
     if (response_started_) {
+      ENVOY_LOG(warn, "reset downstream connection for a partial response");
```
fishcakez, line 303: `warn` level may be a bit high or does http use this level?

JuniorHsu: http would send local reply iirc. however, not much warn log in http, so happy to down level to debug

RyanTheOptimist:
I believe [zuercher](https://github.com/zuercher) is out for another week or two.

source/extensions/filters/network/thrift_proxy/router/upstream_request.cc
rgs1: warn might not be that bad, i'd be curious to know how much it happens in your prod world.

rgs1 (approved):
LGTM, thanks! I'll ask Matt to merge once all is green given Stephan is (apparently) OOO.

rgs1 (approved):
LGTM.

ggreenway (approved):

Issue #21726, opened 2022-06-15
File-System-Buffer multiple threads access the same resource


adisuissa:
_This is a tsan flaky bug in the File-System-Buffer filter, and was cleared with Envoy security team before posted here_

*Title*: File-System-Buffer multiple threads access the same resource

*Description*:
Our filter fuzzer detected a concurrency-related flaky issue with "envoy.filters.http.file_system_buffer".
From an initial look, it seems that 2 threads are trying to update a shared `cancel_in_flight_action_` resource in lines [396](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/file_system_buffer/filter.cc#L396) and [425](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/file_system_buffer/filter.cc#L425).

*Repro steps*:
Create the following file `test/extensions/filters/http/common/fuzz/filter_corpus/issue`:
```
config {
  name: "envoy.filters.http.file_system_buffer"
  typed_config {
    type_url: "type.googleapis.com/envoy.extensions.filters.http.file_system_buffer.v3.FileSystemBufferFilterConfig"
    value: "\n\005\n\001V\022\000\"\014\n\002*\000\022\002\010@\"\002\010@"
  }
}
upstream_data {
  http_body {
    data: "\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177\177"
    data: "envoy.grpc_json_transcodFer"
    data: "aaaaaaaaaaaaaaaaaaaaaaa"
  }
}
```

The above creates a filter configuration which is equivalent to:
```
Filter configuration: name: "envoy.filters.http.file_system_buffer"                                                                                                                                       typed_config {
  [type.googleapis.com/envoy.extensions.filters.http.file_system_buffer.v3.FileSystemBufferFilterConfig] {
    manager_config {
      id: "V"
      thread_pool {
      }
    }
    response {
      behavior {
[...]

PR #21757, opened 2022-06-17
file system buffer: Ensure concurrency is safe.

vehre-x41:
Commit Message: file system buffer: Ensure concurrency is safe.
Additional Description:
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a 
Fixes Issue: [#21726](https://github.com/envoyproxy/envoy/issues/21726)

### DIFF
--- a/test/extensions/filters/http/common/fuzz/uber_filter.cc
+++ b/test/extensions/filters/http/common/fuzz/uber_filter.cc
@@ -99,6 +99,8 @@ void UberFilterFuzzer::fuzz(
     // If we were successful, run any pending events on the main thread's dispatcher loop. These
     // might be, for example, pending DNS resolution callbacks. If they generate exceptions, we want
     // to explode and fail the test, hence we do this outside of the try-catch above.
+    // TODO(ravenblackx): this can potentially exit the loop without the filter having completed.
+    // We should detect stream completion rather than just running one round of dispatcher.
     main_dispatcher_->run(Event::DispatcherImpl::RunType::Block);
   }
   reset();
--- a/test/extensions/filters/http/common/fuzz/uber_filter.cc
+++ b/test/extensions/filters/http/common/fuzz/uber_filter.cc
@@ -73,13 +73,6 @@ void UberFilterFuzzer::fuzz(
         proto_config, factory_context_.messageValidationVisitor(), factory);
     // Clean-up config with filter-specific logic before it runs through validations.
     cleanFuzzedConfig(proto_config.name(), message.get());
-    if (with_main_event_loop_) {
-      // For filters that need main event loop.
-      ON_CALL(encoder_callbacks_, dispatcher())
-          .WillByDefault(testing::ReturnRef(*worker_thread_dispatcher_));
-      ON_CALL(decoder_callbacks_, dispatcher())
-          .WillByDefault(testing::ReturnRef(*worker_thread_dispatcher_));
-    }
     cb_ = factory.createFilterFactoryFromProto(*message, "stats", factory_context_);
     cb_(filter_callback_);
   } catch (const EnvoyException& e) {
--- a/test/extensions/filters/http/common/fuzz/uber_filter.cc
+++ b/test/extensions/filters/http/common/fuzz/uber_filter.cc
@@ -18,7 +18,7 @@ UberFilterFuzzer::UberFilterFuzzer()
     : async_request_{&cluster_manager_.thread_local_cluster_.async_client_},
       thread_factory_(Thread::threadFactoryForTest()) {
   ON_CALL(api_, threadFactory()).WillByDefault(testing::ReturnRef(thread_factory_));
-  main_dispatcher_ =
+  worker_thread_dispatcher_ =
       std::make_unique<Event::DispatcherImpl>("filter_fuzz_test", api_, api_.time_system_);
   // This is a decoder filter.
   ON_CALL(filter_callback_, addStreamDecoderFilter(_))
@@ -73,6 +73,13 @@ void UberFilterFuzzer::fuzz(
         proto_config, factory_context_.messageValidationVisitor(), factory);
     // Clean-up config with filter-specific logic before it runs through validations.
     cleanFuzzedConfig(proto_config.name(), message.get());
+    if (with_main_event_loop_) {
+      // For filters that need main event loop.
+      ON_CALL(encoder_callbacks_, dispatcher())
+          .WillByDefault(testing::ReturnRef(*worker_thread_dispatcher_));
+      ON_CALL(decoder_callbacks_, dispatcher())
+          .WillByDefault(testing::ReturnRef(*worker_thread_dispatcher_));
+    }
     cb_ = factory.createFilterFactoryFromProto(*message, "stats", factory_context_);
     cb_(filter_callback_);
   } catch (const EnvoyException& e) {
@@ -93,15 +100,15 @@ void UberFilterFuzzer::fuzz(
 
   if (with_main_event_loop_) {
     // Ensure the event loop gets at least one event to end the test.
-    auto end_timer =
-        main_dispatcher_->createTimer([]() { ENVOY_LOG_MISC(trace, "server timer fired"); });
+    auto end_timer = worker_thread_dispatcher_->createTimer(
+        []() { ENVOY_LOG_MISC(trace, "server timer fired"); });
     end_timer->enableTimer(std::chrono::milliseconds(5000));
     // If we were successful, run any pending events on the main thread's dispatcher loop. These
     // might be, for example, pending DNS resolution callbacks. If they generate exceptions, we want
     // to explode and fail the test, hence we do this outside of the try-catch above.
     // TODO(ravenblackx): this can potentially exit the loop without the filter having completed.
     // We should detect stream completion rather than just running one round of dispatcher.
-    main_dispatcher_->run(Event::DispatcherImpl::RunType::Block);
+    worker_thread_dispatcher_->run(Event::DispatcherImpl::RunType::Block);
   }
   reset();
 }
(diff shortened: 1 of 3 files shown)


vehre-x41:
CC @ravenblackx can you take a look, whether this way is what you had in mind in #21726 ?

CC @adisuissa because you were on the original issue ;-)

test/extensions/filters/http/common/fuzz/uber_filter.cc
```diff
@@ -85,6 +91,16 @@ void UberFilterFuzzer::fuzz(
     HttpFilterFuzzer::accessLog(access_logger_.get(), stream_info_);
   }
 
+  if (with_main_event_loop_) {
+    // Ensure the event loop gets at least one event to end the test.
+    auto end_timer =
+        main_dispatcher_->createTimer([]() { ENVOY_LOG_MISC(trace, "server timer fired"); });
+    end_timer->enableTimer(std::chrono::milliseconds(5000));
```
ravenblackx, line 98: Is there any way to validate that the filter has completed? Rather than a hard five second loop it would make more sense to do a gentle spin-loop until either the filter completes or a max timeout is reached. As-is, this will pass on the error case "filter never completes", which is probably a worse case than an exception.

I'm thinking similar to this but inside a while (filter not completed) block, with say a 50ms timer, and an if (total time elapsed > threshold) FAIL() branch. That way the test can complete in less than 5 seconds (this method makes it take 5 seconds even if it could have completed immediately), and can detect additional types of failure.

But this does look like it would fix the specific failure case it was targeting.

ravenblackx: (Added bonus of this approach, you shouldn't need to specify whether the dispatcher is even required - if it's not then the filter should have completed the first time so it wouldn't even go into the loop.)

vehre-x41: The `main_dispatcher_->run(Event::DispatcherImpl::RunType::Block);` blocks until it has processed **one** event. That is, when the filter's event is already waiting, when we reach this line, it is processed immediately and the event-loop is terminated. The timer event is deleted when the `end_timer` object goes out of scope. Therefore, when the filter is quick, one does not note the 5 secs at all (they never occur). But when the filter takes forever, then its timeout is 5 seconds, because the event-loop will process the timer event.

One can of course FAIL in the timer callback, when that one is the reason for the event-loop to end. Is that desired?

test/extensions/filters/http/common/fuzz/uber_per_filter.cc
```diff
@@ -176,6 +177,10 @@ void UberFilterFuzzer::perFilterSetup() {
   ON_CALL(decoder_callbacks_, decodingBuffer()).WillByDefault([this]() -> const Buffer::Instance* {
     return decoding_buffer_;
   });
+
+  // For filters that need main event loop.
+  ON_CALL(encoder_callbacks_, dispatcher()).WillByDefault(testing::ReturnRef(*main_dispatcher_));
```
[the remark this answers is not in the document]
vehre-x41: ??? The one callback is for en-coders and the other for de-coders. Yes, both shall return the same main_dispatcher, but the origin objects are different...

test/extensions/filters/http/common/fuzz/uber_filter.cc
ravenblackx: My concern here is that there's no guarantee that there's only supposed to be one dispatched event (I'd be surprised if that's the case for file_system_buffer) - if run only processes one event (that wasn't my understanding from the docs, I thought with block it processed until the queue is empty, but I could be wrong) then without a while loop you're likely just testing up to the file write, and not all the way through the filter. Ideally the test would be able to validate correctly with anywhere from zero to n dispatched events, without long-blocking, which is how it would be with my proposed loop (provided there is a way to validate that the filter terminated!) There might be a race if run consumes the queue until empty, in that if the new events are added quickly enough the filter will complete, but if the new event is not added until after run has terminated then the filter gets left with a queued event (after the other thread adds it) that is never processed.

Though yeah, if Block exits after either one event or at least one event, then an "until filter is terminated" loop with a 5 second fail timeout would be a solid solution.

To clarify: the *two* situations I'm concerned about are:
1. the dispatcher doesn't get an event because the filter doesn't do anything async. In the current implementation you either have to explicitly avoid the dispatcher block entirely, or suffer a 5 second delay. With a loop predicated on the filter terminating, you wouldn't enter the loop in this case because the filter completes before you reach the loop if there's no async action involved.
2. the dispatcher receives more than one event, asynchronously. In the current implementation the test passes, the filter is left in an incomplete state, and there is no validation that the filter completes. With a loop predicated on the filter terminating, all dispatched events get processed and if the filter still doesn't complete in 5 seconds the test can fail.

test/extensions/filters/http/common/fuzz/uber_per_filter.cc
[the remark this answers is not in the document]
ravenblackx: Ah, yes, sorry, that was my misreading.

test/extensions/filters/http/common/fuzz/uber_filter.cc
vehre-x41: I hope I understand your concerns. You are right `run(Block)` will process at least one event. More if possible, but I never have seen this to be deterministic or controlable. Therefore I prefer to regard `run(Block)` to process one event to keep it simple.

For the issue to figure when a filter has terminated: To my understanding are filters created in envoy and live until envoy terminates. They are kept in stand-by and are applied to streams whenever a stream is applicable. Therefore is there no termination criteria. To *my* understanding! Correct me, if I am wrong. The encoder- and decoder-filter both have some kind of `en-/decodeComplete()` method, which *may* be usable here. (I say "may" because I have been surprised by envoy many times, that methods do not as suspected.) To use these (virtual) methods one would have to overload the class for every filter to inject code into that process. With the number of filters available I don't think this is feasible. I have examined the `en-/decoder_callbacks`, which are mocks in the `filter_fuzz_test`. Those could be easily used to indicate a filter "termination", but I could not identify a callback, that would suit our needs (which does not mean none exists; the name could just be counter-intuitive, so that I never got the idea to use that specific callback for that task). Therefore I am kind of stuck here. 

Furthermore do it not know, what the requirements in the design of the `filter_fuzz_test` are/were. I.e. it is unclear, whether it is ok to instantiate a filter and terminated it in the middle of its processing or if the filter's whole task should be carried out in each run of the fuzzer. With the current implementation I doubt this is even possible, because the fuzzer knows nothing about the structure of messages a filter needs to be usable.

After all of this chatter, I am open for any ideas of how to make the processing more reliable. Not having found anything usable yet, I stick to the PR as is. Feel free to add your (code) proposals to make the fuzzer work better.

ravenblackx (dismissed):

test/extensions/filters/http/common/fuzz/uber_filter.cc
vehre-x41: Thx

vehre-x41:
@adisuissa do you approve, too?

test/extensions/filters/http/common/fuzz/uber_filter.cc
ravenblackx: Actually, could you put in this loop, to help me to remember to get to it after this lands...

```
// TODO(ravenblackx): this can potentially exit the loop without the filter having completed.
// We should detect stream completion rather than just running one round of dispatcher.
```

vehre-x41: comment added.

test/extensions/filters/http/common/fuzz/uber_per_filter.cc
```diff
@@ -176,6 +177,10 @@ void UberFilterFuzzer::perFilterSetup() {
...
+  ON_CALL(decoder_callbacks_, dispatcher()).WillByDefault(testing::ReturnRef(*main_dispatcher_));
```
adisuissa, line 183: I think this will prevent filters that use the dispatcher and don't have `with_main_event_loop_` set to true from executing their dispatcher callback code.

I'm contemplating whether the file system buffer filter will (option 1) benefit from its own fuzzer (like we do for ext_authz and ext_proc for example), or (option 2) whether we can hack this by setting these ON_CALLs in the `UberFilterFuzzer::fuzz` method if this filter is being tested, and resetting it when the method ends.

For now let's take option 2 if possible.

vehre-x41: The `ON_CALL`s are now only inserted, when the flag is set.

test/extensions/filters/http/common/fuzz/uber_filter.cc
```diff
@@ -67,6 +73,13 @@ void UberFilterFuzzer::fuzz(
         proto_config, factory_context_.messageValidationVisitor(), factory);
     // Clean-up config with filter-specific logic before it runs through validations.
     cleanFuzzedConfig(proto_config.name(), message.get());
+    if (with_main_event_loop_) {
```
adisuissa, line 76: I think that due to the static definition of the UberFilterFuzzer object, once these are set they won't be cleared.
Executing `VerifyAndClear` when resetting should remove these (but need to add the relevant ON_CALLs that are typically done once in the c'tor).

adisuissa (commented):
Thanks for fixing this!
I've added a minor comment.

wrowe:
Note that US holidays and PTO may delay responses this week. Ping @adisuissa when you are able, thanks!

phlax:
@adisuissa seems like feedback has been addressed, would you mind reviewing again

/wait-any

PR #21302, opened 2022-05-16
route: add stats prefix for route

ramaraochavali:
Commit Message: add stats prefix for route
Additional Description: Adds API for stats generation support per route.
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:

Fixes https://github.com/envoyproxy/envoy/issues/3351

PR #21765, opened 2022-06-18
make route stats context as pure class

ramaraochavali:
Commit Message:make route stats context as pure class
Additional Description:make route stats context as pure class
Risk Level: Low
Testing: Existing tests
Docs Changes: N/A
Release Notes: N/A


ramaraochavali:
@jmarantz I realized there is already a `context.h` in router package that we can use. Changed it like that. PTAL

jmarantz (commented):
nice; thank you!

one tiny comment otherwise lgtm.

coverage failure looks like a flake; it might pass when you remove the 'explicit'.

jmarantz (dismissed):

envoy/router/context.h
```diff
@@ -35,5 +35,23 @@ class Context {
...
+ * RouteStatsContext defines config needed to generate all route level stats.
+ */
+class RouteStatsContext {
+public:
+  virtual ~RouteStatsContext() = default;
+
+  /**
+   * @return stat name for route specific stats.
+   */
+  virtual const Stats::StatName statName() const PURE;
```
htuch, line 48: What does it mean to return a `const` of a value?

ramaraochavali: yeah. it is not needed. Will remove

htuch (commented):
LGTM, but I'm curious on why this change? Is there some testing or multi-implementation use case? Just asking as on its own, it's not an ideal to make everything fully abstract, e.g. implementation local classes can be just that.

ramaraochavali:
@htuch sorry for not providing context on this in PR. @jmarantz suggested this in another PR https://github.com/envoyproxy/envoy/pull/21302#discussion_r895996089 - but did not do it in that PR.

jmarantz:
Yeah I suggested it because this PR eliminates the need to have a somewhat complex constructor inlined into a interface file.

htuch (approved):
Great, thanks!

PR #21767, opened 2022-06-18
ci: Re-organize format jobs

phlax:
Commit Message:
Additional Description:

Now that the format_pre jobs are much faster, i think it makes sense to put the format jobs together, with a view to integrating them longer term

currently, the proto_format job takes a lot longer than everyting else - im working to optimize this separately, but i think for now lets keep this separate and run concurrently

this PR also makes a slight change to the check_format run in that it just runs `fix` immediately rather than running `check` and conditionally running `fix`

the current behaviour with proto_format and check_format grouped together means that if either fails both are run twice - so this at least resolves that (proto_format still runs twice if it fails - but we can address that as part of optimization)

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -13,34 +13,34 @@ pr: none
 stages:
 - stage: precheck
   jobs:
-  - job: format_pre
+  - job: format
     dependsOn: []
     pool:
       vmImage: "ubuntu-18.04"
     steps:
     - task: Cache@2
       inputs:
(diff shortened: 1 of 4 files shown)


phlax:
it also ensures that the protoxform_test job is run in check rather than (conditionally) in fix

lizan (approved):

PR #21750, opened 2022-06-16
proto_format: Use aspect to protoprint formatted api

phlax:
Commit Message:

This switches from calling protoprint in scripts with python multiproc to using a bazel aspect

Both the `proto_format.sh` job and the `protoprint_test` are switched to using a prebuilt protoprinted version of the api/test protos

Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #21769, opened 2022-06-18
proto_format: Prebuild formatted api for `proto_sync`

phlax:
Commit Message:
Additional Description:

Currently `proto_sync` compares 2 "api" directories:

- an ideal one created from what bazel can see
-  what is actually on the local fs

in the case of the former this can be built hermetically and benefit from bazels caching (and potentially incremental builds)

this PR makes the creation of the "formatted api directory" hermetic and separates it from the run job in which the directories are compared

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
this seems to work pretty reliably in CI breaking/fixing protos and testng 

it ~halves the time to run the proto_format job, perhps more if i wasnt blowing the cache

re bazel cache - im not 100% how it will work currently locally and making changes, i will test further

phlax:
i tested this out further and realized that only the "ideal" representation of api fs can be prebuilt - the "actual" needs to use what is currently on the fs so cannot be hermetic

prebuilding only the ideal api (ie the one built from what bazel can see) speeds this up a lot in every situation (uncached run having the least benefit)

i have also taken this a step further and switched the protoprint jobs that are currently run in a python multiproc pool inside proto_sync.py to use an aspect

this seems to speed things up quite a bit - altho in the uncached run it is doing a bit more work as it currently only protoprints changed files, whereas this lets bazel manage it, meaning it has to check all the first time

running protoxform/protoprint against all files there are a few v2 protos that fail formatting - not sure if we would be better just fixing these, or to add an exclusion for them

cc @htuch

phlax:
> running protoxform/protoprint against all files there are a few v2 protos that fail formatting - not sure if we would be better just fixing these, or to add an exclusion for them

after testing this further this seems to be an issue with my impiementation so will debug further - i have for that reason split out the part shifting `protoprint` to an aspect (to #21750 ) as i think this can/should land more quickly and that PR throws up other issues which i am addressing separately

tools/proto_format/BUILD
```diff
@@ -54,9 +56,20 @@ envoy_py_data(
     src = ":data_srcs",
 )
 
+pkg_files(
+    name = "protoprinted_files",
+    srcs = ["//tools/protoxform:api_protoxform"],
+    strip_prefix = strip_prefix.from_pkg(),
```
phlax, line 62: this is pretty annoying - it used to be that to create a `pkg_tar` _and_ preserve dir structure you just gave it `srcs` now to do this you _must_ have 2 rules and this weird `strip_prefix` invokation

tools/proto_format/BUILD
```diff
@@ -54,9 +56,20 @@ envoy_py_data(
...
+)
+
+pkg_tar(
+    name = "protoprinted",
```
phlax, line 66: maybe call this xformed or similar - or maybe desciptors or similar

tools/proto_format/format_api.py
```diff
@@ -0,0 +1,311 @@
+#!/usr/bin/env python3
+
+# Take protoxform artifacts and pretty-print with protoprint.py.
+
+import argparse
+from collections import defaultdict
```
phlax, line 6: shift after direct imports

phlax:
done

PR #19367, opened 2021-12-31
listener: support multiple addresses

soulxu:
Commit Message: listener: support multiple addresses
Additional Description:

This PR adds the ability of the listener to support multiple listening addresses.

The new field `addresses` was added to Listener API. The additional listening addresses can be added. Each address should be unique, and all the addresses should be the same protocol. The Envoy Internal address won't support this implementation since there is no use case for now, but it can be added in the future.

The `ListenerImpl`, `ListenerManagerImpl`, and `ConnectionHandlerImpl` will be aware of the multiple addresses. The `ListenerImpl` will contains multiple addresses and the corresponding socket factories. `ListenerManagerImpl` will be aware of the multiple addresses when adding or updating the listener. `ConnectionHandlerImpl` is responsible for creating the `ActiveListener` for each address.

The `ActiveListener` for UDP and TCP won't be aware of the multiple addresses, each `ActiveListener` will listen to a single address as before. We also can consider one `ActiveListener` to serve multiple addresses, which will need more change. If this is needed, we can do this in the future.

The `ListenerImpl` will create `ConnectionBalancer` and `UdpWorkerRouter` for each address. The `ActiveTcpListener` and `ActiveUdpListener` will reference to the address-specific balancer and worker router.

Also, the admin API `/listeners` is changed to support returning of multiple addresses.

Risk Level: high
Testing: unit tests and integration tests added
Docs Changes: API doc
Release Notes: new feature
Platform Specific Features: no
Fixes #11184

Issue #11184, opened 2020-05-13
Support multiple addresses per listener


mattklein123:
Via https://github.com/envoyproxy/envoy/issues/11124. This would avoid mass listener duplication in cases where a listener config needs to be bound to multiple addresses.

ggreenway:
This may be required for some QUIC use cases. If Envoy is terminating QUIC and sets a Server Preferred Address, packets will start arriving on a different listener, but are part of the same QUIC connection. So packets from different listeners will need to go to the same QUIC connection router.

Same story for QUIC connection migration.

This will also remove duplication if both HTTP/2 and HTTP/3 are needed for the same configuration (UDP vs TCP listeners).

uablrek:
FYI the same for `contour` which also uses envoy https://github.com/projectcontour/contour/pull/1162

soulxu:
I'm trying to work out a PoC for this issue

danzh2010:
> This may be required for some QUIC use cases. If Envoy is terminating QUIC and sets a Server Preferred Address, packets will start arriving on a different listener, but are part of the same QUIC connection. So packets from different listeners will need to go to the same QUIC connection router.
> 
> Same story for QUIC connection migration.

I think supporting dual-stack ipv6 sockets and QUIC server migration are different things. The former only needs a 2nd listener with the same config but listening on a different socket. But QUIC server migration requires one connection lands on the same listener. #19367 seems to implement the former, but I don't see how it helps with the latter. QUIC server migration might be easier to implement in L4 load balancer rather than requires one listener to listen on both sockets. Google is planning to experiment server migration, but we don't have a good story to tell yet.

PR #21774, opened 2022-06-20
listener: enable connection balancer to support multiple addresses listener

soulxu:
Commit Message: listener: enable connection balancer to support multiple addresses listener
Additional Description:
This PR enables the `ListenerImpl` to create connection balancer for each listening address. And pass the specific connection balancer to `ActiveTcpListener`. Also enable the `ConnectionHandlerImpl::getBalancedHandlerByTag` to distinguish different address.

Please reference the full implementation of multiple addresses in listener at https://github.com/envoyproxy/envoy/pull/19367

Risk Level: high
Testing: unittest
Docs Changes: n/a
Release Notes: n/a


envoy/network/connection_handler.h
```diff
@@ -13,6 +13,8 @@
 
 #include "source/common/common/interval_value.h"
 
+#include "address.h"
```
mattklein123, line 16: nit: use fully specified include path. Same elsewhere.

source/server/active_tcp_listener.h
```diff
@@ -81,6 +85,9 @@ class ActiveTcpListener final : public Network::TcpListenerCallbacks,
   // The number of connections currently active on this listener. This is typically used for
   // connection balancing across per-handler listeners.
   std::atomic<uint64_t> num_listener_connections_{};
+
+  Network::ConnectionBalancer& connection_balancer_;
+  Network::Address::InstanceConstSharedPtr address_;
```
mattklein123, line 90: This listener must know it's address because it has a socket, right? It seems like this shouldn't be necessary?

source/server/listener_impl.h
```diff
@@ -464,7 +471,7 @@ class ListenerImpl final : public Network::ListenerConfig,
   const bool continue_on_listener_filters_timeout_;
   std::shared_ptr<UdpListenerConfigImpl> udp_listener_config_;
   std::unique_ptr<Network::InternalListenerConfig> internal_listener_config_;
-  Network::ConnectionBalancerSharedPtr connection_balancer_;
+  absl::flat_hash_map<std::string, Network::ConnectionBalancerSharedPtr> connection_balancers_;
```
mattklein123, line 474: Can you add a comment on what this is hashed on? I assume address string but I'm not sure? It would be nice to eventually add hash support for address if it doesn't already exist so that we don't have to do this anymore but seems fine for now.

test/server/connection_handler_test.cc
```diff
@@ -149,7 +149,9 @@ class ConnectionHandlerTest : public testing::Test, protected Logger::Loggable<L
     void setDirection(envoy::config::core::v3::TrafficDirection direction) {
       direction_ = direction;
     }
-    Network::ConnectionBalancer& connectionBalancer() override { return *connection_balancer_; }
+    Network::ConnectionBalancer& connectionBalancer(const Network::Address::Instance&) override {
```
mattklein123, line 152: Is there a test somewhere in this PR that actually tests multiple addresses and multiple balancers?

mattklein123 (changes requested):
Thanks. Generally LGTM. Flushing out some comments.

/wait

envoy/network/connection_handler.h
soulxu: got it

source/server/active_tcp_listener.h
soulxu: I have think about another two ways:
* We can get the address from the accepted socket, but it won't work with the case of the connection doesn't use original destination.
* The socket object is held by the socket listener, then we can add an interface to the listener to return the socket address.
   https://github.com/envoyproxy/envoy/blob/17d7d7908507888617824d73bc5904a3d79d1d17/envoy/network/listener.h#L398

The second way works also. But let me know if the second way you think better.

source/server/listener_impl.h
soulxu: Got it, I can do that in later PR, I will leave a todo here, then back to this later.

test/server/connection_handler_test.cc
soulxu: I don't have a test for multiple addresses yet. I'm thinking to add integration test in the end
https://github.com/envoyproxy/envoy/blob/17d7d7908507888617824d73bc5904a3d79d1d17/test/integration/listener_lds_integration_test.cc#L1055

I should be able test multiple addresses for ConnectionHandler and ActiveTcpListener. ListenerImpl can't be tested now, since it isn't full support multiple addresses yet.

soulxu: A multiple address test added for connection handler https://github.com/envoyproxy/envoy/pull/21774/files#diff-3df0a7bd0288faa35b611e2bc11ea94713d9ca617ad16dfab81386c814f5e1eeR701

source/server/active_tcp_listener.h
soulxu: I add a comment here https://github.com/envoyproxy/envoy/pull/21774/files#diff-55b19f848d75e4c86be99ac5c87457e24a960e71913de4b60153ea5e47c4d32cR90

But let me know if you prefer another way.

source/server/active_tcp_listener.h
```diff
@@ -87,6 +87,9 @@ class ActiveTcpListener final : public Network::TcpListenerCallbacks,
   std::atomic<uint64_t> num_listener_connections_{};
 
   Network::ConnectionBalancer& connection_balancer_;
+  // This is the address of this listener is listening on. And used for get the correct listener
```
mattklein123, line 90:
```suggestion
  // This is the address this listener is listening on. It's used to get the correct listener
```

mattklein123 (commented):
Thanks a few more questions and comments.

/wait

source/server/active_tcp_listener.h
```diff
@@ -87,6 +87,9 @@ class ActiveTcpListener final : public Network::TcpListenerCallbacks,
...
+  // when rebalancing. The accepted socket can't be used to get the listening address, since
+  // the accepted socket's remote address can be another address than the listening address.
   Network::Address::InstanceConstSharedPtr address_;
```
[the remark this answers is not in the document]
soulxu: After add multiple addresses support for ListenerImpl, then the ListenerImpl's `address()` will changes to `addresses()` (also need to move `address()` interface to base `ListenerConfig`), there still no way for ActiveListener knows which address is listening. I have tried to pass an index of `addresses()` to the ActiveListener, but I thought the `index` isn't readable in the end, hold a  `Network::Address::InstanceConstSharedPtr` is more readable.

mattklein123 (approved):
Thank you!

PR #21781, opened 2022-06-20
fuzz test: Include on_no_match.matcher into validation.

vehre-x41:
Commit Message: fuzz test: Include on_no_match.matcher into validation.
Additional Description:
In an on_no_match not only an action can be set, but also a matcher.
This is now included into checks for valid configuration.
Corpus added.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/common/router/route_fuzz_test.cc
+++ b/test/common/router/route_fuzz_test.cc
@@ -65,13 +65,15 @@ bool validateOnMatchConfig(const xds::type::matcher::v3::Matcher::OnMatch& on_ma
     return !isUnsupportedRouteConfig(on_match_route_action_config);
   }
   case xds::type::matcher::v3::Matcher_OnMatch::ON_MATCH_NOT_SET: {
-    // This alternative shall never occur. It is here to case on all values of Matcher_OnMatch and
-    // this way ensures that the compiler will report new entries to that enum as warning, which
-    // will finally result in an compiler error, when error on all warnings is set.
-    ASSERT(false);
+    // This alternative should never occur. But of course it does. By returning false the
+    // fuzzer will not follow this path any further.
     return false;
   }
+    // By not providing a default: the compiler will indicate new constants in the
+    // Matcher_OnMatch enum as compile warnings.
   }
+  // This is present only to silence compilers that do not analyze the switch above correctly.
+  return false;
 }
 
 bool validateMatcherConfig(const xds::type::matcher::v3::Matcher& matcher) {
--- a/test/common/router/route_fuzz_test.cc
+++ b/test/common/router/route_fuzz_test.cc
@@ -69,11 +69,8 @@ bool validateOnMatchConfig(const xds::type::matcher::v3::Matcher::OnMatch& on_ma
     // fuzzer will not follow this path any further.
     return false;
   }
-    // By not providing a default: the compiler will indicate new constants in the
-    // Matcher_OnMatch enum as compile warnings.
   }
-  // This is present only to silence compilers that do not analyze the switch above correctly.
-  return false;
+  PANIC_DUE_TO_CORRUPT_ENUM;
 }
 
 bool validateMatcherConfig(const xds::type::matcher::v3::Matcher& matcher) {
--- a/test/common/router/route_fuzz_test.cc
+++ b/test/common/router/route_fuzz_test.cc
@@ -48,22 +48,20 @@ cleanRouteConfig(envoy::config::route::v3::RouteConfiguration route_config) {
 bool validateMatcherConfig(const xds::type::matcher::v3::Matcher& matcher);
 
 bool validateOnMatchConfig(const xds::type::matcher::v3::Matcher::OnMatch& on_match) {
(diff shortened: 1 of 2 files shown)


vehre-x41:
/assign-from @envoyproxy/x41-fuzz-triage

Took the shortcut directly to public, because ctec is too old and the change is minor.

tyxia (commented):
Thanks for working on this

test/common/router/route_fuzz_test.cc
```diff
@@ -101,7 +101,8 @@ bool validateMatcherConfig(const xds::type::matcher::v3::Matcher& matcher) {
       }
     }
   }
-  if (matcher.on_no_match().has_action() && !validateOnMatchConfig(matcher.on_no_match())) {
+  if ((matcher.on_no_match().has_action() || matcher.on_no_match().has_matcher()) &&
```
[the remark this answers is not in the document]
vehre-x41: ad 1)  Agreed, changed.

ad 2) I don't know, if I got you there correctly. Please have look at https://github.com/envoyproxy/envoy/pull/21781/commits/8d19fa750b4459fcbf3b99a9fc87dca7b136a48b to see, if I got you correctly.

tyxia (dismissed):
LGTM

test/common/router/route_fuzz_test.cc
```diff
@@ -48,22 +48,20 @@ cleanRouteConfig(envoy::config::route::v3::RouteConfiguration route_config) {
...
-      std::string::npos) {
-    return false;
-  }
-  envoy::config::route::v3::Route on_match_route_action_config;
-  MessageUtil::unpackTo(on_match.action().typed_config(), on_match_route_action_config);
-  ENVOY_LOG_MISC(trace, "typed_config of on_match.action is: {}",
-                 on_match_route_action_config.DebugString());
-  if (isUnsupportedRouteConfig(on_match_route_action_config)) {
-    return false;
+  if (on_match.has_matcher()) {
```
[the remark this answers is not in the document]
vehre-x41: I went for the switch-case way. I like that more, because without a default-alternative the clang-compiler will warn about new entries. And with treat all warnings as errors it will point the developer to the location he missed this way ensuring that he covers the missing cases.

test/common/router/route_fuzz_test.cc
```diff
@@ -48,22 +48,29 @@ cleanRouteConfig(envoy::config::route::v3::RouteConfiguration route_config) {
...
+        std::string::npos) {
+      return false;
+    }
+    envoy::config::route::v3::Route on_match_route_action_config;
+    MessageUtil::unpackTo(on_match.action().typed_config(), on_match_route_action_config);
+    ENVOY_LOG_MISC(trace, "typed_config of on_match.action is: {}",
+                   on_match_route_action_config.DebugString());
+    return !isUnsupportedRouteConfig(on_match_route_action_config);
+  }
+  case xds::type::matcher::v3::Matcher_OnMatch::ON_MATCH_NOT_SET: {
```
tyxia, line 67: you will still need to have return statement here for this case so that all the branches all covered. Otherwise, it should generate build error

test/common/router/route_fuzz_test.cc
```diff
@@ -48,22 +48,29 @@ cleanRouteConfig(envoy::config::route::v3::RouteConfiguration route_config) {
...
+    // This alternative shall never occur. It is here to case on all values of Matcher_OnMatch and
+    // this way ensures that the compiler will report new entries to that enum as warning, which
+    // will finally result in an compiler error, when error on all warnings is set.
+    ASSERT(false);
```
tyxia, line 71: I am not sure if here will work as you explain.  Let's say have new enum `kNew`,  this `xds::type::matcher::v3::Matcher_OnMatch::ON_MATCH_NOT_SET` will not catch it , right?

I think you will need the assert at the end out side of all three cases? something like below : 
switch (on_match.on_match_case()) {
   case kMatcher: {
   .......
   } 
   case kAction: {
   .......
   }
  case ON_MATCH_NOT_SET: {
  .....
  }
  }
  PANIC_DUE_TO_CORRUPT_ENUM;

test/common/router/route_fuzz_test.cc
vehre-x41: Being in a debug build during my patching the assert(false) of course terminated the flow of control each time and was not problem. But for a release build you're right. I fixed that now. Thanks for pointing out.

test/common/router/route_fuzz_test.cc
tyxia: That is only when warnings in clang is enabled, right?
Even though envoy enable most warnings, I think it is better to program in a generic way-- i.e., compiler agnostic or compiler setting agnostic way

adisuissa (commented):
LGTM, thanks.
@tyxia for final pass.

test/common/router/route_fuzz_test.cc
vehre-x41: changed

PR #21786, opened 2022-06-20
fuzz test: Sanitize status header before use, part 2.

vehre-x41:
Commit Message: fuzz test: Sanitize status header before use, part 2.
Additional Description:
In conn_manager_impl_fuzz_test the:status header needs to be sanitized
before use, too, due to the change having the getResponseHeader() no
longer throw but bug.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/common/http/conn_manager_impl_fuzz_test.cc
+++ b/test/common/http/conn_manager_impl_fuzz_test.cc
@@ -483,14 +483,10 @@ public:
         // The client codec will ensure we always have a valid :status.
         // Similarly, local replies should always contain this.
         const auto status = Utility::getResponseStatusOrNullopt(*headers);
-        if (!status.has_value()) {
-          headers->setReferenceKey(Headers::get().Status, "200");
-        }
         // The only 1xx header that may be provided to encodeHeaders() is a 101 upgrade,
         // guaranteed by the codec parsers. See include/envoy/http/filter.h.
-        if (CodeUtility::is1xx(status.value_or(enumToInt(Http::Code::OK))) &&
-            status.value_or(enumToInt(Http::Code::OK)) !=
-                enumToInt(Http::Code::SwitchingProtocols)) {
+        if (!status.has_value() || (CodeUtility::is1xx(status.value()) &&
+                                    status.value() != enumToInt(Http::Code::SwitchingProtocols))) {
           headers->setReferenceKey(Headers::get().Status, "200");
         }
         decoder_filter_->callbacks_->encodeHeaders(std::move(headers), end_stream, "details");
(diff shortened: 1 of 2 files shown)


vehre-x41:
CC @adisuissa @KBaichoo @pradeepcrao

test/common/http/conn_manager_impl_fuzz_test.cc
```diff
@@ -482,15 +482,15 @@ class FuzzStream {
...
-        } catch (const CodecClientException&) {
+        const auto status = Utility::getResponseStatusOrNullopt(*headers);
+        if (!status.has_value()) {
           headers->setReferenceKey(Headers::get().Status, "200");
         }
         // The only 1xx header that may be provided to encodeHeaders() is a 101 upgrade,
         // guaranteed by the codec parsers. See include/envoy/http/filter.h.
-        if (CodeUtility::is1xx(status) && status != enumToInt(Http::Code::SwitchingProtocols)) {
+        if (CodeUtility::is1xx(status.value_or(enumToInt(Http::Code::OK))) &&
+            status.value_or(enumToInt(Http::Code::OK)) !=
```
tyxia, line 492: The `value_or` here is redundant. Once you check `has_value` above, we can do it without `value_or` by small code tweaks. 

But please address my question above first and then we can change here accordingly.

test/common/http/conn_manager_impl_fuzz_test.cc
```diff
@@ -482,15 +482,15 @@ class FuzzStream {
             Fuzz::fromHeaders<TestResponseHeaderMapImpl>(response_action.headers()));
         // The client codec will ensure we always have a valid :status.
         // Similarly, local replies should always contain this.
-        uint64_t status;
-        try {
-          status = Utility::getResponseStatus(*headers);
-        } catch (const CodecClientException&) {
+        const auto status = Utility::getResponseStatusOrNullopt(*headers);
+        if (!status.has_value()) {
           headers->setReferenceKey(Headers::get().Status, "200");
```
tyxia, line 487: Not your code. But I am curious: why do we set status 200 i.e., (OK) status for case that no status in the headers ? Should we consider this as the error (like 400)?

tyxia (commented):
Thanks for working on this!

test/common/http/conn_manager_impl_fuzz_test.cc
tyxia: Yes. exactly.  Thanks！

test/common/http/conn_manager_impl_fuzz_test.cc
tyxia: Yeah, on second thought maybe it doesn't matter too much for fuzzer.
Thanks for the info!

PR #21794, opened 2022-06-21
Fix wasm network filter example

hanxiaop:
Commit Message: fix wasm network filter config example
Additional Description: N/A
Risk Level: N/A
Testing: N/A
Docs Changes: Wasm Network Filter
Release Notes: N/A
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


docs/root/configuration/listeners/network_filters/wasm_filter.rst
```diff
@@ -23,15 +23,14 @@ Example filter configuration:
...
-    config:
-      config:
-        name: "my_plugin"
-        vm_config:
-          runtime: "envoy.wasm.runtime.v8"
-          code:
-            local:
-              filename: "/etc/envoy_filter_http_wasm_example.wasm"
-          allow_precompiled: true
+     config:
```
phlax, line 26: hi @hanxiaop thanks for working on this

with configs in the docs we are moving them to standalone bootstrap yaml files so that they can be validated and tested

would you be able to move this one to a boostrap config ?

if you do `git grep literalinclude` you should be able to find existing examples

hanxiaop: @phlax Thanks for reminding. I've revised to a valid standalone config, but however, I think the config cannot pass CI since it needs a `*.wasm` file, and it looks like the CI system supports multiarch. Do you have any suggestions?

phlax: if you check existing configs for wasm, i think there is a wasm bin in the test build - that should work i think

(see `examples/wasm-cc/envoy.yaml`)

re arch support - i would look how the existing wasm config test is restricted - see `examples/wasm-cc/BUILD`

docs/root/configuration/listeners/network_filters/_include/wasm-filter-proxy.yaml
```diff
@@ -0,0 +1,25 @@
+static_resources:
+  listeners:
+    - name: main
```
phlax, line 3: could you use least indent for yaml formatting please, it makes configs much more readable

docs/root/configuration/listeners/network_filters/_include/wasm-filter-proxy.yaml
```diff
@@ -0,0 +1,25 @@
...
+            - name: envoy.filters.network.wasm
+              typed_config:
+                "@type": type.googleapis.com/envoy.extensions.filters.network.wasm.v3.Wasm
+                config:
+                  name: "my_plugin"
+                  vm_config:
+                    runtime: "envoy.wasm.runtime.v8"
+                    code:
+                      local:
+                        filename: "envoy_filter_http_wasm_example.wasm"
```
phlax, line 19: i think the `filename` needs to be `/etc/envoy_filter_http_wasm_example.wasm`

docs/root/configuration/listeners/network_filters/wasm_filter.rst
```diff
@@ -18,20 +18,8 @@ Example configuration
...
-      config:
-        name: "my_plugin"
-        vm_config:
-          runtime: "envoy.wasm.runtime.v8"
-          code:
-            local:
-              filename: "/etc/envoy_filter_http_wasm_example.wasm"
-          allow_precompiled: true
+.. literalinclude:: _include/wasm-filter-proxy.yaml
+    :language: yaml
```
phlax, line 22: probably we want to add some linenos, caption, etc - see other `literalincludes` for examples

phlax (changes requested):
i think you shouldnt need to add wasm bin and it probably wont work with binary config check - using the existing one in the config env should i think

phlax:
@hanxiaop would be great to land this - shouldnt be too hard to fix up

/wait

hanxiaop:
not stale

phlax:
i can pick this one up again

/retest

hanxiaop:
> 

@jmarantz Sorry for that. Thanks for the reminder!

hanxiaop:
> i can pick this one up again
> 
> /retest

@phlax I've tried different ways but still doesn't work. Could you please take a look? Thanks!

docs/root/configuration/listeners/network_filters/_include/wasm-filter-proxy.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - name: envoy.filters.network.wasm
+        typed_config:
+          "@type": type.googleapis.com/envoy.extensions.filters.network.wasm.v3.Wasm
+          config:
+            name: "my_plugin"
+            vm_config:
+              runtime: "envoy.wasm.runtime.v8"
+              code:
+                local:
+                  filename: "/etc/envoy_filter_http_wasm_example.wasm"
```
phlax, line 19:
```suggestion
                  filename: "lib/envoy_filter_http_wasm_example.wasm"
```

phlax (commented):
hi @hanxiaop as mentioned before the example config needs to point to the wasm binary that is available for testing

hanxiaop:
Force pushed since needs to fix DCO sign-off.

docs/root/configuration/listeners/network_filters/_include/wasm-filter-proxy.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - name: main
+    address:
+      socket_address:
+        address: 127.0.0.1
+        port_value: 10000
+    filter_chains:
+    - filters:
+      - name: envoy.filters.network.wasm
+        typed_config:
+          "@type": type.googleapis.com/envoy.extensions.filters.network.wasm.v3.Wasm
```
phlax, line 12: @hanxiaop not 100% sure what the issue is here - this is the error:

```console
C++ exception with description "Function call: customStatNamespaces
    The mock function has no default action set, and its return type has no default value set." thrown in the test body.
```

i think you probably beed to add a `stat_prefix` for this filter also - altho checking the existing wasm config example for http.wasm it doesnt have it so not sure (`examples/wasm-cc/envoy.yaml`) - defo worth a try

phlax: can you start an Envoy server with this config + a wasm bin ?

hanxiaop: > can you start an Envoy server with this config + a wasm bin ?

@phlax Yes I do.

At least it's not `no directory found` for the wasm bin. I'll try different configs out, thanks for the help!

docs/root/configuration/listeners/network_filters/_include/wasm-filter-proxy.yaml
```diff
@@ -0,0 +1,39 @@
...
+                  filename: "lib/envoy_filter_http_wasm_example.wasm"
+              allow_precompiled: true
+      - name: envoy.filters.network.tcp_proxy
+        typed_config:
+          "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
+          stat_prefix: tcp
+          cluster: tcp
+  clusters:
+    - name: tcp
+      type: strict_dns
```
phlax, line 28: altho this isnt strictly enforced by the server - the spec says that this should be all caps

```suggestion
      type: STRICT_DNS
```

hanxiaop: @phlax I've tried to start Envoy with `stat_prefix` for the filter but it showed that this field was not supported.  So I thought this was the problem.

PR #21795, opened 2022-06-21
protoxform: Use an aspect to build protoxform test

phlax:
Signed-off-by: Ryan Northey <ryan@synca.io>

Commit Message:
Additional Description:

This just adds/make use of an aspect, in a follow up i will shift the test to bazel and clean up this test file further

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #21796, opened 2022-06-21
protoxform: Shift protoxform test to bazel and rename -> `protoprint_test` 

phlax:
Commit Message:
Additional Description:

This shifts the protoxform/protoprint test to bazel and renames it to `//tools/protoprint:protoprint_test` on the basis that what is actually being tested is the protoprint proto visitor

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
this incorporates #21795 which didnt land (yet) due to merge conflict

tools/protoprint/BUILD
```diff
@@ -1,4 +1,9 @@
-load("@rules_python//python:defs.bzl", "py_binary")
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
```
phlax, line 1: this should be failing ci - kicking now...

phlax: fixed

phlax:
@htuch i think this one should be ready to land

it does pretty much the same thing as now - save that the protoxform/print test has been renamed and it makes use of the protoxform aspect rule to derive the xformed protos

behind this i have another PR that adds a `protoprint` aspect and generates the the final protoprinted protos using an aspect chain - `protoxform_aspect` -> `protoprint_aspect` -> `protosync`/`protoprint_test`

this now pretty much works - just tidying it up and trying to figure if we can land just the updates required for the `protoprint_test` first, separating out the changes to `proto_sync` itself after. 

if i can get this to work i think it will give some extra assurance that nothing is changing in terms of outputs/tests and will make the next PR a bit more digestable

phlax:
> if i can get this to work 

i wasnt able to separate out changes to `proto_sync` <> `protoprint_test` as when the `protoprint` module changes they both need to change (and vice versa)

this is mostly for the next PR anyhow - but i think we will get the assurance, because if the protoxform -> protoprint -> proto_sync chain was not working correctly then it would try to add/remove api files (as it has done getting it working)

tools/protoprint/protoprint_test.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    if label.startswith('/'):
+        label = label.replace('//', '/', 1)
+    elif label.startswith('@'):
+        label = re.sub(r'@.*/', '/', label)
+    else:
+        return label
+    label = label.replace(":", "/")
+    splitted_label = label.split('/')
+    return '/'.join(splitted_label[:len(splitted_label) - 1])[1:], splitted_label[-1]
```
htuch, line 36: TBH this is all a bit magical. I think it was before (in the aspect / Bazel rule logic), but still is. Can you include some example paths to show what this is doing?

tools/protoprint/protoprint_test.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def golden_proto_file(tmp, path, filename, version):
+    """Retrieve golden proto file path. In general, those are placed in tools/testdata/protoxform.
+
+    Args:
+        path: target proto path
```
htuch, line 43: Missing `tmp` arg. At some point should this be a class and `tmp` a member?

tools/protoprint/protoprint_test.py
```diff
@@ -0,0 +1,172 @@
...
+    Args:
+        result_file: result proto file
+        golden_file: golden proto file
+
+    Returns:
+        output and status code
+    """
+    command = 'diff -u '
+    command += str(result_file) + ' '
+    command += str(golden_file)
```
htuch, line 101: Prefer a join on a list here..

htuch (changes requested):
Looks good, just a few minor Python nits.

tools/protoprint/protoprint_test.py
phlax: yep for sure - this is a transitional step

tools/protoprint/protoprint_test.py
phlax: will swap to an f-string - should say that i havent made too much of an attempt to cleanup this on the basis of minimum change

tools/protoprint/protoprint_test.py
phlax: sure - again its just what it was doing previously, but copied out (with exception that it now takes off first char)

htuch (approved):

PR #21444, opened 2022-05-25
h3 pool: quic connection honors socket options

danzh2010:
Commit Message: pass 2 cluster config knobs socket options and transport socket options to EnvoyQuicClientSession and EnvoyQuicClientConnection.

Risk Level: low
Testing: new unit tests
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: n/a

PR #21802, opened 2022-06-21
quic: client connection takes socket option in cluster config

danzh2010:
Commit Message: make quic client connection also takes ClusterInfo::clusterSocketOptions() during creation, in addition to connection pool socketOption() added in https://github.com/envoyproxy/envoy/pull/21444

Also make sure that automatic port migration triggered by PTO applies the same socket options from the original socket in maybeMigratePort().

Risk Level: low, h3 uptream only
Testing: added unit tests and integration test
Docs Changes: n/a
Release Notes: n/a


alyssawilk (approved):
LGTM though it'd be lovely if we can come up with an integration test which will validate HTTP/1, HTTP/2 and HTTP/3 support

danzh2010:
> LGTM though it'd be lovely if we can come up with an integration test which will validate HTTP/1, HTTP/2 and HTTP/3 support

Thanks for the quick review! Testing upstream socket options in integration tests is hard overall because they usually are not reflected in any behavior change in fake upstream or client.

PR #21803, opened 2022-06-21
Always use IPv6 sockets on Android

mattklein123:
Investigation shows that Android always uses IPv6 dual stack sockets. There is a chance that some long tail connectivity issues are related to this. This change adds a runtime guard (applicable to Android only) which will force use of IPv6 sockets and translate IPv4 addresses to mapped IPv6 addresses upon connection. This is the same as what is done for Java platform sockets.

Risk Level: None, runtime guarded to false on Android.
Testing: No possible tests here. EM will provide testing.
Docs Changes: N/A
Release Notes: N/A
Runtime guard: envoy.reloadable_features.android_always_use_v6 (default false)


mattklein123:
cc @Reflejo @alyssawilk @Augustyniak

Augustyniak (approved):

source/common/network/io_socket_handle_impl.cc
```diff
@@ -63,6 +64,14 @@ constexpr int messageTruncatedOption() {
 
 namespace Network {
 
+bool IoSocketHandleImpl::alwaysUseV6OnAndroid() {
```
alyssawilk, line 67: if the end goal after runtime guards are gone is to return true on android and false elsewhere maybe forceV6ForAndroid()

source/common/network/io_socket_handle_impl.cc
```diff
@@ -466,7 +475,28 @@ IoHandlePtr IoSocketHandleImpl::accept(struct sockaddr* addr, socklen_t* addrlen
...
+    // connecting.
+    memset(&sin6, 0, sizeof(sin6));
+    sin6.sin6_family = AF_INET6;
+    sin6.sin6_port = sin4.sin_port;
+    sin6.sin6_addr.s6_addr32[2] = htonl(0xffff);
+    sin6.sin6_addr.s6_addr32[3] = sin4.sin_addr.s_addr;
+    ASSERT(IN6_IS_ADDR_V4MAPPED(&sin6.sin6_addr));
+
+    sockaddr_to_use = reinterpret_cast<sockaddr*>(&sin6);
+    sockaddr_len_to_use = sizeof(sin6);
```
alyssawilk, line 495: I think the addresses we get from sockets are now going to be of the style 
ffff:1.2.3.4 which I believe is controlled via config v6only.  Should we also default sanitize getting addresses so we strip the ffff (v6only or forceNormalizeV6ForAndroid) ?

source/common/runtime/runtime_features.cc
```diff
@@ -93,6 +93,8 @@ FALSE_RUNTIME_GUARD(envoy_reloadable_features_thrift_connection_draining);
 FALSE_RUNTIME_GUARD(envoy_reloadable_features_http2_use_oghttp2);
 // Used to track if runtime is initialized.
 FALSE_RUNTIME_GUARD(envoy_reloadable_features_runtime_initialized);
+// TODO(mattklein123): Flip this to true and/or remove completely once verified by Envoy Mobile.
+FALSE_RUNTIME_GUARD(envoy_reloadable_features_android_always_use_v6);
```
alyssawilk, line 97: Also please unit test, or add unit tests to the TODO here if the timing is too tight.

alyssawilk (commented):
yeesh, good find.  Any idea if this is true for the versions of android E-M supports?

source/common/network/io_socket_handle_impl.cc
mattklein123: I need to look and see if we ever actually pull the the peer address. I'm not sure that we do but let me look further.

source/common/runtime/runtime_features.cc
mattklein123: Do you have a suggestion of how to unit test this in this repo? This is all behind an android only ifdef.

mattklein123:
> Any idea if this is true for the versions of android E-M supports?

I haven't looked back through the Android history. It's true on main branch. I can take a look through previous versions of someone can tell me which versions to look back to.

mattklein123:
> I haven't looked back through the Android history. It's true on main branch. I can take a look through previous versions of someone can tell me which versions to look back to.

I looked and this applies all the way back to Android 6: https://cs.android.com/android/platform/superproject/+/android-6.0.1_r81:libcore/luni/src/main/native/libcore_io_Posix.cpp;l=91-128

source/common/network/io_socket_handle_impl.cc
mattklein123: So yeah we do have `Address::InstanceConstSharedPtr IoSocketHandleImpl::peerAddress()`, but I don't think we ever use this in the client context. Let me add a comment/TODO there in case this becomes an issue. I would rather not make this change more complicated than it needs to be since I don't think I can test it (see other comment).

alyssawilk:
Give connect method a boolean argument which defaults to forceV6ForAndroid() and explicitly set the boolean in test? Otherwise test upstream for E-M on android only CI?  Again I'm fine with TODO but I think it's possible to test with a little refactor work.

source/common/runtime/runtime_features.cc
mattklein123: I think I would prefer to not have a build flag for it since that will be harder to test. I think Alyssa's suggestion is good to just refactor it and add a unit test. I can do that but will do it as a follow up.

alyssawilk (approved):
I'm not super confident of the change but as it's off by default and y'all are testing it I'm totally fine letting you folks figure out if there are corner cases missed :-)  Good luck!

alyssawilk (approved):

PR #19588, opened 2022-01-18
DNS: officially (previously only unofficially) allowing null resolutions

alyssawilk:
in https://github.com/envoyproxy/envoy/pull/17645 there was a bunch of discussion around the DNS cache returning null addresses and how to handle it.  After discussion on https://github.com/envoyproxy/envoy/pull/19461 we agreed to keep sending null updates, but to fast-fail if no address was resolved.

Risk Level: Medium (data plane change)
Testing: updated integration tests, unit tests
Docs Changes: n/a
Release Notes: inline

PR #21810, opened 2022-06-21
dfp: consistent behavior on DNS resolution failure

alyssawilk:
Fixes a bug in https://github.com/envoyproxy/envoy/pull/19588 (and previously) where after DNS resolutoin fail, subsequent requests weren't blackholed by the filter.

Risk Level: low
Testing: fixed integration tests
Docs Changes: n/a
Release Notes: n/a
Fixes https://github.com/envoyproxy/envoy/issues/21410


mattklein123 (approved):
Nice find and fix!

PR #21811, opened 2022-06-21
docs: add section for running Envoy via CLI

martinezlucas98:
As a beginner I've had a tough time getting around Envoy. One of the things that got me scratching my head for a while was "how to run Envoy directly from the CLI?".

Although testing Envoy with bazel is magnificently documented (10/10 to whoever wrote it), I couldn't find much information on how to run Envoy from the CLI.

I believe this small section will be of great use for beginners, although I'm not convinced if this doc is the right one for the "Running envoy via CLI" section. What do you thing? 

> PS, in case there is already information on this topic, maybe we should make it more visible or easy to find.

Signed-off-by: Lucas Martinez <57833320+martinezlucas98@users.noreply.github.com>

### DIFF
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -272,12 +272,14 @@ for how to update or override dependencies.
 
 ## Running the built Envoy binary on the host system
 
-After Envoy is built, it can be executed via CLI.
+After Envoy is built, it can be executed via CLI. 
+
+As Bazel's current ability to provide information about where it just built the artifacts is limited (as stated in [bazel's issue #8739](https://github.com/bazelbuild/bazel/issues/8739)), we can use the `bazel info bazel-genfiles` command to get the path where Envoy was built.
 
 For example, if Envoy was built using the `bazel build -c opt //source/exe:envoy-static` command, then it can be executed from the project's root directory by running:
 
 ```console
-bazel-bin/source/exe/envoy-static --config-path /path/to/your/envoy/config.yaml
+$(bazel info bazel-genfiles)/source/exe/envoy-static --config-path /path/to/your/envoy/config.yaml
 ```
 
 ## Building Envoy with the CI Docker image
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -270,6 +270,16 @@ for how to update or override dependencies.
 1. `bazel build envoy` from the Envoy source directory. Add `-c opt` for an optimized release build or
    `-c dbg` for an unoptimized, fully instrumented debugging build.
 
+## Running Envoy via CLI
+
+After Envoy is built, it can be executed via CLI.
+
+For example, if Envoy was built using the `bazel build -c opt //source/exe:envoy-static` command, then it can be executed from the project's root directory by running:
+
+```
+bazel-bin/source/exe/envoy-static --config-path /path/to/your/envoy/config.yaml
+```
+
 ## Building Envoy with the CI Docker image
 
 Envoy can also be built with the Docker image used for CI, by installing Docker and executing the following.
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -270,13 +270,13 @@ for how to update or override dependencies.
 1. `bazel build envoy` from the Envoy source directory. Add `-c opt` for an optimized release build or
    `-c dbg` for an unoptimized, fully instrumented debugging build.
 
-## Running Envoy via CLI
+## Running the built Envoy binary on the host system
 
 After Envoy is built, it can be executed via CLI.
 
 For example, if Envoy was built using the `bazel build -c opt //source/exe:envoy-static` command, then it can be executed from the project's root directory by running:
 
-```
+```console
 bazel-bin/source/exe/envoy-static --config-path /path/to/your/envoy/config.yaml
 ```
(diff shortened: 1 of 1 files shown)


bazel/README.md
```diff
@@ -270,6 +270,16 @@ for how to update or override dependencies.
 1. `bazel build envoy` from the Envoy source directory. Add `-c opt` for an optimized release build or
    `-c dbg` for an unoptimized, fully instrumented debugging build.
 
+## Running Envoy via CLI
+
+After Envoy is built, it can be executed via CLI.
+
+For example, if Envoy was built using the `bazel build -c opt //source/exe:envoy-static` command, then it can be executed from the project's root directory by running:
```
daixiang0, line 277: If use `ci/do_ci.sh <target>` to build, the output directory is different, please add it as well.

martinezlucas98: Sure!

bazel/README.md
```diff
@@ -270,6 +270,16 @@ for how to update or override dependencies.
 1. `bazel build envoy` from the Envoy source directory. Add `-c opt` for an optimized release build or
    `-c dbg` for an unoptimized, fully instrumented debugging build.
 
+## Running Envoy via CLI
```
phlax, line 273:
```suggestion
## Running the built Envoy binary on the host system
```

bazel/README.md
```diff
@@ -270,6 +270,16 @@ for how to update or override dependencies.
...
+
+After Envoy is built, it can be executed via CLI.
+
+For example, if Envoy was built using the `bazel build -c opt //source/exe:envoy-static` command, then it can be executed from the project's root directory by running:
+
+```
```
phlax, line 279:
```suggestion
```console
```

bazel/README.md
```diff
@@ -270,6 +270,16 @@ for how to update or override dependencies.
...
+bazel-bin/source/exe/envoy-static --config-path /path/to/your/envoy/config.yaml
```
phlax, line 280: bazels ability to provide info about where it just built things is limited (cf https://github.com/bazelbuild/bazel/issues/8739) 

as a ~workaround it is possible to use `bazel info` - i believe something like the following should work (and would also hopefully address above comments)

```console

$ $(bazel info bazel-genfiles)/source/exe/envoy-static --config-path /path/to/your/envoy/config.yaml

```

phlax (changes requested):
@martinezlucas98 thanks for working on this, i agree it would be helpful to provide some further information about running the built binary - a couple of comments inline

phlax (changes requested):
@martinezlucas98 i think we want to resolve the bazel path issue - then this should be good to land

martinezlucas98:
> @martinezlucas98 i think we want to resolve the bazel path issue - then this should be good to land

Great 👍 , I just added it. Got any suggestions?

bazel/README.md
```diff
@@ -272,12 +272,14 @@ for how to update or override dependencies.
 
 ## Running the built Envoy binary on the host system
 
-After Envoy is built, it can be executed via CLI.
+After Envoy is built, it can be executed via CLI. 
+
+As Bazel's current ability to provide information about where it just built the artifacts is limited (as stated in [bazel's issue #8739](https://github.com/bazelbuild/bazel/issues/8739)), we can use the `bazel info bazel-genfiles` command to get the path where Envoy was built.
```
phlax, line 277: i dont think we want to include this message - sorry if i was unclear - this was a message to you rather than a suggestion

martinezlucas98: Got it, i'll make the changes

martinezlucas98: What about the line:

>  we can use the `bazel info bazel-genfiles` command to get the path where Envoy was built

Do you think it's worth mentioning? Or seems unnecessary

phlax: on balance, probably unnecessary

bazel/README.md
```diff
@@ -272,9 +272,7 @@ for how to update or override dependencies.
 
 ## Running the built Envoy binary on the host system
 
-After Envoy is built, it can be executed via CLI. 
-
-As Bazel's current ability to provide information about where it just built the artifacts is limited (as stated in [bazel's issue #8739](https://github.com/bazelbuild/bazel/issues/8739)), we can use the `bazel info bazel-genfiles` command to get the path where Envoy was built.
+After Envoy is built, it can be executed via CLI.
```
phlax, line 275: this sentence and one below can be merged i think

phlax: actually nm - i think its fine

phlax (commented):
one last ask - would you mind moving this section below all of the build instructions - just above "Testing Envoy with Bazel"

probs you should promote the header also - i think its kinda its own section

martinezlucas98:
> one last ask - would you mind moving this section below all of the build instructions - just above "Testing Envoy with Bazel"
> 
> probs you should promote the header also - i think its kinda its own section

I was  also thinking about that 🤔, I''ll update it

phlax (approved):
great, lgtm, thanks @martinezlucas98

Issue #18675, opened 2021-10-19
stats: reduce browser and server memory/cpu bursts when viewing stats from admin 


jmarantz:
See related issues:
  * https://github.com/envoyproxy/envoy/issues/16981
  * https://github.com/envoyproxy/envoy/issues/16139

There are a few different cases to consider when dealing with aggregating stats when there are a large number of clusters. I don't think there is a single solution that's ideal for all of them. But I think for for the admin endpoint we should add endpoints help view stats in a single scope. So if someone has 10k clusters they can pick which cluster to look at and then bring that one in.

Any of the solutions that involve iterating over all the stats are going to involve a burst of CPU at least on the main thread, likely while holding the stats allocator lock for a long time.

A prototype scoped access solution exists in 
https://github.com/envoyproxy/envoy/pull/18670

jmarantz:
@pradeepcrao @daixiang0 FYI

PR #18670, opened 2021-10-19
RFC admin: hierarchical stats viewer

jmarantz:
Commit Message: Adds default hierarchical view for admin stats accessed by clicking 'stats' from the admin home page. Note that the existing /stats API is unchanged.

Screen Shot 2022-01-12 at 12 28 52 PM

This is not ready for approval/submit yet as the new endpoints lack unit tests, and the UI (if we keep it) is ugly (unstyled). But I wanted to get some comments on the concept, particularly as it relates to solving #18675 before I clean all that up.

Also needed before this PR can be merged:
 * proper escaping of special chars embedded in scope names, wherever arbitrary text is inserted via C++ or JS.
 * When navigating within the outline, record HTML history
 * Add JS tests of some sort
 * Review Scope lifetime and make sure we don't risk referencing scopes as they are being deleted by a config update
 * When "usedonly" mode is on, find some mechanism to filter out scopes with no used stats. Same for regex filtering.
 * Break up into about 3 PRs.
 * Add option declarations for other admin pages besides stats.
 * test coverage for stats_hander.cc and admin_html_generator.cc.
 
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
Fixes: #18675

PR #21814, opened 2022-06-22
admin: make admin HTML a compile-time option

jmarantz:
Commit Message: Make the HTML view for stats a compile-time option, defaulting to on. This PR doesn't change any code-paths, but it moves the HTML support into a separate file, compiled into //source/server/admin:admin_lib. In the future, and admin_html_lib will be added, built only when admin-html is not disabled. That will allow https://github.com/envoyproxy/envoy/pull/19546 and ultimately https://github.com/envoyproxy/envoy/pull/18670 to proceed.

To disable, compile with `--define=admin_html=disabled`, in which case the admin `"/"` endpoint will just print that the thml mode was disabled.

Additional Description: This PR is a step toward solving https://github.com/envoyproxy/envoy/issues/18675
Risk Level: low -- this just moves code around and doesn't change it at all.
Testing: //test/..., both with and without `--define=admin_html=disabled`
Docs Changes: included
Release Notes: included

### DIFF
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -319,6 +319,11 @@ selects.config_setting_group(
 )
 
 config_setting(
+    name = "disable_admin_html",
+    values = {"define": "admin_html=disabled"},
+)
+
+config_setting(
     name = "disable_hot_restart_setting",
     values = {"define": "hot_restart=disabled"},
     visibility = ["//visibility:private"],
--- a/source/server/admin/admin_html.cc
+++ b/source/server/admin/admin_html.cc
@@ -1,6 +1,5 @@
-#include "source/server/admin/admin.h"
-
 #include "source/common/html/utility.h"
+#include "source/server/admin/admin.h"
 
 namespace Envoy {
 namespace Server {
--- a/source/server/admin/admin_html.cc
+++ b/source/server/admin/admin_html.cc
@@ -1,0 +1,123 @@
+#include "source/server/admin/admin.h"
+
+#include "source/common/html/utility.h"
+
+namespace Envoy {
+namespace Server {
+
+namespace {
+
+/**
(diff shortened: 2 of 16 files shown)


test/integration/integration_admin_test.cc
```diff
@@ -138,8 +138,13 @@ TEST_P(IntegrationAdminTest, Admin) {
   EXPECT_THAT(response->body(), HasSubstr("admin commands are:"));
 
   EXPECT_EQ("200", request("admin", "GET", "/", response));
+#ifdef ENVOY_ADMIN_HTML
```
jmarantz, line 141: note: in other cases I did not need a `-D` for this, but it seemed painful to tweak this particular test without it.

I have this gut feel that this define is costly from a build perspective since it is passed on every compile-command, even though it's needed only by one test.

LMK if you think I should re-work this somehow.

source/server/admin/admin_html.cc
```diff
@@ -0,0 +1,122 @@
+#include "source/common/html/utility.h"
```
jmarantz, line 1: note to reviewers: this code is unchanged; just moved.

bazel/BUILD
```diff
@@ -318,6 +318,11 @@ selects.config_setting_group(
     ],
 )
 
+config_setting(
+    name = "disable_admin_html",
+    values = {"define": "admin_html=disabled"},
+)
+
```
mattklein123, line 325: Can you document this here? https://github.com/envoyproxy/envoy/tree/main/bazel#disabling-optional-features

Probably also worth a release note?

mattklein123 (changes requested):
Thanks LGTM with a small comment. Thank you!

/wait

mattklein123 (approved):
Nice, thanks!

jmarantz:
Thanks Matt!

PR #21818, opened 2022-06-22
examples: fix gzip decompression

giantcroc:
Signed-off-by: giantcroc <changran.wang@intel.com>

Commit Message:
Due to [quickfix of decompression oom](https://github.com/envoyproxy/envoy/commit/d4c39e635603e2f23e1e08ddecf5a5fb5a706338), `decompressed-size` will be truncated at around the size of file.gz * `MaxInflateRatio` . And we could not get fixed `decompressed-size` which is approximately 1130000. 
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/docs/root/start/sandboxes/gzip.rst
+++ b/docs/root/start/sandboxes/gzip.rst
@@ -14,11 +14,12 @@ By enabling compression in Envoy you can save some network bandwidth, at the exp
 
 Envoy supports compression and decompression for both requests and responses.
 
-This sandbox provides an example of response compression served over ``HTTP``. Although ``HTTPS`` is not demonstrated, compression can be used for this also.
+This sandbox provides an example of response compression and request decompression served over ``HTTP``. Although ``HTTPS`` is not demonstrated, compression can be used for this also.
 
-The sandbox covers two scenarios:
+The sandbox covers three scenarios:
 
 - compression of files from an upstream server
+- decompression of files from a downstream client
 - compression of Envoy's own statistics
 
 Step 1: Start all of our containers
@@ -63,7 +64,29 @@ As only files with a content-type of ``application/json`` are configured to be g
 
     $ curl -si -H "Accept-Encoding: gzip" localhost:10000/file.txt | grep "content-encoding"
 
-Step 3: Test compression of Envoy’s statistics
+Step 3: Test Envoy’s decompression of downstream files
+******************************************************
+
+The sandbox is configured with one endpoint on port ``10000`` for uploading downstream files:
+
+- ``/upload``
+
+Use ``curl`` to get compressed file ``file.gz``
+
+.. code-block:: console
+
+    $ curl -s -H "Accept-Encoding: gzip" -o file.gz localhost:10000/file.json
+
+Use ``curl`` to check that the response from uploading ``file.gz`` contains the ``decompressed-size`` header.
+
+You will need to add an ``content-encoding: gzip`` request header.
+
+.. code-block:: console
+
+    $ curl -si -H "Content-Encoding: gzip" localhost:10000/upload --data-binary "@file.gz" | grep "decompressed-size"
+    decompressed-size: 10485760
+
+Step 4: Test compression of Envoy’s statistics
 **********************************************
 
 The sandbox is configured with two ports serving Envoy’s admin and statistics interface:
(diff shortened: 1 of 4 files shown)


giantcroc:
@phlax Hi, I have modified `verify.sh` to make it works.

phlax:
@giantcroc the bit im not clear about is why we are seeing a different decompressed size - surely it should decompress to the original size ?

is there a way we can still verify what we have uploaded as well as that we did with compression ?

giantcroc:
> @giantcroc the bit im not clear about is why we are seeing a different decompressed size - surely it should decompress to the original size ?
> 
> is there a way we can still verify what we have uploaded as well as that we did with compression ?

em.. Because compressed data size is not fixed every time, and [quickfix of decompression oom](https://github.com/envoyproxy/envoy/commit/d4c39e635603e2f23e1e08ddecf5a5fb5a706338)'s logic is using 100*compressed data size to limit output size. So we can't get fixed decompressed data size. Maybe we need more flexible way to limit output size.

phlax:
is the upload failing when this happens ?

giantcroc:
> is the upload failing when this happens ?

Yeah, I think so.

phlax:
hmm - in which case - i think we definitely need to address the issue that is causing the breakage

the example/tests are kinda confusing already i think in terms of files/names/mimetypes etc - i started trying to address this previously here #20735 but didnt so far have time to work on it further

im wondering if in terms of making this example work what we would need to change about the upload file - size ? compression ratio ?

but, im also wondering whether the example could/should demonstrate that this is what would happen given "bad" data

phlax:
@giantcroc i guess this one can progress now - could you merge and update

giantcroc:
> @giantcroc i guess this one can progress now - could you merge and update

OK, I will fix it soon.

giantcroc:
@phlax Hi, I find that `max_inflate_ratio` doesn't work yet due to `envoyproxy/envoy-dev:latest` has not been updated.

phlax:
>  doesn't work yet 

it shouldnt matter - CI tests the image/binary created in CI

phlax:
> envoyproxy/envoy-dev:latest has not been updated.

also, it is continually updated - and has been several times since that PR landed

giantcroc:
@phlax Hi, it seems ready to be merged.

examples/gzip/gzip-envoy.yaml
```diff
@@ -37,6 +37,21 @@ static_resources:
...
+          - name: envoy.filters.http.decompressor
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.http.decompressor.v3.Decompressor
+              decompressor_library:
+                name: upload
+                typed_config:
+                  "@type": "type.googleapis.com/envoy.extensions.compression.gzip.decompressor.v3.Gzip"
+                  window_bits: 9
+                  chunk_size: 8192
+                  max_inflate_ratio: 1000
```
phlax, line 49: i think we should probably say something about the decompression ratio - as we know - if this is set too low (for our expectations) then it will fail

docs/root/start/sandboxes/gzip.rst
```diff
@@ -63,7 +64,29 @@ As only files with a content-type of ``application/json`` are configured to be g
...
+
+Use ``curl`` to get compressed file ``file.gz``
+
+.. code-block:: console
+
+    $ curl -s -H "Accept-Encoding: gzip" -o file.gz localhost:10000/file.json
+
+Use ``curl`` to check that the response from uploading ``file.gz`` contains the ``decompressed-size`` header.
+
+You will need to add an ``content-encoding: gzip`` request header.
```
phlax, line 82:
```suggestion
You will need to add the ``content-encoding: gzip`` request header.
```

docs/root/start/sandboxes/gzip.rst
```diff
@@ -63,7 +64,29 @@ As only files with a content-type of ``application/json`` are configured to be g
...
 
-Step 3: Test compression of Envoy’s statistics
+Step 3: Test Envoy’s decompression of downstream files
+******************************************************
+
+The sandbox is configured with one endpoint on port ``10000`` for uploading downstream files:
+
+- ``/upload``
+
+Use ``curl`` to get compressed file ``file.gz``
```
phlax, line 74:
```suggestion
Use ``curl`` to get the compressed file ``file.gz``
```

docs/root/start/sandboxes/gzip.rst
```diff
@@ -14,11 +14,12 @@ By enabling compression in Envoy you can save some network bandwidth, at the exp
 
 Envoy supports compression and decompression for both requests and responses.
 
-This sandbox provides an example of response compression served over ``HTTP``. Although ``HTTPS`` is not demonstrated, compression can be used for this also.
+This sandbox provides an example of response compression and request decompression served over ``HTTP``. Although ``HTTPS`` is not demonstrated, compression can be used for this also.
```
phlax, line 17:
```suggestion
This sandbox provides examples of response compression and decompression served over ``HTTP``. Although ``HTTPS`` is not demonstrated, compression can be used for this also.
```

docs/root/start/sandboxes/gzip.rst
```diff
@@ -63,7 +64,29 @@ As only files with a content-type of ``application/json`` are configured to be g
 
     $ curl -si -H "Accept-Encoding: gzip" localhost:10000/file.txt | grep "content-encoding"
 
-Step 3: Test compression of Envoy’s statistics
+Step 3: Test Envoy’s decompression of downstream files
+******************************************************
+
+The sandbox is configured with one endpoint on port ``10000`` for uploading downstream files:
```
phlax, line 70:
```suggestion
The sandbox is configured with an endpoint for uploading downstream files:
```

phlax (commented):
looking good @giantcroc 

a few nits inline

examples/gzip/gzip-envoy.yaml
giantcroc: OK, I added a comment in yaml file.Thanks!

docs/root/start/sandboxes/gzip.rst
giantcroc: Done, thanks!

docs/root/start/sandboxes/gzip.rst
giantcroc: Done, thanks!

docs/root/start/sandboxes/gzip.rst
giantcroc: Done, thanks!

docs/root/start/sandboxes/gzip.rst
giantcroc: Done, thanks!

phlax (approved):
lgtm, thanks @giantcroc for following up on this, and resolving the gzip decompression issues

PR #21821, opened 2022-06-22
bazel: Use experimental flags for fetching coverage output

phlax:

Commit Message:
Additional Description:

@lizan this is more of a question than anything

trying to track down why coverage is so much slower in postsubmit then presubmit and it would appear that its because postsubmit is set with `--nocache-test-results`

not sure why this is, whether it makes any difference if it they never pass/fail independently, and whether we can remove this flag

i saw some discussion of using this flag to workaround rbe not giving back all test results so wondered if this was related - hence proposed change, but im not sure how helpful that is

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/build_setup.sh
+++ b/ci/build_setup.sh
@@ -86,7 +86,9 @@ trap cleanup EXIT
 
 "$(dirname "$0")"/../bazel/setup_clang.sh "${LLVM_ROOT}"
 
-[[ "${BUILD_REASON}" != "PullRequest" ]] && BAZEL_EXTRA_TEST_OPTIONS+=("--nocache_test_results")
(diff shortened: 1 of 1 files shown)


phlax:
thanks - i think i understand

> If we're going to do a daily run on main and/or release branches

i was already thinking of adding this for the release branches so we can pick up problems like the ones resolved recently with unpinned deps bitrotting

i guess my main reservation with scheduled tests is that they can be more easily ignored, altho not sure iiuc how caching works here so maybe that is not an issue in this case

ill close this  as it was mostly raised to try and clarify the situation and think about setting up scheduled jobs

PR #21850, opened 2022-06-23
[ext_proc]: Refactor ext proc timeout handling

chaoqin-li1123:
Commit Message: Move setCallbackState and startTimer or disableTimer into the same function call because callback_state and timer have the same lifetime. This will make the state machine management more consistent and easier to reason about.
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/http/ext_proc/processor_state.cc
+++ b/source/extensions/filters/http/ext_proc/processor_state.cc
@@ -301,7 +301,7 @@ void ProcessorState::enqueueStreamingChunk(Buffer::Instance& data, bool end_stre
 }
 
 void ProcessorState::clearAsyncState() {
-  cleanUpTimer();
+  onFinishCall();
   while (auto queued_chunk = dequeueStreamingChunk(false)) {
     auto chunk = std::move(*queued_chunk);
     ENVOY_LOG(trace, "Injecting leftover buffer of {} bytes", chunk->data.length());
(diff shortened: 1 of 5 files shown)


source/extensions/filters/http/ext_proc/processor_state.cc
```diff
@@ -18,13 +18,22 @@ using envoy::service::ext_proc::v3::CommonResponse;
 using envoy::service::ext_proc::v3::HeadersResponse;
 using envoy::service::ext_proc::v3::TrailersResponse;
 
-void ProcessorState::startMessageTimer(Event::TimerCb cb, std::chrono::milliseconds timeout) {
+void ProcessorState::onStartCall(Event::TimerCb cb, std::chrono::milliseconds timeout,
```
snowp, line 21: Maybe name this (and the other function) super clear that we're talking about a call to the processor? e.g. `startProcessorCall`, `onStartProcessorCall` etc

snowp (changes requested):
Nice, seems like an improvement. One possible suggestion, but general lgtm

source/extensions/filters/http/ext_proc/processor_state.cc
chaoqin-li1123: Thanks, I have renamed the method.

snowp (approved):
lgtm, thanks!

PR #21852, opened 2022-06-23
Allow FileSystemBufferFilter to be bypassed without creating a thread

ravenblackx:
Commit Message: Allow FileSystemBufferFilter to be bypassed without creating a thread
Additional Description: FileSystemBufferFilter requires that an AsyncFileManager be configured; this change makes a single exception to that rule, if the filter is configured to do nothing then the AsyncFileManager can be omitted. This allows for dynamic configuration of a FileSystemBufferFilter to not spin up an unused file-management thread.
Risk Level: Low (filter is still experimental, change only makes a previously invalid behavior valid)
Testing: Added test coverage.
Docs Changes: n/a
Release Notes: n/a


mattklein123 (approved):
Thanks!

Issue #21828, opened 2022-06-22
Header Invert Match Rule Does Not Work When Header Does Not Exist


yuhaoliu77:
*Description*:
In https://github.com/envoyproxy/envoy/blob/154a7a865c064870f500a11d2772b2ca106337ac/source/common/http/header_utility.cc#L140-L146, the code shows that when header does not exist, Envoy would ignore any header match rule.

However, when there is a invert match rule, the header match rule could still match even when the header does not exist.

PR #21859, opened 2022-06-23
Add "treat_missing_header_as_empty" API Field

yuhaoliu77:
~~For invert header match rule with any header match type other than present, and the header is not present, always consider it as a match.~~

Addingdd an API field `treat_missing_header_as_empty` for better compatibility (both backward and to other xDS clients like gRPC) and potential to expand (not restricted to `invert_match`)

Signed-off-by: Yuhao Liu <yuhaoliu@google.com>

Risk Level: Low
Testing: Unit test
Docs Changes: inline
Release Notes: inline
[Optional Fixes #Issue] #21828


source/common/http/header_utility.cc
```diff
@@ -139,7 +139,10 @@ bool HeaderUtility::matchHeaders(const HeaderMap& request_headers, const HeaderD
 
   if (!header_value.result().has_value()) {
     if (header_data.invert_match_) {
-      return header_data.header_match_type_ == HeaderMatchType::Present && header_data.present_;
+      // For invert match with any match type but present, it should be considered as a match when
+      // the header is not present
+      return header_data.header_match_type_ == HeaderMatchType::Present
+                 ? header_data.present_ : true;
```
soulxu, line 145: I guess we need runtime flag to protect this.

And it will be great to update the document to describe the behavior
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-headermatcher

We can update the `invert_match_` example, to clarify this case.
```
invert_match
([bool](https://developers.google.com/protocol-buffers/docs/proto#scalar)) If specified, the match result will be inverted before checking. Defaults to false.

Examples:

The regex \d{3} does not match the value 1234, so it will match when inverted.

The range [-10,0) will match the value -1, so it will not match when inverted.
```

test/common/http/header_utility_test.cc
```diff
@@ -416,6 +417,7 @@ invert_match: true
   header_data.push_back(
       std::make_unique<HeaderUtility::HeaderData>(parseHeaderMatcherFromYaml(yaml)));
   EXPECT_TRUE(HeaderUtility::matchHeaders(matching_headers, header_data));
+  EXPECT_TRUE(HeaderUtility::matchHeaders(matching_empty_headers, header_data));
```
soulxu, line 420: Do we have same test but with `invert_match: false`?

soulxu (changes requested):
nice catch, few comments, thanks!

soulxu:
@envoyproxy/api-shepherds  in case of this API change need to be discussed, seems like this undefined behavior in the past.

bryanwux:
Please fix the format. I suggest you running the format script `./tools/code_format/check_format.py fix && ./tools/code_format/format_python_tools.sh fix` according to https://github.com/envoyproxy/envoy/blob/main/support/README.md

alyssawilk:
/wait on CI and review comments :-)

test/common/http/header_utility_test.cc
yuhaoliu77: No. Added them in the latest commit. Thank for pointing it out!

htuch:
I think this makes sense based on what was implied by the existing field comments, but I'd like to check with @markdroth @dfawley @ejona86 that the gRPC set of clients have/want the same semantics here.

soulxu:
> Yes for the last commit I ran the `./ci/run_envoy_docker.sh './ci/do_ci.sh fix_proto_format` script. Does it fix all the format issues?
> 
> Update: I think I found why the script didn't work. Let's try it again

@yuhaoliu77 It failed due to runtime flag required the a changelog https://dev.azure.com/cncf/envoy/_build/results?buildId=111549&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39&l=64

you can add changelog here https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml

source/common/http/header_utility.cc
```diff
@@ -139,7 +139,16 @@ bool HeaderUtility::matchHeaders(const HeaderMap& request_headers, const HeaderD
 
   if (!header_value.result().has_value()) {
     if (header_data.invert_match_) {
-      return header_data.header_match_type_ == HeaderMatchType::Present && header_data.present_;
+      if (Runtime::runtimeFeatureEnabled(
+              "envoy.reloadable_features.return_match_when_invert_header_match_"
+              "and_header_is_not_present")) {
```
soulxu, line 144: maybe name it as `apply_invert_match_when_header_not_present_and_non_present_match_type`. But I'm not good at naming, just a suggestion, and we can see others' opinions here.

yuhaoliu77:
> @yuhaoliu77 It failed due to runtime flag required the a changelog https://dev.azure.com/cncf/envoy/_build/results?buildId=111549&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39&l=64
> 
> you can add changelog here https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml

Ah, thanks for the pointer!

soulxu:
Thanks for the update! This LGTM, waiting for the maintainer and API people to feedback.

soulxu (dismissed):

changelogs/current.yaml
```diff
@@ -116,6 +116,11 @@ minor_behavior_changes:
 - area: logging
   change: |
     changed category name for access log filter extensions to ``envoy.access_loggers.extension_filters``.
+- area: http
+  change: |
+    header invert match rule will apply when the header is not present and the header match rule is not present type.
+    This behavior change can be reverted by setting runtim guard
```
alyssawilk, line 122: runtim -> runtime

also I'd be inclined to put this in the bugfix section

ejona86:
(I see Mark has replied while I wrote this. I've written this in parallel.)

Given that HeaderMatch is used in RBAC, it seems inaccurate to label this as "Risk Level: Low".

gRPC [duplicated Envoy's current behavior](https://github.com/grpc/proposal/blob/master/A41-xds-rbac.md):
> If a header is not present, HeaderMatch will not match except for present_match when present_match == invert_match. This is because HeaderMatcher.invert_match inverts the comparison operation so something like exact_match changes from an == comparison to a != comparison, and both fail to match if the header is not present.

I wouldn't make this change willy-nilly. This is the sort of thing that the expected behavior can differ per-person and even per-situation. This is a common problem with ternary logic (I'm looking at you SQL and NaN). I wasn't necessarily a fan of the current behavior, but it is conceptually consistent.

I will note that with either interpretation of invert the matcher is equally powerful as you can do a header presence check before whatever match operation. This is more about convenience and expectations.

changelogs/current.yaml
yuhaoliu77: Thank you! Fixed the changelog.

yuhaoliu77:
> I suggest the following alternative: Leave the behavior of `invert_match` as-is, but add the following new field:
> 
> ```
> // If true, a missing header will be treated as if it was present and its value is the empty string.
> bool treat_missing_as_empty = N;
> ```
> 
> I think this gives us the best of both worlds:
> 
> * If `treat_missing_as_empty` is false (the default), we continue to get the current behavior: If the header is not present, then `invert_match` is used only for the `present_match` matcher; all other matchers return false, regardless of the value of `invert_match`.
> * If `treat_missing_as_empty` is true, then we get the new behavior: We treat the header as present with value being the empty string, and then apply the matcher (regardless of type) to that value, and then we apply `invert_match` to that result.
> 
> I think the only real difference here is that if the matcher does actually match the empty string (e.g., a regex match allowing the empty string), then the matcher will match and `invert_match` will make it not match, but that may actually be a feature instead of a bug.
> 
> WDYT?

I think this is a great idea. This way we can keep the default behavior to make it backward compatible, also allows easy switching to the new behavior. So I plan to go ahead and make a change to use a feature flag called `treat_missing_header_as_empty_for_invert_match`, and make it default to disabled. WDYT?

markdroth:
> So I plan to go ahead and make a change to use a feature flag called `treat_missing_header_as_empty_for_invert_match`, and make it default to disabled. WDYT?

To be clear, I think this should be a new API field, not a runtime flag.  Note that runtime is an Envoy-specific concept; xDS clients like gRPC do not have any equivalent, but we'll presumably want this new behavior to be possible for all clients.  Also, it's important for clients to know what a control plane is instructing them to do, which means that the control plane needs a way to encode its intent in the xDS resources it sends.

Also, I think the field should not be specific to `invert_match`.  I think it should just be `treat_missing_as_empty`, and if true, it should be honored even if `invert_match` is false.  That way, it can also be used to (e.g.) use a regex match that matches against either an empty value or a specific value.

soulxu:
> > So I plan to go ahead and make a change to use a feature flag called `treat_missing_header_as_empty_for_invert_match`, and make it default to disabled. WDYT?
> 
> To be clear, I think this should be a new API field, not a runtime flag. Note that runtime is an Envoy-specific concept; xDS clients like gRPC do not have any equivalent, but we'll presumably want this new behavior to be possible for all clients. Also, it's important for clients to know what a control plane is instructing them to do, which means that the control plane needs a way to encode its intent in the xDS resources it sends.
> 

Thanks, good to learn this!

Also checks the doc of headers again:

`A match will happen if all the headers in the route are present in the request with the same values (or based on presence if the value field is not in the config).`
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routematch

It seems like the current behavior is documented if I didn't parse it wrong.

So the envoy API doc should be the truth of the source. For anything is not document or undefined in the API doc, we shouldn't use the runtime flag. We only should use runtime flag when the envoy implements a wrong behavior compared to the API doc. Did I understand that correctly?

> Also, I think the field should not be specific to `invert_match`. I think it should just be `treat_missing_as_empty`, and if true, it should be honored even if `invert_match` is false. That way, it can also be used to (e.g.) use a regex match that matches against either an empty value or a specific value.

For both `treat_missing_as_empty` and `present_match` are true, I guess it will consider the header always present, right?

yuhaoliu77:
> For both `treat_missing_as_empty` and `present_match` are true, I guess it will consider the header always present, right?

This is a good point. If we strictly follow the definition of `treat_missing_as_empty`, any header would be present so it is always a `present_match`? Thoughts on the preference here?

yuhaoliu77:
I just committed a version that strictly follows the description of "treat_missing_header_as_empty", which means the header will always be considered as "present" when this option is set to true. Let me know your thoughts and we can change the behavior if needed.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -2247,6 +2247,21 @@ message HeaderMatcher {
...
+  // Examples:
+  //
+  // * The header match rule specified header "header1" to range match of [0,
+  //   10], and invert_match is set to true; The "header1" header is not present.
+  //   The empty header does not match the range, so it will match when
+  //   inverted.
+  // * The header match rule specified header "header2" to a string regex match
+  //   ``^$`` which means an empty string; The "header2" header is not present.
+  //   The match rule will treat the "header2" header as an empty header so it
+  //   will match.
```
soulxu, line 2263: We better to clarify the current behavior (as `treat_missing_header_as_empty` = false) too.

soulxu (commented):
LGTM, only a comment about the doc.

soulxu:
Good to have @markdroth help to ensure the API again.

alyssawilk:
cc @htuch for API pass
/retest

api/envoy/config/route/v3/route_components.proto
```diff
@@ -2247,6 +2247,21 @@ message HeaderMatcher {
...
   // * The range [-10,0) will match the value -1, so it will not match when inverted.
   bool invert_match = 8;
+
+  // If specified, for any header match rule, if the header match rule specified header
+  // does not exist, this header value will be treated as empty. Defaults to false.
+  //
+  // Examples:
+  //
+  // * The header match rule specified header "header1" to range match of [0,
+  //   10], and invert_match is set to true; The "header1" header is not present.
```
htuch, line 2257: Nit: RST link to `invert_match`

api/envoy/config/route/v3/route_components.proto
```diff
@@ -2247,6 +2247,21 @@ message HeaderMatcher {
...
+  //   The empty header does not match the range, so it will match when
+  //   inverted.
```
htuch, line 2259: It would be useful to point out how this behaves in both the invert and non-inverted case when `treat_missing_header_as_empy` is missing. It's pretty subtle for the reader to infer.

htuch (commented):
API looks good, just a couple of docs nits.

htuch:
@markdroth can you take a look from gRPC side and LGTM if this makes sense?

api/envoy/config/route/v3/route_components.proto
yuhaoliu77: Thanks! Added more API doc to include the false case

api/envoy/config/route/v3/route_components.proto
yuhaoliu77: Added the link to invert_match.

api/envoy/config/route/v3/route_components.proto
yuhaoliu77: PTAL at the new doc to see if it is good to go. Thankyou!

alyssawilk (commented):
Looks good from a code perspective.  Please do add a release note referencing the new field and you're good to go from my end!
/wait

markdroth:
/lgtm api

Thanks, this looks great!

alyssawilk (dismissed):
LGTM, ci willing :-)

htuch (commented):
/lgtm api

htuch (approved):
/lgtm api

PR #21884, opened 2022-06-24
Only call ares_library_init()/cleanup() routines in process_wide() if ares extension used in DNS resolving.

yanjunxiang-google:
Only call ares_library_init()/cleanup() routines in process_wide() if ares extension is used in DNS resolving.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -46,8 +46,8 @@ public:
    * @returns the DNS Resolver factory.
    * @param typed_dns_resolver_config: the typed DNS resolver config
    */
-  static Network::DnsResolverFactory& createFactory(
-      const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config);
+  static Network::DnsResolverFactory&
+  createFactory(const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config);
 
   /**
    * Terminate the DNS resolver factory.
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -42,17 +42,17 @@ public:
   virtual void terminate() {}
 
   /**
-   * Find the DNS resolver factory based on the typed config and initialize it.
+   * Create the DNS resolver factory based on the typed config and initialize it.
    * @returns the DNS Resolver factory.
    * @param typed_dns_resolver_config: the typed DNS resolver config
    */
-  static Network::DnsResolverFactory& initializeDnsResolverFactory(
+  static Network::DnsResolverFactory& createFactory(
       const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config);
 
   /**
-   * Terminate the DNS resolver factories.
+   * Terminate the DNS resolver factory.
    */
-  static void terminateDnsResolverFactories();
+  static void terminateFactory();
 };
 
 } // namespace Network
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -26,7 +26,19 @@ public:
       const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) const PURE;
 
   std::string category() const override { return std::string(DnsResolverCategory); }
+
+  /*
+   * Initialize the related data for this type of DNS resolver.
+   * For some DNS resolvers, like c-ares, there are some specific data structure
+   * needs to be initialized before using it to resolve target.
+   */
   virtual void init() {}
+
+  /*
+   * Cleanup the related data for this type of DNS resolver.
+   * For some DNS resolvers, like c-ares, there are some specific data structure
+   * needs to be cleaned up before terminates Envoy.
+   */
   virtual void cleanup() {}
 };
 
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -42,5 +42,14 @@ public:
   virtual void terminate() {}
 };
 
+/**
+ * Terminate the DNS resolver factories.
+ */
+static inline void terminateDnsResolverFactories() {
+  auto& factories = Registry::FactoryRegistry<Network::DnsResolverFactory>::factories();
+  std::for_each(factories.begin(), factories.end(),
+                [](auto& factory_it) { factory_it.second->terminate(); });
+}
+
 } // namespace Network
 } // namespace Envoy
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -26,6 +26,8 @@ public:
       const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) const PURE;
 
   std::string category() const override { return std::string(DnsResolverCategory); }
+  virtual void init() {}
+  virtual void cleanup() {}
 };
 
 } // namespace Network
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -26,7 +26,6 @@ public:
       const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) const PURE;
 
   std::string category() const override { return std::string(DnsResolverCategory); }
-  virtual void init() {}
   virtual void cleanup() {}
 };
 
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -40,16 +40,20 @@ public:
    * needs to be cleaned up before terminates Envoy.
    */
   virtual void terminate() {}
-};
 
-/**
- * Terminate the DNS resolver factories.
- */
-static inline void terminateDnsResolverFactories() {
-  auto& factories = Registry::FactoryRegistry<Network::DnsResolverFactory>::factories();
-  std::for_each(factories.begin(), factories.end(),
-                [](auto& factory_it) { factory_it.second->terminate(); });
-}
+  /**
+   * Find the DNS resolver factory based on the typed config and initialize it.
+   * @returns the DNS Resolver factory.
+   * @param typed_dns_resolver_config: the typed DNS resolver config
+   */
+  static Network::DnsResolverFactory& initializeDnsResolverFactory(
+      const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config);
+
+  /**
+   * Terminate the DNS resolver factories.
+   */
+  static void terminateDnsResolverFactories();
+};
 
 } // namespace Network
 } // namespace Envoy
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -50,9 +50,9 @@ public:
   createFactory(const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config);
 
   /**
-   * Terminate the DNS resolver factory.
+   * Call the terminate method on all the registered DNS resolver factories.
    */
-  static void terminateFactory();
+  static void terminateFactories();
 };
 
 } // namespace Network
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -15,7 +15,7 @@ constexpr absl::string_view DnsResolverCategory = "envoy.network.dns_resolver";
 
 class DnsResolverFactory : public Config::TypedFactory {
 public:
-  /*
+  /**
    * @returns a DnsResolver object.
    * @param dispatcher: the local dispatcher thread
    * @param api: API interface to interact with system resources
@@ -27,19 +27,19 @@ public:
 
   std::string category() const override { return std::string(DnsResolverCategory); }
 
-  /*
+  /**
    * Initialize the related data for this type of DNS resolver.
    * For some DNS resolvers, like c-ares, there are some specific data structure
    * needs to be initialized before using it to resolve target.
    */
-  virtual void init() {}
+  virtual void initialize() {}
 
-  /*
+  /**
    * Cleanup the related data for this type of DNS resolver.
    * For some DNS resolvers, like c-ares, there are some specific data structure
    * needs to be cleaned up before terminates Envoy.
    */
-  virtual void cleanup() {}
+  virtual void terminate() {}
 };
 
 } // namespace Network
--- a/envoy/network/dns_resolver.h
+++ b/envoy/network/dns_resolver.h
@@ -26,6 +26,7 @@ public:
       const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) const PURE;
 
   std::string category() const override { return std::string(DnsResolverCategory); }
+  virtual void init() {}
   virtual void cleanup() {}
 };
 
--- a/source/common/config/utility.h
+++ b/source/common/config/utility.h
@@ -241,17 +241,6 @@ public:
                                    POOL_HISTOGRAM(scope))};
   }
 
-
-
-  /**
-   * Get the hash map for this factory.
-   */
-  template <class Factory>
-  static absl::flat_hash_map<std::string, Factory*>&  getFactoryMap() {
-    return Registry::FactoryRegistry<Factory>::factories();
-  }
-
-
   /**
    * Get a Factory from the registry with a particular name (and templated type) with error checking
    * to ensure the name and factory are valid.
--- a/source/common/config/utility.h
+++ b/source/common/config/utility.h
@@ -241,6 +241,17 @@ public:
                                    POOL_HISTOGRAM(scope))};
   }
 
+
+
+  /**
+   * Get the hash map for this factory.
+   */
+  template <class Factory>
+  static absl::flat_hash_map<std::string, Factory*>&  getFactoryMap() {
+    return Registry::FactoryRegistry<Factory>::factories();
+  }
+
+
   /**
    * Get a Factory from the registry with a particular name (and templated type) with error checking
    * to ensure the name and factory are valid.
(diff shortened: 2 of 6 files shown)


jmarantz:
per IM I am not seeing how this can work.

yanjunxiang-google:
> per IM I am not seeing how this can work.

With c-ares library is an extension now, this change helps fixing a bug that if someone build Envoy with  c-ares is not linked in here, : https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/extensions/extensions_build_config.bzl#L340,  the ares_library_init()/ares_library_cleanup() should not be called.

This is not an issue before since ares library is in core.  It is a minor issue now since ares extension is by default included.

yanjunxiang-google:
> per IM I am not seeing how this can work.

Thanks for the comments. Please take a look the latest fix.

The latest fix will only call ares_library_init() when the createDnsResolver() is called for c-ares. And it will only call ares_library_cleanup() during ~process_wide() when c-ares createDnsResolver() is ever called.  

Overall, this PR can help that if c-ares extension is not linked in, or is linked in, but never used due to configuration,  ares_library_init()/ares_library_cleanup() will not be called.

source/extensions/network/dns_resolver/cares/dns_impl.cc
```diff
@@ -507,6 +508,11 @@ class CaresDnsResolverFactory : public DnsResolverFactory {
                                              typed_dns_resolver_config) const override {
     envoy::extensions::network::dns_resolver::cares::v3::CaresDnsResolverConfig cares;
     std::vector<Network::Address::InstanceConstSharedPtr> resolvers;
+    // Initialize c-ares library in case first time.
+    if (!ares_library_initialized_) {
+      ares_library_init(ARES_LIB_INIT_ALL);
+      ares_library_initialized_ = true;
```
jmarantz, line 514: what threads can this be called on? Consider Envoy-Mobile case where there can be multiple main threads.

source/extensions/network/dns_resolver/cares/dns_impl.cc
```diff
@@ -521,6 +527,16 @@ class CaresDnsResolverFactory : public DnsResolverFactory {
     }
     return std::make_shared<Network::DnsResolverImpl>(cares, dispatcher, resolvers);
   }
+
+  void cleanup() override {
+    // Cleanup c-ares library if initialized.
+    if (ares_library_initialized_) {
+      ares_library_cleanup();
```
jmarantz, line 534: Can we move this cleanup to the ares factory destructor? That would feel a little cleaner.

source/extensions/network/dns_resolver/cares/dns_impl.cc
```diff
@@ -495,6 +495,7 @@ DnsResolverImpl::AddrInfoPendingResolution::availableInterfaces() {
 // c-ares DNS resolver factory
 class CaresDnsResolverFactory : public DnsResolverFactory {
 public:
+  CaresDnsResolverFactory() : ares_library_initialized_(false) {}
```
jmarantz, line 498: are we guaranteed to have only one instance of CaresDnsResolverFactory ?

jmarantz (commented):
Thanks I think this is vectoring toward the solution we discussed.

source/extensions/network/dns_resolver/cares/dns_impl.cc
yanjunxiang-google: Yes, IIUC, there is only one instance of CaresDnsResolverFactory(), which is registered here:
https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/extensions/network/dns_resolver/cares/dns_impl.cc#L527

Every time when we need to createDnsResolver, we first retrieve the factory by, e.g:
https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/server/server.cc#L678

source/extensions/network/dns_resolver/cares/dns_impl.cc
yanjunxiang-google: Good point. Let me move the ares_library_init() to main thread. Please check the latest change, which call the ares_library_init() within factory. init() after factory retrieval. And IIUC,  the factory retrieval is done in maint hread.
https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/common/network/dns_resolver/dns_factory_util.cc#L82

BTW, createDnsResolver() can be called in both main threads, like here:
https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/server/server.cc#L681

Or worker threads, like here
https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/extensions/filters/udp/dns_filter/dns_filter_resolver.h#L28

I tried to move the latter to main thread, but later on found doing it in worker threads is actually by design. Please check : https://github.com/envoyproxy/envoy/pull/21887#issuecomment-1166283363.

source/extensions/network/dns_resolver/cares/dns_impl.cc
jmarantz: It's not obvious to me from looking at the code that there's only one instance of it.

I think it would be better to keep a separate structure that "owns" the initialization/shutdown of c_ares, which is created with MUTABLE_CONSTRUCT_ON_FIRST_USE, and provides start() and finish() methods that update a counter protected by a mutex, and have the init/destruct run inside the mutex on transitions of the count from 0->1 and 1->0.

source/common/network/dns_resolver/dns_factory_util.cc
```diff
@@ -79,8 +79,11 @@ void handleLegacyDnsResolverData(
 Network::DnsResolverFactory& createDnsResolverFactoryFromTypedConfig(
     const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) {
   ENVOY_LOG_MISC(debug, "create DNS resolver type: {}", typed_dns_resolver_config.name());
-  return Config::Utility::getAndCheckFactory<Network::DnsResolverFactory>(
-      typed_dns_resolver_config);
+  auto& factory =
+      Config::Utility::getAndCheckFactory<Network::DnsResolverFactory>(typed_dns_resolver_config);
+  // Perform factory initialization.
+  factory.init();
```
yanavlasov, line 85: If you move this code into the process_wide.cc you will not need complexity with the `ares_library_initialized_` in the factory. You already put the code for clean-up there.

source/extensions/network/dns_resolver/cares/dns_impl.cc
yanavlasov: Can't do it in the destructor since factories are statically initialized and will race with whatever static clean-up happens in c-ares.

yanavlasov (changes requested):
/wait-any

source/common/network/dns_resolver/dns_factory_util.cc
yanjunxiang-google: The reason we do factory.init() here is that only if the factory is c-ares, it will call ares_library_init() and set ares_library_initialized_ to be true.  If c-ares Resolver is never created, then  ares_library_init() will never be called, and ares_library_initialized_  stay false.   This provide a path to never call these ares_library_init()/ares_library_cleanup() functions if proper DNS resolvers are configured.

If we do this in process_wide(), there is no context there to check whether c-ares resolver is configured. That will ends up ares_library_init() get called even c-ares DNS resolver is never created.

envoy/network/dns_resolver.h
```diff
@@ -26,6 +26,8 @@ class DnsResolverFactory : public Config::TypedFactory {
       const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) const PURE;
 
   std::string category() const override { return std::string(DnsResolverCategory); }
+  virtual void init() {}
```
jmarantz, line 29: add doc to these methods, saying why they are needed.

source/exe/process_wide.cc
```diff
@@ -58,7 +57,11 @@ ProcessWide::~ProcessWide() {
 
   ASSERT(init_data.count_ > 0);
   if (--init_data.count_ == 0) {
-    ares_library_cleanup();
+    // Cleanup c-ares library if it is linked in.
```
jmarantz, line 60: Add a comment that it self-inits on demand in a thread-safe way, thus avoiding initializing it if not used.

source/extensions/network/dns_resolver/cares/dns_impl.cc
jmarantz: Makes sense. Can you comment that the destructor is not called in deterministic order as the object is statically constructed.

source/extensions/network/dns_resolver/cares/dns_impl.cc
```diff
@@ -521,6 +521,25 @@ class CaresDnsResolverFactory : public DnsResolverFactory {
     }
     return std::make_shared<Network::DnsResolverImpl>(cares, dispatcher, resolvers);
   }
+
+  void init() override {
+    // Initialize c-ares library in case first time.
+    absl::Mutex mutex;
```
jmarantz, line 527: this mutex is not going to do anything. It needs to be a class member.

Two different threads could call init() concurrently and they'd each get separate local mutexes.

source/extensions/network/dns_resolver/cares/dns_impl.cc
```diff
@@ -521,6 +521,25 @@ class CaresDnsResolverFactory : public DnsResolverFactory {
...
+  }
+  void cleanup() override {
+    // Cleanup c-ares library if initialized.
+    if (ares_library_initialized_) {
+      ares_library_cleanup();
+    }
+  }
+
+private:
+  mutable bool ares_library_initialized_{false};
```
jmarantz, line 542: ABSL_GUARDED_BY(mutex_)

jmarantz: also; why mutable?

envoy/network/dns_resolver.h
yanjunxiang-google: ACK

source/exe/process_wide.cc
yanjunxiang-google: ACK

envoy/network/dns_resolver.h
```diff
@@ -26,6 +26,20 @@ class DnsResolverFactory : public Config::TypedFactory {
...
   std::string category() const override { return std::string(DnsResolverCategory); }
+
+  /*
+   * Initialize the related data for this type of DNS resolver.
+   * For some DNS resolvers, like c-ares, there are some specific data structure
+   * needs to be initialized before using it to resolve target.
+   */
+  virtual void init() {}
+
+  /*
```
jmarantz, line 37: Use `/**` per Doxygen style, here and above.

jmarantz: nit: can you rename them initialize() and terminate() as we tend to avoid abbreviations like `init` and `terminate` or `shutdown` is more of the dual of `initialize` then `cleanup`.

jmarantz: typically you respond "ACK" when the comment was not actionable. If you are actually doing what is asked you'd respond "Done" or just resolve the comment.

It looks like you actually did this.

jmarantz (dismissed):
Thanks!

source/common/network/dns_resolver/dns_factory_util.cc
```diff
@@ -79,8 +79,11 @@ void handleLegacyDnsResolverData(
 Network::DnsResolverFactory& createDnsResolverFactoryFromTypedConfig(
     const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config) {
   ENVOY_LOG_MISC(debug, "create DNS resolver type: {}", typed_dns_resolver_config.name());
-  return Config::Utility::getAndCheckFactory<Network::DnsResolverFactory>(
-      typed_dns_resolver_config);
+  auto& factory =
+      Config::Utility::getAndCheckFactory<Network::DnsResolverFactory>(typed_dns_resolver_config);
+  // Perform factory initialization.
```
jmarantz, line 84: nit: above comment does not appear necessary: `factory.initialize()` says it all to me :)

yanjunxiang-google: done

source/exe/process_wide.cc
```diff
@@ -58,7 +57,12 @@ ProcessWide::~ProcessWide() {
 
   ASSERT(init_data.count_ > 0);
   if (--init_data.count_ == 0) {
-    ares_library_cleanup();
+    // The c-ares library init is done in a thread-safe way when it is actually used for resolving.
+    // Clean it up here in case it is ever used.
+    if (auto* dns_factory = Config::Utility::getAndCheckFactoryByName<Network::DnsResolverFactory>(
+            std::string(Network::CaresDnsResolver), true)) {
+      dns_factory->terminate();
```
mattklein123, line 64: I think this should be done by category vs. specific resolver?

mattklein123 (changes requested):

/wait

source/exe/process_wide.cc
mattklein123: Yes that is correct. The way you wrote the code is that you generically call initialize() on the factory, not calling terminate() on the factory doesn't make sense to me so let's go ahead and do that now. Thank you!

yanjunxiang-google: I see what you mean now. Yeah, that make sense. We should do same thing for initialize and terminate.

Since what we want to do is specific for c-ares factory. I added a c-ares check there before initialize. This make them consistent.  PTAL.

source/common/network/dns_resolver/dns_factory_util.cc
```diff
@@ -81,7 +81,9 @@ Network::DnsResolverFactory& createDnsResolverFactoryFromTypedConfig(
   ENVOY_LOG_MISC(debug, "create DNS resolver type: {}", typed_dns_resolver_config.name());
   auto& factory =
       Config::Utility::getAndCheckFactory<Network::DnsResolverFactory>(typed_dns_resolver_config);
-  factory.initialize();
+  if (factory.name() == std::string(CaresDnsResolver)) {
```
jmarantz, line 84: I think it'd be cleaner if unconditionally call `initialize()`, and unconditionally call `terminate()` also. They do have empty-impls in base classes so it'll work fine and be much cleaner.

If some other DNS resolver wants initialize/terminate semantics they can just override the default and not add special cases.

mattklein123: +1 I was going to say the same thing. Please just do it by category.

yanjunxiang-google: Done.

source/common/config/utility.h
```diff
@@ -241,6 +241,17 @@ class Utility {
                                    POOL_HISTOGRAM(scope))};
   }
 
+
+
+  /**
+   * Get the hash map for this factory.
+   */
+  template <class Factory>
+  static absl::flat_hash_map<std::string, Factory*>&  getFactoryMap() {
```
jmarantz, line 250: I don't like exposing the representation in the API; can we instead use a forEach model, where you call a std::function on every element?

jmarantz (commented):
Remember to look at the format error also.

envoy/network/dns_resolver.h
```diff
@@ -26,7 +26,30 @@ class DnsResolverFactory : public Config::TypedFactory {
...
 
+/**
+ * Terminate the DNS resolver factories.
+ */
+static inline void terminateDnsResolverFactories() {
+  auto& factories = Registry::FactoryRegistry<Network::DnsResolverFactory>::factories();
+  std::for_each(factories.begin(), factories.end(),
+                [](auto& factory_it) { factory_it.second->terminate(); });
+}
+
```
jmarantz, line 53: this looks reasonable. So that the abstraction is somewhat more symmetric can you declare a function `intializeDnsResolverFactory(const envoy::config::core::v3::TypedExtensionConfig& typed_dns_resolver_config)` and put it here, to abstract the call to `Config::Utility::getAndCheckFactory<Network::DnsResolverFactory>(typed_dns_resolver_config);` and `factory.initialize()` ? The impl would also go in dns_resolver_util.cc.

envoy/network/dns_resolver.h
```diff
@@ -26,7 +26,30 @@ class DnsResolverFactory : public Config::TypedFactory {
...
+   * needs to be cleaned up before terminates Envoy.
+   */
+  virtual void terminate() {}
 };
 
+/**
+ * Terminate the DNS resolver factories.
+ */
+static inline void terminateDnsResolverFactories() {
+  auto& factories = Registry::FactoryRegistry<Network::DnsResolverFactory>::factories();
```
jmarantz, line 49: suggest making this a static method of DnsResolverFactory and put the impl in dns_resolver_util.cc, rather than putting an inline impl in the interface.

envoy/network/dns_resolver.h
```diff
@@ -40,16 +40,20 @@ class DnsResolverFactory : public Config::TypedFactory {
...
-  auto& factories = Registry::FactoryRegistry<Network::DnsResolverFactory>::factories();
-  std::for_each(factories.begin(), factories.end(),
-                [](auto& factory_it) { factory_it.second->terminate(); });
-}
+  /**
+   * Find the DNS resolver factory based on the typed config and initialize it.
+   * @returns the DNS Resolver factory.
+   * @param typed_dns_resolver_config: the typed DNS resolver config
+   */
+  static Network::DnsResolverFactory& initializeDnsResolverFactory(
```
jmarantz, line 49: great! Now that we are inside DnsResolverFactory I think we can rename this method to createFactory(), and we can rename terminateDnsResolverFactories to terminateFactory().

The `DnsResolver` part of the names now look redundant at call-sites.

jmarantz (approved):

yanjunxiang-google:
Thanks jmarantz@ for the comments and approval!

mattklein123 (approved):

PR #21887, opened 2022-06-24
Avoid creating DNS resolver for DNS filter in data plane

yanjunxiang-google:
Create DNS resolver for DNS filter in control plane config parsing, not in data plane.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/udp/dns_filter/dns_filter.cc
+++ b/source/extensions/filters/udp/dns_filter/dns_filter.cc
@@ -158,18 +158,23 @@ DnsFilterEnvoyConfig::DnsFilterEnvoyConfig(
   }
 
   forward_queries_ = config.has_client_config();
+  envoy::config::core::v3::TypedExtensionConfig typed_dns_resolver_config;
+  Network::DnsResolverFactory * dns_resolver_factory;
+
   if (forward_queries_) {
     const auto& client_config = config.client_config();
-    dns_resolver_factory_ =
-        &Network::createDnsResolverFactoryFromProto(client_config, typed_dns_resolver_config_);
+    dns_resolver_factory =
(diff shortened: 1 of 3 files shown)


yanjunxiang-google:
This change is causing a crash due to thread safety check.  
"
[2022-06-25 12:42:57.418][27][debug][dns] [source/extensions/network/dns_resolver/cares/dns_impl.cc:341] dns resolution for www.google.com started
[2022-06-25 12:42:57.418][27][critical][assert] [source/common/event/timer_impl.cc:31] assert failure: dispatcher_.isThreadSafe().
[2022-06-25 12:42:57.418][27][critical][backtrace] [./source/server/backtrace.h:104] Caught Aborted, suspect faulting address 0xfffe00000010
"

It looks like DNS filter resolving by design is in worker threads.  It doesn't have the logic as in dynamic forward proxy which post the resolving to main threads:  https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/extensions/common/dynamic_forward_proxy/dns_cache_impl.cc#L110 . 

Also, for DFP, after resolving finished, it posts the results back to all worker threads:
https://github.com/envoyproxy/envoy/blob/326b205af6b16dbf75f6bfbf246082f5966bc0b0/source/extensions/common/dynamic_forward_proxy/dns_cache_impl.cc#L386

DNS filter only needs to send the resolved data back to client, and doesn't need to post the data to other worker threads. Therefore , such worker-thread--->main-thread---->worker-thread resolving process is not needed in DNS filter.  And with that, the DNS resolver has to be created in worker threads during DNS filter creation, not in the main threads during DNS filter config parsing.  Close this PR.

Issue #9958, opened 2020-02-07
tracing: Transition to OpenTelemetry from OpenTracing and OpenCensus


moderation:
*Title*: *Plan to transition to OpenTelemetry*

*Description*: Envoy currently supports OpenTracing (OTr) and OpenCensus (OC) [0]. In May 2019 it was announced that these projects were merging into the new OpenTelemetry (OTel) project [1]. The original plan was to have the legacy project repos moved to read-only by the end of 2019. That hasn't happened but according to the OTel maintainers they are aiming for beta releases in March [2].

Should Envoy:
- keep OTr and OC and add OTel (eventually deprecating the legacy projects)?
- replace OTr and OC with OTel?

https://github.com/envoyproxy/envoy/pull/9955 is planning on adding to the OC capability. The OC service / agent repo says its in maintenance mode and points to [OTel/opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector)

*Relevant Links*:
0 : [config.trace.v3.Tracing.Http](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/trace.proto.html?highlight=opentracing#config-trace-v3-tracing-http)
1 : [OpenTracing, OpenCensus Merge into a Single New Project, OpenTelemetry](https://thenewstack.io/opentracing-opencensus-merge-into-a-single-new-project-opentelemetry/)
2 : [OpenTelemetry Monthly Update: January 2020](https://medium.com/opentelemetry/opentelemetry-monthly-update-january-2020-73579f9cb979)

yvespp:
Probably also depends on the C++ otel library so it can be used in envoy.
https://github.com/open-telemetry/opentelemetry-cpp is still in active development and has no releases yet.

moderation:
[AWS adds observability metrics to the OpenTelemetry C++ library](https://aws.amazon.com/blogs/opensource/aws-adds-observability-metrics-to-the-opentelemetry-c-library/)

moderation:
Getting closer - RC1 - https://opensource.googleblog.com/2020/10/opentelemetrys-first-release-candidates.html

moderation:
Announced at re:invent December 2020 - AWS Distro for OpenTelemetry - https://aws-otel.github.io/

gramidt:
As a member of the OpenTelemetry community, I thought I would share that the OpenTelemetry Tracing specification 1.0.1 has been released. 

Long term support includes:

* API: 3 year support guarantee
[...]

PR #21893, opened 2022-06-25
extension: Add Span::setTag for setting attributes in OTLP Tracer

AlexanderEllis:
Signed-off-by: Alex Ellis <ellisonjtk@gmail.com>

Commit Message:  Add Span::setTag for setting attributes in OTLP Tracer
Additional Description: Adding attributes to the OTLP traces generated in the OTLP tracer (see https://github.com/envoyproxy/envoy/issues/9958). 

Here's a sampling of the attributes Envoy creates in a basic test setup (Envoy making a child span for its call to the Nighthawk test server) by just adding this method:

image

Risk Level: low
Testing: Added a unit test (with fleshed out proto expectations)
Docs Changes: N/A
Part of #9958

### DIFF
--- a/source/extensions/tracers/opentelemetry/tracer.cc
+++ b/source/extensions/tracers/opentelemetry/tracer.cc
@@ -61,6 +61,30 @@ void Span::injectContext(Tracing::TraceContext& trace_context,
   trace_context.setByReferenceKey(kTraceParent, traceparent_header_value);
 }
 
+void Span::setTag(absl::string_view name, absl::string_view value) {
+  // The attribute key MUST be a non-null and non-empty string.
+  if (name.empty()) {
+    return;
+  }
+  // Attribute keys MUST be unique.
+  // If a value already exists for this key, overwrite it.
+  for (auto& key_value : *span_.mutable_attributes()) {
+    if (key_value.key() == name) {
+      key_value.mutable_value()->set_string_value(std::string{value});
+      return;
+    }
+  }
+  // If we haven't found an existing match already, we can add a new key/value.
+  opentelemetry::proto::common::v1::KeyValue key_value =
+      opentelemetry::proto::common::v1::KeyValue();
+  opentelemetry::proto::common::v1::AnyValue value_proto =
+      opentelemetry::proto::common::v1::AnyValue();
+  value_proto.set_string_value(std::string{value});
+  key_value.set_key(std::string{name});
+  *key_value.mutable_value() = value_proto;
+  *span_.add_attributes() = key_value;
+}
+
 Tracer::Tracer(OpenTelemetryGrpcTraceExporterPtr exporter, Envoy::TimeSource& time_source,
                Random::RandomGenerator& random, Runtime::Loader& runtime,
                Event::Dispatcher& dispatcher, OpenTelemetryTracerStats tracing_stats)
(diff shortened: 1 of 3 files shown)


AlexanderEllis:
Hi @htuch! When this auto-added you for review, I realized I added you for this [extension under CODEOWNERS](https://github.com/envoyproxy/envoy/blob/main/CODEOWNERS#L59) in the last PR without checking with you — happy to update that if you'd like, as I'm sure you get plenty of review requests already :)

moderation:
@AlexanderEllis initial testing and it looks like this is working well. Fantastic! And thanks for working on this

AlexanderEllis:
Awesome, glad to hear it, and thanks for testing it out!

moderation:
One thing I'm stuck on is the health check filter. I have a cluster health check on my front Envoy
```json
                "health_checks": [
                    {
                        "always_log_health_check_failures": true,
                        "healthy_threshold": 2,
                        "http_health_check": {
                            "path": "/z/",
                            "codec_client_type": "HTTP2",
                            "service_name_matcher": {
                                "exact": "sync_service"
                            }
                        },
                        "interval": "5s",
                        "interval_jitter": "0.5s",
                        "timeout": "1s",
                        "unhealthy_threshold": 2
                    }
                ],
```
and then a health check filter on the upstream Envoy
```json
                                        {
                                            "name": "envoy.filters.http.health_check",
                                            "typed_config": {
                                                "@type": "type.googleapis.com/envoy.extensions.filters.http.health_check.v3.HealthCheck",
                                                "headers": [
                                                    {
                                                        "name": ":path",
                                                        "string_match": {
                                                            "exact": "/z/"
                                                        }
                                                    }
                                                ],
                                                "pass_through_mode": false
                                            }
                                        },
```
But the OpenTelemetry Collector is still seeing and sending health checks with `user_agent` `Envoy/HC`. 

Feels like this should be handled in the HCM. I'm pretty sure the filter worked when using the Lightstep tracer previously.

Any ideas?

htuch:
LGTM, but we probably should have a non-Googler maintainer consistently reviewing these as well, @mattklein123 can you take a look at this short PR as you also reviewed the main OT extension?

I'm happy to remain CODEOWNERS, but it's possible @yanavlasov or someone else in our Envoy Platform team might take it if they have cycles.

AlexanderEllis:
@moderation Thanks for calling that out! Just to make sure I'm on the same page, you have your front Envoy health checking your upstream Envoy, and the OTel collector is _not_ seeing the traces from that check with the right user agent? I'm going to be afk this week, but I should be able to dig in and try that setup out myself after the 4th 👍 

@htuch Sounds great to me, thanks!

source/extensions/tracers/opentelemetry/tracer.cc
```diff
@@ -61,6 +61,30 @@ void Span::injectContext(Tracing::TraceContext& trace_context,
...
+    return;
+  }
+  // Attribute keys MUST be unique.
+  // If a value already exists for this key, overwrite it.
+  for (auto& key_value : *span_.mutable_attributes()) {
+    if (key_value.key() == name) {
+      key_value.mutable_value()->set_string_value(std::string{value});
+      return;
+    }
+  }
```
mattklein123, line 76: This kind of N^2 behavior seems sub-optimal, though I guess there aren't usually that many tags. It might be worth it to optimize this later.

mattklein123 (approved):
Thanks!

AlexanderEllis:
@moderation Should be all set with https://github.com/envoyproxy/envoy/pull/22047 merged in 👍 Thanks again for trying this out!

PR #21894, opened 2022-06-26
Added content type "application/grpc-web+proto" to compresed-by-default list

dorfire:
This will help me utilize the compression feature via ESPv2.


wbpcode:
Could you check the CI first? Thanks.

/wait

KBaichoo (commented):
test failing: test/extensions/filters/http/compressor/compressor_filter_test.cc:222

source/extensions/filters/http/compressor/compressor_filter.cc
```diff
@@ -30,12 +30,13 @@ const uint64_t DefaultMinimumContentLength = 30;
...
-       "text/javascript", "text/x-javascript", "text/ecmascript", "text/js", "text/jscript",
-       "text/x-js", "application/ecmascript", "application/x-json", "application/xml",
-       "application/json", "image/svg+xml", "text/xml", "application/xhtml+xml"});
+  CONSTRUCT_ON_FIRST_USE(std::vector<std::string>,
+                         {"text/html", "text/plain", "text/css", "application/javascript",
+                          "application/x-javascript", "text/javascript", "text/x-javascript",
+                          "text/ecmascript", "text/js", "text/jscript", "text/x-js",
+                          "application/ecmascript", "application/x-json", "application/xml",
+                          "application/json", "image/svg+xml", "text/xml", "application/xhtml+xml",
+                          "application/grpc-web+proto"});
```
KBaichoo, line 39: From https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2

> the receiver should assume the default is "+proto" when the message format is 
missing in Content-Type (as "application/grpc-web")

seems like application/grpc-web should also be included since it's a synonym of the one added? 

Can you also help me understand why insufficent to config these content-types in https://github.com/envoyproxy/envoy/blob/c58ccaf8e9276ebd095652e56e89c7d56e92e6c0/api/envoy/extensions/filters/http/compressor/v3/compressor.proto#L45 ? Thanks

dorfire: 1. thanks for the clarification, added 'application/grpc-web' as well
2. valid point. truth is, it's just more difficult to configure Envoy via ESPv2, so I'd rather have it "Just Work" like it currently Just Works with JSON :)

KBaichoo (approved):

Issue #21880, opened 2022-06-24
What happened to the `format_pre` ci job ?


phlax:
As part of ongoing efforts to optimize and integrate the formatting and linting tooling, the `format_pre` job has now been integrated with the `format` job - running:

```console

./ci/do_ci.sh format

```

will run all of the existing `format_pre` checks plus the checks that were previously run in `format`, with the exception of the `protobuf` formatting test

wrt the protobuf formatting test, this can in some cases be very slow. So far it has been optimized to make use of bazel's cache. It can be optimized quite a bit further and will be integrated with the other formatting tools once this is done.

going forward, some docs still need updating (#21844 ) - there are some improvements/fixes to error handling/output (eg #21877), and some new tooling is being added #21849)

apologies for any disruption and instability - hopefully the outcome will be improved tooling and dev workflows

phlax:
cc @RyanTheOptimist

PR #21896, opened 2022-06-27
[WIP] tooling: Cleanups/fixes for format errors

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
related to #21880

phlax:
fixed proto_format diff here:

- https://dev.azure.com/cncf/envoy/_build/results?buildId=111470&view=logs&j=f0537f27-2e39-530f-1fd2-a5ecd0924017&t=388a504b-3b8d-52aa-33e0-1e12bca05378&l=787
- https://dev.azure.com/cncf/_apis/resources/Containers/12984312/format?itemPath=format%2Ffix_proto_format.diff

improved format log (diff output added) here:

- https://dev.azure.com/cncf/envoy/_build/results?buildId=111470&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39&l=142

phlax:
cc @tyxia this should improve the error handling if any config cannot be parsed as yaml

phlax:
example local bad yaml handling:

```console

ERROR (validation failed): docs/root/start/quick-start/_include/envoy-demo.yaml
while scanning a simple key
  in "<unicode string>", line 17, column 1:
    zzzzzzzzzzzzzzzzzzzzzzzzzzzz
    ^
could not find expected ':'
  in "<unicode string>", line 18, column 25:
                typed_config:


```

tools/proto_format/proto_sync.py
```diff
@@ -12,7 +12,8 @@
 
 
 def git_status(path):
-    return subprocess.check_output(['git', 'status', '--porcelain', str(path)]).decode()
+    return subprocess.check_output(
+        ['git', 'status', '--porcelain', str(path)], cwd=str(path)).decode()
```
wbpcode, line 16: should we change the workdir by the `cwd`? Looks like the `git_status` is used to check if there is dirty api tree.

wbpcode (commented):
Thanks for your great contribution. Only one comment.

tools/proto_format/proto_sync.py
phlax: im not sure i follow - but the thing that this script is passed is the api_root and that is what it is comparing - so i used this 

the other option would be to use the api_root parent which is the repo - but this felt a bit arbitrary so just used the api_root instead

phlax: `str(path)` is absolute so it will check the (correct) api path - but do it _in_ that directory also

phlax: its the equivalent of the following in bash

```console

x@y:/src/workspace/envoy# cd api
x@y:/src/workspace/envoy/api# git status --porcelain /src/workspace/envoy/api

```

wbpcode: Yep, it's OK if the `str(path)` is absolute. But in the current code base, the default value of `api_root` is `./api`, a normal relatively string may be passed to the script.

So I am worried that this change will result in some unexpected results.

```
    parser.add_argument('--api_root', default='./api')
```

phlax: ok lets remove the default and i guess make the arg compulsory

phlax: hmm, actually wait - rather just make sure the path is absolute - it is always called so in ci, but for sure we should convert whatever we are passed in to being absolute - good catch

wbpcode: Thanks. 😄  

FWIW, how about change the
                 
```
_git_status = git_status(api_root)
```
to
```
_git_status = git_status(api_root_path)
```
where the `api_root_path` always a absolute version.

phlax: oh wait again!

that already happens - see here https://github.com/envoyproxy/envoy/blob/main/tools/proto_format/proto_sync.py#L47

maybe that should happen in main to be more explicit - i can cleanup when i follow up - im working on these files atm

phlax: k - got it - on it - thanks for thorough review

wbpcode (approved):
LGTM. Thanks, just wait for ci to be passed.

PR #21907, opened 2022-06-27
Added Pattern Matching Library for Url Pattern Rewrites

silverstar194:
Commit Message: Added Pattern Matching Library for Url Pattern Rewrites

Additional Description:
Ported code from internal fork.

Risk Level: Low. No impact on current functionality.
   
Testing:
Unit testing. See files.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


stevenzzzz (commented):
First round. Looks good. 
Thanks for working on this!

source/common/common/matching/url_template_matching.cc
```diff
@@ -0,0 +1,169 @@
...
+
+#ifndef SWIG
+// Silence warnings about missing initializers for members of LazyRE2.
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+#endif
+
+inline re2::StringPiece ToStringPiece(absl::string_view text) { return {text.data(), text.size()}; }
+
+bool IsPatternMatch(absl::string_view pattern, absl::string_view capture_regex) {
+  RE2 regex = RE2(ToStringPiece(capture_regex));
```
zhxie, line 30: Can we use `Regex::CompiledMatcher` and `Regex::parseRegex` instead?

api/envoy/config/route/v3/route_components.proto
```diff
@@ -1362,6 +1362,24 @@ message RouteAction {
   MaxStreamDuration max_stream_duration = 36;
 }
 
+// Holds the segments for rewriting urls base on pattern templates
+message RouteUrlRewritePattern {
```
zhxie, line 1366: Could you provide an example of `RouteUrlRewritePattern`? And we have `envoy::type::matcher::v3::RegexMatchAndSubstitute` currently, can we extend it if it does not meet your need?

zhxie (commented):
A high level comment, we currently have `envoy::type::matcher::v3::RegexMatchAndSubstitute` for regex rewrite, can we extent it instead of adding a new API.

api/envoy/config/route/v3/route_components.proto
htuch: Why is this proto in the public API if not referenced anywhere in it?

source/common/common/matching/url_template_matching.cc
```diff
@@ -0,0 +1,169 @@
...
+    if (!url_template_matching_internal::IsValidIndent(segments2[0])) {
+      return absl::InvalidArgumentError("Invalid variable name");
+    }
+    result.emplace_back(segments2[0], RewriteStringKind::kVariable);
+  }
+  return result;
+}
+
+absl::StatusOr<envoy::config::route::v3::RouteUrlRewritePattern>
+ParseRewritePattern(absl::string_view pattern, absl::string_view capture_regex) {
```
htuch, line 90: Please use Envoy style, ie. `parseRewritePattern`.

lizan (changes requested):
Taking a step back, what's the point of using regex engine to implement the URL template matching? I thought the benefit of adding URL template was performance and simplicity of implementation (without relying on regex).

Also we already pull a URL pattern matcher library for gRPC transcoder, which is from Google as well: https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/blob/master/src/include/grpc_transcoding/path_matcher.h . It might makes more sense to extract this part into Envoy or as a separate library?

cc @qiwzhang

source/common/common/matching/url_template_matching.cc
silverstar194: changed

silverstar194:
I'll let @htuch jump in here as well.

The major benefit is better and more streamlined user experience. Users do not want to go off and write regex in order to match paths and do rewrites like: /{item_one}/** or /goo/{item_one}/var/** . The easy use of variables and wildcard matching is the goal.

source/common/common/matching/url_template_matching.cc
```diff
@@ -0,0 +1,167 @@
...
+    return status.status();
+  }
+  std::vector<RewritePatternSegment> processed_pattern = *std::move(status);
+
+  const std::map<std::string, int>& capture_index_map = regex.NamedCapturingGroups();
+
+  for (const auto& [str, kind] : processed_pattern) {
+    switch (kind) {
+    case RewriteStringKind::kLiteral:
+      parsed_pattern.add_segments()->set_literal(std::string(str));
```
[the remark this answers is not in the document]
silverstar194: Yes, cannot guarantee the config is still there

source/common/common/matching/url_template_matching.cc
```diff
@@ -0,0 +1,167 @@
...
+
+absl::StatusOr<std::string> ConvertURLPatternSyntaxToRegex(absl::string_view url_pattern) {
+
+  absl::StatusOr<ParsedUrlPattern> status =
+      url_template_matching_internal::ParseURLPatternSyntax(url_pattern);
+  if (!status.ok()) {
+    return status.status();
+  }
+  struct ParsedUrlPattern pattern = *std::move(status);
+  return url_template_matching_internal::ToRegexPattern(pattern);
```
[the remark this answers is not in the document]
silverstar194: Just separation of logic is my understanding from original author

source/common/common/matching/url_template_matching.cc
```diff
@@ -0,0 +1,167 @@
...
+
+  for (const envoy::config::route::v3::RouteUrlRewritePattern::RewriteSegment& segment :
+       rewrite_pattern.segments()) {
+    if (segment.has_literal()) {
+      absl::StrAppend(&rewritten_url, segment.literal());
+    } else if (segment.has_var_index()) {
+      if (segment.var_index() < 1 || segment.var_index() >= capture_num) {
+        return absl::InvalidArgumentError("Invalid variable index");
+      }
+      absl::StrAppend(&rewritten_url, absl::string_view(captures[segment.var_index()].as_string()));
```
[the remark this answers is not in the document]
silverstar194: Yes, but string_view really internally really just stores the pointer and size so there isn't a copy being made.

source/common/common/matching/url_template_matching.cc
silverstar194: Maybe, I can try it out. What is the advantage?

api/envoy/config/route/v3/route_components.proto
silverstar194: Far enough. Original use before port is gone now... Let me try that.

stevenzzzz:
> > The major benefit is better and more streamlined user experience. Users do not want to go off and write regex in order to match paths and do rewrites like: /{item_one}/** or /goo/{item_one}/var/** . The easy use of variables and wildcard matching is the goal
> 
> I'm really not in favor of adding this feature if it could be implemented in terms of regex. Why can't we just document some example regex patterns then? This seems like a lot of complexity for not much benefit?

@htuch as well. 

I think the point of introducing this somewhat overlapping new feature is to guarantee a bounded runtime cost for regex usage operation as well as covering a fairly large amount of customer requests on easier path components based regex matching.

htuch:
I think this is the standard tension between treating xDS as assembler language and as something consumable by humans @mattklein123 @lizan. For the most part, don't benefit from writing regexes for most common cases - usually you're really interested in either some path component match or a suffix match etc. Regexes are a larger foot gun than needed, and if you look at the APIs provided across the industry, glob/wildcards that match these higher-level semantic goals are the norm (AWS, Cloudflare etc. in addition to GCP).

So, the idea here is really to provide an API that expresses the common case, for most Envoy users. It is strictly redundant with regexes (but then again, so is path prefix and path suffix etc.).

htuch:
@stevenzzzz point is another motivation here - runtime complexity is easier to bound when we have a limited subset of regexes. We know from experience that this limited subset works well, even in multi-tenant systems.

The way I see the use of RE2 here is that it's a simpler choice for now than implementing the engine ourselves. Previous attempts at reuse of another system (blob of Google code) did not perform as well as RE2, but given the syntax, I don't see it as being unfeasible to replace with something simple, fast and easy to reason about directly later.

htuch:
That said, having just looked at the actual implementation, this is pretty complicated and I'm not super excited about the complexity being brought into Envoy core.

Taking a step back, this is another example of us doing parsing by hand of some regular grammar. Is this the place we start to use ANTLR or something, which is long overdue anyway in other parts of the config ingestion?

If we we're doing all this manual parsing, the implementation and logic here should be fairly short and simple.

htuch:
Yeah, let's say we make this an extension point, does that work for everyone? One question I would have then is whether we should just treat it like a regex engine - functionally it would make sense, but it would be a bit confusing as it's not actually regex.

Also, when I look at what was done for the regex engine in https://github.com/envoyproxy/envoy/commit/0a92cc86e7f8b88d5af0eb2e27b5c7ef64719e56, this sets it globally for the proxy. I don't think we want that here - we want to be able to have folks who use both wildcard and REs as needed on different paths.

So, we would have a new path_matcher extension I think. Does this make sense to everyone? CC @markdroth

mattklein123:
> So, we would have a new path_matcher extension I think. Does this make sense to everyone?

This would be fine with me. I would also be fine with building this in if we can at least introduce a real grammar which would provide legwork for fixing other things also (namely access log config).

lizan:
> So, we would have a new path_matcher extension I think. Does this make sense to everyone?

Works for me as well.

markdroth:
Yeah, I think a path_matcher extension point makes a lot of sense.  I think that's what I suggested originally. :)

silverstar194:
Sounds good folks. Will reform this into an extension and circle back.

wrowe:
/wait

on followup

silverstar194:
Moved to extension. I opened a new PR here for review and assigned everyone over: https://github.com/envoyproxy/envoy/pull/22214

Will keep this PR (21907) open for reference and abandon when extension changes are complete.

silverstar194:
Adding new to series to complete this work. https://github.com/envoyproxy/envoy/pull/22459

silverstar194:
Adding PR to series to complete this work. PR contains library for rewrites and matching. https://github.com/envoyproxy/envoy/pull/22468

silverstar194:
Closing out for full PR with two extensions: https://github.com/envoyproxy/envoy/pull/22207

PR #21908, opened 2022-06-27
test: Enable addFakeUpstream() to create AutonomousUpstreams

abeyad:
Previously, the BaseIntegrationTest::addFakeUpstream() function would only create a FakeUpstream, not taking
BaseIntegrationTest::autonomous_upstream_ into account.  This could lead to confusing integration test behavior.

With this change, the addFakeUpstream() function takes a required autonomous_upstream parameter that
indicates whether an AutonomousUpstream should be created instead of a FakeUpstream. 


alyssawilk (approved):
LGTM, CI willing :-)

alyssawilk (approved):

PR #21913, opened 2022-06-27
network: allow Apple platforms to force IPv6 sockets

jpsim:
We suspect that under some conditions, non-Android platforms (i.e. iOS) also would benefit from forcibly using IPv6 sockets, so to validate this this change adds a new runtime feature generalizing the one added for Android.

Risk Level: Low, runtime guard defaults to false.
Testing: Tested on an iPhone 13 Pro via Envoy Mobile, confirmed that an IPv4 address was being mapped to an IPv6 one.
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: Not platform specific, but intended to be used on Apple platforms.
Runtime guard: `envoy.reloadable_features.always_use_v6` (default false)

### DIFF
--- a/source/common/network/io_socket_handle_impl.cc
+++ b/source/common/network/io_socket_handle_impl.cc
@@ -65,11 +65,10 @@ constexpr int messageTruncatedOption() {
 namespace Network {
 
 bool IoSocketHandleImpl::forceV6() {
-#ifndef __ANDROID_API__
+#if defined(__APPLE__) || defined(__ANDROID_API__)
   return Runtime::runtimeFeatureEnabled("envoy.reloadable_features.always_use_v6");
 #else
-  return Runtime::runtimeFeatureEnabled("envoy.reloadable_features.always_use_v6") ||
-         Runtime::runtimeFeatureEnabled("envoy.reloadable_features.android_always_use_v6");
+  return false;
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


jpsim:
Marked as draft as I test in Envoy Mobile.

source/common/network/io_socket_handle_impl.cc
```diff
@@ -477,18 +477,23 @@ IoHandlePtr IoSocketHandleImpl::accept(struct sockaddr* addr, socklen_t* addrlen
...
     // Android always uses IPv6 dual stack. Convert IPv4 to the IPv6 mapped address when
     // connecting.
     memset(&sin6, 0, sizeof(sin6));
     sin6.sin6_family = AF_INET6;
     sin6.sin6_port = sin4.sin_port;
+#if defined(__ANDROID_API__)
     sin6.sin6_addr.s6_addr32[2] = htonl(0xffff);
     sin6.sin6_addr.s6_addr32[3] = sin4.sin_addr.s_addr;
+#elif defined(__APPLE__)
+    sin6.sin6_addr.__u6_addr.__u6_addr32[2] = htonl(0xffff);
```
jpsim, line 494: Linux headers on the left, Darwin on the right.

image

mattklein123 (approved):
LGTM, thanks!

PR #21926, opened 2022-06-28
[ext_proc]: Log per grpc call latency and status for ext_proc filter

chaoqin-li1123:
Commit Message: Log processor grpc call latency and status in the filter state for observability. 
Additional Description:
Risk Level: low, This change doesn't change request/response handing logic but will cause overhead since we get timestamp for each grpc call.
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


bryanwux:
You can run `./tools/code_format/check_format.py fix && ./tools/code_format/format_python_tools.sh fix` to fix the format. See https://github.com/envoyproxy/envoy/blob/main/support/README.md.

wbpcode:
Is this still in WIP?

source/extensions/filters/http/ext_proc/ext_proc.h
```diff
@@ -39,6 +39,32 @@ struct ExtProcFilterStats {
...
+      break;
+    case ProcessorState::CallbackState::TrailersCallback:
+      trailer_call_latencies.emplace_back(latency, call_status);
+      break;
+    default:
+      body_call_latencies.emplace_back(latency, call_status);
+    }
+  }
+
+  std::vector<std::pair<std::chrono::microseconds, absl::Status>> header_call_latencies;
```
mpwarres, line 63: There is a single ExtProcGrpcStats instance per stream/FilterState, that ends up being used by both DecodingProcessorState and EncodingProcessorState. AIUI, as currently stands, with the three *_call_latencies fields, this means that request and response latencies will be mixed in the same vector--e.g. header_call_latencies will end up containing entries for both EncodingProcessorState HeadersCallback as well as DecodingProcessorState HeadersCallback. This will make it difficult to separate these measurements for reporting request headers latency vs. response headers latency.

ISTM there are two possible solutions:
A. Use separate FilterState object types for encoding vs. decoding direction.
B. Adjust the fields of ExtProcGrpcStats so that it keeps encoding vs. decoding values separable.

Of these, I think (B) is preferable--a single class is simpler and more idiomatic / in line with how other Filters use FilterState.

If you go with (B), defining 6 separate fields for (header, body, trailer) x (encode, decode) would IMO be a bit clunky. Instead you could define a CallbackType enum { RequestHeaders, RequestBody, RequestTrailers, ResponseHeaders, ResponseBody, ResponseTrailers }, and have a single field like absl::flat_hash_map<std::tuple<CallbackType, absl::Status>, std::vector<std::chrono::microseconds>> call_latencies.

source/extensions/filters/http/ext_proc/ext_proc.h
```diff
@@ -39,6 +39,32 @@ struct ExtProcFilterStats {
   ALL_EXT_PROC_FILTER_STATS(GENERATE_COUNTER_STRUCT)
 };
 
+inline constexpr absl::string_view ExtProcGrpcStatsName = "ext-proc-filter-state";
+
+class ExtProcGrpcStats : public Envoy::StreamInfo::FilterState::Object {
```
mpwarres, line 44: (optional / subjective): IIUC this is named ExtProcGrpcStats to distinguish it from ExtProcFilterStats, and because all current values in it are gRPC-related. This makes sense and certainly works.

But, ISTM that the more fundamental difference between this new class and ExtProcFilterStats is that the stats in this new class are per-stream. So another possible name for this class would be ExtProcPerStreamStats, which would give more leeway in the future for other fields. For example, if we decide that we want to capture per-stream failure messages or status flags to be included in BDN ULS logging, that would go in this class as well, but would not be strictly gRPC-related.

source/extensions/filters/http/ext_proc/ext_proc.h
```diff
@@ -39,6 +39,32 @@ struct ExtProcFilterStats {
...
+public:
+  void record(std::chrono::microseconds latency, ProcessorState::CallbackState callback_state,
+              absl::Status call_status) {
+    switch (callback_state) {
+    case ProcessorState::CallbackState::Idle:
+      header_call_latencies.emplace_back(latency, call_status);
```
mpwarres, line 50: The CallbackState::Idle state is when the filter is "Not waiting for anything", so ISTM a record() call in this state would be unexpected. Given that, I don't think we should be adding to header_call_latencies in this case. Instead, this could ENVOY_BUG or ENVOY_LOG(error).

source/extensions/filters/http/ext_proc/processor_state.cc
```diff
@@ -25,13 +25,20 @@ void ProcessorState::onStartProcessorCall(Event::TimerCb cb, std::chrono::millis
...
+  call_start_time_ = filter_callbacks_->dispatcher().timeSource().monotonicTime();
 }
 
-void ProcessorState::onFinishProcessorCall(CallbackState next_state) {
-  callback_state_ = next_state;
+void ProcessorState::onFinishProcessorCall(absl::Status call_status, CallbackState next_state) {
   if (message_timer_) {
     message_timer_->disableTimer();
   }
+  if (call_start_time_.has_value()) {
```
mpwarres, line 35: We expect call_start_time.has_value() to always be true when onFinishProcessorCall() is called, right? If so, it might be worth an ENVOY_BUG if this is ever not the case.

Hmm, though reading further, I wonder if there might be cases where onFinishProcessorCall() is invoked at multiple levels for a given stream. If that's the case, never mind the earlier comment.

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -586,7 +594,7 @@ void Filter::onReceiveMessage(std::unique_ptr<ProcessingResponse>&& r) {
     stats_.stream_msgs_received_.inc();
     processing_complete_ = true;
     closeStream();
-    onFinishProcessorCalls();
+    onFinishProcessorCalls(absl::UnknownError(""));
```
mpwarres, line 597: Can 'processing_status' be passed instead of absl::UnknownError here?

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -612,7 +620,7 @@ void Filter::onGrpcError(Grpc::Status::GrpcStatus status) {
     closeStream();
     // Since the stream failed, there is no need to handle timeouts, so
     // make sure that they do not fire now.
-    onFinishProcessorCalls();
+    onFinishProcessorCalls(absl::UnknownError(""));
```
mpwarres, line 623: It would be good to translate the GrpcStatus 'status' value into an absl::Status and pass it into onFinishProcessorCalls here.

mpwarres (changes requested):
The overall approach of capturing per-stream stats in FilterState LGTM, just some comments on lower-level details.

source/extensions/filters/http/ext_proc/ext_proc.cc
chaoqin-li1123: We can pass GrpcStatus instead of absl::Status, this status is only for grpc call result anyway.

source/extensions/filters/http/ext_proc/ext_proc.h
chaoqin-li1123: Thanks, I will separate the request and response stats.

source/extensions/filters/http/ext_proc/processor_state.cc
mpwarres: Thanks, SG. (I'd resolve this comment thread except that I can't seem to find the button to do so.)

source/extensions/filters/http/ext_proc/ext_proc.cc
mpwarres: SGTM!

source/extensions/filters/http/ext_proc/ext_proc.h
chaoqin-li1123: Thanks, I add an assertion in the caller.

source/extensions/filters/http/ext_proc/ext_proc.cc
chaoqin-li1123: Thanks, change made.

chaoqin-li1123:
@PiotrSikora

source/extensions/filters/http/ext_proc/processor_state.h
```diff
@@ -199,6 +207,7 @@ class DecodingProcessorState : public ProcessorState {
       Filter& filter, const envoy::extensions::filters::http::ext_proc::v3::ProcessingMode& mode)
       : ProcessorState(filter) {
     setProcessingModeInternal(mode);
+    processor_type_ = ProcessorType::DecodingProcessor;
```
mpwarres, line 210: Since the processor_type_ field belongs to the ProcessorState superclass, I think it would be cleaner to have it be an explicit parameter in the ProcessorState constructor, so then e.g. DecodingProcessorState could do `explicide DecodingProcessorState(...) : ProcessorState(filter, ProcessorType::DecodingProcessor) {}

This way there is structurally no way to have a ProcessorState instance whose processor_type_ field is unassigned.

mpwarres (changes requested):
LGTM mod comments. Thanks!

wrowe:
/assign @snowp 

as maintainer/codeowner

source/extensions/filters/http/ext_proc/ext_proc.h
```diff
@@ -39,6 +39,34 @@ struct ExtProcFilterStats {
   ALL_EXT_PROC_FILTER_STATS(GENERATE_COUNTER_STRUCT)
 };
 
+inline constexpr absl::string_view ExtProcStreamStatsName = "ext-proc-stream-stats";
+
+class ExtProcStreamStats : public Envoy::StreamInfo::FilterState::Object {
+public:
+  struct ProcessorGrpcStats {
```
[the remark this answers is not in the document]
chaoqin-li1123: renamed.

source/extensions/filters/http/ext_proc/ext_proc.h
```diff
@@ -126,6 +154,7 @@ class Filter : public Logger::Loggable<Logger::Id::ext_proc>,
   const FilterConfig& config() const { return *config_; }
 
   ExtProcFilterStats& stats() { return stats_; }
+  ExtProcStreamStats& grpcStats() { return *grpc_stats_; }
```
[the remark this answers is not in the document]
chaoqin-li1123: Thanks, renamed.

source/extensions/filters/http/ext_proc/processor_state.h
```diff
@@ -55,6 +55,10 @@ class ChunkQueue {
 
 class ProcessorState : public Logger::Loggable<Logger::Id::ext_proc> {
 public:
+  enum class ProcessorType {
```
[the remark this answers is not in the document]
chaoqin-li1123: Thanks, let's use envoy::config::core::v3::TrafficDirection

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -56,6 +56,30 @@ class ImmediateMutationChecker {
...
+                                             ProcessorState::ProcessorType processor_type) {
+  ProcessorGrpcStats& grpc_stats = processorGrpcStats(processor_type);
+  switch (callback_state) {
+  case ProcessorState::CallbackState::HeadersCallback:
+    grpc_stats.header_call_stats_.emplace_back(latency, call_status);
+    break;
+  case ProcessorState::CallbackState::TrailersCallback:
+    grpc_stats.trailer_call_stats_.emplace_back(latency, call_status);
+    break;
+  default:
```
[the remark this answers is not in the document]
chaoqin-li1123: There are many types of body call, I can keep the CallbackState in the Stats.

source/extensions/filters/http/ext_proc/processor_state.h
chaoqin-li1123: Is it ok to assign an UNSPECIFIED if we use the TrafficDirection enum?

source/extensions/filters/http/ext_proc/ext_proc.cc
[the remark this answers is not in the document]
mpwarres: Ok, that works too--SG.

source/extensions/filters/http/ext_proc/processor_state.h
mpwarres: Either way works, so if you prefer that, SGTM.

My individual leaning would be to add the param to the ProcessorState constructor so that it can be set from the beginning to the proper value (and ProcessorState retains responsibility for setting its own fields--plus now you can make the field const), but that's subjective, without much real practical benefit, so up to you.

mpwarres (dismissed):

mpwarres (commented):
The changes here LGTM, but you should also add some tests. Not sure if you wanted to wait for @snowp 's initial comments before adding those or not.

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -77,6 +92,14 @@ void FilterConfigPerRoute::merge(const FilterConfigPerRoute& src) {
 void Filter::setDecoderFilterCallbacks(Http::StreamDecoderFilterCallbacks& callbacks) {
   Http::PassThroughFilter::setDecoderFilterCallbacks(callbacks);
   decoding_state_.setDecoderFilterCallbacks(callbacks);
+  const Envoy::StreamInfo::FilterStateSharedPtr& filter_state =
+      callbacks.streamInfo().filterState();
+  if (!filter_state->hasData<ExtProcStreamStats>(ExtProcStreamStatsName)) {
+    filter_state->setData(ExtProcStreamStatsName, std::make_shared<ExtProcStreamStats>(),
```
snowp, line 98: I might be missing something here but why are we storing this in stream state vs just as a member of on the Filter? The filter itself is scoped to a stream

source/extensions/filters/http/ext_proc/processor_state.cc
```diff
@@ -249,7 +258,7 @@ absl::Status ProcessorState::handleBodyResponse(const BodyResponse& response) {
         }
       }
     } else {
-      onFinishProcessorCall();
+      onFinishProcessorCall(Grpc::Status::FailedPrecondition);
```
snowp, line 261: I wonder if it'll be confusing to inject these fake (I think?) grpc response codes, these aren't actually being returned by the server right? 

I don't have a better solution for this but this should at least be documented I think

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -56,6 +56,21 @@ class ImmediateMutationChecker {
...
+    std::chrono::microseconds latency, Grpc::Status::GrpcStatus call_status,
+    ProcessorState::CallbackState callback_state,
+    envoy::config::core::v3::TrafficDirection traffic_direction) {
+  ASSERT(callback_state != ProcessorState::CallbackState::Idle);
+  grpcStats(traffic_direction).stats_.emplace_back(latency, call_status, callback_state);
+}
+
+ExtProcStreamStats::GrpcStats&
+ExtProcStreamStats::grpcStats(envoy::config::core::v3::TrafficDirection traffic_direction) {
+  return traffic_direction == envoy::config::core::v3::TrafficDirection::INBOUND
```
snowp, line 69: ASSERT that it's not unspecified perhaps?

snowp (changes requested):
Thanks for working on this! Will needs tests for sure, but a few other comments to consider first

source/extensions/filters/http/ext_proc/ext_proc.cc
chaoqin-li1123: In out use case the stream stats is for logging, we want to share the state with other filter for visibility. If it is a member of filter class, we won't be able to access the it from other filter.

source/extensions/filters/http/ext_proc/ext_proc.cc
chaoqin-li1123: Thanks, assertion added.

test/extensions/filters/http/ext_proc/filter_test.cc
```diff
@@ -241,6 +242,22 @@ class HttpFilterTest : public testing::Test {
     stream_callbacks_->onReceiveMessage(std::move(response));
   }
 
+  // The number of processor grpc calls made in the encoding and decoding path.
+  void expectGrpcCalls(const envoy::config::core::v3::TrafficDirection trafic_direction,
```
mpwarres, line 246: nit: s/trafic/traffic/

mpwarres (approved):
LGTM, thanks!

source/extensions/filters/http/ext_proc/ext_proc.cc
mpwarres: How about ExtProcLoggingInfo? CacheFilter defines a similarly-purposed FilterState object for logging/per-request metrics called [CacheFilterLoggingInfo](https://github.com/envoyproxy/envoy/blob/0d1781492579cfa0791babc010bef2555f30d31c/source/extensions/filters/http/cache/cache_filter_logging_info.h#L95).

ExtProcCallResult matches the current payload but I wonder if it might not age well if we start adding other information for logging as well, e.g. some form of status flags, etc.

snowp (changes requested):
A few small suggestions but this is looking good, thanks!

chaoqin-li1123:
./retest

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -56,24 +56,24 @@ class ImmediateMutationChecker {
   std::unique_ptr<Checker> rule_checker_;
 };
 
-void ExtProcStreamStats::recordGrpcCallStats(
+void ExtProcLoggingInfo::recordGrpcCallStats(
```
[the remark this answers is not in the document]
chaoqin-li1123: Thanks, renamed.

snowp (approved):
Thanks!

Issue #13085, opened 2020-09-14
Support generic protocol access log formatter extension


caitong93:
*Title*: *Support generic protocol access log formatter extension*

*Description*:

I want to add access log support for dubbo and thrift filter, but find out current access log formatter only support http,  and it cannot be extended. Any thoughts about implement a generic access log formatter ?

mattklein123:
See https://github.com/envoyproxy/envoy/blob/308619fc7a2c403ff674e2af7dda4e73b00575ed/api/envoy/config/accesslog/v3/accesslog.proto#L318

caitong93:
`ExtensionFilter` seems not what I want. `Envoy::AccessLog::Instance`, `Envoy::AccessLog::Formatter`, `Envoy::AccessLog::FormatterProvider`, and implementations of these interfaces,  only support HTTP.  I'm not sure if envoy have plan to refactor these interfaces, so that other protocols can reuse them.

My current plan for implement a thrift access logger, without modifying envoy code:
1. implement a  access logger extension named `envoy.thrift_file_access_log`
2. convert thrift metadata to Http::HeaderMap to use the method `void log(const Http::HeaderMap* request_headers, const Http::HeaderMap* response_headers,
                   const Http::HeaderMap* response_trailers,
                   const StreamInfo::StreamInfo& stream_info)` 

will try to do this as an internal implementation. If anyone is interested, I can send out a pr.

lizan:
This seems a valid feature request but need design proposal. @mattklein123 the extension filter is about how we filter log entries, but it doesn't let access log support generic protocol, other than TCP / HTTP.

@cpakulski was looking into how to integrate MySQL logs (e.g. transaction logs which is in different layers than TCP), we saw same issue with current logging structure including ALS. It makes sense to have some extensible way to let TCP filters to emit custom logs easily (to access logger extensions, file/gRPC/etc).

mattklein123:
> It makes sense to have some extensible way to let TCP filters to emit custom logs easily (to access logger extensions, file/gRPC/etc).

I agree. We actually need access logs for the DNS filter also at Lyft (cc @abaptiste) so whatever we come up with here can be used for that.

Naively I would think that the general HTTP/TCP system should work with possible protocol specific data extensions?

cpakulski:
I am working on a prototype. The idea is to emit fairly flexible messages describing non-HTTP transactions, like database interactions. I could not fit them into gRPC message, as it is very HTTP specific.

[...]

PR #21932, opened 2022-06-28
router: unified header formatters

### LINKED ISSUE #15036: Add RESP(header-name) header formatter
Specifically I'm looking for support within `response_headers_to_add`.

I'd like to do something like:

```yaml
response_headers_to_add:
- header:
    key: x-served-by
    value: '%HOSTNAME%,%RESP(X-SERVED-BY)%'
```

The usecase is rather petty, because `append: true` can be used. But this yields duplicate headers, which is fine by spec, but it's harder to trace a path since there's no explicit ordering. We use this pattern which ends up being formatted similarly to an `X-Forwarded-For` header where the path keeps getting appended through the stack.

Ultimately, the yielded header would be:

```
x-served-by: lb-01,web-01
```

vs what we can do now as:

```
x-served-by: web-01
x-served-by: lb-01
```

I tried to look through the code for the header formatter, and I'm far from being good at C++. I think the issue, from what I was able to understand, is that the header formatter is shared and also is used for `request_headers_to_add` which means the formatter itself doesn't assume or doesn't even know yet that there are HTTP headers. I'm guessing if that's the case, in general this would be problematic and need guards around doing config like:

```yaml
request_headers_to_add:
- header:
    key: foo
    value: '%RESP(xxx)%'
```

Support for REQ formatting: https://github.com/envoyproxy/envoy/pull/8782

### LINKED ISSUE #20389: unified header formatter and substitution formatter
*Title*: *unified header formatter and substitution formatter*

*Description*:

As the title, now the header formatter and substitution formatter provide not exactly the same command, can we try to unified them?

Related issue: #15036 #7213

cpakulski:
Commit Message:
 unified formatters for access log and header

Additional Description:
At the core of this PR are changes to signature of just two methods:
`HeaderFormatter::evaluateHeader` - it used to take only stream_info as param. Now it also takes RequestHeaderMap and ResponseHeaderMap. This is required to directly map parameters to access log substitution formatters. 
`Router::ResponseEntry::finalizeResponseHeaders` - added RequestHeaderMap to the parameters.
Remaining changes are around changes to those two methods.

Header formatters used to have internal parser and its own set of formatters. Now parsing is done by access log's parser and header formatting is done via access log's substitution formatters.
Runtime guard `envoy_reloadable_features_unified_header_formatter `was added to allow using the previous implementation of parser and formatters.

New changes are fully backward compatible. However, header formatters for DYNAMIC_METADATA, UPSTREAM_METADATA and PER_REQUEST_STATE had a different format than access log's equivalent. Temporary translators have been added for those 3 formatters and will be removed when old format is deprecated.

All substitution formatters are allowed to be used. Some usage does not make sense though. For example, it is possible to modify request headers using a formatter which needs response headers, like ,%RESP(X-SERVED-BY)%. At the moment of modifying request headers, the response headers are not present yet and formatter produces empty string. It is left up for the administrator to select only formatters which make sense to use. This behaviour could be modified by adding attribute flags to each formatter indicating where formatters can be used (for example only in request headers, only in response headers or both), but it would come at some code complexity.

Risk Level: Med
Testing: Added unit tests. Manual testing with runtime guard on and off.
Docs Changes: Yes.
Release Notes: Yes 
Platform Specific Features:
Runtime guard: envoy.reloadable_features.unified_header_formatter
Fixes #20389
Fixes #15036


changelogs/current.yaml
```diff
@@ -242,6 +242,9 @@ new_features:
 - area: ext_proc
   change: |
     added support for per-route :ref:`grpc_service <envoy_v3_api_field_extensions.filters.http.ext_proc.v3.ExtProcOverrides.grpc_service>`.
+- area: header_formatters
+  change: |
+    all access log formatters can be used as custom request/response headers. Custom header's syntax is parsed using access logger's parser and header values are obtained using access log's substitution formatters. This feature can be reversed by setting runtime guard ``envoy.reloadable_features.unified_header_formatter`` to false.
```
phlax, line 247: nit: break line to improve readability (single line breaks are treated as single ws)

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,271 +574,18 @@ used to delimit variable names.
...
-
-%UPSTREAM_REMOTE_ADDRESS_WITHOUT_PORT%
-    Remote address of the upstream connection, without any port component.
-    IP addresses are the only address type with a port component.
-
-%UPSTREAM_REMOTE_PORT%
-    Remote port of the upstream connection.
-    IP addresses are the only address type with a port component.
-
-%PER_REQUEST_STATE(reverse.dns.data.name)%
```
phlax, line 810: are we losing any information here?

not all of these substitution vars seem to appear on the linked page

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,271 +574,18 @@ used to delimit variable names.
...
-
-%START_TIME%
-    Request start time. START_TIME can be customized with specifiers as specified in
-    :ref:`access log format rules<config_access_log_format_start_time>`.
-
-    An example of setting a custom header with current time in seconds with the milliseconds resolution:
+All HTTP :ref:`Command Operators <config_access_log_command_operators>` used for access logging may be specified
+in Custom request/response headers. However, depending where a particular Command Operator is used, the context needed for
+the Operator may not be available and the produced output is empty string. For example, the following configuration
+uses %RESPONSE_CODE% operator to modify request headers using code from the response.
```
phlax, line 580:
```suggestion
uses ``%RESPONSE_CODE%`` operator to modify request headers using code from the response.
```

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,271 +574,18 @@ used to delimit variable names.
...
+The output is an empty string, because request headers are modified
+before the request is sent upstream and the response is not received yet.
 
     .. code-block:: none
```
phlax, line 584:
```suggestion
    .. code-block:: yaml
```

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,271 +574,18 @@ used to delimit variable names.
...
 
-      route:
-        cluster: www
       request_headers_to_add:
         - header:
```
phlax, line 587: nit: cleanup this to use least indent formatting

phlax (commented):
first-pass docs review below

@cpakulski needs main merge

docs/root/configuration/http/http_conn_man/headers.rst
cpakulski: I am using term `header formatters` to describe the separate formatters which were documented in docs/root/configuration/http/http_conn_man/headers.rst. 
As `access log formatters` I mean formatters previously used to create access log entry. 
The idea behind this PR is to use` access log formatters` in both scenarios.
The situation is as follows: all formatters which were separately documented as `header formatters` are also present in access log formatters in exact form. There are 3 exceptions: `header formatter's` DYNAMIC_METADATA and UPSTREAM_METADATA used to take the same parameters as access log equivalent, but in different form.
`header formatter's` PER_REQUEST_STATE is not part of access log formatters (as you noted above), but the same information is accessed using access log's FILTER_STATE formatter.
In order to assure backwards compatibility I added code to translate DYNAMIC_METADATA and UPSTREAM_METADATA from "old" header formatter's form to the one used by access log. Both forms can be used now. Also, I added a translator to translate header formatter's PER_REQUEST_STATE to FILTER_STATE. As translator I mean C++ Envoy code which kicks in when config is received, not an offline translator which processes config files.
It means that whoever is using header formatter's format of *_METADATA and PER_REQUEST_STATE can continue doing so, until it is decided to deprecate those formats.
But I think that from now on, all header formatters should use formats equivalent to access log's one.

Do you think I should document all that logic in docs or release notes?

phlax: probably we need to say something in the changelog, but im still concerned about just removing these from the docs

i guess my first question (hoping iuc above) - are we deprecating the older substitution vars now ? (or have we already)

i think ideally we want something in the docs themselves showing the vars as deprecated for a period of time

phlax: i understand about the de-duplication here - i think defo a good idea

my concern is that either 

a) we are keeping `PER_REQUEST_STATE` + others, but we wont have any docs for them, or
b) we are not keeping them, but we are removing them from the docs before deprecating them

either way, if someone were to search for `PER_REQUEST_STATE` now they _would_ get hits, whereas after this PR they would not

cpakulski: @plax Thanks a lot for your comments and for digging into details of this PR! I agree with you that some sort of documentation is needed about formatters which changed and are going to be deprecated. I added `attention` section listing mentioned 3 formatters and substitutes which should be used.

phlax:
@tonya11en would you mind giving this a review

@cpakulski looks like this needs another main merge

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,191 +574,35 @@ used to delimit variable names.
...
+in Custom request/response headers. However, depending where a particular Command Operator is used, the context needed for
+the Operator may not be available and the produced output is empty string. For example, the following configuration
+uses ``%RESPONSE_CODE%`` operator to modify request headers using code from the response.
+The output is an empty string, because request headers are modified
+before the request is sent upstream and the response is not received yet.
 
-%DOWNSTREAM_REMOTE_ADDRESS%
-    Remote address of the downstream connection. If the address is an IP address it includes both
-    address and port.
+.. code-block:: yaml
```
phlax, line 584: it would be better to move this to a separate yaml file with a full boostrap config - that way it would be tested/validated and wont bitrot

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,191 +574,35 @@ used to delimit variable names.
...
 
-    .. note::
+   request_headers_to_add:
+     - header:
```
phlax, line 587: nit: use least indent formatting for yaml

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,191 +574,35 @@ used to delimit variable names.
...
+.. attention::
 
-%DOWNSTREAM_REMOTE_ADDRESS_WITHOUT_PORT%
-    Remote address of the downstream connection, without any port component.
-    IP addresses are the only address type with a port component.
+  The following legacy header formatters are still supported, but will be deprecated in the future.
+  The equivalent information can be accessed using indicated substitutes.
 
-    .. note::
+  %DYNAMIC_METADATA(["namespace", "key", ...])%
```
phlax, line 597: could you make these either literal or links please

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -771,74 +615,12 @@ Supported variable names are:
...
-    Remote address of the upstream connection, without any port component.
-    IP addresses are the only address type with a port component.
-
-%UPSTREAM_REMOTE_PORT%
-    Remote port of the upstream connection.
-    IP addresses are the only address type with a port component.
-
-%PER_REQUEST_STATE(reverse.dns.data.name)%
+  %PER_REQUEST_STATE(reverse.dns.data.name)%
     Populates the header with values set on the stream info filterState() object. To be
```
phlax, line 621:
```suggestion
    Populates the header with values set on the stream info ``filterState()`` object. To be
```

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -771,74 +615,12 @@ Supported variable names are:
...
     usable in custom request/response headers, these values must be of type
     Envoy::Router::StringAccessor. These values should be named in standard reverse DNS style,
```
phlax, line 623:
```suggestion
    ``Envoy::Router::StringAccessor``. These values should be named in standard reverse DNS style,
```

phlax (commented):
@cpakulski thanks for adding the deprecation note

ive added a few docs nits inline

phlax:
@wbpcode would you be able to take a look at this ?

docs/root/configuration/http/http_conn_man/headers.rst
cpakulski: Good idea. Is there an example how to do that?

docs/root/configuration/http/http_conn_man/headers.rst
cpakulski: Cannot use links, because documentation for obsoleted formatters has been removed. I could use URLs pointing to previous release (1.22), but it could be misleading.

docs/root/configuration/http/http_conn_man/headers.rst
cpakulski: Done.

docs/root/configuration/http/http_conn_man/headers.rst
phlax: if you `git grep literalinclude docs` you should find other examples

docs/root/configuration/http/http_conn_man/headers.rst
phlax: ah, i see

docs/root/configuration/http/http_conn_man/headers.rst
cpakulski: Added full config. Thanks for the idea!

cpakulski:
Code is ready for review. For some reason `lcov` counts comments as code lines, which brings down coverage. I will address it later.

wbpcode:
Thanks. This is a great work. It seems that there are some base interface updates.

Before the further review, I wonder could we update the `StreamInfo` to set and get request/response headers/trailers? The `StreamInfo` provides interfaces to get/set request headers for now, could we extend these interfaces to set and get more info of request/response messages? 🤔 Then all the formatters could work with only one parameter of `StreamInfo`. 
May be also related to #22242.
cc @alyssawilk 

/wait-any

cpakulski:
> Before the further review, I wonder could we update the StreamInfo to set and get request/response headers/trailers? The StreamInfo provides interfaces to get/set request headers for now, could we extend these interfaces to set and get more info of request/response messages? 🤔 Then all the formatters could work with only one parameter of StreamInfo.
May be also related to https://github.com/envoyproxy/envoy/pull/22242.

I was actually hoping to remove `get/set request headers` from `StreamInfo` after this PR is accepted and runtime guard is retired. It was used only by `REQ(X)` header formatter which will be replaced by equivalent `REQ(X)` access log formatter which does not take headers from StreamInfo. 

If you look at the code, whenever formatters are called to manipulate headers, all required params (including request and response headers) are already easy to access and there is no need to take them from StreamInfo.

IMHO it would be cleaner to not to pack everything to StreamInfo.

wbpcode:
> IMHO it would be cleaner to not to pack everything to StreamInfo.

Although I agree with you in this point. I still think the headers/trailers could be set into `StreamInfo` because it's metadata of request/response streams.

Wait some more input from other maintainers. @alyssawilk @envoyproxy/envoy-maintainers

source/common/grpc/async_client_impl.cc
```diff
@@ -95,8 +95,16 @@ void AsyncStreamImpl::initialize(bool buffer_body_for_retry) {
...
-                                            options_.parent_context.stream_info);
+  // TODO(cpakulski): Find a better way to access requestHeaders after runtime guard
+  // envoy_reloadable_features_unified_header_formatter runtime guard is deprecated.
+  // Maybe put it to parent_context?
+  parent_.metadata_parser_->evaluateHeaders(
+      headers_message_->headers(),
+      ((options_.parent_context.stream_info != nullptr) &&
+       (options_.parent_context.stream_info->getRequestHeaders() != nullptr))
+          ? *options_.parent_context.stream_info->getRequestHeaders()
+          : *Http::StaticEmptyHeaders::get().request_headers,
```
alyssawilk, line 106: If it's possible for request headers to be empty, why not just make the argument an OptRef?

source/common/router/header_formatter.h
```diff
@@ -98,5 +100,74 @@ class CompoundHeaderFormatter : public HeaderFormatter {
...
+      : formatter_(std::move(formatter)), append_(append) {}
+
+  // HttpHeaderFormatter::format
+  // Trailers are not available when HTTP headers are manipulated.
+  const std::string format(const Http::RequestHeaderMap& request_headers,
+                           const Http::ResponseHeaderMap& response_headers,
+                           const Envoy::StreamInfo::StreamInfo& stream_info) const override {
+    std::string buf;
+    buf = formatter_->format(request_headers, response_headers,
+                             *Http::StaticEmptyHeaders::get().response_trailers, stream_info, "");
```
alyssawilk, line 139: if trailers are optional, let's make the OptRef as well?

alyssawilk:
regarding stream info, I think everything else accessible from stream info is owned by stream info and the headers AFIK can't be, so I'd either be inclined to remove the APIs and pass those fields explicitly as done here, or make them shared pointers and move underlying ownership to stream info.
Not a strong opinion though and I'm happy to defer to @mattklein123 as he's back today

alyssawilk (commented):
As long as I'm looking, one thought on signatures.  Also it's awesome to have all these unified, thanks for this giant refactor!

cpakulski:
One extra thought. When I was working on Postgres filter, I wanted to add access log for database transactions: type of operation, database name, table name, etc. It was not trivial because API interface to access logging is very HTTP specific. To solve the problem I envisioned type specific access log interface. For example database like Postgres would pass transport specific information via StreamInfo, and database specific information via DB parameters like:

`log(operation, table, db_name, stream_info)`

In other words, streamInfo could be shared between different protocols and remaining parameters could be protocol specific. That is the reason I think it would be cleaner to keep Http outside the StramInfo.

Thank you for all the comments so far. I am away right now and will address them when I get back (in 3 weeks).

wbpcode:
> In other words, streamInfo could be shared between different protocols and remaining parameters could be protocol specific. That is the reason I think it would be cleaner to keep Http outside the StramInfo.

What I expected before is that we add a protocol-independent wrapper to stream info to wrap all protocol-related data.

Then, based on that, we can construct a generic format interface `std::string format(const StreamInfo::StreamInfo&)`.

Based on the generic format interface, we can go a step further and construct the generic access log (#13085).

Envery proxy can reuse the access loggers by implementing as set of formatters that implement the ``std::string format(const StreamInfo::StreamInfo&)`.

This is a very rough idea though, and I haven't figured out how to handle grpc logger yet.

wbpcode:
> Thank you for all the comments so far. I am away right now and will address them when I get back (in 3 weeks).

Thanks for you great work. This PR is very helpful and valuable. I will add `wait` tag before you get back. 😺 

/wait

tools/code_format/check_format.py
```diff
@@ -92,7 +92,8 @@
     "./source/server/admin/utils.cc", "./source/server/admin/stats_params.h",
     "./source/server/admin/stats_request.cc", "./source/server/admin/prometheus_stats.h",
     "./source/server/admin/prometheus_stats.cc", "./tools/clang_tools/api_booster/main.cc",
-    "./tools/clang_tools/api_booster/proto_cxx_utils.cc", "./source/common/version/version.cc")
+    "./tools/clang_tools/api_booster/proto_cxx_utils.cc", "./source/common/version/version.cc",
+    "./source/common/router/header_parser_utils.cc")
```
phlax, line 96: this list just got moved to tools/code_format/config.yaml - you will need to adjust that one instead

cpakulski: Thanks! Adjusted tools/code_format/config.yaml.

cpakulski:
Going back to the issue whether headers should be saved in `stream_info` or not...

> regarding stream info, I think everything else accessible from stream info is owned by stream info and the headers AFIK can't be, so I'd either be inclined to remove the APIs and pass those fields explicitly as done here, or make them shared pointers and move underlying ownership to stream info.
Not a strong opinion though and I'm happy to defer to @mattklein123 as he's back today

The current situation is that only request headers are saved in `stream_info`. They were saved only because "old-style" header formatter (the ones we are trying to deprecate in this PR) took only stream_info as a parameter. Access log formatters do not read request headers from stream_info.

After refactoring the code to use access log formatters to manipulate headers, there is no need to read the headers from `stream_info` anymore, so I was planning to remove them from stream_info after runtime guard is deprecated. In all places I needed request/response headers they were readily available either as local variables, as parameters or as data members.

source/common/router/header_formatter.h
cpakulski: @alyssawilk That is a good idea (to use `optref`), but I think it would be better to do it in a separate PR (maybe after runtime guard is deprecated). `format` method is implemented in around 50 access logs formatters and I would have to change them all. It will add probably hundreds of lines to review to this already large PR. I can open an issue to track that activity.

source/common/grpc/async_client_impl.cc
cpakulski: Agree. As I noted in TODO section just above those lines, that section needs to be reworked after runtime guard is deprecated. I am going to add the suggestion to use optref to that TODO comment.

wbpcode:
@cpakulski Thanks for still working on this. After rethinking your design again, may be it's OK to separate the protocol dependent message from the protocol independent stream info.

But could we wrapper these request and response headers/trailers to a simple `HttpMessage` and unifiled all the formatter interface?

```
// Just for example.
struct HttpMessage {
  Http::RequestHeaerMap* request_headers{};
  Http::RequestTrailerMap* request_trailers{};
  Http::ResponseHeaderMap* response_headers{};
  // ......
};
```

By this way, if we added trailer support in future (update response headers based on request trailers, logging support for the request trailers, etc.), we needn't update all these methods again.

source/common/router/header_formatter.h
```diff
@@ -86,5 +88,73 @@ class CompoundHeaderFormatter : public HeaderFormatter {
...
+
+  virtual const std::string format(const Http::RequestHeaderMap& request_headers,
+                                   const Http::ResponseHeaderMap& response_headers,
+                                   const Envoy::StreamInfo::StreamInfo& stream_info) const PURE;
+
+  /**
+   * @return bool indicating whether the formatted header should be appended to the existing
+   *              headers or replace any existing values for the header
+   */
+  virtual bool append() const PURE;
```
wbpcode, line 106: Seems this `append` method is no longer necessary? See #22723.

cpakulski:
@wbpcode 
> But could we wrapper these request and response headers/trailers to a simple HttpMessage and unifiled all the formatter interface?
// Just for example.
struct HttpMessageInfo {
  const Http::RequestHeaderMap* request_headers{};
  const Http::RequestTrailerMap* request_trailers{};
  const Http::ResponseHeaderMap* response_headers{};
  // ......
};
By this way, if we try to add trailer support in future (update response headers based on request trailers, logging support for the request trailers, etc.), we needn't update all these methods again. 🤔

I agree, but as I noted in https://github.com/envoyproxy/envoy/pull/21932#discussion_r965326784, I would prefer to do it after runtime guard is deprecated. Lots, and I mean lots of code will be removed then and interfaces will be easier to change.
What I would like to achieve in this PR is simply to funnel all `format` calls to access log implementation with minimum "bridging". I can open an issue to track/rethink interface to unified formatters.

mattklein123:
@wbpcode after you are done reviewing can you ping me for a final pass? Thank you.

envoy/http/header_evaluator.h
```diff
@@ -13,12 +13,16 @@ class HeaderEvaluator {
...
    *
    * @param headers the target header map to be mutated.
+   * @param request_headers request headers to be used in the header manipulation.
+   * @param response_headers response headers to be used in the header manipulation.
    * @param stream_info the source of values that can be used in the evaluation.
    */
   virtual void evaluateHeaders(Http::HeaderMap& headers,
+                               const Http::RequestHeaderMap& request_headers,
+                               const Http::ResponseHeaderMap& response_headers,
                                const StreamInfo::StreamInfo& stream_info) const PURE;
```
wbpcode, line 26: Could we keep both the original version and new version methods in this PR to minimize the impact on existing code.

envoy/router/router.h
```diff
@@ -53,9 +53,11 @@ class ResponseEntry {
    * example, adding or removing headers. This should only be called ONCE immediately after
    * obtaining the initial response headers.
    * @param headers supplies the response headers, which may be modified during this call.
+   * @param request_headers request headers to be used in manipulation of the response headers.
    * @param stream_info holds additional information about the request.
    */
   virtual void finalizeResponseHeaders(Http::ResponseHeaderMap& headers,
+                                       const Http::RequestHeaderMap& request_headers,
                                        const StreamInfo::StreamInfo& stream_info) const PURE;
```
wbpcode, line 61: According to previous comment/discussion (https://github.com/envoyproxy/envoy/pull/21932#discussion_r965326784 and https://github.com/envoyproxy/envoy/pull/21932#issuecomment-1241188138), we will do some interface updates in a new PR. So, I am not sure if this change here is necessary for this PR? 

Could we defer this update to the PR that will unify all the interface (after the runtime guard is expired)?

wbpcode (commented):
Thanks. This work is great and thanks again for all these contributions.

I add some comments. And my main concern is that there are too much code changes caused by the interface updates.

And according to previous discussion, we will have a dedicated PR to unify all interfaces and some others. 
So, could we defer all these unnecessary interface updates to the future dedicated PR and minimize the code change in current PR and avoid duplicate interface updates?

The whole work can be divided into three parts:
1. Make the access log formatters work for the header with minimum code change.
2. Unify all the interface after the runtime guard is expired.
3. Remove deprecated interface in the stream info (`getRequestHeaders`).

source/common/http/filter_manager.cc
```diff
@@ -914,7 +914,13 @@ void DownstreamFilterManager::sendLocalReplyViaFilterChain(
       Utility::EncodeFunctions{
           [this, modify_headers](ResponseHeaderMap& headers) -> void {
             if (streamInfo().route() && streamInfo().route()->routeEntry()) {
-              streamInfo().route()->routeEntry()->finalizeResponseHeaders(headers, streamInfo());
+              Http::RequestHeaderMap* request_headers =
+                  filter_manager_callbacks_.requestHeaders().ptr();
```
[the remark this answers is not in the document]
cpakulski: Created helper method and used const in the method body.

source/common/http/filter_manager.cc
```diff
@@ -953,7 +959,13 @@ void DownstreamFilterManager::sendDirectLocalReply(
...
```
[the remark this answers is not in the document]
cpakulski: Created helper method and used const in the method body.

source/common/http/filter_manager.cc
```diff
@@ -953,7 +959,13 @@ void DownstreamFilterManager::sendDirectLocalReply(
...
+              if (request_headers == nullptr) {
+                request_headers = Http::StaticEmptyHeaders::get().request_headers.get();
+              }
+              streamInfo().route()->routeEntry()->finalizeResponseHeaders(headers, *request_headers,
+                                                                          streamInfo());
```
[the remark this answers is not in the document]
cpakulski: Created helper method and used const in the method body.

envoy/http/header_evaluator.h
cpakulski: Unfortunately this is not possible without further complicating the code. `evaluateHeaders` method simply iterates over headers and invokes formatters. The problem is that `evaluateHeaders` does not know if the formatters are "old" style or the "new" style. The new (access log compatible) formatters need response headers to evaluate RESP(X) formatter, so they must be passed as params. If it happens that underlying formatters are "old" style, the call to them goes through [HttpHeaderFormatterBridge](https://github.com/envoyproxy/envoy/blob/17a81b5cddff3f280624c4d42abf8e50787ea77d/source/common/router/header_formatter.h#L128-L147), which simply ignores `request_headers` and `response_headers` and passes only stream_info to the formatter. That in my opinion was the least complicated way to have "old" and "new" formatters and fork between them based on runtime guard. The runtime guard controls a path when config is received. It either forwards config to "old" parser which creates "old" formatters or forwards the config to a "new" (access log's) parser  which creates "new" (actually access log's formatters). After that, during actual formatting, the code is the same. If I kept original version of `evaluateHeaders` I would have to check runtime guard each time `evaluateHeaders` is called, which would complicate code and affect performance.

I understand that this PR is large and code reviewing it is not trivial and I truly appreciate your effort to understand the changes.

envoy/router/router.h
cpakulski: Remember that "new" (access log formatter) does not take `request_headers` from `stream_info` and they have to be passed to `format` method as parameter. The best we could do would be to get request headers from `stream_info` inside the `finalizeResponseHeaders`, so the signature stays the same. The thing is that each time `finalizeResponseHeaders` is called the request headers are easy to get, because they  are either local variables or class's date members.

source/extensions/filters/common/ext_authz/ext_authz_grpc_impl.h
```diff
@@ -50,7 +50,8 @@ class GrpcClientImpl : public Client,
   // ExtAuthz::Client
   void cancel() override;
   void check(RequestCallbacks& callbacks, const envoy::service::auth::v3::CheckRequest& request,
-             Tracing::Span& parent_span, const StreamInfo::StreamInfo& stream_info) override;
+             Tracing::Span& parent_span, const Http::RequestHeaderMap& orig_request_headers,
+             const StreamInfo::StreamInfo& stream_info) override;
```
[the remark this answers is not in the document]
cpakulski: Not sure if you meant interface to `check` method or its implementation. I noticed that in several places of `check` implementation, it always passed `Http::StaticEmptyHeaders::get().response_headers` to `evaluateHeaders` so I added  a helper version of `evaluateHeaders` which does not take `response_headers`.

As for `orig_request_headers` params, all calls to `check` pass request headers received from downstream.

cpakulski:
@wbpcode Thanks for reviewing. I appreciate your effort  to understand the changes and call flow. I know it is not easy to review hundreds of lines.

> And according to previous discussion, we will have a dedicated PR to unify all interfaces and some others.
> 
> So, could we defer all these unnecessary interface updates to the future dedicated PR and minimize the code change in current PR and avoid duplicate interface updates?
> 
> The whole work can be divided into three parts:
> 
> Make the access log formatters work for the header with minimum code change.
> Unify all the interface after the runtime guard is expired.
> Remove deprecated interface in the stream info (getRequestHeaders).

This is still the plan. The whole problem here is that the "new" (access log) formatters require different set of parameters, which are not available in `stream_info`, so some API changes are unavoidable in this PR.

envoy/http/header_evaluator.h
wbpcode: What I mean by `both the original version and new version methods` is to keep the interface rather than the implementation.

We can keep an old version interface to wrapper the new implementation like this:

```
void evaluateHeaders(Http::HeaderMap& headers, const StreamInfo::StreamInfo& stream_info) const override {
       evaluateHeaders(headers, stream_info->getRequestHeaders() != nullptr 
          ? *stream_info->getRequestHeaders() : *Http::StaticEmptyHeaders::get().request_headers, *Http::StaticEmptyHeaders::get().response_headers, stream_info);
}
``` 

By this simple wrapper, we can reduce much of the code change to make this PR more clean.

envoy/router/router.h
wbpcode: > The best we could do would be to get request headers from stream_info inside the finalizeResponseHeaders, so the signature stays the same.

This is what I want for this PR. If we will have a dedicated PR for interface updates, then I'd like delay all these interface updates to the future PR.

wbpcode:
> The whole problem here is that the "new" (access log) formatters require different set of parameters, which are not available in stream_info, so some API changes are unavoidable in this PR.

Yeah, I agree. But still some method signatures could stay the same in this PR. We can update it in the future dedicated PR. At that time, we can do a more uniform design of these interfaces.

envoy/http/header_evaluator.h
cpakulski: This wrapper indeed maps 2 parameters to 4 parameters version, but I do not see how can we use it. We need an opposite operation: take 4 parameters and reduce it to 2 (when runtime guard created "old" style formatters) and this is exactly what is done in [HttpHeaderFormatterBridge.](https://github.com/envoyproxy/envoy/pull/21932#:~:text=HttpHeaderFormatterBridge)

source/common/router/config_impl.cc
```diff
@@ -783,7 +783,8 @@ void RouteEntryImplBase::finalizeRequestHeaders(Http::RequestHeaderMap& headers,
   for (const HeaderParser* header_parser : getRequestHeaderParsers(
            /*specificity_ascend=*/vhost_.globalRouteConfig().mostSpecificHeaderMutationsWins())) {
     // Later evaluated header parser wins.
-    header_parser->evaluateHeaders(headers, stream_info);
+    header_parser->evaluateHeaders(headers, headers,
+                                   *Http::StaticEmptyHeaders::get().response_headers, stream_info);
```
wbpcode, line 787: To https://github.com/envoyproxy/envoy/pull/21932#discussion_r970243905

For example, change of here can be eliminated by the wrapper, right?

source/common/grpc/google_async_client_impl.cc
```diff
@@ -182,7 +182,16 @@ void GoogleAsyncStreamImpl::initialize(bool /*buffer_body_for_retry*/) {
...
+  // TODO(cpakulski): Find a better way to access requestHeaders after runtime guard
+  // envoy_reloadable_features_unified_header_formatter runtime guard is deprecated.
+  // Maybe put it to parent_context?
+  parent_.metadata_parser_->evaluateHeaders(
+      *initial_metadata,
+      ((options_.parent_context.stream_info != nullptr) &&
+       (options_.parent_context.stream_info->getRequestHeaders() != nullptr))
+          ? *options_.parent_context.stream_info->getRequestHeaders()
+          : *Http::StaticEmptyHeaders::get().request_headers,
+      options_.parent_context.stream_info);
```
wbpcode, line 194: And here.

source/common/upstream/health_checker_impl.cc
```diff
@@ -385,7 +385,8 @@ void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::onInterval() {
                                          local_connection_info_provider_);
   stream_info.setUpstreamInfo(std::make_shared<StreamInfo::UpstreamInfoImpl>());
   stream_info.upstreamInfo()->setUpstreamHost(host_);
-  parent_.request_headers_parser_->evaluateHeaders(*request_headers, stream_info);
+  parent_.request_headers_parser_->evaluateHeaders(
+      *request_headers, *Http::StaticEmptyHeaders::get().request_headers, stream_info);
```
wbpcode, line 389: Or here

source/common/router/config_impl.cc
cpakulski: Added a wrapper with 2 params and request headers are taken from stream_info.

source/common/grpc/google_async_client_impl.cc
cpakulski: used `evaluateHeaders` wrapper with 2 params.

source/common/upstream/health_checker_impl.cc
cpakulski: Used `evaluateHeaders` wrapper with 2 params.

source/common/router/upstream_request.cc
```diff
@@ -235,9 +235,11 @@ class UpstreamFilterManager : public Http::FilterManager {
 };
 
 UpstreamRequest::UpstreamRequest(RouterFilterInterface& parent,
+                                 const Http::RequestHeaderMap& request_headers,
                                  std::unique_ptr<GenericConnPool>&& conn_pool,
                                  bool can_send_early_data, bool can_use_http3)
-    : parent_(parent), conn_pool_(std::move(conn_pool)), grpc_rq_success_deferred_(false),
+    : parent_(parent), request_headers_(request_headers), conn_pool_(std::move(conn_pool)),
+      grpc_rq_success_deferred_(false),
```
wbpcode, line 242: This change also unnecessary. The upstream request can access request headers by `downstreamHeaders()` of `RouterFilterInterface`.

source/common/router/upstream_request.h
```diff
@@ -179,6 +180,8 @@ class UpstreamRequest : public Logger::Loggable<Logger::Id::router>,
   StreamInfo::StreamInfo& streamInfo() { return stream_info_; }
   bool hadUpstream() const { return had_upstream_; }
 
+  const Http::RequestHeaderMap& getRequestHeaders() const { return request_headers_; }
+
```
wbpcode, line 184: And this is unnecessary for above reason.

source/common/tcp_proxy/upstream.cc
```diff
@@ -290,7 +290,9 @@ void Http2Upstream::setRequestEncoder(Http::RequestEncoder& request_encoder, boo
                           Http::Headers::get().ProtocolValues.Bytestream);
   }
 
-  config_.headerEvaluator().evaluateHeaders(*headers, downstream_info_);
+  config_.headerEvaluator().evaluateHeaders(
+      *headers, *Http::StaticEmptyHeaders::get().request_headers,
+      *Http::StaticEmptyHeaders::get().response_headers, downstream_info_);
```
wbpcode, line 295: not sure should we also create a simple wrapper function for evaluator to eliminate this change?

source/extensions/filters/http/local_ratelimit/local_ratelimit.cc
```diff
@@ -130,16 +130,18 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
   config->stats().rate_limited_.inc();
 
   if (!config->enforced()) {
-    config->requestHeadersParser().evaluateHeaders(headers, decoder_callbacks_->streamInfo());
+    config->requestHeadersParser().evaluateHeaders(headers, headers,
+                                                   decoder_callbacks_->streamInfo());
```
wbpcode, line 134: May be use the two parameters help function here?

source/extensions/filters/http/local_ratelimit/local_ratelimit.cc
```diff
@@ -130,16 +130,18 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
...
 
   config->stats().enforced_.inc();
 
   decoder_callbacks_->sendLocalReply(
       config->status(), "local_rate_limited",
-      [this, config](Http::HeaderMap& headers) {
-        config->responseHeadersParser().evaluateHeaders(headers, decoder_callbacks_->streamInfo());
+      [this, &headers, config](Http::HeaderMap& response_headers) {
+        config->responseHeadersParser().evaluateHeaders(response_headers, headers,
+                                                        decoder_callbacks_->streamInfo());
```
wbpcode, line 144: ditto.

test/extensions/filters/common/ext_authz/ext_authz_grpc_impl_test.cc
```diff
@@ -90,7 +90,8 @@ TEST_F(ExtAuthzGrpcClientTest, AuthorizationOk) {
 
   envoy::service::auth::v3::CheckRequest request;
   expectCallSend(request);
-  client_->check(request_callbacks_, request, Tracing::NullSpan::instance(), stream_info_);
+  client_->check(request_callbacks_, request, Tracing::NullSpan::instance(),
+                 *Http::StaticEmptyHeaders::get().request_headers, stream_info_);
```
wbpcode, line 94: There are many similar changes. And these changes make no sense for now. I mean, we can delay these to the future dedicated PR.

wbpcode (dismissed):
Thanks for you hard work. :)

My mainly concern is still to keep the PR simpler and cleaner. So, at least in this PR, we should use request headers in `stream info` to avoid unnecessary method signature updates (except the unavoidable updates).

IMO, we should delay these signature updates to the future dedicated PR as much as possible.

envoy/router/router.h
cpakulski: Reverted to original signature and request headers are retrieved from `streamInfo` in `finalizeResponseHeaders` body.

source/common/router/header_parser.cc
```diff
@@ -317,10 +369,13 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
...
                                                          bool do_formatting) const {
   Http::HeaderTransforms transforms;
+  auto empty_req_map = Http::RequestHeaderMapImpl::create();
+  auto empty_response_map = Http::ResponseHeaderMapImpl::create();
 
   for (const auto& [key, entry] : headers_to_add_) {
     if (do_formatting) {
-      const std::string value = entry.formatter_->format(stream_info);
+      const std::string value =
+          entry.formatter_->format(*empty_req_map, *empty_response_map, stream_info);
```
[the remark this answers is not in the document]
cpakulski: Yeah, good idea. Thanks! Done.

source/common/router/header_parser.h
```diff
@@ -53,9 +54,48 @@ class HeaderParser : public Http::HeaderEvaluator {
...
+  void evaluateHeaders(Http::HeaderMap& headers, const Http::RequestHeaderMap& request_headers,
+                       const StreamInfo::StreamInfo& stream_info) const {
+    evaluateHeaders(headers, request_headers,
+                    *Http::StaticEmptyHeaders::get().response_headers.get(), stream_info);
+  }
+  void evaluateHeaders(Http::HeaderMap& headers, const Http::RequestHeaderMap& request_headers,
+                       const StreamInfo::StreamInfo* stream_info) const {
+    evaluateHeaders(headers, request_headers,
+                    *Http::StaticEmptyHeaders::get().response_headers.get(), stream_info);
+  }
```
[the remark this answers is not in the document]
cpakulski: Removed.

source/common/router/router.cc
```diff
@@ -426,7 +426,8 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
           if (!new_path.empty() && add_location) {
             response_headers.addReferenceKey(Http::Headers::get().Location, new_path);
           }
-          direct_response->finalizeResponseHeaders(response_headers, callbacks_->streamInfo());
+          direct_response->finalizeResponseHeaders(response_headers, request_headers,
+                                                   callbacks_->streamInfo());
```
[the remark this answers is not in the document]
cpakulski: request_headers are read from `stream_info` inside the `finalizeResponseHeaders`, so no changes to method's signature.

source/common/router/router.cc
```diff
@@ -1406,10 +1409,11 @@ void Filter::onUpstreamHeaders(uint64_t response_code, Http::ResponseHeaderMapPt
...
       const RetryStatus retry_status = retry_state_->shouldRetryHeaders(
           *headers, *downstream_headers_,
-          [this, can_use_http3 = upstream_request.upstreamStreamOptions().can_use_http3_,
+          [this, &request_headers = upstream_request.getRequestHeaders(),
+           can_use_http3 = upstream_request.upstreamStreamOptions().can_use_http3_,
            had_early_data = upstream_request.upstreamStreamOptions().can_send_early_data_](
               bool disable_early_data) -> void {
-            doRetry((disable_early_data ? false : had_early_data), can_use_http3);
+            doRetry(request_headers, (disable_early_data ? false : had_early_data), can_use_http3);
           });
```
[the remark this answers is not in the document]
cpakulski: Reverted `doRetry` to original signature.

source/common/router/router.cc
```diff
@@ -1502,7 +1506,8 @@ void Filter::onUpstreamHeaders(uint64_t response_code, Http::ResponseHeaderMapPt
   // TODO(zuercher): If access to response_headers_to_add (at any level) is ever needed outside
   // Router::Filter we'll need to find a better location for this work. One possibility is to
   // provide finalizeResponseHeaders functions on the Router::Config and VirtualHost interfaces.
-  route_entry_->finalizeResponseHeaders(*headers, callbacks_->streamInfo());
+  route_entry_->finalizeResponseHeaders(*headers, upstream_request.getRequestHeaders(),
+                                        callbacks_->streamInfo());
```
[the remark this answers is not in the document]
cpakulski: Done. Reverted `finalizeResponseHeaders` to the original signature.

source/common/router/router.cc
```diff
@@ -1788,7 +1793,8 @@ void Filter::runRetryOptionsPredicates(UpstreamRequest& retriable_request) {
   }
 }
 
-void Filter::doRetry(bool can_send_early_data, bool can_use_http3) {
+void Filter::doRetry(const Http::RequestHeaderMap& request_headers, bool can_send_early_data,
+                     bool can_use_http3) {
```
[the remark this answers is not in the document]
cpakulski: Reverted `doRetry` to original form.

source/common/router/upstream_request.cc
cpakulski: Reverted to the  original form.

source/common/router/upstream_request.h
cpakulski: Removed.

source/common/tcp_proxy/upstream.cc
cpakulski: I cannot use `evaluateHeaders` wrappers defined  in `router`, because `headerEvaluator() `returns reference to `HeaderEvaluator` pure virtual class, not `HeaderEvaluatorImpl`. I suggest we leave it as is for now.

source/extensions/filters/http/local_ratelimit/local_ratelimit.cc
cpakulski: Yes. Done.

source/extensions/filters/http/local_ratelimit/local_ratelimit.cc
cpakulski: Done.

test/extensions/filters/common/ext_authz/ext_authz_grpc_impl_test.cc
cpakulski: Reverted check to the original form, so all those changes were also reverted.

cpakulski:
@wbpcode Thanks for reviewing. I addressed all your comments and reverted most methods to the original form which reduced size of this PR by around 300+ lines :).

However, it sometimes comes at a cost of retrieving request headers from `streamInfo` in methods' bodies and adding mocked methods in tests. So in effect the new code relies on `streamInfo's` `getRequestHeaders` more than before. But I agree with you that the code is easier to review and understand with this approach. I also believe that when it comes to removing `streamInfo's` `getRequestHeaders` it will be easier to do it module by module instead of in one big step, so it seems to be the right approach.

test/common/router/header_formatter_test.cc
```diff
@@ -1636,7 +1754,8 @@ request_headers_to_remove: ["x-foo-header"]
   Http::TestRequestHeaderMapImpl header_map{{"x-foo-header", "foo"}};
   NiceMock<Envoy::StreamInfo::MockStreamInfo> stream_info;
 
-  req_header_parser->evaluateHeaders(header_map, stream_info);
+  req_header_parser->evaluateHeaders(
+      header_map, header_map, *Http::StaticEmptyHeaders::get().response_headers, stream_info);
   EXPECT_EQ("bar", header_map.get_("x-foo-header"));
```
wbpcode, line 1759: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1666,23 +1785,26 @@ match: { prefix: "/new_endpoint" }
 
   {
     Http::TestResponseHeaderMapImpl header_map;
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1789: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1666,23 +1785,26 @@ match: { prefix: "/new_endpoint" }
...
     EXPECT_EQ("foo", header_map.get_("x-foo-header"));
     EXPECT_EQ("bar", header_map.get_("x-bar-header"));
     EXPECT_EQ("per", header_map.get_("x-per-header"));
   }
 
   {
     Http::TestResponseHeaderMapImpl header_map{{"x-foo-header", "exist-foo"}};
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1798: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1666,23 +1785,26 @@ match: { prefix: "/new_endpoint" }
...
     EXPECT_EQ(2, header_map.get(Http::LowerCaseString("x-foo-header")).size());
     EXPECT_EQ("bar", header_map.get_("x-bar-header"));
     EXPECT_EQ("per", header_map.get_("x-per-header"));
   }
 
   {
     Http::TestResponseHeaderMapImpl header_map{{"x-bar-header", "exist-bar"}};
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1807: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1691,7 +1813,8 @@ match: { prefix: "/new_endpoint" }
 
   {
     Http::TestResponseHeaderMapImpl header_map{{"x-per-header", "exist-per"}};
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1817: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1720,21 +1843,24 @@ match: { prefix: "/new_endpoint" }
 
   {
     Http::TestResponseHeaderMapImpl header_map;
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1847: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1720,21 +1843,24 @@ match: { prefix: "/new_endpoint" }
...
     EXPECT_EQ("foo", header_map.get_("x-foo-header"));
     EXPECT_EQ("bar", header_map.get_("x-bar-header"));
   }
 
   {
     Http::TestResponseHeaderMapImpl header_map{{"x-foo-header", "exist-foo"}};
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1855: ditto?

test/common/router/header_formatter_test.cc
```diff
@@ -1720,21 +1843,24 @@ match: { prefix: "/new_endpoint" }
...
     EXPECT_EQ(2, header_map.get(Http::LowerCaseString("x-foo-header")).size());
     EXPECT_EQ("bar", header_map.get_("x-bar-header"));
   }
 
   {
     Http::TestResponseHeaderMapImpl header_map{{"x-bar-header", "exist-bar"}};
-    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    resp_header_parser->evaluateHeaders(
+        header_map, *Http::StaticEmptyHeaders::get().request_headers, header_map, stream_info);
```
wbpcode, line 1863: ditto?

wbpcode (commented):
Thanks. It's looks good overall for me. Only some more minor clears are necessary.
It's seems like there are lots of comments, but just some minor problem. :)

**Thanks again for all your contribution and patience.** 🌹

source/extensions/filters/http/local_ratelimit/local_ratelimit.cc
```diff
@@ -138,8 +138,9 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
 
   decoder_callbacks_->sendLocalReply(
       config->status(), "local_rate_limited",
-      [this, config](Http::HeaderMap& headers) {
-        config->responseHeadersParser().evaluateHeaders(headers, decoder_callbacks_->streamInfo());
+      [this, config](Http::HeaderMap& response_headers) {
+        config->responseHeadersParser().evaluateHeaders(response_headers,
+                                                        decoder_callbacks_->streamInfo());
```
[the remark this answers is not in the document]
cpakulski: Change is not required for now. Reverted.

test/common/router/header_formatter_test.cc
```diff
@@ -1107,9 +1136,10 @@ TEST(HeaderParserTest, TestParseInternal) {
   absl::optional<Envoy::Http::Protocol> protocol = Envoy::Http::Protocol::Http11;
   ON_CALL(stream_info, protocol()).WillByDefault(ReturnPointee(&protocol));
 
-  std::shared_ptr<NiceMock<Envoy::Upstream::MockHostDescription>> host(
-      new NiceMock<Envoy::Upstream::MockHostDescription>());
-  stream_info.upstreamInfo()->setUpstreamHost(host);
+  // Get the pointer to MockHostDescription.
+  std::shared_ptr<const Upstream::MockHostDescription> hd =
+      std::dynamic_pointer_cast<const Upstream::MockHostDescription>(
+          stream_info.upstreamInfo()->upstreamHost());
```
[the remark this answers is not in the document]
cpakulski: Not necessary for now. Reverted.

test/common/router/header_formatter_test.cc
```diff
@@ -1203,7 +1309,8 @@ match: { prefix: "/new_endpoint" }
       HeaderParser::configure(parseRouteFromV3Yaml(yaml).request_headers_to_add());
   Http::TestRequestHeaderMapImpl header_map{{":method", "POST"}};
   NiceMock<Envoy::StreamInfo::MockStreamInfo> stream_info;
-  req_header_parser->evaluateHeaders(header_map, stream_info);
+  req_header_parser->evaluateHeaders(
+      header_map, header_map, *Http::StaticEmptyHeaders::get().response_headers, stream_info);
```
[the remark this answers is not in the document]
cpakulski: Not necessary for now. Wrapper with 2 params can be used. The same below.

test/common/router/header_formatter_test.cc
```diff
@@ -1324,13 +1436,15 @@ match: { prefix: "/new_endpoint" }
...
+  std::shared_ptr<const Upstream::MockHostDescription> hd =
+      std::dynamic_pointer_cast<const Upstream::MockHostDescription>(
+          stream_info.upstreamInfo()->upstreamHost());
   auto metadata = std::make_shared<envoy::config::core::v3::Metadata>();
-  stream_info.upstreamInfo()->setUpstreamHost(host);
-  ON_CALL(*host, metadata()).WillByDefault(Return(metadata));
-  req_header_parser->evaluateHeaders(header_map, stream_info);
+  ON_CALL(*hd, metadata()).WillByDefault(Return(metadata));
+  req_header_parser->evaluateHeaders(
+      header_map, header_map, *Http::StaticEmptyHeaders::get().response_headers, stream_info);
```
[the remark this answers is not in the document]
cpakulski: Not necessary. Reverted.

test/common/router/header_formatter_test.cc
cpakulski: Expanded the test to make response headers to be evaluated dependent on both request and response headers. Therefore `evaluateHeaders` must be called with all 4 params.

cpakulski:
@wbpcode Thanks for another review! I reverted most of calls  to `evaluateHeader` in tests to the version with 2 params. It was possible because in those cases headers to be generated did not depend on response headers (did not have for example `RESP(x)` formatter). I extended one test case to be dependent on both, request (`REQ(X)`) and response (`RESP(X)`) headers and therefore all 4 params must be passed to `evaluateHeaders` method.

wbpcode:
Thanks. Great.

wbpcode:
cc @mattklein123 I think this is ready for a final review.

Some background: according to previous discussion, this PR will not change the method signature as far as possible to reduce impact to exist code tree (except some unavoidable updates). We will have a dedicated PR to unify all these interfaces after the runtime guard is expired.

Something I think is important: If the runtime guard (default to true) is OK for the scene. It will break users' configuration immediately if the user does not change the runtime guard value to false. I think this behavior change has a bigger impact.

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,191 +574,34 @@ used to delimit variable names.
...
 
-%DOWNSTREAM_REMOTE_ADDRESS%
-    Remote address of the downstream connection. If the address is an IP address it includes both
-    address and port.
-
-    .. note::
-
-      This may not be the physical remote address of the peer if the address has been inferred from
-      :ref:`Proxy Protocol filter <config_listener_filters_proxy_protocol>` or :ref:`x-forwarded-for
-      <config_http_conn_man_headers_x-forwarded-for>`.
```
wbpcode, line 587: nit: should we just keep these old docs for now and removed after the runtime guard is expired? 🤔

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,191 +574,34 @@ used to delimit variable names.
   doubling it. For example, to emit a header with the value ``100%``, the custom header value in
   the Envoy configuration must be ``100%%``.
 
-Supported variable names are:
+All HTTP :ref:`Command Operators <config_access_log_command_operators>` used for access logging may be specified
+in Custom request/response headers. However, depending where a particular Command Operator is used, the context needed for
```
mattklein123, line 578:
```suggestion
in custom request/response headers. However, depending where a particular command operator is used, the context needed for
```

docs/root/configuration/http/http_conn_man/headers.rst
```diff
@@ -574,191 +574,34 @@ used to delimit variable names.
...
+the Operator may not be available and the produced output is empty string. For example, the following configuration
```
mattklein123, line 579:
```suggestion
the operator may not be available and the produced output is empty string. For example, the following configuration
```

source/common/router/header_parser.cc
```diff
@@ -32,11 +35,41 @@ enum class ParserState {
...
+  // HTTP/1 and HTTP/2. It could arguably be allowed on the response path.
+  if (!Http::HeaderUtility::isModifiableHeader(key)) {
+    throw EnvoyException(":-prefixed or host headers may not be modified");
+  }
+
+  // UPSTREAM_METADATA and DYNAMIC_METADATA must be translated from JSON ["a", "b"] format to colon
+  // format (a:b)
+  std::string final_header_value = HeaderParser::translateMetadataFormat(header_value.value());
+  // Change PER_REQUEST_STATE to FILTER_STATE.
+  final_header_value = HeaderParser::translatePerRequestState(final_header_value);
```
mattklein123, line 58: Do we output warnings for deprecated usages? This is going to break someone in the future if/when any of the deprecated operators get used?

mattklein123 (changes requested):
Sweet! A few small comments, thank you.

/wait

source/common/router/header_parser.cc
cpakulski: I was planning to add warnings when runtime guard is retired. There are translators from "old" format and they will be kept for about 1 year from the moment warnings are added. After that, the "old" format will be completely deprecated. Is this OK?

mattklein123: I would just add them now since the runtime guard defaults to enabled.

cpakulski: Added. Thanks!

docs/root/configuration/http/http_conn_man/headers.rst
cpakulski: Done.

source/common/router/header_parser_utils.cc
```diff
@@ -0,0 +1,82 @@
...
+// Related to issue 20389. Header formatters are parsed and processed by formatters defined in
+// source/common/formatter/substitution_formatter.cc. For backwards compatibility UPSTREAM_METADATA
+// and UPSTREAM_METADATA format must be changed. Those formatters used to take a JSON format like
+// UPSTREAM_METADATA(["a", "b"]) and substitution formatters use UPSTREAM_METADATA(a:b) format.
+// This translator translates UPSTREAM_METADATA and DYNAMIC_METADATA from JSON format to colon
+// format.
+// TODO(cpakulski): Eventually JSON format should be deprecated in favor of colon format.
+std::string HeaderParser::translateMetadataFormat(const std::string& header_value) {
+  const std::regex command_w_args_regex(
+      R"EOF(%(UPSTREAM|DYNAMIC)_METADATA\(\s*(\[(?:.|\r?\n)+?\]\s*)\)%)EOF");
```
jmarantz, line 20: drive-by after-the fact comment.

I'm not sure if this hits the data plane but it would be trivial to use CONSTRUCT_ON_FIRST_USE to compile this regex only once, rather than every time it's called.

Less trivial might be converting it to RE2 for improved speed, or implementing the above regex using direct code. Not sure how important this is.

jmarantz: sorry it's obvious from TRY_ASSERT_MAIN_THREAD this does not happen on the data plane. Nevertheless it does happen during operation on xDS updates. Not sure if it's material relative to all that goes on during one.

cpakulski: Thanks for reviewing it. This does not happen at the data plane, but during the configuration time. I will convert it to CONSTRUCT_ON_FIRST_USE. 

This is a translator between legacy format of specifying which metadata should be extracted and put into header and the new format. I plan to deprecate legacy format in about 1 year. 
But, the potential problem is that each configured header formatter is matched against this regex. Not sure how efficient is regex. Maybe it should be preceded by simple 
```
(if header_value.StartsWith("UPSTREAM) || header_value.StartsWith("DYNAMIC")) {
  // match regex.

```
I will check it with microbenchmark.

jmarantz: The control-plane can also have issues with being compute-bound.

std::regex is particularly bad even if you don't recompile on every call. It allocates memory during parsing, for example. RE2 is much better, but directly parsing in C++ would likely be faster still. Was there any reason not to use RE2? I thought we had close to eliminated std::regex from the codebase.

cpakulski: Thanks for that info. There was no particular reason not to use RE2 - just my knowledge on that topic was limited. Let me benchmark it and rework to either to pure C++ or RE2. I will open a new PR and ask you for review.

> I thought we had close to eliminated std::regex from the codebase.

There are still few files which use std::regex. I had to modify tools/code_format/config.yaml to add a new file with std::regex. Now I know why.

jmarantz: Yeah the purpose of allow-lists in that file is to limited the spread of bad practice :)

PR #21933, opened 2022-06-28
dockerfile: Add default platform and use 0 for uid/gid

phlax:

Commit Message:
Additional Description:

following up on https://github.com/envoyproxy/envoy/issues/21899

switching the `COPY --chown` from `root` -> `0` resolved the following error i hit locally:

```console
unable to convert uid/gid chown string to host mapping: can't find uid for user root: open /var/lib/docker/overlay2/cd074b39f9fa59a8731b1b92200dfa7c652dd32739edaa52d595d75f0184c91d/merged/etc/passwd: no such file or directory             

```

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/Dockerfile-envoy
+++ b/ci/Dockerfile-envoy
@@ -5,12 +5,12 @@ ARG ENVOY_VRP_BASE_IMAGE=envoy
 
 FROM scratch AS binary
 
-ARG TARGETPLATFORM
+ARG TARGETPLATFORM=linux/amd64
 ARG ENVOY_BINARY=envoy
 ARG ENVOY_BINARY_SUFFIX=_stripped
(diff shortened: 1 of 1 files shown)


wbpcode (approved):
LGTM. Thanks. cc @lizan for merge or final pass.

phlax:
@lizan this caused a significant drop in the size of the distroless image - not sure why exactly - perhaps using `0` rather than `root` saves a layer

im going to test the output from the docker build job now

phlax:
tested the image - all seems good - the loaded image size is the same

checked inside the tarballs before/after and there is 1 less layer - all good i think

PR #21940, opened 2022-06-29
listener: enable UdpListenerWorkerRouter to support multiple addresses

soulxu:
Commit Message: listener: enable UdpWorkerRouter to support multiple addresses
Additional Description:
This PR enable the `ListenerImpl` to create `UdpListenerWorkerRouter ` for each address. And pass the correct `UdpListenerWorkerRouter` to the `ActiveUdpListener`.

Risk Level: high
Testing: unittest
Docs Changes: n/a
Release Notes: n/a


source/server/listener_impl.h
```diff
@@ -494,6 +499,7 @@ class ListenerImpl final : public Network::ListenerConfig,
 
   Quic::QuicStatNames& quic_stat_names_;
   MissingListenerConfigStats missing_listener_config_stats_;
+  uint32_t concurrency_;
```
mattklein123, line 502: nit: const. As an aside it would be nice to not have to store this in every listener, but there probably isn't a great way to do that without passing through all server options so I guess it's ok.

mattklein123 (changes requested):
LGTM with small comment, thank you.

/wait

source/server/listener_impl.h
soulxu: I think I can get that from the server instance directly, let me try, also thanks point this out, it indeed better to not pass through it, and it doesn't feel like something the `ListenerImpl` need to hold one.

Issue #11569, opened 2020-06-12
Support tunneling TCP over HTTP/2 behind HTTP proxy


irozzo-1A:
Description:

#1451 enables the usage of envoy to tunnel raw TCP over HTTP/2. I would like to use this feature in a scenario in which the tunnel should pass through an HTTP proxy (supporting HTTP/1.x CONNECT). This would mean that I would need "double tunneling", something like:

```
client--[raw TCP]-->L2 Envoy--[TCP tunneled over HTTP/2 tunneled over HTTP/1.1]-->HTTP Proxy-->[TCP tunneled over HTTP/2]-->L1 Envoy--[raw TCP]-->server
```

I know HTTP/1.x connect are not supported yet (i.e. #11308), but I would like to have an early feedback about the possibility of supporting this.

mattklein123:
cc @alyssawilk

alyssawilk:
Why would you need double tunneling?   Could you simply have the middle proxy forward the connect on untouched?  Envoy at least supports incoming CONNECT being sent as outbound CONNECT, at which point you could just encapsulate your TCP in a CONNECT request at the front proxy, forward it through the middle one, and terminate the CONNECT at the last one in line.

irozzo-1A:
Thx @alyssawilk for your reply. This setup would be needed in environments where we don't have the hands on the middle HTTP proxy, that could be anything e.g. Squid, Bluecoat proxies, Fortigates, etc. The advantage is that it would be transparent to the middle HTTP proxy and wouldn't require any specific configuration on it.
I agree that having full control on the environment would make it easier.

lambdai:
I am adding pluggable backend to tcp proxy, so potentially you could have additional control on terminating raw TCP at L2 Envoy.

I feel the double tunnel is extremely complex. I see your point but it's likely be supported by your own extension using the pluggable framework.

My PR on the pluggable tcp proxy backend will be create today. Now I am examples how it could be used and the compatibility with the current tcp proxy.

irozzo-1A:
Thx @lambdai for your reply! I will take a look to the TCP pluggable backend.

In the meantime I was thinking about a possible workaround (probably less clean than the option you proposed). The idea is to use two stages and a Unix domain socket between them.
* First stage is composed by a listener binding to a TCP socket with a TCP proxy filter taking care of the inner tunnel addressed to the L1 Envoy with a UDS upstream.
[...]

PR #21942, opened 2022-06-29
Adding direct connect support.

alyssawilk:
Creates a transport socket for HTP/1.1 proxy support.

With the combination of the transport socket, and a filter putting the proxy stream info in place this will
1) redirect TCP connections to the proxy IP address
2) prefix TLS connections with cleartext CONECT headers to the destination host, and strip CONNECT response
3) change cleartext HTTP/1.1 to send fully qualified URLs

Risk Level: medium (intended as a no op but it does have data plane refactory)
Testing: new unit, integration tests
Docs Changes: n/a


api/envoy/extensions/transport_sockets/http_11_proxy/v3/upstream_http_11_connect.proto
```diff
@@ -0,0 +1,31 @@
...
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Upstream HTTP/1.1 Proxy]
+// [#extension: envoy.transport_sockets.http_11_proxy]
+
+// Configuration for HTTP/1.1 proxy transport sockets.
+// If this is configured and an intermediate filter adds proxy metadata to the
+// stream info then
+//   - Upstream connections will be directed to the specified proxy address rather
+//     than the host's address
```
RyanTheOptimist, line 23: Does this avoid the DNS lookup for the server's address? I'm guessing not. I think that would be a good optimization to add because the DNS lookup might fail (in the worst case) or might take unnecessary time (in the best case). Not something to do now, but something to maybe add a TODO / issue about?

envoy/network/transport_socket.h
```diff
@@ -230,6 +230,22 @@ class TransportSocketOptions {
    */
   virtual absl::optional<Network::ProxyProtocolData> proxyProtocolOptions() const PURE;
 
+  // Information for use by the http_11_proxy transport socket.
+  struct ProxyInfo {
```
RyanTheOptimist, line 234: nit: Http11ProxyInfo, maybe?

envoy/network/transport_socket.h
```diff
@@ -230,6 +230,22 @@ class TransportSocketOptions {
...
+    // the underlying transport is TLS.
+    std::string hostname;
+    // The address of the proxy, where connections should be routed to.
+    Network::Address::InstanceConstSharedPtr proxy_address;
+  };
+
+  /**
+   * @return any proxy information if sending to an intermediate proxy over HTTP/1.1.
+   */
+  virtual const std::unique_ptr<const ProxyInfo>& proxyInfo() const PURE;
```
RyanTheOptimist, line 247: nit: Should this perhaps be named http11ProxyInfo? Not sure...

RyanTheOptimist, line 247: Instead of returning a ref to a unique ptr, how about returning an OptRef?

source/common/network/filter_state_proxy_info.h
```diff
@@ -0,0 +1,31 @@
...
+#include "absl/strings/string_view.h"
+
+namespace Envoy {
+namespace Network {
+
+/**
+ * Information which filters can add if they detect the stream should go
+ * upstream through an HTTP/1.1 proxy.
+ */
+class FilterStateProxyInfo : public StreamInfo::FilterState::Object {
```
RyanTheOptimist, line 15: nit: It looks like the naming convention for such object is usually <Something>FilterState. So how about "ProxyInfoFilterState"?

source/common/network/filter_state_proxy_info.h
```diff
@@ -0,0 +1,31 @@
...
+public:
+  static const std::string& key();
```
RyanTheOptimist, line 17: nit: comment.

source/common/network/filter_state_proxy_info.h
```diff
@@ -0,0 +1,31 @@
...
+
+  FilterStateProxyInfo(absl::string_view hostname, Network::Address::InstanceConstSharedPtr address)
+      : hostname_(hostname), address_(address) {}
+  Network::Address::InstanceConstSharedPtr address() const { return address_; }
```
RyanTheOptimist, line 21: Totally your call but it does seem like "value()" is the common name for this method. So I'm torn between "address()" being an obviously better named but "value()" being a more consistent name. I guess better should trump consistent, so consider this a stream of consciousness that you should ignore!

source/common/network/filter_state_proxy_info.h
```diff
@@ -0,0 +1,31 @@
...
+
+private:
+  // The hostname of this individual request.
+  const std::string hostname_;
```
RyanTheOptimist, line 25: Is this field used?

source/common/network/transport_socket_options_impl.cc
```diff
@@ -77,9 +79,15 @@ TransportSocketOptionsConstSharedPtr TransportSocketOptionsUtility::fromFilterSt
     proxy_protocol_options.emplace(typed_data->value());
   }
 
+  if (auto typed_data =
+          filter_state->getDataReadOnly<FilterStateProxyInfo>(FilterStateProxyInfo::key());
+      typed_data != nullptr) {
+    proxy_info = std::make_unique<TransportSocketOptions::ProxyInfo>("host", typed_data->address());
```
RyanTheOptimist, line 85: Ah, since this is passing in `"host"` I'm guessing it's an unused argument?

source/common/upstream/upstream_impl.cc
```diff
@@ -279,6 +279,16 @@ Network::TransportSocketFactory& HostDescriptionImpl::resolveTransportSocketFact
...
+  // If the transport socket options indicate the connection should be
+  // redirected to a proxy, create the TCP connection to the proxy's address not
+  // the host's address.
+  if (transport_socket_options && transport_socket_options->proxyInfo()) {
+    return {createConnection(dispatcher, cluster(),
+                             transport_socket_options->proxyInfo()->proxy_address,
+                             {transport_socket_options->proxyInfo()->proxy_address},
+                             transportSocketFactory(), options, transport_socket_options),
+            shared_from_this()};
+  }
```
RyanTheOptimist, line 291: Up to you but consider having only a single call to createConnection(), but pass in the address/address_list as local variables which default to address() and addressList() and are overridden in the case of the proxy, to make it clear that only the addresses are changing. (Or not, if that doesn't seem cleaner)

source/extensions/transport_sockets/http_11_proxy/config.cc
```diff
@@ -0,0 +1,43 @@
...
+Network::TransportSocketFactoryPtr
+UpstreamHttp11ConnectSocketConfigFactory::createTransportSocketFactory(
+    const Protobuf::Message& message,
+    Server::Configuration::TransportSocketFactoryContext& context) {
+  const auto& outer_config = MessageUtil::downcastAndValidate<
+      const envoy::extensions::transport_sockets::http_11_proxy::v3::Http11ProxyUpstreamTransport&>(
+      message, context.messageValidationVisitor());
+  auto& inner_config_factory = Config::Utility::getAndCheckFactory<
+      Server::Configuration::UpstreamTransportSocketConfigFactory>(outer_config.transport_socket());
+  ProtobufTypes::MessagePtr inner_factory_config = Config::Utility::translateToFactoryConfig(
```
RyanTheOptimist, line 24: LOL, why is this not auto? :) (But I'm happy with explicit types so don't ever change! :>)

source/extensions/transport_sockets/http_11_proxy/connect.h
```diff
@@ -0,0 +1,57 @@
...
+  UpstreamHttp11ConnectSocketFactory(Network::TransportSocketFactoryPtr transport_socket_factory);
+
+  // Network::TransportSocketFactory
+  Network::TransportSocketPtr
+  createTransportSocket(Network::TransportSocketOptionsConstSharedPtr options) const override;
+  void hashKey(std::vector<uint8_t>& key,
+               Network::TransportSocketOptionsConstSharedPtr options) const override;
+
+private:
+  //  Http11ConnectConfig config_;
```
RyanTheOptimist, line 51: nit: Should this be removed?

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+
+const std::string& UpstreamHttp11ConnectSocket::key() {
+  CONSTRUCT_ON_FIRST_USE(std::string, "envoy.network.transport_socket.http_11_proxy.address");
+}
+
+UpstreamHttp11ConnectSocket::UpstreamHttp11ConnectSocket(
+    Network::TransportSocketPtr&& transport_socket,
+    Network::TransportSocketOptionsConstSharedPtr options)
+    : PassthroughSocket(std::move(transport_socket)), options_(options) {
+  if (options_ && options_->proxyInfo() && transport_socket_->ssl()) {
```
RyanTheOptimist, line 25: I don't think I have a clear picture of how this interacts with socket reuse for multiple requests. For example if the first request via a proxy is for an http:// URL and the second is for an https:// URL, the first request should be sent with a fully qualified URL and no CONNECT, but the second should be a CONNECT request (and then a non-fully qualified URL). Is such a thing possible in Envoy?

source/extensions/transport_sockets/http_11_proxy/connect.h
```diff
@@ -0,0 +1,57 @@
...
+  Network::IoResult doRead(Buffer::Instance& buffer) override;
+
+private:
+  void generateHeader();
+  Network::IoResult writeHeader();
+
+  Network::TransportSocketOptionsConstSharedPtr options_;
+  Network::TransportSocketCallbacks* callbacks_{};
+  Buffer::OwnedImpl header_buffer_{};
+  bool stripped_connect_{};
```
RyanTheOptimist, line 37: nit: comment? Also, I think = false would be slightly clearer than {}, but up to you. (I'd also recommend = nullptr for callbacks_ and would be inclined to remove {} for header_buffer_)

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+void UpstreamHttp11ConnectSocket::setTransportSocketCallbacks(
+    Network::TransportSocketCallbacks& callbacks) {
+  transport_socket_->setTransportSocketCallbacks(callbacks);
+  callbacks_ = &callbacks;
+}
+
+Network::IoResult UpstreamHttp11ConnectSocket::doWrite(Buffer::Instance& buffer, bool end_stream) {
+  if (header_buffer_.length() > 0) {
+    auto header_res = writeHeader();
+    return header_res;
```
RyanTheOptimist, line 42: nit: perhaps just `return writeHeader()`

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+  } else if (stripped_connect_) {
+    // Don't pass events up until the connect response is read because TLS reads
+    // kick off writes which don't pass through the transport socket.
+    return transport_socket_->doWrite(buffer, end_stream);
+  }
+  return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
```
RyanTheOptimist, line 48: Let's do the early return pattern here:

```
if (header_buffer_.empty()) {
  return writeHeader();
}
if (stripped_connect_) {
}
return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
```

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+}
+
+void UpstreamHttp11ConnectSocket::setTransportSocketCallbacks(
+    Network::TransportSocketCallbacks& callbacks) {
+  transport_socket_->setTransportSocketCallbacks(callbacks);
+  callbacks_ = &callbacks;
+}
+
+Network::IoResult UpstreamHttp11ConnectSocket::doWrite(Buffer::Instance& buffer, bool end_stream) {
+  if (header_buffer_.length() > 0) {
```
RyanTheOptimist, line 40: nit: `length() > 0` => `empty()`

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+    // Don't pass events up until the connect response is read because TLS reads
+    // kick off writes which don't pass through the transport socket.
+    return transport_socket_->doWrite(buffer, end_stream);
+  }
+  return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
+}
+
+Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer) {
+  if (!stripped_connect_) {
+    char peek_buf[200];
```
RyanTheOptimist, line 53: OOC, where does 200 come from? Should we document this anywhere?

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+    auto result = callbacks_->ioHandle().recv(peek_buf, 200, MSG_PEEK);
```
RyanTheOptimist, line 54: nit: 200 => ARRAYSIZE(peek_buf), perhaps?
Also it's not obvious to me what the return value of recv is, so I think the style guide would recommend again auto.

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+    }
+    absl::string_view peek_data(peek_buf, result.return_value_);
+    size_t index = peek_data.find("\r\n\r\n");
+    if (index == absl::string_view::npos) {
+      if (result.return_value_ == 200) {
+        ENVOY_CONN_LOG(trace, "failed to receive CONNECT headers within 200 bytes",
+                       callbacks_->connection());
+        return {Network::PostIoAction::Close, 0, false};
+      }
+      return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
```
RyanTheOptimist, line 66: nit: More early return opportunities here :) (Just remove the `else {}` below)

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+  if (!stripped_connect_) {
+    char peek_buf[200];
+    auto result = callbacks_->ioHandle().recv(peek_buf, 200, MSG_PEEK);
+    if (!result.ok() && result.err_->getErrorCode() != Api::IoError::IoErrorCode::Again) {
+      return {Network::PostIoAction::Close, 0, false};
+    }
+    absl::string_view peek_data(peek_buf, result.return_value_);
+    size_t index = peek_data.find("\r\n\r\n");
+    if (index == absl::string_view::npos) {
+      if (result.return_value_ == 200) {
```
RyanTheOptimist, line 61: Does this 200 represent the 200 status code or the size of the buffer? I suppose it's the latter so perhaps ARRAYSIZE(peek_buf)?

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+      return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
+    } else {
+      result = callbacks_->ioHandle().read(buffer, index + 4);
+      if (!result.ok() || result.return_value_ != index + 4) {
+        ENVOY_CONN_LOG(trace, "failed to drain CONNECT header", callbacks_->connection());
+        return {Network::PostIoAction::Close, 0, false};
+      }
+      // Note this is not in any way proper HTTP/1.1 parsing.
+      // Before this is used with any untrusted upstream, proper checks should
+      // be done rather than this.
```
RyanTheOptimist, line 75: Hah! Yeah, fair enough. Do we need a TODO or an issue for this?

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+      if (!absl::StartsWith(peek_data, "HTTP/1.1 200")) {
+        ENVOY_CONN_LOG(trace, "Response does not match strict connect checks",
+                       callbacks_->connection());
+        return {Network::PostIoAction::Close, 0, false};
+      }
+      buffer.drain(buffer.length());
+      stripped_connect_ = true;
```
RyanTheOptimist, line 82: Ah, so `stripped_connect_ = false` means "We still need to parse the connect response header" (which is why we set it to true when not sending the CONNECT method)?

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -0,0 +1,140 @@
...
+    if (result.ok()) {
+      ENVOY_CONN_LOG(trace, "write returns: {}", callbacks_->connection(), result.return_value_);
+      bytes_written += result.return_value_;
+    } else {
+      ENVOY_CONN_LOG(trace, "write error: {}", callbacks_->connection(),
+                     result.err_->getErrorDetails());
+      if (result.err_->getErrorCode() != Api::IoError::IoErrorCode::Again) {
+        action = Network::PostIoAction::Close;
+      }
+      break;
```
RyanTheOptimist, line 107: It'd be inclined to avoid a bit of nesting and write this as:
```
if (!result.ok()) {
      ENVOY_CONN_LOG(trace, "write error: {}", callbacks_->connection(),
                     result.err_->getErrorDetails());
      if (result.err_->getErrorCode() != Api::IoError::IoErrorCode::Again) {
        action = Network::PostIoAction::Close;
      }
      break;
}

ENVOY_CONN_LOG(trace, "write returns: {}", callbacks_->connection(), result.return_value_);
bytes_written += result.return_value_;
```

test/common/http/http1/codec_impl_test.cc
```diff
@@ -2120,8 +2120,9 @@ TEST_F(Http1ServerConnectionImplTest, ShouldDumpDispatchBufferWithoutAllocatingM
 class Http1ClientConnectionImplTest : public Http1CodecTestBase {
 public:
   void initialize() {
-    codec_ = std::make_unique<Http1::ClientConnectionImpl>(
-        connection_, http1CodecStats(), callbacks_, codec_settings_, max_response_headers_count_);
+    codec_ = std::make_unique<Http1::ClientConnectionImpl>(connection_, http1CodecStats(),
+                                                           callbacks_, codec_settings_,
+                                                           max_response_headers_count_, false);
```
RyanTheOptimist, line 2125: nit: /* passing_through_proxy=*/ false (or use a local variable `bool passing_through_proxy = false`)

test/extensions/transport_sockets/http_11_proxy/connect_integration_test.cc
```diff
@@ -0,0 +1,245 @@
...
+  // The request should be sent to fake upstream 0.
+  ASSERT_TRUE(fake_upstreams_[0]->waitForHttpConnection(*dispatcher_, fake_upstream_connection_));
+  ASSERT_TRUE(fake_upstream_connection_->waitForNewStream(*dispatcher_, upstream_request_));
+  ASSERT_TRUE(upstream_request_->waitForEndStream(*dispatcher_));
+  // No encapsulation.
+  EXPECT_TRUE(upstream_request_->waitForEndStream(*dispatcher_));
+  upstream_request_->encodeHeaders(default_response_headers_, true);
+  ASSERT_TRUE(response->waitForEndStream());
+  EXPECT_EQ("200", response->headers().getStatusValue());
+}
```
RyanTheOptimist, line 242: Should we add a test where we send multiple different requests over the same proxy?

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
```diff
@@ -0,0 +1,353 @@
...
+  Buffer::OwnedImpl buffer("");
+  connect_socket_->doRead(buffer);
+}
+
+TEST_P(Http11ConnectTest, NoInjectTlsAbsent) {
+  ssl_.reset();
+  initialize(false);
+
+  EXPECT_CALL(io_handle_, write(_)).Times(0);
+  auto msg = Buffer::OwnedImpl("initial data");
```
RyanTheOptimist, line 146: How about `Buffer::OwnedImpl msg("initial data")` which is shorter and does not use auto.

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
```diff
@@ -0,0 +1,353 @@
...
+TEST_P(Http11ConnectTest, InjectesHeaderOnlyOnce) {
+  initialize();
+
+  EXPECT_CALL(io_handle_, write(BufferStringEqual(connect_data_.toString())))
+      .WillOnce(Invoke([&](Buffer::Instance& buffer) -> Api::IoCallUint64Result {
+        auto length = buffer.length();
+        buffer.drain(length);
+        return Api::IoCallUint64Result(length, Api::IoErrorPtr(nullptr, [](Api::IoError*) {}));
+      }));
+  auto msg = Buffer::OwnedImpl("initial data");
```
RyanTheOptimist, line 96: How about `Buffer::OwnedImpl msg("initial data")` which is shorter and does not use auto. (here and elsewhere)

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
```diff
@@ -0,0 +1,353 @@
...
+namespace Http11Connect {
+namespace {
+
+class Http11ConnectTest : public testing::TestWithParam<Network::Address::IpVersion> {
+public:
+  Http11ConnectTest() {}
+  void initialize(bool no_proxy_protocol = false) {
+    std::string address_string =
+        absl::StrCat(Network::Test::getLoopbackAddressUrlString(GetParam()), ":1234");
+    auto address = Network::Utility::parseInternetAddressAndPort(address_string);
```
RyanTheOptimist, line 41: Can we use the full type here? Here and elsewhere. (In the make_unique call below the type is obvious so auto is fine)

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
```diff
@@ -0,0 +1,353 @@
...
+
+  EXPECT_CALL(io_handle_, write(BufferStringEqual(connect_data_.toString())))
+      .WillOnce(Invoke([&](Buffer::Instance& buffer) -> Api::IoCallUint64Result {
+        auto length = buffer.length();
+        buffer.drain(length);
+        return Api::IoCallUint64Result(length, Api::IoErrorPtr(nullptr, [](Api::IoError*) {}));
+      }));
+  auto msg = Buffer::OwnedImpl("initial data");
+
+  auto rc1 = connect_socket_->doWrite(msg, false);
```
RyanTheOptimist, line 98: Explicit type, please (here and below)

RyanTheOptimist (commented):
This looks amazing! A couple of legit comments but mostly just nits.

envoy/network/transport_socket.h
kyessenov, line 247: Why not keep this private to the extension? E.g. per https://github.com/envoyproxy/envoy/pull/21700, would be another private hashable object.

source/common/network/transport_socket_options_impl.cc
kyessenov: Suggestion to use well-known endpoint metadata ("tunnel.address" or something). We've been doing that for the internal listener effectively.

api/envoy/extensions/transport_sockets/http_11_proxy/v3/upstream_http_11_connect.proto
alyssawilk: That's a great call out.  It does not - I think we'll have to do that in the DFP filter since the CONNECT code doesn't interact with DNS.  I've added a TODO to do this as a follow-up.

envoy/network/transport_socket.h
alyssawilk: done

source/common/network/transport_socket_options_impl.cc
alyssawilk: using the hostname() sorry for missing this as I fixed things up.

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: done

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: done

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: done

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: the compiler tells me ARRAYSIZE is a googleism, so using const defined variable for clarity

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: ack

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: done

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: I mean we're not using untrusted upstreams (I trust android to not return crap) and the status of the plugin says it should only be used with trusted upstreams/downtreams.  anyone wanting to change that should sure as heck audit the code so I'm inclined to leave as-is

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: inverted for clarity

source/extensions/transport_sockets/http_11_proxy/connect.cc
alyssawilk: of course you would :-P

source/extensions/transport_sockets/http_11_proxy/connect.h
alyssawilk: pretty sure I've been corrected to not use {false} so leaving as is

test/common/http/http1/codec_impl_test.cc
alyssawilk: done.  though for the record I dislike that paradigm

test/extensions/transport_sockets/http_11_proxy/connect_integration_test.cc
alyssawilk: I _had_ one.  Where the heck did it go?
readded, thanks!

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
alyssawilk: ack

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
alyssawilk: done

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
alyssawilk: don

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
alyssawilk: done

source/extensions/transport_sockets/http_11_proxy/connect.cc
RyanTheOptimist: Oh, whoops. I missed that. Follow up is fine, or leave alone.

source/extensions/transport_sockets/http_11_proxy/connect.cc
RyanTheOptimist: Ah! Looks like it's ABSL_ARRAYSIZE() now. but const is great.

RyanTheOptimist (commented):
LGTM, though it looks like this needs a main merge. I'm also curious if the proxy info should be added to the TransportSocketOption's hashKey() computation?

envoy/network/transport_socket.h
kyessenov: These objects are also hashed in transport options and passed along during connection creation to the upstream connection stream info. Maybe I'm missing something?

alyssawilk: I see how I could handle hashing but I don't see how I could handle the createConnection in  
source/common/upstream/upstream_impl.cc properly without explicit visibility of the addresses?

kyessenov: I see. I suppose we could split the two since `address()` override seems more generic than CONNECT (you could do passthrough and all sorts of things with it), but it's up to you.

test/integration/filters/header_to_proxy_filter.cc
```diff
@@ -0,0 +1,54 @@
...
+class HeaderToProxyFilter : public Http::PassThroughFilter {
+public:
+  Http::FilterHeadersStatus decodeHeaders(Http::RequestHeaderMap& request_headers, bool) override {
+    auto connect_proxy = Http::LowerCaseString("connect-proxy");
+    auto hostname = request_headers.getHostValue();
+    ASSERT(!hostname.empty());
+    if (!request_headers.get(connect_proxy).empty()) {
+      std::string address_string(request_headers.get(connect_proxy)[0]->value().getStringView());
+      auto address = Network::Utility::parseInternetAddressAndPort(address_string);
+      decoder_callbacks_->streamInfo().filterState()->setData(
```
kyessenov, line 26: I wonder if we could use `LbEndpoint` itself somehow to determine the proxy address. In our case, each endpoint has an assigned proxy, and the proxy address is inside the endpoint metadata. I'm not sure how we'd make it work with a downstream filter.

RyanTheOptimist (dismissed):

api/envoy/extensions/transport_sockets/http_11_proxy/v3/upstream_http_11_connect.proto
```diff
@@ -0,0 +1,33 @@
...
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/http_11_proxy/v3;http_11_proxyv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Upstream HTTP/1.1 Proxy]
+// [#extension: envoy.transport_sockets.http_11_proxy]
+
+// Configuration for HTTP/1.1 proxy transport sockets.
+// If this is configured and an intermediate filter adds proxy metadata to the
+// stream info then
```
mattklein123, line 21: Can you clarify what "proxy metadata" means in this context?

mattklein123 (changes requested):
Awesome to see this come together. Just a high level comment from an API pass.

Also, I think this should have a release note? Additionally, I think this would benefit from some arch docs somewhere? It's getting kind of confusing to understand all the permutations of tunneling. Perhaps this could be added to wherever we document tunneling and we can flesh that section out to cover different use cases? Feel free to do this arch docs as a follow up if you want. Thank you so much!

/wait

api/envoy/extensions/transport_sockets/http_11_proxy/v3/upstream_http_11_connect.proto
```diff
@@ -17,11 +17,11 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // [#extension: envoy.transport_sockets.http_11_proxy]
 
 // Configuration for HTTP/1.1 proxy transport sockets.
-// If this is configured and an intermediate filter adds proxy metadata to the
-// stream info then
+// If this is configured and an intermediate filter adds the data necessary for
+// proxying to the stream info then
```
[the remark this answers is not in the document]
alyssawilk: yeah this is why I didn't want to do genreal Envoy docs for it for it as it's written for a very narrow use case.  I suspect it will get extended to be general purpose over time but right now, the goal is to provide the infrastructure for E-M sending data through the local android proxy (https://github.com/envoyproxy/envoy-mobile/issues/1622)  The E-M filter will query Android for the address and port of the proxy, then add that information to the stream info filter state so it can be used to redirect the connection and alter the payload going to the proxy.
I'm super happy to put these details in the proto but I've been repeatedly chided by @markdroth for putting too much Envoy information in what is supposed to be generic protos.  Given this one is Envoy-only maybe it's OK to add more detail there?  But this really isn't intended for generic Envoy use. I could also land the proto and extension in E-M but then I can't test any of the changes to core Envoy required to make it all hang together.  Happy to take your advice here.,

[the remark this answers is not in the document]
mattklein123: OK I see. I don't know maybe then just make it a bit more clear in the docs that this is EM specific right now and can be extended later? 

I get the concern about Envoy specific info in the protos, but IMO 1st priority is documentation. We need to sort out the per client docs separately which is a known issue.

moderation:
Does this PR mean you can implement something like https://github.com/envoyproxy/envoy/issues/11569#issuecomment-645623588=  but without the need to use a loopback?

kyessenov:
@moderation  I think adding some well-known metadata in endpoints would be an alternate way to find a tunnel address. That's what we've been doing for H2 tunneling with the internal listener. I think this is probably the right long term evolution away from internal listeners, but the problem is that transport sockets need to support multiplexing (1 to many for downstream, many to 1 for upstream), and right now they are basically one-to-one with TCP streams.

moderation:
Thanks @alyssawilk and @kyessenov

PR #21943, opened 2022-06-29
docs: update flag status

alyssawilk:
Wasn't updated when https://github.com/envoyproxy/envoy/blob/main/source/common/runtime/runtime_features.cc#L35 was defaulted true


phlax (approved):
lgtm, thanks @alyssawilk

PR #21954, opened 2022-06-29
internal listener: fix assertion in passthrough state

kyessenov:
Commit Message: Fix assertion in internal listener user space socket. The assertion would fail if internal upstream transport socket is not used, but it is not necessary.
Additional Description: Added an integration test as well (apparently there was none before without internal upstream transport)
Risk Level: low
Testing: yes
Docs Changes: none
Release Notes: none


kyessenov:
/assign @lizan 
(reviewer of internal upstream)

lizan (approved):

PR #21960, opened 2022-06-30
Split StopIteration into two statuses

zhaohuabing:
The meaning of the StopIteration status is ambiguous: it is used in two scenarios now:

* Pause the current iteration of the l7 filter chain and the execution of the filter chain will be resumed later.  In this case, the current implementation is fine.
* Abort the iteration of the l7 filter chain. In this case, the current implementation results in memory leak since the active message will not be deleted from the connection manager until the downstream connection is closed.

This PR solves this issue by splitting StopIteration into two unambiguous statuses: PauseIteration and AbortIteration

* PauseIteration: the filter iteration is paused and will be resumed later, for e.g, waiting for an available connection. The corresponding message remains in the connection manager.
* AbortIteration: the filter iteration is aborted, for e.g, the upstream host is not found, and the corresponding message will be deleted from the connection manager.

Signed-off-by: zhaohuabing <zhaohuabing@gmail.com>

Commit Message: Split StopIteration into two statuses
Additional Description:
Risk Level: low
Testing: yes
Docs Changes: none
Release Notes: none

### DIFF
--- a/source/extensions/filters/network/dubbo_proxy/active_message.cc
+++ b/source/extensions/filters/network/dubbo_proxy/active_message.cc
@@ -245,16 +245,24 @@ void ActiveMessage::onStreamDecoded(MessageMetadataSharedPtr metadata, ContextSh
   };
 
   auto status = applyDecoderFilters(nullptr, FilterIterationStartState::CanStartFromCurrent);
-  if (status == FilterStatus::StopIteration) {
-    ENVOY_LOG(debug, "dubbo request: stop calling decoder filter, id is {}", metadata->requestId());
+  switch (status) {
+  case FilterStatus::StopIteration:
+    ENVOY_LOG(debug, "dubbo request: pause calling decoder filter, id is {}",
+              metadata->requestId());
     pending_stream_decoded_ = true;
     return;
+  case FilterStatus::AbortIteration:
+    ENVOY_LOG(debug, "dubbo request: abort calling decoder filter, id is {}",
+              metadata->requestId());
+    parent_.deferredMessage(*this);
+    return;
+  case FilterStatus::Continue:
+    ENVOY_LOG(debug, "dubbo request: complete processing of downstream request messages, id is {}",
+              metadata->requestId());
+    finalizeRequest();
+    return;
   }
-
-  finalizeRequest();
-
-  ENVOY_LOG(debug, "dubbo request: complete processing of downstream request messages, id is {}",
-            metadata->requestId());
+  PANIC_DUE_TO_CORRUPT_ENUM
 }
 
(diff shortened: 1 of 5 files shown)


source/extensions/filters/network/dubbo_proxy/active_message.cc
```diff
@@ -81,7 +81,7 @@ FilterStatus ActiveResponseDecoder::applyMessageEncodedFilters(MessageMetadataSh
   auto status = parent_.applyEncoderFilters(
       nullptr, ActiveMessage::FilterIterationStartState::CanStartFromCurrent);
   switch (status) {
-  case FilterStatus::StopIteration:
+  case FilterStatus::PauseIteration:
```
wbpcode, line 84: May be keep the `StopIteration`?

source/extensions/filters/network/dubbo_proxy/active_message.cc
```diff
@@ -81,7 +81,7 @@ FilterStatus ActiveResponseDecoder::applyMessageEncodedFilters(MessageMetadataSh
...
     break;
   default:
     ASSERT(FilterStatus::Continue == status);
```
wbpcode, line 87: Handle `AbortIteration` here?

source/extensions/filters/network/dubbo_proxy/decoder_event_handler.h
```diff
@@ -15,10 +15,12 @@ namespace DubboProxy {
...
-  // Do not iterate to any of the remaining filters in the chain. Returning
-  // FilterDataStatus::Continue from decodeData()/encodeData() or calling
-  // continueDecoding()/continueEncoding() MUST be called if continued filter iteration is desired.
-  StopIteration,
+  // Pause iterating to any of the remaining filters in the chain.
+  // The current message remains in the connection manager to wait to be continued.
+  // ContinueDecoding()/continueEncoding() MUST be called to continue filter iteration.
+  PauseIteration,
+  // Abort the current iterate and remove the current message from the connection manager
+  AbortIteration,
```
wbpcode, line 23: nit: ` // Abort the current iterate and remove the current message from the connection manager.`

wbpcode (commented):
Thanks for your contribution. Some comments are added.

source/extensions/filters/network/dubbo_proxy/active_message.cc
zhaohuabing: Sure, we can keep StopIteration as it is for back-compatible.

wbpcode:
@zhaohuabing is this still in WIP?

wbpcode (approved):
LGTM. Thanks.

PR #21964, opened 2022-06-30
upstream: Add an interface to support applying Android socket tags

RyanTheOptimist:
Add an interface to support applying Android socket tags to upstream sockets. Create a new
Socket::Option for this purpose.

A concrete implementation of the Android socket tag will come in a subsequent PR.

For more detail about motivation, see: https://github.com/envoyproxy/envoy-mobile/issues/1521

Risk Level: Low, off by default
Testing: New integration test
Docs Changes: N/A
Release Notes: Updated


mattklein123 (commented):
This LGTM, but high level question: Should this just leave in the EM repo? Do we need this code in upstream Envoy? Thank you.

/wait-any

PR #8309, opened 2019-09-20
api: protoxform tool and API reformat.

htuch:
This patch introduces a new tool, protoxform, that will be the basis of
the v2 -> v3 migration tooling. It operates as a Python protoc plugin,
within the same framework as protodoc, and provides the ability to
operate on protoc AST input and generate proto output.

As a first step, the tool is applied reflexively on v2, and functions as
a formatting tool. In later patches, this will be added to
check_format/fix_format scripts and CI.

Part of #8082.

Risk level: medium (it's possible that some inadvertent wire changes
  occur, if they do, this patch should be rolled back).
Testing: manual inspection of diff, bazel test //test/..., some
  grep/diff scripts to ensure we haven't lost any comments.

Signed-off-by: Harvey Tuch <htuch@google.com>

PR #21969, opened 2022-06-30
api_proto_plugin: Remove debugging code

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/tools/api_proto_plugin/plugin.py
+++ b/tools/api_proto_plugin/plugin.py
@@ -1,9 +1,5 @@
 """Python protoc plugin for Envoy APIs."""
 
-import cProfile
-import io
-import os
-import pstats
 import sys
 from collections import namedtuple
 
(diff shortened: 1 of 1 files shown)


wrowe:
Requesting review from @htuch who introduced the code now gated by cprofile_enabled.

From the slack, further comments from phlax;

> To give more context, these are FileDescriptorProtos ie files including source codek, atm they are created as outputs but not used anywhere (afaict)
> Also bazel recently added FileDescriptorProtos to ProtoSets so where we do actually need the source, which is going through this code we can grab it.
> Re cprofile_enabled - mho is just remove the code guarded by it as it would require use of an action_env var to get the var there and that is not documented anywhere

It sounds to me that these are redundant and this was early diagnostic logic that never got mopped up from https://github.com/envoyproxy/envoy/pull/8309?

htuch:
Yeah, we can drop the cprofile stuff I think. But the this PR is now only about this, so the title no longer makes sense?

phlax:
title updated

htuch (approved):
LGTM, thanks!

PR #21970, opened 2022-06-30
aws_lambda: allow cross account lambda function invocation

vaibhav7dhage:
Signed-off-by: Vaibhav Dhage <vddhage@amazon.com>

Commit Message:
- Allow cross account lambda function invocation by setting the path in header with complete Lambda ARN instead of the function name.

Additional Description:
- Without this change, lambda filter always calls the lambda function in the AWS account where envoy runs and fails if it can't find the function in that account with the same name.

Risk Level:
- Low

Testing:
- Unit Test
- Integration Test
- Manual testing in AWS cross account setup to make sure function invocation works for both in account and cross account.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
@@ -41,17 +41,11 @@ namespace {
 constexpr auto filter_metadata_key = "com.amazonaws.lambda";
 constexpr auto egress_gateway_metadata_key = "egress_gateway";
 
-void setLambdaHeaders(Http::RequestHeaderMap& headers,
-                      const absl::optional<Arn>& arn,
+void setLambdaHeaders(Http::RequestHeaderMap& headers, const absl::optional<Arn>& arn,
                       InvocationMode mode) {
   headers.setMethod(Http::Headers::get().MethodValues.Post);
-  auto lambda_arn = "arn:" +
-      arn->partition() + ":" +
-      arn->service() + ":" +
-      arn->region() + ":" +
-      arn->accountId() + ":" +
-      arn->resourceType() + ":" +
-      arn->functionName();
+  auto lambda_arn = "arn:" + arn->partition() + ":" + arn->service() + ":" + arn->region() + ":" +
+                    arn->accountId() + ":" + arn->resourceType() + ":" + arn->functionName();
   headers.setPath(fmt::format("/2015-03-31/functions/{}/invocations", lambda_arn));
   if (mode == InvocationMode::Synchronous) {
     headers.setReference(LambdaFilterNames::get().InvocationTypeHeader, "RequestResponse");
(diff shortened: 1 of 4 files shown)


suniltheta:
Can you update [changelogs](https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml) with this as a bug fix?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -41,10 +41,12 @@ namespace {
 constexpr auto filter_metadata_key = "com.amazonaws.lambda";
 constexpr auto egress_gateway_metadata_key = "egress_gateway";
 
-void setLambdaHeaders(Http::RequestHeaderMap& headers, absl::string_view function_name,
+void setLambdaHeaders(Http::RequestHeaderMap& headers, const absl::optional<Arn>& arn,
                       InvocationMode mode) {
   headers.setMethod(Http::Headers::get().MethodValues.Post);
-  headers.setPath(fmt::format("/2015-03-31/functions/{}/invocations", function_name));
+  auto lambda_arn = "arn:" + arn->partition() + ":" + arn->service() + ":" + arn->region() + ":" +
```
suniltheta, line 47: Can we add a method inside [Arn class](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/aws_lambda/aws_lambda_filter.h#L21) so that we can directly call `arn->value()` so that we get the arn string value entirely? Then we don't have to construct the arn value every time. The config has the entire arn as a string in itself.

Also worth considering if we can simplify `class Arn`. Currently we only use `region()` & `functionName()` method inside. With this PR we can even get rid of `functionName` and other unused functions.

vaibhav7dhage:
> Can you update [changelogs](https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml) with this as a bug fix?

Good point!

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
vaibhav7dhage: Sure. Actually, planning to add another member variable `arn_` in `Arn()` class which retains the original ARN string from the config.

I would keep the other functions since they could be used in future after parsing ARN to extract specific fields.

suniltheta: that should work as well.

suniltheta (approved):
LGTM!

wrowe:
Hi @mattklein123 , this has one review by @suniltheta and looks ready for final review, assigning as a codeowner if you could please?

mattklein123 (approved):

PR #21972, opened 2022-07-01
Move regex engine initialization before stats store initialization

xjtian:
Commit Message: Move regex engine initialization before stats store initialization
Additional Description: If the stats store bootstrap config contains regexes, Envoy initialization will fail with `[critical][assert] [external/envoy/source/common/singleton/threadsafe_singleton.h:56] assert failure: loader_ != nullptr. Details: InjectableSingleton used prior to initialization`. This PR initializes the regex engine before the stats store to fix this
Risk Level: low
Testing: updated the existing integration test to include a stats config with regexes
Docs Changes: n/a
Release Notes: n/a

Signed-off-by: Jacky Tian <jtian@lyft.com>

### DIFF
--- a/source/server/server.cc
+++ b/source/server/server.cc
@@ -445,6 +445,23 @@ void InstanceImpl::initialize(Network::Address::InstanceConstSharedPtr local_add
               absl::StrJoin(info.registered_headers_, ","));
   }
 
+  // Initialize the regex engine and inject to singleton.
+  // Needs to happen before stats store initialization because the stats
+  // matcher config can include regexes.
+  if (bootstrap_.has_default_regex_engine()) {
+    const auto& default_regex_engine = bootstrap_.default_regex_engine();
+    Regex::EngineFactory& factory =
+        Config::Utility::getAndCheckFactory<Regex::EngineFactory>(default_regex_engine);
+    auto config = Config::Utility::translateAnyToFactoryConfig(
+        default_regex_engine.typed_config(), messageValidationContext().staticValidationVisitor(),
+        factory);
+    regex_engine_ = factory.createEngine(*config, serverFactoryContext());
+  } else {
+    regex_engine_ = std::make_shared<Regex::GoogleReEngine>();
+  }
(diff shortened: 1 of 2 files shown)


source/server/server.cc
```diff
@@ -445,6 +445,23 @@ void InstanceImpl::initialize(Network::Address::InstanceConstSharedPtr local_add
...
   }
 
+  // Initialize the regex engine and inject to singleton.
+  // Needs to happen before stats store initialization because the stats
+  // matcher config can include regexes.
+  if (bootstrap_.has_default_regex_engine()) {
+    const auto& default_regex_engine = bootstrap_.default_regex_engine();
+    Regex::EngineFactory& factory =
+        Config::Utility::getAndCheckFactory<Regex::EngineFactory>(default_regex_engine);
+    auto config = Config::Utility::translateAnyToFactoryConfig(
```
adisuissa, line 455: IIRC this may increase a stats counter ([deprecated_feature_use](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#runtime-stats)).
cc @mattklein123 can this code be performed before stats_store_ is fully initialized?

adisuissa (commented):
Thanks for fixing this!

source/server/server.cc
mattklein123: I'm not sure we have a choice. IIRC there is code that handles this but I can't remember. I.e. I thought at some point I fixed it to reparse bootstrap to look for deprecated fields but maybe I'm not remembering.

xjtian: This block of code won't increment the stat, but the instantiation of the stats matcher in the block below this could if there's a regex matcher with a deprecated `google_re2: {}` config block. This doesn't block server startup, but the deprecated feature stat won't get incremented in this case.

mattklein123 (approved):
Thanks!

mattklein123:
We should have landed this before the release. This will need to be backported. cc @pradeepcrao @phlax

Issue #20147, opened 2022-03-01
High CPU usage for matching routes (re2::DFA::InlinedSearchLoop)


ZIK-1337:
What are the solutions to reduce CPU usage? There are a lot of routes, each request in the loop is matched - probably because of this.
Envoy v1.15.0
Top 50 usage from pprof:

> Showing top 50 nodes out of 260
      flat  flat%   sum%        cum   cum%
  1842.68s 20.48% 20.48%   1842.82s 20.48%  re2::DFA::InlinedSearchLoop
   484.96s  5.39% 25.87%    484.96s  5.39%  tc_newarray
   391.39s  4.35% 30.22%   4550.41s 50.57%  Envoy::Router::RegexRouteEntryImpl::matches (inline)
   373.74s  4.15% 34.37%    373.74s  4.15%  writev
   372.04s  4.13% 38.51%   5005.04s 55.63%  Envoy::Router::VirtualHostImpl::getRouteFromEntries
   353.18s  3.93% 42.43%   3138.31s 34.88%  re2::RE2::Match
   277.19s  3.08% 45.51%    277.19s  3.08%  absl::string_view::find_first_of
   261.06s  2.90% 48.42%    261.06s  2.90%  __GI___pthread_rwlock_rdlock
   241.56s  2.68% 51.10%   2206.65s 24.52%  re2::DFA::Search
   183.22s  2.04% 53.14%    458.71s  5.10%  Envoy::Router::RouteEntryImplBase::matchRoute
   170.52s  1.90% 55.03%    170.52s  1.90%  __nss_database_lookup
   164.50s  1.83% 56.86%    164.50s  1.83%  re2::Prog::GetDFA
   149.17s  1.66% 58.52%   3290.14s 36.57%  re2::RE2::DoMatch
   123.84s  1.38% 59.89%    123.84s  1.38%  operator delete[]
   110.38s  1.23% 61.12%    110.38s  1.23%  Envoy::Router::RouteEntryImplBase::evaluateRuntimeMatch (inline)
   106.25s  1.18% 62.30%    106.25s  1.18%  readv
    97.59s  1.08% 63.39%    114.70s  1.27%  <unknown>
    96.40s  1.07% 64.46%     96.40s  1.07%  __GI___pthread_rwlock_unlock
    91.51s  1.02% 65.47%     91.51s  1.02%  re2::DFA::AnalyzeSearchHelper
    86.56s  0.96% 66.44%     86.56s  0.96%  epoll_wait
[...]

Issue #6602, opened 2019-04-16
Scalable route configuration


htuch:
The existing Envoy `RouteConfiguration` must be evaluated linearly and contains match criteria with hard to predict execution time (regular expressions). For scale-up Envoy instances where very large route configurations for a host may be present (let's say O(1m)) or multi-tenant Envoy instances, this becomes problematic.

There are many ways this could be solved:
1. Envoy could optimize based on presented configuration, for example a trie could be recovered from repeated patterns or a specific match order. This is challenging if the existing linear semantics are to be preserved and involves Envoy complexity.
2. We could have variants of regular expressions, e.g. prefix/suffix/glob or RE2 that are cheaper and more predictable to evaluate.
3. We could have a parallel `RouteConfiguration` with trie-like semantics, where evaluation order is designed for efficiency.

Some combination of (2) and (3) seems likely to yield a clean solution IMHO, but I'm keen to hear from others.

@jmarantz @mattklein123 @dmitri-d @yanavlasov

mattklein123:
I've thought about this for a while and IMO we need to do (3) which is to have a parallel configuration which supports a much more limited set of matching options designed for sub-linear execution. Optimally, the configuration would allow embedding a "legacy" route configuration inside of it so that once a sub-linear match is done, the user can further do a linear match if desired for extra features.

dmitri-d:
I think a combination of (2) and (3) work. Not sure re: full-on regex implementation in (2), but some form of simple pattern (a subset of bash pattern-matching perhaps?)/prefix and/or suffix matching might be sufficient for our purposes.

jmarantz:
I've been messing around with RE2 vs std::regex benchmarking on my Mac on the plane, and there's a pretty significant speedup. This includes dramatic improvements executing regular expressions that are thought of as pathologically slow. I will publish a bit later.

However, I think sublinear is the only way to scale, one way or another we need to get to at trie.

I was also thinking it might be possible in some cases to compile a restricted subset of existing Envoy config into a trie, but I haven't sorted out the details. There's some value in being able to simply make the existing configurations run faster. This is because there are layers (e.g. Istio) that already mirror the existing Envoy configs.

Going with (3) would be easier though (less of a science project).

htuch:
@louiscryan

rnd4222:
[...]

PR #21973, opened 2022-07-01
contrib: add Hyperscan regex engine

zhxie:

Commit Message: contrib: add Hyperscan regex engine
Additional Description: Hyperscan has been introduced as an input matcher earlier this year. Since the regex engine interface has been completed, the patch extent the usage of Hyperscan into a contrib regex engine.
Risk Level: Low
Testing: Unit
Docs Changes: API
Release Notes: N/A
Platform Specific Features: Requires processor with SSSE3 support (nearly any modern x86 processor)


wrowe:
Ping @moderation for deps review?

moderation:
Metadata only change for the dependency

/lgtm deps

docs/root/api-v3/config/contrib/hyperscan/regex_engine.rst
```diff
@@ -0,0 +1,5 @@
+.. toctree::
```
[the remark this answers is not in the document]
zhxie: Thanks, I will fix that.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +59,88 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
-    ENVOY_LOG_MISC(error, "unable to scan, error code {}.", err);
+    ENVOY_LOG_MISC(error, "unable to scan, error code {}", err);
   }
 
   return matched;
 }
 
+std::string Matcher::replaceAll(absl::string_view value, absl::string_view substitution) const {
+  // Find matched pair.
+  using Found = std::pair<unsigned long long, unsigned long long>;
```
jmarantz, line 70: WDYT of naming this something like Bounds, and commenting whether .second is inclusive or not?

Or is it a start-position and length? Or something else?

I'm wondering why that has to be 'unsigned long long'? Usually we don't spell that out in Envoy and would use uint64_t.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +59,88 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
+      &founds);
+  if (err != HS_SUCCESS && err != HS_SCAN_TERMINATED) {
+    ENVOY_LOG_MISC(error, "unable to scan, error code {}", err);
+    return std::string(value);
+  }
+
+  // Sort founds. Make sure the longest length found in the front will appear first.
+  std::sort(founds.begin(), founds.end(), [](Found a, Found b) {
+    if (a.first == b.first) {
+      return a.second > b.second;
```
jmarantz, line 92: I feel like this sorting semantics could be broken out into its own function and unit-tested in isolation from the regex logic. WDYT?

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +59,88 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
+
+  // Concatenate string and replace matched pair with substitution.
+  std::string result;
+  unsigned long long pos = 0;
+  for (auto& found : founds) {
+    if (found.first < pos) {
+      continue;
+    }
+
+    result += std::string(value.substr(pos, found.first - pos));
```
jmarantz, line 105: `absl::StrCat(&result, value.substr(pos, found.first - pos), substitution);`

That will save a bunch of extraneous string-copying.

Or, even better, make result be a `vector<string_view>` and at the end write `return absl::StrJoin(result, "");`

jmarantz (commented):
Flushing some comments. Lots to dig into here.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -1,91 +1,55 @@
...
+    }
+  }
 }
 
-Matcher::Matcher(
-    const envoy::extensions::matching::input_matchers::hyperscan::v3alpha::Hyperscan& config,
-    ThreadLocal::SlotAllocator& tls)
+Matcher::Matcher(const std::vector<const char*>& expressions,
+                 const std::vector<unsigned int>& flags, const std::vector<unsigned int>& ids,
+                 ThreadLocal::SlotAllocator& tls, bool som)
```
jmarantz, line 26: What's `som`?

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +59,88 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
       },
       &matched);
   if (err != HS_SUCCESS && err != HS_SCAN_TERMINATED) {
-    ENVOY_LOG_MISC(error, "unable to scan, error code {}.", err);
+    ENVOY_LOG_MISC(error, "unable to scan, error code {}", err);
```
jmarantz, line 62: should we include the failed regex string in this message?

contrib/hyperscan/matching/input_matchers/source/config.cc
```diff
@@ -25,7 +25,45 @@ Config::createInputMatcherFactoryCb(const Protobuf::Message& config,
   throw EnvoyException("X86_64 architecture is required for Hyperscan.");
 #else
   return [hyperscan_config, &factory_context]() {
-    return std::make_unique<Matcher>(hyperscan_config, factory_context.threadLocal());
+    int size = hyperscan_config.regexes().size();
+    std::vector<const char*> expressions;
```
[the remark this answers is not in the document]
jmarantz: oh I see below that hs_compile_multi requires parallel vectors. Can you just comment about that here, as it's distant from the API requiring this setup.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +59,88 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
+
+void Matcher::compile(const std::vector<const char*>& expressions,
+                      const std::vector<unsigned int>& flags, const std::vector<unsigned int>& ids,
+                      hs_database_t** database) {
+  hs_compile_error_t* compile_err;
+  hs_error_t err =
+      hs_compile_multi(expressions.data(), flags.data(), ids.data(), expressions.size(),
+                       HS_MODE_BLOCK, nullptr, database, &compile_err);
+  if (err != HS_SUCCESS) {
+    std::string compile_err_message(compile_err->message);
```
jmarantz, line 129: why is this copy needed?

api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -14,14 +14,9 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
...
 
-// `Hyperscan <https://github.com/intel/hyperscan>`_ regex matcher. The engine of the matcher uses
-// hybrid automata techniques to allow simultaneous matching of large numbers of regular
-// expressions across streams of data.
-//
-// The matcher follows PCRE pattern syntax, and the regex string must adhere to the documented
-// `pattern support <https://intel.github.io/hyperscan/dev-reference/compilation.html#pattern-support>`_.
-// The syntax is not compatible with the default RE2 regex matcher. Depending on configured expressions,
-// swapping regex matcher may cause match rules to no longer be valid.
+// `Hyperscan <https://github.com/intel/hyperscan>`_ regex matcher. The matcher uses Hyperscan
```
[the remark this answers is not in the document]
zhxie: I think Hyperscan as a proper noun, maybe we can ignore the article "the"?

api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -14,14 +14,9 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
...
+// engine which allows simultaneous matching of large numbers of regular expressions across streams
```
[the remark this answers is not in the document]
zhxie: This description is copied from the [homepage](https://www.hyperscan.io/) of Hyperscan.

Yes, It uses SIMD for simultaneous matching. And these SIMD instructions are delivered in modern processors (maybe in 15 years).

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -0,0 +1,49 @@
...
+// expressions, swapping regex engine may cause match rules to no longer be valid.
+// [#next-free-field: 7]
+message Hyperscan {
+  reserved 4;
+
+  // Matching will be performed case-insensitively.
+  //
+  // The expression may still use PCRE tokens (notably `(?i)` and `(?-i)`) to switch
+  // case-insensitive matching on and off.
+  bool caseless = 1;
```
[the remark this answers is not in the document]
zhxie: Thanks. These flags are copied from the documentation of Hyperscan. I will change all these options to user friendly phrases.

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -0,0 +1,49 @@
...
+
+  // Matching a `.` will not exclude newlines.
+  bool dot_all = 2;
+
+  // `^` and `$` anchors match any newlines in data.
+  bool multiline = 3;
```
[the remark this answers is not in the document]
zhxie: `/^bar/m` (multiline in Hyperscan) matches "foo\nbar" while `/^bar/` does not.

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -0,0 +1,49 @@
...
+
+  // Treat the pattern as a sequence of UTF-8 characters.
```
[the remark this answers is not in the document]
zhxie: For example, `/^.$/u` (utf8 in Hyperscan) matches "😀", while `/./` does not, since "😀" is interpreted as 2 characters.

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -0,0 +1,49 @@
...
+  bool utf8 = 5;
+
+  // Use Unicode properties for character classes.
+  //
+  // This option instructs Hyperscan to use Unicode properties, rather than the default ASCII
+  // interpretations, for character mnemonics like `\w` and `\s` as well as the POSIX character
+  // classes. It is only meaningful in conjunction with `utf8`.
+  bool ucp = 6;
```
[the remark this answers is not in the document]
zhxie: For example, `/^\w$/` with UCP enabled matches "Á", while `/^\w$/` without does not, since "Á" is not a [a-zA-Z0-9_] word character.

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -0,0 +1,49 @@
...
+// of data.
+//
+// The engine follows PCRE pattern syntax, and the regex string must adhere to the documented
+// `pattern support <https://intel.github.io/hyperscan/dev-reference/compilation.html#pattern-support>`_.
+// The syntax is not compatible with the default RE2 regex engine. Depending on configured
+// expressions, swapping regex engine may cause match rules to no longer be valid.
+// [#next-free-field: 7]
+message Hyperscan {
+  reserved 4;
+
```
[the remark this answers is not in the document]
zhxie: These flags are all `false` by default.

contrib/hyperscan/matching/input_matchers/source/config.cc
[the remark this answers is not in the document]
zhxie: Ok, I will add some descriptions here.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: It is a good idea. I am not good at naming, and "bounds" sounds great.

I will leave some comments here.

It is just because Hyperscan uses `unsigned long long`, I can just change them to `uint64_t`.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Unit tests have covered these codes. But I am willing to extract them into a class member function.

zhxie: Also, I suggest making bounds a struct, so we can implement `operator<` directly for the struct.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: That's cool! I will fix that.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: `som` represents `start of matching`. It is an abbreviation used by Hyperscan. With `som` flag turned on, all the matching will report their start and end position (while increasing matching time), which is used for `CompiledMatcher::replaceAll`.

I can rename to `report_start_of_matching` if it is not clear.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: I am afraid we are not able to include the regex pattern in the message. Hyperscan cannot report the failed regex patterns on matching. In fact, the `else` block is just a insurance, the error can only be produced in the situation like lack memory or something else.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: We must call `hs_free_compile_error` to free the `compile_err` before the calling of `throw`. So we have to store these messages in advance if we want to show in the throw statements.

contrib/hyperscan/matching/input_matchers/source/matcher.h
```diff
@@ -14,26 +13,35 @@ namespace InputMatchers {
...
-class Matcher : public Envoy::Matcher::InputMatcher {
+class Matcher : public Envoy::Regex::CompiledMatcher, public Envoy::Matcher::InputMatcher {
 public:
-  explicit Matcher(
-      const envoy::extensions::matching::input_matchers::hyperscan::v3alpha::Hyperscan& config,
-      ThreadLocal::SlotAllocator& tls);
-  ~Matcher() override { hs_free_database(database_); }
+  Matcher(const std::vector<const char*>& expressions, const std::vector<unsigned int>& flags,
+          const std::vector<unsigned int>& ids, ThreadLocal::SlotAllocator& tls, bool som);
+  ~Matcher() override {
```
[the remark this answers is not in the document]
zhxie: Done and pushed.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
jmarantz: Yes that'd be great. See https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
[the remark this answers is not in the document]
jmarantz: The article is needed because "Hyperscan" is not being used as a noun. It's being used as an adjective and the noun in "engine".

api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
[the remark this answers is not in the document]
jmarantz: Great, in that case, how about "which exploits x86 SIMD instructions to accelerate matching of"

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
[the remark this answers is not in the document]
jmarantz: Sounds good. I'm wondering if we need to propagate these choices to Envoy users; maybe in the context of Envoy we can just decide the right thing?

If you do decide that we need to enable customization of this, adding the PR comments into the proto doc would be helpful.

contrib/hyperscan/matching/input_matchers/source/config.cc
```diff
@@ -24,8 +24,47 @@ Config::createInputMatcherFactoryCb(const Protobuf::Message& config,
 #ifdef HYPERSCAN_DISABLED
   throw EnvoyException("X86_64 architecture is required for Hyperscan.");
 #else
+  // Hyperscan requires vectors of expressions, flags and IDs for matching database compilation.
```
jmarantz, line 27: s/Hyperscan requires vectors/Hyperscan's API requires parallel vectors/

contrib/hyperscan/matching/input_matchers/source/matcher.cc
jmarantz: I see -- I think if we wanted to retain the original regex we could, but given it's only an OOM situation it is not necessary.

Can you add a comment that we don't expect this to occur? Actually, should this be  an IS_ENVOY_BUG invocation rather than a log message?

We won't be able to cover this line in tests right?

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +75,82 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
 
+std::string Matcher::replaceAll(absl::string_view value, absl::string_view substitution) const {
+  // Find matched bounds.
+  std::vector<Bound> bounds;
+  hs_scratch_t* scratch_ = tls_->get()->scratch_;
+  hs_error_t err = hs_scan(
+      som_database_, value.data(), value.size(), 0, scratch_,
+      [](unsigned int, unsigned long long from, unsigned long long to, unsigned int,
+         void* context) -> int {
+        std::vector<Bound>* founds = static_cast<std::vector<Bound>*>(context);
```
jmarantz, line 92: s/founds/bounds/

contrib/hyperscan/matching/input_matchers/test/matcher_speed_test.cc
```diff
@@ -43,11 +43,9 @@ BENCHMARK(BM_CompiledGoogleReMatcher);
 
 // NOLINTNEXTLINE(readability-identifier-naming)
 static void BM_HyperscanMatcher(benchmark::State& state) {
-  envoy::extensions::matching::input_matchers::hyperscan::v3alpha::Hyperscan config;
-  auto regex = config.add_regexes();
-  regex->set_regex(std::string(cluster_re_pattern));
   auto instance = ThreadLocal::InstanceImpl();
-  auto matcher = Extensions::Matching::InputMatchers::Hyperscan::Matcher(config, instance);
+  auto matcher = Extensions::Matching::InputMatchers::Hyperscan::Matcher(
+      {cluster_re_pattern.data()}, {0}, {0}, instance, false);
```
jmarantz, line 48: pre-existing, but you can s/cluster_re_pattern/ClusterRePattern/ per Envoy naming convention for constants?

More importantly, if a null-terminated char* is expected here you can't guarantee you'll get that with .data(). Strictly speaking you need to make use `std::string(ClusterRePattern).c_str()` here.

contrib/hyperscan/matching/input_matchers/test/matcher_test.cc
```diff
@@ -46,6 +48,125 @@ TEST(MatcherTest, RaceScratchCreation) {
   }
 }
 
+// Verify that comparing works correctly for bounds.
+TEST(BoundTest, Compare) {
+  EXPECT_TRUE(Bound(1, 1) < Bound(2, 1));
```
jmarantz, line 53: EXPECT_LT x3

jmarantz (commented):
This looks really good -- a few more comments.

jmarantz:
also needs main merge.

api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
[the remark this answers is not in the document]
zhxie: Thanks, I will fix it.

api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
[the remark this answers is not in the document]
zhxie: Fixed.

contrib/hyperscan/matching/input_matchers/source/config.cc
zhxie: Thanks, fixed.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Thanks, fixed.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Thanks for your suggestion. We do not expect it to occur, and I will replace the log to a `IS_ENVOY_BUG`.

We will not cover these lines in tests.

contrib/hyperscan/matching/input_matchers/test/matcher_speed_test.cc
zhxie: Thanks, fixed.

api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto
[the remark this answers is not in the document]
zhxie: I think you are right. Maybe we can take the decision and align the Hyperscan's behavior as RE2.

contrib/hyperscan/matching/input_matchers/test/matcher_test.cc
zhxie: Fixed all above and pushed.

zhxie:
@moderation could you please review deps again? I do not think there is anything changed since last review (metadata only) but the bot add deps label.

contrib/hyperscan/matching/input_matchers/source/matcher.h
```diff
@@ -14,26 +13,42 @@ namespace InputMatchers {
...
+  std::string replaceAll(absl::string_view value, absl::string_view substitution) const override;
 
   // Envoy::Matcher::InputMatcher
   bool match(absl::optional<absl::string_view> input) override;
 
 private:
-  std::vector<unsigned int> flags_{};
-  std::vector<unsigned int> ids_{};
   hs_database_t* database_{};
+  hs_database_t* som_database_{};
```
jmarantz, line 47: ideally we should spell out whatever som stands for everywhere by renaming "som" as a variable name segment. but at a minimum maybe expand that acronym here?

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -1,91 +1,71 @@
...
+ScratchThreadLocal::ScratchThreadLocal(const hs_database_t* database,
+                                       const hs_database_t* som_database) {
   hs_error_t err = hs_alloc_scratch(database, &scratch_);
   if (err != HS_SUCCESS) {
     throw EnvoyException(fmt::format("unable to allocate scratch space, error code {}.", err));
   }
+  if (som_database) {
+    err = hs_alloc_scratch(som_database, &scratch_);
+    if (err != HS_SUCCESS) {
+      throw EnvoyException(
```
jmarantz, line 18: unreachable?

use IS_ENVOY_BUG?

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +75,82 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
+  if (err != HS_SUCCESS && err != HS_SCAN_TERMINATED) {
+    IS_ENVOY_BUG(fmt::format("unable to scan, error code {}", err));
+    return std::string(value);
+  }
+
+  // Sort bounds. Make sure the longest length bound in the front will appear first.
+  std::sort(bounds.begin(), bounds.end());
+
+  // Concatenate string and replace matched pair with substitution.
+  std::vector<absl::string_view> parts;
```
jmarantz, line 108: parts.reserve(bounds.size() * 2);

contrib/hyperscan/matching/input_matchers/source/matcher.h
```diff
@@ -14,26 +13,42 @@ namespace InputMatchers {
...
   ThreadLocal::TypedSlotPtr<ScratchThreadLocal> tls_;
+
+  void compile(const std::vector<const char*>& expressions, const std::vector<unsigned int>& flags,
```
jmarantz, line 50: add docstring explaining this will throw on failure, and the requirement for parallel vectors.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -1,91 +1,71 @@
...
+    for (unsigned int& som_flag : som_flags) {
+      som_flag = som_flag | HS_FLAG_SOM_LEFTMOST;
     }
+    compile(expressions, som_flags, ids, &som_database_);
   }
 
-  tls_->set(
-      [this](Event::Dispatcher&) { return std::make_shared<ScratchThreadLocal>(this->database_); });
+  tls_->set([this](Event::Dispatcher&) {
+    return std::make_shared<ScratchThreadLocal>(this->database_, this->som_database_);
```
jmarantz, line 55: nit: `this->` not necessary -- I think that would only be needed in a template.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
```diff
@@ -95,12 +75,82 @@ bool Matcher::match(absl::optional<absl::string_view> input) {
...
+  hs_error_t err =
+      hs_compile_multi(expressions.data(), flags.data(), ids.data(), expressions.size(),
+                       HS_MODE_BLOCK, nullptr, database, &compile_err);
+  if (err != HS_SUCCESS) {
+    std::string compile_err_message(compile_err->message);
+    int compile_err_expression = compile_err->expression;
+    hs_free_compile_error(compile_err);
+
+    if (compile_err_expression < 0) {
+      throw EnvoyException(fmt::format("unable to compile database: {}", compile_err_message));
```
jmarantz, line 144: do we need a testcase for this case?

contrib/hyperscan/matching/input_matchers/source/config.cc
```diff
@@ -24,8 +24,48 @@ Config::createInputMatcherFactoryCb(const Protobuf::Message& config,
...
+      if (regex.dot_all()) {
+        flag |= HS_FLAG_DOTALL;
+      }
+      if (regex.multiline()) {
+        flag |= HS_FLAG_MULTILINE;
+      }
+      if (regex.allow_empty()) {
+        flag |= HS_FLAG_ALLOWEMPTY;
+      }
+      if (regex.utf8()) {
```
jmarantz, line 52: should we make the default behavior match RE2 more closely by flipping the defaults for some of these?

Or is that too late because you already checked in a proto-file?

contrib/hyperscan/matching/input_matchers/source/matcher.h
```diff
@@ -14,26 +13,42 @@ namespace InputMatchers {
...
   hs_scratch_t* scratch_{};
 };
 
-class Matcher : public Envoy::Matcher::InputMatcher {
+struct Bound {
+  Bound(uint64_t start, uint64_t end);
+
+  bool operator<(const Bound& other) const;
+
+  uint64_t start_;
```
jmarantz, line 27: I assume end_ is exclusive, like STL end() ?

If that's the case can you call these begin_ and end_ ?

In either case can you comment?

contrib/hyperscan/matching/input_matchers/test/matcher_test.cc
```diff
@@ -46,6 +48,125 @@ TEST(MatcherTest, RaceScratchCreation) {
...
+// Verify that comparing works correctly for bounds.
+TEST(BoundTest, Compare) {
+  EXPECT_LT(Bound(1, 1), Bound(2, 1));
+  EXPECT_LT(Bound(1, 2), Bound(2, 1));
+  EXPECT_LT(Bound(2, 2), Bound(2, 1));
+}
+
+class MatcherTest : public ::testing::Test {
+protected:
+  void setup(const char* expression, unsigned int flag, bool report_start_of_matching) {
```
jmarantz, line 60: Can you expose a helper function that translates protobufs into the flag bitmask and test using the protobufs?

jmarantz (commented):
this is looking pretty good.

What's the usage model? Will someone configuring the matcher control its behavior RE multi-line, utf8 etc?

It seems like we might want the least amount of friction here and potential for operability issues. I'm not sure that we get that by inheriting the Hyperscan defaults.

contrib/hyperscan/matching/input_matchers/source/matcher.h
zhxie: Done.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: No, it is not unreachable. A malformed regex pattern will cause the exception.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Done.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Done.

contrib/hyperscan/matching/input_matchers/source/matcher.h
zhxie: Done.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Oh, I see the incorrect line. Yes, it is unreachable as expected, and only will happen on lacking memories.

BTW, should we use `IS_ENVOY_BUG` so frequently? The exception will be thrown in config time but not runtime, and lacking memories is not a Envoy bug I think.

contrib/hyperscan/matching/input_matchers/source/config.cc
zhxie: I do not want to change behavior of Hyperscan input matcher. I will leave this choice to user in Hyperscan input matcher, and make the default behavior match RE2 closely in Hyperscan regex engine.

contrib/hyperscan/matching/input_matchers/source/matcher.h
zhxie: It does not matter, can be called whatever you want. For me, it is just a struct holds two variables which indicates the start and end number. It will be fine to change to `begin_`.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: It will only happen on lacking memories or something else on config time.

contrib/hyperscan/matching/input_matchers/test/matcher_test.cc
zhxie: In fact, no. The matcher test works for both Hyperscan input matcher and regex engine, so introduce proto here is not reasonable. If you think it is necessary to test protos, we should add them in config_test.cc

All above replied or fixed, thanks.

jmarantz:
What's the usage model? Will someone configuring the matcher control its behavior RE multi-line, utf8 etc?

You said in one of the comments that you don't want to alter Hyperscan defaults, and that makes me think you are optimizing for Hyperscan users that might want to try using Envoy. When my model is an Envoy user wants to try Hyperscan would want the behavior to be as close to RE2 as possible (just faster).

contrib/hyperscan/matching/input_matchers/source/matcher.cc
jmarantz: RE "lacking memory" I would use IS_ENVOY_BUG which is a hint that the lack of coverage is OK

RE "something else" -- like what? Invalid input? We should test that.

contrib/hyperscan/matching/input_matchers/source/matcher.h
```diff
@@ -14,26 +13,44 @@ namespace InputMatchers {
...
   bool match(absl::optional<absl::string_view> input) override;
 
 private:
-  std::vector<unsigned int> flags_{};
-  std::vector<unsigned int> ids_{};
   hs_database_t* database_{};
+  hs_database_t* start_of_match_database_{};
   ThreadLocal::TypedSlotPtr<ScratchThreadLocal> tls_;
+
+  // Compiles the Hyperscan database. It will throw on failure of lacking memories or malformed
```
jmarantz, line 50: s/lacking memories/insufficient memory/

zhxie:
No. There are 2 usages of Hyperscan. Input matcher and regex engine. As an input matcher, users should change their configuration to generic matching, it is a lot of work, so I leave options to users. But as a regex engine, users can just change `default_regex_engine` in bootstrap, and I make it close to RE2.

I think you are right, we should make the decision to remove these flags in regex engine, and it is the implementation now.

jmarantz:
More basic question: who exactly is going to use this feature?

jmarantz:
So @ZIK-1337 is the initial user? I think I'm not asking about why it might be used. I'm asking who (which companies) plan to use this.

Also there are better alternatives to regexes matching for scalable path evaluation: https://github.com/envoyproxy/envoy/issues/6602

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Fixed, I will change it to `IS_ENVOY_BUG`.

For example, invalid parameters (flags), which will not happen.

contrib/hyperscan/matching/input_matchers/source/matcher.h
zhxie: Fixed.

All above fixed, thanks.

zhxie:
Replied offline about end users consideration.

Scalable path evaluation is not conflict with regex engine update. In fact, a regex engine can be used everywhere in Envoy, especially in security region (including WAF-like extensions like RBAC).

contrib/hyperscan/matching/input_matchers/source/matcher.cc
jmarantz: Yeah I think what I really want to express here is that the line of code can't be hit.

There's certainly a concern with lines of code that can't be tested, and whether anyone would catch this assert successfully and what they would do about that if there was no memory left.

We used to have a macro to indicate specifically that a line of code can't be hit, and previously the toolchain running the coverage analysis would not count those lines, but that stopped working at some point when the toolchain evolved.

WDYT of just using RELEASE_ASSERT here rather than throwing an exception?

jmarantz (commented):
looks good -- just a few more small questions.

jmarantz:
I think @zhxie did a great job with the code in this PR -- thanks for all the iteration.

I still have some higher level concerns
 * what are the exact steps someone would take to rebuild Envoy with this feature
 * what changes would a user need to existing re2 regexes?
 * wouldn't the user (eg in #20147) be better off using https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/common/matcher/v3/matcher.proto ?

I am worried about propagating an alternate regex syntax (even as contrib/...) and bifurcating the community. I also wish Hyperscan didn't bring in Boost.

I thought also about whether some of the Hyperscan techniques should just be ported to re2 but that seems unlikely. There was some enlightening discussion on https://github.com/google/re2/issues/383 . Notably -- there are microbenchmarks showing Hyperscan's raw speed but no obvious benchmarks exposed in the context of Envoy operations.

@mattklein123 

In light of the above I'd like to get senior-maintainer opinion on this.


api/contrib/envoy/extensions/matching/input_matchers/hyperscan/v3alpha/hyperscan.proto
```diff
@@ -14,14 +14,9 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // [#protodoc-title: Hyperscan matcher]
 // [#extension: envoy.matching.input_matchers.hyperscan]
 
-// `Hyperscan <https://github.com/intel/hyperscan>`_ regex matcher. The engine of the matcher uses
-// hybrid automata techniques to allow simultaneous matching of large numbers of regular
-// expressions across streams of data.
-//
-// The matcher follows PCRE pattern syntax, and the regex string must adhere to the documented
```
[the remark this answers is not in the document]
zhxie: The paragraph has been moved to  `api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto` since it is the introduction of the engine itself.

contrib/hyperscan/matching/input_matchers/source/config.cc
zhxie: Yes. In Hyperscan regex engine, I will set all these flags to match RE2's behavior.

contrib/hyperscan/matching/input_matchers/source/matcher.cc
zhxie: Then I think it will be fine to use `IS_ENVOY_BUG`, I will change all these unreachable exceptions to `IS_ENVOY_BUG`.

BTW, this is a contrib extension, and the CI would not cover it.

alyssawilk:
I had largely thought that with contrib we didn't have to monitor things like dependency bloat from boost or extra regex engines.  I'm going to throw senior maintianer pass over to Matt as he's back tomorrow and he was driving the recent changes to contrib.

mattklein123:
> In light of the above I'd like to get senior-maintainer opinion on this.

I'm OK with this one coming into contrib as I've heard of multiple requests for using HyperScan to speed up matching (for example in WAF scenarios). IMO it's OK to let this into contrib and see how it gets used? Happy to defer to other opinions if they are strong however.

alyssawilk:
SGTM.  @moderation can we get one more deps LGTM so we can merge this in?

jmarantz:
OK with me then. @zhxie were you going to change defaults in this PR to make it as easy as possible for a configuration that works with re2 to also work with hyperscan?

zhxie:
@jmarantz of course. As I have mentioned, I have removed all options in Hyperscan regex engine proto to align the behavior with RE2. https://github.com/envoyproxy/envoy/blob/e09647f8e258cc9ab18b43cf74767d0b2c20c303/api/contrib/envoy/extensions/regex_engines/hyperscan/v3alpha/hyperscan.proto#L24-L25

jmarantz:
Sorry still not totally getting it.

I was thinking when reviewing code earlier that some of the out-of-the-box options for Hyperscan around embedded newlines and unicode were different than what I think RE2 behavior is, and that we'd need explicit defaults to make them as compatible as practical.

jmarantz (approved):
Per slack discussion the defaults are set to maximize compatibility with RE2. Thanks for pushing this all the way through!

moderation:
/lgtm deps

This is a metadata only dependency change. The initial Hyperscan contrib extension that brings in unwanted dependencies like Boost was added prior to the current contrib extension policy - https://github.com/envoyproxy/envoy/blob/main/EXTENSION_POLICY.md#contrib-extensions. The Hyperscan input matcher and regex engine would likely no longer be approved without an end user sponsor. These extensions appear to be R&D activities for a vendor without end users.

sambercovici:
I see that hyperscan is compiled without "FAT_RUNTIME": "on" 
Was there any reason fore this?

PR #21974, opened 2022-07-01
Plumb header validator into HTTP/1 server codec

yanavlasov:
Additional Description:
This PR plumbs header validator into HTTP/1 codec. The code is disabled in normal builds and is behind the ENVOY_ENABLE_UHV flag, which is tested during compile_time_options build only.

H/1 client and H/2 codecs are to be done in a follow up PR.

Risk Level: Low, no changes for normal Envoy builds.
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A


mattklein123 (dismissed):
LGTM, thanks. Very excited to see this work progressing.

ggreenway (dismissed):
LGTM

source/common/http/http1/codec_impl.cc
```diff
@@ -49,6 +49,12 @@ struct Http1ResponseCodeDetailValues {
   const absl::string_view ChunkedContentLength = "http1.content_length_and_chunked_not_allowed";
   const absl::string_view HttpsInPlaintext = "http1.https_url_on_plaintext_connection";
   const absl::string_view InvalidScheme = "http1.invalid_scheme";
+  const absl::string_view InvalidHost = "http1.invalid_host";
+  const absl::string_view InvalidPath = "http1.invalid_path";
+  const absl::string_view InvalidMethod = "http1.invalid_method";
+  const absl::string_view InvalidStatus = "http1.invalid_status";
+  const absl::string_view InvalidHeader = "http1.invalid_header";
+  const absl::string_view InvalidHeaders = "http1.invalid_headers";
```
alyssawilk, line 57: is there value in disambiguating 1 from 1+?

source/common/http/http1/codec_impl.cc
```diff
@@ -507,6 +515,9 @@ Status ConnectionImpl::completeLastHeader() {
   ASSERT(dispatching_);
```
alyssawilk, line 515: optional as it's not your code, but consider renaming completeLastHeader as I read it as completing _the_ last header, not completing the most recent header to be processed

source/common/http/http1/codec_impl.cc
```diff
@@ -1072,10 +1102,16 @@ Status ServerConnectionImpl::handlePath(RequestHeaderMap& headers, absl::string_
   // request-target. A proxy that forwards such a request MUST generate a
   // new Host field-value based on the received request-target rather than
   // forward the received Host field-value.
+  RETURN_IF_ERROR(checkHeaderEntryAndSendError(Http::HeaderString(header_values.Host),
```
alyssawilk, line 1105: I can imagine these errors being pretty hard to debug without trace logs - I think previously the headers would land in the access logs, so you could debug what the invalid host was, where now we're protecting access logs and response filters from bad header data.  I think the loss of debugability is probably worse though so I'd be inclined to add the header and then check validity so we can debug or come up with some other workaround

source/common/http/http1/codec_impl.cc
```diff
@@ -1323,11 +1369,66 @@ void ServerConnectionImpl::ActiveRequest::dumpState(std::ostream& os, int indent
...
+#ifdef ENVOY_ENABLE_UHV
+  if (active_request_->header_validator_) {
+    auto validation_result =
+        active_request_->header_validator_->validateRequestHeaderMap(header_map);
+    if (validation_result == Http::HeaderValidator::RequestHeaderMapValidationResult::Reject ||
+        validation_result == Http::HeaderValidator::RequestHeaderMapValidationResult::Redirect) {
+      result = codecProtocolError("http/1.1 protocol error: invalid header map");
+      std::function<void(ResponseHeaderMap & headers)> modify_headers;
+      if (validation_result == Http::HeaderValidator::RequestHeaderMapValidationResult::Redirect &&
+          !Grpc::Common::hasGrpcContentType(header_map)) {
```
alyssawilk, line 1411: Also I _think_ before we mostly did checks of each header after processing all, so if we had a gRPC request and a bad header before gRPC content type, response would follow gRPC path, and now we'll send a standard error reply for headers we would have done "in envoy" (same path for prior http_parser errors).  I don't think it's deal breaker but we should consider documenting this wherever we document differences between the old and the new.

It's also a lot more work to do this in all three codecs.  Given we're probably going to have to have a "check these headers for errors" function for headers transformed by WASM or ext_authz and such, I wonder if we should simply do a one time pass on all headers after we read them and before the HCM processes them, rather than doing all this invasive plumbing into client and server codecs for HTTP/1, HTTP/2 and HTTP/3.  We're going to have to do a pass against all headers once we have all headers to make sure we catch things like 'no content length and transfer encoding' so I wonder if we should just do character validation there and skip all the work along the way.  Thoughts?

alyssawilk (commented):
Exciting to see this underway!
/wait

source/common/http/http1/codec_impl.cc
yanavlasov: Removed after refactor

source/common/http/http1/codec_impl.cc
yanavlasov: Will do in a separate PR

source/common/http/http1/codec_impl.cc
```diff
@@ -1056,6 +1084,8 @@ Status ServerConnectionImpl::handlePath(RequestHeaderMap& headers, absl::string_
   // CONNECT "urls" are actually host:port so look like absolute URLs to the above checks.
   // Absolute URLS in CONNECT requests will be rejected below by the URL class validation.
   if (!codec_settings_.allow_absolute_url_ && !is_connect) {
+    RETURN_IF_ERROR(checkHeaderEntryAndSendError(path, active_request_->request_url_,
```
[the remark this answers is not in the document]
yanavlasov: Refactored

source/common/http/http1/codec_impl.cc
yanavlasov: Refactored

source/common/http/http1/codec_impl.cc
yanavlasov: I have moved header validation into the HCM and moved it out of the codec. I think if we need to do checks for individual headers as they are being parsed, I can plumb it back into codecs. It is however performance optimization only, that allows header parsing to be stopped early once bad header is found.

source/common/http/conn_manager_impl.cc
```diff
@@ -896,6 +929,11 @@ void ConnectionManagerImpl::ActiveStream::decodeHeaders(RequestHeaderMapPtr&& he
     request_header_timer_.reset();
   }
 
+  if (!validateHeaders()) {
```
alyssawilk, line 932: one of the gains of moving this to the HCM is making sure we handle connection close properly for say HTTP/1.0 header only requests.  I encourage best-effort checks of the shouldCloseconnection before the validation and an integration test which future-proofs since apparently nothing failed.

source/common/http/conn_manager_impl.cc
```diff
@@ -896,6 +929,11 @@ void ConnectionManagerImpl::ActiveStream::decodeHeaders(RequestHeaderMapPtr&& he
...
+    ENVOY_STREAM_LOG(debug, "request headers validation failed:\n{}", *this, *request_headers_);
```
alyssawilk, line 933: encouraged follow-up (TODO?) we could have a default header validator, and move most of Envoy's current HCM checks there since they should be redundant with UHV
if (!request_headers_->Host())
  if ((!HeaderUtility::isConnect(*request_headers_) || request_headers_->Path()) &&
...

source/common/http/conn_manager_impl.h
```diff
@@ -351,6 +351,8 @@ class ConnectionManagerImpl : Logger::Loggable<Logger::Id::http>,
       return *tracing_custom_tags_;
     }
 
+    bool validateHeaders();
```
alyssawilk, line 354: comment, especially as today it doesn't always validate

source/extensions/filters/network/http_connection_manager/config.cc
```diff
@@ -127,6 +127,32 @@ envoy::extensions::filters::network::http_connection_manager::v3::HttpConnection
       KEEP_UNCHANGED;
 }
 
+Http::HeaderValidatorFactoryPtr
+createHeaderValidatorFactory([[maybe_unused]] const envoy::extensions::filters::network::
+                                 http_connection_manager::v3::HttpConnectionManager& config,
+                             [[maybe_unused]] Server::Configuration::FactoryContext& context) {
+
+  Http::HeaderValidatorFactoryPtr header_validator_factory;
+#ifdef ENVOY_ENABLE_UHV
```
alyssawilk, line 136: this is only ifdef'd out until we have runtime controls of underlying header validation correct?

Should we throw an exception if it's configured on but not supported due to compile time options?

source/server/admin/admin.h
```diff
@@ -211,6 +211,10 @@ class AdminImpl : public Admin,
   const HttpConnectionManagerProto::ProxyStatusConfig* proxyStatusConfig() const override {
     return proxy_status_config_.get();
   }
+  Http::HeaderValidatorPtr makeHeaderValidator(Http::Protocol, StreamInfo::StreamInfo&) override {
+    // TODO(yanavlasov): admin interface should use the default validator
```
alyssawilk, line 215: should this ENVOY_BUG?  seems like it's pretty unsafe to do admin requests until this is fixed

alyssawilk (commented):
Nice, I think this will be much simpler.  
/wait

source/common/http/http1/codec_impl.cc
yanavlasov: Done in PR #22408

source/server/admin/admin.h
yanavlasov: Done

source/common/http/conn_manager_impl.h
yanavlasov: Done

source/extensions/filters/network/http_connection_manager/config.cc
yanavlasov: Done

source/common/http/conn_manager_impl.cc
yanavlasov: This is on my TODO list and PR with these changes will be out soon.

source/common/http/conn_manager_impl.cc
yanavlasov: You right, I missed that. Added tests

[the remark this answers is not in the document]
yanavlasov: Done

source/common/http/conn_manager_impl.cc
```diff
@@ -866,6 +867,38 @@ uint32_t ConnectionManagerImpl::ActiveStream::localPort() {
...
+          response_headers.addReferenceKey(Http::Headers::get().Location, new_path);
+        };
+      } else if (is_grpc) {
+        grpc_status = Grpc::Status::WellKnownGrpcStatus::Internal;
+      }
+
+      auto response_details_opt = filter_manager_.streamInfo().responseCodeDetails();
+
+      sendLocalReply(response_code, "", modify_headers, grpc_status,
+                     response_details_opt
```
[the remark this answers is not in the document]
yanavlasov: Done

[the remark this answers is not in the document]
alyssawilk: sorry missed this but it still looks optional.  We could add details to the function signature and ENVOY_BUG that it's not empty?

alyssawilk (commented):
LGTM modulo one last nit and CI failure
/wait
/retest

alyssawilk (dismissed):
making that last comment optional :-)

source/server/admin/admin.h
yanavlasov: Had to undo ENVOY_BUG as it crashed all integration tests with UHV enabled. This item is on my internal OKR tracker and will not be dropped.

yanavlasov:
Changed the return type of header validator to be a tuple of action and details string and enforced non empty string at construction.

envoy/http/header_validator.h
```diff
@@ -18,12 +21,39 @@ class HeaderValidator {
...
 
-  enum class HeaderEntryValidationResult { Accept, Reject };
+  // A class that holds either success condition or an error condition with tuple of
+  // action and error details.
+  template <typename ActionType> class Result {
+  public:
+    using Action = ActionType;
+
+    // Helper for constructing successful results
+    static Result success() { return Result(ActionType::Accept, absl::string_view()); }
```
alyssawilk, line 31: This needs a TODO for removal since it encourages empty string views, no?

yanavlasov: We only need details if header map was invalid, i.e. when result was either `reject` or `redirect`. If header map passed, the details are empty, which is what this method is used for.

alyssawilk: ugh sorry yes, my bad.

alyssawilk (approved):

Issue #21178, opened 2022-05-06
Support for upstream network filter


surki:
*Title*: We have a need for upstream RBAC or ext-authz filters

*Description*:
We are using Envoy as forward proxy (i.e., "dynamic forward proxy"). We would like to apply some network policy, specifically on the IPs it is connecting to. So we would need support for that in the upstream context.

There was some discussion [here](https://github.com/envoyproxy/envoy/issues/10525), so I went ahead and made the changes to see if existing network RBAC filter would work in the context of upstream. The changes are [here](https://github.com/surki/envoy/commit/25720a3229f1cbbf26487df0a29327121094b715) and it seems to work.

Few quick observations:

- Since the existing RBAC filter assumes ingress/downstream, the names are inverted, it looks bit confusing to use it this way.
```
  clusters:
  - name: foo
    ....
    ....
    ....
    filters:
    - name: envoy.filters.network.rbac
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.filters.network.rbac.v3.RBAC
        rules:
          action: DENY
          policies:
            "loopback-linklocal":
              permissions:
              - any: true
              principals:
              - or_ids:
                  ids:
                  - remote_ip:
                      address_prefix: 127.0.0.0
                      prefix_len: 8
                  - remote_ip:
                      address_prefix: 169.254.0.0
                      prefix_len: 16

```

- When policy is enforced (say by closing connection), the connection termination information is not propagated back (it is kept in the context of upstream connection context)

I am wondering what's the right approach should be here? 

1. Create a new upstream RBAC filter with its own .proto definition with the names fixed/relevant in the upstream context
2. Re-use the ingress filter as it is (like done  [here](https://github.com/surki/envoy/commit/25720a3229f1cbbf26487df0a29327121094b715)) and just fix up the documentation that when used in upstream context, fields are inverted etc.

Option 1 might be better for long term.

lizan:
@yangminzhu @yanavlasov

mattklein123:
I agree that (1) is the right approach, with sharing as much code as possible. It would be great to see this happen!

surki:
FYI I just started looking into this, I will send a draft version for feedback soon

wjrbetts:
We would love to see this feature so that we can limit the damage that a semi-trusted client/server can do via a dynamic forward proxy. 

It looks like the PR that @surki raised was closed while waiting for changes around upstream filters to stabilise. Has this stabilised now @alyssawilk?

[...]

PR #21196, opened 2022-05-09
docs: add warning about confused deputy attacks on DFP.

htuch:
Fixes #19429

Signed-off-by: Harvey Tuch <htuch@google.com>

PR #21985, opened 2022-07-01
[WIP] Add upstream network RBAC filter

surki:
- Reuses the RBAC downstream filter, introduces new matcher/fields for
  upstream ip and port
- Propagates information from upstream filter back to downstream context if
  any

TODO:

- Downstream context is not available in upstream filter, find a way to
  propagate (so we can say `application with IP 1.2.3.4 can talk to upstream
  IP 4.5.6.7` etc)
- RBAC rule should be enforced before connection establishment, currently
  does it after establishment
- Documentation

Signed-off-by: Suresh Kumar <sureshkumar.pp@gmail.com>

Example usage:

```
static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address:
        protocol: TCP
        address: 0.0.0.0
        port_value: 1234
    filter_chains:
    - filters:
        typed_config:
          stat_prefix: ingress_http
          route_config:
            name: local_route
            virtual_hosts:
            - name: local_service
              domains: ["*"]
              routes:
              - match:
                  prefix: "/"
                route:
                  cluster: example_cluster
                  timeout: 1500s
          http_filters:
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
  clusters:
  - name: example_cluster
    type: LOGICAL_DNS
    connect_timeout: 5s
    dns_lookup_family: V4_ONLY
    respect_dns_ttl: true
    dns_refresh_rate: 3000s
    load_assignment:
      cluster_name: example.com|8080
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: example.com
                port_value: 8080
    filters:            
    - name: envoy.filters.network.rbac
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.filters.network.rbac.v3.RBAC
        stat_prefix: rbac_upstream
        rules:
          action: DENY
          policies:
            "loopback":
              principals:
              - any: true
              permissions:
              - upstream_destination_ip:   
                  address_prefix: 127.0.0.1
                  prefix_len: 8
```

Commit Message: Adds new upstream network RBAC filter
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA] #21178
[Optional Deprecated:]

### DIFF
--- a/api/envoy/config/rbac/v3/rbac.proto
+++ b/api/envoy/config/rbac/v3/rbac.proto
@@ -148,7 +148,7 @@ message Policy {
 }
 
 // Permission defines an action (or actions) that a principal can take.
-// [#next-free-field: 13]
+// [#next-free-field: 16]
 message Permission {
   option (udpa.annotations.versioning).previous_message_type = "envoy.config.rbac.v2.Permission";
 
@@ -224,6 +224,18 @@ message Permission {
     // Extension for configuring custom matchers for RBAC.
     // [#extension-category: envoy.rbac.matchers]
     core.v3.TypedExtensionConfig matcher = 12;
+
+    // A CIDR block that describes the upstream destination IP.
+    // Note: Only valid in upstream filter context
+    core.v3.CidrRange upstream_destination_ip = 13;
+
+    // A port number that describes the destination port connecting to.
+    // Note: Only valid in upstream filter context
+    uint32 upstream_destination_port = 14 [(validate.rules).uint32 = {lte: 65535}];
+
+    // A port number range that describes a range of destination ports connecting to.
+    // Note: Only valid in upstream filter context
+    type.v3.Int32Range upstream_destination_port_range = 15;
   }
 }
 
--- a/envoy/matcher/matcher.h
+++ b/envoy/matcher/matcher.h
@@ -17,7 +17,7 @@ namespace Envoy {
 
 namespace Server {
 namespace Configuration {
-class ServerFactoryContext;
+class CommonFactoryContext;
 }
 } // namespace Server
 
@@ -174,7 +174,7 @@ class InputMatcherFactory : public Config::TypedFactory {
 public:
   virtual InputMatcherFactoryCb
   createInputMatcherFactoryCb(const Protobuf::Message& config,
-                              Server::Configuration::ServerFactoryContext& factory_context) PURE;
+                              Server::Configuration::CommonFactoryContext& factory_context) PURE;
 
   std::string category() const override { return "envoy.matching.input_matchers"; }
 };
@@ -293,7 +293,7 @@ template <class DataType> class CustomMatcherFactory : public Config::TypedFacto
 public:
   virtual MatchTreeFactoryCb<DataType>
   createCustomMatcherFactoryCb(const Protobuf::Message& config,
-                               Server::Configuration::ServerFactoryContext& factory_context,
+                               Server::Configuration::CommonFactoryContext& factory_context,
                                DataInputFactoryCb<DataType> data_input,
                                absl::optional<OnMatchFactoryCb<DataType>> on_no_match,
                                OnMatchFactory<DataType>& on_match_factory) PURE;
--- a/envoy/stream_info/stream_info.h
+++ b/envoy/stream_info/stream_info.h
@@ -433,6 +433,10 @@ public:
 
   virtual void setUpstreamProtocol(Http::Protocol protocol) PURE;
   virtual absl::optional<Http::Protocol> upstreamProtocol() const PURE;
+
+  virtual void
+  setUpstreamConnectionTerminationDetails(absl::string_view connection_termination_details) PURE;
+  virtual const absl::optional<std::string>& upstreamConnectionTerminationDetails() const PURE;
 };
 
 /**
--- a/source/common/formatter/substitution_formatter.cc
+++ b/source/common/formatter/substitution_formatter.cc
@@ -974,6 +974,17 @@ const StreamInfoFormatter::FieldExtractorLookupTbl& StreamInfoFormatter::getKnow
                                     return stream_info.connectionTerminationDetails();
                                   });
                             }}},
+                          {"UPSTREAM_CONNECTION_TERMINATION_DETAILS",
+                           {CommandSyntaxChecker::COMMAND_ONLY,
+                            [](const std::string&, const absl::optional<size_t>&) {
+                              return std::make_unique<StreamInfoStringFieldExtractor>(
+                                  [](const StreamInfo::StreamInfo& stream_info) {
+                                    return stream_info.upstreamInfo()
+                                               ? stream_info.upstreamInfo()
+                                                     ->upstreamConnectionTerminationDetails()
+                                               : absl::nullopt;
+                                  });
+                            }}},
                           {"BYTES_SENT",
                            {CommandSyntaxChecker::COMMAND_ONLY,
                             [](const std::string&, const absl::optional<size_t>&) {
--- a/source/common/matcher/matcher.h
+++ b/source/common/matcher/matcher.h
@@ -149,9 +149,9 @@ template <class DataType, class ActionFactoryContext>
 class MatchTreeFactory : public OnMatchFactory<DataType> {
 public:
   MatchTreeFactory(ActionFactoryContext& context,
-                   Server::Configuration::ServerFactoryContext& factory_context,
+                   Server::Configuration::CommonFactoryContext& factory_context,
                    MatchTreeValidationVisitor<DataType>& validation_visitor)
-      : action_factory_context_(context), server_factory_context_(factory_context),
+      : action_factory_context_(context), common_factory_context_(factory_context),
         match_input_factory_(factory_context.messageValidationVisitor(), validation_visitor) {}
 
   // TODO(snowp): Remove this type parameter once we only have one Matcher proto.
@@ -283,8 +283,8 @@ private:
           matcher.matcher_tree().custom_match());
       ProtobufTypes::MessagePtr message = Config::Utility::translateAnyToFactoryConfig(
           matcher.matcher_tree().custom_match().typed_config(),
-          server_factory_context_.messageValidationVisitor(), factory);
-      return factory.createCustomMatcherFactoryCb(*message, server_factory_context_, data_input,
+          common_factory_context_.messageValidationVisitor(), factory);
+      return factory.createCustomMatcherFactoryCb(*message, common_factory_context_, data_input,
                                                   on_no_match, *this);
     }
     }
@@ -323,11 +323,11 @@ private:
       auto& factory = Config::Utility::getAndCheckFactory<ActionFactory<ActionFactoryContext>>(
           on_match.action());
       ProtobufTypes::MessagePtr message = Config::Utility::translateAnyToFactoryConfig(
-          on_match.action().typed_config(), server_factory_context_.messageValidationVisitor(),
+          on_match.action().typed_config(), common_factory_context_.messageValidationVisitor(),
           factory);
 
       auto action_factory = factory.createActionFactoryCb(
-          *message, action_factory_context_, server_factory_context_.messageValidationVisitor());
+          *message, action_factory_context_, common_factory_context_.messageValidationVisitor());
       return [action_factory] { return OnMatch<DataType>{action_factory, {}}; };
     }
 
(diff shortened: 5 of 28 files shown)


surki:
@mattklein123 @snowp @yanavlasov @markdroth Please provide feedback on this, wondering if you can do a quick skim through and see if this  approach (using same filter/config in both downstream and upstream filters) is acceptable?

Additionally, need some input on how to do the following:

1. Currently there seems to be no callback in filter before the connection is established. Did I miss it somewhere? I want to enforce  RBAC rules even before the connection is established
2. I want to pass the information from downstream connection into upstream filter context (so we can say `application with IP 1.2.3.4 can talk to upstream cluster with IP 4.5.6.7` etc). Wondering if you can provide some input on how I can go about it.

zhxie:
> Additionally, need some input on how to do the following:
> 
> 1. Currently there seems to be no callback in filter before the connection is established. Did I miss it somewhere? I want to enforce  RBAC rules even before the connection is established
> 2. I want to pass the information from downstream connection into upstream filter context (so we can say `application with IP 1.2.3.4 can talk to upstream cluster with IP 4.5.6.7` etc). Wondering if you can provide some input on how I can go about it.

Reply questions:

1. Can you check if `Network::ReadFilter::onNewConnection()` satisfies your needs? The method will get called when the connection between downstream and Envoy established.
2. For every connection/request, a corresponding network filter/HTTP filter will be established. So you can just save everything you need in the filter on downstream connection and use them later.

surki:
> > Additionally, need some input on how to do the following:
> > 
> > 1. Currently there seems to be no callback in filter before the connection is established. Did I miss it somewhere? I want to enforce  RBAC rules even before the connection is established
> > 2. I want to pass the information from downstream connection into upstream filter context (so we can say `application with IP 1.2.3.4 can talk to upstream cluster with IP 4.5.6.7` etc). Wondering if you can provide some input on how I can go about it.
> 
> Reply questions:
> 
> 1. Can you check if `Network::ReadFilter::onNewConnection()` satisfies your needs? The method will get called when the connection between downstream and Envoy established.
`onNewConnection` is called after the connection is established. I guess we would have to introduce a new callback, something along the line of "onAccept" callback, so we can enforce rules before the connection is established.

> 2. For every connection/request, a corresponding network filter/HTTP filter will be established. So you can just save everything you need in the filter on downstream connection and use them later.
I guess I wasn't clear enough: since there are two "connections" involved (`connection one` being "client to envoy listener" and `connection two` being "envoy to upstream host"), in the `connection two` context (which is where the upstream filter would be instantiated), I will need access to the information about `connection one`. Just went through [router.cc](https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/common/router/router.cc#L686), we don't seem to pass anything about downstream to upstream_request object, so was wondering if this is already being passed somewhere else.

zhxie:
> > > Additionally, need some input on how to do the following:
> > > 
> > > 1. Currently there seems to be no callback in filter before the connection is established. Did I miss it somewhere? I want to enforce  RBAC rules even before the connection is established
> > > 2. I want to pass the information from downstream connection into upstream filter context (so we can say `application with IP 1.2.3.4 can talk to upstream cluster with IP 4.5.6.7` etc). Wondering if you can provide some input on how I can go about it.
> > 
> > 
> > Reply questions:
> > 
> > 1. Can you check if `Network::ReadFilter::onNewConnection()` satisfies your needs? The method will get called when the connection between downstream and Envoy established.
> 
> `onNewConnection` is called after the connection is established. I guess we would have to introduce a new callback, something along the line of "onAccept" callback, so we can enforce rules before the connection is established.
> 
> > 2. For every connection/request, a corresponding network filter/HTTP filter will be established. So you can just save everything you need in the filter on downstream connection and use them later.
> 
> I guess I wasn't clear enough: since there are two "connections" involved (`connection one` being "client to envoy listener" and `connection two` being "envoy to upstream host"), in the `connection two` context (which is where the upstream filter would be instantiated), I will need access to the information about `connection one`. Just went through [router.cc](https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/common/router/router.cc#L686), we don't seem to pass anything about downstream to upstream_request object, so was wondering if this is already being passed somewhere else.

1. I think `onNewConnection` will be called after the connection between downstream and Envoy gets established. And I think what you need is to avoid establishing connection between Envoy and upstream. There is no need to add something like `onAccept`.

2. Which way? From Envoy to the upstream (like connects to which cluster), you can use metadata. From the upstream to Envoy, you can just use any contexts in the filter.

surki:
> > > 1. Can you check if `Network::ReadFilter::onNewConnection()` satisfies your needs? The method will get called when the connection between downstream and Envoy established.
> > 
> > 
> > `onNewConnection` is called after the connection is established. I guess we would have to introduce a new callback, something along the line of "onAccept" callback, so we can enforce rules before the connection is established.
> > > 2. For every connection/request, a corresponding network filter/HTTP filter will be established. So you can just save everything you need in the filter on downstream connection and use them later.
> > 
> > 
> > I guess I wasn't clear enough: since there are two "connections" involved (`connection one` being "client to envoy listener" and `connection two` being "envoy to upstream host"), in the `connection two` context (which is where the upstream filter would be instantiated), I will need access to the information about `connection one`. Just went through [router.cc](https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/common/router/router.cc#L686), we don't seem to pass anything about downstream to upstream_request object, so was wondering if this is already being passed somewhere else.
> 
> 1. I think `onNewConnection` will be called after the connection between downstream and Envoy gets established. And I think what you need is to avoid establishing connection between Envoy and upstream. There is no need to add something like `onAccept`.

Right, we need to check (via upstream RBAC filter) and potentially stop envoy establishing upstream connection. What's the right filter API/callback to do this? (`onNewConnection` is too late for it, the connection is already established, I don't see any filter callbacks in `ClientConnectionImpl::connect()` or before that).

> 2. Which way? From Envoy to the upstream (like connects to which cluster), you can use metadata. From the upstream to Envoy, you can just use any contexts in the filter.

Not sure I understood this completely, can you please give a pointer to code somewhere that does it? Just to be on the same page: this RBAC filter is instantiated in upstream connection context (so `router.cc` -> `upstream_request.cc` -> `ClientConnectionImpl` etc), once I get a callback in, say `onData`, how do I refer to the downstream listener side connection context? (which I suppose could be on completely another  threadpool, with its own lifecycle etc)

zhxie:
> > > > 1. Can you check if `Network::ReadFilter::onNewConnection()` satisfies your needs? The method will get called when the connection between downstream and Envoy established.
> > > 
> > > 
> > > `onNewConnection` is called after the connection is established. I guess we would have to introduce a new callback, something along the line of "onAccept" callback, so we can enforce rules before the connection is established.
> > > > 2. For every connection/request, a corresponding network filter/HTTP filter will be established. So you can just save everything you need in the filter on downstream connection and use them later.
> > > 
> > > 
> > > I guess I wasn't clear enough: since there are two "connections" involved (`connection one` being "client to envoy listener" and `connection two` being "envoy to upstream host"), in the `connection two` context (which is where the upstream filter would be instantiated), I will need access to the information about `connection one`. Just went through [router.cc](https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/common/router/router.cc#L686), we don't seem to pass anything about downstream to upstream_request object, so was wondering if this is already being passed somewhere else.
> > 
> > 
> > 
> > 1. I think `onNewConnection` will be called after the connection between downstream and Envoy gets established. And I think what you need is to avoid establishing connection between Envoy and upstream. There is no need to add something like `onAccept`.
> 
> Right, we need to check (via upstream RBAC filter) and potentially stop envoy establishing upstream connection. What's the right filter API/callback to do this? (`onNewConnection` is too late for it, the connection is already established, I don't see any filter callbacks in `ClientConnectionImpl::connect()` or before that).
> 
> > 2. Which way? From Envoy to the upstream (like connects to which cluster), you can use metadata. From the upstream to Envoy, you can just use any contexts in the filter.
> 
> Not sure I understood this completely, can you please give a pointer to code somewhere that does it? Just to be on the same page: this RBAC filter is instantiated in upstream connection context (so `router.cc` -> `upstream_request.cc` -> `ClientConnectionImpl` etc), once I get a callback in, say `onData`, how do I refer to the downstream listener side connection context? (which I suppose could be on completely another threadpool, with its own lifecycle etc)

1. No, when `onNewConnection` gets called, the upstream connection has not been established yet.
2. For example, you can check the how this variable is used in local rate limit HTTP filter. https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/extensions/filters/http/local_ratelimit/local_ratelimit.h#L163

surki:
> > > > > 1. Can you check if `Network::ReadFilter::onNewConnection()` satisfies your needs? The method will get called when the connection between downstream and Envoy established.
> > > > 
> > > > 
> > > > `onNewConnection` is called after the connection is established. I guess we would have to introduce a new callback, something along the line of "onAccept" callback, so we can enforce rules before the connection is established.
> > > > > 2. For every connection/request, a corresponding network filter/HTTP filter will be established. So you can just save everything you need in the filter on downstream connection and use them later.
> > > > 
> > > > 
> > > > I guess I wasn't clear enough: since there are two "connections" involved (`connection one` being "client to envoy listener" and `connection two` being "envoy to upstream host"), in the `connection two` context (which is where the upstream filter would be instantiated), I will need access to the information about `connection one`. Just went through [router.cc](https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/common/router/router.cc#L686), we don't seem to pass anything about downstream to upstream_request object, so was wondering if this is already being passed somewhere else.
> > > 
> > > 
> > > 
> > > 1. I think `onNewConnection` will be called after the connection between downstream and Envoy gets established. And I think what you need is to avoid establishing connection between Envoy and upstream. There is no need to add something like `onAccept`.
> > 
> > 
> > Right, we need to check (via upstream RBAC filter) and potentially stop envoy establishing upstream connection. What's the right filter API/callback to do this? (`onNewConnection` is too late for it, the connection is already established, I don't see any filter callbacks in `ClientConnectionImpl::connect()` or before that).
> > > 2. Which way? From Envoy to the upstream (like connects to which cluster), you can use metadata. From the upstream to Envoy, you can just use any contexts in the filter.
> > 
> > 
> > Not sure I understood this completely, can you please give a pointer to code somewhere that does it? Just to be on the same page: this RBAC filter is instantiated in upstream connection context (so `router.cc` -> `upstream_request.cc` -> `ClientConnectionImpl` etc), once I get a callback in, say `onData`, how do I refer to the downstream listener side connection context? (which I suppose could be on completely another threadpool, with its own lifecycle etc)
> 
> 1. No, when `onNewConnection` gets called, the upstream connection has not been established yet.

true in downstream context, but not in upstream filter context. (and there is no reliable way in downstream filter context to check/enforce policy on upstream connections: dns lookups, retries etc happen later...)

> 2. For example, you can check the how this variable is used in local rate limit HTTP filter. https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/source/extensions/filters/http/local_ratelimit/local_ratelimit.h#L163

This is only referring current context, not the upstream.

zhxie:
Sorry, I am still not clear about your requests, maybe seniors can answer your questions.

BTW, if you just want to check upstream IP and port, does not [RBAC upstream IP and port matcher](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/rbac/matchers/upstream_ip_port/v3/upstream_ip_port_matcher.proto.html) satisfy your needs?

surki:
> BTW, if you just want to check upstream IP and port, does not [RBAC upstream IP and port matcher](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/rbac/matchers/upstream_ip_port/v3/upstream_ip_port_matcher.proto.html) satisfy your needs?

thanks @zhxie. `UpstreamIpPortMatcher` is what we are currently using, but it is not safe to use, it is susceptible to the confused-deputy/dns-rebinding attack, please see [here](https://github.com/envoyproxy/envoy/pull/21196/files). Based on the discussions on that, it appears the safest way to check/enforce the policy is to have an upstream filter (which is instantiated during the upstream connection establishment...). This PR is to add that feature. We are trying to re-use the code as much as possible between the downstream and upstream RBAC filters.

surki:
> yeah fwiw I have a series of changes which should largely do away with commonserverfactorycontext and replace with serverfactorycontext which should fix up a lot of the class changes here

Thanks, will wait for your changes.
@alyssawilk Wondering if you are also setting up infra to pass the downstream information to upstream request context (which is needed for upstream filters)?

alyssawilk:
not yet, no.  I think @kyessenov _may_ have done some of that?  But if you want to work on that as a standalone that'd be helpful

kyessenov:
I have done sharing with upstream only for the filter state. Some basic fields like SSH info need separate plumbing (and answer how we handle sharing of upstream connections by multiple downstream connections).

PR #21987, opened 2022-07-01
Implement no throw version of `unpackTo` to avoid throwing exception on dataplane

tyxia:
Return the error message to the caller site and let caller decides how to handle it (e.g., logging etc) instead of throwing the exception.

### DIFF
--- a/source/common/protobuf/utility.cc
+++ b/source/common/protobuf/utility.cc
@@ -562,14 +562,15 @@ void MessageUtil::unpackTo(const ProtobufWkt::Any& any_message, Protobuf::Messag
   }
 }
 
-absl::optional<std::string> MessageUtil::unpackToNoThrow(const ProtobufWkt::Any& any_message,
-                                                         Protobuf::Message& message) {
+ProtobufUtil::Status MessageUtil::unpackToNoThrow(const ProtobufWkt::Any& any_message,
+                                                  Protobuf::Message& message) {
   if (!any_message.UnpackTo(&message)) {
-    return absl::StrCat("Unable to unpack as ", message.GetDescriptor()->full_name(), ": ",
-                        any_message.DebugString());
+    return ProtobufUtil::Status(google::protobuf::util::StatusCode::kInternal,
+                                "Unable to unpack as " + message.GetDescriptor()->full_name() +
+                                    ": " + any_message.DebugString());
   }
-  // No error message is returned if `UnpackTo` succeeded.
-  return absl::nullopt;
+  // Ok Status is returned if `UnpackTo` succeeded.
+  return ProtobufUtil::OkStatus();
 }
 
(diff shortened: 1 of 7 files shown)


tyxia:
/assign @yanavlasov 

PTAL, Thanks!

yanavlasov (changes requested):
/wait

source/common/protobuf/utility.cc
```diff
@@ -562,6 +562,17 @@ void MessageUtil::unpackTo(const ProtobufWkt::Any& any_message, Protobuf::Messag
   }
 }
 
+absl::Status MessageUtil::unpackToNoThrow(const ProtobufWkt::Any& any_message,
+                                          Protobuf::Message& message) {
+  if (!any_message.UnpackTo(&message)) {
+    return {absl::StatusCode::kInternal, "Unable to unpack as " +
```
yanavlasov, line 568:
```suggestion
    return absl::InternalError(absl::StrCat("Unable to unpack as ", message.GetDescriptor()->full_name(), ": ",
                                            any_message.DebugString());
```

yanavlasov (changes requested):
/wait

source/common/protobuf/utility.cc
```diff
@@ -562,6 +562,16 @@ void MessageUtil::unpackTo(const ProtobufWkt::Any& any_message, Protobuf::Messag
   }
 }
 
+absl::optional<std::string> MessageUtil::unpackToNoThrow(const ProtobufWkt::Any& any_message,
```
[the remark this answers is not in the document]
tyxia: Make sense! Done.

tyxia:
Thanks for the review. PTAL @yanavlasov 

I have addressed the comments and the only CI failure seems to be caused by test environment (stopped hearing from agent i-0ba525d18868b93b4).

source/common/protobuf/utility.cc
```diff
@@ -562,6 +562,17 @@ void MessageUtil::unpackTo(const ProtobufWkt::Any& any_message, Protobuf::Messag
   }
 }
 
+absl::Status MessageUtil::unpackToNoThrow(const ProtobufWkt::Any& any_message,
+                                          Protobuf::Message& message) {
+  if (!any_message.UnpackTo(&message)) {
+    return absl::InternalError(absl::StrCat("Unable to unpack as " +
```
yanavlasov, line 568: Concatenating strings with + is very inefficient. If replace `+` with `,` the StrCat will make it a lot better.

yanavlasov (changes requested):
/wait

source/common/protobuf/utility.cc
tyxia: oops, I used strcat but forgot to change the `+`.
Thanks! Done.

PR #21990, opened 2022-07-01
http: Do not add content-length: 0 header to UPGRADE requests

yanavlasov:
Additional Description:
Upgrade requests with `content-length: 0` fail on some proxies as they assume no payload will be sent. With this change Envoy will no longer be adding `content-length: 0` header if it was not present in the upstream UPGRADE request.

Risk Level: Medium, Runtime Guarded
Testing: Unit Tests
Docs Changes: N/A
Release Notes: Yes
Runtime guard: envoy.reloadable_features.http_skip_adding_content_length_to_upgrade


source/common/http/conn_manager_utility.cc
```diff
@@ -87,9 +87,13 @@ ConnectionManagerUtility::MutateRequestHeadersResult ConnectionManagerUtility::m
     // content-length both do not exist this means there is no request body. After transfer-encoding
     // is stripped here, the upstream request becomes invalid. We can fix it by explicitly adding a
     // "content-length: 0" request header here.
-    const bool no_body = (!request_headers.TransferEncoding() && !request_headers.ContentLength());
-    if (no_body) {
-      request_headers.setContentLength(uint64_t(0));
+    if (!Runtime::runtimeFeatureEnabled(
```
alyssawilk, line 90: Move the comment into the new if block or remove?

source/common/http/http1/codec_impl.cc
```diff
@@ -439,6 +439,10 @@ Status RequestEncoderImpl::encodeHeaders(const RequestHeaderMap& headers, bool e
   }
   if (Utility::isUpgrade(headers)) {
     upgrade_request_ = true;
```
alyssawilk, line 441: Do we need to do the same on the response path as well?

source/common/http/http1/codec_impl.cc
```diff
@@ -439,6 +439,10 @@ Status RequestEncoderImpl::encodeHeaders(const RequestHeaderMap& headers, bool e
...
+    if (Runtime::runtimeFeatureEnabled(
+            "envoy.reloadable_features.http_skip_adding_content_length_to_upgrade")) {
+      disableChunkEncoding();
```
alyssawilk, line 444: note: this arguably isn't "safe" to flip from true to false live, as it'll corrupt any upgrade requests which have had headers read, but not sent.  I'm OK with this but I want to make sure we're doing it mindfully.

test/integration/websocket_integration_test.cc
```diff
@@ -421,6 +421,51 @@ TEST_P(WebsocketIntegrationTest, BidirectionalChunkedData) {
   response_headers.removeContentLength();
   performUpgrade(request_headers, response_headers);
 
+  // Client codec should not add TE or CL:0 headers to UPGRADE requests without content-length
+  ASSERT_TRUE(upstream_request_->headers().TransferEncoding() == nullptr);
```
alyssawilk, line 425: I think it's worth having a test in integration_test.cc or a clone of those, which looks at raw headers.  The codec would strip TE so this check is a functional no-op and I believe would pass without the second part of your change.  AFIK the body payload would also be untouched either way, because once we hit the upgrade path we direct-encode all data.

alyssawilk (commented):
Awesome, thanks for picking this up
/wait

source/common/http/conn_manager_utility.cc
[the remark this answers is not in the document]
alyssawilk: If not, please at least file a tracking bug for this?

alyssawilk (commented):
cool, just a couple of test nits and you're good to go
/wait

source/common/http/http1/codec_impl.cc
yanavlasov: Added comment

test/integration/websocket_integration_test.cc
```diff
@@ -449,7 +454,56 @@ TEST_P(WebsocketIntegrationTest, BidirectionalNoContentLengthNoTransferEncoding)
   ASSERT_TRUE(waitForUpstreamDisconnectOrReset());
 }
 
-TEST_P(WebsocketIntegrationTest, BidirectionalChunkedDataLegacyAddTE) {
+TEST_P(WebsocketIntegrationTest, BidirectionalNoContentLengthNoTransferEncoding) {
+  if (downstreamProtocol() == Http::CodecType::HTTP2 ||
```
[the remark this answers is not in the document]
yanavlasov: done

test/integration/websocket_integration_test.cc
```diff
@@ -449,7 +454,56 @@ TEST_P(WebsocketIntegrationTest, BidirectionalNoContentLengthNoTransferEncoding)
...
+
+  tcp_client->waitForData("\r\n\r\n", false);
+  // Make sure Envoy did not add TE or CL on the response path
+  ASSERT_FALSE(absl::StrContains(tcp_client->data(), "content-length"));
+  ASSERT_FALSE(absl::StrContains(tcp_client->data(), "transfer-encoding"));
+
+  fake_upstream_connection->clearData();
+  // Send data and make sure Envoy did not add chunk framing
+  ASSERT_TRUE(tcp_client->write("foo bar", false, false));
+  ASSERT_TRUE(fake_upstream_connection->waitForData(
```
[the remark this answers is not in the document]
yanavlasov: We do not have method like this for raw upstream connections. I however added a helper similar to `waitForInexactMatch` to make more readable.

alyssawilk (dismissed):

alyssawilk (approved):

PR #22005, opened 2022-07-04
listener: add multiple addresses support for ListenerImpl and ListenerManagerImpl

soulxu:
Commit Message: listener: add multiple addresses support for ListenerImpl and ListenerManagerImpl
Additional Description:
This PR enables the ListenerImpl to parse the multiple addresses from the new field `additional_addresses` from the Listener API. Also enable ListenerManagerImpl to create socket for each address in the Listener.

Risk Level: high
Testing: unittest
Docs Changes: n/a
Release Notes: n/a


soulxu:
emm...a little big of this PR, let me know if prefer to separate this PR

source/server/listener_manager_impl.cc
```diff
@@ -373,6 +373,24 @@ ListenerManagerStats ListenerManagerImpl::generateStats(Stats::Scope& scope) {
 
 bool ListenerManagerImpl::addOrUpdateListener(const envoy::config::listener::v3::Listener& config,
                                               const std::string& version_info, bool added_via_api) {
+  std::string name;
+  if (!config.name().empty()) {
+    name = config.name();
+  } else {
+    name = server_.api().randomGenerator().uuid();
```
daixiang0, line 380: Is a random uuid helpful for users as it will be used in the exception below?

envoy/network/listener.h
```diff
@@ -164,7 +164,7 @@ class ListenerConfig {
    *         A listener that doesn't listen on a port can only receive connections
    *         redirected from other listeners.
    */
-  virtual bool bindToPort() PURE;
+  virtual bool bindToPort() const PURE;
```
daixiang0, line 167: Is it necessary or we can get benefit from it?

source/server/listener_impl.cc
```diff
@@ -353,20 +350,22 @@ ListenerImpl::ListenerImpl(const envoy::config::listener::v3::Listener& config,
...
-       (!address_->ip()->isAnyAddress() &&
-        address_->ip()->ipv6()->v4CompatibleAddress() == nullptr))) {
-    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.strict_check_on_ipv4_compat")) {
-      throw EnvoyException(fmt::format(
-          "Only IPv6 address '::' or valid IPv4-mapped IPv6 address can set ipv4_compat: {}",
-          address_->asStringView()));
-    } else {
-      ENVOY_LOG(warn, "An invalid IPv4-mapped IPv6 address is used when ipv4_compat is set: {}",
-                address_->asStringView());
+  // All the addresses should be same socket type, so get the first address's socket type is enough.
```
daixiang0, line 353: Maybe I miss the context, is it designed like this or will it be optimized in the future?

envoy/network/listener.h
soulxu: This is used by `bool hasDuplicatedAddress(const ListenerImpl& other) const` https://github.com/envoyproxy/envoy/pull/22005/files#diff-e7a517ffa60d492c4979c569cc2857ed4bdc89e9d188ae3167a777cdf264e189R322, then I have to change this method to const, since they just return the value of member, so I think it makes sense to mark it to const.

source/server/listener_impl.cc
soulxu: yea, this is defined in the API, you can reference here

https://github.com/envoyproxy/envoy/blob/e5ef611dd9a0e840681cd6e99c6cffc97ecb47b5/api/envoy/config/listener/v3/listener.proto#L124-L128

source/server/listener_manager_impl.cc
soulxu: Actually, this code is moved from https://github.com/envoyproxy/envoy/pull/22005/files#diff-f7639f0659fc4c081810d0a87d2e9e1aa9dc1c1423ea2d683ae491c4479e95f1L402

But yes, you are right, it isn't good for the user in the log or exception. I think there are two ways to improve that:
1. just asking user should specify a name
2. we log both the name and addresses.

If we choose the second one, we can do that in the separated PR, since the whole listener is full of logs with only the name, those are out of scope for this PR.

daixiang0: Both ok to me, as a user I do not want a random uuid.

envoy/network/listener.h
daixiang0: Sound reasonable to me.

mattklein123 (changes requested):
Generally LGTM with a few comments, thank you.

/wait

source/server/listener_impl.cc
```diff
@@ -378,8 +377,12 @@ ListenerImpl::ListenerImpl(const envoy::config::listener::v3::Listener& config,
               cx_limit_runtime_key_, config_.name());
   }
 
+  filter_chain_manager_ = std::make_unique<FilterChainManagerImpl>(
+      addresses_, listener_factory_context_->parentFactoryContext(), initManager()),
```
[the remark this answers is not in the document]
soulxu: It is only for the parameter `addresses_`, since the addresses are filled up late, then I move the creation of filter chain manager late. Actually, the filter chain manager only uses the addresses for the log.

source/server/listener_impl.cc
```diff
@@ -1016,9 +1045,59 @@ bool ListenerImpl::getReusePortOrDefault(Server::Instance& server,
...
-  return *address() == *other.address() &&
-         Network::Utility::protobufAddressSocketType(config_.address()) ==
-             Network::Utility::protobufAddressSocketType(other.config_.address());
+  if ((socket_type_ != other.socket_type_) || (addresses_.size() != other.addresses().size())) {
+    return false;
+  }
+
+  // Since there can be multiple zero port addresses, create a copy
+  // of other's addresses, remove it for any finding to avoid matching same zero port
+  // address to the same one.
```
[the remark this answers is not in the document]
soulxu: Sorry, let me explain. We support listening on zero port for the test, so I need to support multiple addresses with zero port also. When we compare two listeners, we only need to ensure there is the same number of zero port addresses in the two listeners. So when find one zero port address matched, then I removed one from the copy addresses list.

source/server/listener_manager_impl.cc
soulxu: got it

mattklein123 (approved):
Nice, thanks!

PR #22013, opened 2022-07-04
docs: Improve file collection/handling in build

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -56,7 +56,7 @@ genrule(
         $(location //source/extensions:extensions_metadata.yaml) \\
         $(location //contrib:extensions_metadata.yaml) $@
     """,
-    tools = ["//tools/docs:generate_extensions_security_rst"],
+    exec_tools = ["//tools/docs:generate_extensions_security_rst"],
 )
 
 genrule(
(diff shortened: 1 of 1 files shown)


docs/BUILD
```diff
@@ -136,13 +136,24 @@ pkg_files(
...
+        "root/**/*.css",
+        "root/**/*.ico",
+        "root/**/*.js",
+        "root/**/*.json",
+        "root/**/*.pb",
+        "root/**/*.png",
+        "root/**/*.proto",
+        "root/**/*.rst",
+        "root/**/*.svg",
+        "root/**/*.yaml",
```
wrowe, line 149: Do we expect folks will miss this when adding other file types to the docs set? Should it be documented anywhere or is docs/BUILD self-explanatory?

wrowe (approved):
One nit to ask if the change should be documented, otherwise LGTM

phlax:
good question - there is not so much docs for writing docs atm, at least not in one place

in this case i think lets land - there are other similarly undocumented docs/build artefacts and i think probably it would be good to take a more comprehensive view when updating

im also thinking that i need to add _something_ about the ci/docs/format changes to changelog, but probably not specific to this change

unless you shout otherwise, ill land this in a bit @wrowe

PR #22015, opened 2022-07-04
protodoc: Use (more) templates for api build

phlax:
This shifts the rendering code from python into jinja templates.

The logic for parsing protodoc should not change much, but how we want to present it does so this separates the changeable code out, and should make it easier to update, it should also make the code easier to read/understand

This is a further step towards using templates (we use some already) and once it has landed we can look at moving further in this direction.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/tools/protodoc/generate_empty.py
+++ b/tools/protodoc/generate_empty.py
@@ -7,11 +7,11 @@ import pathlib
 import sys
 import tarfile
 
-import protodoc as protodoc
+import protodoc as _protodoc
 from tools.protodoc.jinja import env as jinja_env
 
 
-def generate_empty_extension_docs(extension, details, api_extensions_root):
+def generate_empty_extension_docs(protodoc, extension, details, api_extensions_root):
     extension_root = pathlib.Path(details['path'])
     path = pathlib.Path(api_extensions_root, extension_root, 'empty', extension_root.name + '.rst')
     path.parent.mkdir(parents=True, exist_ok=True)
@@ -19,13 +19,13 @@ def generate_empty_extension_docs(extension, details, api_extensions_root):
     reflink = ''
     if 'ref' in details:
         reflink = '%s %s.' % (
-            details['title'], protodoc.format_internal_link(
-                'configuration overview', details['ref']))
+            details['title'],
+            _protodoc.format_internal_link('configuration overview', details['ref']))
     content = jinja_env.get_template("empty.rst.tpl").render(
         header=details['title'],
         description=description,
         reflink=reflink,
-        extension=protodoc.format_extension(extension))
+        extension=protodoc._extension(extension))
     path.write_text(content)
 
 
@@ -36,8 +36,9 @@ def main():
     api_extensions_root = os.path.join(generated_rst_dir, "api-v3/config")
 
     empty_extensions = json.loads(pathlib.Path(empty_extensions_path).read_text())
+    protodoc = _protodoc.RstFormatVisitor()
     for extension, details in empty_extensions.items():
-        generate_empty_extension_docs(extension, details, api_extensions_root)
+        generate_empty_extension_docs(protodoc, extension, details, api_extensions_root)
 
     with tarfile.open(output_filename, "w:gz") as tar:
         tar.add(generated_rst_dir, arcname=".")
--- a/tools/protodoc/protodoc.py
+++ b/tools/protodoc/protodoc.py
@@ -4,35 +4,26 @@
 # https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html for Sphinx RST syntax.
 
 import logging
-import functools
 import sys
 from collections import defaultdict
-from functools import cached_property
+from functools import cached_property, lru_cache
+from typing import Dict, Iterable, Iterator, Set, Tuple
 
+from google.protobuf.descriptor_pb2 import FieldDescriptorProto as field_proto
 import yaml
-
-from envoy.code.check.checker import BackticksCheck
-
-from tools.api_proto_plugin import annotations, constants
-from tools.api_proto_plugin import plugin
-from tools.api_proto_plugin import visitor
-from tools.protodoc.data import data
-from tools.protodoc.jinja import env as jinja_env
 
 from udpa.annotations import security_pb2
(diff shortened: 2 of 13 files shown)


tools/protodoc/protodoc.py
```diff
@@ -55,23 +49,57 @@
...
+    field_proto.TYPE_BYTES: 'bytes'
+}
+
+PRETTY_LABEL_NAMES = {field_proto.LABEL_OPTIONAL: '', field_proto.LABEL_REPEATED: '**repeated** '}
+
+import contextlib
+
+
+@contextlib.contextmanager
+def debug(name, threshold=None):
```
phlax, line 90: remove

tools/protodoc/protodoc.py
```diff
@@ -55,23 +49,57 @@
...
+"""
+
 WIP_WARNING = (
     '.. warning::\n   This API feature is currently work-in-progress. API features marked as '
     'work-in-progress are not considered stable, are not covered by the :ref:`threat model '
     '<arch_overview_threat_model>`, are not supported by the security team, and are subject to '
     'breaking changes. Do not use this feature without understanding each of the previous '
     'points.\n\n')
 
+PRETTY_TYPE_NAMES = {
```
phlax, line 66: move to constants

tools/protodoc/protodoc.py
```diff
@@ -55,23 +49,57 @@
 # http://www.fileformat.info/info/unicode/char/2063/index.htm
 UNICODE_INVISIBLE_SEPARATOR = u'\u2063'
 
+CONTRIB_MSG = """
```
phlax, line 52: template ?

tools/protodoc/protodoc.py
```diff
@@ -414,293 +196,493 @@ def field_cross_ref_label(field_name):
...
+    def _warnings(self, file_proto):
+        _warnings = ((
+            file_proto.options.HasExtension(udpa_status_pb2.file_status)
+            and file_proto.options.Extensions[udpa_status_pb2.file_status].work_in_progress) or (
+                file_proto.options.HasExtension(xds_status_pb2.file_status)
+                and file_proto.options.Extensions[xds_status_pb2.file_status].work_in_progress))
+        return ([WIP_WARNING] if _warnings else [])
+
+
+class TraversalAdapter:
```
phlax, line 633: dont do this here

tools/api_proto_plugin/type_context.py
```diff
@@ -89,7 +87,8 @@ def leading_comment_path_lookup(self, path):
         """
         location = self.location_path_lookup(path)
         if location is not None:
-            return Comment(location.leading_comments, self.file_level_annotations)
+            comments = "\n".join(s[1:] for s in location.leading_comments.split("\n"))
+            return Comment(comments, self.file_level_annotations)
```
phlax, line 91: revert this for now as it breaks proto_format - probs need to shift this code

jmarantz:
Hi Ryan - do you want to fill in the PR description, or is the title pretty much all there is to say?

/wait

phlax:
description updated

tools/protodoc/generate_empty.py
```diff
@@ -7,25 +7,25 @@
 import sys
 import tarfile
 
-import protodoc as protodoc
+import protodoc as _protodoc
```
htuch, line 10: Why underscore? This feels like some internal name. Can we use something like `protodoc_import` or `pd` etc?

htuch (commented):
Thanks. Overall I like the direction and consider this a net improvement. This is an enormous PR though with a lot of changes to verify. What would make me a lot more comfortable doing some local code reviews focused on style etc. is if you could validate that the before/after docs are identical. 
/wait

phlax:
> I think it would be useful investing in some lightweight scripts to do this in any case, to allow for confidence in these complex protodoc changes; absent unit tests we need to use that for correctness criteria.

this is more complex than some lightweight scripts as the rst/html does change, whereas the rendered output should not change

for this reason ive written a screenshotting spider which you can see here https://github.com/phlax/envoy-docshots

essentially it does the following for 2 given hashes where the hashes correspond to the rendered docs in GCP and represent a before (ie current rendered output) and after (ie rendered output from a PR):

- scrapes all the API urls 
- takes a screenshot of each
- compares the screenshots before/after for each of the scraped pages

after running this i found a few bugs, and i have fixed them here

the rendered (html) docs that the screenshots were taken from can be seen here:

- https://storage.googleapis.com/envoy-pr/a4d6e1f/docs/api-v3/api.html
- https://storage.googleapis.com/envoy-pr/bf3659e/docs/api-v3/api.html

the results of running against these 2 branches are:

```console
TOTAL MATCHED: 132
TOTAL FAILED: 2
```

the remaining failures are due (afaict) to some API urls being put in a TOC outside of the API tree - which means its a gamble as to which TOC sphinx sees as the real one (we might want to fix this, but its outside of the scope of this PR)

The failing pages are:

- https://storage.googleapis.com/envoy-pr/a4d6e1f/docs/api-v3/service/auth/v3/external_auth.proto.html
- https://storage.googleapis.com/envoy-pr/bf3659e/docs/api-v3/service/auth/v3/external_auth.proto.html

and here

- https://storage.googleapis.com/envoy-pr/a4d6e1f/docs/api-v3/service/auth/v3/attribute_context.proto.html
- https://storage.googleapis.com/envoy-pr/bf3659e/docs/api-v3/service/auth/v3/attribute_context.proto.html

tools/protodoc/generate_empty.py
phlax: this _is_ a common thing to do - i understand and agree with your previous objection regarding local vars - but this is pretty idiomatic - ie using an underscore prefix to disambiguate a namespace clashing imported object

if its a blocker i can use `pd` - i have never seen an import disambiguated with `x_import` before so i would rather not do that

htuch:
Thanks for doing the diff work, I think this materially improves confidence in the correctness. Neat approach.

htuch (approved):
LGTM, thanks!

PR #22017, opened 2022-07-04
[WIP] bazel: Set RBE downloads to minimal

phlax:

Commit Message:
Additional Description:

Testing RBE locally and it can in many cases take longer downloading files than it would have done just running locally

~Partly i think (mis)use of `pkg_tars` is to blame, and there are defo some optis that we could add, but i also think we want this set to minimal if possible~ i wasnt using the CI config

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -27,6 +27,8 @@ stages:
     - script: ci/run_envoy_docker.sh 'ci/do_ci.sh format'
       workingDirectory: $(Build.SourcesDirectory)
       env:
+        ENVOY_RBE: "true"
+        BAZEL_BUILD_EXTRA_OPTIONS: "--config=remote-ci --jobs=$(RbeJobs)"
         ENVOY_DOCKER_BUILD_DIR: $(Build.StagingDirectory)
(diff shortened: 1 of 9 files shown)


phlax:
strange CI seems to not dl the pkg files at all - not sure why it is different locally

https://dev.azure.com/cncf/envoy/_build/results?buildId=112063&view=logs&j=d56d8cbf-a81f-59e6-8c88-0420601e1aa0&t=afdd832d-55f2-5116-86b3-7f79d03ddeb4&l=195

phlax:
> not sure why it is different locally

because im building docs and not running any tests that consume the docs output beforehand

phlax:
> because im building docs and not running any tests that consume the docs output beforehand

that doesnt seem to be it - docs fail in CI and dont run anything that consumes the docs beforehand and it is failing - locally never fails

phlax:
so it seems the reason i was downloading so much is that i didnt have the toplevel config activated

i now also understand that `download_minimal` means more like `download_nothing` - so if this is the default then build jobs that use targets locally need to add a flag

not sure exactly why but this still saves 25% from the job to build x86 - i guess this is more to do with `test` than `build`

PR #22024, opened 2022-07-05
redis_proxy: support multiple passwords

surki:
Commit Message: redis_proxy: support multiple passwords
Additional Description: When rotating passwords, we need to support multiple passwords for graceful deployment/rollout. This change adds support for multiple passwords in AUTH, for both old auth and new ACL based one.
Risk Level: Low
Testing: Integration and unit tests
Docs Changes: Added
Release Notes: Added
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

Signed-off-by: Suresh Kumar <sureshkumar.pp@gmail.com>
Signed-off-by: Suresh Kumar <suresh.ponnusamy@freshworks.com>

### DIFF
--- a/api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
+++ b/api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
@@ -23,7 +23,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // Redis Proxy :ref:`configuration overview <config_network_filters_redis_proxy>`.
 // [#extension: envoy.filters.network.redis_proxy]
 
-// [#next-free-field: 9]
+// [#next-free-field: 10]
 message RedisProxy {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.config.filter.network.redis_proxy.v2.RedisProxy";
@@ -252,14 +252,21 @@ message RedisProxy {
   PrefixRoutes prefix_routes = 5;
 
   // Authenticate Redis client connections locally by forcing downstream clients to issue a `Redis
-  // AUTH command <https://redis.io/commands/auth>`_ with this password before enabling any other
-  // command. If an AUTH command's password matches this password, an "OK" response will be returned
-  // to the client. If the AUTH command password does not match this password, then an "ERR invalid
-  // password" error will be returned. If any other command is received before AUTH when this
+  // AUTH command <https://redis.io/commands/auth>`_ with this password or with the one in
+  // :ref:`extra_downstream_auth_passwords<envoy_v3_api_field_extensions.filters.network.redis_proxy.v3.RedisProxy.extra_downstream_auth_passwords>`
+  // before enabling any other command. If an AUTH command's password matches one of these passwords,
+  // an "OK" response will be returned to the client. If the AUTH command password does not match this password,
+  // then an "ERR invalid password" error will be returned. If any other command is received before AUTH when this
   // password is set, then a "NOAUTH Authentication required." error response will be sent to the
   // client. If an AUTH command is received when the password is not set, then an "ERR Client sent
   // AUTH, but no password is set" error will be returned.
   config.core.v3.DataSource downstream_auth_password = 6 [(udpa.annotations.sensitive) = true];
+
+  // Additional passwords to use when authenticating downstream clients locally. Please see
+  // :ref:`downstream_auth_password<envoy_v3_api_field_extensions.filters.network.redis_proxy.v3.RedisProxy.downstream_auth_password>`
+  // details.
+  repeated config.core.v3.DataSource extra_downstream_auth_passwords = 9
+      [(udpa.annotations.sensitive) = true];
 
   // List of faults to inject. Faults currently come in two flavors:
   // - Delay, which delays a request.
--- a/api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
+++ b/api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
@@ -7,6 +7,7 @@ import "envoy/config/core/v3/base.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
+import "envoy/annotations/deprecation.proto";
 import "udpa/annotations/migrate.proto";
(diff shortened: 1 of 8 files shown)


surki:
Question: wondering if I should mark the `downstream_auth_password` field deprecated and just add `downstream_auth_passwords` (notice the plural) rather than maintaining two fields?

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
```diff
@@ -252,15 +252,22 @@ message RedisProxy {
...
-  // password" error will be returned. If any other command is received before AUTH when this
+  // AUTH command <https://redis.io/commands/auth>`_ with this password or with the one in
+  // :ref:`extra_downstream_auth_passwords<envoy_v3_api_field_extensions.filters.network.redis_proxy.v3.RedisProxy.extra_downstream_auth_passwords>`
+  // before enabling any other command. If an AUTH command's password matches one of these passwords,
+  // an "OK" response will be returned to the client. If the AUTH command password does not match this password,
+  // then an "ERR invalid password" error will be returned. If any other command is received before AUTH when this
   // password is set, then a "NOAUTH Authentication required." error response will be sent to the
   // client. If an AUTH command is received when the password is not set, then an "ERR Client sent
   // AUTH, but no password is set" error will be returned.
   config.core.v3.DataSource downstream_auth_password = 6 [(udpa.annotations.sensitive) = true];
```
mattklein123, line 263: Let's go ahead and deprecate this field (and mark it as such in the release notes).

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
```diff
@@ -252,15 +252,22 @@ message RedisProxy {
...
 
+  // Additional passwords to use when authenticating downstream clients locally. Please see
+  // :ref:`downstream_auth_password<envoy_v3_api_field_extensions.filters.network.redis_proxy.v3.RedisProxy.downstream_auth_password>`
+  // details.
+  repeated config.core.v3.DataSource extra_downstream_auth_passwords = 9
```
mattklein123, line 268: Instead I would call this `downstream_auth_passwords` (plural) to take over for the now deprecated field.

changelogs/current.yaml
```diff
@@ -229,6 +229,9 @@ new_features:
 - area: listener
   change: |
     added dynamic listener filter configuration for listener filters. :ref:`dynamic listener filter re-configuration<envoy_v3_api_field_config.listener.v3.ListenerFilter.config_discovery>`. This dynamic listener filter configuration is only supported in TCP listeners.
+- area: redis_proxy
+  change: |
+    added support for multiple passwords to the redis proxy.
```
mattklein123, line 234: Please ref link to the new field, and per the other comment make sure you document the deprecated field.

source/extensions/filters/network/redis_proxy/proxy_filter.cc
```diff
@@ -24,9 +24,25 @@ ProxyFilterConfig::ProxyFilterConfig(
...
+  if (!downstream_auth_password.empty()) {
+    downstream_auth_passwords_.emplace_back(downstream_auth_password);
+  }
+
+  if (config.extra_downstream_auth_passwords_size() > 0) {
+    downstream_auth_passwords_.reserve(downstream_auth_passwords_.size() +
+                                       config.extra_downstream_auth_passwords().size());
+    for (const auto& source : config.extra_downstream_auth_passwords()) {
+      const auto p = Config::DataSource::read(source, true, api);
+      if (!p.empty()) {
```
mattklein123, line 40: Why does this version have the empty check? Seems like it should be the same as above?

mattklein123 (changes requested):
Thanks makes sense to me at a high level. A few comments to get started. Thank you!

/wait

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
surki: Done

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
surki: Done

changelogs/current.yaml
surki: Done

source/extensions/filters/network/redis_proxy/proxy_filter.cc
surki: We were having password empty check (which was on `std::string`) and fail AUTH. To maintain the same behavior, we need to have empty `std:vector` if the provided password is empty. We could probably ask the `Config::Datasource:read` to not allow empty string if we want to avoid empty check.

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
```diff
@@ -252,20 +253,32 @@ message RedisProxy {
...
+  // password" error will be returned. If any other command is received before AUTH when this
   // password is set, then a "NOAUTH Authentication required." error response will be sent to the
   // client. If an AUTH command is received when the password is not set, then an "ERR Client sent
   // AUTH, but no password is set" error will be returned.
-  config.core.v3.DataSource downstream_auth_password = 6 [(udpa.annotations.sensitive) = true];
+  //
+  // .. attention::
+  //   This field is deprecated. Use :ref:`downstream_auth_passwords
+  //   <envoy_v3_api_field_extensions.filters.network.redis_proxy.v3.RedisProxy.downstream_auth_passwords>`.
+  config.core.v3.DataSource downstream_auth_password = 6 [
```
mattklein123, line 267: You should have some test that covers use of this old field. See DEPRECATED_FEATURE_TEST.

mattklein123 (changes requested):
Thanks LGTM with small test comment.

/wait

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
surki: Done

PR #22026, opened 2022-07-05
docs: Fix backticks in api protos

phlax:
Signed-off-by: Albere <alberenorthey@gmail.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/type/matcher/regex.proto
+++ b/api/envoy/type/matcher/regex.proto
@@ -21,11 +21,11 @@ message RegexMatcher {
   // the documented `syntax <https://github.com/google/re2/wiki/Syntax>`_. The engine is designed
   // to complete execution in linear time as well as limit the amount of memory used.
   //
-  // Envoy supports program size checking via runtime. The runtime keys `re2.max_program_size.error_level`
-  // and `re2.max_program_size.warn_level` can be set to integers as the maximum program size or
+  // Envoy supports program size checking via runtime. The runtime keys ``re2.max_program_size.error_level``
+  // and ``re2.max_program_size.warn_level`` can be set to integers as the maximum program size or
   // complexity that a compiled regex can have before an exception is thrown or a warning is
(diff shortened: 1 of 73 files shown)


api/envoy/type/matcher/regex.proto
```diff
@@ -21,11 +21,11 @@ message RegexMatcher {
   // the documented `syntax <https://github.com/google/re2/wiki/Syntax>`_. The engine is designed
   // to complete execution in linear time as well as limit the amount of memory used.
   //
-  // Envoy supports program size checking via runtime. The runtime keys `re2.max_program_size.error_level`
-  // and `re2.max_program_size.warn_level` can be set to integers as the maximum program size or
+  // Envoy supports program size checking via runtime. The runtime keys ``re2.max_program_size.error_level``
+  // and ``re2.max_program_size.warn_level`` can be set to integers as the maximum program size or
```
phlax, line 25: not sure if these are necessary to fix v3 protos - i think the v3 protos are getting this from xds

phlax:
cc @albere

api/envoy/type/matcher/regex.proto
htuch: Yes, these are v2 protos.

htuch (approved):

htuch:
@phlax looking at a reasonable samples this LGTM. Do you have some systematic way to avoid us getting into this pitfall? Can we ban single ticks entirely and use other ways to achieve the same format effect?

phlax:
> Do you have some systematic way to avoid us getting into this pitfall? Can we ban single ticks entirely and use other ways to achieve the same format effect?

yes, and yes

we currently have the `Bad RST` warning in `protodoc` the intention is to turn that to error (perhaps with a `--continue` option)

PR #22027, opened 2022-07-05
ci: propagate NUM_CPUS to bazel

r-bk:
Commit Message: ci: propagate NUM_CPUS to bazel
Additional Description:
This commit ensures that NUM_CPUS env var is correctly propagated to
bazel. Without this change, if NUM_CPUS is set, for example, to a lower
than the actual CPU count number, bazel still consumes all CPUs
available on a host.
Risk Level: low
Testing: PR github workflow
Docs Changes: no
Release Notes: no

### DIFF
--- a/ci/build_setup.sh
+++ b/ci/build_setup.sh
@@ -100,6 +100,7 @@ BAZEL_BUILD_OPTIONS=(
   "--repository_cache=${BUILD_DIR}/repository_cache"
   "--experimental_repository_cache_hardlinks"
   "--action_env=CLANG_FORMAT"
+  "--jobs=${NUM_CPUS}"
   "${BAZEL_BUILD_EXTRA_OPTIONS[@]}"
   "${BAZEL_EXTRA_TEST_OPTIONS[@]}")
 


ci/build_setup.sh
```diff
@@ -100,6 +100,7 @@ BAZEL_BUILD_OPTIONS=(
   "--repository_cache=${BUILD_DIR}/repository_cache"
   "--experimental_repository_cache_hardlinks"
   "--action_env=CLANG_FORMAT"
+  "--jobs=${NUM_CPUS}"
```
phlax, line 103: i think this is already set in `ci/setup_cache.sh` -  it is set to `--jobs=HOST_CPUS*.9`

r-bk: @phlax Thanks a lot for the review!

`--jobs=HOST_CPUS*.9` is set only in case `BAZEL_REMOTE_CACHE` is set too.
I am building locally without remote cache. Is this scenario rare and I am optimizing a non-interesting use case?

As to the PR itself, I can refactor such that both cases are handled similar to the way it is handled in `setup_cache.sh`, by introducing a new env var `NUM_JOBS` with default value`HOST_CPUS*.9` (for example). This way it will be possible to override, and will be handled in a unified way with or without remote cache. Does it sound good?

r-bk: I am running it like this `./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.dev'`.
This takes 8/8 cores. I would like to limit the whole thing to 2 cores at most.

r-bk: I am running it like this `./ci/run_envoy_docker.sh './ci/do_ci.sh bazel.dev'`.
This takes 8/8 cores. I would like to limit the whole thing to 2 cores at most.

r-bk: @phlax OK. Thanks!
Dropping the PR ..

PR #22031, opened 2022-07-06
Example: Add a new Kafka Broker filter

nguyenptk:
Signed-off-by: Nguyen Phan <nguyenptk@gmail.com>

Commit Message: Add a new example for the Kafka Broker filter
Additional Description: 
Risk Level: Low
Testing: README.md guideline
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+export NAME=kafka-broker
+
+# shellcheck source=examples/verify-common.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../verify-common.sh"
+
+run_log "Waiting for Zookeeper and Kafka started."
+sleep 40
```
daixiang0, line 9: Long time to wait, could we use a loop to check them are ready?

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,17 @@
...
+
+TOPIC="envoy.kafka.broker"
+run_log "Creating topic."
+docker exec kafka kafka-topics --bootstrap-server localhost:19092 --create --topic $TOPIC quickstart-events
+sleep 5
+
+run_log "Producer is sending messages. Please type..."
+docker exec -it kafka kafka-console-producer --request-required-acks 1 --broker-list localhost:19092 --topic $TOPIC
```
daixiang0, line 17: Does it need to type by hand? Since the verify.sh will run in the CI, should better use auto way to send and receive.

examples/kafka-broker/verify.sh
nguyenptk: updated, please help review again

examples/kafka-broker/verify.sh
nguyenptk: updated, please help review again

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash -e
+
+export NAME=kafka-broker
+
+# shellcheck source=examples/verify-common.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../verify-common.sh"
+
+run_log "Waiting for Kafka on 9092."
+while true; do if docker-compose logs -f kafka | grep -q 'started (kafka.server.KafkaServer)'; then
```
daixiang0, line 9: It may be failed in CI, could you replace with:
```
while true; do
   if xxx
   else
   fi
done

File: examples/kafka-broker/verify.sh [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+else
+    echo waiting ...; sleep 1;
+fi; done
+run_log "Kafka launched."
+
+# Topic
+TOPIC="envoy.kafka.broker"
+run_log "Creating topic."
+docker-compose exec kafka kafka-topics --bootstrap-server localhost:19092 --create --topic $TOPIC quickstart-events
+sleep 5
```
daixiang0, line 20: Still need a loop for this check.

examples/kafka-broker/verify.sh
nguyenptk: Updated, please help verify again

examples/kafka-broker/verify.sh
nguyenptk: Thank you for your point. I tested it many times and we don't need it anymore.

nguyenptk:
```
unknown type: envoy.extensions.filters.network.kafka_broker.v3.KafkaBroker for type Any
```
https://dev.azure.com/cncf/envoy/_build/results?buildId=112228&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec
Does the `envoyproxy/envoy-build-ubuntu` in the pipeline support `KafkaBroker`?

My local machine works well with `envoyproxy/envoy-contrib-dev:latest`

@phlax Do you have any comments?

phlax:
>  Does the envoyproxy/envoy-build-ubuntu in the pipeline support KafkaBroker?

its not really related to the build image - more the release binary that is being tested - i think that we are not testing against the binary with contribs included

if you check `examples/BUILD` there are listed exclusions there, you can add kafka

nguyenptk:
> if you check `examples/BUILD` there are listed exclusions there, you can add kafka

Hi @phlax, the pipeline is good now, please help me take a look at the MR. 
Thank you

phlax:
@nguyenptk apologies for slow review - as mentioned offline i have been busy with release so will review once that is out of the way

/wait-any

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,61 @@
...
+    docker-compose exec -T kafka kafka-console-consumer --bootstrap-server localhost:19092 --topic $TOPIC --from-beginning >> consumer.txt
+}
+
+check_message() {
+    while true; do
+    if [[ $(< consumer.txt) == "$MESSAGE" ]]; then
+        run_log "Received message succesfully"
+        rm consumer.txt # clean up consumer.txt
+        run_log "Bring down the kafka"
+        docker-compose stop kafka
```
phlax, line 48: not sure why this is here but its not the responsibility of this script to bring down the stack

examples/kafka-broker/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+    container_name: zookeeper
+    networks:
+    - envoymesh
+    ports:
+    - 22181:2181
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
+
+networks:
```
phlax, line 48: you can remove all of this network stuff, they are all on the same network so can use the default

examples/kafka-broker/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+      KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+
+  zookeeper:
+    image: confluentinc/cp-zookeeper:latest
+    container_name: zookeeper
+    networks:
+    - envoymesh
+    ports:
+    - 22181:2181
```
phlax, line 43: remove these as they are not used

examples/kafka-broker/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - 19092:19092
+    - 8001:8001
+
+  kafka:
+    image: confluentinc/cp-kafka:latest
+    container_name: kafka
+    networks:
+    - envoymesh
+    ports:
+    - 127.0.0.1:9092:9092
```
phlax, line 21: remove this

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,61 @@
...
+    else
+        run_log "Checking message"
+        sleep 1
+    fi;
+    done
+}
+
+read_message &
+check_message &
+wait
```
phlax, line 61: its not clear what this is doing - please move any checks/test into a `run_log` section with an explicit test and dont background things unless necessary

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,61 @@
...
+run_log "Kafka launched"
+
+# Topic
+TOPIC="envoy.kafka.broker"
+run_log "Creating topic"
+docker-compose exec -T kafka kafka-topics --bootstrap-server localhost:19092 --create --topic $TOPIC quickstart-events
+
+run_log "Checking topic"
+topic_created=$(docker-compose exec -T kafka kafka-topics --bootstrap-server localhost:19092 --list)
+if [[ "$topic_created" == "$TOPIC" ]]; then
```
phlax, line 19: how could this every be in more than one state?

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash -e
+
+export NAME=kafka-broker
+
+# shellcheck source=examples/verify-common.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../verify-common.sh"
+
+run_log "Waiting for Kafka on port 9092"
+wait_for 120 sh -c "docker-compose logs -f kafka | grep -q 'started (kafka.server.KafkaServer)'"
```
phlax, line 9: rather than doing this can you add healthchecks and waits in the docker-compose file

phlax:
@nguyenptk overall we need to simplify and cleanup the verify script

As discussed offline im happy to add docs for this but it needs to be super-clear step-by-step what is going on for me to do it

phlax:
perhaps you could tell me what the "steps" are that you expect in the documentation, i think that will make it easier to clariify what the tests should be in the verify script

phlax:
> here is the "steps"

couple of points

- from an end users perspective it is not related to verifying - the verification is that the docs do what they say
- starting/stopping the containers is not the responsibility of the verify script

examples/kafka-broker/verify.sh
nguyenptk: I have stuck when I want to exit 
```
docker-compose exec -T kafka kafka-console-consumer --bootstrap-server localhost:19092 --topic $TOPIC --from-beginning >> consumer.txt
```
so I use this way. Do you have any suggestions for it?

examples/kafka-broker/verify.sh
nguyenptk: I only found this way to make sure Kafka launched successfully. Because it takes some seconds to connect to the ZooKeeper and sometimes it gets failed when starting.

examples/kafka-broker/verify.sh
phlax: not sure but normally exec will run a command and exit - if that is not happening then there is something wrong going on

what does the `--bootstrap-server` flag do ?

im guessing whatever is going on here should have happened in the composition

nguyenptk: `kafka-console-consumer` is the listener so it should be there forever to receive new messages. Let me find another way to exit the command.
`--bootstrap-server` is the server of Kafka broker and `localhost:19092` is the host and port of Envoy.

examples/kafka-broker/verify.sh
nguyenptk: Just in case, we cannot create the `TOPIC` successfully, and `verify.sh` will exit with `exit 1`.

phlax: there should be no just in case - it needs to work

nguyenptk: Because from my testing result, rarely we get the timeout error when creating a topic. So I add this logic to make sure the TOPIC was created before we continue to do something else.

examples/kafka-broker/docker-compose.yaml
nguyenptk: Updated, please help verify again

examples/kafka-broker/docker-compose.yaml
nguyenptk: Updated, please help verify again

examples/kafka-broker/docker-compose.yaml
nguyenptk: Updated, please help verify again

examples/kafka-broker/verify.sh
nguyenptk: Updated, please help verify again

examples/kafka-broker/docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - kafka
+    ports:
+    - 19092:19092
+    - 8001:8001
+
+  kafka:
+    image: confluentinc/cp-kafka:latest
+    container_name: kafka
+    depends_on:
+    - zookeeper
```
phlax, line 19: i think add a `condition` here

examples/kafka-broker/verify.sh
phlax: reading kafka docs - it seems like the `--boostrap-server` is probably going to be needed - we can figure out further

examples/kafka-broker/verify.sh
phlax: either way it needs to work reliably - otherwise it will create flakes in CI

examples/kafka-broker/docker-compose.yaml
nguyenptk: Updated

examples/kafka-broker/docker-compose.yaml
```diff
@@ -6,8 +6,6 @@ services:
       context: .
       dockerfile: Dockerfile-proxy
     container_name: proxy
-    depends_on:
-    - kafka
```
phlax: i think we do want this - and we want to make it wait for kafka to be healthy

nguyenptk: I think the Envoy can be started first and the cluster's health check interval can help. Correct me if I'm wrong

phlax: i think best to make envoy start last and only after the others have come up

that way you can be pretty confident that when you call it via the proxy it will respond

phlax:
hi @adamkotwasinski while working on this PR we noticed that when you bring up the docker composition it churns out endless logs with:

```console
proxy           | [2022-08-16 08:18:06.482][1][debug][dns] [source/extensions/network/dns_resolver/cares/dns_impl.cc:354] dns resolution for 127.0.0.1 started
proxy           | [2022-08-16 08:18:06.631][1][debug][dns] [source/extensions/network/dns_resolver/cares/dns_impl.cc:275] dns resolution for 127.0.0.1 completed with status 0
proxy           | [2022-08-16 08:18:06.631][1][debug][upstream] [source/common/upstream/upstream_impl.cc:275] transport socket match, socket default selected for host with address 127.0.0.1:9092
proxy           | [2022-08-16 08:18:06.631][1][debug][upstream] [source/common/upstream/strict_dns_cluster.cc:180] DNS refresh rate reset for 127.0.0.1, refresh rate 5000 ms
proxy           | [2022-08-16 08:18:10.487][1][debug][main] [source/server/server.cc:251] flushing stats
zookeeper       | [2022-08-16 08:18:11,622] WARN Unexpected exception (org.apache.zookeeper.server.NIOServerCnxn)
zookeeper       | EndOfStreamException: Unable to read additional data from client, it probably closed the socket: address = /127.0.0.1:60574, session = 0x0
zookeeper       | 	at org.apache.zookeeper.server.NIOServerCnxn.handleFailedRead(NIOServerCnxn.java:163)
zookeeper       | 	at org.apache.zookeeper.server.NIOServerCnxn.doIO(NIOServerCnxn.java:326)
zookeeper       | 	at org.apache.zookeeper.server.NIOServerCnxnFactory$IOWorkRequest.doWork(NIOServerCnxnFactory.java:522)
zookeeper       | 	at org.apache.zookeeper.server.WorkerService$ScheduledWorkRequest.run(WorkerService.java:154)
zookeeper       | 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
zookeeper       | 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
zookeeper       | 	at java.base/java.lang.Thread.run(Thread.java:829)
```

tracking the zookeeper issue it seems like despite warning about an `Unexpected exception` (not sure if that should be a warning or unexpected etc) that this is "normal" behaviour when a client disconnects

apart from the issue of log hogging and make any real errors hard to track it would fill up a disk/logrotate pretty quickly
 
it seems as though the error/warning is triggered in response to Envoy `flushing stats`

so, we are wondering if there is a way to prevent this either by Envoy not flushing stats (so frequently) or suppressing this warning/error - and i guess whether there is something incorrect in either the config here or Envoy's interaction with kafka and/or zookeeper

adamkotwasinski:
Hey, I'm not too experienced in Kafka's usage of Zookeeper, but it should be a fairly simple one (it only keeps some topology metadata). However I think I've seen something similar happening in the past in case of network failure - so maybe that's a docker thing? (you never know)

I'm on holidays until end-of-week, but if I were to debug, I'd check the behaviour _without_ Envoy in the stack just to remove the potential culprit (so only `9092` instead of `10000`).

The other ideas that come to my mind would be:
* changing logging of `org.apache.zookeeper.server.NIOServerCnxn` in zk's log4j.properties
* (exploratory) getting rid of ZK in the stack and adopting KRaft

phlax:
so we hit a bigger problem with this sandbox - we realized that it wasnt actually using the proxy

probably best to wait until you get back - but could do with a bit of input into what exactly should go through the proxy

the tests/examples we were adding were

- add/check a topic
- send a message
- receive the message

when updating the config to actually use the proxy it kinda falls apart - so not sure how it should work when containerized - ie not using localhost

examples/kafka-broker/docker-compose.yaml
```diff
@@ -0,0 +1,59 @@
...
+      retries: 120
+    environment:
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      # Listener value needs to be equal to cluster value in Envoy config
+      # (will receive payloads from Envoy).
+      KAFKA_LISTENER: PLAINTEXT://kafka-server:9092
+      # Advertised listener value needs to be equal to Envoy's listener
+      # (will make clients discovering this broker talk to it through Envoy).
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka-server:10000
```
adamkotwasinski, line 43: I'm afraid this might be "proxy:${port}", not "kafka-server", because we want clients to talk to proxy, the proxy is going to route client traffic to kafka-server

examples/kafka-broker/verify.sh
```diff
@@ -0,0 +1,30 @@
...
+kafka_client () {
+    docker-compose run kafka-client "$@"
+}
+
+TOPIC="envoy-kafka-broker"
+
+MESSAGE="Welcome to Envoy and Kafka Broker filter!"
+
+run_log "Create a Kafka topic"
+kafka_client kafka-topics --bootstrap-server proxy:10000 --create --topic $TOPIC quickstart-events
```
adamkotwasinski, line 19: I might be out of date with this, but shouldn't it look like?
```
kafka_client kafka-topics --bootstrap-server proxy:10000 --create --topic $TOPIC --partitions 13 --replication-factor 1
```

adamkotwasinski:
Apart from the two comments above (first one is important) everything in this PR feels okay - it is okay to use Envoy for every Kafka-related operation (just like in the commands to create-topic / list-topics / send / receive message).

There's some small manual for configuring Envoy/Kafka - https://adam-kotwasinski.medium.com/deploying-envoy-and-kafka-8aa7513ec0a0 (esp. first part - `Routing all traffic through Envoy (including replication)`)

I do not think that hostname change from localhost to something else is going to hurt, as the docker compose AFAIU should make everything resolve-able.

If that doesn't help, I'll try to take a look next week when I'm back!

examples/kafka-broker/docker-compose.yaml
phlax: yep, that is where we hit problems

if we set it to `proxy:1000` then it seems to go into a loop where kafka tries to check the proxy, but then the proxy fails because it calls kafka and finds it unhealthy

phlax: even removing the healthcheck, and bringing it up manually hit the same problem

examples/kafka-broker/verify.sh
nguyenptk: for the `--replication-factor 1`, we have `KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1` in docker-compose instead.

nguyenptk:
I think we solved this problem here with
```
KAFKA_LISTENER: INTERNAL://kafka-server:9092,EXTERNAL://proxy:10000
# Advertised listener value needs to be equal to Envoy's listener
# (will make clients discovering this broker talk to it through Envoy).
KAFKA_ADVERTISED_LISTENERS: INTERNAL://kafka-server:9092,EXTERNAL://proxy:10000
KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT
KAFKA_INTER_BROKER_LISTENER_NAME: INTERNAL
KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
```

nguyenptk:
I updated the health check script of the Zookeeper and the error of
```
EndOfStreamException: Unable to read additional data from client, it probably closed the socket:
```
was gone.

examples/kafka-broker/docker-compose.yaml
phlax: seems like @nguyenptk figured out the problem (both the main kafka server and the broker need to be listed as listners) - really, really nice work @nguyenptk !!!

magically it also seems to have negated the need for the healthcheck - no warnings seem to be emitted (we might need to revisit this if we find it flakes in CI, but for now looking good)

nguyenptk:
hi @adamkotwasinski, I have a question about the `kafka.kafka_broker.request.create_topics_request` stats. I cannot receive the `kafka.kafka_broker.request.create_topics_request: 1` on my local machine (I saw it for the first time when I grepped the stats but now). Currently, I only have
```
kafka.kafka_broker.request.create_topics_request: 0
```
Could you point me more information about that?

**Note:** other stats look good. For examples:
```
kafka.kafka_broker.request.api_versions_request: 4
kafka.kafka_broker.request.find_coordinator_request: 1
kafka.kafka_broker.request.metadata_request: 4
kafka.kafka_broker.response.api_versions_response: 4
kafka.kafka_broker.response.find_coordinator_response: 1
kafka.kafka_broker.response.metadata_response: 4
cluster.kafka_service.max_host_weight: 1
cluster.kafka_service.membership_healthy: 1
cluster.kafka_service.membership_total: 1
```

adamkotwasinski:
This metric is increased by kafka-topics --create invocations. If you point
the script at the proxy it should go through envoy and increase the metric.
You can try that with an already-setup demo.

niedz., 28 sie 2022, 09:38 użytkownik Nguyên ***@***.***>
napisał:

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,110 @@
...
+
+ Now you can receive the message using the ``kafka-console-consumer`` :
+
+.. code-block:: console
+
+  $ docker-compose run --rm kafka-client kafka-console-consumer --bootstrap-server proxy:10000 --topic $TOPIC --from-beginning --max-messages 1 | grep "$MESSAGE"
+
+
+Step 6: Check admin ``kafka_broker`` stats
+**************************************
```
phlax, line 80: this underline needs to be longer to match the title

examples/kafka/verify.sh
```diff
@@ -0,0 +1,58 @@
...
+kafka_client /bin/bash -c " \
+    echo $MESSAGE \
+    | kafka-console-producer --request-required-acks 1 --broker-list proxy:10000 --topic $TOPIC"
+
+run_log "Receive a message using the Kafka consumer"
+kafka_client kafka-console-consumer --bootstrap-server proxy:10000 --topic $TOPIC --from-beginning --max-messages 1 | grep "$MESSAGE"
+
+run_log "Check admin kafka_broker stats"
+EXPECTED_BROKER_STATS=(
+    # "kafka.kafka_broker.request.create_topics_request: 1"
```
phlax, line 34: cc @adamkotwasinski this one doesnt seem to work as expected (returns 0) so is commented out for now

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,110 @@
...
+  -----------------------------------------------------------------------------------------------------------------------
+  kafka_kafka-server_1   /etc/confluent/docker/run      Up             9092/tcp
+  kafka_proxy_1          /docker-entrypoint.sh /usr ... Up             0.0.0.0:10000->10000/tcp, 0.0.0.0:8001->8001/tcp
+  kafka_zookeeper_1      /etc/confluent/docker/run      Up (healthy)   2181/tcp, 2888/tcp, 3888/tcp
+
+
+Step 2: Create a Kafka topic
+****************************
+
+Let's start by creating a Kafka topic called ``envoy-kafka-broker``:
```
phlax, line 37:
```suggestion
Start by creating a Kafka topic with the name ``envoy-kafka-broker``:
```

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,110 @@
...
+
+  $ export TOPIC="envoy-kafka-broker"
+  $ docker-compose run --rm kafka-client --bootstrap-server proxy:10000 --create --topic $TOPIC
+
+
+Step 3: Check the Kafka topic
+*****************************
+
+You can view the topics that Kafka is aware of with the ``--list`` argument.
+Let's check that the topic you created exists:
```
phlax, line 49:
```suggestion
You can view the topics that Kafka is aware of with the ``kafka-topics --list`` argument.

Check that the topic you created exists:
```

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,110 @@
...
+
+.. code-block:: console
+
+  $ docker-compose run --rm kafka-client kafka-topics --bootstrap-server proxy:10000 --list | grep $TOPIC
+
+
+Step 4: Send a message using the Kafka producer
+***********************************************
+
+ Next, let's send a message using the topic you have created using the ``kafka-console-producer``:
```
phlax, line 59:
```suggestion
 Next, send a message for the topic you have created using the ``kafka-console-producer``:
```

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,110 @@
...
+ Now you can receive the message using the ``kafka-console-consumer`` :
+
+.. code-block:: console
+
+  $ docker-compose run --rm kafka-client kafka-console-consumer --bootstrap-server proxy:10000 --topic $TOPIC --from-beginning --max-messages 1 | grep "$MESSAGE"
+
+
+Step 6: Check admin ``kafka_broker`` stats
+******************************************
+
```
phlax, line 81:
```suggestion

When you proxy to the Kafka broker, Envoy records various stats.

You can check the broker stats by querying the Envoy admin interface:

```

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,110 @@
...
+  kafka.kafka_broker.request.find_coordinator_request: 1
+  kafka.kafka_broker.request.metadata_request: 4
+  kafka.kafka_broker.response.api_versions_response: 4
+  kafka.kafka_broker.response.find_coordinator_response: 1
+  kafka.kafka_broker.response.metadata_response: 4
+
+
+Step 7: Check admin ``kafka_service`` stats
+*******************************************
+
```
phlax, line 96:
```suggestion

Envoy also records stats fot the Kafka service:

```

phlax (commented):
Amazing work @nguyenptk - thanks so much for persevering with this.

Copy-edit suggestions inline

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,117 @@
...
+******************************************
+
+When you proxy to the Kafka broker, Envoy records various stats.
+
+You can check the broker stats by querying the Envoy admin interface:
+
+.. code-block:: console
+
+  $ curl -s "http://localhost:8001/stats?filter=kafka.kafka_broker" | grep -v ": 0"
+  kafka.kafka_broker.request.create_topics_request: 1
```
phlax, line 90: lets remove this unless/until we get it to work

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,117 @@
...
+  kafka.kafka_broker.request.api_versions_request: 4
+  kafka.kafka_broker.request.find_coordinator_request: 1
+  kafka.kafka_broker.request.metadata_request: 4
+  kafka.kafka_broker.response.api_versions_response: 4
+  kafka.kafka_broker.response.find_coordinator_response: 1
+  kafka.kafka_broker.response.metadata_response: 4
+
+
+Step 7: Check admin ``kafka_service`` stats
```
phlax, line 99:
```suggestion
Step 7: Check admin ``kafka_service`` cluster stats
```

you will need to adjust underline also

docs/root/start/sandboxes/kafka.rst
```diff
@@ -0,0 +1,117 @@
...
+*******************************************
+
+Envoy also records stats fot the Kafka service:
```
phlax, line 102:
```suggestion
Envoy also records cluster stats fot the Kafka service:
```

phlax (approved):
lgtm, thanks again @nguyenptk for your hard work on this

PR #22047, opened 2022-07-07
OpenTelemetry Tracer: only export OTLP trace span when sampled

AlexanderEllis:
Commit Message: Export OTLP Trace span only when sampled
Additional Description: Envoy's health checks rely on setting the `sampled` attributed to false for the span for the health check. This should prevent the spurious spans from being created, as we generally don't want to trace every time we health check a cluster.  Before this change, this Tracer just exported them, and it now will check the sampled flag before exporting each span. See https://github.com/envoyproxy/envoy/pull/21893#issuecomment-1167450850 for more context.
Risk Level: Low
Testing: Added a unit test
Part of https://github.com/envoyproxy/envoy/issues/9958


wbpcode (approved):
LGTM. Thanks.

Issue #19526, opened 2022-01-13
slow start mode formula result very big edf dealine_ cause slow start invalid


jiangshantao-dbg:
in slow start window, endpoints need slow start receives no traffic and beyond the slow start window, they still receive no traffic. because endpoints which need slow start will have a very high `deadline_` in edf scheduler.
```
(gdb) ppqueue queue_
$25 = {
  deadline_ = 300000,
  order_offset_ = 1,
  entry_ = {
    __ptr_ = 0x555562060158,
    __cntrl_ = 0x55556205b260
  }
}
$26 = {
  deadline_ = 2,
  order_offset_ = 2,
  entry_ = {
    __ptr_ = 0x555561eed7d8,
    __cntrl_ = 0x555561f4c820
  }
}
Priority queue size = 2
Priority queue capacity = 2
``` 
if host_weight=1, then `weight=host_weight * applyAggressionFactor(time_factor)` will be less than 1, but `deadline_ = current_time_+1/weight` will be too big.
and the endpoints will receive no traffic because it will never in top of the queue_ utils the normal endpoints deadline_ catch up it. this cause beyond the slow start window the endpoints still receive no traffic.

3. istio default set the host weight to 1. so don't work in istio. and may cause bug when deployment rolling update, just old endpoints receive traffic may cause these pods load too high.

> @jiangshantao-dbg @du2016 thanks for reporting the issue, this is unfortunate that the bug has not been caught in tests. I've submitted [patch](https://github.com/envoyproxy/envoy/pull/19518) for fixing slow start config propagation in Round Robin Lb. Would suggest to open a dedicated issue for revisiting the slow start formula. cc @tonya11en @wbpcode

how about change slow start formula to just a static percent. just use `aggression` as `new_weight=weight*aggression`. default aggression to 0.5, which means slow start endpoint just receive 50% of traffic to the normal one.

use formula can't avoid a too small new_weight which will cause unexpected situation.

_Originally posted by @jiangshantao-dbg in https://github.com/envoyproxy/envoy/issues/19437#issuecomment-1011685017_

wbpcode:
cc @nezdolik may be we can create a configurable min factor to compute new weight just as my previous comment?

tonya11en:
The purpose of the aggression term is to control the curve of the ramp-up. Changing the formula to be linear would defeat the purpose. 

[...]

PR #13176, opened 2020-09-18
Support slow Start mode in Envoy

nezdolik:
Signed-off-by: Kateryna Nezdolii <nezdolik@spotify.com>

Support progressive traffic increase in Envoy, implementation is according to design doc: https://docs.google.com/document/d/1NiG1X0gbfFChjl1aL-EE1hdfYxKErjJ2688wJZaj5a0/edit

Additional Description: Please refer to RFC
Risk Level: Medium
Testing: Done
Docs Changes: Done
Release Notes: Done
Fixes #11050

PR #22054, opened 2022-07-07
Big edf entry deadline when slow start enabled.

pyrl247:
it will cause big edf entry deadline, and the entry will not be selected in a
long time.

Signed-off-by: Guang Yang <pyrl247@gmail.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/upstream/load_balancer_impl.cc
+++ b/source/common/upstream/load_balancer_impl.cc
@@ -1004,7 +1004,7 @@ double EdfLoadBalancerBase::applySlowStartFactor(double host_weight, const Host&
     }
 
     ASSERT(aggression_ > 0.0);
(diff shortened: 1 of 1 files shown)


pyrl247:
https://github.com/envoyproxy/envoy/issues/19526 have the same problem,min_weight_percent could fix the problem,but I think change it to 1 second could be more appropriate.

wrowe:
/assign @mattklein123 

Appears you initially had reviewed https://github.com/envoyproxy/envoy/pull/13176 when this was introduced

mattklein123 (changes requested):
Thanks for working on this. I don't fully understand this change so requesting some more information. Thank you!

/wait

PR #22057, opened 2022-07-07
uds: fixing a regression with tls uds upstreams

alyssawilk:
Risk Level: low
Testing: new UDS TLS test
Docs Changes: n/a
Release Notes:
Fixes https://github.com/envoyproxy/envoy/pull/21356 (because we can never take out the optional check)


alyssawilk:
cc @adisuissa we should get this in before we cut the release

moderation:
Wondering if this is why I could never get https://github.com/envoyproxy/envoy/issues/11569#issuecomment-645623588 to work ?

snowp (approved):
Thanks!

Issue #21490, opened 2022-05-27
GrpcJsonTranscoder returns 200 with an empty body when the gRPC message is too deep


bmonnahan:
*Title*: GrpcJsonTranscoder returns 200 with an empty body when the gRPC message is too deep 

*Description*:

Actual: We have response messages with a recursive structure which can lead to deep objects.  Above 64 levels the json transcoder doesn't successfully convert the message to json. Instead it returns an empty body with a 200 HTTP status. Nothing shows up in the logs even with trace logging turned on.

Expected: An error in the log and if possible a HTTP error status is returned to the client.

*Repro steps*:

Using the code & config below...

This call works:
`curl -v http://localhost:9091/testRead/64`

This one doesn't (empty response body, status 200):
`curl -v http://localhost:9091/testRead/65`

In the logs for the service we can see that the gRPC message is generated successfully in both cases.

Our envoy comes from a docker image based on **envoy-alpine:v1.20.1-1.7**

Proto
```
syntax = "proto3";
package com.test.recursion;

import "google/api/annotations.proto";

option java_multiple_files = true;
option java_package = "com.test.recursion";


service TemplateService {

  rpc testRead(TestReadRequest) returns (TestReadResponse) {
    option (google.api.http) = {
      get : "/testRead/{count}"
    };
  };

  rpc testWrite(TestWriteRequest) returns (TestWriteResponse) {
    option (google.api.http) = {
      post : "/testWrite",
      body : "*"
    };
  };
}

message TestReadResponse {
  string message = 1;
  Tree tree = 2;
}

message TestReadRequest {
  int32 count = 1;
}

message TestWriteRequest {
  Tree tree = 1;
}


message TestWriteResponse {
  string message = 1;
}

message Tree {
  oneof left_child {
    Tree left_tree = 1;
    string left_leaf = 2;
  }
  oneof right_child {
    Tree right_tree = 3;
    string right_leaf = 4;
  }
}
```

Service Impl
```
package com.test.recursion.grpc;


import com.google.protobuf.InvalidProtocolBufferException;
import com.google.protobuf.util.JsonFormat;
import com.test.recursion.TemplateServiceGrpc;
import com.test.recursion.TestReadRequest;
import com.test.recursion.TestReadResponse;
import com.test.recursion.TestWriteRequest;
import com.test.recursion.TestWriteResponse;
import com.test.recursion.Tree;
import io.grpc.ManagedChannel;

import io.grpc.stub.StreamObserver;
[...]

PR #22060, opened 2022-07-07
grpc_transcoder: check response transcoder error

qiwzhang:

To fix https://github.com/envoyproxy/envoy/issues/21490

grpc response transcoder will fail with error if the proto message is too deep (>64),  but grpc_transcoder filter encodeData() function did not check the error status.  This pr adds code to check the error status and adds a test.

Risk Level:  Low
Testing:  Unit-test
Docs Changes: No
Release Notes: Yes

### DIFF
--- a/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
+++ b/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
@@ -656,6 +656,9 @@ Http::FilterDataStatus JsonTranscoderFilter::encodeData(Buffer::Instance& data, 
   }
 
   readToBuffer(*transcoder_->ResponseOutput(), data);
+  if (checkIfResponseTranscoderFailed()) {
+    return Http::FilterDataStatus::StopIterationNoBuffer;
+  }
 
   if (!method_->descriptor_->server_streaming() && !end_stream) {
     // Buffer until the response is complete.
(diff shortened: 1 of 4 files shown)


source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -656,6 +656,9 @@ Http::FilterDataStatus JsonTranscoderFilter::encodeData(Buffer::Instance& data,
   }
 
   readToBuffer(*transcoder_->ResponseOutput(), data);
+  if (checkIfResponseTranscoderFailed()) {
```
shuoyang2016, line 659: remove TODO on line:667?

qiwzhang: Done

wrowe:
/assign @lizan 

as code-owner/maintainer

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -757,6 +762,25 @@ bool JsonTranscoderFilter::checkIfTranscoderFailed(const std::string& details) {
   return false;
 }
 
+bool JsonTranscoderFilter::checkAndRejectIfResponseTranscoderFailed() {
+  const auto& response_status = transcoder_->ResponseStatus();
+  if (!response_status.ok()) {
+    ENVOY_LOG(debug, "Transcoding response error {}", response_status.ToString());
+    error_ = true;
+    encoder_callbacks_->sendLocalReply(
+        Http::Code::InternalServerError,
```
lizan, line 771: 502 Bad Gateway is more appropriate for this case?

>    The 502 (Bad Gateway) status code indicates that the server, while
   acting as a gateway or proxy, received an invalid response from an
   inbound server it accessed while attempting to fulfill the request.

lizan (approved):

PR #22061, opened 2022-07-07
api: updating factory APIs

alyssawilk:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/server/BUILD
+++ b/source/server/BUILD
@@ -454,7 +454,7 @@ envoy_cc_library(
         "listener_manager_impl.h",
     ],
     deps = [
-        ":server_factory_context_wrapper_lib",
+        ":listener_factory_context_wrapper_lib",
         "active_raw_udp_listener_config",
         ":configuration_lib",
         ":drain_manager_lib",
@@ -567,8 +567,8 @@ envoy_cc_library(
 )
 
 envoy_cc_library(
-    name = "server_factory_context_wrapper_lib",
-    hdrs = ["server_factory_context_wrapper.h"],
+    name = "listener_factory_context_wrapper_lib",
+    hdrs = ["listener_factory_context_wrapper.h"],
     deps = [
         "//envoy/server:factory_context_interface",
     ],
--- a/source/server/listener_factory_context_wrapper.h
+++ b/source/server/listener_factory_context_wrapper.h
@@ -7,10 +7,10 @@ namespace Server {
 
 // Wraps the global server context with a listener-local version, so that
 // downstream filters can access the downstream factory context.
-class ServerFactoryContextWrapper : public Configuration::ServerFactoryContext {
+class ListenerFactoryContextWrapper : public Configuration::ServerFactoryContext {
 public:
-  ServerFactoryContextWrapper(Configuration::ServerFactoryContext& context,
-                              Configuration::DownstreamFactoryContext& downstream_context)
+  ListenerFactoryContextWrapper(Configuration::ServerFactoryContext& context,
+                                Configuration::DownstreamFactoryContext& downstream_context)
       : context_(context), downstream_context_(downstream_context) {}
 
   // Configuration::ServerFactoryContext
--- a/source/server/listener_impl.h
+++ b/source/server/listener_impl.h
@@ -23,7 +23,7 @@
 #include "source/common/init/target_impl.h"
 #include "source/common/quic/quic_stat_names.h"
 #include "source/server/filter_chain_manager_impl.h"
-#include "source/server/server_factory_context_wrapper.h"
+#include "source/server/listener_factory_context_wrapper.h"
 #include "source/server/transport_socket_config_impl.h"
 
 namespace Envoy {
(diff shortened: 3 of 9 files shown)


source/common/filter/config_discovery_impl.h
```diff
@@ -192,7 +192,8 @@ class HttpDynamicFilterConfigProviderImpl
     auto* factory =
         Registry::FactoryRegistry<Server::Configuration::NamedHttpFilterConfigFactory>::getFactory(
             factory_name);
-    const bool is_terminal_filter = factory->isTerminalFilterByProto(message, factory_context_);
+    const bool is_terminal_filter =
+        factory->isTerminalFilterByProto(message, factory_context_.getServerFactoryContext());
```
kyessenov, line 196: Thanks, this is great for ECDS. We can erase a type template parameter later if all filter factories depend on server context rather than factory context.

envoy/server/factory_context.h
```diff
@@ -177,14 +205,36 @@ class ServerFactoryContext : public virtual CommonFactoryContext {
...
+  /**
+   * @return ProcessContextOptRef an optional reference to the
+   * process context. Will be unset when running in validation mode.
+   */
+  virtual ProcessContextOptRef processContext() PURE;
+
+  /**
+   * @return the downstream factory context, for downstream filters.
+   */
+  virtual OptRef<DownstreamFactoryContext> downstreamContext() PURE;
```
kyessenov, line 228: I'm a bit concerned about subtle lifetime issues here with ECDS. An ECDS filter can outlive the parent listener when shared, so a reference here would become stale in a non-obvious way.

mattklein123: Can you expand more on how a filter would be shared? It seems strange to me that it would work this way if ultimately the filters need to get information for the listener they are responding on? I don't know much about ECDS.

Assuming this is true I guess this needs to be a weak ptr?

source/server/server_factory_context_wrapper.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include "envoy/server/factory_context.h"
+
+namespace Envoy {
+namespace Server {
+
+// Wraps the global server context with a listener-local version, so that
+// downstream filters can access the downstream factory context.
+class ServerFactoryContextWrapper : public Configuration::ServerFactoryContext {
```
mattklein123, line 10: If we are going to bikeshed maybe ListenerFactoryContextWrapper or something? I honestly find all of the different factory contexts very confusing at this point.

source/server/listener_impl.cc
```diff
@@ -280,7 +281,7 @@ ProcessContextOptRef ListenerFactoryContextBaseImpl::processContext() {
 }
 Configuration::ServerFactoryContext&
 ListenerFactoryContextBaseImpl::getServerFactoryContext() const {
```
mattklein123, line 283: Don't we already have a context wrapper in here that could do the wrapping? There are so many wrappers. :joy:

mattklein123:
At a high level this makes sense to me modulo the small comments. Anything you can do to improve names/docs/etc. around all the contexts would be appreciated as IMO it's very confusing at this point! Thank you for working on this.

envoy/server/factory_context.h
kyessenov: It happens when listener 1 opens ECDS stream, then listener 2 shares the stream, then listener 1 is deleted. ECDS filters are dis-associated from listeners in some ways. Filters don't really need much from the parent at runtime, but the factories can capture some listener state if that becomes accessible from root factory context. I think we need to make it more explicit the limited lifetime of those references.

alyssawilk: well right now, the lifetimes are that of the listener, so pretty much guaranteed for the lifetime of all the code calling it.
given this is only actually used in 2 places (wasm / auth, I guess 3 if you count the sds test) I'm inclined to land as-is with no lifetime changes, and land ECDS lifetime changes separately in a follow up as it's needed.  Would that work for you?

source/server/listener_impl.cc
alyssawilk: given this is a new class I don't think so?

mattklein123: I guess the larger point is there are so many wrappers it's incredibly confusing. One general thought: At some point it made sense (I guess) to have all of the inheritance. Does that makes sense anymore? Would it be cleaner / easier to understand if all the factory interfaces were split? Not sure just throwing it out there.

envoy/server/factory_context.h
kyessenov: Is the idea to have this populated sometimes, and leave it empty the other times when the lifetime is not guaranteed? I'm confused why a server context points to the more specific listener context.

alyssawilk: likely abandoning in favor of https://github.com/envoyproxy/envoy/pull/22134

PR #22080, opened 2022-07-08
dns: add a getaddrinfo based system resolver

mattklein123:
This can be used when using the system resolver is desired. For
example, on Android.

Risk Level: None, new extension
Testing: New tests
Docs Changes: Added
Release Notes: Added
Platform Specific Features: TBD


mattklein123:
Not quite finished but I want to see what problems CI gives me.

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,216 @@
...
+      ENVOY_LOG(debug, "popped pending query [{}]", next_query->dns_name_);
+
+      addrinfo hints;
+      memset(&hints, 0, sizeof(hints));
+      hints.ai_flags = AI_ADDRCONFIG;
+      hints.ai_family = AF_UNSPEC;
+      // If we don't specify a socket type, every address will appear twice, once
+      // for SOCK_STREAM and one for SOCK_DGRAM. Since we do not return the family
+      // anyway, just pick one.
+      hints.ai_socktype = SOCK_STREAM;
```
mattklein123, line 160: This is copied from the Android OS implementation.

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,216 @@
...
+    auto new_query = std::make_unique<PendingQuery>(dns_name, dns_lookup_family, callback);
+    absl::MutexLock guard(&mutex_);
+    pending_queries_.emplace_back(std::move(new_query));
+    return pending_queries_.back().get();
+  }
+
+  void resetNetworking() override {}
+
+private:
+  struct PendingQuery : public ActiveDnsQuery {
```
RyanTheOptimist, line 45: nit: ActiveDnsQuery is a class so PendingQuery should be a class too.

mattklein123:
OK I think this is ready.

source/extensions/extensions_build_config.bzl
```diff
@@ -339,9 +339,9 @@ EXTENSIONS = {
 
     # c-ares DNS resolver extension is recommended to be enabled to maintain the legacy DNS resolving behavior.
     "envoy.network.dns_resolver.cares":                "//source/extensions/network/dns_resolver/cares:config",
-
     # apple DNS resolver extension is only needed in MacOS build plus one want to use apple library for DNS resolving.
     "envoy.network.dns_resolver.apple":                "//source/extensions/network/dns_resolver/apple:config",
+    "envoy.network.dns_resolver.getaddrinfo":          "//source/extensions/network/dns_resolver/getaddrinfo:config",
```
alyssawilk, line 344: optional nit - the other two are commented inline - might want to for htis?

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,217 @@
...
+
+#include "source/common/api/os_sys_calls_impl.h"
+#include "source/common/network/address_impl.h"
+
+namespace Envoy {
+namespace Network {
+
+// This resolver uses getaddrinfo() on a dedicated resolution thread. Thus, it is only suitable
+// currently for relatively low rate resolutions. In the future a thread pool could be added if
+// desired.
```
alyssawilk, line 14: Should have a note in the docs as well, yes?

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,217 @@
...
+    for (auto ai = addrinfo_result; ai != nullptr; ai = ai->ai_next) {
+      if (ai->ai_family == AF_INET) {
+        sockaddr_in address;
+        memset(&address, 0, sizeof(address));
+        address.sin_family = AF_INET;
+        address.sin_port = 0;
+        address.sin_addr = reinterpret_cast<sockaddr_in*>(ai->ai_addr)->sin_addr;
+
+        v4_results.emplace_back(DnsResponse(std::make_shared<const Address::Ipv4Instance>(&address),
+                                            std::chrono::seconds(60)));
```
alyssawilk, line 75: where does 60 come from?

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }
+    Api::OsSysCallsSingleton::get().freeaddrinfo(addrinfo_result);
+
+    std::list<DnsResponse> final_results;
+    switch (query.dns_lookup_family_) {
+    case DnsLookupFamily::All: {
+      final_results = std::move(v4_results);
+      final_results.splice(final_results.begin(), v6_results);
```
alyssawilk, line 92: defering to @RyanTheOptimist re happy eyeballs sorting

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,217 @@
...
+  // Must be a shared_ptr for passing around via post.
+  using PendingQuerySharedPtr = std::shared_ptr<PendingQuery>;
+
+  // Parse a GAI response and determine the final address list.
+  std::pair<ResolutionStatus, std::list<DnsResponse>>
+  processGaiResponse(const PendingQuery& query, addrinfo* addrinfo_result) {
+    std::list<DnsResponse> v4_results;
+    std::list<DnsResponse> v6_results;
+    for (auto ai = addrinfo_result; ai != nullptr; ai = ai->ai_next) {
+      if (ai->ai_family == AF_INET) {
```
alyssawilk, line 67: optional (given we commented on perf) but it'd be nice to skip v4/v6 inserts if they're not wanted

test/extensions/network/dns_resolver/getaddrinfo/getaddrinfo_test.cc
```diff
@@ -0,0 +1,292 @@
+#include "envoy/extensions/network/dns_resolver/getaddrinfo/v3/getaddrinfo_dns_resolver.pb.h"
```
alyssawilk, line 1: given cross-thread work let's also make sure one of the e2e tests uses this as well.

alyssawilk (commented):
Looks like coverage isn't happy but here's some first ass comments while you sort that :-)
/wait

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
soulxu, line 67: nit, we can skip this when dns_lookup_family is v6 only

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,217 @@
...
+        sockaddr_in address;
+        memset(&address, 0, sizeof(address));
+        address.sin_family = AF_INET;
+        address.sin_port = 0;
+        address.sin_addr = reinterpret_cast<sockaddr_in*>(ai->ai_addr)->sin_addr;
+
+        v4_results.emplace_back(DnsResponse(std::make_shared<const Address::Ipv4Instance>(&address),
+                                            std::chrono::seconds(60)));
+      } else if (ai->ai_family == AF_INET6) {
```
soulxu, line 76: ditto, we can skip this when dns_lookup_family is v4 only

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,217 @@
...
+
+      addrinfo hints;
+      memset(&hints, 0, sizeof(hints));
+      hints.ai_flags = AI_ADDRCONFIG;
+      hints.ai_family = AF_UNSPEC;
+      // If we don't specify a socket type, every address will appear twice, once
+      // for SOCK_STREAM and one for SOCK_DGRAM. Since we do not return the family
+      // anyway, just pick one.
+      hints.ai_socktype = SOCK_STREAM;
+      addrinfo* addrinfo_result;
```
soulxu, line 160: Should we put this into a helper class, which destructor will call the `OsSysCallsImpl::freeaddrinfo`

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
mattklein123: It seemed like a reasonable default given that this API doesn't support getting TTL. Would you prefer a different default? I will document it and I guess we could make it configurable later if anyone cares.

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
mattklein123: Yes will do.

source/extensions/extensions_build_config.bzl
mattklein123: Sure

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
mattklein123: The sorting is done at a high layer so I don't think it matters for this.

test/extensions/network/dns_resolver/getaddrinfo/getaddrinfo_test.cc
mattklein123: The tests I wrote does actual resolutions with real threads so it's probably OK? Otherwise I can try to make one of the DNS cache integration tests use this.

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
mattklein123: Sure.

mattklein123:
Unfortunately the coverage tool is missing braces so there isn't much I can do about this. I will drop the limit:
https://storage.googleapis.com/envoy-pr/17b2cb1/coverage/source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc.gcov.html

adisuissa (commented):
/lgtm api

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
alyssawilk: yeah definitely want comments.  If this is the default you want for Lyft I'd say you can skip the knob for now.

test/extensions/network/dns_resolver/getaddrinfo/getaddrinfo_test.cc
mattklein123: Looks like no special visibility rules were required so I will add the test.

source/common/api/posix/os_sys_calls_impl.cc
```diff
@@ -404,5 +404,13 @@ SysCallIntResult OsSysCallsImpl::getifaddrs([[maybe_unused]] InterfaceAddressVec
 #endif
 }
 
+SysCallIntResult OsSysCallsImpl::getaddrinfo(const char* node, const char* service,
+                                             const addrinfo* hints, addrinfo** res) {
+  const auto rc = ::getaddrinfo(node, service, hints, res);
```
RyanTheOptimist, line 409: Your call, but I'm not sure offhand what the return value of getaddrinfo is (int?) so I'd recommend being explicit about the return type instead of using auto here.

source/common/api/win32/os_sys_calls_impl.cc
```diff
@@ -458,5 +458,13 @@ SysCallIntResult OsSysCallsImpl::getifaddrs([[maybe_unused]] InterfaceAddressVec
   PANIC("not implemented");
 }
 
+SysCallIntResult OsSysCallsImpl::getaddrinfo(const char* node, const char* service,
+                                             const addrinfo* hints, addrinfo** res) {
+  const auto rc = ::getaddrinfo(node, service, hints, res);
```
RyanTheOptimist, line 463: ditto

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,237 @@
...
+  public:
+    AddrInfoWrapper(addrinfo* info) : info_(info) {}
+    ~AddrInfoWrapper() { Api::OsSysCallsSingleton::get().freeaddrinfo(info_); }
+    const addrinfo* get() { return info_; }
+
+  private:
+    addrinfo* info_;
+  };
+
+  // Parse a GAI response and determine the final address list. We could potentially avoid adding
```
RyanTheOptimist, line 72: GAI = getaddrinfo?

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,237 @@
...
+  // v4 or v6 addresses if we know they will never be used. Right now the final filtering is done
+  // below and this code is kept simple.
+  std::pair<ResolutionStatus, std::list<DnsResponse>>
+  processGaiResponse(const PendingQuery& query, const addrinfo* addrinfo_result) {
```
RyanTheOptimist, line 76: I'd be inclined to remove "Gai" from this method name as the style guide is not a big fan of abbreviations (and since there aren't other kinds of responses which need to be processed)

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,237 @@
...
+          break;
+        }
+
+        next_query = std::move(pending_queries_.front());
+        pending_queries_.pop_front();
+      }
+
+      ENVOY_LOG(debug, "popped pending query [{}]", next_query->dns_name_);
+
+      // For mock testing make sure the GAI response is freed prior to the post.
```
RyanTheOptimist, line 164: nit: GAI => getaddrinfo.

RyanTheOptimist (commented):
Other than a few nits, this looks great (not surprisingly).

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
```diff
@@ -0,0 +1,237 @@
...
+    bool cancelled_{false};
+  };
+  // Must be a shared_ptr for passing around via post.
+  using PendingQuerySharedPtr = std::shared_ptr<PendingQuery>;
+
+  // RAII wrapper to free the `addrinfo`.
+  class AddrInfoWrapper : NonCopyable {
+  public:
+    AddrInfoWrapper(addrinfo* info) : info_(info) {}
+    ~AddrInfoWrapper() { Api::OsSysCallsSingleton::get().freeaddrinfo(info_); }
```
soulxu, line 65: It seems like unsafe to call `OsSysCallsImpl::freeaddrinfo` with nullptr (it will get nullptr when `OsSysCallsImpl::getaddrinfo` failed at https://github.com/envoyproxy/envoy/pull/22080/files#diff-01fdbaa5394bb19e54e31b01894d73299a3578fb2b54abcd8a5722b30db3ef80R175)

The man page didn't clear about this case https://man7.org/linux/man-pages/man3/getaddrinfo.3.html

But after googling, there is someone saw the segment fault when passing null to freeaddrinfo.

Also I'm thinking maybe we needn't pass the pointer to this wrapper, just create a class like:

```
Class AddrInfo : NonCopyable {
public:
    ~AddrInfo() { if (info_) {Api::OsSysCallsSingleton::get().freeaddrinfo(info_); } }

   addrinfo* info_{nullptr};
}
```

Then we pass the `AddrInfo::info_` to the `OsSysCallsImpl::getaddrinfo`. Or we ask `OsSysCallsImpl::getaddrinfo` accept a reference of `AddrInfo`, then no one will use `addrinfo*` nakely. But I don't if we want to add more logic for `OsSysCallsImpl::getaddrinfo` or just want to keep same with os interface.

soulxu (changes requested):

source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.cc
mattklein123: Thanks for the suggestion but I think I'm done making changes at this point. I test both the null and not null case and it seems to work ok. We can change it later if needed.

soulxu: cool, that works for me. This extension was added for android, and you tested that works well.

soulxu (approved):
LGTM, thanks

api/envoy/extensions/network/dns_resolver/getaddrinfo/v3/getaddrinfo_dns_resolver.proto
```diff
@@ -0,0 +1,30 @@
...
+// [#protodoc-title: getaddrinfo DNS resolver]
+// [#extension: envoy.network.dns_resolver.getaddrinfo]
+
+// Configuration for getaddrinfo DNS resolver. This resolver will use the system's getaddrinfo()
+// function to resolve hosts.
+//
+// .. attention::
+//
+//   This resolver uses a single background thread to do resolutions. As such, it is not currently
+//   advised for use in situations requiring a high resolution late. A thread pool can be added
```
ramaraochavali, line 22: in situations requiring a high resolution late -> in situations requiring a high resolution rate?

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
```diff
@@ -225,43 +266,22 @@ INSTANTIATE_TEST_SUITE_P(IpVersions, ProxyFilterIntegrationTest,
...
-  testConnectionTiming(response, true, original_usec);
-  EXPECT_EQ(1, test_server_->counter("dns_cache.foo.dns_query_attempt")->value());
-  EXPECT_EQ(1, test_server_->counter("dns_cache.foo.host_added")->value());
-  // Make sure dns timings are tracked for cache-hits.
-  ASSERT_FALSE(response->headers().get(Http::LowerCaseString("dns_start")).empty());
-  ASSERT_FALSE(response->headers().get(Http::LowerCaseString("dns_end")).empty());
+TEST_P(ProxyFilterIntegrationTest, RequestWithBody) { requestWithBodyTest(); }
+
+// Do a sanity check using the getaddrinfo() resolver.
+TEST_P(ProxyFilterIntegrationTest, RequestWithBodyGetAddrInfoResolver) {
```
alyssawilk, line 272: Can we do RequestWithUnknownDomain as well, to have coverage of non-injected lookup failure?

alyssawilk (commented):
Just the one testing nit and you're good to go.  Thanks!

test/extensions/filters/http/dynamic_forward_proxy/proxy_filter_integration_test.cc
mattklein123: Yup will do.

PR #22081, opened 2022-07-09
multiple addresses: add integration tests

soulxu:
Commit Message: multiple addresses: add integration tests
Additional Description:
Risk Level: low
Testing: integration tests
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/config/utility.cc
+++ b/test/config/utility.cc
@@ -724,7 +724,6 @@ ConfigHelper::ConfigHelper(const Network::Address::IpVersion version, Api::Api& 
   auto* static_resources = bootstrap_.mutable_static_resources();
   for (int i = 0; i < static_resources->listeners_size(); ++i) {
     auto* listener = static_resources->mutable_listeners(i);
-
     if (listener->mutable_address()->has_envoy_internal_address()) {
       ENVOY_LOG_MISC(
           debug, "Listener {} has internal address {}. Will not reset to loop back socket address.",
@@ -737,7 +736,6 @@ ConfigHelper::ConfigHelper(const Network::Address::IpVersion version, Api::Api& 
                      i, listener->mutable_address()->pipe().path());
       continue;
     }
-
     auto* listener_socket_addr = listener->mutable_address()->mutable_socket_address();
     if (listener_socket_addr->address() == "0.0.0.0" || listener_socket_addr->address() == "::") {
       listener_socket_addr->set_address(Network::Test::getAnyAddressString(version));
--- a/test/config/utility.cc
+++ b/test/config/utility.cc
@@ -73,7 +73,22 @@ static_resources:
                      Platform::null_device_path, Platform::null_device_path);
 }
 
-std::string ConfigHelper::baseConfig() {
+std::string ConfigHelper::baseConfig(bool multiple_addresses) {
+  if (multiple_addresses) {
+    return absl::StrCat(baseConfigNoListeners(), R"EOF(
+  listeners:
+  - name: listener_0
+    address:
+      socket_address:
+        address: 127.0.0.1
+        port_value: 0
+    additional_addresses:
+    - address:
+        socket_address:
+          address: 127.0.0.1
+          port_value: 0
+)EOF");
+  }
   return absl::StrCat(baseConfigNoListeners(), R"EOF(
   listeners:
   - name: listener_0
@@ -84,8 +99,9 @@ std::string ConfigHelper::baseConfig() {
 )EOF");
 }
 
-std::string ConfigHelper::baseUdpListenerConfig(std::string listen_address) {
-  return fmt::format(R"EOF(
+std::string ConfigHelper::baseUdpListenerConfig(std::string listen_address,
+                                                bool multiple_addresses) {
+  std::string config = fmt::format(R"EOF(
 admin:
   access_log:
   - name: envoy.access_loggers.file
@@ -108,6 +124,28 @@ static_resources:
               socket_address:
                 address: 127.0.0.1
                 port_value: 0
+)EOF",
+                                   Platform::null_device_path);
+
+  if (multiple_addresses) {
+    return absl::StrCat(config, fmt::format(R"EOF(
+  listeners:
+    name: listener_0
+    address:
+      socket_address:
+        address: {}
+        port_value: 0
+        protocol: udp
+    additional_addresses:
+    - address:
+        socket_address:
+          address: {}
+          port_value: 0
+          protocol: udp
+)EOF",
+                                            listen_address, listen_address));
+  }
+  return absl::StrCat(config, fmt::format(R"EOF(
   listeners:
     name: listener_0
     address:
@@ -116,7 +154,7 @@ static_resources:
         port_value: 0
         protocol: udp
 )EOF",
-                     Platform::null_device_path, listen_address);
+                                          listen_address));
 }
 
 std::string ConfigHelper::tcpProxyConfig() {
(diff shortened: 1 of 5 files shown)


test/integration/listener_lds_integration_test.cc
```diff
@@ -64,43 +67,13 @@ class ListenerIntegrationTest : public HttpIntegrationTest,
...
-          rds_api_config_source->set_transport_api_version(envoy::config::core::v3::V3);
-          envoy::config::core::v3::GrpcService* grpc_service =
-              rds_api_config_source->add_grpc_services();
-          setGrpcService(*grpc_service, "rds_cluster", getRdsFakeUpstream().localAddress());
-        });
-
+    setUpGrpcRds();
     // Note this has to be the last modifier as it nuke static_resource listeners.
     setUpGrpcLds();
+    ENVOY_LOG_MISC(debug, "listener config: {}", listener_config_.DebugString());
```
mattklein123, line 73: Don't we already print this somewhere? I thought we do so just want to make sure we aren't printing this twice?

test/integration/listener_lds_integration_test.cc
```diff
@@ -199,9 +172,100 @@ class ListenerIntegrationTest : public HttpIntegrationTest,
...
+      auto* lds_config_source = bootstrap.mutable_dynamic_resources()->mutable_lds_config();
+      lds_config_source->set_resource_api_version(envoy::config::core::v3::ApiVersion::V3);
+      auto* lds_api_config_source = lds_config_source->mutable_api_config_source();
+      lds_api_config_source->set_api_type(envoy::config::core::v3::ApiConfigSource::GRPC);
+      lds_api_config_source->set_transport_api_version(envoy::config::core::v3::V3);
+      envoy::config::core::v3::GrpcService* grpc_service =
+          lds_api_config_source->add_grpc_services();
+      setGrpcService(*grpc_service, "lds_cluster", getLdsFakeUpstream().localAddress());
+    });
+  }
```
mattklein123, line 260: Can we have some helpers and/or common base class/mixin to avoid duplicating this code?

mattklein123 (commented):
Thanks this is super awesome. A couple of small comments. I'm also going to assign @alyssawilk and @danzh2010 to look at the integration test framework and QUIC changes, respectively. Thank you!

/wait

test/integration/listener_lds_integration_test.cc
soulxu: yea, I add it when I debug, then I forget to remove it, let me cleanup it.

test/integration/listener_lds_integration_test.cc
soulxu: yea, let me give a try

mattklein123:
LGTM but would still like to get a quick sanity check from @alyssawilk or @danzh2010. Thank you!

mattklein123 (dismissed):

test/config/utility.cc
```diff
@@ -73,7 +73,22 @@ std::string ConfigHelper::baseConfigNoListeners() {
...
+std::string ConfigHelper::baseConfig(bool multiple_addresses) {
+  if (multiple_addresses) {
+    return absl::StrCat(baseConfigNoListeners(), R"EOF(
+  listeners:
+  - name: listener_0
+    address:
+      socket_address:
+        address: 127.0.0.1
+        port_value: 0
+    additional_addresses:
```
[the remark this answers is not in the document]
soulxu: I added a flag `multiple_addresses` to the `baseConfig()`, the default value is false:
https://github.com/envoyproxy/envoy/pull/22081/files#diff-a00c6efc42fc025c0cea76be1b3c81c6993b65edb65c77e5001ed22157236386R157

only set it as true for the multiple addresses tests, like:
https://github.com/envoyproxy/envoy/pull/22081/files#diff-7fe1ad3a088b5d0ea44d9bace0dfa0675d5009bcb538afb778ebad50c802ddeaR231

so it won't affect any existing test.

I also checked, we will rename the listener to `http` for legacy test
https://github.com/envoyproxy/envoy/blob/cb78092e9219da3fd77e49aa1273435696c72921/test/integration/http_integration.cc#L327-L329

With this multiple addresses test, the `http` port still points to the first address. But for the multiple addresses test, it will register ports name by itself, like
https://github.com/envoyproxy/envoy/pull/22081/files#diff-7fe1ad3a088b5d0ea44d9bace0dfa0675d5009bcb538afb778ebad50c802ddeaR434

so all the multiple addresses tests won't depend on `http` port.

But yes, I didn't have a strong reason to change `baseConfig()`, I just feel it is easier and readable to write a yaml config than the modifier.

Let me know if you think changing the base config is still good choice.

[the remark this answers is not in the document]
alyssawilk: ehh, I prefer to not have 8 ways to set configuration but I'll accept your arguments for readability given my main concern (that it not affect other tests) is as you point out mistaken, so let's leave as-is :-)
I'll defer to Dan for QUIC comments, then do final pass.

test/integration/quic_http_integration_test.cc
```diff
@@ -347,10 +345,118 @@ class QuicHttpIntegrationTest : public HttpIntegrationTest,
...
+    for (size_t i = 1; i <= concurrency_; ++i) {
+      // The BPF filter and ActiveQuicListener::destination() look at the 1st word of connection id
+      // in the packet header. And currently all QUIC versions support >= 8 bytes connection id. So
+      // create connections with the first 4 bytes of connection id different from each
+      // other so they should be evenly distributed.
+      designated_connection_ids_.push_back(quic::test::TestConnectionId(i << 32));
+      // TODO(sunjayBhatia,wrowe): deserialize this, establishing all connections in parallel
+      // (Expected to save ~14s each across 6 tests on Windows)
+      codec_clients1.push_back(makeHttpConnection(lookupPort("address1")));
+      designated_connection_ids_.push_back(quic::test::TestConnectionId(i << 32));
```
danzh2010, line 380: Here codec_clients1[i] and codec_clients2[i] use the same connection ID and they connect to the same listener (on different ports). I thought this would be a connection ID collision, but the test seems to work. Maybe I missed some obvious facts, but can you comment more on what this test is testing and how it works?

soulxu: When using multiple addresses in a single listener, we will have single ListenerImpl for it. But actually, we create multiple separated ActiveQuicListener for each address.

For the same connection id's `codec_clients1` and `codec_clients2` will go to the different ActiveQuicListener instance, so there is no collision.

The multiple addresses didn't touch QUIC direclty accept the change of UdpWorkerRouter https://github.com/envoyproxy/envoy/pull/21940

That PR ensures the udp/quic request going to the correct worker and correct active listener.

If there is a bug that let the connection with the same id but with a different destination address go to the same listener, then this test will fail (fail with the connection ID collision as you said).

Let me add a comment here to explain this. But also let me know if I didn't understand quic code correctly.

Thanks for the review!

danzh2010: Okay, I thought the same ActiveQuicListener would listen on both addresses, but in fact they are still 1:1 mapping. So this test is verifying that both listeners are dispatching quic packets according to connection IDs correctly. It makes sense! Thanks for explanation!

And it worth pointing out that such implementation further disassociate supporting QUIC server preferred address with what Greg proposed here: https://github.com/envoyproxy/envoy/issues/11184#issuecomment-679214885, which requires the same listener to listen on all the configured addresses. This might not be a issue, but worth clarifying on the PR about that.

soulxu: > And it worth pointing out that such implementation further disassociate supporting QUIC server preferred address with what Greg proposed here: [#11184 (comment)](https://github.com/envoyproxy/envoy/issues/11184#issuecomment-679214885), which requires the same listener to listen on all the configured addresses. This might not be a issue, but worth clarifying on the PR about that.

Thanks for pointing this out, yes, that case isn't the scope of this implementation. But I will think about it in the future steps, it seems like can be done with some change for UdpWorkerRouter, but anyway I will think about it more in the next steps.

soulxu: @danzh2010 I add a comment for that, PTAL again, thanks in advance!

mattklein123 (changes requested):
/wait

mattklein123 (approved):

soulxu:
@alyssawilk gently ping :)

soulxu:
@alyssawilk thanks!

PR #22088, opened 2022-07-10
lua: use pushlstring instead of pushstring

agrawroh:
### Background

Currently, while setting the value of fields for dynamic metadata in the LUA filter we use `pushstring` with no length indication and hence it ends up with a null terminated character array.

### Changes

This PR changes the use of `pushstring` -> `pushlstring` and supplies the length of the explicitly so that we can also set/get binary data.

---
**Commit Message:** change the use of `pushstring` to `pushlstring` for setting the lua metadata values.
**Risk Level:** Low
**Testing:** Unit Tests
**Docs Changes:** N/A
**Release Notes:** N/A


agrawroh:
@mattklein123 Should we change `pushstring` -> `pushlstring` everywhere?

test/extensions/filters/http/lua/lua_filter_test.cc
```diff
@@ -1800,6 +1800,43 @@ TEST_F(LuaHttpFilterTest, GetRequestedServerName) {
...
+      for idx = 1, data_length do
+        local hex_value = string.format('%x', string.byte(bin_data, idx))
+        if string.len(string.format('%x', string.byte(bin_data, idx))) < 2 then
+          hex_table[data_length - idx + 1] = '0'..hex_value
+        else
+          hex_table[data_length - idx + 1] = hex_value
+        end
+      end
+      local hex_data = table.concat(hex_table, '')
+      local int64_data = string.format('%d', tonumber(hex_data, 16))
```
fedya-at-db, line 1819: This feels unrelated to the functionality. While yes, you could encode a number with this data, we don't really need to test that part. Why not just stick some binary data in there, and make sure to include some \0 since that's what pushstring chokes on otherwise, and print those as escaped hex (or whatever other way printing is easy) to verify.

test/extensions/filters/http/lua/lua_filter_test.cc
```diff
@@ -1800,6 +1800,43 @@ TEST_F(LuaHttpFilterTest, GetRequestedServerName) {
...
+      request_handle:logTrace('Data: ' .. int64_data)
+    end
+  )EOF"};
+
+  ProtobufWkt::Value metadata_value;
+  const uint64_t e0 = 9688686178336770;
+  metadata_value.set_string_value(reinterpret_cast<char const*>(&e0), 8);
```
fedya-at-db, line 1826: This is a bit too opaque about what's going on. The thing we really care about is that c_str() has embedded nulls, so let's write some of those. e.g.
constexpr uint8_t buffer[] = {'h', 'e', 0x00, 'l', 'l', 'o'};
metadata_value.set_string_value(buffer, sizeof(buffer));

maybe you need to throw in a reinterpret_cast in there, I didn't test this. And throw in a comment in the test that the prior version would result in only "he" being stored.

fedya-at-db:
> @mattklein123 Should we change pushstring -> pushlstring everywhere?
I would +1 this for consistency and better behavior. The alternative is that non-ascii, non-utf8 characters can still make it in, but we stop on early nulls, and there's no error thrown, so it'll be hard to catch. There are only 3 files calling pushstring within envoy anyway.

wbpcode:
> Currently, while setting the value of fields for dynamic metadata in the LUA filter we use `pushstring` with no length indication and hence it ends up with a null terminated character array.

+1

wbpcode:
> @mattklein123 Should we change `pushstring` -> `pushlstring` everywhere?

+1

mattklein123:
> @mattklein123 Should we change pushstring -> pushlstring everywhere?

+1 also, thanks for working on this.

/wait

source/extensions/filters/common/lua/wrappers.cc
```diff
@@ -121,7 +121,7 @@ void MetadataMapHelper::createTable(lua_State* state,
   lua_createtable(state, 0, fields.size());
   for (const auto& field : fields) {
     int top = lua_gettop(state);
-    lua_pushstring(state, field.first.c_str());
+    lua_pushlstring(state, field.first.c_str(), field.first.length());
```
fedya-at-db, line 124: stick with .size() or .length() for consistency, though they're synonyms. I'm used to .size(), but don't know if envoy uses one or the other more frequently. Basic github search has .size() (~600 matches) win out vs .length() (~100 matches).

test/extensions/filters/http/lua/lua_filter_test.cc
```diff
@@ -1800,6 +1800,39 @@ TEST_F(LuaHttpFilterTest, GetRequestedServerName) {
...
+          str_table[#str_table + 1] = string.char(string.byte(bin_data, idx))
+        end
+      end
+      request_handle:logTrace('Data: ' .. table.concat(str_table, ''))
+    end
+  )EOF"};
+
+  ProtobufWkt::Value metadata_value;
+  constexpr uint8_t buffer[] = {'h', 'e', 0x00, 'l', 'l', 'o'};
+  metadata_value.set_string_value(reinterpret_cast<char const*>(&buffer), sizeof(buffer));
```
fedya-at-db, line 1822: This looks super weird to me, but apparently uint8_t *p = foo; and uint8_t *p = &foo; both work because static arrays are strange and are actually of type array with pointer decay. I'd defer to others if they have an opinion, but I've always thought of buffer as already basically a pointer, so would expect this to say reinterpret_cast<char const*>(buffer) (lack of &).

test/extensions/filters/http/lua/lua_filter_test.cc
```diff
@@ -1800,6 +1800,39 @@ TEST_F(LuaHttpFilterTest, GetRequestedServerName) {
...
+// Verify that binary values could also be extracted from dynamicMetadata() in LUA filter.
+TEST_F(LuaHttpFilterTest, GetDynamicMetadataBinaryData) {
+  const std::string SCRIPT{R"EOF(
+    function envoy_on_request(request_handle)
+      local metadata = request_handle:streamInfo():dynamicMetadata():get("envoy.pp")
+      local bin_data = metadata["bin_data"]
+      local str_table = { }
+      local data_length = string.len(metadata["bin_data"])
+      for idx = 1, data_length do
+        if string.byte(bin_data, idx) ~= 0 then
```
fedya-at-db, line 1812: Why skip the null? Now I think this code isn't really distinguishing whether the 0x00 made it to the lua context or not. Imagine a bug where we instead pass this string through MessageToJson which will drop the 0x00 but keep the rest of the characters.

I like what you had earlier with converting this to hex escape codes since it shows to the reader byte by byte what made it over. Alternatively you could call base64Escape and in your assertion show that the value logged will match calling base64Escape on buffer in your test, but that seems slightly more difficult for the reader to follow.

test/extensions/filters/http/lua/wrappers_test.cc
```diff
@@ -418,6 +418,41 @@ TEST_F(LuaStreamInfoWrapperTest, SetGetAndIterateDynamicMetadata) {
...
+  metadata.mutable_fields()->insert({"bin_data", metadata_value});
+
+  setup(SCRIPT);
+
+  StreamInfo::StreamInfoImpl stream_info(Http::Protocol::Http2, test_time_.timeSystem(), nullptr);
+  (*stream_info.metadata_.mutable_filter_metadata())["envoy.pp"] = metadata;
+  Filters::Common::Lua::LuaDeathRef<StreamInfoWrapper> wrapper(
+      StreamInfoWrapper::create(coroutine_->luaState(), stream_info), true);
+
+  EXPECT_CALL(printer_, testPrint("Data: hello"));
```
fedya-at-db, line 451: Ditto in the test above, but we want to see the 0x00 reflected in our assertion somehow. I recommend the hex escape codes you were using before as they make it pretty clear that everything stayed the same byte by byte, but other methods could work too. This assertion though makes me wonder why we dropped the 0x00, is it your lua code or is it a bug that dropped the 0x00 even before your lua code.

source/extensions/filters/common/lua/wrappers.cc
agrawroh: Changed it to `.size()`. Thanks!

source/extensions/filters/common/lua/wrappers.cc
```diff
@@ -22,7 +22,7 @@ template <typename StringList>
 void createLuaTableFromStringList(lua_State* state, const StringList& list) {
   lua_createtable(state, list.size(), 0);
   for (size_t i = 0; i < list.size(); i++) {
-    lua_pushstring(state, list[i].c_str());
+    lua_pushlstring(state, list[i].c_str(), list[i].size());
```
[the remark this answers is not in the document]
agrawroh: Changed it to `.data()` as suggested. Thanks!

test/extensions/filters/http/lua/lua_filter_test.cc
agrawroh: Changed it to `reinterpret_cast<char const>(buffer)` instead.

test/extensions/filters/http/lua/lua_filter_test.cc
agrawroh: Changed it to hex instead. Thanks!

test/extensions/filters/http/lua/wrappers_test.cc
agrawroh: Changed it to hex instead. Thanks!

test/extensions/filters/http/lua/lua_filter_test.cc
```diff
@@ -1800,6 +1800,40 @@ TEST_F(LuaHttpFilterTest, GetRequestedServerName) {
...
 }
 
+// Verify that binary values could also be extracted from dynamicMetadata() in LUA filter.
+TEST_F(LuaHttpFilterTest, GetDynamicMetadataBinaryData) {
+  const std::string SCRIPT{R"EOF(
+    function envoy_on_request(request_handle)
+      local metadata = request_handle:streamInfo():dynamicMetadata():get("envoy.pp")
+      local bin_data = metadata["bin_data"]
+      local data_length = string.len(metadata["bin_data"])
+      for idx = 1, data_length do
```
fedya-at-db, line 1810: I posted a comment but can't find it, so sorry if this shows up as a duplicate:

Make your Lua something like:
bin_data = "he\0llo"
escaped = ""
for idx = 1, string.len(bin_data) do
  escaped = escaped .. string.format("\\x%02x", string.byte(bin_data, idx))
end

Then you can have a single EXPECT_CALL like:
// Hex values for "he\0llo"
EXPECT_CALL(*filter_, scriptLog(spdlog::level::trace, StrEq(R"Hex Data: \x68\x65\x00\x6c\x6c\x6f")));

fedya-at-db (dismissed):

mattklein123 (dismissed):
Great, thank you.

mattklein123 (approved):

PR #22098, opened 2022-07-11
fix coredump

dorisd0102:
Commit Message: fix the coredump when stream is nullptr, then return directly instead of still sendMessage
Additional Description: N/A
Testing: MT
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/grpc/typed_async_client.h
+++ b/source/common/grpc/typed_async_client.h
@@ -35,8 +35,13 @@ template <typename Request> class AsyncStream /* : public RawAsyncStream */ {
 public:
   AsyncStream() = default;
   AsyncStream(RawAsyncStream* stream) : stream_(stream) {}
-  void sendMessage(const Protobuf::Message& request, bool end_stream) {
-    Internal::sendMessageUntyped(stream_, std::move(request), end_stream);
+  bool sendMessage(const Protobuf::Message& request, bool end_stream) {
+    if (stream_ == nullptr) {
+      return false;
+    } else {
+      Internal::sendMessageUntyped(stream_, std::move(request), end_stream);
+      return true;
+    }
   }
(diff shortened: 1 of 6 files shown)


dorisd0102:
@durd07  Please help to review this change. Thanks.

durd07 (approved):
LGTM

phlax:
@adisuissa would you mind reassigning this as its not really my area

source/common/grpc/typed_async_client.h
```diff
@@ -35,8 +35,13 @@ template <typename Request> class AsyncStream /* : public RawAsyncStream */ {
 public:
   AsyncStream() = default;
   AsyncStream(RawAsyncStream* stream) : stream_(stream) {}
-  void sendMessage(const Protobuf::Message& request, bool end_stream) {
-    Internal::sendMessageUntyped(stream_, std::move(request), end_stream);
+  bool sendMessage(const Protobuf::Message& request, bool end_stream) {
+    if (stream_ == nullptr) {
+      return false;
+    } else {
```
adisuissa, line 41: This feels like the wrong solution in this case (one red flag - why isn't the same check being performed for the other methods, e.g., in `resetStream()`)?
Can the AsyncStream be created using the second c'tor?

adisuissa:
Also can you add some tests to validate the behavior of the different changes?

source/common/grpc/typed_async_client.h
dorisd0102: @adisuissa , I check the nullptr in calling part and restore this code.

mattklein123 (approved):

PR #22103, opened 2022-07-11
repo: Ignore dot files by default (+some code check ignores)

phlax:
Ignore dot files in .gitignore and .bazelignore, unless
explicitly included.

Also ignore dot files in code check script

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


.bazelignore
```diff
@@ -1,3 +1,26 @@
...
+!\.flake8
+!\.gitattributes
+!\.github
+!\.gitignore
+!\.python-version
+!\.style.yapf
+!\.vscode
+!\.yapfignore
+!\.zuul
+!\.zuul.yaml
```
adisuissa, line 22: Is this needed?
Does ".bazelignore" support the "!" notation? (the closest thing to doc I could find is this: https://github.com/bazelbuild/bazel/blob/96c8a9073807c9e97635ddafe2ed0365a9318d6f/site/en/docs/bazelrc.md#other-files-governing-bazels-behavior-bazel-behavior-files and it didn't say anything)

adisuissa (commented):
Thanks!

.bazelignore
phlax: i also cant find any docs - but tbh that is not uncommon with bazel - and in this case trying to search brings up the wrong things - ie `.bazelignore exclude !` doesnt do it

i can set up a minimal repro if you like to confirm - im pretty sure i tested it when i added, but can check again

phlax: my bad - testing this out further, this doesnt do what i had hoped

i had read a conversation somewhere about keeping .bazelignore functionality aligned with .gitignore and incorrectly assumed that it had some equivalence - it does not

afaict, what it actually does is only work with directories, no globs, and no file exclusions

part of the problem i was addressing has been resolved by adding better file collection on the docs build

i still have a massive `.project` folder that i would like to exclude tho, if i exclude that from bazel it makes a massive difference

i think we want the .gitignores in this PR - ill update .bazelignore to just exclude my problem directory - others can use a similar convention or we can evolve the strategy

phlax:
@adisuissa i have cleaned this up a bit - i think the remaining changes are pretty non-controversial

would be great to land, as atm im rebasing other PRs against it so i can test locally - without it a check that takes 2mins takes 20mins

apologies for rebase/force - i messed up locally while working on it

adisuissa (approved):
LGTM, thanks!

PR #22108, opened 2022-07-11
check_format: Add fail_on_diff + small cleanups

phlax:
This speeds up the check_format in ci/format/format_pre job by optimizing the fix/check algo and adding a `fail_on_diff` flag

This allows CI to call `--fix` directly without having to call check, and then fix

The algo for checking/fixing itself is also slightly improved

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
cc @mattklein123 re conv in dev channel, this should be a pretty big improvement over what was there previously

i have thinned out the (~non-ci) runs in the format_pre elsewhere so with these changes applied perf should be better in all situations and running more things

phlax:
actually to qualify this it is probs still slower if no failure as it does more things

i know of several fairly immediate optis, so ill follow up

ci/format_pre.sh
```diff
@@ -57,14 +57,8 @@ if [[ -n "$AZP_BRANCH" ]]; then
...
 
-fix_format () {
-    echo "Fixing format..."
-    "${ENVOY_SRCDIR}"/tools/code_format/check_format.py fix
-    return 1
-}
-
 CURRENT=check_format
-"${ENVOY_SRCDIR}"/tools/code_format/check_format.py check || fix_format
+"${ENVOY_SRCDIR}"/tools/code_format/check_format.py fix
```
mattklein123, line 61: Don't you need to specify fail on diff here? Can you make a format error and double check that CI fails? Thank you.

/wait-any

phlax: good spot, apologies, fixed

mattklein123:
It's fine I just want to make sure that it works in CI. If you are sure it's good that's fine with me otherwise a test can't hurt.

phlax:
ill land and monitor ci - im pretty confident all is good

PR #22110, opened 2022-07-11
check_format: Dont run on api proto files

phlax:
This already happens in proto_format.sh

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


tools/code_format/check_format.py
```diff
@@ -887,11 +887,14 @@ def fix_source_path(self, file_path):
...
         if not file_path.endswith(PROTO_SUFFIX):
             error_messages += self.fix_header_order(file_path)
-        error_messages += self.clang_format(file_path)
-        if file_path.endswith(PROTO_SUFFIX) and self.is_api_file(file_path):
-            package_name, error_message = self.package_name_for_proto(file_path)
-            if package_name is None:
-                error_messages += error_message
+            error_messages += self.clang_format(file_path)
+        else:
+            if self.is_api_file(file_path):
```
phlax, line 892: filter these earlier

htuch (commented):
What checks or format fixes are we losing here?

phlax:
> What checks or format fixes are we losing here?

overall nothing - in this particular checker it wont run clang-tidy on the api proto files, as that is already done as part of proto_format

tools/code_format/check_format.py
```diff
@@ -356,17 +358,6 @@ def check_namespace(self, file_path):
...
         return []
 
-    def package_name_for_proto(self, file_path):
-        package_name = None
-        error_message = []
-        result = PROTO_PACKAGE_REGEX.search(self.read_file(file_path))
-        if result is not None and len(result.groups()) == 1:
-            package_name = result.group(1)
-        if package_name is None:
-            error_message = ["Unable to find package name for proto file: %s" % file_path]
```
phlax, line 366: i belive that proto_format also effectively checks the package name, but i will check this

phlax:
i need to main merge - but checking also for any differences

/wait

tools/code_format/check_format.py
```diff
@@ -888,10 +879,6 @@ def fix_source_path(self, file_path):
         if not file_path.endswith(PROTO_SUFFIX):
             error_messages += self.fix_header_order(file_path)
         error_messages += self.clang_format(file_path)
-        if file_path.endswith(PROTO_SUFFIX) and self.is_api_file(file_path):
-            package_name, error_message = self.package_name_for_proto(file_path)
-            if package_name is None:
```
phlax, line 893: these lines are also wrong - it just adds the same error message a second time, incorrectly

phlax:
conclusion from testing - the current checker ~doubles the time to run the ci/check_format script and does not work as expected

albeit also not with the ideal fix, proto_format certainly does detect this issue - or to be more precise, any build targets that use the apis - proto_format/docs/compiling/tests etc would fail pretty quickly in the presence of the error that the removed code would detect

tldr i think we want to land this asap

~related aside i noticed during my testing that one of the checker/fixers was trying to replace ` `` ` with ` * ` - not sure why that would be, but i will check this now

phlax:
> ... i noticed during my testing that one of the checker/fixers was trying ...

checked this and unable to repro - i think it was something to do with cache

phlax:
testing this again for timings rebased-to-main <> main  - it doesnt save quite as much as i had thought - more like 10-20%

htuch (approved):
Thanks for checking.

sergiitk:
Is it just me, or `check_format_test_helper.sh` now fails because it doesn't detect a proto issue because check_format is no longer checking protos?

```
❯ ./tools/code_format/check_format_test_helper.sh --log=WARN
api/missing_package.proto: Expected failure `Unable to find package name for proto file: ./api/missing_package.proto`, but succeeded
1 FAILURES
```

phlax:
> Is it just me, or check_format_test_helper.sh now fails

not sure - running this locally i hit a different problem - that it expects `clang-format-11` to be available - on resolving this the test passes

i believe this is running in CI also (altho im going to check that again)

im curious as to what my be causing your issue

its also clear that that test file needs some work

phlax:
my bad - i wasnt testing against `main` locally and ci is not currently running this - i will try and address this in #22426

sergiitk:
Thank you for looking into this!

Sidenote: I'm running `clang-format-14`, I think this is the requirement right now. IMO what would be really really useful, is CI printing out what clang version is being used—would make debugging formatting discrepancies much easier.

phlax:
yep, clang-format can be pita for this script - the version changes and is often not compatible with what are available in OS platforms

i would like to get this job into bazel and perhaps use some kinda bazel toolchain, but that is still a little way off

if you want to propose a PR to just print out what the script has found i would be happy to review

Issue #22009, opened 2022-07-04
The oauth2_filter should use client_secret_basic to access token endpoint by default.


figroc:
*Title*:

The oauth2_filter should use client_secret_basic to access token endpoint by default.

*Description*:

Currently the oauth2_filter use client_secret_post as the default authentication method to access oauth2 token endpoint. By RFC6749 and RFC8414, client_secret_basic is the default authentication method. And the client_secret_post method is not recommended and should be limited.

wbpcode:
cc @snowp

Amila-Rukshan:
I'm working on this ticket. going to support Basic auth as the default method and keep the existing method (sending credentials as url encoded body values) to set under a config.
```
config:
  ...
  auth_type: "URL_ENCODED_BODY". // default to "BASIC_AUTH"
  ...
```

Issue #21851, opened 2022-06-23
proposal: add a filter for injecting credentials into outgoing HTTP requests


yskopets:
*Title*: Add a filter for injecting credentials into outgoing HTTP requests

*Description*:

It would be conventient to have a standard filter that can inject credentials into outgoing HTTP requests (as a value of `Authorization` header).

The most common use cases:

1. OAuth2 access token credential
2. basic auth credential
3. opaque bearer token credential

The primary focus of this proposal is on injecting OAuth2 access token credential.

*Proposal*:

Add an HTTP filter with the following configuration model:

```yaml
- name: envoy.filters.network.http_connection_manager
  typed_config:
    "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
    http_filters:
    - name: envoy.filters.http.credentials
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.filters.http.credentials.v3alpha.Injector
        config:
          rules:
          - match:
              ... # HTTP requests to match
            inject:
              credential: { ... } # credential to inject
```

If the list of rules is empty, the filter will have no effect.

With regards to OAuth2 support:
* the filter will allow a user to specify `client_id` and `client_password`
* and let filter to acquire OAuth2 access token through [Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) flow
* the filter will also take care of refreshing access token

*Usage examples*:

#### Injecting OAuth2 access token

```yaml
- name: envoy.filters.network.http_connection_manager
  typed_config:
    "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
    http_filters:
    - name: envoy.filters.http.credentials
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.filters.http.credentials.v3alpha.Injector
        config:
          rules:
          - match:
              prefix: /
            inject:
              credential:
                oauth2:
                  token_endpoint:
                    cluster: oauth
                    uri: oauth.com/token
                    timeout: 3s
                  client_credentials:
                    client_id:
                      secret:
                        name: client-id
                        sds_config:
                          path: "/var/run/secret/credentials/oauth2/client-id.yaml"
                    client_password:
                      secret:
                        name: client-password
                        sds_config:
                          path: "/var/run/secret/credentials/oauth2/client-password.yaml"
                  # (Optional)
                  scopes:
                  - "example"
```

[...]

PR #22122, opened 2022-07-12
oauth: support authentication methods for client to oauth server requests

Amila-Rukshan:
Commit Message: 
oauth: support authentication methods for client for auth server requests

Additional Description:
Added a new config property under `oauth` filter to define how the `client_id` and `client_secret` will be sent in requests to OAuth server.
```
config:
  ...
  auth_type: "BASIC_AUTH" // defaults to "URL_ENCODED_BODY"
  ...
```
Risk Level:
Testing:
Tests added

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]

Fixes: #22009 

[Optional Deprecated:]


htuch (commented):
@snowp WDYT about this? Seems reasonable (even given the dangers of basic auth, if the user is aware of them and the environment is safe). But, there is some nagging thought that we should have some standard way to express HTTP fetch auth etc. for sidecalls. Thoughts?

htuch:
@Amila-Rukshan were you intending for this to be a breaking change, i.e. to start doing basic auth by default? By setting the enum 0 value to this, you create a new default behavior.

@snowp thanks, makes sense. I'm thinking basically anytime we make an HTTP call that a variety of auth methods would be plausibly useful, everything from stats dumps to xDS fetch over HTTP. The said, the fact this is the first time this has come up means this is probably not super useful.

I just noticed https://github.com/envoyproxy/envoy/issues/21851. If we were to implement this we could just leverage it with the OAuth filter making outbound HTTP calls that pick this up.

Amila-Rukshan:
@htuch, yes I wanted to change the existing default method to BASIC_AUTH as it is recommended in the RFC spec. But given that there are people using this filter will have effect on that change, shall I update the PR with existing auth method URL_ENCODED_BODY as the default?

rgs1:
cc: @fishcakez @JuniorHsu since Pinterest might be interested in any breaking behavior here.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -62,8 +62,17 @@ message OAuth2Credentials {
 
 // OAuth config
 //
-// [#next-free-field: 11]
+// [#next-free-field: 12]
 message OAuth2Config {
+  enum AuthType {
+    // The client_id and client_secret will be sent using HTTP Basic authentication scheme.
+    BASIC_AUTH = 0;
```
rgs1, line 69: following up on what @htuch said, we probably want the default value to be a no-op (e.g.: `NO_AUTH` or such).

rgs1 (changes requested):
Sounds like a nice thing to have, but I'd strongly suggest we make the current behavior the default one.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Amila-Rukshan: @rgs1 @htuch, I have made the existing behaviour as the default auth method (`URL_ENCODED_BODY`).

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -62,8 +62,17 @@ message OAuth2Credentials {
 
 // OAuth config
 //
-// [#next-free-field: 11]
+// [#next-free-field: 12]
 message OAuth2Config {
+  enum AuthType {
+    // The client_id and client_secret will be sent in the URL encoded request body.
```
htuch, line 68: The other alternative here is making thie DEFAUlT = 0 and allowing that to indicate "client specific default behavior" from the server. That way we could could safely change the Envoy defualt via runtime later, without breaking the wire protoocl. Not sure if there is a strong case for this, so this is just a suggestion.

htuch, line 68:
```suggestion 
    // The ``client_id`` and ``client_secret`` will be sent in the URL encoded request body.
```

htuch (commented):
/lgtm api

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Amila-Rukshan: @htuch, are you suggesting in the DEFAULT = 0 case we have another config where we have auth type defined for different auth providers. If that is the case we might need new "auth provider" field in the api do support it. Am I getting you right?

source/extensions/filters/http/oauth2/oauth_client.cc
```diff
@@ -21,21 +22,38 @@ namespace HttpFilters {
 namespace Oauth2 {
 
 namespace {
-constexpr const char* GetAccessTokenBodyFormatString =
+constexpr const char* GetAccessTokenBodyFormatString_01 =
```
snowp, line 25: Can you give these two a more explanatory name than 1/2?

source/extensions/filters/http/oauth2/oauth_client.cc
```diff
@@ -21,21 +22,38 @@ namespace HttpFilters {
...
+  std::string body;
+  switch (auth_type) {
+  case AuthType::UrlEncodedBody:
+    body = fmt::format(GetAccessTokenBodyFormatString_01, auth_code, encoded_client_id,
+                       encoded_secret, encoded_cb_url);
+    break;
+  case AuthType::BasicAuth:
+    const auto basic_auth_token = absl::StrCat(client_id, ":", secret);
+    const auto encoded_token = Base64::encode(basic_auth_token.data(), basic_auth_token.size());
+    const auto basic_auth_header_value = absl::StrCat("Basic ", encoded_token);
```
snowp, line 50: Are we able to pre-encode this early on the way we do for the other parameters? That would save us doing this on every request

snowp (changes requested):
At a high level this seems good to me, a few comments

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
htuch: No, just that we could add DEFAULT = 0 and let the implementatoin pick between 1 and 2.

Amila-Rukshan: @htuch, Now I got you. even though auth server supports both 1 and 2 if a request includes auth information in header and encoded body then auth server is supposed to fail (as a validation step). In that way we should avoid sending auth credentials in both types under a default config.

source/extensions/filters/http/oauth2/oauth_client.cc
```diff
@@ -37,20 +37,21 @@ void OAuth2ClientImpl::asyncGetAccessToken(const std::string& auth_code,
   const auto encoded_secret = Http::Utility::PercentEncoding::encode(secret, ":/=&?");
   const auto encoded_cb_url = Http::Utility::PercentEncoding::encode(cb_url, ":/=&?");
 
+  const auto basic_auth_token = absl::StrCat(client_id, ":", secret);
+  const auto encoded_token = Base64::encode(basic_auth_token.data(), basic_auth_token.size());
+  const auto basic_auth_header_value = absl::StrCat("Basic ", encoded_token);
+
```
snowp, line 43: Ah for some reason I thought we were pre-encoding the data here during filter creation. The way it is now this probably should be inside the switch case. Similarly, I think values that are only used by the UrlEncodedBody should be inside that case block, leaving only shared ones (encoded_cb_url, secret) on top. Sorry for the back and forth here!

Amila-Rukshan: @snowp updated the code. `encoded_cb_url` is common for both the cases

snowp (approved):
Thanks!

Issue #14030, opened 2020-11-16
Lua: in Lua filter, should require one of `inline_code` and `source_codes` , but not just require `inline_code`


nic-chen:
Hi, Community

For reusing codes, I am used to writing the code in a file and referencing it through `source_codes`:

```yaml
          http_filters:
          -name: entry.lua
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
              source_codes:
                entry.lua:
                  filename: /apisix/entry.lua
```
But it cause an error

```sh
 Proto constraint validation failed (LuaValidationError.InlineCode: ["value length must be at least " '\x01' " runes"]): source_codes {
   key: "entry.lua"
   value {
     filename: "/apisix/entry.lua"
   }
 }
```

To avoid this error, I had to add an empty function in `inline_code`

```yaml
          http_filters:
          - name: envoy.lua
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
              inline_code: |
                function envoy_on_request(request_handle)
                end
              source_codes:
                entry.lua:
                  filename: /apisix/entry.lua
```

So, I think it should require one of `inline_code` and `source_codes` , but not just require `inline_code`.

What do you think? Thanks.

dio:
cc. @wbpcode

nic-chen:
gentle ping @wbpcode

dio:
I think `inline_code` is required since it is the fallback, the global one. Before we have `source_codes` defined, we only have that `inline_code`. We need to test a couple of use-cases to catch if removing that global function will be fine.

nic-chen:
> I think `inline_code` is required since it is the fallback, the global one. Before we have `source_codes` defined, we only have that `inline_code`. We need to test a couple of use-cases to catch if removing that global function will be fine.

Thanks for the explanation.

I would like to try this.

wbpcode:
@nic-chen  This is designed to be compatible with old APIs.
I do want to remove the `inline_code`, but it is not desirable if it affects some existing users. However, I am in favor of deprecating it gradually.

wbpcode:
@nic-chen Sorry for delayed response. I just realized that this is an issue from ten days ago.

nic-chen:
@wbpcode 
Thanks for your reply and sharing your idea！

afontan:
I've been working on a use case that should have an empty inline_code section.
The end result of this is the following code:
[...]

PR #12516, opened 2020-08-06
support inline Lua code in per filter config

wbpcode:
Signed-off-by: wbpcode <comems@msn.com>

Commit Message:
Additional Description:
This is a supplement to PR #11235.  Check #12460 for more info.

Risk Level: Low
Testing: 
Docs Changes:
Release Notes:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

PR #11235, opened 2020-05-18
lua: Add per filter config for Lua filter

wbpcode:
Commit Message: Add per filter config support for Lua filter.

This allows Lua filter to support per-route configuration. This patch enables the configured Lua filter to have multiple registered codes that can be referenced from each per-route config. Disabling running the global Lua filter for a route is also supported.

Risk Level: Low
Testing: Added.
Docs Changes: Added.
Release Notes: Added.
Fixes: #9279  #3124

PR #22125, opened 2022-07-12
lua: deprecate inline_code and use default_source_code

wbpcode:
Commit Message: lua: deprecate inline_code and use default_source_code
Additional Description:

Lua source codes map and per route inline source code are supported after #11235 and #12516. However, the `inline_code` field is still necessary for the proto validation.
For users who needn't global/default script, they always need to set a `inline_code` with meaningless code (#14030). This PR will fix this problem.

**This PR will make the `inline_code` optional (remove the `string.min_len` validation) and deprecate it. The  `default_source_code`(of type  `core.v3.DataSource`) will be used to set default/global script.**

Risk Level: Low.
Testing: Updated.
Docs Changes: Updated.
Release Notes: Added.

### DIFF
--- a/api/envoy/extensions/filters/http/lua/v3/lua.proto
+++ b/api/envoy/extensions/filters/http/lua/v3/lua.proto
@@ -52,7 +52,7 @@ message Lua {
   //
   map<string, config.core.v3.DataSource> source_codes = 2;
 
-  // The default Lua code that Envoy will execute. If no any per route config is provided
+  // The default Lua code that Envoy will execute. If no per route config is provided
   // for the request, this Lua code will be applied.
   config.core.v3.DataSource default_source_code = 3;
 }
--- a/api/envoy/extensions/filters/http/lua/v3/lua.proto
+++ b/api/envoy/extensions/filters/http/lua/v3/lua.proto
@@ -4,6 +4,7 @@ package envoy.extensions.filters.http.lua.v3;
 
 import "envoy/config/core/v3/base.proto";
 
+import "envoy/annotations/deprecation.proto";
 import "udpa/annotations/status.proto";
 import "udpa/annotations/versioning.proto";
 import "validate/validate.proto";
@@ -26,7 +27,11 @@ message Lua {
   // further loads code from disk if desired. Note that if JSON configuration is used, the code must
   // be properly escaped. YAML configuration may be easier to read since YAML supports multi-line
   // strings so complex scripts can be easily expressed inline in the configuration.
-  string inline_code = 1 [(validate.rules).string = {min_len: 1}];
+  //
+  // This field is deprecated. Please use
+  // :ref:`default_source_code <envoy_v3_api_field_extensions.filters.http.lua.v3.Lua.default_source_code>`
+  string inline_code = 1
+      [deprecated = true, (envoy.annotations.deprecated_at_minor_version) = "3.0"];
 
   // Map of named Lua source codes that can be referenced in :ref:`LuaPerRoute
   // <envoy_v3_api_msg_extensions.filters.http.lua.v3.LuaPerRoute>`. The Lua source codes can be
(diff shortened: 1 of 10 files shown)


source/extensions/filters/http/lua/lua_filter.cc
```diff
@@ -690,9 +690,12 @@ FilterConfig::FilterConfig(const envoy::extensions::filters::http::lua::v3::Lua&
...
-  auto global_setup_ptr = std::make_unique<PerLuaCodeSetup>(proto_config.inline_code(), tls);
-  if (global_setup_ptr) {
-    per_lua_code_setups_map_[GLOBAL_SCRIPT_NAME] = std::move(global_setup_ptr);
+  if (proto_config.has_default_source_code()) {
+    const std::string code =
+        Config::DataSource::read(proto_config.default_source_code(), true, api);
+    default_lua_code_setup_ = std::make_unique<PerLuaCodeSetup>(code, tls);
+  } else if (!proto_config.inline_code().empty()) {
+    default_lua_code_setup_ = std::make_unique<PerLuaCodeSetup>(proto_config.inline_code(), tls);
   }
```
soulxu, line 700: Should we reject both `default_source_code` and `inline_code` specified? If both are specified, it should be a mistake by the user, feel better to tell the user explicitly.

api/envoy/extensions/filters/http/lua/v3/lua.proto
```diff
@@ -46,6 +51,10 @@ message Lua {
   //       filename: /etc/lua/world.lua
   //
   map<string, config.core.v3.DataSource> source_codes = 2;
+
+  // The default Lua code that Envoy will execute. If no any per route config is provided
```
mattklein123, line 55:
```suggestion
  // The default Lua code that Envoy will execute. If no per route config is provided
```

mattklein123 (changes requested):
Makes sense to me, thanks. Check CI?

/wait

api/envoy/extensions/filters/http/lua/v3/lua.proto
```diff
@@ -26,7 +27,11 @@ message Lua {
   // further loads code from disk if desired. Note that if JSON configuration is used, the code must
   // be properly escaped. YAML configuration may be easier to read since YAML supports multi-line
   // strings so complex scripts can be easily expressed inline in the configuration.
-  string inline_code = 1 [(validate.rules).string = {min_len: 1}];
+  //
+  // This field is deprecated. Please use
+  // :ref:`default_source_code <envoy_v3_api_field_extensions.filters.http.lua.v3.Lua.default_source_code>`
```
soulxu, line 32: better to document the conflict behavior when both inline_code and default_source code specified also

soulxu:
seems like you need a force push since this commits missed the sign off 040f164cf8585ceca621145a63138c34292bc6c5

api/envoy/extensions/filters/http/lua/v3/lua.proto
```diff
@@ -46,6 +54,10 @@ message Lua {
   //       filename: /etc/lua/world.lua
   //
   map<string, config.core.v3.DataSource> source_codes = 2;
+
+  // The default Lua code that Envoy will execute. If no per route config is provided
+  // for the request, this Lua code will be applied.
+  config.core.v3.DataSource default_source_code = 3;
```
htuch, line 60: /lgtm api

wbpcode:
> seems like you need a force push since this commits missed the sign off [040f164](https://github.com/envoyproxy/envoy/commit/040f164cf8585ceca621145a63138c34292bc6c5)

It seems that the `sign off and commit suggestion` will create a sign of `Co-authored-by: xxx` which will be reject by the DCO.
I marked it pass manually first.

soulxu (approved):
LGTM, thanks!

PR #22126, opened 2022-07-12
examples: Improve waits/fix flakes

### LINKED ISSUE #22113: Dynamic config examples can be flaky
Both of the dyanmic config examples can fail for a similar reason - essentially not updating quicly enough

not sure why they have both started failing now, but i think the best solution is to use a wait script

phlax:
Fix #22113

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


examples/ext_authz/verify.sh
```diff
@@ -15,7 +15,12 @@ responds_with_header \
 run_log "Restart front-envoy with FRONT_ENVOY_YAML=config/http-service.yaml"
 "$DOCKER_COMPOSE" down
 FRONT_ENVOY_YAML=config/http-service.yaml "$DOCKER_COMPOSE" up -d
-sleep 10
+
+wait_for 15 bash -c "\
+         responds_with_header \
+         'HTTP/1.1 200 OK' \
+         -H 'Authorization: Bearer token1' \
+         http://localhost:${PORT_PROXY}/service" || :
```
phlax, line 23: arrggghhh

examples/verify-common.sh
```diff
@@ -171,3 +174,11 @@ fi
 if [[ -z "$MANUAL" ]]; then
     bring_up_example
 fi
+
+
+# These allow the fun to be used in subshells, eg in `wait_for`
```
KBaichoo, line 179: s/fun/function/ 
s/eg/e.g./

examples/zipkin-tracing/verify.sh
```diff
@@ -13,9 +13,13 @@ responds_with \
     "Hello from behind Envoy (service 1)!" \
     "http://localhost:${PORT_PROXY}/trace/1"
 
-run_log "Test dashboard"
 # this could do with using the healthcheck and waiting
-sleep 20
+wait_for 30 bash -c "\
+         responds_with \
+         '<!doctype html>' \
+         http://localhost:${PORT_ADMIN}/zipkin/"
```
KBaichoo, line 20: Might be missing context, why are we  querying it twice with same query? Here and a few other places

phlax: the most reliable way of checking that the test is ready to be run is just waiting for the test to pass

ditto elsewhere

despite trying i didnt find a way to make this wait for docker correctly - i guess i could just move these to healthchecks on the compositions

my pref is to land and iterate because these make ci less flakey and faster

KBaichoo: iteration sgtm.

KBaichoo (approved):
Thanks!

Issue #10455, opened 2020-03-19
Upstream HTTP filters


snowp:
Today there is no way for an extension writer to access the individual requests that are sent by the router (or any other HTTP client). We should extend the API to support adding upstream HTTP filters: same interface as the existing HTTP filters, but they apply to outbound requests. 

Originally I envisioned these to be on the router (similar to how the upstream logs are defined there), but perhaps it makes the most sense to put them on the cluster level so that they can apply to any HTTP call made through Envoy? Thinking about calls made through the AsyncClient

wbpcode:
But clusters in envoy do not distinguish between protocols. In other words, a cluster may have redis/dubbo/thrift/http as its backend. If an http filter is added to the cluster, we need to make an additional distinction for each cluster.

snowp:
We already have HTTP specific configuration on the cluster, so I don't think this is introducing any new protocol awareness to the cluster. That said, it might make sense to make it a bit more generic, i.e. something like

```
message ProtocolFilters {
  repeated google.protobuf.Any filters = 1;
}

map<string, ProtocolFilters> upstream_protocol_fillters = 1;
```

so that we don't have to add more HTTP specific fields to the cluster. At that point other L7 filters could use the same field to configure this. Alternatively we could use the metadata field, but this might be generic enough that having it be its own field makes sense.

dio:
@snowp this is different from https://github.com/envoyproxy/envoy/blob/4403003b9c0888ce81606a2f45fcc1da126514e7/api/envoy/api/v2/cluster.proto#L776-L779? I can see there is one "interpretation" of it here: https://github.com/envoyproxy/envoy/issues/10525.

snowp:
Those are L4 filters applied to the upstream connection, so not protocol aware.

Augustyniak:
At Lyft, we would benefit from being able to access individual requests that are sent by the router too. I explained our use case in detail in [EnvoyMobile repository issue](https://github.com/lyft/envoy-mobile/issues/782) but Envoy repository may be a better place for an issue like this. 

TL;DR;
[...]

PR #22134, opened 2022-07-12
server: reworking factory contexts

alyssawilk:
This is a large and toil-y refactoring of FactoryContexts in preparation for upstream filters and ECDS. We went through several design options  (e.g. https://github.com/envoyproxy/envoy/pull/22061) and ended up with this, which is not trivial to update to, but provides a cleaner long term structure to factories and ownership.

We apologize for the downstream burden for folks pulling this PR.  In most cases you can
1) change createFilterFactoryFromProtoTyped to have Server::Configuration::FactoryContext& context
2) add a local member variable Server::Configuration::ServerFactoryContext& context = base_context.getServerFactoryContext();
3) update any calls to initManager or scope to use base_context.

(3) is especially important as it ensures that things work as expected for LDS, using the downstream init manager, and downstream stats scope.  It also makes this a very high risk PR - anything which will not persist beyond the downstream context must use the server context and the prior mixed inheritance model made it very unclear which was used where and when.

Risk Level: VERY HIGH
Testing: all the tests _finally_ pass
Docs Changes: n/a
Release Notes: 


envoy/server/factory_context.h
```diff
@@ -264,12 +269,44 @@ class FactoryContext : public virtual CommonFactoryContext {
...
+  FilterFactoryContext(ServerFactoryContext& server_context,
+                       OptRef<DownstreamFactoryContext> downstream_context)
+      : server_context_(server_context), downstream_context_(downstream_context) {}
+
+  /**
+   * @return ServerFactoryContext which lifetime is no shorter than the server.
+   */
+  ServerFactoryContext& getServerFactoryContext() { return server_context_; }
+
+  OptRef<DownstreamFactoryContext> getDownstreamFactoryContext() { return downstream_context_; }
```
alyssawilk, line 286: @kyessenov downstreamFactoryContext would be available for downstream filters, not for upstream filters.  Lifetime would be lifetime of the downstream listener

envoy/server/factory_context.h
```diff
@@ -264,12 +269,44 @@ class FactoryContext : public virtual CommonFactoryContext {
   virtual ProcessContextOptRef processContext() PURE;
 };
 
+/**
+ * Context for downstream network and HTTP filters.
+ */
```
mattklein123, line 274: Is this comment out of date or if it is actually the context for downstream network and http filters why does getDownstreamFactoryContext() return an OptRef? Shouldn't it always have that context?

mattklein123 (changes requested):
Thanks I'm a huge +1 on this approach. IMO it is vastly simpler to understand. Thank you for iterating on this. @kyessenov wdyt?

alyssawilk:
Ok having spent 2 more days swinging at this, I want to check in again.
I agree the abstractions are much more clean, but I will say that the API changes for folks implementing filters are super toily.
I managed to minimize the toil for HTTP filters, by changing the factories such that by default we use the ServerFactoryContext and don't have to getServerFactoryContext() left and right, but ironically because of the way network factories are already split out I couldn't do it there.

I've got test/integration and all extensions building, but just want to check in that we're OK with a more painful API change for the various features we're getting out of this in the long run before I go tackle tests again.

envoy/server/factory_context.h
```diff
@@ -246,56 +257,38 @@ class DownstreamFactoryContext {
...
     if (getDownstreamFactoryContext().has_value()) {
       return getDownstreamFactoryContext()->messageValidationVisitor();
     }
     return getServerFactoryContext().messageValidationVisitor();
   }
 
-private:
-  ServerFactoryContext& server_context_;
-  OptRef<DownstreamFactoryContext> downstream_context_;
+  operator ServerFactoryContext&() { return this->getServerFactoryContext(); }
```
mattklein123, line 291: I get why you did this, but pretty magical and I'm not sure we really want this?

mattklein123 (commented):
Thanks for slogging through this. I'm +1 on this approach even if it causes some build/integration pain for people. I think we can have some good release notes which should make it pretty clear what to do. I think this is going to make things much easier to understand and extend later, and also make the upstream stuff and potential sharing much cleaner. Thank you!

alyssawilk:
Ok, one full week plus some overtime and this is good to go.  All tests pass but the sip proxy test I filed https://github.com/envoyproxy/envoy/issues/22306 for.

Also please see updated PR description.  This PR now scares the heck out of me.  Tests caught 2 lifetime issues and I'm increasingly concerned about hard to find stats leak due to misplaced scope()

jmarantz:
I haven't looked at the 299 files yet :). Scared :)

Can you add some description to the PR about what motivated this change? Is it strictly https://github.com/envoyproxy/envoy/issues/10455 ?

alyssawilk:
@jmarantz "This is a large and toil-y refactoring of FactoryContexts in preparation for upstream filters and ECDS." not clear enough?
Originally this was so we could reuse shared code for upstream HTTP filters but it fixes some issues Kuat had with ECDS.  If it were just upstream filters I think I would have gone the template route by now.

RyanTheOptimist:
Any idea what's up with the failing tests? Are they flakes, or are they related to the changes in this PR?

alyssawilk:
/wait on CI
I think they're largely CI being overloaded because nothing is cached, but I want to land https://github.com/envoyproxy/envoy/pull/22301 and see if I can get away without this change.

alyssawilk:
OK, having thought about this some more (and trying to avoid the sunk cost fallacy) I'm REALLY leery of landing this PR.  
The lifetime issues I hit when debugging the final tests mean that not only is it super toily for users, it's also really high risk. 
Soooo, I went and tried another tack, and spent a day rebasing my upstream filters prototype (which now passes some of the inteegration tests!) with a template based solution
full PR complete with way too many FIXMES here:
https://github.com/envoyproxy/envoy/compare/main...alyssawilk:upstream_filters_clean?expand=1
the template changes are largely here:
https://github.com/envoyproxy/envoy/commit/53d466fac48e1697584b86addda678cca9114ff9

It's a lot uglier (god help whoever has to touch the config discovery templates next) but from buffer/config.* you can see the work to take a downstream factory and make it an upstream factory is really minimal, and having an opt-in system isn't that bad.

So unless @kyessenov or @mattklein123 actively want the filter refactor, I'm inclined to abandon this PR in favor of the template scariness (which is far less toil and risk.)

alyssawilk:
see https://github.com/envoyproxy/envoy/pull/22358 for a cleaner diff

//test/integration:integration test passes, but there might be a few other unit test fix ups I need to make for templates

PR #22135, opened 2022-07-12
rate_limit_quota: Quota based rate limiting, part 1

tyxia:
First part of quota-based rate limiting feature. The majority of this PR is to lay out the basic framework and shape of major components with unit test coverage, other functionalities (e.g., periodically usage report,  thread local cache, quota assignment, etc) will be added in next PRs. This PR includes:

- Rate limit quota filter basic framework: 
   Filter factory, filter creation and registration, filter configuration, etc.
- Bidirectional gRPC basic framework : 
   gRPC stream start and close, message send and receive, callback for the client and filter interaction, etc
- Request matching:  
   1) matcher tree creation 
   2) perform request matching with request header and exact `value_match` field from the config 
   3) `on_no_match` support (empty bucketId with `noassignmentBehavior` for now for illustration purpose, will be 
   modified in the next PRs)
- BucketId generation: 
   Generates the `BucketId` via either static method (`string_value`) or dynamic method (`custom_value` with 
   typed_config).
- Quota assignment caching:  
   Use`BucketId` as the key with customized hash/equal function. Will change to thread local storage 
- Unit tests for the config, filter and grpc client,
- Doc changes has been completed and merged in PR #22539.

Note to reviewers: 
In sprit of small PRs and making review easier, I plan to divide the whole implementation into several parts.  Hence, I 

- Removed some refer link in the proto for the ease of a boilerplate doc generation (this feature is WIP).  Also, I have left 
some TODOs that will be addressed in next PRs
- Temporarily reduced the code coverage requirement for this extension as this is first part of the several PR. Currently, it is ~75% coverage.

Signed-off-by: Tianyu Xia [tyxia@google.com](mailto:tyxia@google.com)


tyxia:
/unassign @mattklein123 

Initial commit that just adds back the docs as part of the implementation. Not ready for review yet.

tyxia:
/unassign @mattklein123

Initial commit that just adds back the docs as part of the implementation. Not ready for review yet.

tyxia:
/assign @yanavlasov 

Could you take a pass on this? Thanks!

wbpcode:
friendly ping @yanavlasov

api/envoy/extensions/filters/http/rate_limit_quota/v3/rate_limit_quota.proto
```diff
@@ -382,11 +377,10 @@ message RateLimitQuotaBucketSettings {
   // The interval at which the data plane (RLQS client) is to report quota usage for this bucket.
   //
   // When the first request is matched to a bucket with no assignment, the data plane is to report
-  // the request immediately in the :ref:`RateLimitQuotaUsageReports
```
bbassingthwaite, line 385: This might be out of the context of this PR but wasn't sure the best place to leave this. In our use case, the RQLS management server needs to know the Envoy node information in order to understand where the usage report is coming from. Should the first `RateLimitQuotaUsageReports` that Envoy sends contain the `Node` similar to other xDS APIs? Thanks!

tyxia: Thanks for the question!

I created an issue for this and we can discuss further over there. Resolving the conversation here....

source/extensions/filters/http/rate_limit_quota/client.h
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Async callbacks used during rateLimit() calls.
+ */
+class RateLimitQuotaCallbacks {
+public:
+  virtual ~RateLimitQuotaCallbacks() = default;
+
+  virtual void
+  onReceive(envoy::service::rate_limit_quota::v3::RateLimitQuotaResponse* response) PURE;
```
yanavlasov, line 24: Please use reference if you do not intend to take ownership of the `response`.

Also I would think that a callback to indicate that the stream was closed is also needed, so you can detect error conditions.

Maybe also make method name more descriptive, like `onQuotaResponse`

source/extensions/filters/http/rate_limit_quota/client_impl.cc
```diff
@@ -0,0 +1,63 @@
...
+#include "source/common/tracing/http_tracer_impl.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace RateLimitQuota {
+
+void RateLimitClientImpl::onRemoteClose(Grpc::Status::GrpcStatus, const std::string&) {
+  // TODO(tyxia) Add implementation later.
+  stream_closed_ = true;
```
yanavlasov, line 12: This will need to be wired into the client, so that stream closures can be retried

source/extensions/filters/http/rate_limit_quota/filter.cc
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// TODO(tyxia) Mostly are example/boilerplate code, polish implementation here.
+Http::FilterHeadersStatus RateLimitQuotaFilter::decodeHeaders(Http::RequestHeaderMap& headers,
+                                                              bool) {
+  // Start the stream on the first request.
+  auto start_stream = rate_limit_client_->startStream(callbacks_->streamInfo());
+  if (!start_stream.ok()) {
+    // TODO(tyxia) Consider adding the log.
+    return Envoy::Http::FilterHeadersStatus::Continue;
```
yanavlasov, line 22: Is this logic correct? Shouldn't it be subject to the no-assignment configuration?

source/extensions/filters/http/rate_limit_quota/filter.h
```diff
@@ -0,0 +1,145 @@
...
+  absl::StatusOr<BucketId> requestMatching(const Http::RequestHeaderMap& headers);
+
+  ~RateLimitQuotaFilter() override = default;
+
+private:
+  // Create the matcher factory and matcher.
+  void createMatcher();
+
+  FilterConfigConstSharedPtr config_;
+  RateLimitClientPtr rate_limit_client_;
```
yanavlasov, line 120: It does not seem to be conceptually correct. Filter should share the thread-local state (sort of cache) of quotas and this state includes the client that gets the bucket assignments.

yanavlasov (changes requested):
/wait

api/envoy/extensions/filters/http/rate_limit_quota/v3/rate_limit_quota.proto
```diff
@@ -25,15 +25,15 @@ option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/fil
 option (udpa.annotations.file_status).package_version_status = ACTIVE;
 option (xds.annotations.v3.file_status).work_in_progress = true;
 
-// [#not-implemented-hide:]
```
yanavlasov, line 28: Do you recall what check was failing that forced you to make these changes to the proto file? We should fix this check instead of having to put work-in-progress into published documentation.

source/extensions/filters/http/rate_limit_quota/client.h
tyxia: The caller site of this function which inherits from grcp async_client's [onReceiveMessage](https://github.com/envoyproxy/envoy/blob/35bf11a0f14a996d4e95b5302b0a953e0beba9f6/source/common/grpc/typed_async_client.h#L87) that has the smart pointer as function arg. So I just use the raw pointer here by simply via `get()`.

From the code [here](https://github.com/envoyproxy/envoy/blob/35bf11a0f14a996d4e95b5302b0a953e0beba9f6/source/common/grpc/typed_async_client.h#L97), this pointer will not be null. So you are right the reference is slightly better here as it can manifest its `non-nullness`.

Will update it

source/extensions/filters/http/rate_limit_quota/client_impl.cc
tyxia: Do you mean `stream_closed_`? It is currently in the client. Could you elaborate on this?

tyxia (commented):
Thanks for the review!

tyxia:
@yanavlasov  PTAL. Thanks!

I have addressed the review comments. The CI doc failure in this PR will be resolved once PR #22539 is merged (That PR is also under review now).

tyxia:
/retest

The CI failure should has nothing to do with this PR

phlax:
@tyxia i think the tsan problem has been fixed on `main` here #24252 

the other problem is coverage, which is saying:

```console
Code coverage 96.0 is lower than limit of 96.1
```

is it possible this fail is genuine ?

yanavlasov (approved):

yanavlasov (approved):

test/per_file_coverage.sh
```diff
@@ -53,6 +53,7 @@ declare -a KNOWN_LOW_COVERAGE=(
 "source/extensions/filters/http/lua:96.4"
 "source/extensions/filters/http/oauth2:96.0"
 "source/extensions/filters/http/wasm:1.9"
+"source/extensions/filters/http/rate_limit_quota:70" # TODO(tyxia) WIP. Improve coverage as development continues
```
alyssawilk, line 56: I believe this was part of brekaing coverage for main
I'm inclined to roll this and https://github.com/envoyproxy/envoy/pull/24304 back as this is new code.
@yanavlasov WDYT?

tyxia: I noticed the main coverage breakage before as well in my previous commits in this PR(from 96%.1 to 96%). Then I fixed it by increasing the coverage of this extension to  [83.1%](https://storage.googleapis.com/envoy-pr/0faa669/coverage/index.html) even though it is still listed as 70 in this check file.

So I think it should be fine if we are referring to same main coverage break. Please let me know if you have any questions @alyssawilk

tyxia: I have sent out a PR for your review and left the comment on open issue.

I will resolved the conversation here. And please let me know if you have any questions or requests on that PR or open issue. Thanks!

PR #22139, opened 2022-07-12
matchers: implement generic matching input container

kyessenov:

Commit Message: Add support for lists and integers to matching inputs.
Additional Description: Add a container to hold an input value that also supports integers and lists. This is needed for one-of semantics for some inputs (e.g. ALPN list) and more efficient for others (ports matching in ranges). In the future, a structured input can be added to support CEL contexts and protobuf values. The meaning of integers and lists is as follows:
- All current matchers treat integers as if they are strings for backwards compatibility.
- All matchers treat lists by checking each element one by one. This is not hard-coded, a matcher can choose to match lists in their entirety. 

Looking for feedback before completing this PR.

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


envoy/matcher/matcher.h
```diff
@@ -151,6 +153,95 @@ template <class DataType> class MatchTree {
...
+      return Kind::List;
+    default:
+      return Kind::Null;
+    }
+  }
+
+  bool isNull() const { return kind() == Kind::Null; }
+
+  /**
+   * Access the stored string value or throw an exception if the kind does not match.
```
snowp, line 190: Throwing an exception possibly on the data plan path seems icky, is there a way to avoid that?

kyessenov:
@snowp Right, the matchers decide how to handle lists. I think matching item-by-item makes sense for maps, for example.

zhxie:
I am thinking if we really need to add `InputValue`, maybe we can just use `std::any` instead? Then in the matcher we can use `type()` and `std::any_cast` to get the expected data or throw an exception if the type is not accepted.

kyessenov:
@zhxie I thought about it. I think it's harder to make code safe with `void *` since both inputs and matchers are extensible. Google style recommends avoid using `any` unless there's a really good fit, and in this case, strings are definitely the most common and benefit from `variant`. I think what we can do is to add another variant for what is essentially `any`. That way we can make code safe by ignoring `any` in most cases, unless the input and the matcher are somehow coupled (which I think will happen for CEL).

kyessenov:
@zhxie `any` has a problem with dis-allowing unique_ptrs since it requires copyability. That imposes limitations since it forces shared pointers for large objects. Not sure how to resolve that, but we can try with `any` for extensiblity purposes.

kyessenov:
Hmm, `any` is PITA to work with because of implicit conversion from any type to any. Would it be better to instead define some base class for "custom input output"? It might be easier for memory management as well if we have stronger expectations on the object.

zhxie:
Yes, I think you are right. `any` is a pain and it will mess up the code and something like `variant` will be good.
BTW, ProtoBuf struct is used to get expected fields in metadata matcher, and we should get it supported.

source/common/http/matching/inputs.h
```diff
@@ -26,17 +26,17 @@ class HttpHeadersDataInputBase : public Matcher::DataInput<HttpMatchingData> {
...
-      return {Matcher::DataInputGetResult::DataAvailability::NotAvailable, absl::nullopt};
+      return {Matcher::DataInputGetResult::DataAvailability::NotAvailable, Matcher::InputValue()};
     }
 
     auto header_string = HeaderUtility::getAllOfHeaderAsString(maybe_headers->get(), name_, ",");
 
     if (header_string.result()) {
       return {Matcher::DataInputGetResult::DataAvailability::AllDataAvailable,
-              std::string(header_string.result().value())};
+              Matcher::InputValue(header_string.result().value())};
```
vikaschoudhary16, line 36: should not we pass a list of values here? currently it is passing a comma separated list as a single string value?

PR #22155, opened 2022-07-13
original_dst_cluster: add override by filter state

kyessenov:
Commit Message: Add an override by filter state to ORIGINAL_DST cluster (which takes priority over header override and restored local address). 

Additional Description: ORIGINAL_DST cluster is used as a generic tunnel upstream for an internal listener (e.g. encoding TCP streams into HTTP/2). When used this way, there are two actual destination addresses. The first address is the logical address for which, for example, an RBAC policy must be applied on the internal listener. The second address is the physical address of the tunnel host. This change allows us to set the destination address for ORIGINAL_DST at the very last step in processing while doing the host selection, so that everything before that can use the logical address. [Prior art](https://github.com/envoyproxy/envoy/pull/21942) couples the same behavior with HTTP/1.1 CONNECT transport, which would be confusing to re-use for ORIGINAL_DST. 

Risk Level: low (opt-in)
Testing: unit
Docs Changes: none (filter states are not documented in general, maybe they should be).
Release Notes: none

### DIFF
--- a/source/common/network/BUILD
+++ b/source/common/network/BUILD
@@ -478,6 +478,17 @@ envoy_cc_library(
 )
 
 envoy_cc_library(
+    name = "filter_state_dst_address_lib",
+    srcs = ["filter_state_dst_address.cc"],
+    hdrs = ["filter_state_dst_address.h"],
+    deps = [
+        "//envoy/network:address_interface",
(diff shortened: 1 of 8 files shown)


mattklein123 (approved):
Thanks!

PR #22156, opened 2022-07-13
socket: allow restoring SSL info

kyessenov:

Commit Message: Permit changing SSL info in extensions and not just from the transport socket SSL info. 
Additional Description: This is needed when using internal sockets since they are raw_buffer but are upstream from TLS streams.
We would like to override TLS info on the internal connection in a custom extension.
Risk Level: low (removal of debug assertion)
Testing: regression
Docs Changes: none
Release Notes: none

### DIFF
--- a/envoy/network/socket.h
+++ b/envoy/network/socket.h
@@ -165,11 +165,6 @@ public:
   virtual void setSslConnection(const Ssl::ConnectionInfoConstSharedPtr& ssl_connection_info) PURE;
 
   /**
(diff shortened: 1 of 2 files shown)


kyessenov:
CC @alyssawilk as the original author of the assertion. Let me know if you want some more complicated mechanism (e.g. restore vs set like for the local address).

phlax:
@kyessenov looks like this was missed in the release - but should be good to land now

/wait

PR #22161, opened 2022-07-13
tooling: Preload CVE data for `dependency.check`

### LINKED ISSUE #22100: Downloading NIST data in dependency.checker can be flakey
currently we download the NIST data on every CI/dep.check run

this can be flakey as the data is huge and the service is not always reliable

im figuring there are a couple of things we can do to resolve

- accept filepath arguments in the dep.checker for NIST urls
- add a bazel job to download the data

as the data rarely changes this should speed things up a lot for the dep.checker

phlax:
Fix #22100

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/check-deps.yml
+++ b/.github/workflows/check-deps.yml
@@ -23,5 +23,7 @@ jobs:
       - name: Run dependency checker
         run: |
           export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
-          bazel run //tools/dependency:check -- -c release_issues --fix
-          bazel run //tools/dependency:check -- -c cves -w error
+          TODAY_DATE=$(date -u -I"date")
+          export TODAY_DATE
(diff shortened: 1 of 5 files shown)


phlax:
While this doesnt entirely fix #22100 (there is still a small risk of download issues once a day) it should reduce the issue to the point of very unlikely

if this crops up again - i can follow up with a retry algo - it may also help to stream the data in chunks to prevent async reactor locking up (if that is the actual issue)

phlax:
@moderation just flagging that it needs a dated action env to force daily expiry - otherwise the CVEs dont ever get updated

mattklein123 (approved):

PR #22170, opened 2022-07-13
docs: new parameter name for the bazel remote cache

mkauf:
docs: new parameter name for the bazel remote cache

The bazel parameter '--remote_http_cache' has been renamed to
'--remote_cache'. Use the new name.

Risk Level: Low

### DIFF
--- a/bazel/README.md
+++ b/bazel/README.md
@@ -1009,5 +1009,5 @@ slower cache performance on macOS due to slow disk performance on Docker for Mac
 Adding the following parameter to Bazel everytime or persist them in `.bazelrc`.
 
 ```
---remote_http_cache=http://127.0.0.1:28080/
+--remote_cache=http://127.0.0.1:28080/
 ```
--- a/ci/README.md
+++ b/ci/README.md
@@ -104,11 +104,11 @@ For a debug version of the Envoy binary you can run:
 The build artifact can be found in `/tmp/envoy-docker-build/envoy/source/exe/envoy-debug` (or wherever
 `$ENVOY_DOCKER_BUILD_DIR` points).
 
-To leverage a [bazel remote cache](https://github.com/envoyproxy/envoy/tree/main/bazel#advanced-caching-setup) add the http_remote_cache endpoint to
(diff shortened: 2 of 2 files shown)


phlax (commented):
@mkauf the flag change in bazel/README looks correct

regarding `BAZEL_REMOTE_CACHE` - is this automatically detected by bazel ?

im also wondering why not use the existing env var for this

cc @lizan

mkauf:
@phlax

> regarding `BAZEL_REMOTE_CACHE` - is this automatically detected by bazel ?

No, this environment variable is handled in `setup_cache.sh` (https://github.com/envoyproxy/envoy/blob/v1.22.2/ci/setup_cache.sh#L23)

phlax:
i see, but then im wondering why change it - iiuc it is just going to set the `BAZEL_BUILD_EXTRA_OPTIONS` and skip the undocumented envoy ci setup

with the original way it gives the end user more flexibility because they can also pass other args that way - no ?

phlax:
cool, lets keep the docs for `BAZEL_BUILD_EXTRA_OPTIONS` and adjust the docs for the RBE flags

bazel/README.md
```diff
@@ -1009,5 +1009,5 @@ slower cache performance on macOS due to slow disk performance on Docker for Mac
 Adding the following parameter to Bazel everytime or persist them in `.bazelrc`.
 
 ```
---remote_http_cache=http://127.0.0.1:28080/
+--remote_cache=http://127.0.0.1:28080/
```
phlax, line 1012: checking these docs futher - this is the docs for http cache - there are docs  futher up for remote_cache - i think we need to leave this as is

mkauf: The option "remote_http_cache" has been renamed to "remote_cache". We should use the new name.

https://github.com/bazelbuild/bazel/blob/5.2.0/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java#L114

phlax (commented):
@mkauf if you can rever the change to the extra_options env var then this should be good to land

mkauf:
I have updated the pull request.

phlax (approved):
great, thanks @mkauf , lgtm

PR #22174, opened 2022-07-13
docs: Cleanup more literals in API docs

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/config/bootstrap/v3/bootstrap.proto
+++ b/api/envoy/config/bootstrap/v3/bootstrap.proto
@@ -189,7 +189,7 @@ message Bootstrap {
 
   // Optional watchdog configuration.
   // This is for a single watchdog configuration for the entire system.
-  // Deprecated in favor of *watchdogs* which has finer granularity.
+  // Deprecated in favor of ``watchdogs`` which has finer granularity.
   Watchdog watchdog = 8
       [deprecated = true, (envoy.annotations.deprecated_at_minor_version) = "3.0"];
 
@@ -249,7 +249,7 @@ message Bootstrap {
   // when :ref:`dns_resolvers <envoy_v3_api_field_config.cluster.v3.Cluster.dns_resolvers>` and
   // :ref:`use_tcp_for_dns_lookups <envoy_v3_api_field_config.cluster.v3.Cluster.use_tcp_for_dns_lookups>` are
   // specified.
-  // This field is deprecated in favor of *dns_resolution_config*
+  // This field is deprecated in favor of ``dns_resolution_config``
   // which aggregates all of the DNS resolver configuration in a single message.
   bool use_tcp_for_dns_lookups = 20
       [deprecated = true, (envoy.annotations.deprecated_at_minor_version) = "3.0"];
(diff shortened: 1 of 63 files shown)


docs/root/configuration/http/http_filters/aws_lambda_filter.rst
```diff
@@ -22,7 +22,8 @@ If :ref:`payload_passthrough <envoy_v3_api_field_extensions.filters.http.aws_lam
 However, if :ref:`payload_passthrough <envoy_v3_api_field_extensions.filters.http.aws_lambda.v3.Config.payload_passthrough>`
 is set to ``false``, then the HTTP request is transformed to a JSON payload with the following schema:
 
-.. code-block::
+.. code-block:: json
+   :force:
```
phlax, line 26: requires css hack

api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -267,12 +267,12 @@ message Bootstrap {
...
-  // can be packed into this *typed_dns_resolver_config*. This configuration replaces the
+  // can be packed into this ``typed_dns_resolver_config``. This configuration replaces the
   // :ref:`dns_resolution_config <envoy_v3_api_field_config.bootstrap.v3.Bootstrap.dns_resolution_config>`
   // configuration.
-  // During the transition period when both *dns_resolution_config* and *typed_dns_resolver_config* exists,
-  // when *typed_dns_resolver_config* is in place, Envoy will use it and ignore *dns_resolution_config*.
-  // When *typed_dns_resolver_config* is missing, the default behavior is in place.
+  // During the transition period when both ``dns_resolution_config`` and ``typed_dns_resolver_config`` exists,
+  // when ``typed_dns_resolver_config`` is in place, Envoy will use it and ignore ``dns_resolution_config``.
+  // When ``typed_dns_resolver_config`` is missing, the default behavior is in place.
```
htuch, line 275: Is this covered by the RST warnings? Is there a style guide entry here?

docs/root/configuration/http/http_filters/aws_lambda_filter.rst
htuch: Can you unbundle the changes with interesting things from the `**` --> quotes mass change. Hard to spot what to pay attention to.

phlax: yep, that was kinda my intention

api/envoy/config/bootstrap/v3/bootstrap.proto
phlax: how could it - this is me editing prose to make words that should be literal, so - until we get AI in the pipeline this is not something we can test for

more generally - the use of `*` for literals relates to historic advice which i changed long ago

we could perhaps check for `*word*` and spellcheck the word as literals should not be marked up this way - it would catch some - ill look at it when i get to the spellcheck stuff

docs/root/configuration/http/http_filters/aws_lambda_filter.rst
phlax: code changes are here and here:

- #22198 
- #22197

api/envoy/config/bootstrap/v3/bootstrap.proto
adisuissa: I guess an alternative would be to clarify it in https://github.com/envoyproxy/envoy/blob/main/api/STYLE.md, but as long as we don't detect this in tooling, it will be hard to catch.
Can we hook a validation to the comments parser somehow?

phlax: this is what we do now for single backticks

for `*` or bare literals its much harder to detect - there are genuine uses of `*` - ie for italics/emphasis - and there is nothing to grep for with bare literals

spellchecking could work and catch most issues but not all and with some false +ves etc

this is something i intend to work, but not imminently - some preparatory work has been done over in pytooling tho

wrt to the STYLE guide i can update/review if you think it needs it but would suggest we follow up with that

phlax: also - mho is that the best style guide is what is there - the more we clean stuff up the more people are working from examples that exemplify what we want

adisuissa: > also - mho is that the best style guide is what is there - the more we clean stuff up the more people are working from examples that exemplify what we want

I also tend to follow other examples.

adisuissa (approved):
LGTM, thanks!
@htuch for a final say.

phlax:
one thing im thinking - we could do with a "literals" policy (myabe that should be `literals`)

my rule of thumb is "would this be localized" - but that doesnt always help

there are some grey areas, like HTTP or `HTTP` - which on the basis of it being more acronym than literal it probably should not be a literal - but im not sure if that is ever localized any other way (checking eg https://pontoon.mozilla.org/machinery/) 

tldr - i think we just need to agree some rules/priorites regarding literals and code blocks and put those in the style guide - but lets follow up with that

htuch (approved):
Thanks for the cleanup. I do think it would be ideal to add at least a style guide entry in a followup, since this is the kind of thing that will regress. Examples are powerful, but some guidance is even better :P

PR #22176, opened 2022-07-13
bazel: fix quiche compilation error on Android

jpsim:
`llround` is not constexpr on Android.

### DIFF
--- a/bazel/quiche.patch
+++ b/bazel/quiche.patch
@@ -1,0 +1,38 @@
+# llround is not constexpr on Android
+
+diff --git a/quiche/quic/core/quic_bandwidth.h b/quiche/quic/core/quic_bandwidth.h
+index b6572281..06cbaa30 100644
+--- a/quiche/quic/core/quic_bandwidth.h
++++ b/quiche/quic/core/quic_bandwidth.h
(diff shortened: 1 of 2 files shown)


jpsim:
> Should we add a TODO or file an issue somewhere to make sure don't lose track of this?

I can open a PR to apply this change directly in quiche if you think it's likely to be considered.

jpsim:
I opened https://github.com/google/quiche/pull/18, let's see what the quiche folks say.

moderation:
Definitely preferable fix upstream vs. carry a patch here. Agree with @RyanTheOptimist on adding a TODO so we can track to removal

RyanTheOptimist:
> Definitely preferable fix upstream vs. carry a patch here. Agree with @RyanTheOptimist on adding a TODO so we can track to removal

I should have mentioned that there is an upstream fix in flight. (though it will probably not land in envoy until next week)

jpsim:
I added a todo comment.

I'm glad to hear that an upstream fix is in flight, although I'd appreciate if we could merge this to unblock Envoy Mobile from being up to date in the meantime.

RyanTheOptimist (approved):
LGTM, but perhaps we should wait for https://github.com/envoyproxy/envoy/pull/22186 to land instead?

jpsim:
Closing in favor of https://github.com/envoyproxy/envoy/pull/22186

PR #22188, opened 2022-07-13
Explicitly initialize atomics

vitalybuka:
Before C++20 default constructor {} is trivial.
https://en.cppreference.com/w/cpp/atomic/atomic/atomic

It's a standard compliance nit-picking as I failed to reproduce uninitialized value.

Signed-off-by: Vitaly Buka <vitalybuka@google.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/access_log/access_log_manager_impl.h
+++ b/source/common/access_log/access_log_manager_impl.h
@@ -114,8 +114,8 @@ private:
                    // high performance. It is always local to the process.
   Thread::ThreadPtr flush_thread_;
   Thread::CondVar flush_event_;
-  std::atomic<bool> flush_thread_exit_{};
-  std::atomic<bool> reopen_file_{};
+  std::atomic<bool> flush_thread_exit_{false};
+  std::atomic<bool> reopen_file_{false};
   Buffer::OwnedImpl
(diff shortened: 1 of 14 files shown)


zhxie:
In fact, I think the standard means a `std::atomic<T> v` will not be initialized, but for `std::atomic<T> v{}`, it will be initialized using zero-initialization.

vitalybuka:
> In fact, I think the standard means a `std::atomic<T> v` will not be initialized, but for `std::atomic<T> v{}`, it will be initialized using zero-initialization.

pre C++20 zero initialization is only for static and thread-local
{} still calls default constructor https://godbolt.org/z/Yc4dW5oTa

vitalybuka:
> In fact, I think the standard means a `std::atomic<T> v` will not be initialized, but for `std::atomic<T> v{}`, it will be initialized using zero-initialization.

You are correct:
std::atomic<int> t; is default initialization, and it can be undefined https://en.cppreference.com/w/cpp/language/default_initialization
std::atomic<int> t{}; is a value initialization https://en.cppreference.com/w/cpp/language/value_initialization

PR #22212, opened 2022-07-14
http: Add content-based health check for upstream

botengyao:

Added the HTTP content-based health check for upstream.
- The health check should pass if the checker finds the expected string within the first `response_buffer_size` bytes of the HTTP response body. 
- There is still status code check after the response check.
- This is an optional feature for health check. Here is some [context](https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-protocol-http).

Commit Message: http: Add HTTP content-based health check for upstream
Additional Description:
Risk Level: M
Testing: fuzz & integration test
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


tonya11en (changes requested):
LG overall, but I had a few questions/comments. 

If there's a healthcheck benchmark, can you also post the before/after results? Also, can you add some more context and/or a pointer to an issue that contains the context in the PR description? Thanks!

test/common/upstream/health_check_fuzz.proto
```diff
@@ -10,6 +10,7 @@ import "google/protobuf/empty.proto";
 message HttpRespond {
   test.fuzz.Headers headers = 1;
   uint64 status = 2 [(validate.rules).uint64.lt = 1000];
+  optional string body = 3;
```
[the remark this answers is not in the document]
botengyao: Good catch! Done!

source/common/upstream/health_checker_impl.cc
```diff
@@ -347,9 +360,25 @@ void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::onGoAway(
 HttpHealthCheckerImpl::HttpActiveHealthCheckSession::HealthCheckResult
 HttpHealthCheckerImpl::HttpActiveHealthCheckSession::healthCheckResult() {
   const uint64_t response_code = Http::Utility::getResponseStatus(*response_headers_);
-  ENVOY_CONN_LOG(debug, "hc response={} health_flags={}", *client_, response_code,
+  ENVOY_CONN_LOG(debug, "hc response_code={} health_flags={}", *client_, response_code,
                  HostUtility::healthFlagsToString(*host_));
 
+  absl::string_view expected_response = parent_.expected_response_;
```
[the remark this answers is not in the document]
botengyao: Done!

source/common/upstream/health_checker_impl.cc
```diff
@@ -256,6 +261,14 @@ void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::decodeHeaders(
   }
 }
 
+void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::decodeData(Buffer::Instance& data,
+                                                                     bool end_stream) {
+  absl::StrAppend(&body_, data.toString());
+  if (end_stream) {
+    onResponseComplete();
+  }
+}
```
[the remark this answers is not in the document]
botengyao: Good point. Done.

source/common/upstream/health_checker_impl.cc
```diff
@@ -200,6 +204,7 @@ bool HttpHealthCheckerImpl::HttpStatusChecker::inRetriableRanges(uint64_t http_s
 }
 
 bool HttpHealthCheckerImpl::HttpStatusChecker::inExpectedRanges(uint64_t http_status) const {
+  ENVOY_LOG_MISC(info, "Boteng {}, expected {}", http_status, expected_ranges_);
```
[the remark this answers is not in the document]
botengyao: Yeah, deleted now.

botengyao:
> If there's a healthcheck benchmark, can you also post the before/after results? Also, can you add some more context and/or a pointer to an issue that contains the context in the PR description? Thanks!

This should be an optional feature. Google Cloud has this [content-based health check](https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-protocol-http) feature. Basically, we want to add the support for Envoy. There could be some benchmark in Google when the feature was added before. There should be some additional cost because we match the maximum 1024 bytes string.

phlax:
@botengyao - requires main merge

/wait

api/envoy/config/core/v3/health_check.proto
```diff
@@ -145,6 +145,11 @@ message HealthCheck {
     // <envoy_v3_api_msg_type.matcher.v3.StringMatcher>`. See the :ref:`architecture overview
     // <arch_overview_health_checking_identity>` for more information.
     type.matcher.v3.StringMatcher service_name_matcher = 11;
+
+    // An optional string to match anywhere in the first 1024 bytes of the response body.
+    // The response data can only be ASCII. If it is set, both the expected response check and status code determine the health check.
+    google.protobuf.StringValue response = 13
```
yanavlasov, line 151: I do not think these constraints are correct as the body can have arbitrary characters. Also the type should be `bytes`, not `string` as the protobuf `string` implies UTF-8 encoding.

source/common/upstream/health_checker_impl.cc
```diff
@@ -256,6 +260,16 @@ void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::decodeHeaders(
   }
 }
 
+void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::decodeData(Buffer::Instance& data,
+                                                                     bool end_stream) {
+  if (!parent_.expected_response_.empty()) {
+    absl::StrAppend(&body_, data.toString());
```
yanavlasov, line 266: We should accumulate up to 1Kb of data only here to avoid keeping large amounts of data if health response is large. Also you can make the `body_` to be a `Buffer` too, and move slices from `data` into `body_` to avoid copying. When you check for presence of the string you will need to call `linearize` on the `body_` first.

test/common/upstream/health_checker_impl_test.cc
```diff
@@ -600,7 +633,7 @@ class HttpHealthCheckerImplTest : public Event::TestUsingSimulatedTime,
   }
 
   void respond(size_t index, const std::string& code, bool conn_close, bool proxy_close = false,
-               bool body = false, bool trailers = false,
+               bool body = false, absl::string_view response_body = "", bool trailers = false,
```
yanavlasov, line 636: You can overload the `respond` method and add the `response_body` parameter to avoid changing every call site that does not need to set the `response_body`

test/common/upstream/health_checker_impl_test.cc
```diff
@@ -1025,6 +1058,130 @@ TEST_F(HttpHealthCheckerImplTest, SuccessWithMultipleHostSets) {
   EXPECT_EQ(Host::Health::Healthy, cluster_->prioritySet().getMockHostSet(1)->hosts_[0]->health());
 }
 
+TEST_F(HttpHealthCheckerImplTest, SuccessExpectedResponseCheck) {
```
yanavlasov, line 1061: Consider adding tests that send body in multiple small portions to validate that you concatenate it correctly in the health checker.

yanavlasov (changes requested):
/wait

api/envoy/config/core/v3/health_check.proto
botengyao: Thank you for reviewing! Relaxed the limitation, and changed to bytes.

source/common/upstream/health_checker_impl.cc
botengyao: Good point. Done! Thank you!

test/common/upstream/health_checker_impl_test.cc
botengyao: Done.

test/common/upstream/health_checker_impl_test.cc
botengyao: Create a new method to send body and send multiple small portions.

source/common/upstream/health_checker_impl.cc
```diff
@@ -347,9 +360,26 @@ void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::onGoAway(
...
 HttpHealthCheckerImpl::HttpActiveHealthCheckSession::healthCheckResult() {
   const uint64_t response_code = Http::Utility::getResponseStatus(*response_headers_);
-  ENVOY_CONN_LOG(debug, "hc response={} health_flags={}", *client_, response_code,
+  ENVOY_CONN_LOG(debug, "hc response_code={} health_flags={}", *client_, response_code,
                  HostUtility::healthFlagsToString(*host_));
 
+  if (!parent_.expected_response_.empty()) {
+    // If the expected response is set, check the first 1024 bytes of actual response if contains
+    // the expected response.
+    std::string response_data =
```
yanavlasov, line 369: Using std::string here causes the data to be copied out of the `response_body_` buffer. Using `absl::string_view` instead avoids this extra copy.

test/common/upstream/health_checker_impl_test.cc
```diff
@@ -454,6 +455,43 @@ class HttpHealthCheckerImplTest : public Event::TestUsingSimulatedTime,
...
+      path: /healthcheck
+      response: {0}
+    )EOF",
+                                   response);
+
+    allocHealthChecker(yaml);
+    addCompletionCallback();
+  }
+
+  void setupServiceHTTP2ValidationWithExpectedResponseHC() {
```
yanavlasov, line 477: Is this function used anywhere?

yanavlasov (changes requested):
/wait

source/common/upstream/health_checker_impl.cc
botengyao: Good point, thank you!

test/common/upstream/health_checker_impl_test.cc
botengyao: Yes, in a test case and addressed.

source/common/upstream/health_checker_impl.cc
```diff
@@ -256,13 +258,24 @@ void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::decodeHeaders(
   }
 }
 
+void HttpHealthCheckerImpl::HttpActiveHealthCheckSession::decodeData(Buffer::Instance& data,
+                                                                     bool end_stream) {
+  if (!parent_.expected_response_.empty() && response_body_->length() < 1024) {
```
yanavlasov, line 263: Sorry, missed this one. Please define a named constant for `1024`

yanavlasov (changes requested):
/wait

source/common/upstream/health_checker_impl.cc
botengyao: Thank you! Done.

yanavlasov (dismissed):

botengyao:
@envoyproxy/api-shepherd for anther look. Thank you!

api/envoy/config/core/v3/health_check.proto
```diff
@@ -145,6 +145,10 @@ message HealthCheck {
     // <envoy_v3_api_msg_type.matcher.v3.StringMatcher>`. See the :ref:`architecture overview
     // <arch_overview_health_checking_identity>` for more information.
     type.matcher.v3.StringMatcher service_name_matcher = 11;
+
+    // An optional string to match anywhere in the first 1024 bytes of the response body.
+    // If it is set, both the expected response check and status code determine the health check.
+    bytes response = 13 [(validate.rules).bytes = {max_len: 1024}];
```
adisuissa, line 151: Please move next to `expected_statuses`.

Also renaming to `response_prefix_contents` or `response_prefix_substring` may represent this field better.

adisuissa (commented):
Thanks for adding this feature!
Left an API comment.

api/envoy/config/core/v3/health_check.proto
botengyao: Thank you! Done.

api/envoy/config/core/v3/health_check.proto
```diff
@@ -145,6 +145,10 @@ message HealthCheck {
     // <envoy_v3_api_msg_type.matcher.v3.StringMatcher>`. See the :ref:`architecture overview
     // <arch_overview_health_checking_identity>` for more information.
     type.matcher.v3.StringMatcher service_name_matcher = 11;
+
+    // An optional string to match anywhere in the first 1024 bytes of the response body.
+    // If it is set, both the expected response check and status code determine the health check.
```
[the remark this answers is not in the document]
yanavlasov: Interesting. I think to make this generic, we can add BinaryMatcher similar to StringMatcher https://github.com/envoyproxy/envoy/blob/main/api/envoy/type/matcher/v3/string.proto 

We can also have `one of` here of BinaryMatcher and StringMatcher for convenience.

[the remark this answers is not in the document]
botengyao: Thank you! Can the generic job be a new PR from Api's point of view?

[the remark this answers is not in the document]
yanavlasov: It will require deprecation of the existing field and writing code that takes care of compatibility. It is more time efficient to do it right from the start, instead of pushing expected API changes to a later date.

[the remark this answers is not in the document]
botengyao: Got it, let me look into it. Thank you!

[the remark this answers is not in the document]
htuch: Sorry, I'm confused. Why doesn't this look more like the `Payload` message already used by `TcpHealthCheck`? That seems the place to try and harmonize.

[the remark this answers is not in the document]
botengyao: We need to choose one of [health checkers](https://github.com/envoyproxy/envoy/blob/1cf5603dc5239c92e5bc38ef321f59ccf6eabc6e/api/envoy/config/core/v3/health_check.proto#L288). here we mainly want to support HTTP for Google Cloud [feature](https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-protocol-http). 

Added a BinaryMacher, and make them generic.

changelogs/current.yaml
```diff
@@ -70,6 +70,6 @@ new_features:
     updated command operator ``%GRPC_STATUS%`` to suppoprt the snake case.
 - area: http
   change: |
-    added the support of :ref:`expected response string matcher <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_string_matcher>` and :ref:`expected response bytes matcher <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_bytes_matcher>`for HTTP health check.
+    added the support of :ref:`expected response string matcher <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_string_matcher>` and :ref:`expected response bytes matcher <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_bytes_matcher>` for HTTP health check.
```
phlax, line 73: if you split these lines it will make it easier to review

botengyao: Thank you! Done.

api/envoy/config/core/v3/health_check.proto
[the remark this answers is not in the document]
htuch: Not sure I follow - why can't we teach the `Payload` message, which is already in the business of binary matching, to do a better job for both TCP and HTTP body health check? This is the same problem I think and we should have similar levels of capability in both place.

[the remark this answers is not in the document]
botengyao: Good point, thank you for pointing it out! 

I think we can reuse the un-implemented [Payload](https://github.com/envoyproxy/envoy/blob/1cf5603dc5239c92e5bc38ef321f59ccf6eabc6e/api/envoy/config/core/v3/health_check.proto#L106-L107) in HTTP health check proto. And then create our own `HttpHealthCheckMatcher` class, and probably in the future we can add the `match_pattern` proto filed in [message HttpHealthCheck]() and TcpHealthCheck proto to extend.  The default behavior for HTTP health check is the contains pattern, and the default behavior for TCP is the Mongo HEX pattern for now.

We probably can make this [TcpHealthCheckMatcher](https://github.com/envoyproxy/envoy/blob/23ff23d0954eb9e6e4ed7ed60ec0e0d05174f6da/source/common/upstream/health_checker_impl.h#L194-L244) generic. Tcp health check payload match is using MongoDB health check and is only using HEX string, and doesn't implement bytes. 

How do you think? Thank you!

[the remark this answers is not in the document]
htuch: I think the idea of standardizing on Payload makes sense, and then we can extend Payload for any more sophisticated matching, on string or bytes.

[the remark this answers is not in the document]
botengyao: Hi all, can you review again? I used Payload api, and the match pattern can be extendable in future. Thank you!

api/envoy/config/core/v3/health_check.proto
```diff
@@ -163,6 +170,10 @@ message HealthCheck {
     // When checking the response, “fuzzy” matching is performed such that each
     // binary block must be found, and in the order specified, but not
     // necessarily contiguous.
+    //
+    // .. note::
+    //
+    //   The text payload should be Hex encoded payload. E.g., "000000FF", and binary is not implemented.
```
htuch, line 176: Why aren't thee constraints left on `Payload`? Aren't they true wherever `Payload` is used?

api/envoy/config/core/v3/health_check.proto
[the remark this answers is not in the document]
htuch: Yes, makes sense now just one docs comment, thanks.

api/envoy/config/core/v3/health_check.proto
botengyao: The health check for TCP is using repeated Hex string Payload which will be decoded to bytes. While for HTTP payload, users can specify one UTF-8 string payload as an expected response body for convenient purpose, or they can set the binary to follow the HTTP standard. The UTF-8 string is using by Google right-now.

htuch: Can we be consistent here and support the full range of options? It seems this is just a protobuf config utility thing and easy to support the same formats across both if the matcher is normalized internally?

botengyao: Good point. Yes, extended the Payload, and will not impact the current TCP health check usage.

botengyao: Emm.. looks like an enum is better, changed it. Thank you!

kyessenov:
/assign @adisuissa 
for finishing API review.

api/envoy/config/core/v3/health_check.proto
htuch: Looks fine to me now I think, thanks. I think if we did it from scratch we would use some standard generic matchers but given the Payload prior art in HCs this is the right way to go. Over to you @adisuissa.

api/envoy/config/core/v3/health_check.proto
```diff
@@ -69,15 +69,23 @@ message HealthCheck {
...
+    enum Decoder {
+      HEX = 0;
+      UTF8 = 1;
+    }
+
     oneof payload {
       option (validate.required) = true;
 
-      // Hex encoded payload. E.g., "000000FF".
+      // String payload. Default is Hex encoded payload, eg. "000000FF".
```
adisuissa, line 80:
```suggestion
      // String payload. Default is Hex encoded payload, e.g., "000000FF".
```

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +111,16 @@ message HealthCheck {
     // [#not-implemented-hide:] HTTP specific payload.
     Payload send = 3;
 
-    // [#not-implemented-hide:] HTTP specific response.
-    Payload receive = 4;
+    // HTTP specific expected response.
```
adisuissa, line 114: This is a repeated field, but IIUC the doc refers to it as a single entry.

api/envoy/config/core/v3/health_check.proto
```diff
@@ -69,15 +69,23 @@ message HealthCheck {
     option (udpa.annotations.versioning).previous_message_type =
         "envoy.api.v2.core.HealthCheck.Payload";
 
+    enum Decoder {
```
adisuissa, line 72: nit:
```suggestion
    enum DecoderType {
```

api/envoy/config/core/v3/health_check.proto
```diff
@@ -69,15 +69,23 @@ message HealthCheck {
...
-      // Hex encoded payload. E.g., "000000FF".
+      // String payload. Default is Hex encoded payload, eg. "000000FF".
       string text = 1 [(validate.rules).string = {min_len: 1}];
 
-      // [#not-implemented-hide:] Binary payload.
+      // Binary payload.
       bytes binary = 2;
     }
+
+    // Specify the text decode format. It doesn't impact on the binary field.
```
adisuissa, line 87:
```suggestion
    // Specify the text decode format. Only relevant if ``text`` payload is used. If not specified, text payload is assumed to be Hex encoded.
```

api/envoy/config/core/v3/health_check.proto
adisuissa: I was under the impression that if `min_len` validator is at least 1, this must have a config, no?

adisuissa (commented):
Thanks, overall looks good to me, left a few nits.

api/envoy/config/core/v3/health_check.proto
botengyao: Changed the doc. Thanks!

api/envoy/config/core/v3/health_check.proto
botengyao: Thank you, done!

api/envoy/config/core/v3/health_check.proto
botengyao: Not sure if I follow. Logically, it is fine that this is an empty string I think. But it is under oneof validate.required, so min_len looks redundant. There must be some config there.
```
http_health_check {
  path: "/healthcheck"
  receive {
  }
}
``` 
This will fail.

api/envoy/config/core/v3/health_check.proto
botengyao: Thank you, done!

api/envoy/config/core/v3/health_check.proto
```diff
@@ -69,15 +69,24 @@ message HealthCheck {
...
+      HEX = 0;
+      UTF8 = 1;
+    }
+
     oneof payload {
       option (validate.required) = true;
 
-      // Hex encoded payload. E.g., "000000FF".
-      string text = 1 [(validate.rules).string = {min_len: 1}];
+      // String payload. Default is Hex encoded payload, e.g., "000000FF".
```
adisuissa, line 80: I don't think this field should have a default value, but I may be misunderstanding something.
IIUC there should either be `text` (and if so it should have a minimum length of 1) or `binary`.
If `text` is provided, there isn't a default value. If `binary` is provided, the `text` part shouldn't be accessed, so it doesn't need to have a default value.

api/envoy/config/core/v3/health_check.proto
```diff
@@ -69,15 +69,24 @@ message HealthCheck {
     option (udpa.annotations.versioning).previous_message_type =
         "envoy.api.v2.core.HealthCheck.Payload";
 
+    enum DecoderType {
+      HEX = 0;
+      UTF8 = 1;
+    }
+
     oneof payload {
```
adisuissa, line 77: @htuch IMHO the use of Payload as text comparator is not so straight-forward. The `Payload` type is highly targeted towards binary comparisons. If one looks at the current state it seems that there are 2 modes: text and binary, but actually there should be 3 modes: text-plain, text-hex, binary.
Would it be better if we explicitly supply 3 states: text, binary, text-plain?

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +112,15 @@ message HealthCheck {
     // [#not-implemented-hide:] HTTP specific payload.
     Payload send = 3;
 
-    // [#not-implemented-hide:] HTTP specific response.
-    Payload receive = 4;
+    // Specifies a list of HTTP expected responses to match in the first 1024 bytes of the response body.
+    // If it is set, both the expected response check and status code determine the health check.
+    // When checking the response, “fuzzy” matching is performed such that each binary block must be found,
+    // and in the order specified, but not necessarily contiguous.
```
adisuissa, line 118: This only applies to binary blocks, but after this change there can be textual blocks (IIUC).
This line can probably be generalized as follows:
```suggestion
    // When checking the response, each payload must be matched,
    // and in the order specified, but not necessarily contiguous.
```

adisuissa (commented):
Thanks!

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +112,15 @@ message HealthCheck {
     // [#not-implemented-hide:] HTTP specific payload.
     Payload send = 3;
 
-    // [#not-implemented-hide:] HTTP specific response.
-    Payload receive = 4;
+    // Specifies a list of HTTP expected responses to match in the first 1024 bytes of the response body.
```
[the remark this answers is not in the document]
botengyao: For now, I think it is fine as 1024 default. It is mainly from Google's usage and from the efficiency point of view. We can make it configurable if we think it is necessary.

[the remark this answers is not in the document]
botengyao: Made it configurable. Done.

api/envoy/config/core/v3/health_check.proto
botengyao: There could also be Base64 encode text maybe in future, and so on. I am fine with ether the current approach or use multiple oneof text fields.

yanavlasov: I'm confused. Why do we need oneoff between bytes and string? bytes should be able to cover all use case where string may be used.

yanavlasov (commented):
/wait

api/envoy/config/core/v3/health_check.proto
botengyao: Thanks Yan, talked offline. It makes sense that UTF8 string can be actually carried by the bytes binary.  We can get rid of DecoderType. We will just keep the text HEX usage as TCP did before.  HTTP can use HEX text and bytes binary.

botengyao: Can you check the most recent changes? Thank you all!

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +103,18 @@ message HealthCheck {
...
+    // If it is set, both the expected response check and status code determine the health check.
+    // When checking the response, “fuzzy” matching is performed such that each payload block must be found,
+    // and in the order specified, but not necessarily contiguous.
+    //
+    // .. note::
+    //
+    //   Only the first 1024 bytes of the response body is used to match when ``response_buffer_size`` is not set.
+    repeated Payload receive = 4;
+
+    // Specifies the size of response buffer in bytes that is used to Payload match. Default value is 1024.
```
adisuissa, line 116: IMHO there should be a way to define an "infinite" buffer size, just in case some config wants to avoid setting a defined threshold.

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +103,18 @@ message HealthCheck {
...
+    uint64 response_buffer_size = 14 [(validate.rules).uint64 = {gte: 0}];
```
adisuissa, line 117: Per [API style-guide](https://github.com/envoyproxy/envoy/blob/main/api/STYLE.md), this should be a well-known-type (e.g., `google.protobuf.UInt32Value`).

adisuissa (commented):
Overall looks good, thanks!

api/envoy/config/core/v3/health_check.proto
botengyao: Good point. In that case, changing the default behavior to be infinite and getting rid of all 1024 definition will make more sense from user's perspective.

api/envoy/config/core/v3/health_check.proto
botengyao: I removed the default value. But in case future, changed it to the wrapper.

yanavlasov (dismissed):

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +103,20 @@ message HealthCheck {
...
-    Payload receive = 4;
+    // Specifies a list of HTTP expected responses to match in the first ``response_buffer_size`` bytes of the response body.
+    // If it is set, both the expected response check and status code determine the health check.
+    // When checking the response, “fuzzy” matching is performed such that each payload block must be found,
+    // and in the order specified, but not necessarily contiguous.
+    //
+    // .. note::
+    //
+    //   It is recommended to set ``response_buffer_size`` based on the Payload size for efficiency,
+    //   otherwise all response body will be recorded.
```
adisuissa, line 114:
```suggestion
    //   otherwise the matching will be performed against the entire response body.
```

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +103,20 @@ message HealthCheck {
...
+    repeated Payload receive = 4;
+
+    // Specifies the size of response buffer in bytes that is used to Payload match.
+    // All the response body will be used for matching if it is not set and ``receive`` is not empty.
```
adisuissa, line 118: @htuch the default value can be "safe" (e.g., set max-size to 1024), or it could be unbounded (to keep parity with tcp). WDYT?

@botengyao please add what will happen when the value is set to 0.

adisuissa (commented):
Thanks, api is almost there.

api/envoy/config/core/v3/health_check.proto
botengyao: Thank you! 

The current approach is if it is not set or 0 it will be unbounded. Also put a comment above to say it is recommended to set the buffer size.

Or we can set the default value is 1024 when it is not set, and to use unbound by setting to 0. 

Both should work.

htuch: Prefer sensible default bound as client-specific value like 1024 when not set. When set, use the set value, 0 as unlimited.

botengyao: Thank you, done!

api/envoy/config/core/v3/health_check.proto
```diff
@@ -103,8 +103,20 @@ message HealthCheck {
...
+    // and in the order specified, but not necessarily contiguous.
+    //
+    // .. note::
+    //
+    //   It is recommended to set ``response_buffer_size`` based on the total Payload size for efficiency.
+    //   The default buffer size is 1024 bytes when it is not set.
+    repeated Payload receive = 4;
+
+    // Specifies the size of response buffer in bytes that is used to Payload match.
+    // The default value is 1024 when not set. When set, use the set value, and 0 is as unlimited.
```
adisuissa, line 118:
```suggestion
    // The default value is 1024. Setting to 0 implies that the Payload will be matched against the entire response.
```

changelogs/current.yaml
```diff
@@ -95,5 +95,8 @@ new_features:
 - area: access_log
   change: |
     added support for number values in substitution format string in json_format.
+- area: http
+  change: |
+    added the support of :ref:`expected response <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.receive>` for HTTP health check.
```
adisuissa, line 100: This should be updated.

adisuissa (commented):
Thanks!
Left a couple of minor comments.
/wait

api/envoy/config/core/v3/health_check.proto
botengyao: Done.

changelogs/current.yaml
botengyao: Thank you, done!

changelogs/current.yaml
```diff
@@ -104,6 +104,10 @@ new_features:
 - area: access_log
   change: |
     added support for number values in substitution format string in json_format.
+- area: http
+  change: |
+    added the expected :ref:`receive <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.receive>` payload check for HTTP health check.
+    made the HTTP response buffer size configurable through :ref:`response_buffer_size <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_buffer_size>`.
```
adisuissa, line 110: nit:
```suggestion
    added :ref:`response_buffer_size <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_buffer_size>` to configure the maximal HTTP response buffer size.
```

adisuissa (commented):
One small nit, thanks!
/lgtm api

changelogs/current.yaml
```diff
@@ -104,6 +104,10 @@ new_features:
 - area: access_log
   change: |
     added support for number values in substitution format string in json_format.
+- area: http
+  change: |
+    added the expected :ref:`receive <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.receive>` payload check for HTTP health check.
+    added :ref:`response_buffer_size <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_buffer_size>` to configure the maximum HTTP response buffer size.
```
adisuissa, line 110: Apologies, one small nit after looking at the rendered docs (just realized that these are the same entry in the version history).
```suggestion
    added the expected :ref:`receive <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.receive>` payload check for HTTP health check.
    Added :ref:`response_buffer_size <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.response_buffer_size>` to configure the maximum HTTP health check response buffer size.
```

yanavlasov (approved):

PR #22214, opened 2022-07-14
Isolate pattern_template library changes to extensions

silverstar194:
Commit Message:
Isolate pattern_template extension and rename to pattern_template.

This is first PR to add separate extension and functionality for URL rewrites. 

Additional Description:
Ported code from internal fork.

Risk Level: Low. No impact on current functionality.

Testing:
Unit testing. See files.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


stevenzzzz:
when we talked about extension, I was thinking that we extend the RouteMatch to have a customizable match extension and a path-rewrite extension, which enables Envoy user to provide their own implementations for route-match and path-rewrite.

talked with Sean @silverstar194 , I realized that we are doing extension on the new field from RouteMatch::path_template, which will definitely work but seems less "extensible". 

It's not too late to extend RouteMatch::path_specifier tho since this PR is under review, and the required change is just a API change. 

@htuch thoughts?

stevenzzzz:
one thing I want to call out: kudos to @silverstar194 's flexibility.

htuch:
Yes, I was expecting a `TypedExtensionConfig` field to be in the `path_specifier oneof`..

silverstar194:
@htuch @stevenzzzz No worries. Will change to TypedExtensionConfig

silverstar194:
@stevenzzzz  @htuch I've changed the API to except the TypedExtensionConfig. Please review at your leisure. Will have the control plane and data planes changes in separate PR as before

silverstar194:
What are @htuch thoughts about the current API using two TypedExtensionConfig: one for match and one for action. I think we have two choices: 1. Put match pattern and rewrite under a single policy. 2. Create two extensions and polices (one for custom match and one for custom rewrite).

I think the downside of one unified policy being it seems strange to include a rewrite under a matching section.

The downside of two polices is added complexity and it's unclear how you'd validate match pattern with rewrite pattern. I.e. all rewrite variables are present in the matching pattern. Would be strange to have two extensions cross validating input to each-other.

That being said I have the two config/policy API in the PR now. I can change to one config/policy based on feedback...

htuch:
The split config LGTM - @mattklein123 @markdroth any thoughts?

silverstar194:
https://github.com/envoyproxy/envoy/pull/22326 Proto only change for easy review

PR #22244, opened 2022-07-16
tracer: fix the coredump when instances of skywalking type tracer wer…

zhouyujt:
…e destructed in TLS.

The reason for the crash is that the main thread posts function to each worker thread and does not wait for the worker thread to finish executing.  Let's assume that the main thread executes faster than the worker thread has had time to execute, and the main thread continues to execute, destructing the Envoy::Extensions::Tracers::SkyWalking::Driver class, and the tracing_stats_ member in the Driver class is also destructed.  At this point, if the worker thread starts executing, the reference to the tracing_stats_ member held by the worker thread is invalidated.

Signed-off-by: zhouyu <zhouyu-jt@163.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/tracers/skywalking/skywalking_stats.h
+++ b/source/extensions/tracers/skywalking/skywalking_stats.h
@@ -17,7 +17,7 @@ struct SkyWalkingTracerStats {
   SKYWALKING_TRACER_STATS(GENERATE_COUNTER_STRUCT)
 };
 
-using SkyWalkingTracerStatsPtr = std::shared_ptr<SkyWalkingTracerStats>;
+using SkyWalkingTracerStatsSharedPtr = std::shared_ptr<SkyWalkingTracerStats>;
 
 } // namespace SkyWalking
 } // namespace Tracers
--- a/source/extensions/tracers/skywalking/skywalking_stats.h
+++ b/source/extensions/tracers/skywalking/skywalking_stats.h
@@ -17,6 +17,8 @@ struct SkyWalkingTracerStats {
   SKYWALKING_TRACER_STATS(GENERATE_COUNTER_STRUCT)
 };
 
+using SkyWalkingTracerStatsPtr = std::shared_ptr<SkyWalkingTracerStats>;
+
 } // namespace SkyWalking
 } // namespace Tracers
 } // namespace Extensions
--- a/source/extensions/tracers/skywalking/skywalking_tracer_impl.cc
+++ b/source/extensions/tracers/skywalking/skywalking_tracer_impl.cc
@@ -28,8 +28,9 @@ using cpp2sky::TracerException;
 
 Driver::Driver(const envoy::config::trace::v3::SkyWalkingConfig& proto_config,
                Server::Configuration::TracerFactoryContext& context)
-    : tracing_stats_{SKYWALKING_TRACER_STATS(
-          POOL_COUNTER_PREFIX(context.serverFactoryContext().scope(), "tracing.skywalking."))},
+    : tracing_stats_(std::make_shared<SkyWalkingTracerStats>(
+          SkyWalkingTracerStats{SKYWALKING_TRACER_STATS(POOL_COUNTER_PREFIX(
+              context.serverFactoryContext().scope(), "tracing.skywalking."))})),
       tls_slot_ptr_(context.serverFactoryContext().threadLocal().allocateSlot()) {
   loadConfig(proto_config.client_config(), context.serverFactoryContext());
   tracing_context_factory_ = std::make_unique<TracingContextFactory>(config_);
--- a/source/extensions/tracers/skywalking/skywalking_tracer_impl.h
+++ b/source/extensions/tracers/skywalking/skywalking_tracer_impl.h
@@ -43,7 +43,7 @@ private:
   };
 
   TracerConfig config_;
-  SkyWalkingTracerStats tracing_stats_;
+  SkyWalkingTracerStatsPtr tracing_stats_;
   ThreadLocal::SlotPtr tls_slot_ptr_;
   std::unique_ptr<TracingContextFactory> tracing_context_factory_;
 };
--- a/source/extensions/tracers/skywalking/skywalking_tracer_impl.h
+++ b/source/extensions/tracers/skywalking/skywalking_tracer_impl.h
@@ -43,7 +43,7 @@ private:
   };
 
   TracerConfig config_;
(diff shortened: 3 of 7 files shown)


wbpcode:
Yeah, all the tracers have similar problems. Thanks for your fix. Could you check the CI?

zhouyujt:
> Yeah, all the tracers have similar problems. Thanks for your fix. Could you check the CI?

okay, i'm doing it now

source/extensions/tracers/skywalking/skywalking_tracer_impl.cc
```diff
@@ -28,8 +28,9 @@ using cpp2sky::TracerException;
 
 Driver::Driver(const envoy::config::trace::v3::SkyWalkingConfig& proto_config,
                Server::Configuration::TracerFactoryContext& context)
-    : tracing_stats_{SKYWALKING_TRACER_STATS(
-          POOL_COUNTER_PREFIX(context.serverFactoryContext().scope(), "tracing.skywalking."))},
+    : tracing_stats_(std::make_shared<SkyWalkingTracerStats>(
+          SkyWalkingTracerStats{SKYWALKING_TRACER_STATS(POOL_COUNTER_PREFIX(
+              context.serverFactoryContext().scope(), "tracing.skywalking."))})),
```
wbpcode, line 33: Thanks for your contribution. But will the stats is the only problem? If the Driver object is deconstructed before the worker callbacks are called, seems that all the access to the `this` are illegal operations.

wbpcode (commented):
Thanks. A comment is added.

source/extensions/tracers/skywalking/skywalking_tracer_impl.cc
zhouyujt: This does not happen because draining avoids this situation when the listener is modified via LDS.

wbpcode: To make it more clear, I mean the initialize callback of TLS slot will captures `this` and access it when running the callback. But because the problem you pointed out in this PR, may be the Driver objects have been deconstructed before the callback is called.

```
  tls_slot_ptr_->set([proto_config, &factory_context, this](Event::Dispatcher& dispatcher) {
    TracerPtr tracer = std::make_unique<Tracer>(std::make_unique<TraceSegmentReporter>(
        factory_context.clusterManager().grpcAsyncClientManager().factoryForGrpcService(
            proto_config.grpc_service(), factory_context.scope(), true),
        dispatcher, factory_context.api().randomGenerator(), tracing_stats_,
        config_.delayed_buffer_size(), config_.token()));
    return std::make_shared<TlsTracer>(std::move(tracer));
  });
```

zhouyujt: Sorry, maybe my PR comment is not detailed enough.the code that initializes the TLS slot callback is not the cause of the crash, it is fine.see the call stack from the crash scene below.
At the A tag, the main thread executes the removeSlot function(the code at the C tag) and posts messages to each worker threads without waiting for them to complete.  The Driver class is destructed when the call stack falls back to B, and the tracing_stats_ member in the Driver class is also destructed.  At this point, if the worker thread starts executing, the reference to the tracing_stats_ member held by the worker thread is invalidated.
image

wbpcode: I got your point. 
But I mean the crash is caused by the `use-after-free` and the stats is not the only struct that might be accessed after Driver objects are deconstructed.
Because these callbacks will be called in the worker threads asynchronously. It's possible that the deconstructing of Driver objects and calling of callbacks happen at same time. Then the accesses to both the `tracing_stats_` or `config_` may be illegal operations.

zhouyujt: I think it is true that only tracing_stats_ will illegal. The reasons are as follows: 
1. The return values of config_.delayed_buffer_size and config_. token are stored as values in TraceSegmentReporter object. 
2. In ListenerManagerImpl: : drainFilterChains have code like this 
```c++
        for (const auto& worker : workers_) {
          // Once the drain time has completed via the drain manager's timer, we tell the workers
          // to remove the filter chains.
          worker->removeFilterChains(
              draining_group->getDrainingListenerTag(), draining_group->getDrainingFilterChains(),
              [this, draining_group]() -> void {
                // The remove listener completion is called on the worker thread. We post back to
                // the main thread to avoid locking. This makes sure that we don't destroy the
                // listener while filters might still be using its context (stats, etc.).
                server_.dispatcher().post([this, draining_group]() -> void {
                  if (draining_group->decWorkersPendingRemoval() == 0) {
                    draining_group->getDrainingListener().debugLog(
                        absl::StrCat("draining filter chains from listener ",
                                     draining_group->getDrainingListener().name(), " complete"));
                    stats_.total_filter_chains_draining_.sub(
                        draining_group->numDrainingFilterChains());
                    draining_filter_chains_manager_.erase(draining_group);
                  }
                });
              });
        }
```
This ensures that the Driver object will not be destroyed before each worker thread completes the TLS initialization slot.

wbpcode: Thanks. It seems that you are right. Sorry I miss it. Thanks again for you patient explaination. 👍

source/extensions/tracers/skywalking/skywalking_stats.h
```diff
@@ -17,6 +17,8 @@ struct SkyWalkingTracerStats {
   SKYWALKING_TRACER_STATS(GENERATE_COUNTER_STRUCT)
 };
 
+using SkyWalkingTracerStatsPtr = std::shared_ptr<SkyWalkingTracerStats>;
```
wbpcode, line 20: Please use `SkyWalkingTracerStatsSharedPtr` here to ensure that the style is consistent with the community.

wbpcode (commented):
Thanks. It's great. Only one last comment.

source/extensions/tracers/skywalking/skywalking_stats.h
zhouyujt: ok,thanks.

wbpcode (approved):
LGTM. Thanks.

PR #22246, opened 2022-07-16
lua: add new methods to access network connection streamInfo() & dynamicMetadata()

agrawroh:
### Background

Currently, there seems to be no way to access the dynamic metadata set by any network filters in LUA or any other HTTP filters. `streamInfo()` is accessible on the connection and could be used to fetch the `dynamicMetadata()` from the network filters but it's is currently not exposed in HTTP LUA filter.

### Use Case

We want to read the TLV values set by the `proxy_protocol` filter in the dynamic metadata in the LUA filter so that we can propagate these TLV pairs to the upstream services for some further processing. 

### Changes

This PR adds a new `connectionStreamInfo()` methods to LUA filter that could be used to access the Stream Info object on the connection and then fetch the dynamic metadata from network filters like this:
```
connectionStreamInfo():dynamicMetadata():get("envoy.proxy_protocol")
```
---
**Commit Message:** add new methods to access network connection streamInfo() & dynamicMetadata().
**Risk Level:** Low
**Testing:** Unit Tests
**Docs Changes:** Added the description of the new methods added in LUA. 
**Release Notes:** Added

### DIFF
--- a/source/extensions/filters/http/lua/lua_filter.cc
+++ b/source/extensions/filters/http/lua/lua_filter.cc
@@ -134,6 +134,9 @@ PerLuaCodeSetup::PerLuaCodeSetup(const std::string& lua_code, ThreadLocal::SlotA
   lua_state_.registerType<DynamicMetadataMapIterator>();
   lua_state_.registerType<StreamHandleWrapper>();
   lua_state_.registerType<PublicKeyWrapper>();
+  lua_state_.registerType<ConnectionStreamInfoWrapper>();
+  lua_state_.registerType<ConnectionDynamicMetadataMapWrapper>();
+  lua_state_.registerType<ConnectionDynamicMetadataMapIterator>();
 
   const Filters::Common::Lua::InitializerList initializers(
       // EnvoyTimestampResolution "enum".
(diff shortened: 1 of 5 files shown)


agrawroh:
cc @ggreenway

dynajoe:
My team is interested in seeing this PR get through. We too need to get proxy protocol dynamic metadata at the HCM to forward to upstream. Is there an alternative?

source/extensions/filters/http/lua/lua_filter.cc
```diff
@@ -134,6 +134,9 @@ PerLuaCodeSetup::PerLuaCodeSetup(const std::string& lua_code, ThreadLocal::SlotA
   lua_state_.registerType<DynamicMetadataMapIterator>();
   lua_state_.registerType<StreamHandleWrapper>();
   lua_state_.registerType<PublicKeyWrapper>();
+  lua_state_.registerType<ConnectionStreamInfoWrapper>();
+  lua_state_.registerType<ConnectionDynamicMetadataMapWrapper>();
+  lua_state_.registerType<ConnectionDynamicMetadataMapIterator>();
```
wbpcode, line 139: should the stream info and dynamic metadata need independent type?

wbpcode (commented):
thanks for working on this. it make sense to me.

source/extensions/filters/http/lua/lua_filter.cc
agrawroh: That's a good question. I wasn't aware of any implications for re-using the one we have for HTTP-level stream info object so I created separate ones for the listener-level stream info. If you think it's fine to reuse the existing ones then we can remove it.

wbpcode: > That's a good question. I wasn't aware of any implications for re-using the one we have for HTTP-level stream info object so I created separate ones for the listener-level stream info. If you think it's fine to reuse the existing ones then we can remove it.

Yeah, I think it would be better to just reuse this wrapper.

wbpcode (commented):
It's LGTM overall, except I think it would be better to reuse the exist type wrapper. Thanks.

And sorry for the delayed review.

dynajoe:
@agrawroh what's the next step to enable access to metadata?

PR #22260, opened 2022-07-18
Make stream idle timeouts refresh with new routes.

paul-r-gall:

Commit Message: Make stream idle timeouts refresh with new routes.
Additional Description: This fixes an outstanding TODO in the HCM. Currently, if the route is reselected by the filter chain, and the new route has a longer idle timeout than the old route, the new idle timeout will not be respected. This fixes that bug.
Risk Level: Low
Testing: Integration tested.


yanavlasov:
I'm unsure why you need this complicated test. I think it should be possible to simply create two routes with different timeouts and use the already existing filter to simply select a different route.

/wait-any

paul-r-gall:
> use the already existing filter to simply select a different route.

@yanavlasov Not sure what you mean by this -- the pre-existing SetRouteFilter won't override anything except the cluster name

source/common/http/conn_manager_impl.cc
```diff
@@ -1088,30 +1088,6 @@ void ConnectionManagerImpl::ActiveStream::decodeHeaders(RequestHeaderMapPtr&& he
     // Allow non websocket requests to go through websocket enabled routes.
   }
 
-  if (hasCachedRoute()) {
```
adisuissa, line 1091: One of the differences I see here is that before this change, the timer was created after lines 1075-1089, which IIUC could send a local reply and not create the timer. In the new code the timer could be created (as part of the call from `refreshCachedRoute()` to `setRoute()`) and then a local-reply will be sent.
What will happen in this case if a timeout occurs?

adisuissa (commented):
Thanks.
It LG overall, just want to better understand the impact of this change.

source/common/http/conn_manager_impl.cc
paul-r-gall: Actually, a default `stream_idle_container_` was already created in the `ActiveStream` constructor. So the possibility of timing out during a local reply already existed.

PR #22270, opened 2022-07-18
Add gRPC status into LocalReplyData

TAOXUY:
Signed-off-by: Xuyang Tao <taoxuy@google.com>

Commit Message: 
- also add gRPC status into LocalReplyData so it can be accessed during `onLocalReply`.

Risk Level: now
Testing: unit test doc
Docs Changes: no need
Release Notes: no need
Platform Specific Features: no need
[Optional Runtime guard:] no need
[Optional Fixes #Issue]


TAOXUY:
@shuoyang2016 @qiwzhang @nareddyt

source/common/stream_info/stream_info_impl.h
```diff
@@ -174,12 +174,20 @@ struct StreamInfoImpl : public StreamInfo {
 
   absl::optional<uint32_t> responseCode() const override { return response_code_; }
 
+  absl::optional<Grpc::Status::GrpcStatus> localReplyGrpcStatus() const override {
```
qiwzhang, line 177: Since this is specific for "localReply",  are we only allowed it to be called by sendLocalReply()?

source/common/stream_info/stream_info_impl.h
```diff
@@ -174,12 +174,20 @@ struct StreamInfoImpl : public StreamInfo {
...
+    return local_reply_grpc_status_;
+  }
+
   const absl::optional<std::string>& responseCodeDetails() const override {
     return response_code_details_;
   }
 
   void setResponseCode(uint32_t code) override { response_code_ = code; }
 
+  void setLocalReplyGrpcStatus(Grpc::Status::GrpcStatus status) override {
```
qiwzhang, line 187: I don't see how it is called.

TAOXUY:
> The changes LGTM. Should any of the current filters be modified to call `setLocalReplyGrpcStatus`?

Done. It will be called by `FilterManager::sendLocalReply`.

source/common/stream_info/stream_info_impl.h
TAOXUY: Done. It will be called by FilterManager::sendLocalReply.

TAOXUY:
> The changes LGTM. Should any of the current filters be modified to call `setLocalReplyGrpcStatus`?

qiwzhang (approved):

nareddyt (dismissed):

shuoyang2016 (approved):

TAOXUY:
@phlax @cpakulski PTAL

cpakulski (dismissed):
LGTM.

phlax:
cc @alyssawilk for final stamp

source/common/http/filter_manager.cc
```diff
@@ -860,6 +860,10 @@ void FilterManager::sendLocalReply(
   }
 
   stream_info_.setResponseCodeDetails(details);
+  if (grpc_status.has_value()) {
```
alyssawilk, line 863: I'm likely missing something here, but why not add the status into the headers/trailers at this point, rather than adding it to stream into and then pulling it back out to add it to headers?

alyssawilk (commented):
thanks for picking this up!  If this results in a functional change to the data plane it should be runtime guardded, per CONTRIBUTING.md

/wait

source/common/http/filter_manager.cc
TAOXUY: As I mentioned in this PR description, it won't work for the case for HTTP downstream, gRPC upstream and using gRPC transcoder in envoy, where gRPC status cannot be passed in headers or trailers.

TAOXUY:
> thanks for picking this up! If this results in a functional change to the data plane it should be runtime guardded, per CONTRIBUTING.md
> 
> /wait

DONE. PTAL.

source/common/http/filter_manager.cc
alyssawilk: So the problem is you're setting the grpc_status in your sendLocalReply, it's being dropped because Envoy is sending a non-gRPC reply, but you want the local reply's grpc status in your access logs? 
I'm having a bit of a hard time justifying adding stream info size for all Envoy users to capture gRPC status for non-gRPC responses.  

Given that local replies will traverse your response filters, how about you have a filter which implements onLocalReply(const LocalReplyData&)   That'll let you capture the grpc_status which is sent in the local reply, insert it using the extensible fields in stream info.

alternately while Envoy isn't going to add a grpc-status for non-grpc responses there's nothing from stopping you from doing so which I believe would also solve the problem.

cc @snowp @lizan for second thoughts.
/wait

TAOXUY: In fact, gRPC status isn't in [LocalReplyData](https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L685-L693) used in `onLocalReply(const LocalReplyData&)`. I guess you mean this [one](https://github.com/envoyproxy/envoy/blob/main/source/common/http/utility.h#L369).

I am fine with this approach and I can add gRPC status into [LocalReplyData](https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L685-L693)  in this PR.

lizan: > I'm having a bit of a hard time justifying adding stream info size for all Envoy users to capture gRPC status for non-gRPC responses.

+1, I'm not clear either why do you want to capture this.

> alternately while Envoy isn't going to add a grpc-status for non-grpc responses there's nothing from stopping you from doing so which I believe would also solve the problem.

This also works, note that you can even emit trailers to HTTP/1.1 request and the codec will disregard them by default.

TAOXUY: Thanks guys! Can you please take another look?

RyanTheOptimist:
Looks like there is a format problem. Can you please fix?

https://dev.azure.com/cncf/envoy/_build/results?buildId=113971&view=logs&jobId=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39

ERROR: From ./source/common/runtime/runtime_features.cc
ERROR: RUNTIME_GUARD(envoy_reloadable_features_use_cannoical_grpc_status_in_local_reply_body_rewrite); and RUNTIME_GUARD(envoy_reloadable_features_validate_connect); are out of order

alyssawilk:
actually forgot Matt was out today.  @lizan can you take a look?

TAOXUY:
> actually forgot Matt was out today. @lizan can you take a look?

Great! Thanks!

lizan (approved):

Issue #22217, opened 2022-07-15
Can't generate compilation DB on M1 Mac


xjtian:
*Description*:

`/tools/vscode/refresh_compdb.sh` doesn't succeed on an M1 MBP due to toolchain selection failure. This happens if I run the script directly on the host and inside the devcontainer, although for different reasons. Unfortunately I don't have an Intel Mac to attempt a repro on.

Toolchain debug output on the *devcontainer*:

```
[output truncated for brevity]

INFO: ToolchainResolution: Target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Selected execution platform @envoy_build_tools//toolchains:rbe_linux_clang_platform, type @bazel_tools//tools/cpp:toolchain_type -> toolchain @envoy_build_tools//toolchains/configs/linux/clang/cc:cc-compiler-k8
INFO: ToolchainResolution:     Type @io_bazel_rules_go//go:toolchain: target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Rejected toolchain @go_sdk//:go_freebsd_386-impl; mismatching values: freebsd, x86_32
INFO: ToolchainResolution:     Type @io_bazel_rules_go//go:toolchain: target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Rejected toolchain @go_sdk//:go_freebsd_amd64-impl; mismatching values: freebsd
INFO: ToolchainResolution:     Type @io_bazel_rules_go//go:toolchain: target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Rejected toolchain @go_sdk//:go_freebsd_arm-impl; mismatching values: freebsd, arm
INFO: ToolchainResolution:     Type @io_bazel_rules_go//go:toolchain: target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Rejected toolchain @go_sdk//:go_freebsd_arm64-impl; mismatching values: freebsd, arm64
INFO: ToolchainResolution:     Type @io_bazel_rules_go//go:toolchain: target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Rejected toolchain @go_sdk//:go_illumos_amd64-impl; mismatching values: illumos
INFO: ToolchainResolution:     Type @io_bazel_rules_go//go:toolchain: target platform @envoy_build_tools//toolchains:rbe_linux_clang_platform: Rejected toolchain @go_sdk//:go_ios_amd64-impl; mismatching values: ios
[...]

PR #22272, opened 2022-07-18
bazel: disable uring with compilation database

keith:
This is a workaround for
https://github.com/envoyproxy/envoy/issues/22217 with the downside that
it disables it for all platforms.


phlax (approved):
lgtm, thanks @keith

PR #22273, opened 2022-07-18
bazel: add target_compatible_with to envoy_cc_library

keith:
This attribute allows us to specify what constraints a library is
compatible with, such as only usable on linux, or not windows etc. It
is viral in that any targets that explicitly depend on it up the tree
will respect those constraints, and not be built, unless they're
specified explicitly on the command line, in which case bazel will fail
instead. This helps us a fix a problem with the compilation database
script where it virtually does `bazel build //...`, but some libraries
don't build at all on some platforms. Theoretically we could also solve
this case by marking those libraries as `manual`, but that would have
other side effects in that they wouldn't be built on the platforms that
they support either (unless they were somehow otherwise part of the
dependency tree, since unlike this attribute, the `manual` tag is not
viral).

### DIFF
--- a/bazel/envoy_library.bzl
+++ b/bazel/envoy_library.bzl
@@ -101,7 +101,8 @@ def envoy_cc_library(
         strip_include_prefix = None,
         include_prefix = None,
         textual_hdrs = None,
-        defines = []):
+        defines = [],
+        target_compatible_with = None):
     if tcmalloc_dep:
         deps += tcmalloc_external_deps(repository)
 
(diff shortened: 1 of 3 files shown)


keith:
This would require pulling in https://github.com/grailbio/bazel-compilation-database/pull/100 to actually solve  https://github.com/envoyproxy/envoy/issues/22217

keith:
I think with the newest rolling release this should work without the change linked above

Issue #20665, opened 2022-04-05
corrupt stats for internal listener


kyessenov:
Stats produced by an internal listener seem to have "envoy://" in their name, which is confusing. Example:
```
listener.envoy_//internal_outbound.worker_39.downstream_cx_total: 0
````

/cc @lambdai

**Comment by lambdai** [comment]
As a workaround, stat_prefix can be specified 
```
      listener_scope_(server_.stats().createScope(
          fmt::format("listener.{}.",
                      !config.stat_prefix().empty()
                          ? config.stat_prefix()
                          : Network::Address::resolveProtoAddress(config.address())->asString()))),
```

The goal of address `envoy://` as listener name is to avoid conflicting. 
There is no guaranteed collision by default, I think I will simply remove "//" in the default address representation.

Note that `:` is replaced to `_` as part of the metric name sanitize. This sanitize is enforce anyway b/c ipv6 address contains ":"

**Comment by lambdai** [comment]
CC @daixiang0 
Same for pipe: when the address if "/tmp/testpipe". See below

So we'd better sanitize the address string in `listener_impl` rather than changing the `Address::asString()`;
```
listener./tmp/testpipe.downstream_cx_length_ms: No recorded values
```

daixiang0:
I will take a look and update, thanks!

Issue #20335, opened 2022-03-14
Consume bootstrap_extension in config_test


lambdai:
im wondering if its possible that config_test cant handle `bootstrap_extension` ?

i can see one of the example yaml files in the docs has a `layered_runtime` but i dont see any with `bootstrap_extension` so im wondering if that is the culprit

_Originally posted by @phlax in https://github.com/envoyproxy/envoy/pull/19467#discussion_r824343681_

kyessenov:
not stale. I'm wondering it myself as I'm noticing presumably validated samples that shouldn't be right.

PR #21132, opened 2022-05-03
xds: Listener.address field is not required when api_listener is populated

markdroth:
Signed-off-by: Mark D. Roth <roth@google.com>

Commit Message: xds: Listener.address field is not required when api_listener is populated
Additional Description: This removes the overly restrictive PGV annotation that the address field is required, since there are cases where it does not need to be populated.
Risk Level: Low
Testing: N/A
Docs Changes: Included in PR
Release Notes: N/A
Platform Specific Features: N/A

The right long-term solution is to use a oneof for the different types of listeners, as described in #15372.  But in the interim, it should not be necessary to populate the address field for non-socket listeners.

CC @costinm

PR #22274, opened 2022-07-18
api: mark internal listener as implemented and change internal listener binding

kyessenov:

Commit Message: Expose internal listener functionality. Marks the fields as implemented. Adds the following functional changes prior to publishing:
* do not use the `address` field per https://github.com/envoyproxy/envoy/pull/21132, and instead use the listener `name` field as originally expected.
* add an `endpoint_id` field to upstream address to fix https://github.com/envoyproxy/envoy/pull/22420; it's not possible to have multiple internal addresses to the same listener in a single pool without some extra identifier beyond the listener name, because pools identify addresses using `asString()`.
* fixes https://github.com/envoyproxy/envoy/issues/20665 by using stat prefix `listener.envoy_internal_<listener_name>`.

Adds a few sample configs (and fixes https://github.com/envoyproxy/envoy/issues/20335):
* minimal two chained `tcp_proxy` listeners;
* encap HTTP in HTTP CONNECT;
* decap HTTP in HTTP CONNECT;

Risk Level: low, hidden prior
Testing: yes
Docs Changes: yes
Release Notes: yes


phlax:
@adisuissa would you mind taking a look

/wait-any

api/envoy/config/core/v3/address.proto
```diff
@@ -38,7 +38,6 @@ message EnvoyInternalAddress {
   oneof address_name_specifier {
     option (validate.required) = true;
 
-    // [#not-implemented-hide:] The :ref:`listener name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the destination internal listener.
```
adisuissa, line 41: Please keep the comment describing the field.

api/envoy/config/listener/v3/listener.proto
```diff
@@ -334,23 +333,20 @@ message Listener {
 
   // The exclusive listener type and the corresponding config.
   // TODO(lambdai): https://github.com/envoyproxy/envoy/issues/15372
```
adisuissa, line 335:
```suggestion
  // [#comment: TODO(lambdai): https://github.com/envoyproxy/envoy/issues/15372]
```

adisuissa (changes requested):
Thanks for cleaning this up, left a few minor comments.
/wait

api/envoy/config/core/v3/address.proto
kyessenov: This comment is not true, address has nothing to do with the listener name.

api/envoy/config/listener/v3/listener.proto
kyessenov: Done.

api/envoy/config/core/v3/address.proto
adisuissa: IIUC the value here needs to point to another listener.
Would it be possible to add a reference to the field that this value will be matched to?

soulxu: seems like it changed what the value expect to from the beginning. Maybe worth to ask again whether this field still name correctly, this can be a chance to change the name before expose to the end user.

alyssawilk:
Looks like a simple API change: throwing over to @htuch as Adi is out.
@kyessenov should this also have a release note as well?

adisuissa (dismissed):
/lgtm api,
thanks!

adisuissa:
@kyessenov following Alyssa's comment, if the `server_listener_name` field now implemented, then a release note should be added.

alyssawilk:
/wait on review notes - I'm happy to do that pass.  I think if we don't have example configs we're going to want that as part of making this non-hidden.

api/envoy/config/core/v3/address.proto
```diff
@@ -143,7 +142,6 @@ message Address {
 
     Pipe pipe = 2;
 
-    // [#not-implemented-hide:]
```
alyssawilk, line 146: we should have comments here and above explaining what the field is for?

kyessenov:
Affected by https://github.com/envoyproxy/envoy/issues/22483.

kyessenov:
Per https://github.com/envoyproxy/envoy/pull/21132, I'm cleaning up the usage of `address` to replace with a field in `InternalListener` instead.

I'm also making a provision to address https://github.com/envoyproxy/envoy/pull/22420 with a dedicated field in upstream address for a port-like identifier.

api/envoy/config/core/v3/address.proto
kyessenov: Yes, I agree we need to fix this. I think we should use a dedicated field in InternalListener instead of internal address in the generic `address` field in Listener. WDYT?

kyessenov:
I'll add a release note and update config samples. There're several issues that could be resolved in this PR, so I'm trying to be systematic before exposing this.

api/envoy/config/core/v3/address.proto
kyessenov: Done.

api/envoy/config/core/v3/address.proto
kyessenov: The simplest thing is to just really use the listener name. That keeps LDS to a minimum and I can't think of any reason why more flexibility would be needed here (multiple addresses don't work anyways, but can always be added to `InternalListener`).

soulxu: I'm not sure the usecase of using multiple internal addresses, it seems useless. I guess the multiple addresses usecase is internal address mix with IP address. From this point of view, just use the listener name should be fine.

api/envoy/config/core/v3/address.proto
```diff
@@ -32,15 +32,19 @@ message Pipe {
...
-// [#comment: TODO(lambdai): Make this address available for listener and endpoint.
 // TODO(asraa): When address available, remove workaround from test/server/server_fuzz_test.cc:30.]
 message EnvoyInternalAddress {
   oneof address_name_specifier {
     option (validate.required) = true;
 
-    // [#not-implemented-hide:] The :ref:`listener name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the destination internal listener.
+    // Specifies the :ref:`name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the
+    // internal listener.
     string server_listener_name = 1;
```
soulxu, line 42: so this is only used for upstream endpoint?

api/envoy/config/core/v3/address.proto
```diff
@@ -32,15 +32,19 @@ message Pipe {
...
   }
+
+  // Specifies an endpoint identifier to distinguish between multiple endpoints for the same internal listener in a
+  // single upstream pool.
+  string endpoint_id = 2;
```
soulxu, line 47: Is this kind of multiple internal addresses? if yes, then another option is just to support multiple internal addresses here, then you needn't add `endpoint_id` field for multiple endpoints, just specify multiple internal addresses. but yes, you can't limit the internal address to the listener name anymore.

configs/encapsulate_http_in_http2_connect.yaml
```diff
@@ -0,0 +1,81 @@
+# This configuration takes incoming HTTP requests on port 10000 and encapsulates it in a CONNECT
+# request which is sent upstream port 10001.
+# It can be used to test HTTP tunneling as described in
+# https://envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades
+# and running `curl -x 127.0.0.1:10000 https://www.google.com`
+
+admin:
```
phlax, line 3: is admin required for this config? if not can we remove - i have tried to make the configs contain the least amount of configuration possible, esp when it comes to including admin

im also wondering the purpose of including these configs here - and whether this would be better served by adding a tunneling sandbox

api/envoy/config/core/v3/address.proto
soulxu: > I'm not sure the usecase of using multiple internal addresses, it seems useless. I guess the multiple addresses usecase is internal address mix with IP address. From this point of view, just use the listener name should be fine.

suddenly found this may be the multiple internal addresses usecase https://github.com/envoyproxy/envoy/pull/22274/files#r935253240

source/server/connection_handler_impl.cc
```diff
@@ -131,8 +131,11 @@ void ConnectionHandlerImpl::addListener(absl::optional<uint64_t> overridden_list
       }
     } else if (absl::holds_alternative<std::reference_wrapper<Network::InternalListener>>(
                    per_address_details->typed_listener_)) {
-      internal_listener_map_by_address_.insert_or_assign(
-          per_address_details->address_->asStringView(), per_address_details);
+      if (per_address_details->address_->type() == Network::Address::Type::EnvoyInternal) {
```
soulxu, line 134: the line 132 should be already ensure this is internal address

source/server/listener_impl.cc
```diff
@@ -526,31 +549,25 @@ void ListenerImpl::buildInternalListener() {
         config_.has_tcp_backlog_size() || config_.has_tcp_fast_open_queue_length() ||
         (config_.has_transparent() && config_.transparent().value())) {
       throw EnvoyException(
-          fmt::format("error adding listener '{}': has unsupported tcp listener feature",
-                      absl::StrJoin(addresses_, ",", Network::AddressStrFormatter())));
+          fmt::format("error adding listener '{}': has unsupported tcp listener feature", name_));
```
soulxu, line 552: The name isn't a good choice also, since 
https://github.com/envoyproxy/envoy/blob/a53743b20351288a37a61811f27bcda463dd32ad/source/server/listener_manager_impl.cc#L381-L383

And when we support multiple addresses which mix internal and IP address, log the addresses is better.

api/envoy/config/core/v3/address.proto
soulxu: after looking at another PR, multiple addresses will bind the number of listener addresses and the number of endpoints together, I guess that isn't what people want.

source/server/listener_impl.cc
```diff
@@ -512,11 +531,15 @@ void ListenerImpl::buildAccessLog() {
 }
 
 void ListenerImpl::buildInternalListener() {
-  if (config_.address().has_envoy_internal_address()) {
+  if (config_.has_internal_listener()) {
+    if (config_.has_address()) {
```
soulxu, line 535: this check https://github.com/envoyproxy/envoy/blob/c4837b52cebc46022e536061755ee29797c97c32/source/server/listener_manager_impl.cc#L386-L394

is useless anymore.

or we can move this check to 
https://github.com/envoyproxy/envoy/blob/c4837b52cebc46022e536061755ee29797c97c32/source/server/listener_manager_impl.cc#L387
and replace the original check, then it will avoid ListenerImpl's construct to do the useless parsing for the internal addresses
https://github.com/envoyproxy/envoy/pull/22274/files#diff-b0d5e1a1317df8e22a556ee74d6f728379c5f85ce640a9404acc14bfe8192ad3R371

source/server/listener_impl.cc
```diff
@@ -350,22 +362,29 @@ ListenerImpl::ListenerImpl(const envoy::config::listener::v3::Listener& config,
...
-      throw EnvoyException(fmt::format("listener {}: has different socket type. The listener only "
-                                       "support same socket type for all the addresses.",
-                                       name_));
+  if (config.has_internal_listener()) {
+    addresses_.emplace_back(
+        std::make_shared<Network::Address::EnvoyInternalInstance>(config.name()));
+  } else {
+    // All the addresses should be same socket type, so get the first address's socket type is
+    // enough.
+    auto address = Network::Address::resolveProtoAddress(config.address());
```
soulxu, line 371: there should be a case, `config.has_internal_listener()` is false, but internal address is specified in the address field.

configs/encapsulate_http_in_http2_connect.yaml
phlax: thinking on this further - why not just put these in the docs and add as literalinclude examples - it would be simpler than adding a sandbox, but at least they would be visible/used

alyssawilk (commented):
This is shaping up really well - thanks for your patience on iterating :-)
/wait

api/envoy/config/core/v3/address.proto
kyessenov: Correct, we want to maximize reuse of the internal listeners - e.g. tunneling shouldn't be coupled to the origin of the traffic. So that implies a possibility of multiple endpoints to the same internal listener in one pool. Upstream de-duplicates endpoints by computing `asString` so it breaks there.

api/envoy/config/core/v3/address.proto
```diff
@@ -32,15 +32,19 @@ message Pipe {
...
     option (validate.required) = true;
 
-    // [#not-implemented-hide:] The :ref:`listener name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the destination internal listener.
+    // Specifies the :ref:`name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the
+    // internal listener.
     string server_listener_name = 1;
   }
+
+  // Specifies an endpoint identifier to distinguish between multiple endpoints for the same internal listener in a
+  // single upstream pool.
```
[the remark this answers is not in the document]
kyessenov: It is an opaque identifier. It's only used so that `asString` is different when `server_listener_name` is the same. Otherwise, EDS will delete repeated endpoints to the same internal listener.

configs/encapsulate_http_in_http2_connect.yaml
kyessenov: 1. I'll remove admin, I thought it's required.
2. They are too large as literal include because they need the entirety to be meaningful. Snippets work best for isolated config samples.
3. Yes, users will copy paste so having a working example is really helpful.

source/server/listener_impl.cc
kyessenov: Address is empty for internal listener I think in this branch.

configs/encapsulate_http_in_http2_connect.yaml
kyessenov: Let's just do both, add snippets and fully worked out examples.

source/server/listener_impl.cc
kyessenov: Oh you mean we don't check for that early enough? We should throw in buildInternalListener but that might be late.

soulxu: yea, I saw you add the check here https://github.com/envoyproxy/envoy/pull/22274/files#diff-b0d5e1a1317df8e22a556ee74d6f728379c5f85ce640a9404acc14bfe8192ad3R569, that is what I said.

source/server/listener_impl.cc
soulxu: You append an internal addresses here https://github.com/envoyproxy/envoy/pull/22274/files#diff-b0d5e1a1317df8e22a556ee74d6f728379c5f85ce640a9404acc14bfe8192ad3R366-R367, so it shouldn't be empty

api/envoy/config/core/v3/address.proto
```diff
@@ -32,15 +32,19 @@ message Pipe {
 }
 
 // The address represents an envoy internal listener.
-// [#comment: TODO(lambdai): Make this address available for listener and endpoint.
 // TODO(asraa): When address available, remove workaround from test/server/server_fuzz_test.cc:30.]
 message EnvoyInternalAddress {
```
soulxu, line 36: Maybe worth to note that this is only used for upstream endpoint address.

api/envoy/config/listener/v3/listener.proto
```diff
@@ -103,7 +103,6 @@ message Listener {
   }
 
   // Configuration for envoy internal listener. All the future internal listener features should be added here.
-  // [#not-implemented-hide:]
   message InternalListenerConfig {
```
soulxu, line 106: Also maybe worth to note, this will make the listener use the listener name as the internal address. And when this is internal listener, the address field should keep empty.

api/envoy/config/listener/v3/listener.proto
```diff
@@ -333,24 +332,18 @@ message Listener {
...
+    // There are some limitations that are derived from the implementation. The known limitations include:
     //
     // * :ref:`ConnectionBalanceConfig <envoy_v3_api_msg_config.listener.v3.Listener.ConnectionBalanceConfig>` is not
-    //   allowed because both cluster connection and listener connection must be owned by the same dispatcher.
+    //   allowed because both the cluster connection and the listener connection must be owned by the same dispatcher.
     // * :ref:`tcp_backlog_size <envoy_v3_api_field_config.listener.v3.Listener.tcp_backlog_size>`
     // * :ref:`freebind <envoy_v3_api_field_config.listener.v3.Listener.freebind>`
     // * :ref:`transparent <envoy_v3_api_field_config.listener.v3.Listener.transparent>`
-    // [#not-implemented-hide:]
     InternalListenerConfig internal_listener = 27;
```
soulxu, line 347: Also maybe worth to note, this will make the listener use the listener name as the internal address. And when this is internal listener, the address field should keep empty.

source/server/listener_impl.cc
```diff
@@ -512,11 +531,16 @@ void ListenerImpl::buildAccessLog() {
 }
 
 void ListenerImpl::buildInternalListener() {
-  if (config_.address().has_envoy_internal_address()) {
+  if (config_.has_internal_listener()) {
+    if (config_.has_address() || !config_.additional_addresses().empty()) {
+      throw EnvoyException(fmt::format("error adding listener '{}': address should not be used "
+                                       "when an internal listener config is provided",
+                                       name_));
```
soulxu, line 538: looks like missing a testcase for this case.

soulxu (commented):
overall LGTM, only left few comments about doc and test

alyssawilk (commented):
LGTM modulo outstanding comments
/wait

configs/encapsulate_http_in_http2_connect.yaml
kyessenov: Resolving since we can have both snippets and full configs.

api/envoy/config/core/v3/address.proto
kyessenov: Added.

api/envoy/config/listener/v3/listener.proto
kyessenov: Added.

source/server/listener_impl.cc
kyessenov: Internal listeners are always deliberately named since that's the only way to refer to them in EnvoyInternalAddress. I think we can rework this if/when we add multiple address support for internal listeners.

source/server/listener_impl.cc
kyessenov: Good catch. This is too late to check for `api_listener` so I moved it earlier.

api/envoy/config/core/v3/address.proto
```diff
@@ -32,15 +32,19 @@ message Pipe {
...
     option (validate.required) = true;
 
-    // [#not-implemented-hide:] The :ref:`listener name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the destination internal listener.
+    // Specifies the :ref:`name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the
+    // internal listener.
     string server_listener_name = 1;
   }
+
+  // Specifies an endpoint identifier to distinguish between multiple endpoints for the same internal listener in a
+  // single upstream pool. Only used in the upstream addresses.
```
adisuissa, line 46: Is empty string the default value here, or some unique name for the endpoint?
Is there a requirement that each endpoint id will be unique?

adisuissa (commented):
Thanks for working on this, exciting feature!

api/envoy/config/core/v3/address.proto
```diff
@@ -32,15 +32,19 @@ message Pipe {
 }
 
 // The address represents an envoy internal listener.
-// [#comment: TODO(lambdai): Make this address available for listener and endpoint.
 // TODO(asraa): When address available, remove workaround from test/server/server_fuzz_test.cc:30.]
```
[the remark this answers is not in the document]
kyessenov: I think that comment is stale, this address is already available in both places.

docs/root/configuration/other_features/internal_listener.rst
```diff
@@ -3,23 +3,38 @@
...
+
+.. validated-code-block:: yaml
+  :type-name: envoy.config.listener.v3.Listener
+
+  name: internal_listener
+  internal_listener: {}
+  filter_chains:
+  - filters: []
+
+Third, an upstream cluster must include an endpoint with an internal address referencing the internal listener by name:
```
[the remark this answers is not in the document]
kyessenov: It'll be an upstream connect error, same as if server is not listening.

api/envoy/config/core/v3/address.proto
kyessenov: Yes, the empty string is the default. There's no requirement per-se, but it's de-duplicated in the same way as when EDS has duplicate IP addresses.

soulxu:
LGTM

api/envoy/config/core/v3/address.proto
lambdai: Is there any typical value of endpoint_id? e.g. the actually L4 address?

source/common/network/address_impl.cc
```diff
@@ -404,10 +404,11 @@ absl::Status PipeInstance::initHelper(const sockaddr_un* address, mode_t mode) {
 }
 
 EnvoyInternalInstance::EnvoyInternalInstance(const std::string& address_id,
+                                             const std::string& endpoint_id,
                                              const SocketInterface* sock_interface)
     : InstanceBase(Type::EnvoyInternal, sockInterfaceOrDefault(sock_interface)),
-      internal_address_(address_id) {
-  friendly_name_ = absl::StrCat("envoy://", address_id);
+      internal_address_(address_id, endpoint_id) {
+  friendly_name_ = absl::StrCat("envoy://", address_id, ":", endpoint_id);
```
lambdai, line 411: Given the endpoint_id in the real world contains ":" (ipv6 address and port). It might be good to use another separator here.

lambdai (commented):
Thank you!

Only nits

api/envoy/config/core/v3/address.proto
stevenctl: @lambdai yes exactly

api/envoy/config/core/v3/address.proto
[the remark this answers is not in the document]
adisuissa: IIUC, the 2nd TODO (`TODO(asraa)`) seems to still be relevant (the code is still there:
https://github.com/envoyproxy/envoy/blob/5e4c50b0c50db8877cf7197c1b6c1853e27e783f/test/server/server_fuzz_test.cc#L29-L31
).

If it is stale, then the test code should be fixed. If it is not, then the `[#comment` tag should be around that TODO.

api/envoy/config/core/v3/address.proto
adisuissa: Ah, I see. I think it may be beneficial to state that this is used for tracking purposes (opposed to an ID that will be used as a foreign key).

api/envoy/config/core/v3/address.proto
kyessenov: I think it'd be confusing to say "L4 address should be here". We don't say a listener corresponds to VIP, or a cluster corresponds to a VIP. That's too prescriptive. I will add an example of the usage.

api/envoy/config/core/v3/address.proto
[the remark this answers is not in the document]
kyessenov: Ah, missed that, thanks.

api/envoy/config/core/v3/address.proto
kyessenov: Done.

source/common/network/address_impl.cc
kyessenov: Sure, will use slash instead.

kyessenov:
/retest
(emsdk download returned 500)

adisuissa (commented):
/lgtm api

docs/root/configuration/other_features/internal_listener.rst
```diff
@@ -39,85 +55,46 @@ This extension emits the following statistics:
...
-
-Encap HTTP GET requests in a HTTP CONNECT request
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Endpoint disambiguation
+-----------------------
+
+In case there are multiple endpoints referencing the same internal listener in a single upstream cluster, use
+:ref:`endpoint ID field <envoy_v3_api_field_config.core.v3.EnvoyInternalAddress.endpoint_id>` to improve change tracking
+in the cluster pool. This field in combination with the internal listener name uniquely identify an endpoint in the
```
snowp, line 64: What happens if you don't use the endpoint id in case? What do you end up losing out on? Or does it just not work?

snowp (changes requested):
Not familiar enough with the internal listener work to give a true in depth review, but high level this looks good to me, just a docs comment

docs/root/configuration/other_features/internal_listener.rst
kyessenov: It'll de-duplicate the internal addresses. The same happens for IP:PORTs, but not well-documented. This is due to `HostMap` using strings as a key for host tracking. Should we add this comment here?

snowp: I think it would be helpful to either make the language prescriptive (you should do this when multiple IPs to ensure X) or help the reader understand what this does if we want to talk about this as an optional thing to do

kyessenov: Done.

snowp (approved):
Thanks!

adisuissa (commented):
/lgtm api

snowp (approved):

kyessenov:
Thanks @lambdai for doing the actual work! Hopefully we can get more usage and value now that it's public and easily available.

PR #22285, opened 2022-07-19
bazel: Remove uses of `depset.to_list` where possible

phlax:

Use of depsets, and `to_list` are recomended against unless actually necessary (https://docs.bazel.build/versions/main/skylark/performance.html)

The recommended advice to simulate a set in starlark is to use dictionary keys (https://docs.bazel.build/versions/main/skylark/lib/depset.html)

> The elements of a depset must be hashable and all of the same type (as defined by the built-in type(x) function), but depsets are not simply hash sets and do not support fast membership tests. If you need a general set datatype, you can simulate one using a dictionary where all keys map to True.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/bazel/api_build_system.bzl
+++ b/api/bazel/api_build_system.bzl
@@ -179,14 +179,10 @@ def api_proto_package(
 
     # Because RBAC proro depends on googleapis syntax.proto and checked.proto,
     # which share the same go proto library, it causes duplicative dependencies.
-    # Thus, we use depset().to_list() to remove duplicated depenencies.
-    go_proto_library(
-        name = name + _GO_PROTO_SUFFIX,
-        compilers = compilers,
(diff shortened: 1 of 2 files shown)


phlax:
cc @kyessenov

kyessenov (approved):
I wonder why they just didn't implement `to_list` with dictionaries.

phlax:
i think the depset itself is not so efficient to use as a set - the `to_list` part is generally bad because it explodes eg cli arg lists that would otherwise be created lazily and gobbles ram

in this case the `to_list` part is probably not making that much of a difference, altho its hard to tell - the code is hit a lot and does create large lists, but its basically still doing that just slightly more efficently

PR #22286, opened 2022-07-19
add integration test for lds warming

ramaraochavali:
Commit Message: add integration test for lds warming
Additional Description: add integration test for lds warming to validate cases that require RDS update.
Risk Level: Low
Testing: N/A
Docs Changes: N/A
Release Notes: N/A


test/integration/ads_integration_test.cc
```diff
@@ -1004,6 +1004,62 @@ TEST_P(AdsIntegrationTest, ClusterWarmingOnNamedResponse) {
   test_server_->waitForGaugeEq("cluster_manager.warming_clusters", 0);
 }
 
+// Verify Listener warming is finished only on named RDS response for new routes
+// and does not get in to warming state for existing route names.
```
snowp, line 1008: Rephrasing to make sure that I get this: This is verifying that if a listener is updated with new route names, it finishes warming when it receives those routes. Additionally, it tests that if the update has no new routes the listener is immediately usable (no warming phase)?

Maybe the comment could be clearer that it's testing two flows, at first read it seemed like the second part was a consequence of the first

test/integration/ads_integration_test.cc
```diff
@@ -1004,6 +1004,62 @@ TEST_P(AdsIntegrationTest, ClusterWarmingOnNamedResponse) {
...
+      {buildClusterLoadAssignment("cluster_1")}, {"cluster_0"}, "2");
+  sendDiscoveryResponse<envoy::config::listener::v3::Listener>(
+      Config::TypeUrl::get().Listener, {buildListener("listener_0", "route_config_0", "rds_test")},
+      {buildListener("listener_0", "route_config_0", "rds_test")}, {}, "2");
+
+  // Validate that listener is updated correctly and does not get in to warming state.
+  test_server_->waitForCounterGe("listener_manager.listener_create_success", 2);
+  test_server_->waitForGaugeEq("listener_manager.total_listeners_warming", 0);
+
+  // Update listener with routing config name change.
```
snowp, line 1046: `with a new route` perhaps?

snowp (changes requested):
Thanks this makes sense, just some wording suggestions

test/integration/ads_integration_test.cc
ramaraochavali: yes that is correct

ramaraochavali:
@snowp Fixed. PTAL.

snowp (approved):
Thanks!

PR #22277, opened 2022-07-19
deps: Bump `build_bazel_rules_apple` -> 1.0.1

phlax:
Fix #21866

Signed-off-by: Ryan Northey <ryan@synca.io>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #22295, opened 2022-07-19
build: update bazel to latest rolling release

jpsim:
Recent changes to rules_apple require this commit: https://github.com/bazelbuild/bazel/commit/68f29c67098cedbfaa3ef011eef4df661860ac36

So bumping to the latest rolling release of bazel now to unblock further downstream updates.

Risk Level: Low
Testing: CI
Docs Changes: N/A
Release Notes: N/A


keith (approved):
Thanks I have been meaning to do this. Let's see what CI says. We might have to bump some rules at the same time

phlax:
related #22277

keith:
great, i think we should be pretty safe on rules_apple since i don't think we hit much of that code here, we just pull it transitively for grpc

PR #22301, opened 2022-07-19
hcm: factoring out filter chain creation code

alyssawilk:
Risk Level: low (code refactor)
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/filter_chain_helper.cc
+++ b/source/common/http/filter_chain_helper.cc
@@ -15,6 +15,7 @@
 namespace Envoy {
 namespace Http {
 
+// Allows graceful handling of missing configuration for ECDS.
 class MissingConfigFilter : public Http::PassThroughDecoderFilter {
 public:
   Http::FilterHeadersStatus decodeHeaders(Http::RequestHeaderMap&, bool) override {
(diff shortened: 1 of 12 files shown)


source/common/http/filter_chain_helper.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "source/common/common/fmt.h"
+#include "source/common/config/utility.h"
+#include "source/common/http/utility.h"
+#include "source/common/protobuf/utility.h"
+#include "source/extensions/filters/http/common/pass_through_filter.h"
+
+namespace Envoy {
+namespace Http {
+
+class MissingConfigFilter : public Http::PassThroughDecoderFilter {
```
RyanTheOptimist, line 18: nit: Can you add a comment which explains the purpose of a MissingConfigFilter?

source/common/http/filter_chain_helper.cc
```diff
@@ -0,0 +1,153 @@
...
+      ENVOY_LOG(trace, "Provider {} missing a filter config", filter_config_provider->name());
+    }
+  }
+}
+
+void FilterChainHelper::processFilters(const FiltersList& filters, const std::string& prefix,
+                                       const std::string& filter_chain_type,
+                                       FilterFactoriesList& filter_factories) {
+  DependencyManager dependency_manager;
+  for (int32_t i = 0; i < filters.size(); i++) {
```
RyanTheOptimist, line 61: I could be wrong, but it looks like Protobuf::RepeatedPtrField::size() returns int, not int32_t. Should we use int to avoid truncation? (Not that we're likely to have more than 2^32 fields)

RyanTheOptimist (approved):
LGTM, but two minor comments. i realize this code is merely being moved but it would be nice, I think, to fix.

RyanTheOptimist (approved):

PR #22302, opened 2022-07-19
contrib: require an end-user sponsor

mattklein123:


mattklein123:
cc @envoyproxy/envoy-maintainers @moderation for comment

EXTENSION_POLICY.md
```diff
@@ -147,6 +147,14 @@ included by default in the main image builds. Consumers need to pull directly fr
 images described in the installation guide. Please read the linked document in detail to determine
 whether contrib extensions are the right choice for a newly proposed extension.
 
+**NOTE:** Contrib extensions **require** an end-user sponsor. The sponsor is someone who will run
```
mattklein123, line 150: Note: need to further define end-user and probably cross link to the security process which also talks about end-users. We should use the same definition in both places. cc @moderation

mattklein123: Referenced https://github.com/envoyproxy/envoy/blob/main/SECURITY.md#membership-criteria but not really well defined. So let's carefully define and then cross link.

Very roughly, and end-user is someone that uses Envoy to directly serve traffic for a 1st party offering. A vendor can also be an end-user (Google would be the canonical example of this within the community.)

jmarantz (approved):
This makes sense to me. Thanks for writing this up.

ggreenway (approved):

phlax (approved):

kyessenov (dismissed):

adisuissa (approved):

wrowe (dismissed):

alyssawilk (approved):

mattklein123:
I'm going to take a stab at better defining end user today and will ping back for another review.

mattklein123:
@envoyproxy/envoy-maintainers I updated this PR and attempted to define end user which also is relevant to the security list membership policy. PTAL.

alyssawilk (approved):

KBaichoo (approved):

PR #22305, opened 2022-07-19
code_format: Shift config to yaml

phlax:
this will make it easier to 

- work on, and to factor out parts of the checker
- resolve the awkward path normalization in this file
- shift the format checker to bazel

and fixes some policy-breaking language use

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
@htuch this one is now pretty well tested, and has a couple of PRs behind it

tools/code_format/check_format.py
```diff
@@ -16,232 +16,120 @@
...
-    ".Times(1).WillRepeatedly": ".WillOnce",
-    "Stats::ScopePtr": "Stats::ScopeSharedPtr",
-}
-
-UNSORTED_FLAGS = {
-    "envoy.reloadable_features.activate_timers_next_event_loop",
-    "envoy.reloadable_features.grpc_json_transcoder_adhere_to_buffer_limits",
-}
+# The way this script is currently used (ie no bazel) it relies on system deps.
+# As `pyyaml` is present in `envoy-build-ubuntu` it should be safe to use here.
```
htuch, line 22: This is not just a Docker tool, this is part of normal developer flow.

tools/code_format/check_format.py
```diff
@@ -16,232 +16,120 @@
...
+        paths["build_fixer_py"] = self._build_fixer_path
+        paths["header_order_py"] = self._header_order_path
+        return paths
+
+    @cached_property
+    def re(self) -> Dict[str, Pattern[str]]:
+        return {k: re.compile(v) for k, v in self["re"].items()}
+
+    @cached_property
+    def re_multiline(self) -> Dict[str, Pattern[str]]:
```
htuch, line 64: Can you add some comments to help readability for these properties?

tools/code_format/check_format.py
```diff
@@ -372,40 +260,44 @@ def check_namespace(self, file_path):
...
+        return name in self.config.suffixes["histogram_with_si"]["include"]
 
     def allow_listed_for_std_regex(self, file_path):
-        return file_path.startswith("./test") or file_path in STD_REGEX_ALLOWLIST
+        return file_path.startswith(
+            "./test") or file_path in self.config.paths["std_regex"]["include"]
 
     def allow_listed_for_grpc_init(self, file_path):
-        return file_path in GRPC_INIT_ALLOWLIST
+        return file_path in self.config.paths["grpc_init"]["include"]
```
htuch, line 300: Why not make this a proto? That way we can still do YAML -> JSON -> proto3, but we end up with a typed schema, instead of this dictionary stuff. It would strictly be more readable and scalable.

htuch (changes requested):
Overall makes a lot of sense. I do feel fairly strongly about using a typed underlying schema for the config YAML, since this is non-trivial and will grow in complexity. We have good results across Envoy when making this proto3 and using the canonical JSON -> proto3 transform.

/wait

tools/code_format/check_format.py
phlax: at the momemt its run from hooks and ci

it doesnt use bazel and it doesnt have _any_ deps

this change adds a dep - which atm can only be done if it _is_ available in the docker image

tools/code_format/check_format.py
phlax: proto is great for wire/message/serialization formats - none of that is relevant here

i think the point is moot anyhow because as it stands this _cannot_ add any deps

when this gets moved to bazel (and probs pytooling) we can add deps - i still see no reason even then to add protobuf - its a lot of overhead for no real benefit

in this PR i added type hinting - mostly for my/developers benefit - its not linted/checked anywhere. In pytooling i have added some code (based on upstream `typecast`) that does ensure the parsed config files are valid/well-formed

what this PR does do is take some of the worst code in codebase (readability/usability/performance - and up until recently code errors) and factors out a big chunk of it into a format that end users/devs can actually work with in a reasonable fashion

phlax:
@htuch this is an important step in fixing a long term problem - and is very much not the end goal

it has taken me a considerable amount of time to get it this far - can we please get this landed so we can move on

tools/code_format/check_format.py
phlax: docstrings added and the erroneously named `subdir_set` has been shifted to config and renamed

htuch:
The proto3 wire thing is not relevant, what's relevant is that we have some typed description of the structure. I would like a plan for a typed schema for the YAML not the source code. Happy to be pragmatic here and land if we can agree on a plan for typed YAML schema. Having properly structured configuration schemas is really important to scalability. Please propose a plan to get there @phlax.

htuch:
@phlax SG, I wil ltake another pass later today. However, I think we have a fundamental confusion:

> on this point i have to strongly disagree - the code is much more complex, more likely to change/scale/etc and type checking is critical to doing this in a stable fashion

I absolutely agree we need Python type checking. But we _also_ need a human readable schema please.

tools/code_format/check_format.py
htuch: Are you saying it's run under Docker from the hooks? Or locally?

I'm basically wondering if we need to document this somewhere like https://github.com/envoyproxy/envoy/blob/main/bazel/README.md#quick-start-bazel-build-for-developers or have a venv?

I realize this connects to the overall Bazel story but I don't want to regress the devX.

tools/code_format/config.yaml
```diff
@@ -0,0 +1,297 @@
...
+  codeowners_contrib: (/contrib/[^@]*\s+)(@.*)
+  codeowners_extensions: .*(extensions[^@]*\s+)(@.*)
+  comment: //|\*
+  designated_initializer: \{\s*\.\w+\s*\=
+  # Check for punctuation in a terminal ref clause, e.g.
+  # :ref:`panic mode. <arch_overview_load_balancing_panic_threshold>`
+  dot_multi_space: \. +
+  duration_value: \b[Dd]uration\(([0-9.]+)
+  # C++17 feature, lacks sufficient support across various libraries / compilers.
+  for_each_n: for_each_n\(
```
htuch, line 237: I'm not sure how I feel about having every constant moved out of the Python file into YAML. I certainly agree all the includes/excludes and things we want to keep changing belong there. Let me get a second opinion from @jmarantz, I don't feel that strongly.

htuch (commented):
LGTM, just a few comments to wrap.

tools/code_format/check_format.py
phlax: > Are you saying it's run under Docker from the hooks? Or locally?

in CI it uses the docker container (and it seems in some devs workflows)

in the hooks it is called directly

> I'm basically wondering if we need to document this somewhere like https://github.com/envoyproxy/envoy/blob/main/bazel/README.md#quick-start-bazel-build-for-developers

pyyaml needs to be added to the deps i think

> or have a venv?

please lets focus on using bazel - i have removed (almost) all venvs as their use was messy and complex

> I realize this connects to the overall Bazel story 

its a step towards getting this into bazel

> I don't want to regress the devX.

this PR specifically should barely touch the devX (pyyaml becomes a dep of the script, it becomes easier to add in/exclusions etc)

phlax: i should add that once we do get this into bazel the possibility of running this incrementally opens up which would greatly improve the devX

tools/code_format/config.yaml
phlax: > I'm not sure how I feel about having every constant moved out of the Python file into YAML.

apart from making the script and data/config a lot clearer (separately) this allows the script to move to pytooling while the data/config it uses can remain in envoy itself where it can be seen/adjusted/etc by dev without having to worry about how eg substitutions are actually substituted

htuch (approved):

PR #22323, opened 2022-07-20
ci: Timeout test result publishing if agent is hung

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/bazel.yml
+++ b/.azure-pipelines/bazel.yml
@@ -72,6 +72,7 @@ steps:
     testResultsFiles: "**/bazel-out/**/testlogs/**/test.xml"
     testRunTitle: "${{ parameters.ciTarget }}"
     searchFolder: $(Build.StagingDirectory)/tmp
+  timeoutInMinutes: 10
   condition: always()
 
 - task: PublishBuildArtifacts@1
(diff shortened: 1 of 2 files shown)


phlax:
checking azp analytics the upload job can legitimately take a while - up to around 5 mins afaict - if it has taken 10 mins then i think we can assume it has hung

phlax:
do the same for `PublishBuildArtefacts`

/wait

lizan:
I doubt it has effect since the timeout is implemented in agent, if agent hungs it is likely no-op.

phlax:
> the timeout is implemented in agent

can you point me to the docs that says this is a timeout _in_ agent

my reading of https://developercommunity.visualstudio.com/t/azure-devops-timeoutinminutes-is-being-ignored-in/567714 and https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts is that this should work

should we not at least try this - the arm agent is hanging fairly frequently during the publish steps - do you think we can set the timeout somewhere else to tackle this ?

lizan:
https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/StepsRunner.cs

If the agent didn't die but the upload task times out this works so we can definitely try.

lizan:
@phlax did this work out?

phlax:
no way to tell until we land it

lizan (approved):

Issue #22324, opened 2022-07-20
`@envoy//test/extensions/transport_sockets/http_11_proxy:connect_integration_test` is flaky (or always fails?)


phlax:
Checking azp analytics the test posted in title appears to always fail (3/3 times in last 14 days)

https://dev.azure.com/cncf/envoy/_test/analytics?definitionId=11&contextType=build

recent azp failure:

https://dev.azure.com/cncf/envoy/_build/results?buildId=113339&view=logs&j=e969334a-0e55-5c18-ac96-8b546753391e&t=b5329a9e-93c7-5412-9f4c-6f8d659e1b57&l=9267

The error log doesnt seem to have any errors:

https://source.cloud.google.com/results/invocations/aaecf512-1507-4b36-86f0-18479974724a/targets/@envoy%2F%2Ftest%2Fextensions%2Ftransport_sockets%2Fhttp_11_proxy:connect_integration_test;shard=1;run=1;attempt=2/log

but the test returns code `134` which im guessing is killed by timeout

https://source.cloud.google.com/results/invocations/aaecf512-1507-4b36-86f0-18479974724a/targets/@envoy%2F%2Ftest%2Fextensions%2Ftransport_sockets%2Fhttp_11_proxy:connect_integration_test;shard=1;run=1;attempt=2/artifacts

its not clear why this test has only run three times, or if that is correct - either way i think there is an issue with this test

resultstore has a massive artefact bundle but i dont have space atm to look in it

phlax:
cc @alyssawilk

alyssawilk:
weird, almost always passes locally, must be a resource bug.  looks like time_system.waitfor doesn't work in this circumstance because of the readDisable.  Have a fix in the works.

PR #22328, opened 2022-07-20
test: fixing a flake

alyssawilk:
Fixes https://github.com/envoyproxy/envoy/issues/22324

### DIFF
--- a/test/integration/fake_upstream.cc
+++ b/test/integration/fake_upstream.cc
@@ -959,7 +959,10 @@ FakeHttpConnection::waitForInexactRawData(const char* data, std::string* out,
     auto result = dynamic_cast<Network::ConnectionImpl*>(&connection())
                       ->ioHandle()
                       .recv(peek_buf, 200, MSG_PEEK);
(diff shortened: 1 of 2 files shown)


test/integration/fake_upstream.cc
```diff
@@ -972,10 +975,15 @@ FakeHttpConnection::waitForInexactRawData(const char* data, std::string* out,
     }
     return false;
   };
-  if (!time_system_.waitFor(lock_, absl::Condition(&reached), timeout)) {
-    return AssertionFailure() << "timed out waiting for raw data";
+  // Because the connection must be read disabled to not auto-consume the
+  // underlying data, waitFor hangs with no events to force the time system to
+  // continue. Break it up into smaller chunks.
```
RyanTheOptimist, line 980: Is the idea the calling waitFor() in smaller chunks causes time to move forward somehow? Does the event loop get pumped somehow? I suspect this all works, but I don't quite understand it.

RyanTheOptimist (approved):

PR #22330, opened 2022-07-20
fm: handling decode data better

alyssawilk:
moving filter end decode from the FM (an upstream /downstream concept) to HCM as it should only be done once.

Risk Level: low
Testing: existing tests
Docs Changes: n/a
Release Notes: n/a


snowp (approved):
Nice cleanup, thanks!

PR #22350, opened 2022-07-21
Add support for intermediate CA as trusted_ca

### LINKED ISSUE #19326: Add support for intermediate CA as trusted_ca with X509_V_FLAG_PARTIAL_CHAIN
*Title*: Add support for intermediate CA as trusted_ca with X509_V_FLAG_PARTIAL_CHAIN

*Description*:
I am attempting to use Envoy with a `common_tls_context.validation_context.trusted_ca` where the PEM encoded certificate at that path is an intermediate (not self signed) certificate. Envoy is failing to communicate with the upstream with error:

> DeltaAggregatedResources gRPC config stream closed: 14, upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: TLS error: 268435581:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED

After looking around a bit, and trying to use `openssl s_client -connect` with the certs I found that `openssl` only works if I use the `-partial_chain` flag. If I understand correctly the equivalent flag for the library is `X509_V_FLAG_PARTIAL_CHAIN`, and I don't see any way to set that in Envoy.

Is there some other way to allow an intermediate CA in `trusted_ca` ? Would you consider adding another option to `common_tls_context.validation_context` (or something around there) to pass this `X509_V_FLAG_PARTIAL_CHAIN` flag to the certificate verification?

Docs reference for that flag ([link](https://www.openssl.org/docs/man1.1.1/man3/X509_VERIFY_PARAM_set1_ip_asc.html)):

> The X509_V_FLAG_PARTIAL_CHAIN flag causes intermediate certificates in the trust store to be treated as trust-anchors, in the same way as the self-signed root CA certificates. This makes it possible to trust certificates issued by an intermediate CA without having to trust its ancestor root CA. With OpenSSL 1.1.0 and later and <X509_V_FLAG_PARTIAL_CHAIN> set, chain construction stops as soon as the first certificate from the trust store is added to the chain, whether that certificate is a self-signed "root" certificate or a not self-signed intermediate certificate. Thus, when an intermediate certificate is found in the trust store, the verified chain passed to callbacks may be shorter than it otherwise would be without the X509_V_FLAG_PARTIAL_CHAIN flag.

LuyaoZhong:
Add support for intermediate CA as trusted_ca with
X509_V_FLAG_PARTIAL_CHAIN

Risk Level: Low
Testing: unit tests
Docs Changes: ssl
Release Notes: yes
Fixes #19326


ggreenway:
Before proceeding with this, we need to conclude discussion in the issue to make sure this won't have any negative side effects.

/wait

LuyaoZhong:
@ggreenway @lizan 

I just tried to get the CI results. Unfortunately, the cert verification test fail, I need to do some investigation.

LuyaoZhong:
@ggreenway @lizan @dnephin

Without `X509_V_FLAG_PARTIAL_CHAIN`, intermediate ca cert will not be treated as trust-anchors, it requires the trust-ca being a cert chain which containning root ca cert. `max_verify_depth` is config for the max number of intermediate certificates in chain that are parsed during verification. So it expects a failure when setting `max_verify_depth` no longer than root ca cert depth.

https://github.com/envoyproxy/envoy/blob/18212bb6395af308d895f75352f82df522b038b4/test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc#L68-L73

After adding `X509_V_FLAG_PARTIAL_CHAIN` as a default flag, this test can not pass. Because the intermediate ca cert can be treated as trust-anchor with this flag, the verification will succeed.

Except above, I don't think there are other negative side effects. I propose making this flag as default, and refine the testcases a little bit, what's your suggestion?

lizan:
1. sounds reasonable to me.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
```diff
@@ -64,8 +64,8 @@ TEST_P(SslCertValidatorIntegrationTest, CertValidated) {
   codec->close();
 }
 
-// With verify-depth set, certificate validation fails
-TEST_P(SslCertValidatorIntegrationTest, CertValidationFailed) {
+// With verify-depth set, certificate validation succeeds since we allow partial chain
+TEST_P(SslCertValidatorIntegrationTest, CertValidatedWithVerifyDepth) {
```
LuyaoZhong, line 68: It seems `max_verify_depth` doesn't work well with `X509_V_FLAG_PARTIAL_CHAIN`.
Here the depth is set to 1, and intermediate_partial_ca_cert_chain.pem is set to trusted ca, which contains intermediate_ca_3  and intermediate_ca_2, the certificate is signed by intermediate_ca_2, see
https://github.com/envoyproxy/envoy/blob/bfc2cc21b19f9d2b4beb44d6f3cc2a9fa9ed4afd/test/integration/ssl_utility.cc#L50-L56

The intermediate_ca_2 depth is 2 in the chain so I thought it should fail if specifying depth to 1, but it succeeds.

This makes me confusing, do you have any ideas about this?

LuyaoZhong: Whoops, I misunderstood the `max_verify_depth`, it does not mean the depth of cert chain from the trust_ca file, it actually is the maximum number of intermediate certificate issuers. So this test's behavior is correct.

changelogs/current.yaml
```diff
@@ -9,6 +9,9 @@ behavior_changes:
 
 minor_behavior_changes:
 # *Changes that may cause incompatibilities for some users, but should not for most*
+- area: tls
+  change: |
+    added support for intermediate CA as trusted ca.
```
ggreenway, line 14: Please elaborate more on what change the user (of envoy) will see: which certs would previously be untrusted but now are trusted? And please link to the proto config field where the trusted CA is configured.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
```diff
@@ -64,8 +64,8 @@ TEST_P(SslCertValidatorIntegrationTest, CertValidated) {
   codec->close();
 }
 
-// With verify-depth set, certificate validation fails
-TEST_P(SslCertValidatorIntegrationTest, CertValidationFailed) {
+// With verify-depth set, certificate validation succeeds since we allow partial chain
```
ggreenway, line 67: It looks like setting this PR breaks the existing feature for limiting verify depth. Is that correct?

ggreenway (changes requested):
/wait

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
ggreenway: Or is this test configured to trust all the intermediate CAs, and so the end of the max_verify_depth points to a trusted intermediate?

If that's the case, please add another test case with not-trusted intermediates and a max_verify_depth and validate that it fails.

LuyaoZhong: max_verify_depth limits the certificate chain size when constructing chain during verification.

Assuming there is a full certificate chain like this: cert -> intermediate 2 ca < intermediate 1 ca < root ca.
If max_verify_depth is set to 2 , the chain constructed will be cert -> intermediate 2 ca < intermediate 1 ca
with X509_V_FLAG_PARTIAL_CHAIN, we don't need to trust all intermediate CAs, only trust intermediate 2 ca is enough (intermediate 2 ca is treated as trust anchor).
without X509_V_FLAG_PARTIAL_CHAIN, validation fails because intermediate CAs are not treated as trust anchor.

I can add a test with max_verify_depth set and intermediate 2 ca not in trust_ca, it that you required?

changelogs/current.yaml
LuyaoZhong: Got it, will do that.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
ggreenway: > I can add a test with max_verify_depth set and intermediate 2nd ca not in trust_ca, is that what you required?

Yes, please add that test, otherwise there isn't a test case anymore for failing to validate due to max_depth anymore.

LuyaoZhong: I need to clarify a little bit before adding this test.

I think the failing for this case you required is not due to max_depth, it 's due to no valid partial chain can be constructed based on trust ca.
That is to say if we don't set max_depth at all, and trust_ca does not contain intermediate 2nd ca, it still fails because we can't construct valid partial chain without trust intermediate 2nd ca.

The description for [max_verify_depth](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-certificatevalidationcontext) is not accurate:

> If configured, and the certificate chain is longer than allowed, the certificates above the limit are ignored, and certificate validation will fail.

This is just the old behavior in Envoy, not exact meaning of the max_verify_depth.

We should reword like [openssl doc](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify_depth.html):
> Certificate chain must reach a trusted issuer no longer than allowed. Otherwise, certificate validation will fail.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
```diff
@@ -74,6 +74,37 @@ TEST_P(SslCertValidatorIntegrationTest, CertValidationFailed) {
...
+}
+
+// With broken chain set, certificate validation fails with max depth
+TEST_P(SslCertValidatorIntegrationTest, CertValidationFailedWithBrokenChainAndDepth) {
+  config_helper_.addSslConfig(ConfigHelper::ServerSslOptions()
+                                  .setRsaCert(true)
+                                  .setTlsV13(true)
+                                  .setClientWithIntermediateCert(true)
+                                  .setBrokenChain(true)
+                                  .setVerifyDepth(1));
```
soulxu, line 104: It doesn't seems a useful case to test, it still failed due to broken chain not due to the depth. Or maybe comment more about this test case.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
LuyaoZhong: > not sure I understand correctly. I'm thinking the case of the user setting a chain which longer than max_depth, but with `X509_V_FLAG_PARTIAL_CHAIN`, it only validates part of them. And that happened silently.
> 
It happened silently.

ggreenway: The case I'm interested in here is leaf->intermediate1->intermediate2->root, with max_verify_depth set to 1 (or 2?). If only the root is trusted, this should fail. If the intermediate is trusted, it should successfully verify.

LuyaoZhong: > The case I'm interested in here is leaf->intermediate1->intermediate2->root, with max_verify_depth set to 1 (or 2?). If only the root is trusted, this should fail. If the intermediate is trusted, it should successfully verify.

Thus we need to make this as an option instead of force enabling, introducing a field "partial_chain" into [`CertificateValidationContext`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-certificatevalidationcontext), and set its default value to true in tests. Then we can choose if we only trust root or trust intermediate. Is it OK?

api/envoy/extensions/transport_sockets/tls/v3/common.proto
```diff
@@ -478,9 +478,8 @@ message CertificateValidationContext {
   bool only_verify_leaf_cert_crl = 14;
 
   // Config for the max number of intermediate certificates in chain that are parsed during verification.
-  // This does not include the leaf certificate. If configured, and the certificate chain is longer than allowed, the certificates
-  // above the limit are ignored, and certificate validation will fail. The default limit is 100,
-  // though this can be system-dependent.
+  // This does not include the leaf certificate. If configured, certificate chain must reach a trusted issuer no longer than allowed.
+  // Otherwise, certificate validation will fail. The default limit is 100, though this can be system-dependent.
```
adisuissa, line 482: Maybe something like the following will be better:
```
// Defines the max number of intermediate certificates in the chain that are parsed during verification.
// This does not include the leaf certificate. If a trusted issuer appears in the chain, but in a depth larger than configured, the certificate validation will fail.
// The default limit is 100, though this can be system-dependent. See
// `SSL set_verify_depth
// <https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify_depth.html>`_.
```

adisuissa (commented):
I've added a suggestion for the API comment, let me know if this doesn't capture this field correctly.

api/envoy/extensions/transport_sockets/tls/v3/common.proto
LuyaoZhong: Thanks for refine the words, this is correct.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
LuyaoZhong: Yeah, maybe we need introduce a config option for this feature, otherwise the depth controls nothing.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
LuyaoZhong: @ggreenway what do you think about adding a config option for `X509_V_FLAG_PARTIAL_CHAIN`, if no and setting `X509_V_FLAG_PARTIAL_CHAIN` silently, the depth can not control anything.

new config option can provide flexibility, silent setting can keep API consistent, not sure one is better.

ggreenway: I don't think an option is needed. You just need to set the trusted ca certs appropriately in the test. Two tests, both setting max-verify-depth to a length less than the full chain. In one of the tests, you set the intermediate as trusted. The validation will succeed in this test. In the other test, you only set the root of the chain as trusted, and not the intermediates. In this test, the validation will fail.

LuyaoZhong: @ggreenway 

> you only set the root of the chain as trusted, and not the intermediates

Do you mean only root ca in trust_ca file? In this case, validation will always fail, it has nothing to do with max-verify-depth. Let me explain case by case.

trust_ca should contain left partial chain or full chain. For a leaf cert -> intermediate2 -> intermediate1 -> root,

1. trust_ca contains intermediate2 -> intermediate1 or intermediate2 (partial chian)
if we don't set X509_V_FLAG_PARTIAL_CHAIN, verification always fails since only root can be trust anchor
If we set X509_V_FLAG_PARTIAL_CHAIN, verification always succeeds since intermediate ca can be trust anchor

2. trust_ca contains  intermediate2 -> intermediate1 -> root (full chain)
if we don't set X509_V_FLAG_PARTIAL_CHAIN, verification result depends on max-verify-depth, it succeeds when the depth >= root ca depth
If we set X509_V_FLAG_PARTIAL_CHAIN, verification always succeeds since intermediate ca can be trust anchor

4. trust_ca contains intermediate1 -> root or intermediate1 or root (broken chain)
validation always fails, because we can not build chain with out intermediate 2 in trusted ca.

If I understand correctly, it will try to build a chain during velidation, all certs in this chain must be in trust ca otherwise chain building will stop. max-verify-depth also have effects on the length of the chain.
X509_V_FLAG_PARTIAL_CHAIN just to indicate if the intermediate ca can be trust anchor, if not set our chain must contain root ca.

Thus to satisfy your requirement on the test case, we must make the X509_V_FLAG_PARTIAL_CHAIN to be an option.

ggreenway: > If I understand correctly, it will try to build a chain during velidation, all certs in this chain must be in trust ca otherwise chain building will stop.

This is incorrect. The chain-building tries to find a chain from the leaf to one of the certs in `trusted_ca`. Without FLAG_PARTIAL_CHAIN, it will only consider self-signed certs when looking for a trusted cert (it treats anything that isn't self-signed as if it weren't in the `trusted_ca` at all). With FLAG_PARTIAL_CHAIN, it will also consider intermediate certs IFF they're in `trusted_ca`. But the intermediates do NOT need to be in `trusted_ca` as long as something in their cert-chain is in `trusted_ca`.

LuyaoZhong: >  But the intermediates do NOT need to be in trusted_ca as long as something in their cert-chain is in trusted_ca.

I don't think it's correct.
See
https://github.com/envoyproxy/envoy/blob/5261b4285d1487f93e7aece94786de9a8d3a2529/test/config/integration/certs/certs.sh#L80-L81

https://github.com/envoyproxy/envoy/blob/5261b4285d1487f93e7aece94786de9a8d3a2529/test/config/utility.cc#L1374-L1380

https://github.com/envoyproxy/envoy/blob/18212bb6395af308d895f75352f82df522b038b4/test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc#L53-L65

The test will fail if we remove any one of the `cacert.pem` `intermediate_cacert.pem`  `intermediate_ca_2cert.pem` from trust ca. This is what I observe from testing (without FLAG_PARTIAL_CHAIN) 

So I infer that
1. If the `intermediate_ca_2cert.pem` is not in trust ca, the built chain will only contains leaf
2. if the `intermediate_cacert.pem` is not in trust ca, the built chain will be leaf -> intermediate_ca_2cert
3. if the `cacert.pem` is not in trust ca, the built chain will be leaf -> intermediate_ca_2cert -> intermediate_cacert.
Without FLAG_PARTIAL_CHAIN, only root ca(self-signed) can be trust anchor, so we have to build a full chain for successful validation.

For the testing with FLAG_PARTIAL_CHAIN, 2 and 3 will succeed, since the intermediate ca can be trust anchor.

ggreenway: There's a lot going on here, but it's all related to how this specific test is setup.

* The client is not sending a chain; it's only sending the leaf cert
* Because the client doesn't send the chain, validation will fail on the server unless the server has the full chain. Putting the full chain in the trusted ca accomplishes this.
* The chain is actually malformed here: intermediate_ca is configured with `pathlen:0`, which means that it is not valid for it to sign intermediate_2. This misconfiguration requires that intermediate2 be a trusted CA for validation to succeed. I'm pretty sure that setting `pathlen:0` on intermediate CA is a bug in the test setup, and should be fixed by setting `pathlen:1` both locations in `intermediate_cacert.cfg`.

To test this new flag, I think you'll need to generate the client cert full chain, via `cat client2cert.pem intermediate_ca_2cert.pem intermediate_cacert.pem cacert.pem > client2_chain.pem`. Then you should be able to only add the root CA to trusted CA configuration.

LuyaoZhong: @ggreenway 

When I fix the `pathlen` issue, generate the client full chain, and only add root CA to trusted CA configuration, the test still succeeds to validate cert with` max_verify_depth = 1` (FLAG_PARTIAL_CHAIN is set).  It seems no difference that we put chain on client side or server side.

If we put the chain on client cert, which level does it start to calculate depth? Does it still start from leaf cert?

ggreenway: I would have assumed it would start from the leaf cert, and would fail in this configuration. Can you post the test code for this?

LuyaoZhong: The test code for this is: 
https://github.com/envoyproxy/envoy/blob/765e7ab10b2cd72742a238de6e75648d07a6a985/test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc#L101

api/envoy/extensions/transport_sockets/tls/v3/common.proto
adisuissa: I think this comment wasn't updated, maybe missed a commit.
Please let me know if I'm missing anything.

LuyaoZhong: @adisuissa Whoops.. I forgot it, will update in next change.

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
LuyaoZhong: @ggreenway 

Dig deeper... It seems boringssl hiding the error, there is a FIXME in `X509_verify_cert` implementation:

https://github.com/google/boringssl/blob/d45d8933e61ccd8bd50fcf58f33a604627e6552c/crypto/x509/x509_vfy.c#L228-L233

jmarantz:
/retest

@adisuissa @ggreenway this looks ready for another round. Thanks!

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
soulxu: @LuyaoZhong There is a mistake in your test, the `max_verify_depth` doesn't set currently at https://github.com/envoyproxy/envoy/pull/22350/files#diff-ce04c82dd2b166c6df41e514e7c253ceeb276f7db1d186ddc16f8f1841ad9496R1391

In the end the `max_verify_depth` missed in the downstream transport config.

That is why your test doesn't work.

test/config/utility.cc
```diff
@@ -1373,15 +1373,23 @@ void ConfigHelper::initializeTls(
...
+      } else {
+        cert_yaml = R"EOF(
+        trusted_ca:
+          filename: "{{ test_rundir }}/test/config/integration/certs/intermediate_partial_ca_cert_chain.pem"
       )EOF";
       }
+      if (options.max_verify_depth_.has_value()) {
+        cert_yaml += R"EOF(
+        max_verify_depth:
+      )EOF" + std::to_string(options.max_verify_depth_.value());
```
soulxu, line 1391: This doesn't work, in the end you got a the value in the next line, like
```
    max_verify_depth:
1

```

The expected output should be:
"""
    max_verify_depth: 1 
"""

soulxu (changes requested):
The generated config doesn't include the `max_verify_depth` in the end, after fixing that, it should work as your expected.

test/config/utility.cc
LuyaoZhong: I think you're right, I did have set it correctly before, but I don't know when I change it back... 

Thanks a lot. I'll try again.

LuyaoZhong: Done

api/envoy/extensions/transport_sockets/tls/v3/common.proto
LuyaoZhong: done

api/envoy/extensions/transport_sockets/tls/v3/common.proto
```diff
@@ -477,10 +477,10 @@ message CertificateValidationContext {
   // certificate chain will be subject to validation by :ref:`CRL <envoy_v3_api_field_extensions.transport_sockets.tls.v3.CertificateValidationContext.crl>`.
   bool only_verify_leaf_cert_crl = 14;
 
-  // Config for the max number of intermediate certificates in chain that are parsed during verification.
-  // This does not include the leaf certificate. If configured, and the certificate chain is longer than allowed, the certificates
-  // above the limit are ignored, and certificate validation will fail. The default limit is 100,
-  // though this can be system-dependent.
-  // https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify_depth.html
+  // Defines the max number of intermediate certificates in the chain that are parsed during verification.
+  // This does not include the leaf certificate. If a trusted issuer appears in the chain, but in a depth larger than configured, the certificate validation will fail.
```
soulxu, line 481: instead of saying `This does not include the leaf certificate.`, we can say that `the max_verify_depth defines the limit on the number of certificates between the end-entity and trust-anchor certificates. the trust-anchor was specified at 'trust_ca' https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-certificatevalidationcontext`

test/config/utility.cc
```diff
@@ -1397,15 +1397,22 @@ void ConfigHelper::initializeTls(
...
-        max_verify_depth: 1
+      } else {
+        cert_yaml = R"EOF(
+        trusted_ca:
+          filename: "{{ test_rundir }}/test/config/integration/certs/intermediate_partial_ca_cert_chain.pem"
       )EOF";
       }
+      if (options.max_verify_depth_.has_value()) {
+        cert_yaml += R"EOF(
+        max_verify_depth: )EOF" + std::to_string(options.max_verify_depth_.value());
```
soulxu, line 1414: in case people make the same mistake in the future :) you can move this after the line 1416, then set max_verify_depth by `validation_context.set_max_verify_depth()`

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
```diff
@@ -74,9 +74,42 @@ TEST_P(SslCertValidatorIntegrationTest, CertValidationFailed) {
...
+  initialize();
+  auto conn = makeSslClientConnection({});
+  IntegrationCodecClientPtr codec = makeRawHttpConnection(std::move(conn), absl::nullopt);
+  ASSERT_TRUE(codec->connected());
+  test_server_->waitForCounterGe(listenerStatPrefix("ssl.handshake"), 1);
+  EXPECT_EQ(test_server_->counter(listenerStatPrefix("ssl.fail_verify_error"))->value(), 0);
+  codec->close();
+}
+
+// With only root ca trusted, certificate validation is expected to fail with max depth,
```
soulxu, line 99: could you add more comments about the server/client cert chain? that help people read this test in the future.

dynajoe:
> Before proceeding with this, we need to conclude discussion in the issue to make sure this won't have any negative side effects.
> 

What's the general mood on having this become part of main?

test/extensions/transport_sockets/tls/cert_validator/default_validator_integration_test.cc
LuyaoZhong: done

LuyaoZhong (commented):
Sorry for delayed response.

api/envoy/extensions/transport_sockets/tls/v3/common.proto
```diff
@@ -484,10 +488,13 @@ message CertificateValidationContext {
...
-  // Config for the max number of intermediate certificates in chain that are parsed during verification.
-  // This does not include the leaf certificate. If configured, and the certificate chain is longer than allowed, the certificates
-  // above the limit are ignored, and certificate validation will fail. The default limit is 100,
-  // though this can be system-dependent.
-  // https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify_depth.html
+
+  // Defines maximum depth of a certificate chain accepted in verification. This number does not include the leaf, so a depth of 1 allows the leaf and one CA certificate.
+  // If a trusted issuer appears in the chain, but in a depth larger than configured, the certificate validation will fail.
+  // See `BoringSSL SSL_CTX_set_verify_depth <https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_verify_depth>`
+  // If you use OpenSSL, its behavior is different from BoringSSL, this will define a limit on the number of certificates between the end-entity and trust-anchor certificates.
```
soulxu, line 495: emm...I wasn't sure about this since it passes a long time, I'm a little forget about the background. need to double check

LuyaoZhong: I just found this difference today, the comment mixed boringssl behavior and openssl behavior before. They are not the same
BoringSSL:
This number does not include the leaf

OpenSSL:
Neither the end-entity nor the trust-anchor certificates count against depth. If the certificate chain needed to reach a trusted issuer is longer than depth+2

you can check the behavior of testcase https://github.com/envoyproxy/envoy/pull/22350/files#diff-f86b77f38f9ce6f2295761c59e534cb90cf8c095454db7106e6d66cd87e209d3R133, it should fail with openssl but succeed with boringssl.

soulxu: have you actually test that on the openssl?

If yes, there are probably two things need to think about. One is the test you added is going to break https://github.com/envoyproxy/envoy-openssl. The second is whether we want to expose different behavior in the API, maybe we can let the OpenSSL match the same behavior like it will process the depth as `max_depth+1`.

But I don't know whether those things should be processed in the Envoy main repo, or it should be done at Envoy OpenSSL repo. Those probably need maintainer help here.

LuyaoZhong: No, I didn't test it with openssl. I just copy the official documentation here. @ggreenway I need your help, how should we do for openssl?

wrowe:
/retest

as there is no current CI result.

wrowe:
It seems this is blocked on CodeQL, merge main should unblock CI, @LuyaoZhong

KBaichoo:
seems like CI is stuck and this needs a merge from main @LuyaoZhong 

/wait

soulxu:
/wait

This probably need maintainer feedback on this https://github.com/envoyproxy/envoy/pull/22350#discussion_r1001224029 before merge.

api/envoy/extensions/transport_sockets/tls/v3/common.proto
ggreenway: I'm not sure if this is the same issue or not, but is this due to https://bugs.chromium.org/p/boringssl/issues/detail?id=533 (which refers to https://github.com/openssl/openssl/commit/0daccd4dc1f1ac62181738a91714f35472e50f3c)?

I wouldn't block this change on the potential to break envoy-openssl.

LuyaoZhong: It seems not due the issue https://bugs.chromium.org/p/boringssl/issues/detail?id=533.

The issue here is the 'depth' has different means in boringssl and openssl,

in boringssl, only leaf cert not count against depth. There are test cases verifying its behavior.
in openssl, neither leaf cert nor trust-anchor count against depth. I did not test it, just copy the openssl documentation to let users aware of the difference. Is that enough?

LuyaoZhong: @ggreenway ping.

ggreenway: Yes, I think documenting the difference is good enough.

ggreenway (changes requested):
This looks good overall. I think we should add a runtime flag (default enabled) to disable the new behavior in the off chance this causes an issue for someone. Other than that, I think this looks good.

/wait

api/envoy/extensions/transport_sockets/tls/v3/common.proto
```diff
@@ -484,10 +488,12 @@ message CertificateValidationContext {
...
-  // This does not include the leaf certificate. If configured, and the certificate chain is longer than allowed, the certificates
-  // above the limit are ignored, and certificate validation will fail. The default limit is 100,
-  // though this can be system-dependent.
-  // https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify_depth.html
+  // Defines maximum depth of a certificate chain accepted in verification. This number does not include the leaf, so a depth of 1 allows the leaf and one CA certificate.
+  // If a trusted issuer appears in the chain, but in a depth larger than configured, the certificate validation will fail.
+  // See `BoringSSL SSL_CTX_set_verify_depth <https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_verify_depth>`
+  // If you use OpenSSL, its behavior is different from BoringSSL, this will define a limit on the number of certificates between the end-entity and trust-anchor certificates.
+  // Neither the end-entity nor the trust-anchor certificates count against depth. The default limit is 100, though this can be system-dependent.
```
adisuissa, line 495: Is the default value is only for OpenSSL?
I'm guessing it's also for BoringSSL, so I suggest moving the default-value description before describing the differences between BoringSSL and OpenSSL in the comment.

LuyaoZhong: Yes, you're is right. Thanks.

FYI,
https://github.com/openssl/openssl/blob/7489ada9f3fd902c5bc3c58cc03a90de2800d0ab/crypto/x509/x509_vpm.c#L523
https://github.com/google/boringssl/blob/a1dffbfed9b78d77ab17c15fe2d189ca44704fd8/crypto/x509/x509_vpm.c#L519

adisuissa (commented):
/lgtm api

soulxu (approved):
LGTM, since @ggreenway confirmed different default value between openssl and boringssl is ok.

soulxu (changes requested):
oh, wait, we should test when runtime flag is false.

LuyaoZhong:
> oh, wait, we should add test for runtime flag is false.

@soulxu For this feature, I don't think we need test for runtime flag is false, since the old code path was running for a long time and already tested before.

LuyaoZhong:
> > > oh, wait, we should add test for runtime flag is false.
> > 
> > 
> > @soulxu For this feature, I don't think we need test for runtime flag is false, since the old code path(other than the code path that the runtime flag controls) was running for a long time and already tested before.
> 
> we should keep the old test to ensure there is no regression in the future

Sorry, what regression we might have? How to avoid regression by keeping old test?

soulxu:
> > > > oh, wait, we should add test for runtime flag is false.
> > > 
> > > 
> > > @soulxu For this feature, I don't think we need test for runtime flag is false, since the old code path(other than the code path that the runtime flag controls) was running for a long time and already tested before.
> > 
> > 
> > we should keep the old test to ensure there is no regression in the future
> 
> Sorry, what regression we might have? How to avoid regression by keeping old test?

Sorry, I may not describe it clearly.  We use runtime flag for switch two logic. You add the tests for the new logic, but the old logic still needs to keep here. But if there is no test for the old logic, it could happen regression if someone missed change the code. Just like, you already test the new code today and the code works well, but it doesn't mean you can remove those tests for the new code, you still need to keep the tests for it in case we won't break the code in the future. In the day we deprecate the runtime flag, then we can remove the old logic/code and the old tests.

LuyaoZhong:
> Sorry, I may not describe it clearly. We use runtime flag for switch two logic. You add the tests for the new logic, but the old logic still needs to keep here. But if there is no test for the old logic, it could happen regression if someone missed change the code. Just like, you already test the new code today and the code works well, but it doesn't mean you can remove those tests for the new code, you still need to keep the tests for it in case we won't break the code in the future. In the day we deprecate the runtime flag, then we can remove the old logic/code and the old tests.

@soulxu Got it. Thanks.

ggreenway (approved):
LGTM. Thanks!

PR #22353, opened 2022-07-21
Added big-endian support into TestByteOrder

knm3000:


test/common/network/utility_test.cc
```diff
@@ -530,22 +530,28 @@ TEST(PortRangeListTest, Normal) {
   }
 }
 
-// TODO(ccaraman): Support big-endian. These tests operate under the assumption that the machine
-// byte order is little-endian.
 TEST(AbslUint128, TestByteOrder) {
+#if defined(ABSL_IS_BIG_ENDIAN)
+  auto BSwapIfRequired = [](const absl::uint128& input) {
```
htuch, line 535: * This should be something like `flip_order_for_endianess`, it's a local definition.
* There already is a `Network::Utility::flipOrder` you can use.

htuch (approved):
LGTM, thanks!

PR #22358, opened 2022-07-21
http: refactoring all the templates to allow for the creation of upst…

alyssawilk:
Changing the filter factory classes to allow for upstream http factories, which take a ServerFactoryContext rather than a FactoryContext.

This also changes the type of isTerminalFilter.* which is not strictly speaking necessary, but makes for less copy-pasted template code.

Risk Level: low
Testing: existing tests
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/filter/config_discovery_impl.h
+++ b/source/common/filter/config_discovery_impl.h
@@ -70,8 +70,7 @@ class DynamicFilterConfigProviderImpl : public DynamicFilterConfigProviderImplBa
 public:
   DynamicFilterConfigProviderImpl(
       FilterConfigSubscriptionSharedPtr& subscription,
-      const absl::flat_hash_set<std::string>& require_type_urls,
-      ThreadLocal::SlotAllocator& tls,
+      const absl::flat_hash_set<std::string>& require_type_urls, ThreadLocal::SlotAllocator& tls,
       ProtobufTypes::MessagePtr&& default_config, bool last_filter_in_filter_chain,
       const std::string& filter_chain_type, absl::string_view stat_prefix,
       const Network::ListenerFilterMatcherSharedPtr& listener_filter_matcher)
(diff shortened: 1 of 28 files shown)


alyssawilk:
(I could also do this without reworking the type of the terminal filter callbacks, but it involved more template gorp and none of the implementations needed the full factory context)

alyssawilk:
Matt, PTAL Friday if you're up for it.  The code is in the final state, modulo a possible minor refactor due to the fix format he

@yanavlasov do you know why we disallow throw in .h files?  I do not want to rewrite this entire class to use status just because making it a template made me move code from a .cc to an .h file.  Is it possible to add an exception to these checks, or will it be problematic in which case I'll probably add a static utility function I can stick in the cc file which does the throw.

alyssawilk:
Ok, looks like I could add the file to EXCEPTION_ALLOWLIST but working around it for now just to get CI to run.

source/common/filter/config_discovery_impl.h
```diff
@@ -173,40 +172,42 @@ struct NamedHttpFilterFactoryCb {
 };
 
 // Implementation of a HTTP dynamic filter config provider.
+template <class FactoryCtx, class NeutralHttpFilterConfigFactory>
```
snowp, line 175: Should Neutral be `Named`? Or this is a new concept introduced?

source/common/http/filter_chain_helper.h
```diff
@@ -13,48 +13,146 @@
 namespace Envoy {
 namespace Http {
 
-using FilterConfigProviderManager =
+using DownstreamFilterConfigProviderManager =
     Filter::FilterConfigProviderManager<Filter::NamedHttpFilterFactoryCb,
                                         Server::Configuration::FactoryContext>;
+using UpstreamFilterConfigProviderManager =
+    Filter::FilterConfigProviderManager<Filter::NamedHttpFilterFactoryCb,
+                                        Server::Configuration::ServerFactoryContext>;
```
snowp, line 21: So this means the upstream ones get passed two ServerFactoryContext - is this just temporary until we get an upstream specific factory context?

snowp (commented):
Nice!

source/common/filter/config_discovery_impl.h
alyssawilk: It's upstream/downstream neutral.  Can rename to UpstreamOrDownstreamHttpFilterConfigFactory if that's more clear?

source/common/http/filter_chain_helper.h
alyssawilk: Correct.  90% sure we're going to want an UpstreamFactoryContext too, so that we can hand it the cluster scope and cluster init manager to not leak stats and correctly wait on initialization when we reload clusters.

source/common/filter/config_discovery_impl.h
snowp: Neutral is fine with an explanation. Maybe a comment on top so it's possible to tell what this is without paying the cost of huge type name?

alyssawilk:
CI finally happy.  Let me know if you want any other changes!

snowp (approved):
lgtm, thanks!

PR #22364, opened 2022-07-22
changelog: fix misplaced log

giantcroc:
Signed-off-by: giantcroc <changran.wang@intel.com>

Commit Message: move misplaced logs to current.yaml
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax (approved):
hmm, trying to think the best way forward here, this file should never have been changed

i think at least we need some CI so you can only ever change current.yaml without an error

anyhow, thanks for cleanup @giantcroc

PR #22367, opened 2022-07-22
stream_info: add downstream handshake completion timing

pksohn:
Signed-off-by: Paul Sohn <paulsohn@google.com>

Commit Message: Adds a downstreamTiming field to the connection-level streamInfo for when the TLS handshake completed for HTTP 1/2/3, for monitoring handshake latency. 
Additional Description: Note that for HTTP/3 zero-RTT, the handshake may complete well into the life of the request, which is why this field is not copied into the stream-level streamInfo when a stream is created.
Risk Level: Low
Testing: Integration tests included in the PR.
Docs Changes: None
Release Notes: Added


phlax:
@pksohn this needs some formatting fixes

you can use the diff/artefacts here to fix

https://dev.azure.com/cncf/envoy/_build/results?buildId=113538&view=results

pksohn:
@ggreenway the checks are finally passing for this PR. Any chance you might be able to take a look? Thanks!

changelogs/current.yaml
```diff
@@ -50,6 +50,49 @@ new_features:
 - area: dns_resolver
   change: |
     added DNS stats for c-ares DNS resolver. Detailed documentation is available :ref:`here <arch_overview_dns_resolution>`.
+    added :ref:`include_unroutable_families<envoy_v3_api_field_extensions.network.dns_resolver.apple.v3.AppleDnsResolverConfig.include_unroutable_families>`
```
ggreenway, line 53: Looks like you picked up some extra release note changes. Bad merge? Please resolve.

ggreenway (changes requested):
Overall looks similar to the upstream handshake timing. How are you planning to access it? In a separate change to add it to access logging or similar?

@alyssawilk it looks like you added the upstream handshake timing. Any comments?

changelogs/current.yaml
pksohn: Done, thanks.

pksohn:
@ggreenway @alyssawilk One thing I meant to ask about:

I removed the assertion that the handshake timestamp can only be written once, in this commit: https://github.com/envoyproxy/envoy/pull/22367/commits/61a8618bc246063886be9eba6ecbd4e19521b57e, upon discovering that the SslSocketTest::ClientSessionResumption... tests were failing due to SslSocket::onSuccess being called multiple times. 

In practice do we actually expect SslSocket::onSuccess to be called multiple times for the same connection (e.g. on resumption)? If so, would we want the handshake-complete timestamp to be overwritten?

envoy/stream_info/stream_info.h
```diff
@@ -304,6 +311,8 @@ class DownstreamTiming {
   absl::optional<MonotonicTime> first_downstream_tx_byte_sent_;
   // The time when the last byte of the response was sent downstream.
   absl::optional<MonotonicTime> last_downstream_tx_byte_sent_;
+  // The time the TLS handshake completed. Set at connection level.
+  absl::optional<MonotonicTime> downstream_handshake_complete_;
```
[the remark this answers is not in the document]
pksohn: There was some concern about the handshake timing being confusing to log at the request level, so the thought was to leave this at the connection level for possible use by custom logging/monitoring L7 filters. If it makes more sense to just copy to stream-level info, I'm happy to add that here or in a follow-up. Do you have a preference?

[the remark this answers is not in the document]
alyssawilk: As said I'm fine with it at the connection level, but I'm not sure how you're planning to log it? Do you have a connection level logger?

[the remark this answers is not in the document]
alyssawilk: offline, using L4 logging so this SGTM

alyssawilk (dismissed):
throwing over to Greg for a second review

alyssawilk:
@ggreenway ping!

changelogs/current.yaml
```diff
@@ -62,6 +62,9 @@ new_features:
 - area: gzip
   change: |
     added support for :ref:`max_inflate_ratio<envoy_v3_api_msg_extensions.compression.gzip.decompressor.v3.Gzip>`.
+- area: http
```
ggreenway, line 65: Based on discussion about l4 vs l7 logging, I think the `area` is incorrect. Maybe it should be `access_log`?

ggreenway (changes requested):
I think the code looks fine. 

In terms of whether this can be logged from an http request, I think there's plenty of prior art around transport-socket fields being loggable from http requests. For example `%DOWNSTREAM_TLS_SESSION_ID%`. 

Would you be willing to add an access log formatter in `source/common/formatter/substitutions_formatter.cc`, at least for tcp-proxy access logs? I think this new field is generally useful so I'd like to make it more easily available.

changelogs/current.yaml
pksohn: Done, thanks.

pksohn:
>Would you be willing to add an access log formatter in source/common/formatter/substitutions_formatter.cc, at least for tcp-proxy access logs? I think this new field is generally useful so I'd like to make it more easily available.

Thanks -- I think I've added what you were suggesting.

docs/root/configuration/observability/access_log/usage.rst
```diff
@@ -397,6 +397,18 @@ The following command operators are supported:
 
   Renders a numeric value in typed JSON logs.
 
+%DOWNSTREAM_HANDSHAKE_DURATION%
+  HTTP
+    Not implemented ("-").
```
ggreenway, line 402: This could be treated as a generic transport socket field, as it is now, and be tcp-only. But if you moved it into the `sslConnection()` and made it TLS-specific, it would work for http requests as well, without any additional work to copy it from connection to request StreamInfo.  (See `StreamInfoSslConnectionInfoFieldExtractor` in substitution_formatter.cc). I'd be inclined to move it to TLS-specific, unless you think someone is going to implement this on other transport sockets. WDYT? Also @alyssawilk any opinion here?

pksohn: That works, my only argument for leaving as-is is that it's nicely grouped and discoverable with other timing metrics in downstreamTiming. Another option is to thread it through to copy to streaminfo (shouldn't be too tricky if we think it's worth it). I'm pretty new to Envoy so defer to you and Alyssa on this one.

jmarantz: @ggreenway @alyssawilk can you give @pksohn some guidance on this so we can move this forward? Thanks!

jmarantz:
@pksohn can you fix CI? Then I'll ping the reviewers again to try to make some progress.

/wait

pksohn:
@alyssawilk: there was a question above that we were hoping you could weigh in on (namely whether to move this metric into the SSL connection or leave it in downstream timing): https://github.com/envoyproxy/envoy/pull/22367#discussion_r941792000

docs/root/configuration/observability/access_log/usage.rst
alyssawilk: May be a question for @ianswett .  I believe that Google will want timing metrics for LOAS eventually, and while that plug-in is in house and I don't know if we'll upstream the similar ALTS implementation, AFIK it makes sense to keep it generic.

ianswett: Yes, we'd want it for LOAS/ALTS.  I'm not sure if it's relevant, but we also want it for QUIC and HTTP/3.

alyssawilk:
I think you need to merge main to pick up #22792

docs/root/configuration/observability/access_log/usage.rst
pksohn: @ggreenway seems like the consensus so far is to leave the implementation as-is, OK with you?

ggreenway (approved):
LGTM

PR #22371, opened 2022-07-22
original_dst: fix leak of hosts in pools

kyessenov:

Commit Message: Address memory leak in concurrent host additions in ORIGINAL_DST clusters.
Additional Description: This is an attempt to avoid a global mutex on the cluster. The idea is to collect synthetic hosts for the same address and manage them collectively. Unfortunately, there's another issue present: GC might delete a host and post drain request before the connection to that host is fully established and violate an assertion (GC sees unset `used_` right before `chooseHost` returns it).

Risk Level: high
Testing: manual so far, fixes memory leak but another issue uncovered
Docs Changes: none
Release Notes: none

### DIFF
--- a/source/common/upstream/original_dst_cluster.cc
+++ b/source/common/upstream/original_dst_cluster.cc
@@ -45,9 +45,11 @@ HostConstSharedPtr OriginalDstCluster::LoadBalancer::chooseHost(LoadBalancerCont
       // Check if a host with the destination address is already in the host set.
       auto it = host_map_->find(dst_addr.asString());
       if (it != host_map_->end()) {
-        HostSharedPtr host(it->second); // takes a reference
-        ENVOY_LOG(debug, "Using existing host {}.", host->address()->asString());
-        host->used(true); // Mark as used.
+        SyntheticHostsSharedPtr hosts(
+            it->second); // takes a reference in case main deletes the address
+        HostSharedPtr host(hosts->hosts_.front());
+        ENVOY_LOG(trace, "Using existing host {}.", host->address()->asString());
+        hosts->used_ = true; // Mark as used (TODO: will this leak if main concurrenty deleted it?)
         return host;
       }
       // Add a new host
@@ -138,7 +140,7 @@ OriginalDstCluster::OriginalDstCluster(
       cleanup_interval_ms_(
           std::chrono::milliseconds(PROTOBUF_GET_MS_OR_DEFAULT(config, cleanup_interval, 5000))),
       cleanup_timer_(dispatcher_.createTimer([this]() -> void { cleanup(); })),
-      host_map_(std::make_shared<HostMap>()) {
+      host_map_(std::make_shared<HostMultiMap>()) {
   if (const auto& config_opt = info_->lbOriginalDstConfig(); config_opt.has_value()) {
     if (config_opt->use_http_header()) {
       http_header_name_ = config_opt->http_header_name().empty()
@@ -160,46 +162,48 @@ OriginalDstCluster::OriginalDstCluster(
 }
 
 void OriginalDstCluster::addHost(HostSharedPtr& host) {
-  HostMapSharedPtr new_host_map = std::make_shared<HostMap>(*getCurrentHostMap());
-  auto pair = new_host_map->emplace(host->address()->asString(), host);
-  bool added = pair.second;
+  HostMultiMapSharedPtr new_host_map = std::make_shared<HostMultiMap>(*getCurrentHostMap());
+  auto [it, added] =
+      new_host_map->emplace(host->address()->asString(), std::make_shared<SyntheticHosts>());
   if (added) {
     ENVOY_LOG(debug, "addHost() adding {}", host->address()->asString());
-    setHostMap(new_host_map);
-    // Given the current config, only EDS clusters support multiple priorities.
-    ASSERT(priority_set_.hostSetsPerPriority().size() == 1);
-    const auto& first_host_set = priority_set_.getOrCreateHostSet(0);
-    HostVectorSharedPtr all_hosts(new HostVector(first_host_set.hosts()));
-    all_hosts->emplace_back(host);
-    priority_set_.updateHosts(0,
-                              HostSetImpl::partitionHosts(all_hosts, HostsPerLocalityImpl::empty()),
-                              {}, {std::move(host)}, {}, absl::nullopt);
   }
+  it->second->hosts_.push_back(host);
+  setHostMap(new_host_map);
+  // Given the current config, only EDS clusters support multiple priorities.
+  ASSERT(priority_set_.hostSetsPerPriority().size() == 1);
+  const auto& first_host_set = priority_set_.getOrCreateHostSet(0);
+  HostVectorSharedPtr all_hosts(new HostVector(first_host_set.hosts()));
+  all_hosts->emplace_back(host);
+  priority_set_.updateHosts(0,
+                            HostSetImpl::partitionHosts(all_hosts, HostsPerLocalityImpl::empty()),
+                            {}, {std::move(host)}, {}, absl::nullopt);
 }
 
 void OriginalDstCluster::cleanup() {
   HostVectorSharedPtr keeping_hosts(new HostVector);
   HostVector to_be_removed;
-  ENVOY_LOG(trace, "Stale original dst hosts cleanup triggered.");
+  // ENVOY_LOG(trace, "Stale original dst hosts cleanup triggered.");
   auto host_map = getCurrentHostMap();
   if (!host_map->empty()) {
     ENVOY_LOG(trace, "Cleaning up stale original dst hosts.");
     for (const auto& pair : *host_map) {
       const std::string& addr = pair.first;
-      const HostSharedPtr& host = pair.second;
-      if (host->used()) {
+      if (pair.second->used_) {
         ENVOY_LOG(trace, "Keeping active host {}.", addr);
-        keeping_hosts->emplace_back(host);
-        host->used(false); // Mark to be removed during the next round.
+        keeping_hosts->insert(keeping_hosts->end(), pair.second->hosts_.begin(),
+                              pair.second->hosts_.end());
+        pair.second->used_ = false; // Mark to be removed during the next round.
       } else {
         ENVOY_LOG(trace, "Removing stale host {}.", addr);
-        to_be_removed.emplace_back(host);
+        to_be_removed.insert(to_be_removed.end(), pair.second->hosts_.begin(),
+                             pair.second->hosts_.end());
       }
     }
   }
 
   if (!to_be_removed.empty()) {
-    HostMapSharedPtr new_host_map = std::make_shared<HostMap>(*host_map);
+    HostMultiMapSharedPtr new_host_map = std::make_shared<HostMultiMap>(*host_map);
     for (const HostSharedPtr& host : to_be_removed) {
       new_host_map->erase(host->address()->asString());
     }
--- a/source/common/upstream/original_dst_cluster.h
+++ b/source/common/upstream/original_dst_cluster.h
@@ -18,8 +18,18 @@
 namespace Envoy {
 namespace Upstream {
 
-using HostMapSharedPtr = std::shared_ptr<HostMap>;
-using HostMapConstSharedPtr = std::shared_ptr<const HostMap>;
+// Container for synthetic hosts produced by workers. Because each worker binds
+// the connection pool to the pointer value of a host, there may be multiple
+// Host structs for the same address.
+struct SyntheticHosts {
+  std::list<Upstream::HostSharedPtr> hosts_;
+  std::atomic<bool> used_{true};
+};
+
+using SyntheticHostsSharedPtr = std::shared_ptr<SyntheticHosts>;
+using HostMultiMap = absl::flat_hash_map<std::string, SyntheticHostsSharedPtr>;
+using HostMultiMapSharedPtr = std::shared_ptr<HostMultiMap>;
+using HostMultiMapConstSharedPtr = std::shared_ptr<const HostMultiMap>;
 
 /**
  * The OriginalDstCluster is a dynamic cluster that automatically adds hosts as needed based on the
@@ -77,7 +87,7 @@ public:
     const std::shared_ptr<OriginalDstCluster> parent_;
     // The optional original host provider that extracts the address from HTTP header map.
     const absl::optional<Http::LowerCaseString>& http_header_name_;
-    HostMapConstSharedPtr host_map_;
+    HostMultiMapConstSharedPtr host_map_;
   };
 
   const absl::optional<Http::LowerCaseString>& httpHeaderName() { return http_header_name_; }
@@ -105,12 +115,12 @@ private:
     const std::shared_ptr<OriginalDstCluster> cluster_;
   };
 
-  HostMapConstSharedPtr getCurrentHostMap() {
+  HostMultiMapConstSharedPtr getCurrentHostMap() {
     absl::ReaderMutexLock lock(&host_map_lock_);
     return host_map_;
   }
 
-  void setHostMap(const HostMapConstSharedPtr& new_host_map) {
+  void setHostMap(const HostMultiMapConstSharedPtr& new_host_map) {
     absl::WriterMutexLock lock(&host_map_lock_);
     host_map_ = new_host_map;
   }
@@ -126,7 +136,7 @@ private:
   Event::TimerPtr cleanup_timer_;
 
   absl::Mutex host_map_lock_;
-  HostMapConstSharedPtr host_map_ ABSL_GUARDED_BY(host_map_lock_);
+  HostMultiMapConstSharedPtr host_map_ ABSL_GUARDED_BY(host_map_lock_);
   absl::optional<Http::LowerCaseString> http_header_name_;
   friend class OriginalDstClusterFactory;
 };
--- a/source/common/upstream/upstream_impl.h
+++ b/source/common/upstream/upstream_impl.h
@@ -230,8 +230,7 @@ public:
            uint32_t priority, const envoy::config::core::v3::HealthStatus health_status,
            TimeSource& time_source)
       : HostDescriptionImpl(cluster, hostname, address, metadata, locality, health_check_config,
-                            priority, time_source),
-        used_(true) {
+                            priority, time_source) {
     setEdsHealthFlag(health_status);
     HostImpl::weight(initial_weight);
   }
@@ -285,8 +284,6 @@ public:
 
   uint32_t weight() const override { return weight_; }
   void weight(uint32_t new_weight) override;
-  bool used() const override { return used_; }
-  void used(bool new_used) override { used_ = new_used; }
 
 protected:
   static CreateConnectionData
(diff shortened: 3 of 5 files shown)


kyessenov:
CC @wbpcode @mattklein123 
Need more thought on how to deal with racey "new connection" / "drain connection" requests.

kyessenov:
I cannot find what actually protected Hosts from being removed while connecting in other types of clusters. I found `lifetimeCallbacks` that are appears to not actually be used anywhere.

kyessenov:
@alyssawilk do you have any suggestions on how to properly handle host removals while connecting? with ORIGINAL_DST this can happen very rapidly, and I hit the race that drain is called before connection is fully established.

alyssawilk:
without having looked at the code, would it make sense to count the host as in-use, or some other guard against reaping, for some multiple (2) of the cluster connect timeout?

kyessenov:
That should work as long as the connect timeout O(s) not O(ms). I'm a bit worried that relying on the wall clock adds uncertainty (if CPU is starved, we might hit GC and timeout simultaneously). Do you know how EDS protects against immediate host removal?

alyssawilk:
I thought with EDS we kept the hosts around until removal, even if unused?  cc @adisuissa

kyessenov:
A couple of experiments:
* I can't make EDS run fast enough to trigger this issue. It requires two quick updates within micros for a request to pick an upstream host, prepare a request and then drain the upstream host
* This is not an issue with tcp_proxy, just HCM with original_dst upstream.

lambdai:
> That should work as long as the connect timeout O(s) not O(ms). I'm a bit worried that relying on the wall clock adds uncertainty (if CPU is starved, we might hit GC and timeout simultaneously). Do you know how EDS protects against immediate host removal?

EDS cluster type maintains a host vector mainly for LB to pick up, and the next EDS update keeps the desired hosts to live (and to delete).

original_dst cluster doesn't have the luxury so a timer was used.  alternatively could be a weak_ptr style: host map as typed map<TCP_ADDRESS, weak_ptr<Host>>, with an nontrivial host ptr deleter.

kyessenov:
I think the crash is actually due to re-use of memory addresses for hosts! Because we create so many synthetic hosts, addresses happen to be re-used:

```
[2022-08-10 19:04:14.599][25911][debug][upstream] [source/common/upstream/original_dst_cluster.cc:67] Created host 127.0.0.1:8001 0x147d7fd61be8.
[2022-08-10 19:04:14.599][25911][trace][pool] [source/common/conn_pool/conn_pool_base.cc:110] TryCreateNewConnections 0x147d7fd61be8
[2022-08-10 19:04:14.600][25904][trace][upstream] [source/common/upstream/original_dst_cluster.cc:201] Remove host 0x147d7fd61be8
[2022-08-10 19:04:14.600][25911][debug][pool] [source/common/conn_pool/conn_pool_base.cc:326] onUpstreamReady 0x147d7fd61be8
[2022-08-10 19:04:14.601][25911][trace][pool] [source/common/conn_pool/conn_pool_base.cc:110] TryCreateNewConnections 0x147d7fd61be8
[2022-08-10 19:04:14.603][25911][trace][pool] [source/common/conn_pool/conn_pool_base.cc:110] TryCreateNewConnections 0x147d7fd61be8
[2022-08-10 19:04:16.424][25912][debug][upstream] [source/common/upstream/original_dst_cluster.cc:67] Created host 127.0.0.1:8001 0x147d7f161be8.
[2022-08-10 19:04:16.424][25912][trace][pool] [source/common/conn_pool/conn_pool_base.cc:110] TryCreateNewConnections 0x147d7f161be8
[2022-08-10 19:04:16.424][25904][trace][upstream] [source/common/upstream/original_dst_cluster.cc:201] Remove host 0x147d7f161be8
[2022-08-10 19:04:16.425][25912][debug][pool] [source/common/conn_pool/conn_pool_base.cc:326] onUpstreamReady 0x147d7f161be8
[2022-08-10 19:04:16.426][25912][trace][upstream] [source/common/upstream/cluster_manager_impl.cc:1742] Erasing idle pool for host 0x147d7f161be8
[2022-08-10 19:04:16.426][25912][trace][upstream] [source/common/upstream/cluster_manager_impl.cc:1749] Pool container empty for host 0x147d7f161be8, erasing host entry
```

There're two hosts created within 2 seconds with the same memory address. This somehow causes corruption in some data structure I think that then violates the draining invariant.

kyessenov:
Ok, I think the right way to fix this is as follows:
* we keep this fix with concurrent `addHost` for the same address leaking;
* we remove the logic to forcefully drain on `cleanup`; this is surprising and contradicts https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#arch-overview-service-discovery-types-original-destination. Moreover, the crash I am seeing means that sometimes drain/tryNewConnection takes longer than creating two synthetic hosts at the same memory address, which messes up the internal maps from `&host`;
* we switch to remove hosts per thread cluster on idle like EDS;
* we find a way to keep synthetic hosts in memory for `cleanup` duration for GC purposes; not quite sure the best way, weak_ptr is not enough.

kyessenov:
@lambdai @mattklein123 hope the plan above sounds good?

mattklein123:
I just spent some time paging this code back in. Sorry for the delay.

At a high level, I'm not sure all this atomic stuff is worthwhile. We already have a lock in there and grab it when mutating the host map, which actually requires copying the entire thing every time we add/remove a host. If we really have a super high rate of change my suspicion is that the copying will dwarf the lock hold time for map changes. This is ultimately what we saw with DFP. If we don't have a high rate of change, I don't think lock versus/copying matters. Perhaps we should just move to a global lock since it's much simpler and then make it better later if it's really needed?

Now, per your analysis (nice work!) I agree that cleanup() is just broken and breaks a lot of other invariants. I agree we should just remove that functionality entirely and rely on idle timeouts to do the cleanups. Perhaps for back compat we could somehow read the cleanup interval and use that to auto configure the HTTP idle timeout for HTTP cases? I'm not sure about TCP and other cases? How would that work so we avoid leaks? Do we need to keep it there which seems bad? In general though I agree that if we remove cleanup() and rely on the existing cluster manager idle functionality I don't think we need to deal with the reaping issue. (I haven't looked at the code in a while, but at it's core IIRC there are maps in the cluster manager that use the raw pointer of a host as a key. It's possible you could figure out a way to use a weak_ptr somehow, but it would be tricky for sure.)

kyessenov:
I could imagine short TCP connections could benefit from pre-constructed synthetic hosts. But it's not worth it IMHO, the cost of making a new host is small, and having weak_ptr handle deletion is a lot easier to reason about.

mattklein123:
> But it's not worth it IMHO, the cost of making a new host is small, and having weak_ptr handle deletion is a lot easier to reason about.

Well the issue is that the cluster itself stores the host with a strong pointer, so I don't think it would ever get deleted, unless you somehow make the cluster itself only store weak_ptrs, but I don't know that would work. I just don't recall enough of the details here without spending a bunch more time. If it would help I'm happy to video chat to brainstorm solutions.

mattklein123:
Or are you suggesting just not storing the hosts in the cluster at all for TCP since there are no conn pools? Maybe that would work but I'm not sure how we would switch on and off that behavior without it being annoying to configure.

kyessenov:
I have a proposal to improve ORIGINAL_DST in this PR. It goes as follows:
* We extend the cluster manager to keep a reference-counted token while it has active pools.
* ORIGINAL_DST cluster will keep a list of (host, host token weak_ptr) it gives to the cluster manager.
* When a host token expires, the host is deleted from the priority set on the main thread. This is done at both `addHost` and `cleanup` time. The cleanup is purely for bookkeeping when a particular address is no longer chosen.

I think the host token is generally useful even if we go with a locking solution. Without locks, I think it's still possible to create two connections to the same host on one worker (if `addHost` races for the same address and then one worker chooses another worker's host). Whether or not we start locking the host collection, we need to know when it is safe to delete them.

mattklein123:
At a high level this seems like a good path forward, but I need to spend more time on this next week. I'm mostly OOO today. Can you fix CI and I can look on Tuesday? Thank you.

/wait

kyessenov:
Thanks, this can wait. The CI breakage is mostly just the change in signature in `load_balancer.h` (which would be a slog to modify so I wanted some confidence we should do that). Currently, I'm mostly load testing with a full envoy.

kyessenov:
@mattklein123 do you still want me to fix all the tests first? Do you think the interface change to load_balancer.h is good?

mattklein123:
Sorry I forgot about this. I will look more today.

mattklein123:
Discussed this offline with @kyessenov and we iterated to a slightly different solution (but still similar) that will hopefully make this PR smaller and a bit more targeted. Thank you for iterating!

/wait

source/common/upstream/upstream_impl.h
```diff
@@ -304,7 +303,13 @@ class HostImpl : public HostDescriptionImpl,
 
   std::atomic<uint32_t> health_flags_{};
   std::atomic<uint32_t> weight_;
-  std::atomic<bool> used_;
+
+  struct HostHandleImpl : HostHandle {
+    HostHandleImpl(const HostImpl& parent) : parent_(parent) { parent.handle_count_++; }
+    ~HostHandleImpl() override { parent_.handle_count_--; }
```
mattklein123, line 309: ASSERT handle_count_ > 0 before decrementing

source/common/upstream/upstream_impl.h
```diff
@@ -304,7 +303,13 @@ class HostImpl : public HostDescriptionImpl,
...
+    const HostImpl& parent_;
```
mattklein123, line 310: This is probably fine given the use, but would it be better to store a weak_ptr to the real host to avoid use after free?

source/common/upstream/original_dst_cluster.cc
```diff
@@ -160,48 +161,72 @@ OriginalDstCluster::OriginalDstCluster(
...
+          keep = true;
+        } else {
+          for (const auto& host : hosts->hosts_) {
+            if (host->used()) {
+              keep = true;
+              break;
+            }
+          }
+        }
+      }
```
mattklein123, line 209: Can you add more comments in here? It's a little hard for me to understand what is going on now from a quick read. The feature flag also makes it more complicated so if you could touch on what can be removed later when the feature flag goes away that would be good. Naively I'm not sure why we have to actually store used_ on the host tracker in here? Can't that just be inferred during each cleanup iteration?

mattklein123 (changes requested):
Thanks this generally LGTM. Thanks for iterating. A few small comments.

/wait

source/common/upstream/original_dst_cluster.cc
kyessenov: Ok, will add more comments. I wanted to keep the two-stage cleanup because we need a small time gap without a cleanup between `chooseHost` and worker making a handle.

changelogs/current.yaml
```diff
@@ -18,6 +18,12 @@ behavior_changes:
     stream. This supports the server half-close semantics for TCP tunneling with CONNECT as well as bi-directional
     streaming calls. This behavior change can be reverted by setting the
     ``envoy.reloadable_features.http_response_half_close`` runtime flag to false.
+- area: original_dst
+  change: |
+    ORIGINAL_DST cluster will not attempt to remove and drain the stale hosts during cleanup if they are still used by
+    the connection pools. For HTTP pools, please use ``idle_timeout`` to limit the duration of the open upstream
+    connections. This behavior change can be reverted by setting runtime guard
```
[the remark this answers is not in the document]
kyessenov: 1h is the default, I think 5min is fine?

source/common/upstream/upstream_impl.h
kyessenov: Done.

source/common/upstream/upstream_impl.h
kyessenov: Ok.

source/common/upstream/original_dst_cluster.cc
```diff
@@ -166,48 +167,76 @@ OriginalDstCluster::OriginalDstCluster(
...
-      const HostSharedPtr& host = pair.second;
-      if (host->used()) {
-        ENVOY_LOG(trace, "Keeping active host {}.", addr);
-        keeping_hosts->emplace_back(host);
-        host->used(false); // Mark to be removed during the next round.
+    for (const auto& [addr, hosts] : *host_map) {
+      // Address is kept in the cluster if either of the two things happen:
+      // 1) a host has been recently selected for the address; 2) none of the
+      // hosts are currently in any of the connection pools.
+      // The set of hosts for a single address are treated as a unit.
```
mattklein123, line 202: Sorry I still don't understand why we are doing this? Can you add more comments? Is this just for legacy behavior or is there some other reason? Basically can the used_ thing be removed once the feature flag is removed?

source/common/upstream/original_dst_cluster.cc
```diff
@@ -166,48 +167,76 @@ OriginalDstCluster::OriginalDstCluster(
...
+        if (hosts->host_->used()) {
+          keep = true;
+        } else {
+          for (const auto& host : hosts->hosts_) {
+            if (host->used()) {
+              keep = true;
+              break;
+            }
+          }
+        }
```
mattklein123, line 218: Can you add comments here on why we need to check the top level and all the other ones?

source/common/upstream/original_dst_cluster.h
```diff
@@ -18,8 +18,21 @@
...
 
-using HostMapSharedPtr = std::shared_ptr<HostMap>;
-using HostMapConstSharedPtr = std::shared_ptr<const HostMap>;
+struct HostsForAddress {
+  HostsForAddress(HostSharedPtr& host) : host_(host), used_(true) {}
+
+  // Primary host for the address.
+  const HostSharedPtr host_;
+  // Hosts that are added concurrently with host_ are stored in this list.
+  std::vector<HostSharedPtr> hosts_;
```
mattklein123, line 27: I'm still not convinced this is worth it versus just using a lock, given that we have to copy everything over to avoid using the lock? If we want to keep this can you add more comments on what concurrency is involved and how it happens?

source/common/upstream/original_dst_cluster.h
```diff
@@ -57,7 +70,7 @@ class OriginalDstCluster : public ClusterImplBase {
     // Upstream::LoadBalancer
     HostConstSharedPtr chooseHost(LoadBalancerContext* context) override;
     // Preconnecting is not implemented for OriginalDstCluster
-    HostConstSharedPtr peekAnotherHost(LoadBalancerContext*) override { return nullptr; }
+    HostConstSharedPtr peekAnotherHost(LoadBalancerContext*) override { return {nullptr}; }
```
mattklein123, line 73: revert

mattklein123 (changes requested):
Few more comments, thanks

/wait

source/common/upstream/original_dst_cluster.cc
kyessenov: Here's a sequence I'm worried about:
1) worker1 picks a host H in `chooseHost`
2) main runs cleanup that calls drain on H
3) worker1 creates a handle for H

If we continue using this `used_` bool then 2) would not happen between 1)-3), as long as time between 1)-3) is less than `cleanup_interval` (which it is).

Happy to hear any other way around it? It doesn't really hurt to preserve the legacy behavior AFAICT.

source/common/upstream/original_dst_cluster.h
kyessenov: The lock is still there for the big table. The fine lock is just not necessary since it's fine for each worker to create their own host. We just need to keep track of all of them. Grabbing a lock on table entry has hard-to-measure effects on performance.

source/common/upstream/original_dst_cluster.h
kyessenov: Thanks.

source/common/upstream/original_dst_cluster.cc
kyessenov: Done.

source/common/upstream/original_dst_cluster.cc
mattklein123: > If we continue using this used_ bool then 2) would not happen between 1)-3), as long as time between 1)-3) is less than cleanup_interval (which it is).

How does the time interval matter here? If time matters I think it could still race? Or the boolean will prevent cleanup?

I guess in general I'm still thinking that switching to locking would be easier to reason about and it's unclear to me that there will be much/any perf delta?

kyessenov: The boolean adds a minimum delay (seconds?). It'll take a really slow execution to execute these two lines with seconds delay:
```
auto h = lb.chooseHost(&context);
auto handle = h.acquireHandle();
```

kyessenov: If we want to use locking, we'd have to rewrite most of this cluster. That is likely not backwards compatible in one way or another, so I'm worried that it's better to create another cluster with better characteristics instead.

mattklein123: Alright fair enough. Let's stick with this.

source/common/upstream/original_dst_cluster.cc
```diff
@@ -166,48 +167,82 @@ OriginalDstCluster::OriginalDstCluster(
...
-                              HostSetImpl::partitionHosts(all_hosts, HostsPerLocalityImpl::empty()),
-                              {}, {std::move(host)}, {}, absl::nullopt);
+  std::string address = host->address()->asString();
+  HostMultiMapSharedPtr new_host_map = std::make_shared<HostMultiMap>(*getCurrentHostMap());
+  auto it = new_host_map->find(address);
+  if (it != new_host_map->end()) {
+    it->second->hosts_.push_back(host);
+  } else {
+    new_host_map->emplace(address, std::make_shared<HostsForAddress>(host));
   }
```
mattklein123, line 177: Sorry to be really annoying, but this code is very confusing and I just want to make sure the next person doesn't have to take as long to understand it. Can you add comments here also about these 2 cases in terms of which concurrency leads to what branch? I get it now but would rather it be more clear to the next reader.

source/common/upstream/original_dst_cluster.cc
```diff
@@ -166,48 +167,82 @@ OriginalDstCluster::OriginalDstCluster(
...
-        host->used(false); // Mark to be removed during the next round.
+    for (const auto& [addr, hosts] : *host_map) {
+      // Address is kept in the cluster if either of the two things happen:
+      // 1) a host has been recently selected for the address; 2) none of the
+      // hosts are currently in any of the connection pools.
+      // The set of hosts for a single address are treated as a unit.
+      //
+      // Using the used_ bit is preserved for backwards compatibility and to
+      // add a delay between load balancers choosing a host and grabbing a
+      // handle on the host.
```
mattklein123, line 206: Can you spell out the interleaving that you put in the comment thread here? Basically that the used_ bit "gaurantees" that for at least one cleanup interval the host will remain. Obviously there is no real hard guarantee, but it's very likely given the time values involved.

mattklein123 (changes requested):
Thanks LGTM modulo remaining comment comments.

/wait

source/common/upstream/original_dst_cluster.cc
kyessenov: Ack. LMK if this is sufficient, I spent too much time grokking this logic myself.

source/common/upstream/original_dst_cluster.cc
kyessenov: Ack.

mattklein123 (approved):
Thanks!

PR #22373, opened 2022-07-22
tls: set BIO::num to the underlying socket file descriptor

yzhao1012:
Signed-off-by: yzhao1012 <yzhao@pixielabs.ai>

Commit Message: Set BIO::num from IoHandle::fdDoNotUse()
Additional Description: This is to make sure it's a tracer can read the file descriptor from BIO::num.
Risk Level: Low
Testing: Manually built envoy and examine the results
Docs Changes: None
Release Notes: None
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/transport_sockets/tls/io_handle_bio.cc
+++ b/source/extensions/transport_sockets/tls/io_handle_bio.cc
@@ -133,7 +133,7 @@ BIO* BIO_new_io_handle(Envoy::Network::IoHandle* io_handle) {
   RELEASE_ASSERT(b != nullptr, "");
 
   // Initialize the BIO
-  b->num = -1;
+  b->num = io_handle->fdDoNotUse();
   b->ptr = io_handle;
   b->shutdown = 0;
   b->init = 1;


ggreenway:
Can you elaborate on what you're trying to do with this change?

/wait-any

yzhao1012:
> Can you elaborate on what you're trying to do with this change?
> 
> /wait-any

* eBPF [0] is a Linux Kernel technology. An eBPF program is executed by the eBPF VM [1] inside Linux Kernel.
* Pixie [2] is an Observability platform. Pixie deploys eBPF programs to collect network IO data from inside Linux Kernel.
* To collect clear text data from Envoy-encrypted connections, Pixie uses uprobes [3] attached onto `SSL_{write,read}` APIs  and collect clear-text data from the data buffer given to the API calls' arguments.
* Pixie's uprobes need to obtain the FD (file descriptor) of the socket connection. The FD is used to resolve remote IP address through netlink [4].
* Usually the FD is written to SSL::{rbio,wbio}.num field. But Envoy is not setting this field.
* This diff therefore tries to set that field correctly.

[0] https://ebpf.io/
[1] https://lwn.net/Articles/599755/
[2] https://px.dev/
[3] https://www.brendangregg.com/blog/2016-02-08/linux-ebpf-bcc-uprobes.html
[4] https://man7.org/linux/man-pages/man7/netlink.7.html

ggreenway:
But this BIO isn't one of the boringssl built-in ones that operates on an fd. This is a custom implementation. 

What code will read this value back out? Help me understand exactly how this gets used.

yzhao1012:
Let me know if you are available for a video call, I am more than happy to quickly go through what I am working on.
Should take <30 mins.

My email is yzhao@pixielabs.ai.

> But this BIO isn't one of the boringssl built-in ones that operates on an fd. This is a custom implementation.

IIUC, there is only one BIO data structure.

TLS uses different IO handling setups, BIO::ptr stores a custom IoSoecketHandleImpl pointer. BIO::method is assigned from https://github.com/google/boringssl/blob/955ef7991e41ac6c0ea5114b4b9abb98cc5fd614/crypto/bio/bio.c#L72 and https://github.com/envoyproxy/envoy/blob/482050b90aaa55d970d91712fa90bd98e156128b/source/extensions/transport_sockets/tls/io_handle_bio.cc#L129

> What code will read this value back out? Help me understand exactly how this gets used.

Some custom eBPF program will read this value. One such example is https://github.com/pixie-io/pixie/blob/410c1fb2d867da87ac20beb64f5e6e23208bf6ad/src/stirling/source_connectors/socket_tracer/bcc_bpf/openssl_trace.c#L65

Other eBPF program will benefit from this in the same way, i.e., able to read filedescriptor from rbio directly, at the time when the uprobe attached to SSL_{write,read} is triggered.

ggreenway:
Very interesting; I have a rough understanding now of what you're trying to accomplish.

Will your code work with BoringSSL (Envoy doesn't use OpenSSL)? It's not ABI-compatible with OpenSSL.

If you have to dig through memory to find the fd, can you just write some eBPF code that searches through Envoy data structures to find the fd, instead of needing to copy the value to a location that serves no other purpose than to be found from outside the process?

yzhao1012:
> Will your code work with BoringSSL (Envoy doesn't use OpenSSL)? It's not ABI-compatible with OpenSSL.

We'll need adjustment to eBPF program to work with boringssl and openssl.
The main thing is to use different offsets when chasing pointers from the function arguments.

> If you have to dig through memory to find the fd, can you just write some eBPF code that searches through Envoy data structures to find the fd, instead of needing to copy the value to a location that serves no other purpose than to be found from outside the process?

eBPF program runs in the VM inside Kernel, and it has no information about the application process' own memory space. We try to mimic debugger in eBPF program, i.e., read application process' executable's debug info to determine how to read application process' memory variables starting from where eBPF program is invoked. I.e., when an eBPF program gets invoked, it has access to registers of the process, so the eBPF program can get the pointers to the input argument, and from there the (nested) member variables of the argument, and so forth.

In this particular case, we precomputed the offsets for different versions of openssl releases, so the eBPF programs get the offsets before runtime. In other words, no search is done during the runtime.

ggreenway:
> In this particular case, we precomputed the offsets for different versions of openssl releases, so the eBPF programs get the offsets before runtime. In other words, no search is done during the runtime.

Can you make the same logic for envoy (different versions, etc) and chase the appropriate pointers at the appropriate offsets?

One of my concerns with this PR as-is is that it is using a function that we're trying to remove. When it is removed, there won't be a way to populate the fd in the BIO anymore. 

My other concern is that anyone looking at the code may analyze use in envoy and not see any reason to set the BIO fd field and remove that code because it appears unused.

So that's why I'm probing for other ways to solve this.

yzhao1012:
> using a function that we're trying to remove.

What's the motivation of removing fdDoNotUse() and completely hide fd from callers of IoHandle?

I thought some IoHandle's subclasses actually have not implemented `fdDoNotUse()`, but IoSocketHandleImpl does return the correct fd. So I guessed that this might still be a question of removing fdDoNotUse().

ggreenway:
> What's the motivation of removing fdDoNotUse() and completely hide fd from callers of IoHandle?

Some things we communicate with aren't OS sockets. Examples of this are internal (within the envoy process) connections, and there has been work to use something like VPP/DPDK to do IO and bypass the kernel.

So we've worked to move all code that needs to handle IO into the IOHandle and have an interface to that, so that things outside this layer don't need to know what the underlying implementation is.

yzhao1012:
> What's the motivation of removing fdDoNotUse() and completely hide fd from callers of IoHandle?
> 
> Some things we communicate with aren't OS sockets. Examples of this are internal (within the envoy process) connections, and there has been work to use something like VPP/DPDK to do IO and bypass the kernel.
> 
> So we've worked to move all code that needs to handle IO into the IOHandle and have an interface to that, so that things outside this layer don't need to know what the underlying implementation is.

Makes sense. I'll close this PR. We'll investigate more and see what could be done.

Issue #13009, opened 2020-09-08
xds: Consistency while handling CDS updates and EDS


sschepens:
When a cluster is created or updated envoy it enters warming phase and needs a related ClusterLoadAssignement response to fully initialize.
During Envoy startup phase envoy sends requests for those resources to the management server and so the management servers knows it has to respond those.
But, when updating a Cluster via CDS, no EDS re-request is sent to the management server and management server doesn't really know it should send a ClusterLoadAssignement for that Cluster, even if the resource hasn't really changed.
This berhavior introduces subtle bugs in management servers, in our case, our resource versioning scheme somehow included the cluster version, yesterday we introduced a change to remove that, and that unexpectedly broke our cluster updates, leaving some clusters without traffic.

This should probably be handled by envoy, currently go-control-plane and java-control-plane don't really handle this since it's kind of hard to induce this behavior of always pushing an EDS update for a Cluster even if the resource hasn't really changed, and if not using ADS, envoy is possibly connected to multiple management servers which increases the difficulty.

@htuch suggested that envoy should probably unsubscribe from EDS for the updated cluster and immediately subscribe again, since this is what's actually happening inside envoy, a new Cluster is being created and it wants to subscribe to the resources, and the old one wants to unsubscribe. Some more thought should be done on this idea and the consecuences of the old cluster being unsubscribed from resources.

Some discussion already happened on Slack, opening this issue to continue discussion.

This issue also applies to LDS updates with RDS and probably SRDS with RDS.

htuch:
CC @envoyproxy/api-shepherds

markdroth:
> But, when updating a Cluster via CDS, no EDS re-request is sent to the management server and management server doesn't really know it should send a ClusterLoadAssignement for that Cluster, even if the resource hasn't really changed.

Why would the management server need to send an update for the EDS resource that has not changed, just because the CDS resource that pointed to it has changed?  This seems like a pretty clear Envoy bug: the CDS and EDS resources should be tracked independently from the perspective of the xDS client code, so that when a new Cluster object is created, it can just pick up the existing data for the EDS resource it wants.

[...]

PR #22377, opened 2022-07-23
clarify resource warming docs in xds protocol

ramaraochavali:
Clusters expect EDS response even if they do not change so EDS response is mandatory to finish warming. But listener use route configuration that Envoy is aware of and response is only needed for new routes. This PR clarifies that part in resource warming docs.

Commit Message:clarify resource warming docs in xds protocol
Additional Description:clarify resource warming docs in xds protocol
Risk Level:N/A
Testing:N/A
Docs Changes:N/A
Release Notes:N/A
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/docs/root/api-docs/xds_protocol.rst
+++ b/docs/root/api-docs/xds_protocol.rst
@@ -669,13 +669,15 @@ Resource warming
 go through warming before they can serve requests. This process
 happens both during :ref:`Envoy initialization <arch_overview_initialization>`
 and when the ``Cluster`` or ``Listener`` is updated. Warming of
-``Cluster`` is completed only when a ``ClusterLoadAssignment`` response
-is supplied by management server. Similarly, warming of ``Listener`` is
-completed only when a ``RouteConfiguration`` is supplied by management
-server if the listener refers to an RDS configuration. Management server
-is expected to provide the EDS/RDS updates during warming. If management
-server does not provide EDS/RDS responses, Envoy will not initialize
-itself during the initialization phase and the updates sent via CDS/LDS
+``Cluster`` is completed only when a new ``ClusterLoadAssignment`` response
+is supplied by management server even though there is no change in endpoints.
+Similarly, warming of ``Listener`` is completed only when a ``RouteConfiguration``
+is supplied by management server if the listener refers to an RDS configuration.
+If Envoy is already aware of ``RouteConfiguration`` referenced by ``Listener``,
+it will use it otherwise Envoy will initiate a new RDS request with the new 
+route name referenced by ``Listener``. Management server is expected to provide
+the EDS/RDS updates during warming. If management server does not provide EDS/RDS responses, 
+Envoy will not initialize itself during the initialization phase and the updates sent via CDS/LDS
 will not take effect until EDS/RDS responses are supplied.
 
 .. _xds_protocol_eventual_consistency_considerations:
--- a/docs/root/api-docs/xds_protocol.rst
+++ b/docs/root/api-docs/xds_protocol.rst
@@ -680,14 +680,14 @@ will not take effect until EDS/RDS responses are supplied.
 
 .. note::
 
-   Envoy specific implementation note: Warming of ``Cluster`` is completed only
-   when a new ``ClusterLoadAssignment`` response is supplied by management server
-   even if there is no change in endpoints. 
-   Warming of ``Listener`` is completed even if management server does not send a 
-   response for ``RouteConfiguration`` referenced by ``Listener``. Envoy will use the
-   previously sent ``RouteConfiguration`` to finish ``Listener`` warming. Management Server
-   has to send the ``RouteConfiguration`` response only if it has changed or it was never 
-   sent in the past.
+   Envoy specific implementation notes: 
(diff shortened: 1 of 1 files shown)


markdroth:
I'm not sure how much of the current EDS/RDS warming behavior we really want to document, given that we have agreed in https://github.com/envoyproxy/envoy/issues/13009 that the current behavior needs to change.

ramaraochavali:
> I'm not sure how much of the current EDS/RDS warming behavior we really want to document, given that we have agreed in https://github.com/envoyproxy/envoy/issues/13009 that the current behavior needs to change.

I agree that is is going to change - but having the current state is updated is important IMO. We can add a disclaimer that it is going to change.

alyssawilk:
@envoyproxy/api-shepherds looks like this needs a tie-breaker.  @adisuissa as the other reviewer do you have thoughts?

htuch:
My $0.02 is that this is implementation specific, not generic xDS protocol documentation. There should be a place for it and we can accept it now and update later. It might be fine to just put it in a Sphinx RST informational box as an implementation consideration. I agree with @markdroth not to conflate the Envoy specifics with desired protocol behavior.

ramaraochavali:
>  I agree with @markdroth not to conflate the Envoy specifics with desired protocol behaviour

I am not suggesting to conflate the Envoy specifics with the desired protocol behaviour. However I think it is important to document what Envoy does as it caused confusion in Istio on multiple occasions . I thought of this place because we documented resource warming here. But happy to move it over to another place if that helps

mattklein123:
I like the idea of just doing some type of "client specific note" box for now. We can move this kind of thing out later, and it will be more obvious that it isn't part of the main spec. So for example somewhere near this do something like:

```
.. note::

  Envoy specific implementation note: ...
```

I think that should be OK for now?

ramaraochavali:
@mattklein123 changed as your suggestion. PTAL

docs/root/api-docs/xds_protocol.rst
```diff
@@ -678,6 +678,14 @@ server does not provide EDS/RDS responses, Envoy will not initialize
 itself during the initialization phase and the updates sent via CDS/LDS
 will not take effect until EDS/RDS responses are supplied.
 
+.. note::
+
+   Envoy specific implementation note: Warming of ``Cluster`` is completed only
+   when a new ``ClusterLoadAssignment`` response is supplied by management server
+   even though there is no change in endpoints. However, If Envoy is already aware of
```
mattklein123, line 685:
```suggestion
   even if there is no change in endpoints. However, If Envoy is already aware of
```

?

docs/root/api-docs/xds_protocol.rst
```diff
@@ -678,6 +678,14 @@ server does not provide EDS/RDS responses, Envoy will not initialize
...
+   ``RouteConfiguration`` referenced by ``Listener``, it will use it otherwise Envoy will
+   initiate a new RDS request with the new route name referenced by ``Listener``.
```
mattklein123, line 687: I can't parse this. Is this related to Cluster or unrelated and a different note about warming? Can you clarify and if it's different move it to a separate note?

mattklein123 (changes requested):
Thanks a few comments.

/wait

docs/root/api-docs/xds_protocol.rst
ramaraochavali: It is about listener warming with route. It is under same resource warming category.

mattklein123: OK then I don't understand this and how it relates to the cluster part. Can you clarify? Thanks.

ramaraochavali: It is not related to cluster warming. It is related to resource warming only but talks about how listeners are warmed.

mattklein123: Then please split it out into a separate note and/or make it a clearly separate statement. I'm having a hard time parsing what this means. Thank you!

ramaraochavali: I clarified it. PTAL

docs/root/api-docs/xds_protocol.rst
```diff
@@ -678,6 +678,17 @@ server does not provide EDS/RDS responses, Envoy will not initialize
...
+.. note::
+
+   Envoy specific implementation note: Warming of ``Cluster`` is completed only
+   when a new ``ClusterLoadAssignment`` response is supplied by management server
+   even if there is no change in endpoints. 
+   Warming of ``Listener`` is completed even if management server does not send a 
+   response for ``RouteConfiguration`` referenced by ``Listener``. Envoy will use the
+   previously sent ``RouteConfiguration`` to finish ``Listener`` warming. Management Server
+   has to send the ``RouteConfiguration`` response only if it has changed or it was never 
+   sent in the past.
```
mattklein123, line 690: nit: split this into 2 notes, or a bulleted list of 2 notes. It's not super clear which parts are related.

mattklein123 (changes requested):
LGTM with small nit. Also please check format, thanks.

/wait

mattklein123 (approved):
Thanks!

PR #22387, opened 2022-07-25
filter:manager change metadata and downstream APIs.

alyssawilk:
This PR does two things.  

1) it moves the translation of metadata map types to the HCM, so we consistently use MetadataMapPtr&& everywhere in between the upstream codec and the downstream codec and
2) start moving downstream-only code out of the generic FM base class into a downstream-specific base class.  Specifically I'm claiming (and documenting) that if an upstream filter sends a local reply, as the response is not generated by the cluster so should only traverse the downstream filter chain, so UpstreamFM sendLocal Reply will just forward sendLocalReply to the DownstreamFM and HCM. 

Risk Level: low (minor refactors)
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/envoy/stream_info/stream_info.h
+++ b/envoy/stream_info/stream_info.h
@@ -729,6 +729,14 @@ public:
     upstream_info.setDownstreamBytesMeter(downstream_info.getDownstreamBytesMeter());
   }
 
+  /**
+   * Dump the info to the specified ostream.
+   *
+   * @param os the ostream to dump state to
+   * @param indent_level the depth, for pretty-printing.
+   *
+   * This function is called on Envoy fatal errors so should avoid memory allocation.
+   */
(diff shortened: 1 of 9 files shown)


alyssawilk:
ping?

envoy/stream_info/stream_info.h
```diff
@@ -728,6 +728,8 @@ class StreamInfo {
     downstream_info.setUpstreamBytesMeter(upstream_info.getUpstreamBytesMeter());
     upstream_info.setDownstreamBytesMeter(downstream_info.getDownstreamBytesMeter());
   }
+
+  virtual void dumpState(std::ostream& os, int indent_level = 0) const PURE;
```
snowp, line 732: doc string?

snowp (changes requested):
Makes sense to me, just one comment

snowp (approved):
Thanks!

PR #22393, opened 2022-07-26
Thrift: refactor the response flags into one enum class

JuniorHsu:

Commit Message:
We have too many boolean state for response state which makes the code is hard to read.

Additional Description:
Risk Level: low
Testing: unit
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/network/thrift_proxy/router/upstream_request.h
+++ b/source/extensions/filters/network/thrift_proxy/router/upstream_request.h
@@ -72,13 +72,13 @@ struct UpstreamRequest : public Tcp::ConnectionPool::Callbacks,
   ThriftObjectPtr upgrade_response_;
 
   enum class ResponseState : uint8_t {
-    ResponseNonStarted = 0,
-    ResponseStarted = 1,
-    ResponseCompleted = 2,
+    None = 0,
+    Started = 1,
+    Completed = 2,
     ConnectionReleased = 3
   };
 
-  ResponseState response_state_{ResponseState::ResponseNonStarted};
+  ResponseState response_state_{ResponseState::None};
   bool request_complete_ : 1;
   bool response_underflow_ : 1;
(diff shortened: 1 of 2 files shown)


source/extensions/filters/network/thrift_proxy/router/upstream_request.h
```diff
@@ -71,15 +71,20 @@ struct UpstreamRequest : public Tcp::ConnectionPool::Callbacks,
   ProtocolPtr protocol_;
   ThriftObjectPtr upgrade_response_;
 
+  enum class ResponseState : uint8_t {
+    ResponseNonStarted = 0,
+    ResponseStarted = 1,
+    ResponseCompleted = 2,
+    ConnectionReleased = 3
+  };
```
rgs1, line 79: nit: I don't think we need to repeat Response on every enum member, so maybe something like:

```
  enum class ResponseState : uint8_t {
    NotStarted = 0,
    Started = 1,
    Completed = 2,
    ConnectionReleased = 3
  };
```

I don't like `NotStarted` that much, maybe `None` is better?

rgs1 (changes requested):
Generally LGTM, one nit.

source/extensions/filters/network/thrift_proxy/router/upstream_request.h
JuniorHsu: Fixed. Thanks!

rgs1 (approved):
Thanks!

rgs1 (approved):
(don't know if i need to re-approve after i've been assigned)

zuercher (approved):

zuercher:
Huh. Github is mystifying, but it's merged.

Issue #21941, opened 2022-06-29
local reply modification: support non-string values in Substitution format string in json_format


kfaseela:
*Description*:

We configure “local reply modification” to modify status codes for certain local replies (“local” means that Envoy generates these replies, not an upstream host).
Documentation here: https://www.envoyproxy.io/docs/envoy/v1.22.2/configuration/http/http_conn_man/local_reply.html

We also add a body with content-type application/problem+json according to RFC 7807 (https://www.rfc-editor.org/rfc/rfc7807.html). For this, we use a “SubstitutionFormatString” in “json_format” (documented here: https://www.envoyproxy.io/docs/envoy/v1.22.2/api-v3/config/core/v3/substitution_format_string.proto#envoy-v3-api-field-config-core-v3-substitutionformatstring-json-format).

In this documentation there is an example that shows the JSON object to be created with a “status” field formatted as an integer:

```
                   status_code: 401
                    body_format_override:
                      json_format: 
                        title: "Bad request"
                        status: 400
                        cause: "MANDATORY_IE_INCORRECT"
                       detail: "%RESPONSE_CODE_DETAILS%"
```

However, when we configure a json_format with an integer value, Envoy rejects the configuration because it expects a string. If we put a string there, we get a string value in the JSON output.  The RFC mentioned above specifies an integer value for the “status” field (see chapter 3.1 of the [RFC 7807](https://www.rfc-editor.org/rfc/rfc7807.html), and that is our problem.  The screenshot below shows the error from Envoy (on envoylint.com):

![envoy-json-format-bug](https://user-images.githubusercontent.com/7901446/176433901-ea5e8f51-4328-4cd4-99ac-4e1421a3213c.jpg)

We have checked the Envoy issues on Github to see if somebody else reported this problem already but did not find an existing bug.  We did find however something that is related where somebody fixed a very similar issue in the access log: https://github.com/envoyproxy/envoy/issues/8374

We are using Envoy 1.22.0/1.22.2 (latest released version).

wbpcode:
Yeah, it seems that it's a bug indeed. For now, only string/struct/list are supported for the Json formatter and this is not matched with the docs.

wbpcode:
Looks like the "%RESPONSE_CODE%" can be handled as number, and the "400" can only be treated as string.

kfaseela:
> Looks like the "%RESPONSE_CODE%" can be handled as number, and the "400" can only be treated as string.

[...]

PR #22396, opened 2022-07-26
Support non-string values in Substitution format string in json_format

ZiyangXiao:
Signed-off-by: Xiao, Ziyang <ziyang.xiao@intel.com>

Commit Message: implement support for integer values
Additional Description: Hi, I'm trying to add support of non-string values in substitution format string in json_format mentioned in https://github.com/envoyproxy/envoy/issues/21941. I implement implement support for number values and re-use String formatter instead of add a integer Formatter. Because finally values are changed  into string and plain String formatter can work right in this process if we input an string value number like "400".WDYT?Can you give me some suggestions?
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] https://github.com/envoyproxy/envoy/issues/21941
[Optional Deprecated:]


source/common/formatter/substitution_formatter.cc
```diff
@@ -183,8 +183,13 @@ StructFormatter::FormatBuilder::toFormatMapValue(const ProtobufWkt::Struct& stru
       output->emplace(pair.first, toFormatListValue(pair.second.list_value()));
       break;
 
+    case ProtobufWkt::Value::kNumberValue:
+      output->emplace(pair.first,
+                      toFormatStringValue(std::to_string(uint32_t(pair.second.number_value()))));
+      break;
```
zuercher, line 189: I don't think it's correct to force the value to be a uint32_t. Number Values are [stored as signed doubles](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L68).

I think it would be better to provide a `class PlainNumberFormatter : FormatterProvider` (similar to the existing `PlainStringFormater`). That requires implement `format` (which should emit a string using with the same semantics as the `%g` printf specifier) and `formatValue` (which can just return the number Value).

You can then provide a `toFormatNumberValue` method in this class and use it to create a PlainNumberFormatter which the concrete structured output class can emit in the appropriate format. I think that's limited to JSON right now, which uses the same type for numeric output.

kfaseela:
cc @mattklein123

source/common/formatter/substitution_formatter.cc
ZiyangXiao: Done.

source/common/formatter/substitution_formatter.cc
```diff
@@ -1541,6 +1551,36 @@ ProtobufWkt::Value PlainStringFormatter::formatValue(const Http::RequestHeaderMa
...
+
+absl::optional<std::string> PlainNumberFormatter::format(const Http::RequestHeaderMap&,
+                                                         const Http::ResponseHeaderMap&,
+                                                         const Http::ResponseTrailerMap&,
+                                                         const StreamInfo::StreamInfo&,
+                                                         absl::string_view) const {
+  std::string str = std::to_string(num_.number_value());
+  str = str.substr(0, str.find_last_not_of('0') + 1);
+  str = str.substr(0, str.find_last_not_of('.') + 1);
+  return str;
```
zuercher, line 1564: I think we can do better than this. Please have a look at `absl::StrFormat`.

source/common/formatter/substitution_formatter.cc
```diff
@@ -1541,6 +1551,36 @@ ProtobufWkt::Value PlainStringFormatter::formatValue(const Http::RequestHeaderMa
   return str_;
 }
 
+PlainNumberFormatter::PlainNumberFormatter(const double& num) { num_.set_number_value(num); }
```
zuercher, line 1554: I think just `double num` is fine here. I don't see any benefit to using a reference type.

source/common/formatter/substitution_formatter.cc
```diff
@@ -1541,6 +1551,36 @@ ProtobufWkt::Value PlainStringFormatter::formatValue(const Http::RequestHeaderMa
...
+ProtobufWkt::Value PlainNumberFormatter::formatValue(const Http::RequestHeaderMap&,
+                                                     const Http::ResponseHeaderMap&,
+                                                     const Http::ResponseTrailerMap&,
+                                                     const StreamInfo::StreamInfo&,
+                                                     absl::string_view) const {
+  return num_;
+}
+
+std::vector<FormatterProviderPtr>
+PlainNumberFormatter::toFormatNumberValue(const double& number_format) {
```
zuercher, line 1576: Let's make this a method on StructFormatter like toFormatStringValue. It also doesn't need to take a const reference. Just a `double` is fine.

source/common/formatter/substitution_formatter.cc
ZiyangXiao: Done.

source/common/formatter/substitution_formatter.cc
ZiyangXiao: Done.

source/common/formatter/substitution_formatter.cc
ZiyangXiao: Done.

zuercher (commented):
There are a couple of small cleanups I'd like to see, but otherwise looks good.

source/common/formatter/substitution_formatter.cc
```diff
@@ -253,6 +263,15 @@ ProtobufWkt::Value StructFormatter::providersCallback(
   return ValueUtil::stringValue(str);
 }
 
+std::vector<FormatterProviderPtr>
+StructFormatter::FormatBuilder::toFormatNumberValue(double number_format) const {
+  double current_number;
+  current_number = number_format;
```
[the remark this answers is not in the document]
ZiyangXiao: Done.

source/common/formatter/substitution_formatter.cc
```diff
@@ -253,6 +263,15 @@ ProtobufWkt::Value StructFormatter::providersCallback(
   return ValueUtil::stringValue(str);
 }
 
+std::vector<FormatterProviderPtr>
+StructFormatter::FormatBuilder::toFormatNumberValue(double number_format) const {
```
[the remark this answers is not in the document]
ZiyangXiao: Done.

zuercher (dismissed):
Thanks!

kfaseela:
Thanks @zuercher @ZiyangXiao !

ZiyangXiao:
friendly ping @mattklein123

mattklein123 (approved):

mattklein123 (changes requested):
Actually sorry this change needs a release note, thank you.

/wait

mattklein123 (approved):
Thanks!

PR #22408, opened 2022-07-26
Rename H/1 codec method to reduce confusion

yanavlasov:
Risk Level: Low
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A


alyssawilk (commented):
Now that I think about it, do you think completeCurrentHeader would be less confusing?
completeLast also reads as the last header, not the last one to be parsed.
/wait

yanavlasov:
done

alyssawilk (approved):
thanks for the clean up

PR #22415, opened 2022-07-26
wasm: fix crash in gRPC/HTTP handlers calling proxy_done

kyessenov:

Commit Message: A fix to address an issue in istio/proxy. There's a major problem in ABI implementation dealing with `proxy_done`. The code is written so that `ContextBase` may delete `Context` while executing methods on `Context`. This PR addresses known use of `proxy_done` when called from gRPC handler by wrapping continuation into deferred actions before calling `ContextBase`.
Risk Level: low
Testing: manual verification, regression test
Docs Changes:
Release Notes:

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1844,22 +1844,23 @@ void Context::onHttpCallFailure(uint32_t token, Http::AsyncClient::FailureReason
 
 void Context::onGrpcReceiveWrapper(uint32_t token, ::Envoy::Buffer::InstancePtr response) {
   ASSERT(proxy_wasm::current_context_ == nullptr); // Non-reentrant.
+  auto cleanup = [this, token] {
+    if (wasm()->isGrpcCallId(token)) {
+      grpc_call_request_.erase(token);
+    }
+  };
   if (wasm()->on_grpc_receive_) {
     grpc_receive_buffer_ = std::move(response);
     uint32_t response_size = grpc_receive_buffer_->length();
     // Deferred "after VM call" actions are going to be executed upon returning from
     // ContextBase::*, which might include deleting Context object via proxy_done().
-    wasm()->addAfterVmCallAction([this, token] {
+    wasm()->addAfterVmCallAction([this, cleanup] {
       grpc_receive_buffer_.reset();
-      if (wasm()->isGrpcCallId(token)) {
-        grpc_call_request_.erase(token);
-      }
+      cleanup();
     });
     ContextBase::onGrpcReceive(token, response_size);
   } else {
-    if (wasm()->isGrpcCallId(token)) {
-      grpc_call_request_.erase(token);
-    }
+    cleanup();
   }
 }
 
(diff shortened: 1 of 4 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1836,14 +1836,17 @@ void Context::onHttpCallFailure(uint32_t token, Http::AsyncClient::FailureReason
 
 void Context::onGrpcReceiveWrapper(uint32_t token, ::Envoy::Buffer::InstancePtr response) {
   ASSERT(proxy_wasm::current_context_ == nullptr); // Non-reentrant.
+  if (wasm()->isGrpcCallId(token)) {
+    grpc_call_request_.erase(token);
```
ingwonsong, line 1840: @PiotrSikora If we remove grpc_call_request object here, I think the wasm callback corresponding to `on_grpc_receive_` may crash if it want to access the request object. Is it OK?

PiotrSikora: `grpc_call_request_` is only tracked on Envoy side, and it's not used in Proxy-Wasm C++ Host and Wasm plugin.

However, I think we're removing it prematurely (triggering destructor) by calling `erase` here. Should we instead `extract` it, so that it's freed after the function is done?

source/extensions/common/wasm/context.cc
```diff
@@ -1836,14 +1836,17 @@ void Context::onHttpCallFailure(uint32_t token, Http::AsyncClient::FailureReason
...
+  }
   if (wasm()->on_grpc_receive_) {
     grpc_receive_buffer_ = std::move(response);
     uint32_t response_size = grpc_receive_buffer_->length();
+    wasm()->addAfterVmCallAction([this] {
+      grpc_receive_buffer_.reset();
+    });
+    // on_grpc_receive_ may indirectly delete Context via proxy_done().
     ContextBase::onGrpcReceive(token, response_size);
```
PiotrSikora, line 1849: Nit: consider changing the comment to:
```
// Deferred "after VM call" actions are going to be executed upon returning from
// ContextBase::*, which might include deleting Context object via proxy_done().
```

source/extensions/common/wasm/context.cc
kyessenov: Updated. We can defer that, too.

source/extensions/common/wasm/context.cc
kyessenov: Done.

PiotrSikora (commented):
Thanks, this looks good.

Could you apply the same fixes to other HTTP and gRPC callout callbacks?

kyessenov:
Update response handlers.

PiotrSikora:
cc @mpwarres @chaoqin-li1123

PiotrSikora (dismissed):
LGTM, thanks for investigating and the fix!

kyessenov:
Fixes: https://github.com/istio/istio/issues/40101

PiotrSikora (approved):

ggreenway (approved):
Approved based on @PiotrSikora 's review as senior extension maintainer.

Issue #13951, opened 2020-11-10
deprecate config using the One True ADS Server(tm)


markdroth:
Today, the way ADS is configured is as follows:
- The bootstrap file has an [`ads_config`](https://github.com/envoyproxy/envoy/blob/d9af0e985a2fef2c921641b12f4786a658ad0923/api/envoy/config/bootstrap/v3/bootstrap.proto#L100) field indicating the One True ADS Server(tm).
- The bootstrap file has its [`lds_resource_locator`](https://github.com/envoyproxy/envoy/blob/d9af0e985a2fef2c921641b12f4786a658ad0923/api/envoy/config/bootstrap/v3/bootstrap.proto#L82) and [`cds_resources_locator`](https://github.com/envoyproxy/envoy/blob/d9af0e985a2fef2c921641b12f4786a658ad0923/api/envoy/config/bootstrap/v3/bootstrap.proto#L91) fields set to a `ConfigSource` with the [`ads`](https://github.com/envoyproxy/envoy/blob/e1c138b427ff9bcdf5ebcb2639cc2d22f98ac1a8/api/envoy/config/core/v3/config_source.proto#L185) field set.
- In resources sent from the management server, `ConfigSource` protos use the same [`ads`](https://github.com/envoyproxy/envoy/blob/e1c138b427ff9bcdf5ebcb2639cc2d22f98ac1a8/api/envoy/config/core/v3/config_source.proto#L185) field.

This approach seems incompatible with the new xDS naming scheme, which was designed to support federation.  As part of the new naming scheme, the `ApiConfigSource.ApiType` enum now has [`AGGREGATED_GRPC`](https://github.com/envoyproxy/envoy/blob/e1c138b427ff9bcdf5ebcb2639cc2d22f98ac1a8/api/envoy/config/core/v3/config_source.proto#L68) and [`AGGREGATED_DELTA_GRPC`](https://github.com/envoyproxy/envoy/blob/e1c138b427ff9bcdf5ebcb2639cc2d22f98ac1a8/api/envoy/config/core/v3/config_source.proto#L73) values, which allows configuring any `ConfigSource` to use a single ADS stream instead of using a separate stream for each resource type.  This way of configuring ADS seems much more in-line with the direction we're heading in and will make it easier to transition to the new federated world.

Therefore, I propose replacing this notion of the One True ADS Server(tm) with use of this new `AGGREGATED_GRPC` enum value.  The typical ADS configuration described above would instead be represented as follows:
[...]

Issue #15072, opened 2021-02-17
allow google grpc local shutdown


lambdai:
Destroying GoogleAsyncClientThreadLocal put the reset streams in dispatcher deferred deleted list and expect the stream object not destroyed. If these streams are destroyed, the GrpcStream, which hides google grpc and envoy grpc, holds a dangling pointer to the destroyed stream. The xds client maybe call sendMessage and will crash Envoy. 

Essentially dispatcher should be able to clean up the deferred deleted list because the deferred deleted object may hold references to ssl ctx object. We want to clean it up in shutdown. Also this allows dispatcher to easier to detect unexpected behavior after shutdown.

lambdai:
Stack frames: please ignore the dispatcher behavior. 
```
[20][debug][grpc] [source/common/grpc/google_async_client_impl.cc:233] resetStream
  [20][debug][grpc] [source/common/grpc/google_async_client_impl.cc:401] Stream cleanup with 1 in-flight tags
  [24][debug][grpc] [source/common/grpc/google_async_client_impl.cc:73] completionThread exiting
  [20][debug][grpc] [source/common/grpc/google_async_client_impl.cc:40] Joining completionThread
  [20][debug][grpc] [source/common/grpc/google_async_client_impl.cc:42] Joined completionThread
  [20][debug][grpc] [source/common/grpc/google_async_client_impl.cc:390] Deferred delete
  [20][debug][upstream] [source/common/upstream/cluster_manager_impl.cc:1065] shutting down thread local cluster manager
  [20][info][main] [source/server/server.cc:801] exiting
  [20][debug][init] [source/common/init/watcher_impl.cc:31] RunHelper destroyed
  [20][debug][main] [source/server/server.cc:139] destroying listener manager
  [20][debug][init] [source/common/init/target_impl.cc:34] target LDS destroyed
  [20][debug][main] [source/common/event/dispatcher_impl.cc:81] destroying dispatcher worker_0
  [20][debug][upstream] [source/common/upstream/upstream_impl.cc:928] Schedule destroy cluster info ads_cluster
  [20][debug][upstream] [source/common/upstream/upstream_impl.cc:928] Schedule destroy cluster info dummy_cluster
  [20][debug][upstream] [source/common/upstream/upstream_impl.cc:928] Schedule destroy cluster info eds_cluster
  [20][debug][main] [source/server/server.cc:141] destroyed listener manager
[...]

PR #5026, opened 2018-11-13
Fixes #2943 - implements shared grpc connection per multiple EDS subscriptions. 

dmitri-d:
Was originally started in https://github.com/envoyproxy/envoy/pull/4823. 

*Description*: Fixes #2943 by implementing shared grpc connection per multiple EDS subscriptions. 
*Risk Level*:
*Testing*:
*Docs Changes*:
*Release Notes*:
[Optional Fixes #Issue]
[Optional *Deprecated*:]

@htuch: I removed ClusterManager dependency on GrpcMuxFactory, as there was a circular dependency after some recent changed. The extacted code now lives in EdsSubscriptionFactory, and all changes I made are constrained within EDS.

PR #18195, opened 2021-09-21
LEDS: integrating with EDS

adisuissa:
Commit Message: LEDS: integrating with EDS
Additional Description:
Allows the usage of the LEDS protocol with EDS messages.
Most of the changes are for the integration tests.
One key question is whether we can minimize the amount of memory used when LEDS isn't used.

Risk Level: Medium - Impact EDS behavior.
Testing: Many integration tests.
Docs Changes: TBD.
Release Notes: TBD.
Platform Specific Features: N/A.

Signed-off-by: Adi Suissa-Peleg <adip@google.com>

PR #22419, opened 2022-07-26
Optimize EDS stream allocation

### LINKED ISSUE #2943: Optimize EDS stream allocation
Today, EDS attempts to open one stream per cluster to the management server when not being multiplexed over ADS. EDS streams support subscribing to more than one cluster, so this should not be necessary if the `ApiConfigSource` points to the same management server.

This is actually very noticeable when we have a large number of clusters. It shows up as stream resource exhaustion in H2 (https://github.com/envoyproxy/envoy/issues/2941) as well as in internal performance benchmarks of configuration scalability.

The fix seems pretty easy, just need to do some level of aggregation similar to what ADS does.

nezdolik:
Signed-off-by: Kateryna Nezdolii <nezdolik@spotify.com>

Commit Message: Optimize streams allocation for Eds Grpc
Additional Description: With this change Eds Grpc subscriptions towards the same management server reuse a shared grpc connection, which greatly reduces the number of concurrent H2 streams towards a single management server. The work is built on top of [5026](https://github.com/envoyproxy/envoy/pull/5026), with bugs fixed and an addition of integration test for Eds over Grpc. I could not port 3 existing Eds integration tests for warming clusters, as they were blocked by [non graceful shutdown of GoogleAsyncClient](https://github.com/envoyproxy/envoy/issues/15072). 

The change has also been tested in production Spotify environment, we use Eds Grpc. Envoy successfully resolved cluster load assignments via optimized Eds.

Risk Level: High
Testing: Unit tests + integration tests
Docs Changes: TBD
Release Notes: TBD
Platform Specific Features:
Fixes #2943


nezdolik:
cc @htuch we have chatted about this change long time ago prior to my parental leave.

nezdolik:
cc @htuch we have chatted about this change long time ago prior to my parental leave.

htuch:
Thanks so much @nezdolik, this is a definitely long asked for feature from the community and will be a welcome contribution. Assigning @adisuissa for review, since Adi is now leading work around xDS architecture in Envoy.

adisuissa:
Apologies for the late reply.
Thanks for working on this, and thanks for trying this in prod and giving us the feedback!
Overall I think we would like to have this support for any xDS service, not only for EDS.

I'll start by giving some background on where I think we should be heading, and then answer your questions.

In the future, as we move towards the [new xDS-TP](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md#transport-network-addressing), the subscriptions to the same authority will essentially be multiplexed using this feature. Your optimized EDS stream allocation will work for each ADS subscription to a specific authority.
For now, we have a single gRPC stream when ADS is used (which doesn't need this multiplexing), and for non-ADS we have per-subscription stream.

IIUC the current design is: The cluster manager has a single `EdsSubscriptionFactory` which maps between config and a gRPC stream.
In the future, we will probably have a ConfigSubscriptionsManager in the server, that will own all the subscriptions, and a map between a config-source (and authority) to its actual GrpcMux. The GrpcMux can be configured to either SotW or Delta, and will abstract over it (just as is being done today).

For this PR I've got a few high-level comments to start with:
1. Should the EdsSubscriptionFactory extend SubscriptionFactoryImpl, and just have a method to return the correct GrpcMux (overriding the current create methods, and calling them if the subscription isn't there already)?
2. Can we use a different name for this factory, maybe `MultiplexedSubscriptionFactory`? (other ideas to make it more generic are welcome :) )
3. Can we add some integration tests?
4. For now I do agree that this factory should co-exist with the current (non-multiplexed) subscription factory, but we should gradually move away from using the original one. I think that introducing (and guarding) it for each xDS-type is the right way forward.

Addressing your questions:

> Do we need support for Delta Grpc APi type?
Yes, we should support both SotW and Delta, but I don't think it will add too much to this implementation.

> Do we need support for unified mux?
The unified-mux code essentially uses the same code-base for both SotW and Delta implementations, so yes we would like to support it. In the near future we will move to that code-base and abandon the old one. The unified mux code should implement the GrpcMux interface, so I don't think it will require much work.

> Implement collectionSubscriptionFromUrl method? (part of udpa/resource federation)
Yes, it should be implemented, but I think it is very straight-forward.

> Hide change behind feature flag in api? (as it is high risk)
I think this is highly desired, at least temporarily, as we want to allow the use of the current subscription code if we break anyone.

LMK if there's something unclear, and if I can help with anything.

nezdolik:
@adisuissa thanks for a very detailed feedback and apologies for late reply, I have been stuck in figuring out mocks fix for the remaining Eds tests.
I agree that MultiplexedSubscriptionFactory should be a general feature that could be enabled for any XDS type (since RDS and SDS currently expose the same perf problem for a large number of resources).
We could start by implementing a generic MultiplexedSubscriptionFactory and exposing it as a config option in EDS api as a start, does that sound good?
I will start addressing your feedback in a series of commits and will ping you when all parts are done.
May need some help later with the integration test, I was not completely sure on which signal/stats to rely to verify that grpc channel is being multiplexed in the test(may figure it out later in the process).
The original idea was to look at Envoy stats for successfully completed requests for an H2 Eds cluster, since in H2 each stream maps to a single logical request/response.

adisuissa:
> I agree that MultiplexedSubscriptionFactory should be a general feature that could be enabled for any XDS type (since RDS and SDS currently expose the same perf problem for a large number of resources).We could start by implementing a generic MultiplexedSubscriptionFactory and exposing it as a config option in EDS api as a start, does that sound good?
> 

Sounds good to me, thanks for all the efforts!
 
> May need some help later with the integration test, I was not completely sure on which signal/stats to rely to verify that grpc channel is being multiplexed in the test(may figure it out later in the process). The original idea was to look at Envoy stats for successfully completed requests for an H2 Eds cluster, since in H2 each stream maps to a single logical request/response.

I wonder if the [upstream statistics](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#general) can be used to count the number of connections.
If you point me to the specific test, I can try to take a look.

nezdolik:
Applied portion of review comments:
* EdsSubscriptionFactory renamed to MultiplexedSubscriptionFactory
* MultiplexedSubscriptionFactory extends SubscriptionFactoryImpl and overrides method for creating mux
* MultiplexedSubscriptionFactory shares mux for subscriptions per management server per xds type url.
* Updated MultiplexedSubscriptionFactory test
I will now work on integration test, add tests for new feature and make it parametrised for xds type (sotw, delta) and grpc flavour (envoy, google).
@adisuissa I am not sure if upstream stats for number of connections is good enough signal to rely on. Since mux manages a single grpc stream (according to [docs](https://github.com/envoyproxy/envoy/blob/main/source/docs/xds.md)) and grpc channel may have many streams.

test/integration/eds_integration_test.cc
```diff
@@ -437,5 +451,471 @@ TEST_P(EdsIntegrationTest, StatsReadyFilter) {
...
+  cleanupUpstreamAndDownstream();
+}
+
+TEST_P(EdsOverGrpcIntegrationTest, ReuseMuxAndStreamForMultipleClusters) {
+  initializeTest(false, 3, false);
+  EXPECT_EQ(0, test_server_->gauge("cluster.cluster_0.membership_total")->value());
+  EXPECT_EQ(0, test_server_->gauge("cluster.cluster_1.membership_total")->value());
+  EXPECT_EQ(0, test_server_->gauge("cluster.cluster_2.membership_total")->value());
+  switch (clientType()) {
+  case Grpc::ClientType::EnvoyGrpc:
```
nezdolik, line 897: @adisuissa figured out the approach to verify the number of streams for both Grpc client types. For Envoy Grpc number of total streams towards specific cluster equals to total number of H2 requests. Google Grpc actually exposes stream metrics per client instance, which is nice. Please let me know wdyt.

test/integration/eds_integration_test.cc
```diff
@@ -1,3 +1,5 @@
+#include <iostream>
```
nezdolik, line 1: to be cleaned up

nezdolik:
* Added integration test to verify that discovery requests for multiple EDS clusters reuse one stream/mux towards same management server.

Looks like there is merge conflict that needs to be fixed. Will work now on adding api feature switch in EDS config and extending tests for the config property.

nezdolik:
@adisuissa this is ready for review

yanavlasov:
@adisuissa and @lizan this is ready for review

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -205,6 +205,11 @@ message Cluster {
     // have the same restrictions as cluster name, i.e. it may be arbitrary
     // length. This may be a xdstp:// URL.
     string service_name = 2;
+
+    // If set, all discovery requests for EDS will utilize a single GRPC stream towards the same management server.
+    // Defaults to false. When unset, Envoy will allocate a GRPC stream per EDS cluster which can lead to the exhaustion of
+    // the limit for max concurrent streams in HTTP2.
+    bool multiplex_eds = 3;
```
markdroth, line 212: I don't think this should be done by adding a new field to the API like this, especially not an EDS-specific field.  The fact that multiple EDS subscriptions are not currently multiplexed onto a single stream is basically a bug in Envoy, and I think we should just fix the bug.

If you explicitly want to find a way to make this configurable via the API, then a better approach would be to fix #13951.  That would allow a `ConfigSource` to specify whether it's an ADS stream, in which case Envoy will multiplex the EDS subscriptions onto a single ADS stream.  And that approach would work for all resource types, not just EDS.

nezdolik: @markdroth the config option was introduced to reduce risk of rolling out the change, as previously merged [fix](https://github.com/envoyproxy/envoy/pull/5272) has broken EDS for bunch of users. I agree is not the best usage of API, as eventually everyone should use the multiplexed version of subscription factory for EDS. Perhaps we could instead guard new feature with runtime flag? We have increased confidence for current bug fix, as it is now covered by integration tests and has been tested in prod environment, but there is still risk involved.

I would like to focus first on fixing EDS bug, as it has significant impact on us. I can further help out with https://github.com/envoyproxy/envoy/issues/13951 but as a separate effort.

markdroth: A runtime flag sounds fine.

I would really prefer not to add an EDS-specific option to the API for this, since it's really not an API issue, it's just an Envoy implementation bug.

mattklein123:
Sorry just catching up on this issue. I would like to get this fixed for the folks at Spotify. Are we good with this overall change if we call it a bug and fix it under runtime flag? That seems fine to me. cc @markdroth @nezdolik

markdroth:
Triggering this via a runtime flag instead of a new API field SGTM.

test/common/grpc/grpc_client_integration.h
```diff
@@ -150,6 +173,11 @@ class DeltaSotwIntegrationParamTest
   testing::Combine(testing::ValuesIn(TestEnvironment::getIpVersionsForTest()),                     \
                    testing::ValuesIn(TestEnvironment::getsGrpcVersionsForTest()),                  \
                    testing::Values(Grpc::LegacyOrUnified::Legacy, Grpc::LegacyOrUnified::Unified))
-
+#define EDS_MODE_DELTA_SOTW_GRPC_CLIENT_INTEGRATION_PARAMS                                         \
```
nezdolik, line 176: integration test for eds over grpc has cardinality of 16, it takes some time to run it.

nezdolik:
@mattklein123 @markdroth switched to using runtime feature flag for eds multiplexing.

source/common/runtime/runtime_features.cc
```diff
@@ -93,6 +93,8 @@ FALSE_RUNTIME_GUARD(envoy_reloadable_features_runtime_initialized);
 // TODO(mattklein123): Also unit test this if this sticks and this becomes the default for Apple &
 // Android.
 FALSE_RUNTIME_GUARD(envoy_reloadable_features_always_use_v6);
+// TODO(nezdolik): Flip this to true to multiplex eds by default.
+FALSE_RUNTIME_GUARD(envoy_reloadable_features_multiplex_eds);
```
mattklein123, line 97: Is there any reason not to have this be true by default?

mattklein123 (commented):
At a high level this makes sense to me but I will defer to @adisuissa for the full review. Thank you!

source/common/runtime/runtime_features.cc
nezdolik: it makes sense, am just being overcautious

phlax:
@adisuissa could you give this another pass please

source/common/runtime/runtime_features.cc
mattklein123: If you can test it in prod ahead of time, I would flip to true by default. If not, I think it's fine to leave as false and come back and switch to true after you test it.

test/common/grpc/grpc_client_integration.h
adisuissa: Can these tests be sharded?

test/common/grpc/grpc_client_integration.h
```diff
@@ -125,6 +127,27 @@ class DeltaSotwIntegrationParamTest
   SotwOrDelta sotwOrDelta() const { return std::get<2>(GetParam()); }
 };
 
+class EdsModeDeltaSotwIntegrationParamTest
```
adisuissa, line 130: Consider renaming to `MultiplexedDeltaSotwIntegrationParamTest` as this is not only for EDS.

test/integration/eds_integration_test.cc
```diff
@@ -437,5 +438,486 @@ TEST_P(EdsIntegrationTest, StatsReadyFilter) {
...
+    }
+    cluster.set_name(name);
+    cluster.set_type(envoy::config::cluster::v3::Cluster::EDS);
+    cluster.mutable_connect_timeout()->CopyFrom(Protobuf::util::TimeUtil::SecondsToDuration(5));
+    auto* eds_cluster_config = cluster.mutable_eds_cluster_config();
+    eds_cluster_config->mutable_eds_config()->set_resource_api_version(
+        envoy::config::core::v3::ApiVersion::V3);
+    if (edsUpdateMode() == Grpc::EdsUpdateMode::Multiplexed) {
+      config_helper_.addRuntimeOverride("envoy.reloadable_features.multiplex_eds", "true");
+    }
```
adisuissa, line 616: Should be moved to `initializeTest()` as it not per-cluster behavior.

test/integration/eds_integration_test.cc
```diff
@@ -437,5 +438,486 @@ TEST_P(EdsIntegrationTest, StatsReadyFilter) {
...
+                nullptr);
+      break;
+    default:
+      PANIC("reached unexpected code");
+    }
+  } else {
+    test_skipped = true;
+  }
+}
+
```
adisuissa, line 921: I'm wondering how difficult will it be to add a test that changes modes between "envoy.reloadable_features.multiplex_eds" enabled to disabled and back?

test/integration/eds_integration_test.cc
```diff
@@ -437,5 +438,486 @@ TEST_P(EdsIntegrationTest, StatsReadyFilter) {
...
+                nullptr);
+      EXPECT_EQ(TestUtility::findCounter(test_server_->statStore(),
+                                         "cluster.cluster_2.grpc.eds_cluster.streams_total"),
+                nullptr);
+      break;
+    default:
+      PANIC("reached unexpected code");
+    }
+  } else {
+    test_skipped = true;
```
adisuissa, line 918: nit: just return at the beginning of the test if the edsUpdateMode isn't multiplexed.

test/integration/eds_integration_test.cc
```diff
@@ -437,5 +438,486 @@ TEST_P(EdsIntegrationTest, StatsReadyFilter) {
   cleanupUpstreamAndDownstream();
 }
 
+class EdsOverGrpcIntegrationTest : public Grpc::EdsModeDeltaSotwIntegrationParamTest,
```
adisuissa, line 441: If this class and tests do not share code with the above class/tests, I suggest refactoring it into a different file.

test/integration/eds_integration_test.cc
```diff
@@ -437,5 +438,486 @@ TEST_P(EdsIntegrationTest, StatsReadyFilter) {
...
+    for (uint32_t i = 0; i < num_clusters_to_add; ++i) {
+      EXPECT_TRUE(compareDiscoveryRequest(
+          Config::TypeUrl::get().ClusterLoadAssignment, "",
+          {resource_names_.begin(), resource_names_.begin() + i + 1}, {}, {}, true));
+    }
+    sendDiscoveryResponse<envoy::config::endpoint::v3::ClusterLoadAssignment>(
+        Config::TypeUrl::get().ClusterLoadAssignment, clas_to_add_, clas_to_add_, {},
+        std::to_string(eds_version_));
+    // Receive EDS ack.
+    EXPECT_TRUE(compareDiscoveryRequest(Config::TypeUrl::get().ClusterLoadAssignment,
```
adisuissa, line 592: I may be missing something here, but shouldn't this test file verify that both the multiplexed and non-multiplexed are working correctly?
If so, shouldn't there be multiple xds streams, and sending updates on multiple streams (in the case of non-multiplexed), and use of a single stream in the multiplexed case?
Please let me know if I'm missing something.

adisuissa (commented):
Apologies for the late reply, and thanks for working on this.
Overall the code LGTM, and added some comments on the tests.

I was considering the tradeoffs between your design, and having the multiplexed implementation as part of the current subscription factory implementation.
I think that it makes sense to use your current design, as it allows to gradually introduce the use for different xDS types. Eventually, I believe, we'll just use the multiplexed subscription for all gRPC-based xDS types.

test/integration/eds_integration_test.cc
nezdolik: this variable is needed for clean shutdown of the test, eg if test and its initialisation logic were skipped then cleanup logic (tear down xds connection, tear down conns to upstreams) is also skipped.

test/common/grpc/grpc_client_integration.h
nezdolik: yes

test/integration/eds_integration_test.cc
nezdolik: That is correct. I intentionally omitted tests for multiple streams/clusters with no multiplexing, because the code is async and is not possible to predict the order of discovery requests. Is there is an established way to test out of order expected set of discovery requests/responses?

adisuissa: That's a great question.
I don't think there is something ready. I've implemented this once for delta-xDS in a previous PR (https://github.com/envoyproxy/envoy/pull/18195/commits/20b1674193e62b6087dd13340b7bc91c03de6301#diff-5227f6492ddd6badd76af87ce9f5f1c8acc349eed191e65a2bc05ab5d43b65f6R643) but eventually it wasn't needed for the tests.

test/common/grpc/grpc_client_integration.h
nezdolik: after extracting eds grpc test to separate file, test running time have decreased, don't think we need to shard

test/integration/eds_integration_test.cc
nezdolik: @adisuissa it will require writing lots of code. Maybe not tricky, but large code wise

nezdolik:
@adisuissa this is ready for review (applied review comments, added test for stream per cluster, enabled multiplexing by default). Only piece not done is integration test for turning on/off the flag via runtime.

test/extensions/clusters/eds/eds_test.cc
```diff
@@ -2520,6 +2536,99 @@ TEST_F(EdsTest, OnConfigUpdateLedsAndEndpoints) {
...
+          api_config_source:
+            api_type: REST
+            cluster_names:
+            - eds
+            refresh_delay: 1s
+    )EOF",
+               Cluster::InitializePhase::Secondary);
+}
+
+TEST_F(EdsTest, MultiplexEdsDisabledByDefault) {
```
adisuissa, line 2609:
```suggestion
TEST_F(EdsTest, MultiplexEdsEnabledByDefault) {
```

adisuissa (commented):
Thanks for working on this!
Code LGTM, and I still need to go through the last integration test (sorry, it is quite large).
RE: "Only piece not done is integration test for turning on/off the flag via runtime", I thought that `MultiplexedDeltaSotwIntegrationParamTest` (which `EdsOverGrpcIntegrationTest` extends) takes care of that, no?

nezdolik:
> Thanks for working on this! Code LGTM, and I still need to go through the last integration test (sorry, it is quite large). RE: "Only piece not done is integration test for turning on/off the flag via runtime", I thought that `MultiplexedDeltaSotwIntegrationParamTest` (which `EdsOverGrpcIntegrationTest` extends) takes care of that, no?

It does yes, what i meant is there is no test that turns flag on and then off via runtime in a single test.

nezdolik:
This bug fix is important for our company as it unblocks significant cost cut down for cross-zone traffic for all services. 
We intended to use the fix early Q1, hopefully we can get it merged by then and not use internal fork with patch included.

adisuissa (approved):
LGTM, thanks!

nezdolik:
The patch has been successfully tested in Spotify prod environment, @lizan

KBaichoo:
friendly ping @lizan 

I think this is waiting for your input.

@nezdolik this needs a main merge

/wait

adisuissa (approved):
LGTM, thanks!

changelogs/current.yaml
```diff
@@ -122,6 +122,9 @@ bug_fixes:
 - area: jwt_authn
   change: |
     fix a bug that jwt_cache breaks the :ref:`provider_and_audiences <envoy_v3_api_field_extensions.filters.http.jwt_authn.v3.JwtRequirement.provider_and_audiences>` JWT requirement.
+- area: eds
+  change: |
+    added ``envoy.reloadable_features.multiplex_eds`` to enable eds multiplexing by default.
```
mattklein123, line 127: Can you say more about what this does? Also, can you flip the text about turning it off via runtime flag because I think this is on by default now?

This will need another rebase as we will land this after the release cut today due to risk. Thank you!

/wait

mattklein123 (approved):
Thanks!

Issue #22417, opened 2022-07-26
EDS deduping endpoints with same address but different metadata


stevenctl:
*Title*: *EDS deduping endpoints with same address but different metadata*

*Description*:

When testing LB behavior and config_dump?include_eds for an EDS cluster with endpoints that have the same address, but different metadata I see that only one endpoint is actually preserved. This behavior makes sense to me, but I can't find documentation that says it should be happening. Assuming it's not correct, the goal is to send traffic to the same internal listener and then using a custom filter on that internal listener extract destination info from passthrough metadata. 

Notes:
*  I have not tested this with socketaddress 
* STATIC clusters with similar sets of endpoints seem to be ok
cc @kyessenov 

*Repro steps*:

Using the static config below, and a modified [example xds server](https://github.com/envoyproxy/go-control-plane/tree/main/internal/example) that just returns endpoints for "outbound_service"

<details>
<summary>
static config (admin on 15000, points to xds on 15012)
</summary>

```yaml
node:
  id: test-id
  cluster: outbound_service
admin:
  address:
    socket_address: { address: 127.0.0.1, port_value: 15000 }
bootstrap_extensions:
- name: envoy.bootstrap.internal_listener
  typed_config:
    "@type": type.googleapis.com/envoy.extensions.bootstrap.internal_listener.v3.InternalListener
layered_runtime:
  layers:
  - name: global config
    static_layer:
      envoy.reloadable_features.internal_address: 'true'
static_resources:
  listeners:
  - name: outbound_capture
    use_original_dst: true
    address:
      socket_address: { address: 127.0.0.1, port_value: 15001 }
    filter_chains:
    - filters:
      - name: envoy.filters.network.tcp_proxy
        typed_config:
          "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
          stat_prefix: ingress_tcp
          cluster: outbound_service
  - name: tunnel_lis
    internal_listener: {}
    address:
      envoy_internal_address:
        server_listener_name: tunnel_lis
    filter_chains:
    - filters:
      - name: envoy.filters.network.tcp_proxy
        typed_config:
          "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
          stat_prefix: ingress_tcp
          cluster: outbound_tunnel
  clusters:
  - name: outbound_service
    type: EDS
    lb_policy: ROUND_ROBIN
    eds_cluster_config:
[...]

PR #22420, opened 2022-07-26
eds: allow duplicate internal addresses

stevenctl:
Commit Message: allow duplicate internal endpoints
Additional Description: endpoints that use internal address should not be subject to deduplication as passthrough metadata is usually what distinguishes them
Risk Level: Low
Testing: TODO unit test
Docs Changes: N/A
Release Notes: should I add one? are there releases including internal address at this point? 

Fixes https://github.com/envoyproxy/envoy/issues/22417

### DIFF
--- a/envoy/network/address.h
+++ b/envoy/network/address.h
@@ -161,6 +161,7 @@ public:
    * For IPv4 addresses: "1.2.3.4:80"
    * For IPv6 addresses: "[1234:5678::9]:443"
    * For pipe addresses: "/foo"
+   * For internal addresses: "serverListenerName-<hash of metadata>"
    */
   virtual const std::string& asString() const PURE;
 
--- a/api/envoy/config/core/v3/address.proto
+++ b/api/envoy/config/core/v3/address.proto
@@ -41,6 +41,10 @@ message EnvoyInternalAddress {
     // [#not-implemented-hide:] The :ref:`listener name <envoy_v3_api_field_config.listener.v3.Listener.name>` of the destination internal listener.
     string server_listener_name = 1;
   }
+
+  // An optional arbitrary string that can be used to differentiate multiple internal addresses with the same
+  // :ref:`listener name <envoy_v3_api_field_config.core.v3.EnvoyInternalAddress.address_name_specifier>`.
+  string logical_port = 2;
 }
 
 // [#next-free-field: 7]
--- a/source/common/network/address_impl.cc
+++ b/source/common/network/address_impl.cc
@@ -404,10 +404,11 @@ absl::Status PipeInstance::initHelper(const sockaddr_un* address, mode_t mode) {
 }
 
 EnvoyInternalInstance::EnvoyInternalInstance(const std::string& address_id,
-                                             const SocketInterface* sock_interface)
+                                             const SocketInterface* sock_interface,
+                                             const std::string& logical_port)
     : InstanceBase(Type::EnvoyInternal, sockInterfaceOrDefault(sock_interface)),
-      internal_address_(address_id) {
-  friendly_name_ = absl::StrCat("envoy://", address_id);
+      internal_address_(address_id, logical_port) {
+  friendly_name_ = absl::StrCat("envoy://", address_id, ":", logical_port);
 }
 
 bool EnvoyInternalInstance::operator==(const Instance& rhs) const {
--- a/source/common/network/address_impl.h
+++ b/source/common/network/address_impl.h
@@ -338,7 +338,8 @@ public:
    * Construct from a string name.
    */
   explicit EnvoyInternalInstance(const std::string& address_id,
-                                 const SocketInterface* sock_interface = nullptr);
+                                 const SocketInterface* sock_interface = nullptr,
+                                 const std::string& logical_port = "");
 
   // Network::Address::Instance
   bool operator==(const Instance& rhs) const override;
@@ -352,10 +353,13 @@ public:
 
 private:
   struct EnvoyInternalAddressImpl : public EnvoyInternalAddress {
-    explicit EnvoyInternalAddressImpl(const std::string& address_id) : address_id_(address_id) {}
+    explicit EnvoyInternalAddressImpl(const std::string& address_id,
+                                      const std::string& logical_port)
+        : address_id_(address_id), logical_port_(logical_port) {}
     ~EnvoyInternalAddressImpl() override = default;
     const std::string& addressId() const override { return address_id_; }
     const std::string address_id_;
+    const std::string logical_port_;
   };
   EnvoyInternalAddressImpl internal_address_;
 };
--- a/source/common/network/resolver_impl.cc
+++ b/source/common/network/resolver_impl.cc
@@ -2,6 +2,7 @@
 
 #include "envoy/common/exception.h"
 #include "envoy/config/core/v3/address.pb.h"
+#include "envoy/config/core/v3/base.pb.h"
 #include "envoy/network/address.h"
 #include "envoy/network/resolver.h"
 #include "envoy/registry/registry.h"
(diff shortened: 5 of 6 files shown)


zuercher:
You'll need to fix the DCO: https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#dco-sign-your-work

source/common/upstream/eds.cc
```diff
@@ -133,9 +133,11 @@ void EdsClusterImpl::BatchUpdateHelper::updateLocalityEndpoints(
     const envoy::config::endpoint::v3::LocalityLbEndpoints& locality_lb_endpoint,
     PriorityStateManager& priority_state_manager, absl::flat_hash_set<std::string>& all_new_hosts) {
   const auto address = parent_.resolveProtoAddress(lb_endpoint.endpoint().address());
+
   // When the configuration contains duplicate hosts, only the first one will be retained.
+  // An exception is made for internal addresses as passthrough metadata is usually what makes them unique.
   const auto address_as_string = address->asString();
-  if (all_new_hosts.count(address_as_string) > 0) {
+  if (address->envoyInternalAddress() == nullptr && all_new_hosts.count(address_as_string) > 0) {
```
kyessenov, line 140: I think this could be a bit improved:
```
if (address->type() != Type::EnvoyInternal) {
  const auto address_as_string = address->asString();
  if (all_new_hosts.count(...)) {
    return;
  }
  all_new_hosts.emplace(address_as_string);
}
```

That will avoid counting internal address and unnecessarily creating a string.

stevenctl: as a cpp newb.. wouldn't internaladdr != null already avoid the counting? Just double checking.

kyessenov: Yeah, but you're still putting internal hosts into a map for no reason.

stevenctl: The emplace call has to happen regardless of address->type(), so seems we should always create the string.

stevenctl: Didn't see this comment ^ 

Are we certain this map is only used for dedupe? It eventually gets passed here:

https://github.com/envoyproxy/envoy/blob/86eef10a935b9fb3e241c3c998c8d5babb4b3781/source/common/upstream/upstream_impl.cc#L1814-L1824

kyessenov: It looks like it's trying to correlate the priority set and the new hosts map, so we should keep them consistent. I wonder if we should make the map be from a pointer instead. Let me look deeper at the usage.

kyessenov (changes requested):
I suspect this is not going to work as expected for addition/removal of internal address hosts with the same address. All the complex logic for diffing host sets relies on `HostMap` which itself relies on `addressAsString`. So something is likely to break (either removal not causing drain, or addition not affecting priority set). I don't see how we can easily change to not use `addressAsString` because hosts are re-created it seems on every EDS update. So what we need is a structural hash `hash(internal_address_name, endpoint metadata)` to be the address-as-string. I think we could simply append hash of the endpoint metadata to internal addresses? WDYT?

kyessenov:
Another idea: change host address to support "extensions" to internal address identifier. That would be a protobuf change, but we could just add an opaque string as a designator of "host uniqueness" without affecting "network identity".

stevenctl:
```
2022-08-01T11:08:32.074168Z     debug   envoy filter    original_dst: new connection accepted
2022-08-01T11:08:32.074204Z     debug   envoy filter    workload metadata: new connection accepted
2022-08-01T11:08:32.074238Z     debug   envoy filter    original_dst: new connection accepted
2022-08-01T11:08:32.074325Z     debug   envoy filter    [C18] new tcp proxy session
2022-08-01T11:08:32.074358Z     debug   envoy filter    [C18] Creating connection to cluster spiffe://cluster.local/ns/echo/sa/a_to_http_b.echo.svc.cluster.local_outbound_internal
2022-08-01T11:08:32.074426Z     debug   envoy misc      Allocating TCP conn pool
2022-08-01T11:08:32.074468Z     debug   envoy pool      trying to create new connection
2022-08-01T11:08:32.074481Z     debug   envoy pool      creating a new connection (connecting=0)
2022-08-01T11:08:32.074626Z     debug   envoy connection        [C19] connecting to envoy://outbound_tunnel_lis_spiffe://cluster.local/ns/echo/sa/a:10.244.1.5_18080
2022-08-01T11:08:32.074651Z     debug   envoy io        user namespace handle 0x55ff14f12000 connect to previously closed peer envoy://outbound_tunnel_lis_spiffe://cluster.local/ns/echo/sa/a:10.244.1.5_18080.
2022-08-01T11:08:32.074664Z     debug   envoy connection        [C19] immediate connect error: Invalid argument
2022-08-01T11:08:32.074696Z     debug   envoy conn_handler      [C18] new connection from 10.244.1.4:42088
2022-08-01T11:08:32.074740Z     debug   envoy connection        [C19] raising immediate error
2022-08-01T11:08:32.074747Z     debug   envoy connection        [C19] closing socket: 0
2022-08-01T11:08:32.074758Z     debug   envoy pool      [C19] client disconnected, failure reason: immediate connect error: Invalid argument
2022-08-01T11:08:32.074777Z     debug   envoy filter    [C18] Creating connection to cluster spiffe://cluster.local/ns/echo/sa/a_to_http_b.echo.svc.cluster.local_outbound_internal
2022-08-01T11:08:32.074792Z     debug   envoy connection        [C18] closing data_to_write=0 type=1
2022-08-01T11:08:32.074802Z     debug   envoy connection        [C18] closing socket: 1
2022-08-01T11:08:32.074859Z     debug   envoy conn_handler      [C18] adding to cleanup list
2022-08-01T11:08:32.074899Z     debug   envoy pool      invoking idle callbacks - is_draining_for_deletion_=false
[2022-08-01T11:08:32.074Z] "- - -" 0 UF,URX - - "immediate_connect_error:_Invalid_argument" 0 0 0 - "-" "-" "-" "-" "envoy://outbound_tunnel_lis_spiffe://cluster.local/ns/echo/sa/a:10.244.1.5_18080" spiffe://cluster.local/ns/echo/sa/a_to_http_b.echo.svc.cluster.local_outbound_internal - 10.96.8.66:80 10.244.1.4:42088 - - outbound capture listener
[2022-08-01T11:08:32.074Z] "- - -" 0 UF,URX - - "immediate_connect_error:_Invalid_argument" 0 0 0 - "-" "-" "-" "-" "envoy://outbound_tunnel_lis_spiffe://cluster.local/ns/echo/sa/a:10.244.1.5_18080" spiffe://cluster.local/ns/echo/sa/a_to_http_b.echo.svc.cluster.local_outbound_internal - 10.96.8.66:80 10.244.1.4:42088 - - capture outbound (no pep)

```

I think changing asString on address has other implications.

kyessenov:
I can't tell what the issue is from the log. Can you reproduce it in a test with a standalone envoy?

kyessenov:
There's a difference between downstream and upstream. You don't want to have logical port on the downstream, so we need to strip the port there.

kyessenov:
I think we need to separate listener binding from address like in https://github.com/envoyproxy/envoy/pull/22274/files

envoy/network/address.h
```diff
@@ -161,6 +161,7 @@ class Instance {
    * For IPv4 addresses: "1.2.3.4:80"
    * For IPv6 addresses: "[1234:5678::9]:443"
    * For pipe addresses: "/foo"
+   * For internal addresses: "serverListenerName-<hash of metadata>"
```
htuch, line 164: Is the port directly available as a property of the Address?

kyessenov: I have another API PR that also changes how binding works https://github.com/envoyproxy/envoy/pull/22274, because we don't want to bind to individual "ports". It's merely an endpoint identifier for diff tracking.

htuch: @kyessenov will let you guide the review for API here and let me know when ready for another pass. It  seems #22274 is solving the broader problem here.

stevenctl:
Seems like https://github.com/envoyproxy/envoy/pull/22274 is the fix

Issue #22404, opened 2022-07-26
Tests flaking due to dockerhub issues - eg `run_envoy_docker` and sandboxes


phlax:
CI fails fairly frequently due to bad dockerhub endpoint:

```console
Generating script.
Script contents:
ci/run_envoy_docker.sh 'ci/upload_gcs_artifact.sh /source/generated/fuzz_coverage fuzz_coverage'
========================== Starting Command Output ===========================
/bin/bash --noprofile --norc /srv/azure-pipelines/_work/_temp/8788096c-b5e6-4a3b-8f8a-3c150786c997.sh
Error response from daemon: Get https://registry-1.docker.io/v2/: EOF

```
https://dev.azure.com/cncf/envoy/_build/results?buildId=113688&view=logs&j=7b6d60aa-b5de-5f9b-678c-063d31881e76&t=0c8bcd19-14ac-5aba-0af8-facf9399c298&l=14

as this seems to be a transient issue that resolves quickly, we could probably mitigate this with a retry in `run_envoy_docker`

phlax:
we could also remove any unnecessary use of it - eg in the artefact upload jobs

phlax:
cc @lizan

daixiang0:
Another idea is that we host image tarball at our side if run in CI env, load it when use since CI image won't change often.

phlax:
> Another idea is that we host image tarball at our side

its not that simple - the docker image is huge and the cache is not much faster than the pulling directly from dockerhub

essentially, it _can_ be faster doing this but iff there are no additonal steps required (a big iff)

see #22093 for an implementation

phlax:
This problem seems to be happening across images and randomly but fairly frequently so ive opened an upstream ticket here https://github.com/docker/hub-feedback/issues/2262

daixiang0:
@phlax seems we need to add retry step to fix flake tests.

phlax:
> @phlax seems we need to add retry step to fix flake tests.

the problem with adding a retry - apart from being an imperfect solution that may/not work, is that the envoy-build image is one of many cases where we are pulling images, so just adding a retry on that is not going to resolve the problem

phlax:
one observation is that the bigger image that its pulling, the more likely to fail - which i think is why we see this more frequently with the envoy-build image and skywalking, jaeger, etc

another point of note here is that while the dockerhub issue is by far the biggest flake atm for verify examples - this can also flake on debian/ubuntu updates and pip downloads 

[...]

PR #22423, opened 2022-07-27
ci: retry docker pull

daixiang0:
Signed-off-by: Loong Dai <loong.dai@intel.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #22404
[Optional Deprecated:]

### DIFF
--- a/ci/run_envoy_docker.sh
+++ b/ci/run_envoy_docker.sh
@@ -78,7 +78,12 @@ if ! is_windows; then
     VOLUMES+=(-v "${SHARED_TMP_DIR}":"${SHARED_TMP_DIR}")
 fi
 
-time docker pull "${ENVOY_BUILD_IMAGE}"
+# Retry docker pull with 3 times
+i=0
+while [[ i -lt 3 ]]; do
+  time docker pull "${ENVOY_BUILD_IMAGE}" && break
+  sleep 5
+done
 
 
 # Since we specify an explicit hash, docker-run will pull from the remote repo if missing.


ci/run_envoy_docker.sh
```diff
@@ -78,7 +78,12 @@ if ! is_windows; then
     VOLUMES+=(-v "${SHARED_TMP_DIR}":"${SHARED_TMP_DIR}")
 fi
 
-time docker pull "${ENVOY_BUILD_IMAGE}"
+# Retry docker pull with 3 times
+i=0
+while [[ i -lt 3 ]]; do
+  time docker pull "${ENVOY_BUILD_IMAGE}" && break
```
phlax, line 84: not sure this would work - its not catching the error

also i think we want some fail condition  if it fails after all retries

see https://github.com/envoyproxy/envoy/blob/14567bf0d16de6e052913a6582c369ac518c360e/examples/verify-common.sh#L149-L165 for a possible implementation that does both

daixiang0: I test it with two image, one is existed another is not, both OK.

If you want to record more info, I will update it.

phlax: hmm, doesnt make sense to me - its just not catching the error 

either way - i think we may be able to avoid adding this (for now) - see https://github.com/docker/hub-feedback/issues/2262#issuecomment-1197828295

daixiang0: Covert to draft, let's see recent CI results.

phlax:
fail here: https://dev.azure.com/cncf/envoy/_build/results?buildId=113915&view=logs&j=1c04237c-5f52-5959-5f83-5522b3556331&t=db3f3889-c922-566b-fa2e-4bc4948bc467&l=15

~1am utc 29/7

phlax:
and another https://dev.azure.com/cncf/envoy/_build/results?buildId=114001&view=logs&j=681c7800-c42c-5f1e-f2ee-699120aa5784&t=9a535103-4675-5a3d-3e9c-59a6e6997567&l=14

PR #22424, opened 2022-07-27
admin: support HTML UI for admin post params

jmarantz:
Commit Message: follow-up to #19546 -- supports parameter declarations for admin POST endpoints. This required an infrastructural improvement to generate queryParams from the AdminStream, which has access to request body, as POST params from HTML form show up as POSTs params.

That in turn required that we ensure that request_headers_->getPathValue() be consistent with the path in test flows, which needed to be sorted out in the test infrastructure. Added an ASSERT to ensure they stay that way.
Additional Description:
Risk Level: low -- incremental change to support POST admin params.
Testing: //test/..., and manually ran server to verify html post behavior, fixing a couple of cases where the HTML controls didn't pass through params as expected by the handlers.
Docs Changes: n/a
Release Notes: n/a


source/server/admin/admin.cc
```diff
@@ -76,6 +76,17 @@ void AdminImpl::startHttpListener(const std::list<AccessLog::InstanceSharedPtr>&
   }
 }
 
+namespace {
+std::vector<absl::string_view> prepend(const absl::string_view first,
+                                       const std::vector<absl::string_view>& strings) {
+  std::vector<absl::string_view> v;
```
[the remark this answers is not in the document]
jmarantz: Oh wait --  what did you mean? It was using vector::insert already. I think this is would be slightly more efficient than copying the vector as is and then inserting a new beginning, as it right-size allocs it first.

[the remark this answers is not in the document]
pradeepcrao: My hunch is insert will be faster, as it will result in moves v/s the explicit copying within prepend. Can this function not just take an r-value `strings`, insert `first` at the beginning and return it (since I don't see any other use case for it in the PR).

source/server/admin/admin.cc
```diff
@@ -139,16 +154,29 @@ AdminImpl::AdminImpl(const std::string& profile_path, Server::Instance& server,
...
           // re-plumbing through the admin callback API. See also drain_listeners.
           makeHandler("/logging", "query/change logging levels",
-                      MAKE_ADMIN_HANDLER(logs_handler_.handlerLogging), false, true),
-
+                      MAKE_ADMIN_HANDLER(logs_handler_.handlerLogging), false, true,
+                      {{Admin::ParamDescriptor::Type::String, "paths",
+                        "Change multiple logging levels by setting to "
+                        "<logger_name1>:<desired_level1>,<logger_name2>:<desired_level2>."},
+                       {Admin::ParamDescriptor::Type::Enum, "level", "desired logging level",
+                        prepend("", LogsHandler::levelStrings())}}),
```
pradeepcrao, line 162: what does empty level do? I could not find this info here https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging

source/server/admin/admin.cc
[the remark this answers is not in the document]
jmarantz: Sorry -- I'm confused. I *am* using *insert*.

BTW this code doesn't need to be fast at all; it just happens once on startup. I think the main argument here is not whether to use insert, but whether it's worth doing the reserve.

I don't think move vs copy makes much difference: this is a string-view.

But I see Pradeep's point: this is a local helper so I could just make it modify the strings in-place not not bother doing the copy, which makes this non-perf-critical code have fewer lines :)

source/server/admin/logs_handler.cc
```diff
@@ -29,28 +28,34 @@ absl::flat_hash_map<absl::string_view, spdlog::level::level_enum> buildLevelMap(
...
-
-      response.add("usage: /logging?<name>=<level> (change single level)\n");
-      response.add(
-          "usage: /logging?paths=name1:level1,name2:level2,... (change multiple levels)\n");
-      response.add("usage: /logging?level=<level> (change all levels)\n");
-      response.add("levels: ");
-      for (auto level_string_view : spdlog::level::level_string_views) {
-        response.add(fmt::format("{} ", level_string_view));
-      }
+Http::Code LogsHandler::handlerLogging(absl::string_view, Http::ResponseHeaderMap&,
```
pradeepcrao, line 40: Is the first parameter used for any handler, since it's now available on the AdminStream? Is this no longer needed on `MAKE_ADMIN_HANDLER`? Also several functions in admin.cc pass `path_and_query` around as arguments. Should all of these just get it from AdminStream?

envoy/server/admin.h
```diff
@@ -102,7 +110,7 @@ class Admin {
     enum class Type { Boolean, String, Enum };
     const Type type_;
     const std::string id_;   // HTML form ID and query-param name (JS var name rules).
-    const std::string help_; // Help text rendered into UI.
+    const std::string help_; // Rendered into home-page HTML and /help text.
```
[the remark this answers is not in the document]
pradeepcrao: Sorry, I meant in the help text output.

source/server/admin/logs_handler.cc
pradeepcrao: This would require AdminFilter to store the query params once parsed.

jmarantz: done as a pending follow-up PR: https://github.com/envoyproxy/envoy/pull/22571 -- it's not reviewable now but once this merges, that one will be non-functional so that its sprawl is easier to reason about.

pradeepcrao (dismissed):
LGTM

jmarantz:
Yan, can you do a senior maintainer review?

Thanks!

.bazelversion
```diff
@@ -1 +1 @@
-6.0.0-pre.20220706.4
+5.0.0
```
yanavlasov, line 1: This looks like merge error

yanavlasov (changes requested):
/wait

.bazelversion
jmarantz: yeah I'm not sure how that snuck in; done.

yanavlasov (approved):

PR #22428, opened 2022-07-27
test: changing an 8s timeout to 2s

alyssawilk:
Risk Level: low
Testing: still passes
Docs Changes: n/a
Release Notes: n/a


RyanTheOptimist (approved):
Nice!

RyanTheOptimist (commented):
Is this really test only since it changes code in source/?

PR #22433, opened 2022-07-27
Fix the analisys of getsockopt() call

jwendell:
The syscall might change the optlen value to hold the actual value of
the resturned value in optval.

In my case it's reducing that value to 192, and thus some tests are
failing.

Signed-off-by: Jonh Wendell <jonh.wendell@redhat.com>

### DIFF
--- a/source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
+++ b/source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
@@ -79,7 +79,7 @@ absl::optional<struct tcp_info> TcpStatsSocket::querySocketInfo() {
   memset(&info, 0, sizeof(info));
   socklen_t optlen = sizeof(info);
   const auto result = callbacks_->ioHandle().getOption(IPPROTO_TCP, TCP_INFO, &info, &optlen);
-  if ((result.return_value_ != 0) || (optlen < sizeof(info))) {
+  if (result.return_value_ != 0) {
     ENVOY_LOG(debug, "Failed getsockopt(IPPROTO_TCP, TCP_INFO): rc {} errno {} optlen {}",
               result.return_value_, result.errno_, optlen);
     return absl::nullopt;
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -12,6 +12,9 @@ minor_behavior_changes:
 
 bug_fixes:
 # *Changes expected to improve the state of the world and are unlikely to have negative effects*
+- area: transport_socket
+  change: |
+    fixed a bug that prevented the tcp stats to be retrieved when running on kernels different than the kernel where Envoy was built.
 
 removed_config_or_runtime:
 # *Normally occurs at the end of the* :ref:`deprecation period <deprecated>`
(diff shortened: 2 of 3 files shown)


jwendell:
@ggreenway Please take a look.

source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
```diff
@@ -79,7 +79,7 @@ absl::optional<struct tcp_info> TcpStatsSocket::querySocketInfo() {
   memset(&info, 0, sizeof(info));
   socklen_t optlen = sizeof(info);
   const auto result = callbacks_->ioHandle().getOption(IPPROTO_TCP, TCP_INFO, &info, &optlen);
-  if ((result.return_value_ != 0) || (optlen < sizeof(info))) {
+  if (result.return_value_ != 0) {
```
ggreenway, line 82: This change would silently not count the unpopulated fields.

What kernel version did this fail on? I'm guessing it's quite old. I think if you want to handle older kernels, you should figure out what is and is not valid in the tcp_info, and not create the stats for the fields that aren't supported, instead of publishing zeros for them with no indication that the values aren't being set correctly.

/wait

ggreenway (changes requested):

source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
jwendell: 4.15.0-162-generic.

If I understand it right, this code assumes the kernel where Envoy is built must match the kernel where it runs?

ggreenway: Oh, I see what went wrong, I think. I looked at the version of `linux/tcp.h` in the build image and didn't think about other build environments.

The current assumption is that the running kernel version be >= kernel headers built against. I think we can substantially relax that by computing the offset of the last field we use in the struct, and setting optlen equal to that. Then we'll only require a kernel that provides the exact fields we're using. The kernel ABI guarantees that fields can be added, but no old fields will change size or be removed.

yanavlasov: We may not need to adjust the length, I think it may just be ok that unpopulated fields are 0. Some stats may be wrong on kernels which do not support all options but it should not be a problem otherwise.

jwendell:
@ggreenway @yanavlasov I'm not sure what's the best approach here, see an alternative PR, as suggested  by Greg: https://github.com/envoyproxy/envoy/pull/22451

source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
yanavlasov: I personally like this approach better. The approach with offset of the last field we use is brittle. Someone can use a different field and have dig for why it is always 0. Functionally this approach and #22451 are the same but this one seems to be less error prone.

ggreenway: @yanavlasov raises good points. I'm ok with keeping this approach.

ggreenway (approved):

ggreenway (changes requested):
Please add a release note

/wait

jwendell:
@ggreenway done.

ggreenway (approved):

PR #22434, opened 2022-07-27
WIP: adding upstream filters

alyssawilk:
This is NOT READY FOR SUBMISSION

needs
1) all the tests
2) better ownership on the request path, to handle filters with pauses
3) watermark fixes
4) all the TODOs addressed
5) runtime guard (default off)

For an explanation of how to fill out the fields, please see the relevant section
in [PULL_REQUESTS.md](https://github.com/envoyproxy/envoy/blob/main/PULL_REQUESTS.md)
-->

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Fixes https://github.com/envoyproxy/envoy/issues/10455


alyssawilk:
cc @adisuissa @paul-r-gall 

Paul: as said while all the tests are passing, it's still going to be O(month) before this is ready (a few weeks of coding, and I'm out for 10 days in August).  That said I think it's passing all the tests with the major caveats above.

@adisuissa sending this out as a draft early because I wanted to get your thoughts on config.  I'd envisioned this as a cluster config thing, but @paul-r-gall pointed out it'd be nice to have a default upstream filter chain, for the filters one always wants on, so you don't have to copy per cluster.  Would we we OK adding that and if so I assume it'd go directly in bootstrap?
cc @markdroth @envoyproxy/api-shepherds for other interested parties

adisuissa:
> @adisuissa sending this out as a draft early because I wanted to get your thoughts on config. I'd envisioned this as a cluster config thing, but @paul-r-gall pointed out it'd be nice to have a default upstream filter chain, for the filters one always wants on, so you don't have to copy per cluster. Would we we OK adding that and if so I assume it'd go directly in bootstrap? 

A few high-level questions:
- Does using upstream http filters override the "downstream http filters" or will these filters be concatenated.
- For the bootstrap defined variant, what will be the semantics? Will it be a default list of filters that could be overridden by an explicit settings of a per-cluster upstream filters? Will these filters always be concatenated to the per-cluster upstream filters?

alyssawilk:
also cc @snowp for these and other thoughts in case he has input from when he was working on it.

1) Does using upstream http filters override the "downstream http filters" or will these filters be concatenated.
AFIK concatenated, as I don't see an easy way to make upstream filter chain creation aware and deduping downstream filter chain.

2) For the bootstrap defined variant, what will be the semantics? Will it be a default list of filters that could be overridden by an explicit settings of a per-cluster upstream filters? Will these filters always be concatenated to the per-cluster upstream filters?
Ugh, that's a really good question which hadn't even occurred to me.  I assume we'd always fully list all filter chains, since if we try to mix and match it's not obvious if we'd prepend or append.  I think I'm going to start with absolute filters but it makes me wonder if instead of adding a manual filter chain list in the bootstrap* I should have a message which currently only contains the default filter chain (which gets 100% overridden) but we could extend to add a prefix filter chain or suffix filter chain, which would do merging?  And if we never do those two, enh, we just have an extra proto :-P

*also now that I think about it, I think I'd put the default cluster filter chain, or ClusterFilterChainDefaults method in the clusters proto rather than the bootstrap proto, as it is cluster specific and it keeps bootstrap clean.  And having a message for it may not be bad as we can also stash the upgrade filter chain in there when we inevitably get around to implementing it.

PR #22435, opened 2022-07-27
Thrift: stats for downstream connection close

JuniorHsu:

Commit Message:
We want to keep downstream connections as much as possible since SR drops usually come from cx close.
Therefore, we want more stats for reasoning.

Moreover, adding more debug log to see reset reason.

Unfortunately we didn't test for partial response during connection close, which might be an followup.

Additional Description: 
Risk Level: low
Testing: unit
Docs Changes: current.yaml
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


zuercher:
`Code coverage for source/extensions/filters/network/thrift_proxy/router is lower than limit of 96.4 (96.1)`

I expect you need to exercise the path for `downstream_cx_partial_request_close_` in tests to get the coverage up.

JuniorHsu:
> `Code coverage for source/extensions/filters/network/thrift_proxy/router is lower than limit of 96.4 (96.1)`
> 
> I expect you need to exercise the path for `downstream_cx_partial_request_close_` in tests to get the coverage up.

I add a test. Hopefully it passes the threshold. Thanks.

source/extensions/filters/network/thrift_proxy/router/upstream_request.cc
```diff
@@ -321,6 +322,13 @@ bool UpstreamRequest::onResetStream(ConnectionPool::PoolFailureReason reason) {
     // downstream.
     if (response_underflow_ || response_state_ == ResponseState::Started) {
       ENVOY_LOG(debug, "reset downstream connection for a partial or underflow response");
+      if (response_underflow_) {
+        std::cout << "XXX response underflow" << std::endl;
```
rgs1, line 326: drop the cout

source/extensions/filters/network/thrift_proxy/router/upstream_request.cc
```diff
@@ -321,6 +322,13 @@ bool UpstreamRequest::onResetStream(ConnectionPool::PoolFailureReason reason) {
...
+        stats_.incCloseUnderflowResponse(parent_.cluster());
+      } else {
+        std::cout << "XXX partial response" << std::endl;
```
rgs1, line 329: drop the cout (or make it a log statement)

test/extensions/filters/network/thrift_proxy/router_test.cc
```diff
@@ -548,7 +548,7 @@ class ThriftRouterTestBase {
   }
 
   void returnResponse(MessageType msg_type = MessageType::Reply, bool is_success = true,
-                      bool is_drain = false) {
+                      bool is_drain = false, bool is_partial_response = false) {
```
rgs1, line 551: nit: `is_partial` is probably enough since we know we are talking about responses

rgs1 (approved):
Thanks!

alyssawilk:
looks like this needs a main merge?

JuniorHsu:
Did an upstream merge to resolve the conflict

zuercher (approved):

Issue #22336, opened 2022-07-20
Combined RBAC IP filter with Unified Matcher usage


znqjnny:
Hi! We are trying to do an IP filter for thousands of IPs. We planning to use RBAC for IP filtering and combined it with the Unified Matcher.

We have some questions about RBAC and Unified Matcher:

1. Question about RBAC:
Could you give us some idea about how to extract the IP for matching, and also how to match the IP with the subnet for filtering?

2. Question about UnifiedMatcher
From what I learned about the Unified Matcher, looks like the data structure needs to be built on our own,  if we would like to do a Trie match, we should build with nested matcher_tree using prefix_match_map and exact_match_map. And since we have thousands IP to block, that seems results to a disaster since we need to build the matcher with a subtree for each IP.
I’m wondering do you have any suggestions on how to build this feature efficiently? And please correct me if I’m understanding wrong on it.

3. Question about combining RBAC filter with UnifiedMatcher
We would like to use UnifiedMatcher to enhance the performance of RBAC IP filter matching. But I didn’t find an example of how we put the Unified Matcher in the RBAC configuration. Could you give us an example if possible?

Thanks so much for the help!

phlax:
cc @snowp

kyessenov:
@zhxie for RBAC and unified matcher integration.

Re 2: We have an LcTrie-based IP matcher implemented https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/ip.proto#envoy-v3-api-msg-xds-type-matcher-v3-ipmatcher that could represent large IP sets efficiently. 

We don't have any concrete solution for sharing "sub-trees" in the matcher tree in general (e.g. multiple identical OnMatch in prefix/exact map). There was some discussion in the past.

zhxie:
Here is an example for using RBAC with generic matching in network/HTTP RBAC filter which will extract the source IP as inputs for matching.

```yaml
# Avoid using `rules` and use `matcher` instead.
matcher:
  matcher_tree:
    input:
      name: envoy.matching.inputs.source_ip
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.matching.common_inputs.network.v3.SourceIPInput
    # Change to `custom_match` if you want to use the IP trie.
    exact_map_match:
      map:
        "127.0.0.1":
          action:
            name: action
            typed_config:
              "@type": type.googleapis.com/envoy.config.rbac.v3.Action
              name: allow-localhost
              action: ALLOW
```

[...]

PR #22439, opened 2022-07-28
examples: add RBAC sandbox

zhxie:

Commit Message: examples: add RBAC sandbox
Additional Description: Envoy does not have a RBAC example currently and there are some issues (e.g., #22336) about how to use the RBAC filter. The patch adds a RBAC sandbox with generic matching.
Risk Level: Low
Testing: N/A
Docs Changes: Sandbox
Release Notes: N/A


examples/rbac/verify.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+export NAME=rbac
+export PORT_PROXY="${RBAC_PORT_PROXY:-10000}"
```
phlax, line 4: the reason for doing this is that we want all of the examples to use different ports (this will allow the examples to be verified simultaneously)

my very hokey system for doing it involves the first letter and its place in the alphabet + an additional integer (prefixed by 1)

in this case lets use:

`11810` and `11811`

examples/rbac/verify.sh
```diff
@@ -0,0 +1,17 @@
...
+export PORT_ADMIN="${RBAC_PORT_ADMIN:-10001}"
+
+# shellcheck source=examples/verify-common.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../verify-common.sh"
+
+run_log "Test upstream with access denied response"
+responds_with "RBAC: access denied" "http://localhost${PORT_PROXY}"
+
+run_log "Test upstream without access denied response"
```
phlax, line 13:
```suggestion
run_log "Test authorized upstream response"
```

examples/rbac/envoy.yaml
```diff
@@ -0,0 +1,79 @@
...
+                              google_re2: {}
+                              regex: https?://(www.)?envoyproxy.io/docs/envoy.*
+                    on_match:
+                      action:
+                        name: action
+                        typed_config:
+                          "@type": type.googleapis.com/envoy.config.rbac.v3.Action
+                          name: illegal-referer
+                          action: DENY
+                on_no_match:
```
phlax, line 50: it seems odd that the `on_match` and `on_no_match` have different indentations - but i guess that is just how the matcher works

examples/rbac/docker-compose.yaml
```diff
@@ -0,0 +1,16 @@
...
+      dockerfile: Dockerfile-envoy
+    ports:
+    - "${PORT_PROXY:-10000}:10000"
+    - "${PORT_ADMIN:-10001}:10001"
+
+  backend:
+    build:
+      context: .
+      dockerfile: Dockerfile-backend
+    ports:
```
phlax, line 15: this isnt required

phlax (changes requested):
@zhxie thanks for working on this, i think its a very welcome addition, and the sandbox is in pretty good shape already

one thing im wonder is where "roles" come into it if we are just matching a header

phlax:
hmm

```console
envoy_1    | [2022-07-28 08:37:06.877][1][warning][misc] [source/common/protobuf/message_validator_impl.cc:35] field 'envoy.extensions.filters.http.rbac.v3.RBAC.matcher' is marked as work-in-progress. API features marked as work-in-progress are not considered stable, are not covered by the threat model, are not supported by the security team, and are subject to breaking changes. Do not use this feature without understanding each of the previous points.
```

phlax:
altho, im not sure above warning is the cause of the failure

docs/root/start/sandboxes/rbac.rst
```diff
@@ -0,0 +1,93 @@
...
+Step 2: Test RBAC of upstream service
+*************************************
+
+The sandbox is configured with ``10000`` port for upstream service.
+
+If a request is denied, Envoy will refuse the connection with a 403 HTTP
+response code and with the content ``RBAC: access denied``.
+
+In the example, requests should only be allowed if its ``Origin`` header
+matches the regex pattern ``https?://(www.)?envoyproxy.io/docs/envoy.*``.
```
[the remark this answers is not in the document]
zhxie: Fixed, thanks.

docs/root/start/sandboxes/rbac.rst
```diff
@@ -0,0 +1,93 @@
...
+
+  $ curl -si localhost:10000
+  HTTP/1.1 403 Forbidden
+  content-length: 19
+  content-type: text/plain
+  date: Thu, 28 Jul 2022 06:48:43 GMT
+  server: envoy
+
+  RBAC: access denied
+
```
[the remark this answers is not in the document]
zhxie: Added, thanks

docs/root/start/sandboxes/rbac.rst
```diff
@@ -0,0 +1,93 @@
...
+And make another request with proper headers set.
+
+.. code-block:: console
+
+  $ curl -si -H "Referer: https://www.envoyproxy.io/docs/envoy" localhost:10000 | grep HTTP/1.1
```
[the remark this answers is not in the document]
zhxie: Oh, it is a mistake, I have changed it to `Referer`.

docs/root/start/sandboxes/rbac.rst
```diff
@@ -0,0 +1,93 @@
...
+.. code-block:: console
+
+  $ curl -s "http://localhost:10001/stats?filter=rbac"
+  http.ingress_http.rbac.allowed: 1
+  http.ingress_http.rbac.denied: 1
+  http.ingress_http.rbac.shadow_allowed: 0
+  http.ingress_http.rbac.shadow_denied: 0
+
+.. seealso::
+
```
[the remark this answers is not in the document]
zhxie: I have a consideration here. Though this example only presents the usage of RBAC HTTP filter, it still has some introduction about RBAC network filter (e.g., `Envoy supports 2 types for RBAC`). In fact, I want to align the example with local ratelimit, which also has both HTTP and network filter.

So, I am wondering,

1. if we should change the doc title and add the work `HTTP`? and
2. if we should add a ref link to API? If should, which one, HTTP, network, or both?

examples/rbac/verify.sh
zhxie: Thanks. I have been puzzled by these ports in examples. I will update them.

examples/rbac/envoy.yaml
zhxie: Yes, it is the design. In fact, `on_no_match` is a catch-all, so it has a higher level.

examples/rbac/docker-compose.yaml
zhxie: Fixed, thanks.

examples/rbac/Dockerfile-backend
```diff
@@ -0,0 +1 @@
+FROM kennethreitz/httpbin@sha256:2c7abc4803080c22928265744410173b6fea3b898872c01c5fd0f0f9df4a59fb
```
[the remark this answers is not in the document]
zhxie: Changed, thanks.

zhxie:
Ah, `Sign off and commit suggestion` by GitHub breaks the DCO. It is sad and I have to force push them.

zhxie:
> one thing im wonder is where "roles" come into it if we are just matching a header

In generic matching, we have HTTP headers and trailers, source and destination, server name, protocols and TLS cert fields. I am worried that adding more policies would make the example too complicated.

phlax:
>  I am worried that adding more policies would make the example too complicated.

i think it could tolerate a little more complexity - and i guess im just concerned that it doesnt demonstrate too much the things i would expect/hope for

that said - i would be happy to land and iterate with just this example so far, esp as it is addressing an issue that has caused confusion

alyssawilk:
cc @phlax status on this one?  it got flagged as stalled and your last comment was you're happy to land?

phlax (commented):
thanks for iterating so far

> if we should add a ref link to API? If should, which one, HTTP, network, or both?

lets add - i would say add both - even tho this is currently only targeting HTTP

/wait

docs/root/start/sandboxes/rbac.rst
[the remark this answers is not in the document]
zhxie: Fixed, thanks.

phlax (approved):
lgtm, thanks @zhxie for working on this

PR #22442, opened 2022-07-28
subset_lb: add metadata_fallback_policy

### LINKED ISSUE #20063: Load Balancer Subsets: support fallback metadata
When using Load Balancer Subsets we can define a fallback policy which tells envoy what to do if no endpoint matching route/dynamic metadata is found.

But I see no way to define fallback route/dynamic metadata, which will be used if no endpoint matching preferred metadata is found.

Let me illustrate it with following example:

<details>
  <summary>Config - Click to expand!</summary>

```yaml
static_resources:
  listeners:
  - name: listener
    address:
      socket_address:
        address: 0.0.0.0
        port_value: 6001
    filter_chains:
    - filters:
        typed_config:
          stat_prefix: ingress_http
          route_config:
            name: routes
            virtual_hosts:
            - name: backend
              domains: ["*"]
              routes:
              - match: { prefix: "" }
                route:
                  cluster: backend
          http_filters:
          - name: envoy.filters.http.header_to_metadata
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.header_to_metadata.v3.Config
              request_rules:
                - header: x-version
                  on_header_present:
                    metadata_namespace: envoy.lb
                    key: version
                    type: STRING
                  remove: false
  clusters:
  - name: backend
    type: STRICT_DNS
    lb_subset_config:
      fallback_policy: ANY_ENDPOINT
      subset_selectors:
      - keys: ["version"]
        fallback_policy: NO_FALLBACK
    load_assignment:
      cluster_name: backend
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: backend
                port_value: 10001
          metadata:
            filter_metadata:
              envoy.lb:
                version: v1
        - endpoint:
            address:
              socket_address:
                address: backend
                port_value: 10002
          metadata:
            filter_metadata:
              envoy.lb:
                version: v2
        - endpoint:
            address:
              socket_address:
                address: backend
                port_value: 10003
          metadata:
            filter_metadata:
              envoy.lb:
                version: v3
```

</details>

There we have three endpoints of the `backend` cluster:

| endpoint | metadata    |
|----------|-------------|
| 1        | version: v1 | 
| 2        | version: v2 |
| 3        | version: v3 |

We also map header `x-version` to dynamic metadata using `header_to_metadata` filter.

I can issue following requests:

| request header `x-version` | responding endpoint         |
|----------------------------|-----------------------------|
| v1                         | 1                           |
| v2                         | 2                           |
| \<missing>                 | 1, 2 or 3                   |
| ???                        | 1 preferred, 2 if 1 is gone |

Last row with `???` is is the subject of the issue. I think it's currently not possible in envoy.

It can be achieved on the client side. Client can issue a request with `x-version: v1` first, then send a second request with `x-version: v2` if the response for the first request returned `no healthy upstream`.

But it would be nice if such flow would be possible entirely on the envoy side. Client would send a single request with 'preferred' and 'fallback' metadata and envoy would execute a fallback logic.

## Question

Is something like this possible in envoy? The solution doesn't have to be limited to lb subsets - if there is any other way please let me know.

Maybe lua script can re-issue a request with another metadata, just like it was send a second time by the client?

## Proposal

However, if this is currently impossible, I propose following extension to Load Balancer Subsets:

1. Introduce new config option in `clusters[].lb_subset_config`: `metadata_fallback_policy` with possible values: `NO_FALLBACK` (default), `FALLBACK_LIST`
  
    Example:
    ```yaml
    lb_subset_config:
        fallback_policy: ANY_ENDPOINT
        metadata_fallback_policy: FALLBACK_LIST
        subset_selectors:
        - keys: ["version"]
    ```

2. When `metadata_fallback_policy` is set to `FALLBACK_LIST`, route/dynamic metadata can have a field `fallback_list` which is a list and every list element is a metadata struct ordered by priority. In our example it would be:

    ```
    fallback_list:
    - version: v1
    - version: v2
    ```

    Such metadata can be set for example by lua (`dynamicMetadata:set`), `header_to_metadata` filter (`PROTOBUF_VALUE` type) or in a route (`metadata_match.filter_metadata`)

    Add `fallback_list` to [well known names](https://github.com/envoyproxy/envoy/blob/c75c1410c8682cb44c9136ce4ad01e6a58e16e8e/source/common/config/well_known_names.h#L44)

2. In [subset_lb.cc](https://github.com/envoyproxy/envoy/blob/c75c1410c8682cb44c9136ce4ad01e6a58e16e8e/source/common/upstream/subset_lb.cc) implement outer loop, that will repeat whole `chooseHost` logic for every metadata struct in `fallback_list` until a host is found

(All names are of course only a proposal and subject to change)

Let me know what you think. Maybe there is a better way to solve this problem?

I can try to implement this if you give a green light for the proposal

MarcinFalkowski:
Signed-off-by: Marcin Falkowski <marcin.falkowski@allegro.pl>

Commit Message: subset_lb: add metadata_fallback_policy

Additional Description:
The problem, context and solution idea is explained here: https://github.com/envoyproxy/envoy/issues/20063

Risk Level:
Low/Medium - a new feature that is disabled by default, but also very small changes to existing logic (no behavior changes expected)

Testing:
* unit tests
* manual testing

Docs Changes:
new protos API field described (`cluster.v3.Cluster.LbSubsetConfig.metadata_fallback_policy`)

Release Notes: added

Fixes #20063

new API feature: metadata_fallback_policy:
* the default value does not cause behavior changes for existing users
* users have a way to disable this change if it causes problems
* it is not plausible that its default value may change in the future


MarcinFalkowski:
https://github.com/envoyproxy/envoy/issues/20063#issuecomment-1070335927

@zuercher I hope this implementation is manageable, or we can make it manageable in the review process :) 

https://github.com/envoyproxy/envoy/issues/20063#issuecomment-1065525691

@rgs1 I would appreciate your review 

BTW. The original issue is closed now due to inactivity. I don't know if that's a problem, but in case it is - I don't know how to reopen it.

zuercher:
Reassigned @markdroth as api shepherd

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -349,6 +395,16 @@ message Cluster {
     // endpoint metadata if the endpoint metadata matches the value exactly OR it is a list value
     // and any of the elements in the list matches the criteria.
     bool list_as_any = 7;
+
+    // Fallback mechanism that allows to try different route metadata until a host is found.
```
zuercher, line 399: One thing that should be made clear (and I'm not sure if that should be here or somewhere else) is how this interacts with the existing LbSubsetFallbackPolicy.

source/common/config/well_known_names.h
```diff
@@ -47,6 +47,8 @@ class MetadataEnvoyLbKeyValues {
   const std::string CANARY = "canary";
   // Key in envoy.lb filter namespace for the key to use to hash an endpoint.
   const std::string HASH_KEY = "hash_key";
+  // Key in envoy.lb filter namespace for providing fallback metadata
+  const std::string FALLBACK_LIST = "fallback_list";
```
zuercher, line 51: Probably should note that this key appears in the RouteAction metadata_match field and has no special meaning elsewhere. (The other two appear in endpoint metadata.)

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -225,6 +225,52 @@ message Cluster {
...
+      //     - version: 2.0
+      //       hardware: c64
+      //     - hardware: c32
+      //     - version: 3.0
+      //
+      // at first, metadata:
+      //
+      // .. code-block:: json
+      //
+      //   {"version": "2.0", "hardware": "c64"}
```
zuercher, line 256: I find it surprising the initial metadata (e.g. `{"version": "1.0"}`) isn't checked first. 

We can't make RouteAction's `metadata_match` a repeated field (see api/API_VERSIONING.md), but I wonder if it wouldn't be better to add a new field in RouteAction that contains the fallbacks. Call it `metadata_match_fallbacks` (or whatever) and it can just be a list of the successive sets of metadata to try after the initial `metadata_match` values.

In that case your example would be 
```
metadata_match: { "envoy.lb": { "version": "2.0", "hardware": "c64"} }
metadata_match_fallbacks: [
  { "envoy.lb": {"version": "1.0", "hardware": "c32" } },
  { "envoy.lb": {"version": "3.0"} }
]
```
and it removes the need for another flag in the subset load balancer.

source/common/config/well_known_names.h
zuercher: It might also be prudent to add a runtime guard around this feature in the unlikely event that someone happens to have metadata with this key and is using the subset load balancer to match on it.

api/envoy/config/cluster/v3/cluster.proto
MarcinFalkowski: > I find it surprising the initial metadata (e.g. {"version": "1.0"}) isn't checked first.

The idea behind it is to allow for having common metadata (in this case `version: 1.0`) that doesn't have to be repeated in every fallback metadata. I think in most cases the keys of fallback metadata and common metadata will be disjoint, but the example shows what happens when they are not disjoint (fallback value overrides the common one).

But I see no problem if we change this logic to: "don't merge common and fallback metadata, just replace it all". 
It will require to specify all metadata keys in every fallback entry, even the constant ones, but IMO it's not a big problem.

Let me know if I should change it, or keep the current behavior.

>  add a new field in RouteAction that contains the fallbacks

What about the other ways for providing route metadata? Namely:
* lua (dynamicMetadata:set)
* header_to_metadata filter (PROTOBUF_VALUE type)

source/common/config/well_known_names.h
MarcinFalkowski: > It might also be prudent to add a runtime guard around this feature in the unlikely event that someone happens to have metadata with this key

`fallback_list` has no special meaning, if the user doesn't explicitly enable `metadata_fallback_policy: FALLBACK_LIST` in config. 
Do you mean we should take another precaution and check if endpoints don't have this metadata key set when user enabled `metadata_fallback_policy: FALLBACK_LIST`?

api/envoy/config/cluster/v3/cluster.proto
MarcinFalkowski: I tried to make it clear in API docs:

```
    // If load balancing process, including all its mechanisms (like
    // :ref:`fallback_policy<envoy_v3_api_field_config.cluster.v3.Cluster.LbSubsetConfig.fallback_policy>`)
    // fails to select a host, this policy decides if and how the process is repeated using another metadata.
```
https://github.com/envoyproxy/envoy/pull/22442/files#diff-cd281142432bf3479ab7fbd00d13d37d5f7dbfc47e6724a8f8fdeb1b6f86553eR400

Do you mean to add this note also here, or to generally make it more clear / explain using other words?

kyessenov:
Needs a main merge.
/assign @lizan 
(for API review as @markdroth is OoO)

source/common/config/well_known_names.h
zuercher: > fallback_list has no special meaning (and can be used as a regular metadata key), if user doesn't explicitly enable metadata_fallback_policy: FALLBACK_LIST in config.

Ok, as long as the keys isn't used unless explicitly configured, that's probably fine.

> LMK if this is what you have in mind

I meant a comment on the constant about where it gets used.

api/envoy/config/cluster/v3/cluster.proto
zuercher: > It is checked, but it's also merged with fallback metadata, so if a key is repeated in both places, fallback value takes precedence.

But it's never checked by itself, right? It's always merged first which was surprising to me.

> I see that fallback may be not a correct word for this behavior.

Yes, maybe a different name would make the behavior clearer.

I don't feel strongly about whether this should be fixed by changing the logic or changing the name.

MarcinFalkowski: OK, thanks for explanation. I will change the logic to work "like a fallback".

I just need a clarification about your idea of adding a new field in RouteAction that contains the fallbacks. I don't see how it can work with mechanisms to provide dynamic metadata, e.g. (lua dynamicMetadata:set)[https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#set]

I think dynamic/route metadata is a generic solution used by many filters, each separated by namespace e.g. "envoy.lb". There are already a code to inject this metadata (lua script, header_to_metadata filter) and pass it to filters.

Creating similar mechanism again (new method in lua API, etc.) for new RouteAction field seems like overkill to me.

zuercher: I suppose it probably doesn't work for other sources.

api/envoy/config/cluster/v3/cluster.proto
zuercher: I think that's fine. Not sure why I missed it before.

api/envoy/config/cluster/v3/cluster.proto
zuercher: To be clear, you can drop that idea.

zuercher (approved):
This looks good to me, but still needs an API approver to stamp it. @markdroth are you available for that?

Issue #21245, opened 2022-05-11
Use Balsa HTTP/1 codec


bencebeky:
The current HTTP/1 codec is [http-parser](https://github.com/nodejs/http-parser).  The plan is to transition Envoy to use [Balsa](https://github.com/google/quiche/tree/main/quiche/balsa) instead.  [edit: fix link]

daixiang0:
@bencebeky hi, could you compare Balsa with http-parser to show the advantage of transition?

wbpcode:
Yeah, is there any reason to make this transition?

adisuissa:
> 

This is related to #10988.
The current http_parser is no longer maintained and adding features such as #18819 is non-trivial.

alyssawilk:
Ah, I should have cross-linked when I was triaging.  The decision points largely documented at https://docs.google.com/document/d/1Z8neuR6GTL61EU8jC88SD2as6Nmq15B-BNWC-SBVKns/edit after a couple of rounds of discussion at the community meeting (notes here https://docs.google.com/document/d/1Yc4zkV-A_cC_R3C0u6D15WQAsRSxADs8p2l8UMMa4P4/edit# as it's not convenient for all time zones)

kanongil:
FYI, the balsa link is broken. They *just* moved it to: https://github.com/google/quiche/tree/main/quiche/balsa.

bencebeky:
no stalebot

bencebeky:
Note that BalsaParser is being developed to provide feature (and bug) parity with http-parser to the highest reasonable extent.  This includes identical error messages, some of which have "HPE_" prefix (for "http-parser error"), and some of which are hardcoded in an awkward manner, see #22447 for an example.  This is to minimize friction during the transition.  After the transision is complete, these error messages, along with other peculiarities, can be eliminated by behavior-changing PR, using runtime flag protection as appropriate.

JuniorHsu:
> Ah, I should have cross-linked when I was triaging. The decision points largely documented at https://docs.google.com/document/d/1Z8neuR6GTL61EU8jC88SD2as6Nmq15B-BNWC-SBVKns/edit 

@alyssawilk This doc is not public or doesn't exist

alyssawilk:
or @yanavlasov codec analysis?  I can't recall :-/

JuniorHsu:
We're keen to do some verification on balsa however black box testing is risky for us. Some document is helpful for us to see the impact/possible regression

[...]

PR #22447, opened 2022-07-28
[balsa] Fix header size limit error messages.

bencebeky:
Change error messages for too large headers that match
ConnectionImpl::checkMaxHeadersSize() behavior.

When http-parser parses a piece of data, it immediately calls
ParserCallbacks::onHeaderField() or ParserCallbacks::onHeaderValue()
with the current fragment of header key or value, and
ConnectionImpl::checkMaxHeadersSize() checks size limit each time.

On the other hand, BalsaFrame buffers the header key and value and only
calls BalsaVisitorInterface::OnHeader() when the given header completes.
In order to prevent from memory attacks, the header size limit is passed
to BalsaFrame through BalsaParser constructor, and BalsaFrame internally
enforces the limit while parsing.

Tracking issue: #21245

Commit Message: [balsa] Fix header size limit error messages.
Additional Description: n/a
Risk Level: low, BalsaParser is behind default-false flag
Docs Changes: n/a
Release Notes: n/a


diannahu (dismissed):
Small functional delta for large test enablement!

bencebeky:
In fact this PR needs a little more work.

/unassign @diannahu @birenroy

bencebeky:
PTAL.  Now it's ready for review.  (Changed error code based on error message returned by Balsa, which allowed for enabling two more tests.)

/assign @diannahu @birenroy

source/common/http/http1/codec_impl.cc
```diff
@@ -687,7 +687,13 @@ Envoy::StatusOr<size_t> ConnectionImpl::dispatchSlice(const char* slice, size_t
 
   const ParserStatus status = parser_->getStatus();
   if (status != ParserStatus::Ok && status != ParserStatus::Paused) {
-    RETURN_IF_ERROR(sendProtocolError(Http1ResponseCodeDetails::get().HttpCodecError));
+    absl::string_view error = Http1ResponseCodeDetails::get().HttpCodecError;
+    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.http1_use_balsa_parser") &&
+        (parser_->errorMessage() == "headers size exceeds limit" ||
+         parser_->errorMessage() == "trailers size exceeds limit")) {
```
diannahu, line 693: optional: What do you think of changing "headers size exceeds limit" and "trailers size exceeds limit" to Http1ResponseCodeDetails::get().HeadersTooLarge in balsa_parser.cc? Then here it can be:

if (feature_enabled && parser_->errorMessage() == HeadersTooLarge) {
  error = HeadersTooLarge;
}

Which reads just a tiny bit simpler here, even if we're coupling a bit more tightly with balsa_parser this way.

If wanting to keep as-is, maybe could define "headers size exceeds limit" and "trailers size exceeds limit" as constants for reuse here and in balsa_parser.cc.

diannahu (approved):

source/common/http/http1/codec_impl.cc
bencebeky: Unfortunately there are two different kind of error messages, see, for example:
https://sourcegraph.com/github.com/envoyproxy/envoy@a00c314/-/blob/test/common/http/http1/codec_impl_test.cc?L2999-3000

My current goal with BalsaParser is feature parity with http-parser, so I'm crafting each error to match.

For reference, when using http-parser, it is not http-parser itself emitting the error, but ConnectionImpl synthesizing it at https://sourcegraph.com/github.com/envoyproxy/envoy@a00c314/-/blob/source/common/http/http1/codec_impl.cc?L551-555, the two kinds of error at the same time.

However, when using BalsaParser, BalsaFrame emits a BalsaFrameEnums::ErrorCode, which is not surfaced directly to ConnectionImpl, but instead converted into an error message string in BalsaParser::HandleError() to match status.message().  Then ConnectionImpl has to synthesize the other kind of error message.

The reason BalsaParser must enforce the size limit internally is that it only calls back to ConnectionImpl when a header key-value pair is parsed completely.  Since we do not want BalsaParser to do unlimited buffering, I'm setting its internal limit to the same as ConnectionImpl's at https://sourcegraph.com/github.com/envoyproxy/envoy@a00c314/-/blob/source/common/http/http1/balsa_parser.cc?L97.  If it wasn't for this complication, we could just rely on the same code path as http-parser does to emit the appropriate error messages.

One option is to surface BalsaFrameEnums::ErrorCode, and move conversion to string logic to ConnectionImpl.  Then ConnectionImpl could set both error message strings based on an enum, not setting one string based on the other.  Note that what you're looking at in this PR is the special case: in all other cases, the responseDetails() kind of error message is HttpCoderError, emitted by ConnectionImpl::dispatchSlice().

Another option is to move this logic to BalsaParser, and let it emit two kinds of error messages: one can be HttpCoderError except when it needs to be HeadersTooLarge, the other BalsaFrameEnums::ErrorCodeToString(error_code) except when http-parser-matching error string is required.

I would prefer to land this PR as-is, and revisit this when all error cases are covered.  A reasonable approach could be changing the behavior so that the error is always HttpCoderError, even when currently it is HeadersTooLarge.  Or a more bold move would be to make status.message() always match responseDetails().  In any case, some cleanup will be in order, because BalsaParser currently emits a number of all-uppercase error messages with HPE_ prefix, which is http-parser specific.  This I plan to do after the BalsaParser transition is done, guarded by a different runtime flag if necessary.  That might be a good time to clean up error generating logic as well.

birenroy:
LGTM

source/common/http/http1/codec_impl.cc
birenroy: Looks like the error message is constructed here: https://sourcegraph.com/github.com/envoyproxy/envoy@a00c314b8a004d1519a464089fc5e0dfc41547fa/-/blob/source/common/http/http1/codec_impl.cc?L589

The changes make sense to me.

birenroy (approved):

source/common/http/http1/codec_impl.cc
diannahu: Thanks for the in-depth explanation of your response! Thanks also for the pointers to the different error messages, interesting. I see a lot of careful crafting to maintain parity with http-parser :D: 

Full support on landing this PR as-is and thinking about revisiting this for clean-up when all error cases are covered.

bencebeky:
Thank you both for the reviews.

bencebeky:
Kevin: PTAL

/assign @KBaichoo

source/common/http/http1/codec_impl.cc
KBaichoo: SGTM, can you track this in the issue for integrating with Balsa, so we do end up cleaning this up? Thanks 
> I would prefer to land this PR as-is, and revisit this when all error cases are covered.

KBaichoo (approved):
lgtm; I think CI is a flake; restarted it

source/common/http/http1/codec_impl.cc
bencebeky: Thanks.  I added comment https://github.com/envoyproxy/envoy/issues/21245#issuecomment-1209476687 for this.  Let me know if you think this needs to be tracked in a more prominent way.

PR #22451, opened 2022-07-28
Fix the analisys of getsockopt() syscall

jwendell:
The syscall might change the optlen value to hold the actual value of
the resturned value in optval. In my case it's reducing that value to 192,
and thus some tests are failing.

We can relax this comparison by comparing the result with the value that
we actually use. If the runtime kernel has a shorter tcp_info struct
than what we need, then we must fail. Otherwise, it's fine.

Signed-off-by: Jonh Wendell <jonh.wendell@redhat.com>

### DIFF
--- a/source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
+++ b/source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
@@ -78,8 +78,14 @@ absl::optional<struct tcp_info> TcpStatsSocket::querySocketInfo() {
   struct tcp_info info;
   memset(&info, 0, sizeof(info));
   socklen_t optlen = sizeof(info);
+
+  // Minimal supported size of the `struct tcp_info`. We rely on that field at build time, so it
+  // must be present in the kernel at runtime.
+  size_t size = offsetof(struct tcp_info, LAST_TCP_INFO_FIELD_WE_USE) +
+                sizeof(info.LAST_TCP_INFO_FIELD_WE_USE);
+
   const auto result = callbacks_->ioHandle().getOption(IPPROTO_TCP, TCP_INFO, &info, &optlen);
-  if ((result.return_value_ != 0) || (optlen < sizeof(info))) {
+  if ((result.return_value_ != 0) || (optlen < size)) {
     ENVOY_LOG(debug, "Failed getsockopt(IPPROTO_TCP, TCP_INFO): rc {} errno {} optlen {}",
               result.return_value_, result.errno_, optlen);
     return absl::nullopt;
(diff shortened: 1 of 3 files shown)


source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
```diff
@@ -78,8 +78,14 @@ absl::optional<struct tcp_info> TcpStatsSocket::querySocketInfo() {
   struct tcp_info info;
   memset(&info, 0, sizeof(info));
   socklen_t optlen = sizeof(info);
+
+  // Minimal supported size of the `struct tcp_info`. We rely on that field at build time, so it
+  // must be present in the kernel at runtime.
+  size_t size = offsetof(struct tcp_info, LAST_TCP_INFO_FIELD_WE_USE) +
```
ggreenway, line 84: I think it would be even better if you set `optlen = size;`, because then the kernel won't fill in later fields, so if someone accidentally adds a use of one of those fields it will always be zero, making it obvious that there's a bug.

test/extensions/transport_sockets/tcp_stats/tcp_stats_test.cc
```diff
@@ -127,14 +127,14 @@ TEST_F(TcpStatsTest, SyscallFailureShortRead) {
   tcp_info_.tcpi_notsent_bytes = 42;
   EXPECT_CALL(io_handle_, getOption(IPPROTO_TCP, TCP_INFO, _, _))
       .WillOnce(Invoke([this](int, int, void* optval, socklen_t* optlen) {
-        *optlen = *optlen - 1;
-        memcpy(optval, &tcp_info_, sizeof(*optlen));
+        *optlen = offsetof(struct tcp_info, LAST_TCP_INFO_FIELD_WE_USE);
```
ggreenway, line 130: Does this need a `- 1` to make this test work as intended?

ggreenway (changes requested):
This approach looks good to me.

/wait

source/extensions/transport_sockets/tcp_stats/tcp_stats.cc
jwendell: agreed

test/extensions/transport_sockets/tcp_stats/tcp_stats_test.cc
jwendell: note that this is just the offset, not offset+sizeof(field). So, it's already a less value than we expect.

jwendell:
@ggreenway Pushed a v2 of this PR.

ggreenway:
Let's go with #22433

Issue #22429, opened 2022-07-27
GrpcJsonTranscoder support other HTTP/Json format for gRPC error status


TAOXUY:
GrpcJsonTranscoder support other HTTP/Json format for gRPC error status

*Description*:
Currently JsonGrpcTranscoder filter only [translates google.rpc.Status into its default Json format](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L856). We want it can support translating into other Json format. The context is that we want to achieve this [HTTP Error format](https://cloud.google.com/apis/design/errors#http_mapping).

TAOXUY:
One solution is we have our custom filter before GrpcJsonTranscoder be responsible for translate gRPC error specifically but we still want all the translation work can be done within GrpcJsonTranscoder. One approach I can think of is to have some new proto format extension(input is a proto and ouput is a String) to be called during [translating gRPC error](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L856) in the GrpcJsonTranscoder.

TAOXUY:
@lizan WDYT?

TAOXUY:
@nareddyt @shuoyang2016

lizan:
As long as it is configurable I think we can do that, can you submit a PR?

PR #22456, opened 2022-07-28
Support custom gRPC error format in GrpcJsonTranscoder

TAOXUY:
Signed-off-by: Xuyang Tao <taoxuy@google.com>

Commit Message: add ProtoTransformer to support custom format for gRPC error beyond the default google.rpc.Status.
Risk Level:no
Testing:done
Docs Changes:done
Release Notes:done
[Optional Runtime guard:]
[Optional Fixes #Issue]https://github.com/envoyproxy/envoy/issues/22429


TAOXUY:
@nareddyt 
@shuoyang2016 
@dafudeng

TAOXUY:
Hi @lizan, the main part is done. Could you take a quick look? Thanks!

lizan:
Taking a step back, do you really need an extensible marshaller other than supporting the [HTTP Error format](https://cloud.google.com/apis/design/errors#http_mapping)?

TAOXUY:
> tensible marshal

Do you have any better idea? 

The another way I can think of is to have another filter before transcoder in encoder filter chain to transcoder gRPC error but I prefer we can do within transcoder filter.

lizan:
If we don't anticipate any other formats, it can be just a flag in transcoder filter to emit the format, which simplifies code a lot.

TAOXUY:
> If we don't anticipate any other formats, it can be just a flag in transcoder filter to emit the format, which simplifies code a lot.

I don't think there would be any other formats. 

That would be great if you are OK to add this error format with a field to toggle in transcoder. Let me quickly update this PR.

TAOXUY:
Hi @lizan, any opinions? If you insist on directly adding that in transcoder filter, I am also fine.

adisuissa (commented):
A high-level question - why is the `ProtoJsonMarshaller` interface not part of the extension code (i.e., should it be moved from source/common/proto_json/ to be part of the grpc_json_transcoder or some other source/extension/common sub-dir)?

TAOXUY:
> I don't think we want to add an extension point given there's only one use case. After re-thinking of this, another filter in front the transcoder filter might be a better fit. If you don't want the configuration to be spread to two filters, you can instantiate a transcoder filter in that filter?

Ohh, I love this idea! Let me try it out locally!

TAOXUY:
Closed this PR as we're going with another solution.

wbpcode:
Please set it ready fro review if it's ready. Thanks. 😸

source/extensions/common/proto_json/proto_json_marshaller.h
```diff
@@ -0,0 +1,37 @@
...
+
+/*
+ * A marshaller for marshaling protobuf into JSON format.
+ */
+template <class ProtoType> class ProtoJsonMarshaller {
+public:
+  virtual ~ProtoJsonMarshaller() = default;
+
+  // Outputs the protobuf into JSON format.
+  virtual Protobuf::util::Status marshal(const ProtoType& proto, std::string& output) const PURE;
```
qiwzhang, line 16: Maybe just Protobuf::Message. not need class template?

test/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter_test.cc
```diff
@@ -1608,6 +1639,46 @@ TEST_F(GrpcJsonTranscoderFilterConvertGrpcStatusTest, SkipTranscodingStatusIfBod
...
+    auto proto_config = bookstoreProtoConfig();
+    proto_config.set_convert_grpc_status(true);
+    proto_config.mutable_grpc_status_json_marshaller()->set_name(
+        "envoy.proto_json_marshaller.transcoder_test_proto_json_marshaller");
+    return proto_config;
+  }
+};
+
+TEST_F(GrpcJsonTranscoderFilterConvertGrpcStatusByExtensionTest,
+       TranscodingBinaryHeaderInTrailerOnlyResponse) {
```
qiwzhang, line 1660: can we add a integration test for this?

TAOXUY: You mean grpc_json_transcoder_integration test. Seems unnecessary to me. The json_transcoder_filter_test should already be enough.

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -897,10 +899,12 @@ bool JsonTranscoderFilter::maybeConvertGrpcStatus(Grpc::Status::GrpcStatus grpc_
...
   std::string json_status;
-  auto translate_status =
-      config_.grpc_status_json_marshaller()
-          ? config_.grpc_status_json_marshaller()->marshal(*status_details, json_status)
-          : per_route_config_->translateProtoMessageToJson(*status_details, &json_status);
+  std::unique_ptr<Protobuf::Message> transformed_grpc_status =
+      config_.grpc_status_proto_transformer() && status_details
+          ? config_.grpc_status_proto_transformer()->transform(*status_details)
+          : nullptr;
+  auto translate_status = per_route_config_->translateProtoMessageToJson(
```
qiwzhang, line 906: How can we make sure translateProtoMessageToJson() is able to translate the proto message into JSON?  Does it have the proto-descriptor to do that translation?

TAOXUY: Update transcoder API to emphasize that descriptor should include the custom error proto type just like the error  details.

qiwzhang (approved):
LGTM.  Thanks

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -225,6 +227,12 @@ message GrpcJsonTranscoder {
   // :ref:`proto descriptor set <config_grpc_json_generate_proto_descriptor_set>`.
   bool convert_grpc_status = 9;
 
+  // Specifies a proto transformer extension that transforms grpc response status into a custom protobuf format.
+  // This field is effective only when ``convert_grpc_status`` is set to true.
+  // The target custom prototype type must be in ``proto_descriptor``.
+  // This field should be an extension of ``envoy.proto_transformer``.
```
adisuissa, line 233: Should be of the form:
`[#extension: envoy.abc.def.ghi]`

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -225,6 +227,12 @@ message GrpcJsonTranscoder {
   // :ref:`proto descriptor set <config_grpc_json_generate_proto_descriptor_set>`.
   bool convert_grpc_status = 9;
 
+  // Specifies a proto transformer extension that transforms grpc response status into a custom protobuf format.
```
adisuissa, line 230: From a high-level I agree that there should be support for transforming the response status, however I'm not sure if we gain much be specifying this to return only a protobuf message. Why not return any valid header value?

cc @lizan that also commented on the issue (#22429).

source/extensions/common/proto_transformer/proto_transformer.h
```diff
@@ -0,0 +1,36 @@
...
+ * A factory for creating ProtoTransformer.
+ */
+template <class ProtoType> class ProtoTransformerFactory : public Envoy::Config::TypedFactory {
+public:
+  ~ProtoTransformerFactory() override = default;
+
+  virtual std::unique_ptr<ProtoTransformer<ProtoType>>
+  createProtoTransformer(const Protobuf::Message& config) PURE;
+
+  std::string category() const override { return "envoy.proto_transformer"; };
```
adisuissa, line 28: Should probably be in a place like `envoy.common.transformers`

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -225,6 +226,15 @@ JsonTranscoderConfig::JsonTranscoderConfig(
   match_incoming_request_route_ = proto_config.match_incoming_request_route();
   ignore_unknown_query_parameters_ = proto_config.ignore_unknown_query_parameters();
   request_validation_options_ = proto_config.request_validation_options();
+  if (proto_config.has_grpc_status_proto_transformer()) {
+    auto* factory = Envoy::Config::Utility::getFactoryByName<
```
adisuissa, line 230: Any reason not to use `getAndCheckFactoryByName`?
This way if there's an invalid extension name the config will be rejected.

adisuissa (commented):
Thanks for working on this!
Left a high-level comment.
Also, would it make sense to add such extension as part of Envoy's extensions?

TAOXUY:
@shuoyang2016 @nareddyt @dafudeng

lizan:
> > I don't think we want to add an extension point given there's only one use case. After re-thinking of this, another filter in front the transcoder filter might be a better fit. If you don't want the configuration to be spread to two filters, you can instantiate a transcoder filter in that filter?
> 
> Ohh, I love this idea! Let me try it out locally!

@TAOXUY What's the context from this suggestion, how it didn't work?

TAOXUY:
> > > I don't think we want to add an extension point given there's only one use case. After re-thinking of this, another filter in front the transcoder filter might be a better fit. If you don't want the configuration to be spread to two filters, you can instantiate a transcoder filter in that filter?
> > 
> > 
> > Ohh, I love this idea! Let me try it out locally!
> 
> @TAOXUY What's the context from this suggestion, how it didn't work?

The wrapper filter solution turns out:
- low testability. Our tests can only cover our intercept logic while we still want to make sure the rest behavior is not changed by copying/maintaining the transcoder filter, which is lots of toil.
- the intercept itself is a little bit risky. For example, the grpc status transcoding could happen in `encodeHeaders/encodeTrailers` and the transcoder itself is a stateful, so we don't want mess up the transcoder introducing undesirable behavior

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
TAOXUY: Sorry, I don't quite follow. We are trying to transform the grpc status format, the output of which should be written in body. Headers won't changed.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
TAOXUY: I noticed there is presubmit checking extension_metadata.yaml about the extension existence. For `envoy.common.proto_transformer`, we only have the extension API no real extension now. Is skipping that presubmit the only way?

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
TAOXUY: Done.

lizan:
> * low testability. Our tests can only cover our intercept logic while we still want to make sure the rest behavior is not changed by copying/maintaining the transcoder filter, which is lots of toil.
> * the intercept itself is a little bit risky. For example, the grpc status transcoding could happen in `encodeHeaders/encodeTrailers` and the transcoder itself is a stateful, so we don't want mess up the transcoder introducing undesirable behavior

I don't quite follow those disadvantage, what I was saying is that you can implement a filter like you said in https://github.com/envoyproxy/envoy/pull/22456#issuecomment-1199566025, then just add that filter before Transcoder, in the config phase like https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/grpc_json_transcoder/config.cc#L22 , then it shouldn't allow your filter to interfere with transcoder status, they will be two separate filter, but just in one config. In that way you can test your filter that just change how status is marshalled.

TAOXUY:
@lizan  IIUC not that simple, if we go with 2 stand-alone filters, there will be much code copy from transcoder for the not trailer only case, like this [test](https://github.com/envoyproxy/envoy/blob/92808ba1960f88016f65511376263ed39a1f8cc5/test/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter_test.cc#L1451-L1475). It needs the encodeHeaders to `StopIteration` selectively, which requires us to copy this logic and the implicitly introduced dependency(like we need to init the method_). I feel  that is very error-prone.

source/extensions/common/proto_transformer/proto_transformer.h
TAOXUY: Done.

TAOXUY:
@lizan, we are also OK if you want us to directly support google apis error format in https://github.com/envoyproxy/envoy/pull/22456#issuecomment-1200050055

lizan:
> @lizan IIUC not that simple, if we go with 2 stand-alone filters, there will be much code copy from transcoder for the not trailer only case(header and then followed by a grpc-status trailers), like this [test](https://github.com/envoyproxy/envoy/blob/92808ba1960f88016f65511376263ed39a1f8cc5/test/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter_test.cc#L1451-L1475). It needs the encodeHeaders to `StopIteration` selectively, which requires us to copy the transcoder's [encodeHeaders](https://github.com/envoyproxy/envoy/blob/0fae6970ddaf93f024908ba304bbd2b34e997a51/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L594-L627) logic and its introduced dependency(like we need to init the method_). I feel that is very error-prone.

Can you push that prototype to some branch? I don't think the other filter need to involve that because the transcoder filter already did the job by merging the header and trailers into one, so the filter in front of them just see one `encodeHeader`.

TAOXUY:
> > @lizan IIUC not that simple, if we go with 2 stand-alone filters, there will be much code copy from transcoder for the not trailer only case(header and then followed by a grpc-status trailers), like this [test](https://github.com/envoyproxy/envoy/blob/92808ba1960f88016f65511376263ed39a1f8cc5/test/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter_test.cc#L1451-L1475). It needs the encodeHeaders to `StopIteration` selectively, which requires us to copy the transcoder's [encodeHeaders](https://github.com/envoyproxy/envoy/blob/0fae6970ddaf93f024908ba304bbd2b34e997a51/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L594-L627) logic and its introduced dependency(like we need to init the method_). I feel that is very error-prone.
> 
> Can you push that prototype to some branch? I don't think the other filter need to involve that because the transcoder filter already did the job by merging the header and trailers into one, so the filter in front of them just see one `encodeHeader`.

Ohh, I was trying to transcode the gRPC status protobuf->json directly before the transcoder and it could be feasible to modify the json payload in the filter after the transcoder.

We still think it makes sense to directly support custom transcoded error JSON format  by extension with google.rpc.Status as underlying wire format.  For example, some users may have HTTP backends with some HTTP error format and then migrate to gRPC backends by using the transcoder. They can use this extension for backwards compatibility.

TAOXUY:
> > For example, some users may have HTTP backends with some HTTP error format and then migrate to gRPC backends by using the transcoder. They can use this extension for backwards compatibility.
> 
> I don't take the theoretical use case as a reasoning behind this. We do see people use local_reply_mapper or implement their own filter (most cases in Lua) to do that. Adding a C++ extension point doesn't reduce their work to keep backward compatibility.

OK, I guess we have to workaround on our side. Anyway thanks for all the reviews!

TAOXUY:
> > For example, some users may have HTTP backends with some HTTP error format and then migrate to gRPC backends by using the transcoder. They can use this extension for backwards compatibility.
> 
> I don't take the theoretical use case as a reasoning behind this. We do see people use local_reply_mapper or implement their own filter (most cases in Lua) to do that. Adding a C++ extension point doesn't reduce their work to keep backward compatibility.

HI @lizan, just realized simply adding a filter in front of transcoder to overwrite payload won't work elegantly as gRPC status info has been removed in transcoder, which we need for our error format. To make that work, AFAIK, we need to change some filter after transcoder to record the gRPC status. This solution feels hacky and error-prone.

Could you please reconsider this PR? Adding an extension is definitely the most elegant solution IMHO. We can use flag to toggle this feature if you don't like the extension approach.

wrowe:
This seems to be stalled on the discussion between @TAOXUY and @lizan (and coverage failure, it's worth re-merging main)

/wait

PR #22457, opened 2022-07-28
http2: guarding against stream reset before stream creation

alyssawilk:
Right now if we call newStream and then reset the stream before sending headers, we attempt to have the codec send a reset stream to stream_id -1.  nghttp2 appears to silently swallow this without complaint. oghttp2 will soon as well but either way it doesn't make sense to reset the stream before a stream id has been assigned.

Risk Level: low
Testing: new unit test
Docs Changes: n/a
Release Notes: n/a


alyssawilk:
cc @birenroy thanks for helping me confirm this weirdness :-)

RyanTheOptimist (approved):
Nice catch!

RyanTheOptimist:
@wbpcode ping

PR #22460, opened 2022-07-29
clarify maglev docs

ramaraochavali:
Clarify Maglev docs that it aims for minimal disruption and point to table_size to control the behaviour.

Commit Message: clarify maglev docs
Additional Description:
Risk Level: N/A
Testing: N/A
Docs Changes: N/A
Release Notes: N/A
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/config/cluster/v3/cluster.proto
+++ b/api/envoy/config/cluster/v3/cluster.proto
@@ -462,8 +462,8 @@ message Cluster {
   // Specific configuration for the :ref:`Maglev<arch_overview_load_balancing_types_maglev>`
   // load balancing policy.
   message MaglevLbConfig {
-    // The table size for Maglev hashing. The Maglev aims for ‘minimal disruption’ rather than an absolute guarantee.
-    // Minimal disruption means that when the set of upstreams changes, a connection will likely be sent to the same
+    // The table size for Maglev hashing. Maglev aims for ‘minimal disruption’ rather than an absolute guarantee.
+    // Minimal disruption means that when the set of upstream hosts change, a connection will likely be sent to the same
     // upstream as it was before. Increasing the table size reduces the amount of disruption.
     // The table size must be prime number limited to 5000011. If it is not specified, the default is 65537.
     google.protobuf.UInt64Value table_size = 1 [(validate.rules).uint64 = {lte: 5000011}];
--- a/api/envoy/config/cluster/v3/cluster.proto
+++ b/api/envoy/config/cluster/v3/cluster.proto
@@ -462,7 +462,7 @@ message Cluster {
   // Specific configuration for the :ref:`Maglev<arch_overview_load_balancing_types_maglev>`
   // load balancing policy.
   message MaglevLbConfig {
-    // The table size for Maglev hashing. Maglev aims for ‘minimal disruption’ rather than an absolute guarantee.
+    // The table size for Maglev hashing. Maglev aims for "minimal disruption" rather than an absolute guarantee.
     // Minimal disruption means that when the set of upstream hosts change, a connection will likely be sent to the same
     // upstream as it was before. Increasing the table size reduces the amount of disruption.
     // The table size must be prime number limited to 5000011. If it is not specified, the default is 65537.
--- a/docs/root/intro/arch_overview/upstream/load_balancing/load_balancers.rst
+++ b/docs/root/intro/arch_overview/upstream/load_balancing/load_balancers.rst
@@ -141,8 +141,8 @@ table lookup build times as well as host selection times (approximately 10x and 
 when using a large ring size of 256K entries). While Maglev aims for minimal disruption, it is not
 as stable as ring hash when upstream hosts change. More keys will move position when hosts are removed
 (simulations show approximately double the keys will move). The amount of disruption can be minimized
- by increasing the :ref:`table_size<envoy_v3_api_field_config.cluster.v3.Cluster.MaglevLbConfig.table_size>`.
- With that said, for many applications
+by increasing the :ref:`table_size<envoy_v3_api_field_config.cluster.v3.Cluster.MaglevLbConfig.table_size>`.
+With that said, for many applications
 including Redis, Maglev is very likely a superior drop in replacement for ring hash. The advanced reader can use
 :repo:`this benchmark </test/common/upstream/load_balancer_benchmark.cc>` to compare ring hash
 versus Maglev with different parameters.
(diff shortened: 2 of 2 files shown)


docs/root/intro/arch_overview/upstream/load_balancing/load_balancers.rst
```diff
@@ -132,16 +132,18 @@ host B has a weight of 2, then host A will have 21,846 entries and host B will h
...
-are underrepresented or missing.
+regardless of weight.
 
 In general, when compared to the ring hash ("ketama") algorithm, Maglev has substantially faster
 table lookup build times as well as host selection times (approximately 10x and 5x respectively
 when using a large ring size of 256K entries). The downside of Maglev is that it is not as stable
 as ring hash. More keys will move position when hosts are removed (simulations show approximately
-double the keys will move). With that said, for many applications including Redis, Maglev is very
-likely a superior drop in replacement for ring hash. The advanced reader can use
+double the keys will move). Maglev aims for minimal disruption when set of upstream hosts change.
```
ramaraochavali, line 141: Main point to explain this with table_size config.  This is what https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-maglevlbconfig says.
Does this contradict with previous statement 
"The downside of Maglev is that it is not as stable
as ring hash. More keys will move position when hosts are removed (simulations show approximately
double the keys will move)" ?

Is there a better way to clarify it?

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -462,8 +462,8 @@ message Cluster {
   // Specific configuration for the :ref:`Maglev<arch_overview_load_balancing_types_maglev>`
   // load balancing policy.
   message MaglevLbConfig {
-    // The table size for Maglev hashing. The Maglev aims for ‘minimal disruption’ rather than an absolute guarantee.
-    // Minimal disruption means that when the set of upstreams changes, a connection will likely be sent to the same
+    // The table size for Maglev hashing. Maglev aims for ‘minimal disruption’ rather than an absolute guarantee.
```
phlax, line 465:
```suggestion
    // The table size for Maglev hashing. Maglev aims for "minimal disruption" rather than an absolute guarantee.
```

docs/root/intro/arch_overview/upstream/load_balancing/load_balancers.rst
```diff
@@ -132,16 +132,18 @@ host B has a weight of 2, then host A will have 21,846 entries and host B will h
...
+Best practice is to set
+:ref:`table_size<envoy_v3_api_field_config.cluster.v3.Cluster.MaglevLbConfig.table_size>` and monitor
+the :ref:`min_entries_per_host and max_entries_per_host gauges <config_cluster_manager_cluster_stats_maglev_lb>`
+to ensure no hosts are underrepresented or missing. With that said, for many applications
```
phlax, line 145:
```suggestion
to ensure no hosts are under-represented or missing. With that said, for many applications
```

seems this can be spelt both ways - but i think its more readable this way

htuch:
LGTM, @phlax feel free to merge when PR changes look good to you.

docs/root/intro/arch_overview/upstream/load_balancing/load_balancers.rst
ramaraochavali: @mattklein123 any thoughts?

mattklein123: I'm a little confused what we are trying to clarify here. From a very quick glance this is mostly just moving text around?

/wait-any

ramaraochavali: "Maglev aims for minimal disruption when set of upstream hosts change."  this part of the text is copied from Maglev config. Also linked the table_size from here. So you are right - this part is moving the text around.

My question is we have this statement 
" The downside of Maglev is that it is not as stable
as ring hash. More keys will move position when hosts are removed (simulations show approximately
double the keys will move)" in this section 

and 
""Maglev aims for minimal disruption when set of upstream hosts change."  is there is in MaglevLbConfig section.

So when folks reading this section think that MagLev does not handle upstream hosts change well. But the second statement seem to contradict that.

The goal of this PR is to clarify that part and add table_size reference here.

mattklein123: I think the original idea behind this text is to say something like "Maglev aims for minimal disruption when a set of upstream hosts change, but the disruption is generally greater than that achieved by Ketama/Ring Hash." I think the text does basically say this though, so feel free to clarify in a way that makes more sense to you?

/wait

ramaraochavali: Sure. Will change. Another thing I got confused with this is this https://docs.google.com/document/d/1IU4b76AgOXijNa4sew1gfBfSiOMbZNiEt5Dhis8QpYg/edit?disco=AAAAMzmbMm8.

So basically it looks like the MagLev disruption is more compared to Ring Hash.  I will reword as you suggested

ramaraochavali: @mattklein123 updated. PTAL if this change makes sense.

docs/root/intro/arch_overview/upstream/load_balancing/load_balancers.rst
```diff
@@ -132,16 +132,18 @@ host B has a weight of 2, then host A will have 21,846 entries and host B will h
 regardless of the configured host and locality weights, so in some extreme cases the actual
 proportions may differ from the configured weights. For example, if the total number of hosts is
 larger than the fixed table size, then some hosts will get 1 entry each and the rest will get 0,
-regardless of weight. Best practice is to monitor the :ref:`min_entries_per_host and
-max_entries_per_host gauges <config_cluster_manager_cluster_stats_maglev_lb>` to ensure no hosts
-are underrepresented or missing.
```
mattklein123, line 137: Sorry I don't think it makes sense to move this part, as this is related to ring population, not the section below on ring stability.

In general I'm really not sure this PR makes anything that much clearer but if after moving this part back you still want an edit below feel free.

/wait

ramaraochavali: I agree this does not makes sense to move that part below. Not sure why I moved that. I wanted to refer table_size config and let users set it.

Given that my original intention of clarifying whether Maglev performs  better or worse (that was the confusion I was having) when hosts change is not needed any more, I will close this.

ramaraochavali: @mattklein123 Sorry. Now I remember - I wanted to refer table_size here for the folks who wanted to reduce the amount of disruption using MagLev. Can you please see if this change makes sense? Sorry for the run around.

mattklein123 (approved):
LGTM, thanks!

PR #22463, opened 2022-07-29
Add validation of trailers to UHV

yanavlasov:
Also consolidate error codes in UHV interface.

Risk Level: Low, not in production build
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A


envoy/http/header_validator.h
```diff
@@ -81,9 +81,21 @@ class HeaderValidator {
...
    */
-  using ResponseHeaderMapValidationResult = RejectResult;
-  virtual ResponseHeaderMapValidationResult
-  validateResponseHeaderMap(ResponseHeaderMap& header_map) PURE;
+  virtual ValidationResult validateResponseHeaderMap(ResponseHeaderMap& header_map) PURE;
+
+  /**
+   * Validate the entire request trailer map.
+   * Returning the Reject value causes HTTP requests to be rejected with the 502 status,
+   * and gRPC requests with the the UNAVAILABLE (14) error code.
```
alyssawilk, line 89: should the other functions also include gRPC resposnes or is it trailers specific?

alyssawilk (commented):
one nit otherwise LGTM.
/wait

snowp (dismissed):
LGTM after other comments, thanks!

wrowe:
It's my understanding that this shouldn't be staled-out, correct @yanavlasov?

alyssawilk:
Yan what's the state on ths one?
/wait on CI if nothing else

yanavlasov:
> Yan what's the state on ths one? /wait on CI if nothing else

Still WiP. It needs some PRs to come in first that I'm working on.

yanavlasov:
Blocked by #24497

envoy/http/header_validator.h
yanavlasov: All other functions include gRPC responses if they are possible. Response trailers error can only result on reset.

yanavlasov:
Ready for review now. @wrowe @alyssawilk

envoy/http/header_validator.h
```diff
@@ -95,6 +95,21 @@ class HeaderValidator {
   using ResponseHeaderMapValidationResult = RejectResult;
   virtual ResponseHeaderMapValidationResult
   validateResponseHeaderMap(ResponseHeaderMap& header_map) PURE;
+
+  /**
+   * Validate the entire request trailer map.
+   * Returning the Reject value causes HTTP requests to be rejected with the 502 status,
```
alyssawilk, line 101: nit: should this be HTTP request to be rejected (no s) here and below?

source/common/http/conn_manager_impl.cc
```diff
@@ -909,6 +909,40 @@ bool ConnectionManagerImpl::ActiveStream::validateHeaders() {
   return true;
 }
 
+bool ConnectionManagerImpl::ActiveStream::validateTrailers() {
+  if (header_validator_) {
```
alyssawilk, line 913: if (!header_validator_) {
  return true
}
and unindent all below?

source/common/http/conn_manager_impl.cc
```diff
@@ -909,6 +909,40 @@ bool ConnectionManagerImpl::ActiveStream::validateHeaders() {
...
+    auto validation_result = header_validator_->validateRequestTrailerMap(*request_trailers_);
+    if (!validation_result.ok()) {
```
alyssawilk, line 915: ditto here: if ok() return true?

source/common/http/conn_manager_impl.cc
```diff
@@ -909,6 +909,40 @@ bool ConnectionManagerImpl::ActiveStream::validateHeaders() {
...
+      if (validation_result.details() == UhvResponseCodeDetail::get().InvalidUnderscore &&
+          connection_manager_.codec_->protocol() == Protocol::Http2) {
+        filter_manager_.streamInfo().setResponseCodeDetails(validation_result.details());
+        resetStream();
+      } else {
+        if (connection_manager_.codec_->protocol() < Protocol::Http2) {
+          sendLocalReply(response_code, "", nullptr, grpc_status, validation_result.details());
+        } else {
+          filter_manager_.streamInfo().setResponseCodeDetails(validation_result.details());
+          resetStream();
```
alyssawilk, line 934: why not send a local response for H2 in this case?

source/extensions/http/header_validators/envoy_default/header_validator.cc
```diff
@@ -434,6 +434,48 @@ bool HeaderValidator::hasChunkedTransferEncoding(const HeaderString& value) {
   return false;
 }
 
+// For both H/1 and H/2 trailers should only contain generic headers. As such a common validation
```
alyssawilk, line 437: Is this not true for H3 as well?

source/extensions/http/header_validators/envoy_default/header_validator.cc
```diff
@@ -434,6 +434,48 @@ bool HeaderValidator::hasChunkedTransferEncoding(const HeaderString& value) {
...
+// method can be used. For H/2, per https://www.rfc-editor.org/rfc/rfc9113#section-8.1 trailers MUST
```
alyssawilk, line 438: I'm missing where we validate pseudo headers?
Also if we don't remove pseudo headers for HTTP/1.1 will we somewhere where we're doing protocol downgrade?

alyssawilk (commented):
/wait

envoy/http/header_validator.h
yanavlasov: Done

source/common/http/conn_manager_impl.cc
yanavlasov: Done

source/common/http/conn_manager_impl.cc
yanavlasov: Done

source/common/http/conn_manager_impl.cc
yanavlasov: I have filed and Issue and left to TODO to harmonize the behavior. I will do it in a separate PR as it will affect a lot of tests, where the reset for H/2 and H/3 is expected.

source/extensions/http/header_validators/envoy_default/header_validator.cc
yanavlasov: Updated comment

source/extensions/http/header_validators/envoy_default/header_validator.cc
yanavlasov: We validate them in [H/1](https://github.com/envoyproxy/envoy/blob/bc3d03886cd4ca8b9677cbaba7976fc2840b9597/source/extensions/http/header_validators/envoy_default/http1_header_validator.cc#L53) and [H/2](https://github.com/envoyproxy/envoy/blob/bc3d03886cd4ca8b9677cbaba7976fc2840b9597/source/extensions/http/header_validators/envoy_default/http2_header_validator.cc#L50) specific validators. Although the function is the same in both cases and should be refactored (I'll file an Issue for this). The `validateGenericHeaderName` fails if there is ':' in the name.

For H/1 we never remove pseudoheaders, since the H/1 parser can not produce them. We only add them for the request and response using the first line from the wire. And we never add them for trailers.

alyssawilk: ah, the comment reads like pseudo headers are allowed for HTTP/1.1   maybe make that clear?

envoy/http/header_validator.h
```diff
@@ -77,23 +77,38 @@ class HeaderValidator {
   /**
    * Validate the entire request header map.
    * This method may mutate the header map as well, for example by normalizing URI path.
-   * Returning the Reject value form this method causes HTTP requests to be rejected with 400
-   * status, and gRPC requests with the the INTERNAL (13) error code. Returning the Redirect value
-   * causes HTTP requests to be redirected to the :path presudo header in the request map. gRPC
-   * requests will still be rejected with the INTERNAL (13) error code.
+   * Returning the Reject value form this method causes HTTP request to be rejected with 400
```
alyssawilk, line 80: sorry my correction needs a correction :-(
I was thinking "the HTTP request"
"http requests" was also fine

source/common/http/conn_manager_impl.h
```diff
@@ -354,6 +354,7 @@ class ConnectionManagerImpl : Logger::Loggable<Logger::Id::http>,
     // Call header validator extension to validate request header map after it was deserialized.
     // If header map failed validation, it sends an error response and returns false.
     bool validateHeaders();
+    bool validateTrailers();
```
alyssawilk, line 357: comment?

source/extensions/http/header_validators/envoy_default/header_validator.cc
```diff
@@ -434,6 +434,49 @@ bool HeaderValidator::hasChunkedTransferEncoding(const HeaderString& value) {
   return false;
 }
 
+// For all (H/1, H/2 and H/3) protocols, trailers should only contain generic headers. As such a
+// common validation method can be used. For H/2, per
+// https://www.rfc-editor.org/rfc/rfc9113#section-8.1 trailers MUST NOT contain H/2 pseudo header
+// fields. For H/1 per https://www.rfc-editor.org/rfc/rfc9110#section-6.5 there are no specific
+// prohibitions.
```
alyssawilk, line 441: for H2 and H3 no pseudo headerrs.  if we disallow for H1 as well let's say there's no prohibitions but we disallow anyway?

alyssawilk (commented):
LG modulo a few remaining nits

envoy/http/header_validator.h
yanavlasov: Changed to "the HTTP request".

source/common/http/conn_manager_impl.h
yanavlasov: Done

source/extensions/http/header_validators/envoy_default/header_validator.cc
yanavlasov: Yes, the comment is confusing. I've tried to clarify it, see if it is better. I'd like to provide more details from RFCs on why the common function would work.

source/extensions/http/header_validators/envoy_default/header_validator.cc
yanavlasov: Clarified comment. I can also remove the more detailed explanation and just link to RFCs.

alyssawilk (approved):

PR #22469, opened 2022-07-29
wasm: update proxy-wasm to apply canary always

ingwonsong:
Signed-off-by: Ingwon Song <igsong@google.com>

This PR update "proxy-wasm" to include https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/289.
After https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/289, proxy-wasm will always apply the canary.

Due to this change, the second creation of Wasm filter with invalid config will raise the exception as the first trial.
So, remove the second creation from test cases.

In addition, to deal with the coverage, test cases are added. 

Please note that the comments in `wasm_filter.h` are moved to right after of the `return` statements because `lcov` weirdly detects the comment lines as uncovered statements.


test/extensions/access_loggers/wasm/config_test.cc
```diff
@@ -180,8 +180,11 @@ TEST_P(WasmAccessLogConfigTest, YamlLoadFromFileWasmInvalidConfig) {
   filter_instance->log(nullptr, nullptr, nullptr, log_stream_info);
 
   TestUtility::loadFromYaml(invalid_yaml, proto_config);
-  filter_instance = factory->createAccessLogInstance(proto_config, nullptr, context_);
-  filter_instance->log(nullptr, nullptr, nullptr, log_stream_info);
+  // When trying again with invalid config, it should raise the exception again because the canary
+  // will fail.
+  EXPECT_THROW_WITH_MESSAGE(factory->createAccessLogInstance(proto_config, nullptr, context_),
+                            Envoy::Extensions::Common::Wasm::WasmException,
+                            "Unable to create Wasm access log ");
```
PiotrSikora, line 187: Note that we already verify that this exception is thrown earlier in this test (there are: `invalid_yaml` and `valid_yaml` cases), so perhaps this should be completely removed instead.

(Same in the other 2 tests)

ingwonsong: OK. it makes sense.
Will change it.

ingwonsong:
FYI, as I mentioned above, `lcov` detects the comment lines as a regular line, so that the coverage is mis-calculated.

image

bazel/repository_locations.bzl
```diff
@@ -1119,8 +1119,8 @@ REPOSITORY_LOCATIONS_SPEC = dict(
         project_name = "WebAssembly for Proxies (C++ host implementation)",
         project_desc = "WebAssembly for Proxies (C++ host implementation)",
         project_url = "https://github.com/proxy-wasm/proxy-wasm-cpp-host",
-        version = "2cf68d6c00b4141ecf2afb5e5d51ebd21a27ed43",
-        sha256 = "6314d3e36e8aa9c4e792772b66f773fcfd58affcd4fd437c6d3dc540165e7fc6",
+        version = "4ddbed3c8c8c1aa76db8157e22ae3be676e76ac1",
+        sha256 = "e9c05ed5e827e256f4725735c2ed3d5420fc95f5bce052b42f93aa37248a1f42",
```
PiotrSikora, line 1123: You need to update the `release_date` below. I'm not sure why the format check didn't catch it.

source/extensions/filters/http/wasm/wasm_filter.h
```diff
@@ -40,11 +40,11 @@ class FilterConfig : Logger::Loggable<Logger::Id::wasm> {
     }
     if (!wasm || wasm->isFailed()) {
       if (handle->plugin()->fail_open_) {
-        // Fail open skips adding this filter to callbacks.
-        return nullptr;
+        return nullptr; // Fail open skips adding this filter to callbacks.
+
```
PiotrSikora, line 44: Nit: unnecessary empty line.

PiotrSikora (dismissed):
LGTM, sans nit. Thanks!

source/extensions/filters/http/wasm/wasm_filter.h
ingwonsong: Thanks! Interesting part is that this empty line is also counted as a uncovered line of codes.

bazel/repository_locations.bzl
ingwonsong: Done.

PiotrSikora (approved):
Thanks!

mattklein123 (approved):

PR #22470, opened 2022-07-29
Make filter fuzz tester support filters that process off-thread

ravenblackx:
Commit Message: Make filter fuzz tester support filters that process off-thread
Additional Description: This is an alternative approach to #21757  
This option also makes the tester verify that filters complete their task - previously a filter could go into a "waiting" mode and stay there, and still pass the fuzz test.
Risk Level: Low (test-only)
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
Fixes Issue: #21726

### DIFF
--- a/test/extensions/filters/http/common/fuzz/uber_filter.cc
+++ b/test/extensions/filters/http/common/fuzz/uber_filter.cc
@@ -9,7 +9,6 @@
 #include "source/common/protobuf/utility.h"
 
 #include "test/test_common/utility.h"
-#include <chrono>
 
 namespace Envoy {
 namespace Extensions {
--- a/test/extensions/filters/http/common/fuzz/uber_filter.cc
+++ b/test/extensions/filters/http/common/fuzz/uber_filter.cc
@@ -1,6 +1,8 @@
 #include "test/extensions/filters/http/common/fuzz/uber_filter.h"
 
+#include "source/common/common/thread_impl.h"
 #include "source/common/config/utility.h"
+#include "source/common/event/dispatcher_impl.h"
 #include "source/common/http/message_impl.h"
 #include "source/common/http/utility.h"
 #include "source/common/protobuf/protobuf.h"
@@ -13,7 +15,11 @@ namespace Extensions {
 namespace HttpFilters {
 
 UberFilterFuzzer::UberFilterFuzzer()
-    : async_request_{&cluster_manager_.thread_local_cluster_.async_client_} {
+    : async_request_{&cluster_manager_.thread_local_cluster_.async_client_},
+      thread_factory_(Thread::threadFactoryForTest()) {
+  ON_CALL(api_, threadFactory()).WillByDefault(testing::ReturnRef(thread_factory_));
+  worker_thread_dispatcher_ =
+      std::make_unique<Event::DispatcherImpl>("filter_fuzz_test", api_, api_.time_system_);
   // This is a decoder filter.
   ON_CALL(filter_callback_, addStreamDecoderFilter(_))
       .WillByDefault(Invoke([&](Http::StreamDecoderFilterSharedPtr filter) -> void {
@@ -77,12 +83,26 @@ void UberFilterFuzzer::fuzz(
   // Data path should not throw exceptions.
   if (decoder_filter_ != nullptr) {
     HttpFilterFuzzer::runData(decoder_filter_.get(), downstream_data);
+  } else {
+    decoding_finished_ = true;
   }
   if (encoder_filter_ != nullptr) {
     HttpFilterFuzzer::runData(encoder_filter_.get(), upstream_data);
+  } else {
+    encoding_finished_ = true;
   }
   if (access_logger_ != nullptr) {
     HttpFilterFuzzer::accessLog(access_logger_.get(), stream_info_);
+  }
+
+  // Most filters should have finished processing during runData, but filters that
+  // rely on an additional thread (e.g. for file system interaction) may need to wait
+  // for the worker thread to complete the filter's task.
+  while (!isFilterFinished()) {
+    auto end_timer = worker_thread_dispatcher_->createTimer(
+        []() { throw EnvoyException("filter did not finish processing within timeout"); });
+    end_timer->enableTimer(std::chrono::milliseconds(5000));
+    worker_thread_dispatcher_->run(Event::DispatcherImpl::RunType::Block);
   }
 
   reset();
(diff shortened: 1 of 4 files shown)


test/extensions/filters/http/common/fuzz/uber_filter.cc
```diff
@@ -77,14 +83,34 @@ void UberFilterFuzzer::fuzz(
...
+    auto end_timer = worker_thread_dispatcher_->createTimer(
+        []() { throw EnvoyException("filter did not finish processing within timeout"); });
+    // This timer doesn't work because, from the log,
+    // "Simulated timer enabled. Use advanceTimeWait or advanceTimeAsync functions to ensure it is
+    // called."
+    end_timer->enableTimer(std::chrono::milliseconds(5000));
+    worker_thread_dispatcher_->run(Event::DispatcherImpl::RunType::Block);
+    // Apparently RunType::Block doesn't actually block if there's no events ready to
+    // fire, so this loop spins. Yield to keep it from spinning too wildly.
+    std::this_thread::yield();
```
adisuissa, line 111: This seems somewhat broken to me.
I wonder if we can use some synchronization between the additional threads and this one. Do you think the [ThreadSynchronizer](https://github.com/envoyproxy/envoy/blob/main/source/common/common/thread_synchronizer.h#L20) be useful in this case?

adisuissa (commented):
Thanks for tackling this. Left a high-level question on the approach.

test/extensions/filters/http/common/fuzz/uber_filter.cc
ravenblackx: Oops, yes, it was broken, I left the comment in there hoping for a good suggestion (it's surprising that RunType::Block doesn't block at all if there's a future timer event, not sure if it blocks if there's no event).

I don't think ThreadSynchronizer can be useful for this because the goal here is:
1. block until there's an event and do that event (the normal flow)
2. fail the test if there is no event for some amount of time

ThreadSynchronizer would potentially (with changes to non-test code) allow us to block until there's an event, but without a timeout, so path 2 would become "the 300 second test timeout with no useful output".

I've updated this to a spinning loop that *does* eventually fail, vs. the previous spinning loop that never failed because the timer event would never be triggered due to both nonblocking *and* fake-time-not-passing.

Another option that would work but would essentially amount to the same thing would be to set a timeout event from outside the loop, and advance fake time by some amount each time the loop cycles - this would still just amount to a fixed number of passes through the loop, so we might as well just do that directly.

I don't think, in the absence of an actual blocking 'run' call and a functioning real-time timeout, there's a solution to this that doesn't involve some amount of spin-waiting. And we can't even reduce the spin with sleep, because sleep is also forbidden. (But we probably also *don't want* to spin any slower than yield anyway, since ending the test quickly is more important than a brief, rare processor spike while testing.)

(NB. we could actually do a real-time deadline using absl::Mutex's `Await`, but this would involve either changes to ThreadSynchronizer or creating another thread just to wait for ThreadSynchronizer and touch a condition variable, *and* modifying non-test code. Seems less clean than even a spin-loop. And probably using that real-time deadline implementation is violating the spirit of the linter's "don't use real clock" by just finding a loophole it hasn't covered!)

alyssawilk:
@adisuissa ping!

test/extensions/filters/http/common/fuzz/uber_filter.cc
adisuissa: IIUC what you are describing here is that we'd want to use a barrier, to ensure all the threads reach some point, before cleanly exiting.

Looking at [`AsyncFileManagerThreadPool`](https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/async_files/async_file_manager_thread_pool.cc#L78) I think that If there's a way to enqueue a "Terminate" action to all the internal worker threads, that should probably fix this.

ravenblackx: I don't understand how that could help.

Enqueuing a terminate action would almost certainly do the wrong thing, because... well, in the context of this test, there are essentially two threads that pass pieces of work back and forth between them, one file manager thread and one dispatcher thread.
If we assume one thread with actions completes its actions for each pass around the activity loop, what happens here is
1. Before the loop, dispatcher thread passes some work to file thread.
2. File thread does some work, passes work back to dispatcher thread.
3. Dispatcher thread does some work, passes work back to file thread.
4. File thread does some work, passes work back to dispatcher thread.
If we enqueued a terminate action at step 2, for example, then step 4 wouldn't work because the thread has been terminated, no?

The problem is that there isn't any "some point" at which we know the threads are done with their work, other than the condition used to exit the loop.

Maybe adding a timeout option to barrierOn is the answer? Then thread_synchronizer could be used without the "just timing out the test" problem.

ravenblackx: After looking into that option, it doesn't look great either - the problem there is ThreadSynchronizer has been implemented to be target-specific rather than sharing a global/static instance (even though its named events would have made a singleton/static design more sensible), which would mean the test needs to know *a specific* filter to wait for, and part of the point of the fuzz test is that it doesn't even know what filter it's testing.

We could potentially instantiate e.g. a FilterWorkerThreadSynchronizerSingleton through the singleton factory, so the threads and test could all share an instance. The thing here is I don't want to make this loop specifically wait for AsyncFileManager threads, I'd want it to wait for any worker threads. Even though AsyncFileManager is currently the only source of external worker threads, it seems likely that e.g. a database-based cache implementation would want to do work in other threads too, and would still want to be supported for fuzz testing.

ravenblackx: Anything like that seems like a lot of work (and also making the live code have more complexity in it) just to avoid 5 seconds of gentle spin-looping in the rare failure case of a test. (Which would still take 5 seconds to fail anyway, even using ThreadSynchronizer modified to support a timeout.)

ravenblackx: (Sent #22607 since that much at least seems like a useful ability even if we don't want to use it for this.)

adisuissa:
> @adisuissa ping again. I would contend that having the fuzz test work arguably-non-optimally is better than leaving the fuzz test unable to cover any filter that processes out-of-band. (If someone has a better idea of how to make it work _and_ not spin-loop without adding complexity to the production code, it's not like landing this as-is would prevent improving it later!)

I think there could be a bigger underlying issue, where a worker-thread will spawn some utility threads and may terminate before the utility threads will finish their job. I'm assuming that the implementation handles this by only terminating the utility threads after all the work has been done, but it might be some fragile assumption.

Other than that, the fix you suggest looks good to me.
I'll defer to @mattklein123 as the other code-owner of this extension.

mattklein123 (approved):
Seems fine to ship and iterate. We can improve as needed. Thank you!

PR #22472, opened 2022-07-30
Add service.name to OpenTelemetry tracer

### LINKED ISSUE #22183: OpenTelemetry extension should set service.name resource attribute
*Title*: *Set the OpenTelemetry service.name resource attribute*

*Description*:
OpenTelemetry specifies semantic conventions for services to meet, [the most basic being the service's name](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service). 

*Repro steps*:
First reported here: https://github.com/envoyproxy/envoy/pull/21370#issuecomment-1165023985

The Envoy should emit well-formed OTLP span data. The OpenTelemetry conventions require a service.name resource attribute, so one should be set either using configuration or a default value. OpenTelemetry specifies how to pick a default value, for example:

> [1]: MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fallback to unknown_service: concatenated with [process.executable.name](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/process/#process), e.g. unknown_service:bash. If process.executable.name is not available, the value MUST be set to unknown_service.

AlexanderEllis:
Commit Message: Add service.name to OpenTelemetry tracer
Additional Description: This is a required field for OTLP trace span resources. This pull requests adds an optional service name to the OpenTelemetry tracer config proto. When the tracer is initialized with that proto, the name is saved and used as the `service.name` attribute in the ResourceSpan's Resource. If a service name is not provided, this defaults to `unknown_service:envoy` (see [spec](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service)). 

This `service.name` attribute is considered [required](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#semantic-attributes-with-sdk-provided-default-value), so we should include it here.

Risk Level: Low
Testing: Unit tests
Fixes #22183


moderation:
I've tested this PR on Linux x64 and aarch64 and MacOS aarch64 and it looks good!

mattklein123 (approved):
Thanks!

Issue #22303, opened 2022-07-19
Add support for an optional TTL in KeyValueStore


abeyad:
It would be nice to have the [KeyValueStore](https://github.com/envoyproxy/envoy/blob/main/envoy/common/key_value_store.h) support a TTL (time-to-live) for entries.  If a TTL is set, then when the TTL has elapsed, the entry would be removed from the KV store, and if it's a KV store that persists to disk (e.g. [FileBasedKeyValueStore](https://github.com/envoyproxy/envoy/blob/main/source/extensions/key_value/file_based/config.h)), then the entry would be deleted from disk as well.

One use case is for storing xDS resources and evicting them when the TTL has expired.

We can add support for a TTL in a couple ways: 
1. As a first class construct within the KV store, similar to how KV store's support a [flush interval](https://github.com/envoyproxy/envoy/blob/main/source/common/common/key_value_store_base.h#L19) and timer.
2. Leave it up to each KV store implementation to implement TTL support as it sees fit.

Interested in hearing others' thoughts on this.

cc @alyssawilk @goaway @jpsim @RyanTheOptimist

RyanTheOptimist:
Seems like a great idea. Alt-Svc has an explicit TTL so if the KV store itself implement it, that would be simpler.

alyssawilk:
SGTM

abeyad:
This was implemented in https://github.com/envoyproxy/envoy/pull/22884, so I'm closing this issue.

Issue #21910, opened 2022-06-27
Per-XDS type callbacks


paul-r-gall:
*Title*: Add per-xds callback extensions that can run arbitrary code after an XDS response has been accepted.

*Description*:
Currently, Envoy contains `ClusterUpdateCallbacks` which allows users to supply extensions that are run when clusters are added or removed. Similar functionality would be very useful for other XDS resources. This could be implemented either as a general XDS update hook, or on a per-resource basis.

wbpcode:
cc @mattklein123

wbpcode:
It would be great if we can make it some things like resource monitors. If there are some new config are accepted or rejected, we can post a event to the monitors. And the monitor then can do some extendable actions based on the type of event/xds.

wu-sheng:
Would like to see the progress in this direction. An xds-oriented callback hook seems a great idea.

kyessenov:
An interesting use-case is "xDS redirect" where a resource name can be translated between two xDS authority servers. That simplifies federation if their use inconsistent naming schemes.

PR #22473, opened 2022-08-01
xds: Adds extension points to receive config update notifications

abeyad:

This is an experimental feature and the API will not be published at this time.

Risk Level:
Testing:
Docs Changes:
Release Notes:

### DIFF
--- a/api/envoy/config/bootstrap/v3/bootstrap.proto
+++ b/api/envoy/config/bootstrap/v3/bootstrap.proto
@@ -41,7 +41,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // <config_overview_bootstrap>` for more detail.
 
 // Bootstrap :ref:`configuration overview <config_overview_bootstrap>`.
-// [#next-free-field: 35]
+// [#next-free-field: 36]
 message Bootstrap {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.config.bootstrap.v2.Bootstrap";
@@ -335,6 +335,11 @@ message Bootstrap {
   // If the value is not specified, Google RE2 will be used by default.
   // [#extension-category: envoy.regex_engines]
   core.v3.TypedExtensionConfig default_regex_engine = 34;
+
+  // TODO(abeyad): comment this
+  // [#extension-category: envoy.xds]
+  // [#not-implemented-hide:]
+  core.v3.TypedExtensionConfig persistent_xds_extension = 35;
 }
 
 // Administration interface :ref:`operations documentation
--- a/source/common/config/grpc_mux_impl.cc
+++ b/source/common/config/grpc_mux_impl.cc
@@ -37,16 +37,20 @@ GrpcMuxImpl::GrpcMuxImpl(const LocalInfo::LocalInfo& local_info,
                          const Protobuf::MethodDescriptor& service_method,
                          Random::RandomGenerator& random, Stats::Scope& scope,
                          const RateLimitSettings& rate_limit_settings, bool skip_subsequent_node,
-                         CustomConfigValidatorsPtr&& config_validators)
+                         CustomConfigValidatorsPtr&& config_validators,
+                         ConfigUpdatedCallbackList&& config_updated_callbacks,
+                         KeyValueStore* xds_config_store, const std::string& target_control_plane)
     : grpc_stream_(this, std::move(async_client), service_method, random, dispatcher, scope,
                    rate_limit_settings),
       local_info_(local_info), skip_subsequent_node_(skip_subsequent_node),
-      config_validators_(std::move(config_validators)), first_stream_request_(true),
+      config_validators_(std::move(config_validators)),
+      config_updated_callbacks_(std::move(config_updated_callbacks)) first_stream_request_(true),
       dispatcher_(dispatcher),
       dynamic_update_callback_handle_(local_info.contextProvider().addDynamicContextUpdateCallback(
           [this](absl::string_view resource_type_url) {
             onDynamicContextUpdate(resource_type_url);
-          })) {
+          })),
+      xds_config_store_(xds_config_store), target_control_plane_(target_control_plane) {
   Config::Utility::checkLocalInfo("ads", local_info);
   AllMuxes::get().insert(this);
 }
@@ -272,6 +276,13 @@ void GrpcMuxImpl::onDiscoveryResponse(
         watch->callbacks_.onConfigUpdate(found_resources, message->version_info());
       }
     }
+
+    // All config updates have been applied without throwing an exception, so we'll call the update
+    // callbacks.
+    for (const auto& callback : config_updated_callbacks_) {
+      callback->onConfigUpdated(target_control_plane_, type_url, all_resource_refs);
+    }
+
     // TODO(mattklein123): In the future if we start tracking per-resource versions, we
     // would do that tracking here.
     api_state.request_.set_version_info(message->version_info());
--- a/source/common/config/grpc_mux_impl.h
+++ b/source/common/config/grpc_mux_impl.h
@@ -4,8 +4,10 @@
 #include <memory>
 #include <queue>
 
+#include "envoy/common/key_value_store.h"
 #include "envoy/common/random_generator.h"
 #include "envoy/common/time.h"
+#include "envoy/config/config_updated_callback.h"
 #include "envoy/config/grpc_mux.h"
 #include "envoy/config/subscription.h"
 #include "envoy/event/dispatcher.h"
@@ -37,7 +39,9 @@ public:
               Event::Dispatcher& dispatcher, const Protobuf::MethodDescriptor& service_method,
               Random::RandomGenerator& random, Stats::Scope& scope,
               const RateLimitSettings& rate_limit_settings, bool skip_subsequent_node,
-              CustomConfigValidatorsPtr&& config_validators);
+              CustomConfigValidatorsPtr&& config_validators,
+              ConfigUpdatedCallbackList&& config_updated_callbacks, KeyValueStore* xds_config_store,
+              const std::string& target_control_plane);
 
   ~GrpcMuxImpl() override;
 
@@ -171,6 +175,7 @@ private:
   const LocalInfo::LocalInfo& local_info_;
   const bool skip_subsequent_node_;
   CustomConfigValidatorsPtr config_validators_;
+  ConfigUpdatedCallbackList config_updated_callbacks_;
   bool first_stream_request_;
 
   // Helper function for looking up and potentially allocating a new ApiState.
@@ -189,6 +194,10 @@ private:
 
   Event::Dispatcher& dispatcher_;
   Common::CallbackHandlePtr dynamic_update_callback_handle_;
+
+  // Not-owned; can be nullptr.
+  KeyValueStore* xds_config_store_;
+  const std::string target_control_plane_;
 
   // True iff Envoy is shutting down; no messages should be sent on the `grpc_stream_` when this is
   // true because it may contain dangling pointers.
--- a/api/envoy/extensions/config/xds/v3/BUILD
+++ b/api/envoy/extensions/config/xds/v3/BUILD
@@ -1,0 +1,11 @@
+load("@envoy_api//bazel:api_build_system.bzl", "api_proto_package")
+
+licenses(["notice"])  # Apache 2
+
+api_proto_package(
+    deps = [
+        "//envoy/config/common/key_value/v3:pkg",
+        "//envoy/config/core/v3:pkg",
+        "@com_github_cncf_udpa//udpa/annotations:pkg"
+    ],
+)
--- a/api/envoy/extensions/config/xds/v3/persistent_xds_extension_config.proto
+++ b/api/envoy/extensions/config/xds/v3/persistent_xds_extension_config.proto
@@ -1,0 +1,24 @@
+syntax = "proto3";
+
+package envoy.extensions.config.xds.v3;
+
+import "envoy/config/common/key_value/v3/config.proto"
+import "envoy/config/core/v3/extension.proto";
+
+import "udpa/annotations/status.proto";
+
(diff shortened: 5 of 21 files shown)


source/common/config/grpc_mux_impl.h
```diff
@@ -171,6 +174,9 @@ class GrpcMuxImpl : public GrpcMux,
   const LocalInfo::LocalInfo& local_info_;
   const bool skip_subsequent_node_;
   CustomConfigValidatorsPtr config_validators_;
+  // Not owned; can be nullptr.
+  XdsResourcesDelegate* xds_resources_delegate_;
```
alyssawilk, line 178: Generally Envoy leans towards OptRef for optional references, rather than pointers.

test/integration/xds_delegate_extension_integration_test.cc
```diff
@@ -0,0 +1,236 @@
...
+  std::string name() const override { return "envoy.config.xds.test_delegate"; };
+
+  Config::XdsResourcesDelegatePtr createXdsResourcesDelegate(const ProtobufWkt::Any&,
+                                                             ProtobufMessage::ValidationVisitor&,
+                                                             Api::Api&) override {
+    return std::make_unique<TestXdsResourcesDelegate>();
+  }
+};
+
+std::string bootstrapConfig() {
```
alyssawilk, line 59: is this mostly cargo culted from somewhere else and if so can we snag it from there rather than copying it?

test/integration/xds_delegate_extension_integration_test.cc
```diff
@@ -0,0 +1,236 @@
...
+    Json::ObjectSharedPtr loader = TestEnvironment::jsonLoadFromString(response->body());
+    auto entries = loader->getObject("entries");
+    if (entries->hasObject(key)) {
+      return entries->getObject(key)->getString("final_value");
+    }
+    return "";
+  }
+
+  void waitforOnConfigUpdatedCount(const int expected_count) {
+    TestUtility::waitForCondition(
```
alyssawilk, line 179: would time_system.waitFor() work here?

alyssawilk (commented):
Awesome!  This largely looks good to me, but I'm going to defer to Adi on general structure

source/common/config/grpc_mux_impl.h
abeyad: Switched to OptRef

test/integration/xds_delegate_extension_integration_test.cc
abeyad: Awesome didn't know about it! Switched to using that.

test/integration/xds_delegate_extension_integration_test.cc
abeyad: So, I did lift most of this from the rtds_integration_test's bootstrap config, but I'm wondering what's the best way to share this config, if any.  My test isn't really related to rtds_integration_test, other than using RTDS for config made setting the test's bootstrap easier, so I don't think my test should depend on rtds_integration_test. Is there a good place to put "similar config strings" like this for the test infra?  Should I add a new file/library to https://github.com/envoyproxy/envoy/blob/main/test/test_common ?

envoy/config/xds_resources_delegate.h
```diff
@@ -0,0 +1,78 @@
...
+   * @param authority_id Unique id for the authority/control plane from which the resources are
+   *   obtained. Typically either a cluster name or a host name.
+   * @param resource_type_url The URL for the type of the resource (e.g. Secrets, Clusters, etc).
+   * @return A set of xDS resources for the given authority and type.
+   */
+  virtual std::vector<envoy::service::discovery::v3::Resource>
+  getResources(const std::string& authority_id, const std::string& resource_type_url) PURE;
+
+  /**
+   * Invoked when SotW xDS configuration updates have been received from an xDS authority, have been
```
adisuissa, line 39: I'm trying to think whether what we want here is a mechanism that's called on each update, or just a way to serialize the state (config).
On the one-hand I can see the simplicity in serializing the current state and loading it when the Envoy restarts (albeit some serialization work will be required). On the other-hand saving the config makes this more resilient to binary upgrades and makes saving deltas possible.
WDYT?

adisuissa (commented):
Thanks for working on this. Left a few high-level comments.

envoy/config/xds_resources_delegate.h
abeyad: Adi & I chatted a bit about this.  A problem with the way Envoy handles xDS now is that there is no unified notion of the state of the dynamic config that can be serialized and re-loaded.  We receive xDS resources from a control plane and play those updates to the various consumers (e.g. SdsApi, CdsApi, etc).  This makes it harder to handle the Delta protocol with this mechanism, though it works reasonably well for SotW.  Since our initial intention with this extension point was to support Envoy Mobile use cases, we feel okay with limiting the feature to SotW for now.

Would like to know more senior maintainers thoughts on this.  /cc @alyssawilk

abeyad: Yes, the plan is to keep the extension point API hidden for the foreseeable future, so we should hopefully not disrupt users if we need to change things when the unified config xDS client in Envoy becomes a reality.

/cc @lizan for another API maintainer on their thoughts.

paul-r-gall:
https://github.com/envoyproxy/envoy/issues/21910 is also relevant here I think.

abeyad:
> #21910 is also relevant here I think.

Good point, I referenced it in the change description.

abeyad:
@lizan friendly ping for your review of this PR, thanks!

lizan (approved):

api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -335,6 +335,11 @@ message Bootstrap {
   // If the value is not specified, Google RE2 will be used by default.
   // [#extension-category: envoy.regex_engines]
   core.v3.TypedExtensionConfig default_regex_engine = 34;
+
+  // Optional XdsResourcesDelegate configuration.
+  // If a value is not specified, no XdsResourcesDelegate will be used.
```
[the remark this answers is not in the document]
abeyad: I explained the purpose of XdsResourcesDelegate in the XdsResourcesDelegate header file.  Would you like a comment here as well?

envoy/config/xds_resources_delegate.h
abeyad: Lizan approved the PR, so will leave it to Adi for final comments

test/integration/xds_delegate_extension_integration_test.cc
abeyad: I realized what you meant with using ConfigHelper, so I updated the test to use ConfigHelper + addConfigModifier

envoy/config/xds_resources_delegate.h
```diff
@@ -0,0 +1,83 @@
...
+   * applied on the Envoy instance, and are about to be ACK'ed.
+   *
+   * @param authority_id Unique id for the authority/control plane from which the resources are
+   *   obtained. Typically either a cluster name or a host name.
+   * @param resource_type_url The URL for the type of the resource (e.g. Secrets, Clusters, etc).
+   * @param resources The resources sent by the authority for the given type.
+   */
+  virtual void onConfigUpdated(const std::string& authority_id,
```
adisuissa, line 52: I think that `authority_id` should not be used here.
Eventually the authority ID will be part of the resource name. At the moment it seems you want to have a single place where a SotW message is confirmed to be the right one, so to me it is unclear when will the authority-id will be useful? Moreover, using the cluster/host name seems to not capture having multiple resources from a single source (that has the same cluster/host name).

envoy/config/xds_resources_delegate.h
```diff
@@ -0,0 +1,83 @@
...
+   * implementation to return a set of resources to be loaded and used by the Envoy instance
+   * (e.g. persisted resources in local storage).
+   *
+   * @param authority_id Unique id for the authority/control plane from which the resources are
+   *   obtained. Typically either a cluster name or a host name.
+   * @param resource_type_url The URL for the type of the resource (e.g. Secrets, Clusters, etc).
+   * @return A set of xDS resources for the given authority and type.
+   */
+  virtual std::vector<envoy::service::discovery::v3::Resource>
+  getResources(const std::string& authority_id, const std::string& resource_type_url) PURE;
```
adisuissa, line 41: nit: consider using `absl::string_view` over `const std::string&`

envoy/config/xds_resources_delegate.h
abeyad: I've been thinking about it some more and I am not sure how we can do away with authority_id for SotW.  First as some background, I ran an integration test that delivers the same resource type and resource name from two different xDS sources and the Envoy processing code for xDS seemed to have no problem with that.

Ultimately, I think we need the authority (or some id for the ApiConfigSource, maybe a hash of the ApiConfigSource?) for two reasons:
(1) because of the possibility, however unlikely, that the same resource type and resource name could get served from two different xDS sources
(2) More importantly, because we have no other way to keep track of what the SotW is for a given resource type from a particular xDS source, so that we can overwrite it with a new list of resources when we receive an update.  This is the case for both named and wildcard subscriptions.  There's no way to know if we should append or remove and add the received resources on a DiscoveryResponse.  Even looping through the resource names to check each time we get a DiscoveryResponse may not be sufficient in the unlikely case of #1.

So I think we still need some notion of the authority/ApiConfigSource, but I think it would be better abstracted into a class, so instead of passing in authority_id + resource_type_url to the functions in this class, we take in an abstract `XdsSourceId`, of which there could be different implementations.

WDYT?

abeyad:
@adisuissa I pushed a commit using a new `XdsSourceId` class that encapsulates the resource id state.  PTAL.

envoy/config/xds_resources_delegate.h
abeyad: @adisuissa @alyssawilk 
I pushed up a commit that adds a `reset()` function to the `XdsResourcesDelegate`, the purpose of which is to clear the existing resources in the xDS store when there is a change in the resource subscription, before adding the updated resources to the store.  This will enable us to handle updates to resources (update, not clear + add), while also clearing out resources that are no longer subscribed to.

We still need the authority id because as the tests in https://github.com/envoyproxy/envoy/pull/22823 show, it's possible to have two resources of the same type and name coming from two separate xDS servers.  We will need to be able to distinguish the resources to be added/updated/deleted in that situation.

abeyad: I changed this again to remove the `reset()` function, based on Adi astutely pointing out that we could have two or more GrpcMuxImpls for the same authority + type (different ApiConfigSources).  Instead, I think it makes sense to just receive the config updates, and allow the TTL'ing mechanism to handle eviction of stale resources (see https://github.com/envoyproxy/envoy/issues/22303 for details).

source/common/config/xds_source_id.cc
```diff
@@ -0,0 +1,23 @@
...
+                                     absl::string_view resource_type_url)
+    : authority_id_(authority_id), resource_type_url_(resource_type_url) {}
+
+std::string XdsConfigSourceId::toKey() const {
+  // The delimiter between parts of the key.
+  static constexpr char DELIMITER[] = "+";
+
+  // Base64 URL-encoding the authority_id and resource_type_url so that the resulting string does
+  // not contain a "+" character, so that we can use it as a delimiter.
+  return absl::StrCat(Base64Url::encode(authority_id_.data(), authority_id_.size()), DELIMITER,
```
adisuissa, line 18: OOC if this is no longer human-readable, why not hash the concatenated string (w/ "+" or "," as delimiter) or hash each string and combine the hash values?

adisuissa (commented):
Overall LGTM, thanks!
We can start with this and change things if the design changes down the road.
/lgtm api

Should be reviewed by non-Googler. @mattklein123 / @alyssawilk to suggest a reviewer.

source/common/config/grpc_mux_impl.cc
```diff
@@ -272,6 +284,19 @@ void GrpcMuxImpl::onDiscoveryResponse(
...
+
+    // All config updates have been applied without throwing an exception, so we'll call the xDS
+    // resources delegate, if any.
+    if (xds_resources_delegate_.has_value()) {
+      const XdsConfigSourceId source_id(target_xds_authority_, type_url);
+      if (mux_started_) {
+        // Reset the delegate for the type and source, before populating with the new resources.
+        xds_resources_delegate_->reset(source_id);
+      }
+      xds_resources_delegate_->onConfigUpdated(source_id, all_resource_refs);
```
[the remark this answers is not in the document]
abeyad: I think it's okay for returned resources to be overridden.  If two gRPC muxes return resource `R` of type `T` from xDS server `X`, then either they're the same object or the latest one got updated on the server, in which case it makes sense to override what's already in the cache with the new value.

abeyad:
> Overall LGTM, thanks! We can start with this and change things if the design changes down the road. /lgtm api
> 
> Should be reviewed by non-Googler. @mattklein123 / @alyssawilk to suggest a reviewer.

Thanks Adi!  Note: Lizan already approved as a non-Googler maintainer.

source/common/config/xds_source_id.cc
abeyad: That's possible too, though I'm not sure what it adds?  i.e. what are the pros and cons of doing that over what's in this function?  I figured it doesn't matter as long as: (1) there's a function that can produce the same key for the same authority id and type and (2) the delimiter doesn't use characters in the hash, so as to not create a possibly rare but confusing situation where the authority id and type get mixed up if the DELIMITER char is present in the authority id.

mattklein123:
Yes @lizan can do the final approval.

lizan (approved):

lizan:
Defer to @adisuissa for final merge.

adisuissa (approved):
LGTM, thanks!

abeyad:
Thanks for working through this one with me, @adisuissa !

PR #22513, opened 2022-08-02
tools: bootstrap: add support for symlinks in pwd

rmstoi:
Currently bootstrap does not support paths with symlinks.

For example:

ln -s envoy/ envoy-symlink
cd envoy-symlink
support/bootstrap
ERROR: this script must be run at the root of the envoy source tree

$ echo $PWD
/envoy-symlink
$ echo "$(git rev-parse --show-toplevel)"
/envoy

Solution is to use pwd -P (physical: avoid all symlinks) instead of $PWD

$ echo "$(pwd -P)"
/envoy

Signed-off-by: Ruslan Mstoi <ruslan.mstoi@intel.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/support/bootstrap
+++ b/support/bootstrap
@@ -9,7 +9,7 @@
 # TODO(hausdorff): If compatibility becomes an issue here, come back and do this
 # "the right way". This is hard to do in general, since `realpath` is not
 # standard.
-if [ ! "$PWD" == "$(git rev-parse --show-toplevel)" ]; then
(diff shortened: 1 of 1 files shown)


phlax (approved):
lgtm, thanks @rmstoi

PR #22525, opened 2022-08-02
[balsa] Modify two Http1ServerConnectionImplTest.

bencebeky:
http-parser signals an error as soon as the method parsed so far is not
a prefix of any valid method name. BalsaParser on the other hand only
verifies the method when the first line is terminated. Relax the greedy
validation requirement in two tests by adding newline to the end of test
input data. This will allow for some behavioral divergence between the
two parsers.

Tracking issue: #21245

Commit Message: [balsa] Modify two Http1ServerConnectionImplTest.
Additional Description: n/a
Risk Level: low
Docs Changes: n/a
Release Notes: n/a


diannahu (approved):
I think this is a reasonable relaxation. I suppose if BalsaParser never gets the firstline \r\n, then the layer above would eventually hit a timeout of some sort, yeah?

bencebeky:
> I think this is a reasonable relaxation. I suppose if BalsaParser never gets the firstline \r\n, then the layer above would eventually hit a timeout of some sort, yeah?

That is correct for real word data if the peer leaves us hanging.  (This does not sound realistic: why would any peer ever send an incomplete firstline?)  In case of this test, the expectations fail and the test ends, tearing down BalsaParser and  BalsaFrame.

KBaichoo (commented):
For Balsa, do we have a limit in balsa for how many bytes will be read in the request line before \r\n? pointers?

bencebeky:
> For Balsa, do we have a limit in balsa for how many bytes will be read in the request line before \r\n? pointers?

Yes.  The `max_header_length` limit is passed to BalsaFrame at https://sourcegraph.com/github.com/envoyproxy/envoy@a00c314/-/blob/source/common/http/http1/balsa_parser.cc?L97, and it applies to the headers globally, at https://github.com/google/quiche/blob/31448f4274053a18a0dbeb2e14d0aa15ba0507e5/quiche/balsa/balsa_frame.cc#L830.  Here's the test demonstrating the behavior:
https://github.com/google/quiche/blob/31448f4274053a18a0dbeb2e14d0aa15ba0507e5/quiche/balsa/balsa_frame_test.cc#L2404-L2421

KBaichoo (approved):
lgtm, thanks

PR #22540, opened 2022-08-03
test: allowing many test filters to be upstream filters

alyssawilk:

Risk Level: n/a (test only)
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/extensions/filters/http/common/empty_http_filter_config.h
+++ b/test/extensions/filters/http/common/empty_http_filter_config.h
@@ -44,7 +44,7 @@ private:
   const std::string name_;
 };
 
-class UpstreamFilterConfig : public virtual Server::Configuration::UpstreamHttpFilterConfigFactory {
+class UpstreamFilterConfig : public Server::Configuration::UpstreamHttpFilterConfigFactory {
 public:
   virtual Http::FilterFactoryCb
   createDualFilter(const std::string& stat_prefix,
--- a/test/extensions/filters/http/common/empty_http_filter_config.h
+++ b/test/extensions/filters/http/common/empty_http_filter_config.h
@@ -44,6 +44,29 @@ private:
   const std::string name_;
 };
 
+class UpstreamFilterConfig : public virtual Server::Configuration::UpstreamHttpFilterConfigFactory {
+public:
+  virtual Http::FilterFactoryCb
+  createDualFilter(const std::string& stat_prefix,
+                   Server::Configuration::ServerFactoryContext& context) PURE;
+
+  Http::FilterFactoryCb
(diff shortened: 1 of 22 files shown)


test/extensions/filters/http/common/empty_http_filter_config.h
```diff
@@ -44,6 +44,29 @@ class EmptyHttpFilterConfig : public Server::Configuration::NamedHttpFilterConfi
   const std::string name_;
 };
 
+class UpstreamFilterConfig : public virtual Server::Configuration::UpstreamHttpFilterConfigFactory {
+public:
+  virtual Http::FilterFactoryCb
+  createDualFilter(const std::string& stat_prefix,
```
KBaichoo, line 50: Can you chime in on the naming "Dual"? Is this the terminology we're going with for filter config that works for both upstream and downstream filters? It was a bit confusing at first

test/extensions/filters/http/common/empty_http_filter_config.h
```diff
@@ -44,6 +44,29 @@ class EmptyHttpFilterConfig : public Server::Configuration::NamedHttpFilterConfi
   const std::string name_;
 };
 
+class UpstreamFilterConfig : public virtual Server::Configuration::UpstreamHttpFilterConfigFactory {
```
KBaichoo, line 47: I don't think this needs virtual inheritance, can you clarify where you ended up needing it?

test/integration/filters/add_body_filter.cc
```diff
@@ -142,5 +142,9 @@ class AddBodyFilterFactory : public Extensions::HttpFilters::Common::FactoryBase
   }
 };
 
+using UpstreamAddBodyFilterFactory = AddBodyFilterFactory;
+
 REGISTER_FACTORY(AddBodyFilterFactory, Server::Configuration::NamedHttpFilterConfigFactory);
+REGISTER_FACTORY(UpstreamAddBodyFilterFactory,
```
KBaichoo, line 148: nit: for dual factories, perhaps we should have a `REGISTER_DUAL_FACTORY` or similar to enforce both registrations?

KBaichoo (commented):
exciting, mostly looks good

KBaichoo:
/wait

Seems like this isn't yet ready for next pass

test/extensions/filters/http/common/empty_http_filter_config.h
KBaichoo: Yea a follow up makes sense. Make we'll get a better sense of a good name for it.

KBaichoo (approved):
looks good, thanks

Issue #13877, opened 2020-11-03
Rate limiting support for web socket connections


NomadXD:
### Rate limiting support for web socket connections

Is there a recommended way to rate limit the web socket connections ? I went through the docs and some other resources and seems like there's no straight forward way to have the rate limiting for web socket connections.(Maybe use the network rate limit filter? ) **So basically my requirement is to capture web socket frames and possibly decode them to extract the information I need for rate limiting.** Is it possible to achieve my requirement by writing a network filter to **capture, extract and send information to an external rate limiting service ?**

mattklein123:
cc @alyssawilk I think using the HTTP ratelimit filter in the upgrade filter chain would work here? cc @phlax also for potential doc opportunities.

alyssawilk:
So your first request was to limit WebSocket connections, which as Matt says can be done with the existing rate limit filter.  Your follow-up comment makes it look like you actually want to rate limit not just WebSocket connections but actual WebSocket work, which can't be done with Envoy today as Envoy doesn't frame WebSocket.  Which are you hoping for?  You could arguably write a filter which does the second, if that's the granularity you're looking for.

NomadXD:
@mattklein123 @alyssawilk Thanks a lot for the quick response. So yeah I want to rate limit web socket work not just the number of web socket connections. So can I achieve that by writing a network filter to capture the web socket frames ? I mean will the web socket frames hit the network filters before proxying to upstream clusters ?

alyssawilk:
Again I think you want to write an HTTP filter, not a network filter, since you want to take the HTTP payload and parse WebSocket frames from there.
There may also be some work to get RateLimit code to do what you want - I'm less familiar with that part of the stack.

NomadXD:
@alyssawilk okay but will the web socket frames go that far to the http filters ? As I read only the initial upgrade request go to http filters and then after the connection establishment it behaves more like a raw TCP proxy.

mattklein123:
You can use an upgrade filter chain to use normal HTTP filters before the data hits the tcp_proxy layer.

I think your best bet would be to write a custom filter that sets dynamic metadata with some business logic, and then uses the normal HTTP rate limit filter to rate limit on dynamic metadata.

NomadXD:
@mattklein123 thanks a lot. will try it out and see.

NomadXD:
[...]

PR #22542, opened 2022-08-03
websocket: websocket codec implementation

Amila-Rukshan:
Commit Message:
websocket: websocket codec implementation

Testing: Unit tests added

Additional Description:
Related issue comment => https://github.com/envoyproxy/envoy/issues/13877#issuecomment-755722888
This is as the initial step to support rate limiting for WebSocket.

Risk Level:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/websocket/codec.cc
```diff
@@ -0,0 +1,215 @@
...
+  frame_.payload_length_ = 0;
+  frame_.payload_ = nullptr;
+  frame_.is_masked_ = false;
+  frame_.masking_key_ = 0;
+}
+
+uint64_t FrameInspector::inspect(const Buffer::Instance& data) {
+  uint64_t delta = 0;
+  for (const Buffer::RawSlice& slice : data.getRawSlices()) {
+    uint8_t* mem = reinterpret_cast<uint8_t*>(slice.mem_);
```
RyanTheOptimist, line 109: nit: Can we make this const, since we're not modifying the data (I think)

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,215 @@
...
+  output_->push_back(std::move(frame_));
+  frame_.flags_and_opcode_ = 0;
+  frame_.payload_length_ = 0;
+  frame_.payload_ = nullptr;
+  frame_.is_masked_ = false;
+  frame_.masking_key_ = 0;
+}
+
+uint64_t FrameInspector::inspect(const Buffer::Instance& data) {
+  uint64_t delta = 0;
```
RyanTheOptimist, line 107: Can you add a comment which explains what delta represents? Is this the number of bytes parsed, perhaps? If so, maybe `num_bytes_parsed` would be a better name?

source/common/websocket/codec.h
```diff
@@ -0,0 +1,179 @@
...
+  virtual void frameDataStart() {}
+  virtual void frameData(uint8_t*, uint64_t) {}
+  virtual void frameDataEnd() {}
+
+  State state_{State::FhFlagsAndOpcode};
+  uint64_t length_{0};
+  uint8_t length_of_extended_length_{0};
+  uint32_t masking_key_{0};
+  uint8_t masking_key_length_{0};
+  uint64_t count_{0};
```
RyanTheOptimist, line 135: what does `count_` represent?

RyanTheOptimist (commented):
Some initial comments. But I do have a high-level question. How will this code be used? Is there a github issue which explains the design we're planning to follow for integrating Websocket code into Envoy? If so, can you link it to the PR?

source/common/grpc/codec.h
```diff
@@ -49,9 +49,9 @@ class Encoder {
 // Wire format (http://www.grpc.io/docs/guides/wire.html) of GRPC data frame
 // header:
 //
-// -----------------------------------------------------------------------
-// |R|R|R|R|R|R|R|R|C|      L     |      L     |      L     |      L     |
-// -----------------------------------------------------------------------
```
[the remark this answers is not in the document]
Amila-Rukshan: It's a minor change as 1 byte was mistakenly depicted using 9 bits with an additional reserved bit (R). It's independent from WebSocket. will send a separate PR for this.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,179 @@
...
+};
+
+class Decoder : public FrameInspector {
```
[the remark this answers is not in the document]
Amila-Rukshan: Decoder actually decode the bytes in the buffer into `WebSocket::Frame` struct while inspector only traverse the bytes by counting the frames. It only looks at mask bit and length of a frame and move to the start of the next frame without creating actual frames (only few methods FrameInspector has a method body).

source/common/websocket/codec.h
```diff
@@ -0,0 +1,179 @@
...
+#include <cstdint>
+#include <vector>
+
+#include "envoy/buffer/buffer.h"
+
+namespace Envoy {
+namespace WebSocket {
+
+// opcodes
+constexpr uint8_t FRAME_OPCODE_CONT = 0x0;
```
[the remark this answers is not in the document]
Amila-Rukshan: Updated the opcodes with decimal values

source/common/websocket/codec.h
```diff
@@ -0,0 +1,179 @@
...
+
+  // Determine the length of the current frame being decoded. This is useful when supplying a
+  // partial frame to decode() and wanting to know how many more bytes need to be read to complete
+  // the frame.
+  uint32_t length() const { return frame_.payload_length_; }
+
+  // Indicates whether it has buffered any partial data.
+  bool hasBufferedData() const { return state_ != State::FhFlagsAndOpcode; }
+
+  Frame& getFrame() { return frame_; };
```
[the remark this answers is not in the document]
Amila-Rukshan: it is used to check the vales of a halfway decoded frames in tests

source/common/websocket/codec.h
Amila-Rukshan: That is the total frames count when identified when inspect was called. I have renamed it to `total_frames_count_`.

source/common/websocket/codec.cc
Amila-Rukshan: It is the number of frames parsed in a single inspect call. I have renamed it as `frames_count_` and it's self descriptive now.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,179 @@
...
+  virtual void frameMaskFlag(uint8_t mask_and_length) {
+    masking_key_length_ = (mask_and_length & 0x80) ? MASKING_KEY_LENGTH : 0;
+    length_ = mask_and_length & 0x7F;
+  }
+  virtual void frameMaskingKey() {}
+  virtual void frameDataStart() {}
+  virtual void frameData(uint8_t*, uint64_t) {}
+  virtual void frameDataEnd() {}
+
+  State state_{State::FhFlagsAndOpcode};
```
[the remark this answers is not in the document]
Amila-Rukshan: Some members are shared between `Decoder` and `FrameInspector`. Made other private members.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,215 @@
...
+void Decoder::frameDataEnd() {
+  output_->push_back(std::move(frame_));
+  frame_.flags_and_opcode_ = 0;
+  frame_.payload_length_ = 0;
+  frame_.payload_ = nullptr;
+  frame_.is_masked_ = false;
+  frame_.masking_key_ = 0;
+}
+
+uint64_t FrameInspector::inspect(const Buffer::Instance& data) {
```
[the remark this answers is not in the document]
Amila-Rukshan: since I have already organised it using visitor pattern with methods `frameStart`, `frameMaskFlag`, `frameMaskingKey`, `frameDataStart`, ...etc, If it is again grouped with methods, then the readability will reduce as it needs to check nested function calls. This is same as how grpc codec is organised actually.

test/per_file_coverage.sh
```diff
@@ -32,6 +32,7 @@ declare -a KNOWN_LOW_COVERAGE=(
 "source/common/tracing:96.1"
 "source/common/upstream:96.0"
 "source/common/watchdog:58.6" # Death tests don't report LCOV
+"source/common/websocket:92.3"
```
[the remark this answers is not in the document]
Amila-Rukshan: Added more tests and increased the coverage

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,215 @@
...
+#include <vector>
+
+#include "source/common/buffer/buffer_impl.h"
+
+#include "absl/container/fixed_array.h"
+
+namespace Envoy {
+namespace WebSocket {
+
+Encoder::Encoder() = default;
```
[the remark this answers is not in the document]
RyanTheOptimist: Un-resolving because this is still in the .cc file, not .h.

source/common/websocket/codec.cc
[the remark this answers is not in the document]
RyanTheOptimist: I don't think I understand. I'm looking at the code in this method in each the various case statements. For example case State::FrameHeaderMaskFlagAndLength: has 22 lines in the body. I'm suggesting factoring this out to a helper method. Then the main body of this method would be:

```
switch(state_) {
  case kFoo:
    doInspectFoo();
    break;
  case kBar:
   doInspectBar();
   break;
}
```

Or some such. I'm not sure how that would result in less readable code. Can you say more?

source/common/websocket/codec.h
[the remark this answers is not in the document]
RyanTheOptimist: Do we expect there to be users of Inspector other than Decoder?

source/common/websocket/codec.h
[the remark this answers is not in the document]
RyanTheOptimist: In general it's not a great idea to expose internal state simply for tests. But I don't think I see this method being used in any tests. Maybe we can just remove it?

test/common/websocket/codec_test.cc
```diff
@@ -266,31 +232,178 @@ TEST(WebSocketCodecTest, decode16BitBinaryFrame) {
...
+    Decoder decoder;
+    std::vector<Frame> frames;
+    EXPECT_TRUE(decoder.decode(buffer, frames));
+    EXPECT_EQ(true, frames[0].masking_key_.has_value());
+    EXPECT_EQ(0x4f5a12e2, frames[0].masking_key_.value());
+    EXPECT_EQ(1, frames.size());
+    EXPECT_EQ(0, frames[0].payload_length_);
+  }
+
+  // Non-masked 7 bit
```
RyanTheOptimist, line 326: nit: Please break these out into individual named tests.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,219 @@
...
+
+#include "source/common/buffer/buffer_impl.h"
+#include "source/common/common/scalar_to_byte_vector.h"
+
+namespace Envoy {
+namespace WebSocket {
+
+Encoder::Encoder() = default;
+
+void Encoder::newFrameHeader(const Frame& frame, std::vector<uint8_t>& output) {
```
RyanTheOptimist, line 17: nit: Instead of returning void and taking output as an outparam, let's return the output vector.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,219 @@
...
+  if (frame.payload_length_ <= 125) {
+    // Set mask bit and 7-bit length
+    pushScalarToByteVector(frame.masking_key_ ? static_cast<uint8_t>(frame.payload_length_ | 0x80)
+                                              : static_cast<uint8_t>(frame.payload_length_),
+                           output);
+  } else if (frame.payload_length_ <= 65535) {
+    // Set mask bit and 16-bit length indicator
+    pushScalarToByteVector(static_cast<uint8_t>(frame.masking_key_ ? 0xfe : 0x7e), output);
+    // Set 16-bit length
+    for (uint8_t i = 1; i <= PAYLOAD_LENGTH_SIXTEEN_BIT; i++) {
```
RyanTheOptimist, line 31: pushScalarToByteVector already knows how to add the individual bytes one at a time to the vector. Ca we do something like:

```
uint16_t len = frame.payload_length_;
pushScalarToByteVector(len, output)
```

source/common/websocket/codec.h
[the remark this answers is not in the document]
RyanTheOptimist: Members need to be private, according to the style guide. If we need to share state with a subclass, we can do this via protected methods.

That being said, I think we should either:
* Unify Inspector and Decoder into a single class (if there are no other consumers of Inspector), or
* Have Inspector define a Visitor interface that Decoder implements

source/common/websocket/codec.h
```diff
@@ -0,0 +1,178 @@
...
+#include <cstdint>
+#include <vector>
+
+#include "envoy/buffer/buffer.h"
+
+namespace Envoy {
+namespace WebSocket {
+
+// Opcodes (https://datatracker.ietf.org/doc/html/rfc6455#section-11.8)
+constexpr uint8_t FRAME_OPCODE_CONTINUATION = 0;
```
RyanTheOptimist, line 13: All these constexpr globals should be named kSomeName as per the style guide:

https://google.github.io/styleguide/cppguide.html#Constant_Names

RyanTheOptimist (commented):
A few more comments. I've not dug too deeply into the tests yet.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,179 @@
...
+//  | Masking-key (continued)       |          Payload Data         |
+//  +-------------------------------- - - - - - - - - - - - - - - - +
+//  : .... Payload Data continued .... Payload Data continued ..... :
+//  + - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - +
+//  | .... Payload Data continued .... Payload Data continued ..... |
+//  +---------------------------------------------------------------+
+
+struct Frame {
+  // |F|R|R|R| opcode(4) |
+  uint8_t flags_and_opcode_;
```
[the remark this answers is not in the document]
Amila-Rukshan: split into  `bool final_fragment_` and `uint8_t opcode_`

source/common/websocket/codec.h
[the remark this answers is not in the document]
Amila-Rukshan: removed it

source/common/websocket/codec.cc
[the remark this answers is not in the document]
Amila-Rukshan: updated the code as suggested

source/common/websocket/codec.h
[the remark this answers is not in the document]
Amila-Rukshan: I changed the code as the first way "Unify Inspector and Decoder into a single class"

source/common/websocket/codec.h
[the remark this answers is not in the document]
Amila-Rukshan: I think decode is enough and made inspector the the base class

source/common/websocket/codec.cc
Amila-Rukshan: I tried it but it didn't work as expected when payload length is 256 test cases

source/common/websocket/codec.h
```diff
@@ -0,0 +1,168 @@
...
+  Encoder() = default;
+
+  // Creates a new Websocket data frame header with the given frame data.
+  // @param frame supplies the frame to be encoded.
+  // @return std::vector<uint8_t> buffer with encoded header data.
+  std::vector<uint8_t> newFrameHeader(const Frame& frame);
+};
+
+// Current state of the frame that is being processed.
+enum class State {
```
RyanTheOptimist, line 81: nit: Since this is basically an implementation detail for Decoder (or DecoderBase currently) consider moving this to be a private inner class of Decoder (or DecoderBase).

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,227 @@
...
+#include "source/common/common/scalar_to_byte_vector.h"
+
+namespace Envoy {
+namespace WebSocket {
+
+std::vector<uint8_t> Encoder::newFrameHeader(const Frame& frame) {
+  std::vector<uint8_t> output;
+  // Set flags and opcode
+  pushScalarToByteVector(
+      static_cast<uint8_t>(frame.final_fragment_ ? (0x80 ^ frame.opcode_) : frame.opcode_), output);
```
RyanTheOptimist, line 19: Out of curiosity, why `^` vs `|`?

source/common/websocket/codec.cc
Amila-Rukshan: yes, endianness was the issue. `htobe16`, `htobe64` and `htobe32` could be used including other places.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  for (const Buffer::RawSlice& slice : data.getRawSlices()) {
+    const uint8_t* mem = reinterpret_cast<uint8_t*>(slice.mem_);
+    for (uint64_t slice_index = 0; slice_index < slice.len_;) {
+      uint8_t c = *mem;
+      switch (state_) {
+      case State::KFrameHeaderFlagsAndOpcode:
+        if (!frameStart(c)) {
+          return frames_count_;
+        }
+        total_frames_count_ += 1;
```
RyanTheOptimist, line 193: Is this value ever used?

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  inspect(input);
+  if (decoding_error_) {
+    return absl::nullopt;
+  }
+  input.drain(input.length());
+  return std::move(output_);
+}
+
+uint64_t Decoder::inspect(const Buffer::Instance& data) {
+  uint64_t frames_count_ = 0;
```
RyanTheOptimist, line 183: nit: local variables should not end with `_`, that's reserved for class members.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  masking_key_length_--;
+  if (masking_key_length_ == 0) {
+    frameMaskingKey();
+    frameDataStart();
+  }
+  slice_index++;
+  mem++;
+}
+
+void Decoder::doDecodePayload(const Buffer::RawSlice& slice, const uint8_t*& mem,
```
RyanTheOptimist, line 152: It looks like `slice` is only used to compute `remain_in_buffer`. Instead of passing in 3 different arguments how about we pass in an `absl::Span<uint8_t> data` which is basically a pointer and a length. If we did this, then `remain_in_buffer` would simply be `data.length()`

source/common/websocket/codec.h
```diff
@@ -0,0 +1,146 @@
...
+  // Current frame that is being decoded.
+  Frame frame_;
+  // Data holder for successfully decoded frames.
+  absl::optional<std::vector<Frame>> output_ = absl::nullopt;
+  bool decoding_error_ = false;
+  State state_ = State::KFrameHeaderFlagsAndOpcode;
+  uint64_t total_frames_count_ = 0;
+  uint8_t length_of_extended_length_ = 0;
+  uint32_t masking_key_ = 0;
+  uint64_t length_ = 0;
```
RyanTheOptimist, line 141: Is this field different from `frame_.payload_length_`? If not, let's remove it.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,146 @@
...
+  };
+  // Current frame that is being decoded.
+  Frame frame_;
+  // Data holder for successfully decoded frames.
+  absl::optional<std::vector<Frame>> output_ = absl::nullopt;
+  bool decoding_error_ = false;
+  State state_ = State::KFrameHeaderFlagsAndOpcode;
+  uint64_t total_frames_count_ = 0;
+  uint8_t length_of_extended_length_ = 0;
+  uint32_t masking_key_ = 0;
```
RyanTheOptimist, line 140: Is this field different from `frame_.masking_key`? If not, let's remove it.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  }
+  return output;
+}
+
+bool Decoder::frameStart(uint8_t flags_and_opcode) {
+  // Validate opcode (last 4 bits)
+  uint8_t opcode = flags_and_opcode & 0x0f;
+  if (std::find(kFrameOpcodes.begin(), kFrameOpcodes.end(), opcode) != kFrameOpcodes.end()) {
+    frame_.opcode_ = opcode;
+    // Set final fragment flag
```
RyanTheOptimist, line 50: nit: let's remove this comment as the code is already self-describing

https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  decoding_error_ = true;
+  return false;
+}
+
+void Decoder::frameMaskFlag(uint8_t mask_and_length) {
+  // Set masked flag
+  if (mask_and_length & 0x80) {
+    masking_key_length_ = kMaskingKeyLength;
+  } else {
+    masking_key_length_ = 0;
```
RyanTheOptimist, line 63: nit: instead of an int which only holds 2 different values, how about making this a bool?

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  }
+  // Set length (0 to 125) or length flag (126 or 127)
+  length_ = mask_and_length & 0x7F;
+}
+
+void Decoder::frameMaskingKey() {
+  frame_.masking_key_ = masking_key_;
```
RyanTheOptimist, line 70: If we can use `frame_.masking_key_` instead of `masking_key_` then I think we can eliminate this method completely.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+    length_of_extended_length_ = kPayloadLength64Bit;
+    length_ = 0;
+    state_ = State::KFrameHeaderExtendedLength;
+  } else if (masking_key_length_ > 0) {
+    state_ = State::KFrameHeaderMaskingKey;
+  } else {
+    frameDataStart();
+  }
+  slice_index++;
+  mem++;
```
RyanTheOptimist, line 116: It looks like these two argument are passed in simply to increment then. Instead, let's increment them at the caller.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  }
+  output_.value().push_back(std::move(frame_));
+  frame_.final_fragment_ = false;
+  frame_.opcode_ = 0;
+  frame_.payload_length_ = 0;
+  frame_.payload_ = nullptr;
+  frame_.masking_key_ = absl::nullopt;
+}
+
+void Decoder::doDecodeMaskFlagAndLength(const uint8_t c, const uint8_t*& mem,
```
RyanTheOptimist, line 99: nit: here and elsewhere `c` is not a very readable name for a function parameter. Let's make it something more descriptive. Perhaps `mask`?

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  length_of_extended_length_--;
+  if (length_of_extended_length_ == 0) {
+    if (masking_key_length_ > 0) {
+      state_ = State::KFrameHeaderMaskingKey;
+    } else {
+      frameDataStart();
+    }
+  }
+  slice_index++;
+  mem++;
```
RyanTheOptimist, line 134: ditto.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  } else {
+    masking_key_ |= static_cast<uint32_t>(c) << 8 * (masking_key_length_ - 1);
+  }
+  masking_key_length_--;
+  if (masking_key_length_ == 0) {
+    frameMaskingKey();
+    frameDataStart();
+  }
+  slice_index++;
+  mem++;
```
RyanTheOptimist, line 149: ditto

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+}
+
+void Decoder::doDecodeMaskingKey(const uint8_t c, const uint8_t*& mem, uint64_t& slice_index) {
+  if (masking_key_length_ == 1) {
+    masking_key_ |= static_cast<uint32_t>(c);
+  } else {
+    masking_key_ |= static_cast<uint32_t>(c) << 8 * (masking_key_length_ - 1);
+  }
+  masking_key_length_--;
```
RyanTheOptimist, line 143: So is this method call N times, each with one byte of the masking key in `c`?

RyanTheOptimist (commented):
Thanks for the changes. We're starting to get there. I've added a bunch of comments which I think will simplify the code. I think there will be an opportunity to continue to simplify.

I'm also curious about the expected behavior of the Decoder with respect to partial frames. Is the input required to contain only whole frames, or can it also contain partial frames. If partial frames are allowed, what is the expected behavior. Relatedly, can a particular slice contain a partial frame? Like is it possible for the length field to be split across two different slices? I don't think I saw tests for this behavior.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,146 @@
...
+  // @return std::vector<uint8_t> buffer with encoded header data.
+  std::vector<uint8_t> encodeFrameHeader(const Frame& frame);
+};
+
+// Decoder decodes bytes in input buffer into in-memory WebSocket frames.
+class Decoder {
+public:
+  // Decodes the given buffer with WebSocket frame. Drains the input buffer when
+  // decoding succeeded (returns true). If the input is not sufficient to make a
+  // complete WebSocket frame, it will be buffered in the decoder. If a decoding
```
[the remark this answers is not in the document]
Amila-Rukshan: There were few with inspector code earlier.  Now added these `decodeIncompleteFrameWithgOnlyFrameHeader`, `decodeIncompleteFrameUptoPartialLength`, `decodeIncompleteFrameUptoPartialMaskingKey`,  and `decodeIncompleteFrameWithPartialPayload`

source/common/websocket/codec.cc
Amila-Rukshan: yes, that's how complete masking key is created

source/common/websocket/codec.h
Amila-Rukshan: `length_` is the processing length, `length_` will get the final value after enough bytes are consumed and at the end of the payload is also processed, then we set the `frame_.payload_length_` to its value.

source/common/websocket/codec.cc
Amila-Rukshan: it has to be an int value as we decrement it when the `frame.masking_key_` is generated from byte to byte

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+
+absl::optional<std::vector<Frame>> Decoder::decode(Buffer::Instance& input) {
+  decoding_error_ = false;
+  inspect(input);
+  if (decoding_error_) {
+    return absl::nullopt;
+  }
+  input.drain(input.length());
```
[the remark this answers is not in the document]
Amila-Rukshan: yes, it will drain the input for an incomplete frame and the state of the halfway decoded frame is saved in Decoder. When the next call to decode with the rest of the data it will return the completed frame in the vector.

source/common/websocket/codec.cc
Amila-Rukshan: simplified after passing `slice.len_` instead of `slice`. could to figure out a way to get `absl::Span<uint8_t>` from `RawSlice`.

Amila-Rukshan:
> Is the input required to contain only whole frames, or can it also contain partial frames.

both ways are valid inputs. `input` will drain only up to the complete frames (successfully decoded frames).

> Relatedly, can a particular slice contain a partial frame? Like is it possible for the length field to be split across two different slices? I don't think I saw tests for this behaviour.

Yes these are possibilities, tests have been added now.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,143 @@
...
+    // length/length flag (7 bit) of the WebSocket data frame.
+    FrameHeaderMaskFlagAndLength,
+    // Waiting for decoding the extended length of the frame if length read previously
+    // is either 126 or 127. Respectively 2 bytes or 8 bytes will be decoded from the
+    // WebSocket data frame.
+    FrameHeaderExtendedLength,
+    // Waiting for decoding the masking key (4 bytes) only if the mask bit is set.
+    FrameHeaderMaskingKey,
+    // Waiting for decoding the payload (both extension data and application data).
+    FramePayload,
```
RyanTheOptimist, line 131: Let's add an explicit state for `FrameFinished` which will allow the switch statement to have an explicit place for processing the end of the frame.

source/common/websocket/codec.cc
RyanTheOptimist: I think we can do something like this:

```
uint64_t Decoder::doDecodePayload(absl::Span<uint8_t> data) {
  if (data.length() >= length_) {
    frameData(data.data(), length_);
    state_ = State::FrameFinished;
    return length_;
  }
  frameData(data.data(), data.length());  
  return data.length();
}
```

Then when we call this method in from `inspect()` we can do:

```
  case State::FramePayload:
        uint64_t bytes_processed = doDecodePayload(absl_span<uint8_t>(mem, slice.len_ - slice_index);
        slice_index += bytes_processed;
        mem += bytes_processed;
        break;
```

source/common/websocket/codec.h
RyanTheOptimist: I think you intended to remove this field but I still see it.

source/common/websocket/codec.cc
RyanTheOptimist: Ah, so this is really the number of bytes of the key left to read? In that case how about `remaining_masking_key_length_` or `num_remaining_masking_key_bytes_`

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,215 @@
...
+  // Set the mask length to be read
+  masking_key_length_ = mask_and_length & 0x80 ? kMaskingKeyLength : 0;
+  // Set length (0 to 125) or length flag (126 or 127)
+  length_ = mask_and_length & 0x7F;
+}
+
+void Decoder::frameDataStart(Buffer::Instance& input, absl::optional<std::vector<Frame>>& output) {
+  frame_.payload_length_ = length_;
+  if (length_ == 0) {
+    frameDataEnd(input, output);
```
RyanTheOptimist, line 67: Instead of doing the "End" processing here, let's add an explicit Finished state. Then we can do:

```
void Decoder::frameDataStart() {
  frame_.payload_length_ = length_;
  if (length_ == 0) {
    state_ = State::FrameFinished;
  } else {
    frame_.payload_ = std::make_unique<Buffer::OwnedImpl>();
    state_ = State::FramePayload;
  }
}
```

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,215 @@
...
+    return absl::nullopt;
+  }
+  return output;
+}
+
+bool Decoder::inspect(Buffer::Instance& input, absl::optional<std::vector<Frame>>& output) {
+  for (const Buffer::RawSlice& slice : input.getRawSlices()) {
+    const uint8_t* mem = reinterpret_cast<uint8_t*>(slice.mem_);
+    for (uint64_t slice_index = 0; slice_index < slice.len_;) {
+      uint8_t byte_of_frame = *mem;
```
RyanTheOptimist, line 177: There are quite a few local variables here which need to be tracked. How about something like:
```
  for (const Buffer::RawSlice& slice : input.getRawSlices()) {
    absl::Span<const uint8_t> data(reinterpret_cast<uint8_t*>(slice.mem_), slice_len_);
    while (!data.empty()) {
      uint64_t bytes_decoded = 0;
      switch (state_) {
      case State::FrameHeaderFlagsAndOpcode:
        bytes_decoded = doDecodeFlagsAndOpcode(data);
        if (bytes_decoded = 0) {
          return false;
        }
        break;
      case State::FrameHeaderMaskFlagAndLength:
        bytes_decoded = doDecodeMaskFlagAndLength(data);
        break;
      case State::FrameHeaderExtendedLength:
        bytes_decoded = doDecodeExtendedLength(data);
        break;
      case State::FrameHeaderMaskingKey:
        bytes_decoded = doDecodeMaskingKey(data);
        break;
      case State::FramePayload:
        bytes_decoded = doDecodePayload(data);
        break;
      case State::FrameFinished:
        doFrameEnd(input, output);
        break;
      }
      data.remove_prefix(bytes_decoded);
      bytes_consumed_by_frame_ += bytes_decoded;
```

Notice that all of the `doDecodeFoo()` methods now simply take the span, and return the number of bytes they processed. The exception is `doFrameEnd()` which drains the input, adds the frame to the output, and resets state.

source/common/websocket/codec.cc
RyanTheOptimist: As I suggest in another comment i think we should pass in the absl::Span<> to all of these doDecode methods. That would allow us to copy as many bytes as we have into masking key. If they're all ready (common) we'll just call this once and do a simple memcpy.

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,766 @@
...
+
+  ASSERT_TRUE(frames.has_value());
+  EXPECT_EQ(2, frames.value()[0].opcode_);
+  EXPECT_TRUE(frames.value()[0].final_fragment_);
+  EXPECT_EQ(65536, frames.value()[0].payload_length_);
+  EXPECT_EQ(false, frames.value()[0].masking_key_.has_value());
+}
+
+// Frame with only frame header data
+TEST(WebSocketCodecTest, decodeIncompleteFrameWithgOnlyFrameHeader) {
```
RyanTheOptimist, line 291: I think it would be a good idea to have a test which does something like:
* Start with a complete frame in some array.
* Call decode 1 byte at a time to make sure we end up correctly parsing the frame at the end.

RyanTheOptimist (commented):
Getting closer. I provided some code which I think will greatly simplify the implementation.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,220 @@
...
+        frameDataEnd(bytes_consumed_by_frame, input, output);
+        break;
+      }
+      data.remove_prefix(bytes_decoded);
+      bytes_consumed_by_frame += bytes_decoded;
+    }
+    // Handles when slice ended with a complete frame scenario
+    if (state_ == State::FrameFinished) {
+      frameDataEnd(bytes_consumed_by_frame, input, output);
+    }
```
RyanTheOptimist, line 214: Hm, it'd be nice if we could avoid this since this is basically identical to the `case State::FrameFinished:` block earlier. Can we change `while (!data.empty()) {` to while (!data.empty() || state_ != State::FrameFinished) {`? I *think* that will do what we want.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,220 @@
...
+    length_ = 0;
+  }
+  if (length_ == 0) {
+    state_ = State::FrameFinished;
+  }
+  return bytes_decoded;
+}
+
+absl::optional<std::vector<Frame>> Decoder::decode(Buffer::Instance& input) {
+  absl::optional<std::vector<Frame>> output = absl::nullopt;
```
RyanTheOptimist, line 178: Let's initialize this to a valid vector here 
```
absl::optional<std::vector<Frame>> output(std::vector<Frame>());
```
Then we can remove the corresponding code from `frameDataEnd()`;

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+// mask bit - 0
+// length - 5 bytes 0000101
+TEST(WebSocketCodecTest, decodeSingleUnmaskedFrame) {
+  Buffer::OwnedImpl buffer;
+  Buffer::addSeq(buffer, {0x81, 0x05, 0x48, 0x65, 0x6c, 0x6c, 0x6f});
+
+  Decoder decoder;
+  absl::optional<std::vector<Frame>> frames = decoder.decode(buffer);
+
+  ASSERT_TRUE(frames.has_value());
```
RyanTheOptimist, line 33: Let's also assert that `frames` has the right number of frames before we attempt to access any of the frames:
```
ASSERT_EQ(1, frames->size());
```
We should do this elsewhere in the file. (And we can remove the corresponding EXPECT_EQ on line 35, and elsewhere)

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+  EXPECT_EQ(false, frames.value()[0].masking_key_.has_value());
+  EXPECT_EQ(1, frames.value().size());
+  EXPECT_EQ(5, frames.value()[0].payload_length_);
+
+  std::string text_payload;
+  text_payload.resize(5);
+  (*(frames.value()[0].payload_)).copyOut(0, 5, text_payload.data());
+  EXPECT_EQ("Hello", text_payload);
```
RyanTheOptimist, line 41: Here and elsewhere, can we simply do:
```
EXPECT_EQ("Hello", (*frames)[0].payload_->toString());
```
I think that would be easier than using `copyOut()`

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+
+// A single-frame unmasked text message
+// 0x81 0x05 0x48 0x65 0x6c 0x6c 0x6f (contains "Hello")
+//                   H        e        l        l        o
+// 10000001 00000101 01001000 01100101 01101100 01101100 01101111
+// fin - 1
+// opcode - 0x1 - text data
+// mask bit - 0
+// length - 5 bytes 0000101
+TEST(WebSocketCodecTest, decodeSingleUnmaskedFrame) {
```
RyanTheOptimist, line 26: nit: the 2nd argument to TEST() should start with a Capital:
```
TEST(WebSocketCodecTest, DecodeSingleUnmaskedFrame) {
```
Please apply here and elsewhere.

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+// frame-1: 0x01 0x03 0x48 0x65 0x6c (contains "Hel")
+// 00000001 00000011 H e l
+// fin - 0 (not final frame - either first or not last)
+// text frame - 1 (type info is in initial frame)
+//
+// frame-2: 0x80 0x02 0x6c 0x6f (contains "lo")
+// 10000000 00000010 l o
+// fin - 1 (final frame)
+// continuation frame
+TEST(WebSocketCodecTest, decodeFragmentedUnmaskedTextMessage) {
```
RyanTheOptimist, line 55: From the perspective of this case, this is simply decoding 2 frames, right? Since "Fragments" are not really a concept that the Decoder knows about, I think it would be simpler to call this test "DecodeTwoUnmaskedFrames".

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+  EXPECT_EQ(2, frames.value()[1].payload_length_);
+
+  std::string text_payload_second_frame;
+  text_payload_second_frame.resize(2);
+  (*(frames.value()[1].payload_)).copyOut(0, 2, text_payload_second_frame.data());
+  EXPECT_EQ("lo", text_payload_second_frame);
+}
+
+// A fragmented and unfragmented text message
+TEST(WebSocketCodecTest, decodeFragmentedAndUnfragmentedTextMessages) {
```
RyanTheOptimist, line 84: nit: Same comment about the name. Something like "DecodeThreeFrames"

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+  EXPECT_EQ(0x37fa213d, frames.value()[0].masking_key_.value());
+
+  std::string text_payload;
+  text_payload.resize(5);
+  (*(frames.value()[0].payload_)).copyOut(0, 5, text_payload.data());
+  // Unmasking the text payload
+  for (size_t i = 0; i < text_payload.size(); ++i) {
+    text_payload[i] ^= (frames.value()[0].masking_key_.value() >> (8 * (3 - i % 4))) & 0xff;
+  }
+  EXPECT_EQ("Hello", text_payload);
```
RyanTheOptimist, line 157: This block of logic is not actually testing the Decoder. It's testing that the input provided to the Decoder is able to be unmasked. But that's just testing the test, not testing the Decoder. All the decoder knows how to do is extract the masking key correctly (which is checked on line 148) and that it extracts the payload correctly. So I'd just do: `EXPECT_EQ("\7f\9f\4d\51\58", ...)` or some such.

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+}
+
+TEST(WebSocketCodecTest, decodeInvalidFrame) {
+  Encoder encoder;
+  Buffer::OwnedImpl buffer;
+  std::vector<uint8_t> header;
+  // Invalid opcode
+  header = encoder.encodeFrameHeader({false, 0x0b, absl::nullopt, 3, nullptr});
+  buffer.add(header.data(), header.size());
+  buffer.add("Hey");
```
RyanTheOptimist, line 652: Since this is a Decoder test, I don't think we really need the Encoder, do we? Can we instead just provide the literal frame input like we do for the other tests?

At the same time, this test makes me wonder if the Encoder should really really be serialized an invalid frame? It makes me think that it should possibly error out of the opcode is invalid. What do you think?

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+
+  Decoder decoder;
+  size_t size = buffer.length();
+  absl::optional<std::vector<Frame>> frames = decoder.decode(buffer);
+
+  ASSERT_FALSE(frames.has_value());
+  EXPECT_EQ(frames, absl::nullopt);
```
RyanTheOptimist, line 659: This check is redundant with the ASSERT_FALSE() just above it.

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+  Decoder decoder;
+  absl::optional<std::vector<Frame>> frames = decoder.decode(buffer);
+
+  ASSERT_TRUE(frames.has_value());
+  EXPECT_EQ(false, frames.value()[0].masking_key_.has_value());
+  EXPECT_EQ(1, frames.value().size());
+  EXPECT_EQ(0, frames.value()[0].payload_length_);
+}
+
+TEST(WebSocketCodecTest, encodeFrameHeader) {
```
RyanTheOptimist, line 578: Since the Encoder is first in the source file, let's put the encoder tests at the top of the test.cc

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+  Buffer::OwnedImpl buffer;
+  Buffer::addSeq(buffer, {0x81, 0x05, 0x48, 0x65, 0x6c, 0x6c, 0x6f});
+
+  Decoder decoder;
+  absl::optional<std::vector<Frame>> frames = decoder.decode(buffer);
+
+  ASSERT_TRUE(frames.has_value());
+  EXPECT_EQ(false, frames.value()[0].masking_key_.has_value());
+  EXPECT_EQ(1, frames.value().size());
+  EXPECT_EQ(5, frames.value()[0].payload_length_);
```
RyanTheOptimist, line 36: This chunk of EXPECT_EQs does not appear to be checking that all the fields are correct. In particular, it's missing `final_fragment_` and `opcode_`. Since we're going to do lots of checking of frames, I wonder if it might make sense do add a function like:
```
bool areFramesEqual(const Frame& a, const Frame& b) {
  return a.final_fragment_ == b.final_fragment_ &&
    a.opcode_ == b.opcode_ &&
    a.masking_key_ == b.masking_key_ &&
    a.payload_length_ == b.payload_length_ &&
    a.payload_.toString() == b.payload.toString()
}
```
Then we can use it here like:
```
EXPECT_TRUE(areFramesEqual(*frames[0], {false, kFrameOpcodeText, {}, 5, "Hello"}));
```
Or something like that.

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+                          0x04, 0xdb, 0x91, 0xd6, 0x9a, 0xc6, 0xa1, 0xe8, 0xd7, 0xb0, 0x49, 0x6d,
+                          0xd8, 0x00, 0x96, 0x41, 0x8d, 0x54, 0xdf, 0x0a, 0x87, 0xea, 0xe7, 0x4c,
+                          0xdf, 0x16, 0x9a, 0x30, 0xa0, 0x26, 0x1c, 0xc4, 0x2c, 0xad, 0xfd});
+
+  Decoder decoder;
+  absl::optional<std::vector<Frame>> frames = decoder.decode(buffer);
+
+  ASSERT_TRUE(frames.has_value());
+  EXPECT_EQ(1, frames.value().size());
+  EXPECT_EQ(2, frames.value()[0].opcode_);
```
RyanTheOptimist, line 234: Here and elsewhere, when checking an opcode, let's use the named constants such as `kFrameOpcodeText`

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,824 @@
...
+  }
+  EXPECT_EQ(" from", text_payload);
+
+  text_payload.resize(6);
+  (*(frames.value()[2].payload_)).copyOut(0, 6, text_payload.data());
+  // Unmasking the text payload
+  for (size_t i = 0; i < text_payload.size(); ++i) {
+    text_payload[i] ^= (frames.value()[2].masking_key_.value() >> (8 * (3 - i % 4))) & 0xff;
+  }
+  EXPECT_EQ("client", text_payload);
```
RyanTheOptimist, line 819: Let's remove all of the test code which does unmasking since that's not actually testing any production code.

RyanTheOptimist (commented):
Looking good. We're almost there.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,220 @@
...
+uint8_t Decoder::doDecodeFlagsAndOpcode(absl::Span<const uint8_t>& data) {
+  // Validate opcode (last 4 bits)
+  uint8_t opcode = data.front() & 0x0f;
+  if (std::find(kFrameOpcodes.begin(), kFrameOpcodes.end(), opcode) == kFrameOpcodes.end()) {
+    return false;
+  }
+  frame_.opcode_ = opcode;
+  frame_.final_fragment_ = data.front() & 0x80;
+  state_ = State::FrameHeaderMaskFlagAndLength;
+  return 1;
```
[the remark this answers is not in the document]
Amila-Rukshan: `return false` line made as `return 0` as the `doDecode..` functions returns number of decoded bytes.

test/common/websocket/codec_test.cc
Amila-Rukshan: yes it should validate it. opcode validation added to both encoder and decoder and error log also added.

source/common/websocket/codec.cc
```diff
@@ -86,7 +83,8 @@ uint8_t Decoder::doDecodeFlagsAndOpcode(absl::Span<const uint8_t>& data) {
   // Validate opcode (last 4 bits)
   uint8_t opcode = data.front() & 0x0f;
   if (std::find(kFrameOpcodes.begin(), kFrameOpcodes.end(), opcode) == kFrameOpcodes.end()) {
-    return false;
+    ENVOY_LOG(error, "Failed to decode websocket frame with invalid opcode: {}", opcode);
```
RyanTheOptimist, line 86: Hm, I'm not sure about this being an error. Since this input will be coming from the peer (who is potentially an attacker) this means that an attacker could trigger this message repeatedly. As I understand it, logging can be quite expensive so this might be a DoS risk. How about debug?

test/common/websocket/codec_test.cc
```diff
@@ -23,22 +95,15 @@ namespace {
...
-  EXPECT_EQ(false, frames.value()[0].masking_key_.has_value());
-  EXPECT_EQ(1, frames.value().size());
-  EXPECT_EQ(5, frames.value()[0].payload_length_);
-
-  std::string text_payload;
-  text_payload.resize(5);
-  (*(frames.value()[0].payload_)).copyOut(0, 5, text_payload.data());
-  EXPECT_EQ("Hello", text_payload);
+  EXPECT_TRUE(areFramesEqual(frames.value()[0], {true, kFrameOpcodeText, absl::nullopt, 5,
+                                                 std::make_unique<Buffer::OwnedImpl>("Hello", 5)}));
```
RyanTheOptimist, line 106: nit: It's slightly annoying to have to specify both the data and the length in all of these cases. How bout adding a helper method like:

```
BufferPtr makeBuffer(absl::string_view data) {
  return std::make_unique<Buffer::OwnedImpl>(data.data(), data.length());
}
```

And then using it throughout this test..

RyanTheOptimist (commented):
Looks great! 2 small comments, then it'll be good to go!

RyanTheOptimist (dismissed):

RyanTheOptimist:
@ggreenway for a non-Google review

ggreenway:
I'm going to be out the rest of this week. @wbpcode can you do a non-google review on this?

RyanTheOptimist:
@wbpcode friendly ping

wbpcode:
Hi, thanks for this great contribution. I have taken a quick look and it's great.

But I am not expert of websocket and this is a big PR. So I will review it again more carefully this week. Thanks again.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,133 @@
...
+  bool final_fragment_;
+  // Frame opcode.
+  uint8_t opcode_;
+  // The 4 byte fixed size masking key used to mask the payload. Masking/unmasking should be
+  // performed as described in https://datatracker.ietf.org/doc/html/rfc6455#section-5.3
+  absl::optional<uint32_t> masking_key_;
+  // Length of the payload as the number of bytes.
+  uint64_t payload_length_;
+  // WebSocket payload data (extension data and application data).
+  Buffer::InstancePtr payload_;
```
soulxu, line 66: Two questions:
* From the case of the issue https://github.com/envoyproxy/envoy/issues/13877#issuecomment-816508238, people may just want to parse the opcode, and don't care about the payload. And the length of payload can be huge (seems it can be 64bit), can we provide an option to the caller to not buffer the payload data, that will save a lot of memory. (since you make a copy of the payload to the struct Frame)
* Since the length of payload can be huge, should we enable option for the caller limits the length of payload to buffer.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+void Decoder::frameData(const uint8_t* mem, uint64_t length) { frame_.payload_->add(mem, length); }
+
+void Decoder::frameDataEnd(uint64_t& bytes_consumed_by_frame, Buffer::Instance& input,
+                           absl::optional<std::vector<Frame>>& output) {
+  output->push_back(std::move(frame_));
+  resetDecoder();
+  input.drain(bytes_consumed_by_frame);
```
soulxu, line 70: We shouldn't drain the input buffer, the caller can pass the original buffer from decodeData(), if we drain the buffer here, the later HTTP filter won't see the data.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,212 @@
...
+}
+
+uint8_t Decoder::doDecodeExtendedLength(absl::Span<const uint8_t>& data) {
+  uint64_t bytes_to_decode = data.length() <= num_remaining_extended_length_bytes_
+                                 ? data.length()
+                                 : num_remaining_extended_length_bytes_;
+  uint8_t* destination = reinterpret_cast<uint8_t*>(&length_) +
+                         (state_ == State::FrameHeaderExtendedLength16Bit ? kPayloadLength16Bit
+                                                                          : kPayloadLength64Bit) -
+                         num_remaining_extended_length_bytes_;
```
soulxu, line 120: Maybe worth an assertion here ensure destination over the boundary like (destination - length_) < 8

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,212 @@
...
+
+uint8_t Decoder::doDecodeMaskingKey(absl::Span<const uint8_t>& data) {
+  if (!frame_.masking_key_.has_value()) {
+    frame_.masking_key_ = 0;
+  }
+  uint64_t bytes_to_decode = data.length() <= num_remaining_masking_key_bytes_
+                                 ? data.length()
+                                 : num_remaining_masking_key_bytes_;
+  uint8_t* destination = reinterpret_cast<uint8_t*>(&(frame_.masking_key_.value())) +
+                         kMaskingKeyLength - num_remaining_masking_key_bytes_;
```
soulxu, line 143: ditto, maybe worth an assertion like what I said here https://github.com/envoyproxy/envoy/pull/22542/files#r950115326

soulxu (commented):
Only did an overall review. The two questions are majorly about the payload buffer.

source/common/websocket/codec.h
Amila-Rukshan: added another constructor with a bool to stop the payload copying to `Frame`

soulxu:
@Amila-Rukshan thanks for the updating to add an option not to buffer the payload!

But my second comment seems ignored https://github.com/envoyproxy/envoy/pull/22542#discussion_r950022717
`Since the length of payload can be huge, should we enable option for the caller limits the length of payload to buffer.`
I still think it isn't safe to buffer the payload unlimited. but you could let me know your opinion on this.

Other things coming to my mind after thinking of the use case of this WebSocket codec, its can be:
* the filter only care about the metadata of WebSocket frame, that is can be done by the `decode_payload` option which you just add in the last update
* The filter wants the payload, but won't stop the filter iteration, the WebSocket codec will buffer the payload for it. That is available from your implement also.
* the filter wants the payload but will stop the filter iteration until it got the expected data.

For the last case, the HTTP filter actually can stop the filter iteration and ask the HCM(filter manager) to buffer the data, the HTTP filter just need to return those status:
https://github.com/envoyproxy/envoy/blob/068861364d84ba86f8b6bdc57c637afb10eb6692/envoy/http/filter.h#L118-L138

Then the filter can access the buffered decode data through:
https://github.com/envoyproxy/envoy/blob/068861364d84ba86f8b6bdc57c637afb10eb6692/envoy/http/filter.h#L365-L369

So I'm thinking about this case, whether we can ask the WebSocket codec to reference the office in the existing buffered decode data than making a copy?

soulxu:
> @Amila-Rukshan thanks for the updating to add an option not to buffer the payload!
> 
> But my second comment seems ignored [#22542 (comment)](https://github.com/envoyproxy/envoy/pull/22542#discussion_r950022717) `Since the length of payload can be huge, should we enable option for the caller limits the length of payload to buffer.` I still think it isn't safe to buffer the payload unlimited. but you could let me know your opinion on this.
> 
> Other things coming to my mind after thinking of the use case of this WebSocket codec, its can be:
> 
> * the filter only care about the metadata of WebSocket frame, that is can be done by the `decode_payload` option which you just add in the last update
> * The filter wants the payload, but won't stop the filter iteration, the WebSocket codec will buffer the payload for it. That is available from your implement also.
> * the filter wants the payload but will stop the filter iteration until it got the expected data.
> 
> For the last case, the HTTP filter actually can stop the filter iteration and ask the HCM(filter manager) to buffer the data, the HTTP filter just need to return those status:
> 
> https://github.com/envoyproxy/envoy/blob/068861364d84ba86f8b6bdc57c637afb10eb6692/envoy/http/filter.h#L118-L138
> 
> Then the filter can access the buffered decode data through:
> 
> https://github.com/envoyproxy/envoy/blob/068861364d84ba86f8b6bdc57c637afb10eb6692/envoy/http/filter.h#L365-L369
> 
> So I'm thinking about this case, whether we can ask the WebSocket codec to reference the office in the existing buffered decode data than making a copy?

ping @RyanTheOptimist @ggreenway @wbpcode maintainers, ensure I'm dragging the PR in the wrong direction :)

wbpcode:
@soulxu Thanks. Your reviews are very valuable.

> Since the length of payload can be huge, should we enable option for the caller limits the length of payload to buffer. I still think it isn't safe to buffer the payload unlimited. but you could let me know your opinion on this.

I agree with you. A more specific limit configuration would be better.

> So I'm thinking about this case, whether we can ask the WebSocket codec to reference the office in the existing buffered decode data than making a copy?

IMO, it's OK to create a copy if the specific `decode_payload` is set to true. It will ensure the payload can be accessed safely after the buffer is drained.
We can do the memory optimization in the future if it's necessary indeed.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,219 @@
...
+    state_ = State::FramePayload;
+  }
+}
+
+void Decoder::frameData(const uint8_t* mem, uint64_t length) { frame_.payload_->add(mem, length); }
+
+void Decoder::frameDataEnd(absl::optional<std::vector<Frame>>& output) {
+  output->push_back(std::move(frame_));
+  resetDecoder();
+}
```
wbpcode, line 71: If the output is optional, should we check it before call the `push_back`? Or the output cannot be optional?

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,219 @@
...
+  frame_.final_fragment_ = data.front() & 0x80;
+  state_ = State::FrameHeaderMaskFlagAndLength;
+  return 1;
+}
+
+uint8_t Decoder::doDecodeMaskFlagAndLength(absl::Span<const uint8_t>& data) {
+  frameMaskFlag(data.front());
+  if (length_ == 0x7e) {
+    num_remaining_extended_length_bytes_ = kPayloadLength16Bit;
+    length_ = 0;
```
wbpcode, line 98: Tell the true, the `length_ = 0` make me confused at first glance. I think we can simply merge `frameMaskFlag` to this function and set a correct `state_` based on the flag/length indicator value.

wbpcode (commented):
Thanks for your great work. Some initial comments are added.

source/common/websocket/codec.cc
Amila-Rukshan: It can be optional. It was initially as the way like you request now. We have changed it in the review. Now it is initialised to an empty vector.
```
absl::optional<std::vector<Frame>> output = std::vector<Frame>();
```
and do the empty check in the return to make it optional.
```
return !output->empty() ? std::move(output) : absl::nullopt;
```
this will avoid multiple empty check calls.

source/common/websocket/codec.cc
Amila-Rukshan: If `length_` 7 bit value <= 253 then it is the length. If it is 254 or 255 then it is a length indicator and we reset it to zero to read the actual value with the bytes we read in. I added a comment to make it more clear. `frameMaskFlag` is merged to the caller.

source/common/websocket/codec.cc
wbpcode: If the ouput parameter here always has value, may be just use `std::vector<Frame>&` here as the parameter type.

source/common/websocket/codec.cc
```diff
@@ -0,0 +1,220 @@
...
+  }
+  if (length_ == 0) {
+    state_ = State::FrameFinished;
+  }
+  return bytes_decoded;
+}
+
+absl::optional<std::vector<Frame>> Decoder::decode(const Buffer::Instance& input) {
+  absl::optional<std::vector<Frame>> output = std::vector<Frame>();
+  resetDecoder();
```
wbpcode, line 184: The `decodeData` of HTTP filter will be called if there is a read event for body. Will it possible that a frame is split and multiple `decodeData` are called for single frame? Then this resetDecoder seems like will break something?

wbpcode (commented):
Thanks for you quick responses. 👍

soulxu:
> @soulxu Thanks. Your reviews are very valuable.
> 
> > Since the length of payload can be huge, should we enable option for the caller limits the length of payload to buffer. I still think it isn't safe to buffer the payload unlimited. but you could let me know your opinion on this.
> 
> I agree with you. A more specific limit configuration would be better.
> 
Thanks!

> > So I'm thinking about this case, whether we can ask the WebSocket codec to reference the office in the existing buffered decode data than making a copy?
> 
> IMO, it's OK to create a copy if the specific `decode_payload` is set to true. It will ensure the payload can be accessed safely after the buffer is drained. We can do the memory optimization in the future if it's necessary indeed.

Yea, agree that can be implemented in the next step. Also, the HCM's existing decoded data buffer also provides the buffer limit protection. And that is configurable. seems a good choice also. Anyway, totally fine for next step.

source/common/websocket/codec.h
```diff
@@ -0,0 +1,134 @@
...
+  // Creates a new Websocket data frame header with the given frame data.
+  // @param frame supplies the frame to be encoded.
+  // @return std::vector<uint8_t> buffer with encoded header data.
+  absl::optional<std::vector<uint8_t>> encodeFrameHeader(const Frame& frame);
+};
+
+// Decoder decodes bytes in input buffer into in-memory WebSocket frames.
+class Decoder : public Logger::Loggable<Logger::Id::websocket> {
+public:
+  Decoder(bool decode_payload) : decode_payload_{decode_payload} {};
```
soulxu, line 83: remember a previous version you change this to uint64 max_payload, but now it seems to revert back

Amila-Rukshan: @soulxu, updated the PR with that change

test/common/websocket/codec_test.cc
```diff
@@ -0,0 +1,885 @@
...
+  Decoder decoder;
+  absl::optional<std::vector<Frame>> frames = decoder.decode(buffer);
+
+  ASSERT_TRUE(frames.has_value());
+  EXPECT_EQ(2, frames->size());
+  EXPECT_TRUE(areFramesEqual(frames.value()[0],
+                             {true, kFrameOpcodePing, absl::nullopt, 5, makeBuffer("Hello")}));
+  EXPECT_TRUE(areFramesEqual(frames.value()[1], {true, kFrameOpcodePong, 0x37fa213d, 5,
+                                                 makeBuffer("\x7f\x9f\x4d\x51\x58")}));
+} // NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
```
soulxu, line 238: May I ask why we need this?

soulxu (changes requested):
overall LGTM, last one comment and a question

source/common/websocket/codec.h
```diff
@@ -0,0 +1,136 @@
...
+  absl::optional<std::vector<uint8_t>> encodeFrameHeader(const Frame& frame);
+};
+
+// Decoder decodes bytes in input buffer into in-memory WebSocket frames.
+class Decoder : public Logger::Loggable<Logger::Id::websocket> {
+public:
+  Decoder(uint64_t max_payload_length = kMaxPayloadBufferLength)
+      : max_payload_buffer_length_{std::min(max_payload_length, kMaxPayloadBufferLength)} {};
+  // Decodes the given buffer with WebSocket frame. Drains the input buffer when
+  // decoding succeeded (returns true). If the input is not sufficient to make a
```
soulxu, line 88: Those comment `Drains the input buffer when decoding succeeded (returns true)` should be updated now.

test/common/websocket/codec_test.cc
Amila-Rukshan: That is due to `makeBuffer` util function return a pointer. This is to avoided the linter error. I saw some other tests have added like this.

soulxu:
LGTM, thanks!

@wbpcode @ggreenway  for maintainer review

wbpcode (approved):
LGTM. Thanks.

PR #22544, opened 2022-08-03
upstream: adding hidden config and creation for upstream filter facto…

alyssawilk:
adding hidden config for upstream filters, and setting the cluster up to create upstream filters.
upstream filters are as yet unused, still cleaning up (https://github.com/envoyproxy/envoy/pull/22434) but this makes it a much cleaner PR

Also adding 2 random tweaks to router code because coverage was failing despite there previously being no changes to router code :-(

Risk Level:  low (no-op if not configured)
Testing: unit tests
Docs Changes: n/a
Release Notes: n/a


envoy/upstream/upstream.h
```diff
@@ -713,7 +718,7 @@ class TypedLoadBalancerFactory;
 /**
  * Information about a given upstream cluster.
```
RyanTheOptimist, line 719: It might be nice to add a comment which alludes to the fact that this is also a factor for creating upstream filters?

source/common/http/filter_chain_helper.h
```diff
@@ -108,11 +108,13 @@ class FilterChainHelper : Logger::Loggable<Logger::Id::config> {
         proto_config, factory_context_.messageValidationVisitor(), *factory);
     Http::FilterFactoryCb callback =
         factory->createFilterFactoryFromProto(*message, stats_prefix_, factory_context_);
-    dependency_manager.registerFilter(factory->name(), *factory->dependencies());
-    bool is_terminal = factory->isTerminalFilterByProto(*message, server_context_);
-    Config::Utility::validateTerminalFilters(proto_config.name(), factory->name(),
-                                             filter_chain_type, is_terminal,
-                                             last_filter_in_current_config);
+    if (!last_filter_in_current_config || validate_last_filter) {
```
RyanTheOptimist, line 111: What's the story with this new conditional? Is this part of the mysterious router code problems you hit?

source/common/upstream/upstream_impl.h
```diff
@@ -757,6 +758,15 @@ class ClusterInfoImpl : public ClusterInfo,
   std::vector<Http::Protocol>
   upstreamHttpProtocol(absl::optional<Http::Protocol> downstream_protocol) const override;
 
+  // Http::FilterChainFactory
+  void createFilterChain(Http::FilterChainManager& manager) const override {
+    Http::FilterChainUtility::createFilterChainForFactories(manager, http_filter_factories_);
+  }
+  bool createUpgradeFilterChain(absl::string_view, const UpgradeMap*,
+                                Http::FilterChainManager&) const override {
+    return false;
```
RyanTheOptimist, line 767: Mind adding a comment? Presumably something like "// Upgrade filters are not supported on upstream connections"?

source/extensions/upstreams/http/config.h
```diff
@@ -53,6 +53,9 @@ class ProtocolOptionsConfigImpl : public Upstream::ProtocolOptionsConfig {
   bool use_http2_{};
   bool use_http3_{};
   bool use_alpn_{};
+  using FiltersList = Protobuf::RepeatedPtrField<
```
RyanTheOptimist, line 56: Totally your call, but if this type is only used on the line below, maybe there's not much value in the "using"? (Or maybe it's somehow used elsewhere?) Anyway, totally your call. This is fine as is.

RyanTheOptimist (dismissed):
Looks great! A few trivial comments, but LGTM

source/common/upstream/upstream_impl.cc
```diff
@@ -975,6 +978,20 @@ ClusterInfoImpl::ClusterInfoImpl(
...
+            server_factory_context_);
+    Http::FilterChainUtility::FiltersList http_filters = http_protocol_options_->http_filters_;
+
+    std::string prefix = stats_scope_->symbolTable().toString(stats_scope_->prefix());
+    Http::FilterChainHelper<Server::Configuration::ServerFactoryContext,
+                            Server::Configuration::UpstreamHttpFilterConfigFactory>
+        helper(*filter_config_provider_manager, server_factory_context_, server_factory_context_,
+               prefix);
+    helper.processFilters(http_filters, "http", "http", http_filter_factories_, false);
+  }
```
mattklein123, line 994: Does this print any logs when it's building the chain similar to how the downstream logs print? Might be useful for debugging / just letting people know that the config was correct? It also occurs to me that obviously with the downstream upstream/split there is no guarantee these will ever get called if we don't end up doing HTTP on the cluster but I don't think there is much we can do about that, much like the other settings I guess. Mostly just thinking about how to let people know what they configured is correct and working. Fine to have a TODO / follow up in a future PR.

source/common/http/filter_chain_helper.h
mattklein123: Was just about to ask the same thing. :) Maybe some comments?

mattklein123 (approved):
LGTM with some small comments. Feel free to address in this one or in a follow up depending on @RyanTheOptimist comments.

source/common/http/filter_chain_helper.h
alyssawilk: haha, this is an artifact of how upstream filters are created.  I already had a TODO in the main PR to clean it up but I'll add an explanation and a TODO here as well.

RyanTheOptimist (approved):

source/common/upstream/upstream_impl.h
alyssawilk: 1) wasn't needed at the time and 2) when it's added we'll have to be careful to document so folks don't get confused.  We can apply filters to upgrades where we're still doing L7 (passthrough upgrades) but not where we're handling the upgrade (terminating upgrades) because the upstream is then a TCP upstream and only handles L4 filters.

alyssawilk: for upgrades you can have something like
[connect headers][non-http-body payload]
Envoy can either proxy this through as a connect request, forwarding on both the headers and payload, or terminate the connect request, strip the headers and forward only the payload on.  The former is proxied as HTTP, so gets upstream filters.  the latter is forwarded as TCP so no upstream L7 filters

see https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades for more details

EItanya: Ahh I see what you're saying, in my case I'm specifically working with `WebSocket HTTP/1.1` upgrades so I should be fine?

alyssawilk: yeah I think the filters would apply but the docs still need to be clear for folks with other use cases

PR #22550, opened 2022-08-03
router: rename a badly named function

alyssawilk:


alyssawilk:
(it's been bugging me that upstream filter encodeHeaders then does UpstreamFilterManager decodeHeaders.  Confusing!)

source/common/router/router.cc
```diff
@@ -1602,7 +1602,7 @@ void Filter::onUpstreamComplete(UpstreamRequest& upstream_request) {
     }
   }
 
-  upstream_request.removeFromList(upstream_requests_);
+  callbacks_->dispatcher().deferredDelete(upstream_request.removeFromList(upstream_requests_));
```
snowp, line 1605: Is this change to defer delete the request intentional? Seems unrelated to the name change

snowp (changes requested):
Rename lgtm, one question

source/common/router/router.cc
alyssawilk: nope sorry, bad PR split thanks for catching

snowp (approved):
Thanks!

PR #22553, opened 2022-08-04
http: implementing upstream filters

alyssawilk:
Major refactors to the upstream request to allow for upstream filters. This is runtime guarded but the refactor itself is high risk either way, due to the number of code points for data flow.

There's still a bunch of refactoring and clean up to do after this.  Upstream filters can do anything (except refresh route and do internal redirects) that downstream filters can, but have no ability to affect or access the upstream connection.  That will come as the upstream interfaces are added but I figured this was a large enough PR without that extra work.

Risk Level: high
Testing: LOADS of integration testing: all the filter integration tests now run for upstream/downstream.
Docs Changes: n/a (will once it's unhidden)
Release Notes: inline
Runtime guard: envoy_reloadable_features_allow_upstream_filters
Most of https://github.com/envoyproxy/envoy/issues/10455


alyssawilk:
@RyanTheOptimist feel free to ignore this one for tonight - you already did a large PR for me today and I suspect it'll still fail coverage.  I'll work on unit tests in the morning but I'm still super stoked to have this "done"

changelogs/current.yaml
```diff
@@ -59,6 +59,8 @@ new_features:
 - area: http
   change: |
     added default-false ``envoy.reloadable_features.http1_use_balsa_parser`` for experimental BalsaParser.
+  change: |
+    added ``envoy_reloadable_features_allow_upstream_filters`` for experimental upstream filters.
```
RyanTheOptimist, line 63: Should this be `envoy.reloadable_features.allow_upstream_filters` instead of `envoy_reloadable_features_allow_upstream_filters`

source/common/http/conn_manager_impl.h
```diff
@@ -279,7 +279,7 @@ class ConnectionManagerImpl : Logger::Loggable<Logger::Id::http>,
     Router::RouteConstSharedPtr route(const Router::RouteCallback& cb) override;
     void setRoute(Router::RouteConstSharedPtr route) override;
     void clearRouteCache() override;
-    absl::optional<Router::ConfigConstSharedPtr> routeConfig() override;
+    absl::optional<Router::ConfigConstSharedPtr> routeConfig();
```
RyanTheOptimist, line 282: nit: If this is no longer an override, we should probably move it out of the section of methods which all come from a subclass?

source/common/http/filter_manager.cc
```diff
@@ -1335,7 +1335,9 @@ void FilterManager::setBufferLimit(uint32_t new_limit) {
 }
 
 void FilterManager::contextOnContinue(ScopeTrackedObjectStack& tracked_object_stack) {
-  tracked_object_stack.add(connection_);
+  if (connection_) {
```
RyanTheOptimist, line 1338: OOC, is this a bug fix, or an artifact of adding upstream filters?

RyanTheOptimist: Oh, I guess it's because connection_ is now a pointer instead of a ref and it's a pointer that's allowed to be null? I think the Envoy idiom would be OptRef<Network::Connection> instead of a nullable pointer. Would that work?

All of THAT having been said, why is connection_ allowed to be null now? (This is probably obvious)

source/common/http/filter_manager.h
```diff
@@ -1016,6 +1013,7 @@ class FilterManager : public ScopeTrackedObject,
     ActiveStreamDecoderFilter* latest_data_decoding_filter_{};
   };
 
+protected:
```
RyanTheOptimist, line 1016: Does this make state_ protected instead of private? If so, can we avoid that in some way? Protected members are not really style guide kosher.

source/common/router/router.cc
```diff
@@ -805,8 +804,8 @@ Http::FilterTrailersStatus Filter::decodeTrailers(Http::RequestTrailerMap& trail
   // a backoff timer.
   ASSERT(upstream_requests_.size() <= 1);
   downstream_trailers_ = &trailers;
-  for (auto& upstream_request : upstream_requests_) {
-    upstream_request->acceptTrailersFromRouter(trailers);
+  if (!upstream_requests_.empty()) {
```
RyanTheOptimist, line 807: OOC, why this change?

source/common/router/upstream_request.cc
```diff
@@ -41,18 +41,190 @@
 namespace Envoy {
 namespace Router {
 
+// This is the last stop in the filter chain: take the headers and ship them to the codec.
```
RyanTheOptimist, line 44: This is the start of a big block of newly added code. Is it copied from somewhere else, or is this all new?

source/common/router/upstream_request.h
```diff
@@ -30,8 +31,52 @@ class GenericUpstream;
 class GenericConnectionPoolCallbacks;
 class RouterFilterInterface;
 class UpstreamRequest;
+class UpstreamRequestFilterManagerCallbacks;
+class UpstreamFilterManager;
+class CodecFilter;
 
-// The base request for Upstream.
+/* The Upstream request is the base class for forwarding HTTP upstream.
```
RyanTheOptimist, line 38: What an awesome comment! Thanks for writing it.

RyanTheOptimist (commented):
There's a lot going on here, so only a few small comments.

(While I definitely feel encouraged to ignore this one for tonight... this is too exciting to ignore! :>)

source/common/http/conn_manager_impl.h
alyssawilk: good call

source/common/http/filter_manager.cc
alyssawilk: so we can have filters on async client requests (no downstream)

source/common/http/filter_manager.h
alyssawilk: fixed

source/common/router/upstream_request.h
alyssawilk: it'd be kinda cruel to make you figure it out on your own :-)

changelogs/current.yaml
alyssawilk: yep.  though this reminds me: @phlax in this instantiation of the PR the guard is true but even when it's false the scripts require comments.  I don't think we should require docs for guards which are false.

source/common/router/upstream_request.cc
alyssawilk: this is all new

alyssawilk:
cc @paul-r-gall

alyssawilk:
looks like an unrelated fuzz flake
/retest

source/common/router/upstream_request.cc
```diff
@@ -72,6 +244,28 @@ UpstreamRequest::UpstreamRequest(RouterFilterInterface& parent,
   if (cluster_info.has_value()) {
     stream_info_.setUpstreamClusterInfo(*cluster_info);
   }
+  if (!allow_upstream_filters_) {
+    return;
+  }
+
+  // Set up the upstream filter manager.
+  fm_callbacks_ = std::make_unique<UpstreamRequestFilterManagerCallbacks>(*this);
```
RyanTheOptimist, line 252: nit: I would call this filter_manager_callbacks_ to avoid abbreviations, but your call as that's pretty verbose.

source/common/router/upstream_request.cc
```diff
@@ -588,6 +881,14 @@ void UpstreamRequest::encodeBodyAndTrailers() {
   }
 }
 
+UpstreamToDownstream& UpstreamRequest::upstreamToDownstream() {
+  if (allow_upstream_filters_) {
+    return codec_filter_->bridge_;
+  } else {
```
RyanTheOptimist, line 887: nit: early return to avoid the else?

RyanTheOptimist (approved):
A couple of nits, but LGTM otherwise. This is awesome!

RyanTheOptimist (dismissed):

RyanTheOptimist (dismissed):

source/common/router/upstream_request.cc
```diff
@@ -248,12 +465,40 @@ void UpstreamRequest::acceptHeadersFromRouter(bool end_stream) {
   router_sent_end_stream_ = end_stream;
 
   conn_pool_->newStream(this);
```
paul-r-gall, line 467: Why don't we delay constructing a new stream from the connection pool until the upstream filter chain has been traversed? i.e. move this to the end of `acceptHeadersFromRouter`?

alyssawilk: generally many streams are blocking, so kicking off the stream early would be important to many users.  I think we can make it configurable in the long run though.

alyssawilk:
@mattklein123 ping?

alyssawilk:
(working one morning this week from Poland, would love your comments by then)

source/common/router/upstream_request.h
```diff
@@ -184,15 +236,191 @@ class UpstreamRequest : public Logger::Loggable<Logger::Id::router>,
...
+
+// This is the last filter in the upstream filter chain.
+// It takes request headers/body/data from the filter manager and encodes them to the upstream
+// codec. It also registers the CodecBridge with the upstream stream, and takes response
+// headers/body/data from the upstream stream and sends them to the filter manager.
+//
+// Right now, the CodecFilter is implicitly configured as the last filter in the
+// chain and manually created. TODO(alyssawilk) fix this before removing alpha
+// on upstream filters.
+class CodecFilter : public Http::StreamDecoderFilter, public Logger::Loggable<Logger::Id::router> {
```
soulxu, line 371: Just as the above comment said, this filter encodes headers/data/data to the upstream.
` It takes request headers/body/data from the filter manager and encodes them to the upstream codec.`

But why this filter is the subclass of `Http::StreamDecoderFilter`, not the `Http::StreamEncoderFilter`?

We take headers/body/data from the filter manager, then we should call `encodeHeader`, `encodeData`, etc...

source/common/router/upstream_request.cc
```diff
@@ -72,6 +244,28 @@ UpstreamRequest::UpstreamRequest(RouterFilterInterface& parent,
...
+      parent_.callbacks()->connection(), parent_.callbacks()->streamId(),
+      parent_.callbacks()->account(), true, parent_.callbacks()->decoderBufferLimit(),
+      *parent_.cluster(), *this);
+  parent_.cluster()->createFilterChain(*filter_manager_);
+
+  // Similar to how the downstream filter chain ends with the router, the
+  // upstream filter chain ends with the encoder filter, which sends data to the
+  // codec. Create that manually here.
+  Http::FilterFactoryCb cb = [this](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+    codec_filter_ = std::make_shared<CodecFilter>(*this);
```
soulxu, line 264: a question on why we didn't pass the `Http::FilterChainFactoryCallbacks` to the filter instead of UpstreamRequest object, I thought `UpstreamRequest` is Envoy internal implement, shouldn't expose to the filter.

source/common/router/upstream_request.h
```diff
@@ -130,6 +180,9 @@ class UpstreamRequest : public Logger::Loggable<Logger::Id::router>,
   bool hadUpstream() const { return had_upstream_; }
 
 private:
+  friend class UpstreamFilterManager;
+  friend class CodecFilter;
```
soulxu, line 184: Emm...this is related to my question at https://github.com/envoyproxy/envoy/pull/22553/files/e755fee044c1ca990736a3ed605dfb729d6052d4#r945441918

Seems the filter shouldn't reference to the UpstreamRequest, it should use the filter manager callback.

source/common/router/upstream_request.h
soulxu: I guess the reason is we defined `Http::StreamDecoderFilter` for the requesting process, `Http::StreamEncoderFIlter` for the response process. 

But It just feels strange put the word decoder/encoder in the upstream part.

source/common/router/upstream_request.cc
soulxu: I think I get answer by my self https://github.com/envoyproxy/envoy/pull/22553/files/e755fee044c1ca990736a3ed605dfb729d6052d4#diff-889698b89a1023b7d44c1c6a715e0b05b7d92c399e0104998290ba4a605fadfdR73-R77, you plan to clean up this later :)

mattklein123:
> @mattklein123 ping?

Sorry I didn't realize this was waiting on me. I will review today.

source/common/router/upstream_request.cc
```diff
@@ -248,12 +465,40 @@ void UpstreamRequest::acceptHeadersFromRouter(bool end_stream) {
...
+  }
+
+  // Make sure that when we are forwarding CONNECT payload we do not do so until
+  // the upstream has accepted the CONNECT request.
+  if (headers->getMethodValue() == Http::Headers::get().MethodValues.Connect) {
+    paused_for_connect_ = true;
+  }
 
+  filter_manager_->requestHeadersInitialized();
+  //  filter_manager_.setDownstreamRemoteAddress(mutate_result.final_remote_address);
```
mattklein123, line 496: ?

source/common/router/upstream_request.h
```diff
@@ -30,8 +31,52 @@ class GenericUpstream;
...
+ *
+ * This supports both classic forwarding, and upstream filter mode, dictated by
+ * allow_upstream_filters_ which is set via envoy.reloadable_features.allow_upstream_filters.
+ *
+ * In 'classic' mode, the UpstreamRequest requests a new stream, internally
+ * buffers data and trailers in |buffered_request_body_| and |downstream_metadata_map_vector_|
+ * and forwards the router's headers, buffered body, buffered metadata, and
+ * router's trailers to the |upstream_| once the stream is established.
+ *
+ * In 'classic; mode, when the stream is established, the UpstreamRequest
```
mattklein123, line 48:
```suggestion
 * In 'classic' mode, when the stream is established, the UpstreamRequest
```

source/common/router/upstream_request.cc
mattklein123: I think this is fine, but might be worth a comment here on the implication of this. I have a vague worry that this might cause issues for idle timeouts or something if there are filters that pause the request flow for a while. I guess in general all upstream timeouts would be impacted due to filters so probably not a big deal, but might be worth a comment/TODO here either way.

changelogs/current.yaml
```diff
@@ -62,6 +62,8 @@ new_features:
 - area: http
   change: |
     added default-false ``envoy.reloadable_features.http1_use_balsa_parser`` for experimental BalsaParser.
+  change: |
+    added ``envoy.reloadable_features.allow_upstream_filters`` for experimental upstream filters.
```
mattklein123, line 66: Can we get some stuff added to arch overview somewhere? Fine if you want to do this as a follow up if people are eager to land this. I'm thinking we should talk about this in the general HTTP section and maybe have a configuration overview sample for some upstream use case? With all the existing filters I'm sure we can come up with something?

mattklein123 (changes requested):
Thanks this is super cool. Amazing work on such a long requested feature! :tada: 

TBH I mostly skimmed the logic at a high level. As you say, this is super high risk, and I think the chance of catching all logic issues during code review is zero so we might as well ship and get some test coverage and :crossed_fingers:. For this change it might be nice to get someone to canary it first but up to you if you want to go to that effort. 

Thank you!

source/common/router/router.cc
```diff
@@ -805,8 +804,8 @@ Http::FilterTrailersStatus Filter::decodeTrailers(Http::RequestTrailerMap& trail
...
+    upstream_requests_.front()->acceptTrailersFromRouter(trailers);
```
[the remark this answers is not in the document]
alyssawilk: yep - we switched all the other loops but this one and if the connection goes down during handling one stream it's problematic.

source/common/router/upstream_request.cc
alyssawilk: I don't think there's much more code sharing that can be done.
Last filter in the downstream chain is the router, and that's pretty different from the codec.

source/common/router/upstream_request.cc
alyssawilk: so here I would like to move over to the HCM creating the filter chain, which will be shared code.  I've got a TODO for that already insofar as it can't be done until all the direct links between codec filter and upstream request go away.  Got 90% of that refactor done in the next follow-up :-)

source/common/router/upstream_request.h
alyssawilk: Agreed!  the filter ideally wouldn't reference the upstream request, but fixing that requires giving upstream filters access to an interface which allows them upstream-specific information.  I've got a TODO to clean up the codec filter so it can be created by the factory, and to accomplish that I have to remove all direct access, and can remove this friend class.    It's a lot more complicated code so I planned on adding it as a follow up.

source/common/router/upstream_request.h
alyssawilk: because decode filters handle request path.  encode filters _alter_ the response path. This is essentially the same  as the router filter which is a decide only filter.  but yeah agree the encode/decode codec vs encode/decode filter is confusing.

mattklein123 (dismissed):
LGTM! Sweet!

alyssawilk:
@mattklein123 can I get a stamp on the merge resolution?

mattklein123 (approved):

PR #22563, opened 2022-08-04
[v1.22] connection handler: handle the update of listener's ipv4_compat flag correctly

### LINKED ISSUE #22449: Envoy crashes when IPv4-compat listeners are used
Hello everyone,
I recently stumbled across an issue that is leading Envoy (tested on the 1.22.2 release) to crash when IPv4-compat listeners are employed.
I already contacted the envoy security team that confirmed that this crash is not exploitable, therefore I'm opening a public GitHub issue.

The crash itself happens quite often (although not every single time) when I initialize multiple connections targeting an IPv4-compat cluster. Specifically, I noticed that this seems to happen when I do multiple burst of traffic (each with a new set of connections) targeting the same cluster.
I made a debug build of Envoy and captured multiple core-dump, all of them seem to suggest that this is related to the changes introduced in #19716.
I have also verified that reverting the aforementioned commit is preventing the issue from showing up (although this leads to the issue that got fixed by that commit).

I attach the [backtrace](https://github.com/envoyproxy/envoy/files/9212232/backtrace.log) and the [configdump](https://github.com/envoyproxy/envoy/files/9212250/configdump_crash_ipv4compat.log). Please note that the configdump was captured after my Envoy container restarted.

If needed I can share the Envoy debug build + coredump file (although the build itself is 1.6GB).
I think I should also note that I'm using the runtime flag "envoy.reloadable_features.strict_check_on_ipv4_compat".

soulxu:

Commit Message: connection handler: handle the update of listener's ipv4_compat flag correctly
Additional Description:
Currently, when updating a listener's ipv4_compat flag from true to false, the mapping on the corresponding ipv4 address for this listener will be kept. After the old listener is drained, that mapping will be left on the map also. That mapping will reference a destructed listener config, which will lead to envoy crash at any time.

This PR fixes that by cleanup those mapping when found there is an update for ipv4_compat flag.

Risk Level: high
Testing: unittest
Docs Changes: n/a
Release Notes: n/a
Fixes #22449

### DIFF
--- a/api/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
+++ b/api/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
@@ -391,8 +391,7 @@ message FilterStateRule {
 
   // A map of string keys to requirements. The string key is the string value
   // in the FilterState with the name specified in the *name* field above.
-  map<string, JwtRequirement>
-  requires = 3;
+  map<string, JwtRequirement> requires = 3;
 }
 
 // This is the Envoy HTTP filter config for JWT authentication.
--- a/api/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
+++ b/api/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
@@ -391,7 +391,8 @@ message FilterStateRule {
 
   // A map of string keys to requirements. The string key is the string value
   // in the FilterState with the name specified in the *name* field above.
-  map<string, JwtRequirement> requires = 3;
+  map<string, JwtRequirement>
+  requires = 3;
 }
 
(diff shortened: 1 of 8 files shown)


soulxu:
waiting for @LithiumSR do a test on the fix.

phlax:
@soulxu is this only an issue on 1.22?

if not we should probably land on main then backport to affected branches

soulxu:
> @soulxu is this only an issue on 1.22?
> 
> if not we should probably land on main then backport to affected branches

Got it, thanks. This should be fix on main also, I fix the 1.22 first for @LithiumSR do a test, if things go right, I will fix main first

soulxu:
This isn't right fix, try another fix https://github.com/envoyproxy/envoy/pull/22616

PR #22570, opened 2022-08-04
http: moving upstream filters to use an upstream context

alyssawilk:
The one no-op bit of this PR is moving the message validation visitor from the Listener's to the Server's.
Talked with @adisuissa yesterday and we both think this is fine.

Risk Level: medium
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


snowp (approved):
Makes sense to me, thanks!

Issue #15152, opened 2021-02-23
A proposal of high-performance L7 network GoLang extension for Envoy.


wangfakang:
Background
=======
Envoy is an excellent Sidecar in the field of Service Mesh. The use of modern C++ guarantees its high performance, but also increases the difficulty for developers. Although Envoy also supports Lua and WASM to enhance its scalability, there are some shortcomings in some scenarios (as shown in the table below); With the growth of GoLang's cloud native language ecology and popularity, We introduce a proposal that allows Envoy to support GoLang extension capabilities.

     

<p align="center"><strong>Lua vs GoLang vs WASM Extension on Envoy</strong></p>

 Name | Advantage | Disadvantage
-- | -- | --
Lua Extension | easier to develop | fewer SDKs supports and smaller ecosystem
GoLang Extension | richer ecology and more SDKs than Lua, less memory copies and higher performance than WASM | Isolation is not as strong as WASM，go runtime may cause more performance overhead
WASM Extension | Cross-language language support (C/C++/Rust), isolation, security, agility, etc | In the experimental stage, [performance loss is large](https://istio.io/latest/docs/reference/config/proxy_extensions/wasm_telemetry/), Can not re-use the existing golang sdk, need to transform the network io

Design
=======

By adding the Http GoLang extension filter extension to Envoy's HTTP filter, users can implement Envoy's http filter using GoLang via the GoLang L7 extension SDK.

#### Http Golang extension filter
GoLang http extension filter on the Envoy side implemented in C++, this module is used to call the http filter implemented by the developer through GoLang.

#### GoLang L7 extension SDK
GoLang L7 extension SDK exports some CGO APIs for interaction between the Http GoLang extension filter and the GoLang http filter.

The specific architecture is as follows:

![image](https://user-images.githubusercontent.com/8215757/108832601-b3999400-7606-11eb-87f4-ab2f85716def.png)

Best practice
=======

#### Improve Dapr gRPC performence by Envoy L7 GoLang extension

[...]

PR #18823, opened 2021-10-29
api: MetaProtocol Proxy

zhaohuabing:
This PR proposes a MetaProtocol proxy that provides a framework for layer-7 protocols. The common capabilities, including routing, tracing, metrics, logging, etc., will be built into the MetaProtocol proxy. This approach has the potential to significantly lower the barrier to writing new Envoy filters: instead of writing fully functional network filters and duplicating similar codes, now we only need to implement the codec interface. Furthermore, MetaProtocol makes it much easier to build a unified control plane and it could be very helpful for the service mesh ecosystem.

Screen Shot 2021-10-29 at 4 09 49 PM

This PR is a merged efforts of https://github.com/envoyproxy/envoy/pull/18768 and https://github.com/envoyproxy/envoy/issues/18209 

This PR is the APIs of the MetaProtocol Proxy. The implementation will be pushed later after the community agrees on the design and finalizes the API. 

Related issues: 
Common framework for layer-7 protocols #18761
generic proxy support: #18209
Generic Access Log: #13085
Generic Routing & Rds: #9656

Commit Message: Add API for MetaProtocol Proxy
Additional Description:
Design doc: https://docs.google.com/document/d/1y_ALyjGp_H1_TRrrRxZxnqENISqeZI6eLn71FAFsZPk/edit?usp=sharing
Istio blog about MetaProtocol: https://istio.io/latest/blog/2021/aeraki/
This design has been proven working in Aeraki : https://github.com/aeraki-framework/meta-protocol-proxy
Risk Level: low
Testing:
Docs Changes:  
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #22573, opened 2022-08-05
Envoy's L7 Go extension API

wangfakang:
This PR is about [Envoy's Go extension API](https://github.com/envoyproxy/envoy/issues/15152) , which provides a API for extending Envoy's L7 filter via go. Developers can use the Go to implement filter processing of request or response streams, and the final effect is the same as that of Envoy's native filter, which makes it easier to extend Envoy. The implementation will be pushed after the community agrees on the design and completes the API.

The following describes the interaction between Envoy and the filter implemented with the Go extension API:

moe api

@mattklein123 @htuch @lizan @bkgs @tgraf @jrajahalme 
Looking forward to your ideas/comments on this PR!

Commit Message: Add API for Envoy's L7 Go extension
Design doc: https://docs.google.com/document/d/1noApyS0IfmOGmEOHdWk2-BOp0V37zgXMM4MdByr1lQk/edit?usp=sharing
Additional Description: https://github.com/envoyproxy/envoy/issues/15152
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] 
[Optional Deprecated:]


api/envoy/extensions/dso/v3/dso.proto
```diff
@@ -0,0 +1,38 @@
...
+
+option java_package = "io.envoyproxy.envoy.extensions.dso.v3";
+option java_outer_classname = "DsoProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/dso/v3;dsov3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: Dso]
+// Dso config.
```
mattklein123, line 18: This seems unrelated to the Go extension, and is a much larger conversation. Can you remove this for now and we can have a separate conversation about this? I'm not sure why this is required for base level Go plugin support.

Edit: Or maybe it is related? TBH I don't really understand how this is to be used by someone, why it requires SO support, etc. If the idea is that the Go modules are compiled as SOs, there needs to be a large conversation about versions, versions checks, ABI, etc. Additionally, can the SO config just be part of the Go filter config so it's self contained from an API perspective.

Would it be possible to push some more documentation on how someone might actually use this? It's really not clear to me from the API. I don't think we need the full implementation right now but a bit more fleshed out usage guide with perhaps a filter example and how one would compile it would be nice as a point of discussion?

Thank you.

mattklein123 (changes requested):
Thanks for working on this. A high level question to get started.

/wait

api/envoy/extensions/dso/v3/dso.proto
wangfakang: > This seems unrelated to the Go extension, and is a much larger conversation. Can you remove this for now and we can have a separate conversation about this? I'm not sure why this is required for base level Go plugin support.
> 
> Edit: Or maybe it is related? TBH I don't really understand how this is to be used by someone, why it requires SO support, etc. If the idea is that the Go modules are compiled as SOs, there needs to be a large conversation about versions, versions checks, ABI, etc. Additionally, can the SO config just be part of the Go filter config so it's self contained from an API perspective.
> 
> Would it be possible to push some more documentation on how someone might actually use this? It's really not clear to me from the API. I don't think we need the full implementation right now but a bit more fleshed out usage guide with perhaps a filter example and how one would compile it would be nice as a point of discussion?
> 
> Thank you.
-----

First, developer's  Go plugins were integrated into Envoy via dynamic libraries, which can avoided rebuilding Envoy when adding new Go plugins.
By design, the SO config could be part of the Go filter config, but there is a user experience issue that requires the complete SO config for each Go plugin, which seems a bit complicated. Therefore, we separate the SO config from the Go plugin as a general API for SO management. Finally, we can refer to the specific SO via `so_id` in the go plugin config.

Take the following example:

develop go-plugin
====

Envoy developers can implement the interface of `StreamFilter` [API](https://github.com/envoyproxy/envoy/blob/9499dbb0941464528653914cb9eb31a77ec7c1da/contrib/go_plugin_api/filter.go#L5) by the GO language to achieve the effects of Envoy native filter.

Then put the GO plugin source code into the `${OUTPUT}/src/` directory with the name of the plugin for GO plugin builds. The following examples implement `limit` and `auth` GO plugins.
```
tree /home/admin/envoy/go-plugins/src/
├── auth
│   ├── factory.go
│   └── filter.go
└── limit
    ├── factory.go
    └── filter.go

````

build go-plugin
====
go_plugin_generate.sh
``` bash
#!/bin/bash
if [ $#  != 2 ]; then 
   echo "need input the go plugin name"
   exit 1
fi

PLUGINNAME=$1
OUTPUT=/home/admin/envoy/go-plugins/
PLUGINSRCDIR=${OUTPUT}/src/${PLUGINNAME}
go build --buildmode=c-shared  -v -o $OUTPUT/lib${PLUGINNAME}.so $PLUGINSRCDIR
```
Build the Go plugin so by `go_plugin_generate.sh` script,  below example the `liblimit.so` and `liblimit.so` will be generated in the `/home/admin/envoy/go-plugins/` directory.
* go_plugin_generate.sh limit
* go_plugin_generate.sh auth

config go-plugin
====

``` yaml
bootstrap_extensions:
  - name: envoy.bootstrap.dso
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.dso.v3.dso
      so_id: plugin-limit
      so_path: "/home/admin/envoy/go-plugins/liblimit.so"
  - name: envoy.bootstrap.dso
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.dso.v3.dso
      so_id: plugin-auth
      so_path: "/home/admin/envoy/go-plugins/libauth.so"      
static_resources:
  listeners:
    - name: ingress
      address:
        socket_address:
          protocol: TCP
          address: 0.0.0.0
          port_value: 8080
      filter_chains:
        - filters:
            …………
                http_filters:
                  - name: envoy.filters.http.golang
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.golang.v3.Config
                      so_id: plugin-limit
                      plugine_name: limit
                      plugin_config:
                        "@type": type.googleapis.com/envoy.extensions.filters.http.golang.plugins.limit.v3.Config
                        xxx1: xx1
                        xxx2: xx2
                  - name: envoy.filters.http.header_to_metadata
                  - name: envoy.filters.http.golang
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.golang.v3.Config
                      so_id: plugin-auth
                      plugine_name: auth
                      plugin_config:
                        "@type": type.googleapis.com/envoy.extensions.filters.http.golang.plugins.auth.v3.Config
                        xxx1: xx1
                        xxx2: xx2
    - name: egress
      address:
        socket_address:
          protocol: TCP
          address: 0.0.0.0
          port_value: 8081
      filter_chains:
        - filters:
            …………
                http_filters:
                  - name: envoy.filters.http.golang
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.golang.v3.Config
                      so_id: plugin-auth
                      plugine_name: auth
                      plugin_config:
                        "@type": type.googleapis.com/envoy.extensions.filters.http.golang.plugins.auth.v3.Config
                        xxx1: xx1
                        xxx2: xx2
```

wangfakang: ping @mattklein123

mattklein123: If we go with this approach there needs to be a lot more in here in terms of documentation, an example filter, etc. Can you please flesh this out more in terms of this PR?

api/envoy/extensions/dso/v3/dso.proto
```diff
@@ -0,0 +1,150 @@
+syntax = "proto3";
```
mattklein123, line 1: All of this should be in contrib. Since it's going in contrib, for now I would just merge the DSO functionality directly into the Go filter. We can move it out later if needed.

contrib/go_plugin_api/filter.go
```diff
@@ -0,0 +1,70 @@
+package plugin
```
mattklein123, line 1: I would like to see a complete working example, exercised in CI. If you have to include all the code temporarily that's fine, but I still want to understand what the developer experience looks like, and have it be documented. Thank you!

mattklein123 (changes requested):
/wait

wangfakang:
Hi @mattklein123  I have made some updates based on your suggestion, can you help me reopen this PR? thx.

api/envoy/extensions/dso/v3/dso.proto
wangfakang: OK. I have merged the DSO functionality into the Go filter.

contrib/go_plugin_api/filter.go
wangfakang: Now we have implemented a simple version example ` <https://github.com/mosn/envoy-go-filter-samples/blob/master/simple/filter.go>` based on this API that can use go-plugins to affect request processing in Envoy.

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
```diff
@@ -0,0 +1,195 @@
...
+// In the below example, we configured the go plugin 'auth' and 'limit' dynamic libraries into
+// Envoy, which can avoid rebuilding Envoy.
+//
+// * Develop go-plugin
+//
+// We can implement the interface of :ref:`StreamFilter <contrib.go_plugin_api.plugin.StreamFilter>`
+// API by the GO language to achieve the effects of Envoy native filter.
+//
+// The filter based on the `APIs implementation
+// <https://github.com/mosn/envoy-go-filter-samples/blob/master/simple/filter.go>`_ for details.
```
mattklein123, line 32: Sorry what I mean is there should be some test filter in this repo, as well as a test that actually exercises that it works. Can we make that happen? Thank you.

/wait

wangfakang: > Sorry what I mean is there should be some test filter in this repo, as well as a test that actually exercises that it works. Can we make that happen? Thank you.
> 

@mattklein123  Well, at the moment we put this working example in the external [repository](https://github.com/mosn/envoy-go-filter-samples/), because the implementation will continue to iterate, so it is easy to manage, and we can migrate it when it is stable. Here's a quick start doc that can be use go-plugins to affect request processing in Envoy : https://github.com/mosn/envoy-go-filter-samples/blob/master/README.md .

wangfakang: > OK well we won't merge here until there is something stable so feel free to report back when it's ready. Thank you.

@mattklein123  So could you please review the API first? At present, it is only a specific implementation based on the API that is  continuously iterating, and the API itself has been basically completed.  The implementation we will be pushed after completes the API. Thank you.

mattklein123: At a high level it looks OK but it needs a lot more comments, and a working example will be helpful. I will wait until there is something more fleshed out to look at.

wangfakang: > At a high level it looks OK but it needs a lot more comments, and a working example will be helpful. I will wait until there is something more fleshed out to look at.

Hi @mattklein123 Here is a working example  https://github.com/mosn/envoy-go-filter-samples/blob/master/README.md, we can try it in the following way.

Build go plugin
======
```
git clone git@github.com:mosn/envoy-go-filter-samples.git  && cd envoy-go-filter-samples && make build
```

Run go plugin
======
```
Notice: only works on amd64 architecture CPU yet.

docker run -d -p 10000:10000 \
    -v `pwd`/libgolang.so:/usr/local/envoy-go-extension/libgolang.so \
    -v `pwd`/envoy.yaml:/etc/envoy/envoy-golang.yaml \
    mosnio/envoy-go-extension:latest
# 1. it will proxy to httpbin.org/header
# we will see the request header and response header setted by Go.

$ curl 'http://127.0.0.1:10000/headers' -v
*   Trying 127.0.0.1:10000...
* Connected to 127.0.0.1 (127.0.0.1) port 10000 (#0)
> GET /headers HTTP/1.1
> Host: 127.0.0.1:10000
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< date: Sun, 30 Oct 2022 02:02:16 GMT
< content-type: application/json
< content-length: 260
< server: envoy
< access-control-allow-origin: *
< access-control-allow-credentials: true
< x-envoy-upstream-service-time: 1014
< rsp-header-from-go: bar-test
<
{
  "headers": {
    "Accept": "*/*",
    "Host": "httpbin.org",
    "Req-Header-From-Go": "foo-test",
    "User-Agent": "curl/7.79.1",
    "X-Amzn-Trace-Id": "Root=1-635ddb28-7c27d0446b0281c10c3170d8",
    "X-Envoy-Expected-Rq-Timeout-Ms": "15000"
  }
}

# 2. local reply from Go
$ curl 'http://127.0.0.1:10000/localreply'
forbidden from go, path: /localreply
```

phlax:
cc @nguyenptk as potential sandbox

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
mattklein123: The example needs to be self contained and in this repo.

wangfakang: > The example needs to be self contained and in this repo.

@mattklein123 I've pushed the example to this repo in this [commit](https://github.com/envoyproxy/envoy/pull/22573/commits/eeda2e9b6ba11a16175f436e83230462fb6fc933).

wangfakang: Now it can be exercised in CI and verify golang plugin examples succeeded. we can look at this [log](https://dev.azure.com/cncf/envoy/_build/results?buildId=120207&view=logs&j=f1a49366-51c0-540f-0294-f73c2bb311a9&t=7fd036c9-61ec-5b59-3645-e90950dc67a6&l=7479) for details.

mattklein123: I want to see a completely self contained example in this repo, that is exercised in tests. Having the sandbox is also good.

wangfakang: @mattklein123 I have switched the mode of compiling the example to `go vendor` mode and avoid pulling external code during CI tests. And also migrated all the code of the example to the directory `examples/golang/envoy-go-filter-samples/`. Thank you.

mattklein123: 1) Please figure out how to avoid vendoring proto code, etc.
2) Remove all references to MOSN, and simplify the example to something as small as possible.

Thanks

/wait

examples/golang/Dockerfile-proxy
```diff
@@ -0,0 +1,5 @@
+FROM mosnio/envoy-go-extension:beta-v1
```
mattklein123, line 1: Why is this pulling from mosnio? this should pull from envoy.

examples/golang/envoy-go-filter-samples/pkg/http/api.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
mattklein123, line 1: I ton of code that is in here as an example seems like core functionality? Shouldn't most of this be built into the extension and then the extension user just needs to provide the go code?

examples/golang/Dockerfile-proxy
wangfakang: > Why is this pulling from mosnio? this should pull from envoy.

Because we put the implementation of the go extension API in this image, so that users olny need to provide the go code to work, just like this [example](https://github.com/envoyproxy/envoy/tree/57ffc1cab6fb151f47f064e829bd924b0b9737a9/examples/golang/envoy-go-filter-samples/pkg/plugins/simple).  We could be pushed implementation after the community agrees on the design and completes the API. Just like the contribution of [MetaProtocol Proxy API](https://github.com/envoyproxy/envoy/pull/18823).

examples/golang/envoy-go-filter-samples/pkg/http/api.h
wangfakang: > I ton of code that is in here as an example seems like core functionality? Shouldn't most of this be built into the extension and then the extension user just needs to provide the go code?

Yes, the extension users only need to provide the [plugin code](https://github.com/envoyproxy/envoy/tree/57ffc1cab6fb151f47f064e829bd924b0b9737a9/examples/golang/envoy-go-filter-samples/pkg/plugins/simple). This use case is mainly used to verify the correctness of API and the user's experience of writing go plugin.  So the current focus is to complete the design of the go extension API and merge it into the master, and then build a SIG to iterate on the go extension implementation together. Do you think it is okay?

examples/golang/Dockerfile-proxy
mattklein123: Ultimately the code needs to live in this repo. So if you want to iterate on it more let's just hold this PR until you get some more consensus.

/wait

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
wangfakang: @mattklein123   I've already committed the API and implementation to this PR. The sandbox example will be pushed after the PR merge because it depends on the image `envoyproxy/envoy-contrib-dev:latest` which contains the go extension feature.

examples/golang/Dockerfile-proxy
wangfakang: @mattklein123   I've already committed the API and implementation to this PR. The sandbox example will be pushed after the PR merge because it depends on the image `envoyproxy/envoy-contrib-dev:latest` which contains the go extension feature.

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
wangfakang: Resolved it. Thanks.

examples/golang/Dockerfile-proxy
phlax: i think its better to push the sandbox separately (it requires documentation and verification etc) but i should say that the `envoy-contrib-dev` in ci should already have the go extension

wangfakang: > i think its better to push the sandbox separately (it requires documentation and verification etc) but i should say that the `envoy-contrib-dev` in ci should already have the go extension

@phlax  Yes, the sandbox will be pushed in another PR.

wangfakang:
> CC @envoyproxy/dependency-shepherds: Your approval is needed for changes made to `(bazel/.*repos.*\.bzl)|(bazel/dependency_imports\.bzl)|(api/bazel/.*\.bzl)|(.*/requirements\.txt)|(.*\.patch)`. envoyproxy/dependency-shepherds assignee is @mattklein123
> 
> 🐱

@mattklein123 I've already committed the API and implementation to this PR. Now the extension API implementation needs to depende on the repository `google.golang.org/protobuf` and `github.com/cncf/xds/go`, so need your approval for change it `(bazel/dependency_imports.bzl)` .  PTAL, Thanks.

contrib/golang/filters/http/source/cgo.cc
```diff
@@ -0,0 +1,134 @@
...
+  if (str == nullptr) {
+    return "";
+  }
+  auto goStr = reinterpret_cast<GoString*>(str);
+  return absl::string_view(goStr->p, goStr->n); // NOLINT(modernize-return-braced-init-list)
+}
+
+extern "C" {
+
+void moeHandlerWrapper(void* r, std::function<void(std::shared_ptr<Filter>&)> f) {
```
mattklein123, line 23: I'm not sure what `moe` means. I would replace with something more obvious like `envoy_go_filter_` or something like that. Here and elsewhere. This will make the exports more clear.

contrib/golang/filters/http/source/cgo.cc
```diff
@@ -0,0 +1,134 @@
...
+  moeHandlerWrapper(r, [status](std::shared_ptr<Filter>& filter) {
+    filter->continueStatus(static_cast<GolangStatus>(status));
+  });
+}
+
+void moeHttpSendLocalReply(void* r, int response_code, void* body_text, void* headers,
+                           long long int grpc_status, void* details) {
+  moeHandlerWrapper(r, [response_code, body_text, headers, grpc_status,
+                        details](std::shared_ptr<Filter>& filter) {
+    (void)headers;
```
mattklein123, line 41: nit: use UNREFERENCED_PARAMETER or just delete the parameter name from the call signature. Here and elsewhere.

contrib/golang/filters/http/source/common/dso/dso.cc
```diff
@@ -0,0 +1,155 @@
...
+  moe_on_http_destroy_ = nullptr;
+
+  if (handler_ != nullptr) {
+    dlclose(handler_);
+    handler_ = nullptr;
+  }
+}
+
+GoUint64 DsoInstance::moeNewHttpPluginConfig(GoUint64 p0, GoUint64 p1) {
+  // TODO: use ASSERT instead
```
mattklein123, line 128: Let's fix this now. Same elsewhere.

contrib/golang/filters/http/source/config.cc
```diff
@@ -0,0 +1,51 @@
...
+    const envoy::extensions::filters::http::golang::v3alpha::Config& proto_config,
+    const std::string&, Server::Configuration::FactoryContext& context) {
+
+  FilterConfigSharedPtr config = std::make_shared<FilterConfig>(proto_config);
+
+  handler_ = context.lifecycleNotifier().registerCallback(
+      Server::ServerLifecycleNotifier::Stage::PostInit,
+      [lib_id = config->soId(), lib_path = config->soPath()] {
+        Envoy::Dso::DsoInstanceManager::pub(lib_id, lib_path);
+        // ENVOY_LOG(info, "open golang library at postInit: {} {}", lib_id, lib_path);
```
mattklein123, line 23: delete commented out code, here and elsewhere.

contrib/golang/filters/http/source/go/Makefile
```diff
@@ -0,0 +1,20 @@
...
+build:
+	docker run --rm -v $(shell pwd):/go/src/${PROJECT_NAME} -w /go/src/${PROJECT_NAME} ${BUILD_IMAGE} make build-local
+
+sync:
+	cp export.go pkg/api/export.txt
+
+clean:
+	@rm -rf libgolang.*
+
+.PHONY: sync build-local build
```
mattklein123, line 20: What is this for? Shouldn't we be using bazel here?

contrib/golang/filters/http/source/config.cc
```diff
@@ -0,0 +1,51 @@
...
+      });
+
+  return [&context, config](Http::FilterChainFactoryCallbacks& callbacks) {
+    auto filter =
+        std::make_shared<Filter>(context.grpcContext(), config, Filter::global_stream_id_++,
+                                 Dso::DsoInstanceManager::getDsoInstanceByID(config->soId()));
```
mattklein123, line 29: How is error handling done if the DSO doesn't load? That part is not clear to me.

contrib/golang/filters/http/source/go/pkg/api/filter.go
```diff
@@ -0,0 +1,102 @@
...
+    // response stream
+    StreamEncoderFilter
+    // TODO add more
+    /*
+    // stream complete
+    OnStreamComplete()
+    // error log
+    Log(LogType, string)
+    */
+    // destroy filter
```
mattklein123, line 53: delete commented out code here and elsewhere

contrib/golang/filters/http/source/go/pkg/plugins/echo/config.go
```diff
@@ -0,0 +1,15 @@
+package echo
```
mattklein123, line 1: This should be moved to somewhere in the test hierarchy and used for a full integration test.

mattklein123 (changes requested):
Thanks this looks like it's on the right track. The code needs some cleanup and I would also like to see a full integration test running with some sample filter (echo, passthrough, etc.). Thank you.

/wait

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Done.

contrib/golang/filters/http/source/cgo.cc
wangfakang: Done. use `envoy_go_filter_` or `envoyGoFilter` instead of all `moe`

contrib/golang/filters/http/source/cgo.cc
wangfakang: Done.

contrib/golang/filters/http/source/config.cc
wangfakang: Done

contrib/golang/filters/http/source/go/Makefile
wangfakang: Done. use bazel instead of Makefile.

contrib/golang/filters/http/source/go/pkg/plugins/echo/config.go
wangfakang: Done

wangfakang:
> Thanks this looks like it's on the right track. The code needs some cleanup and I would also like to see a full integration test running with some sample filter (echo, passthrough, etc.). Thank you.
> 
> /wait

@mattklein123  I have cleaned up the code according to your review suggestions and added some integration tests for golang extension.  PTAL, Thanks.

mattklein123:
Please never force push. It makes reviews more difficult. Just add commits and merge main. Thanks.

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
```diff
@@ -0,0 +1,203 @@
...
+  // MERGE_VIRTUALHOST_ROUTER_FILTER: pass all configuration into go plugin.
+  // MERGE_VIRTUALHOST_ROUTER: pass Virtual-Host and Router configuration into go plugin.
+  // OVERRIDE: override according to Router > Virtual_host > Filter priority and pass the
+  // configuration to the go plugin.
+  MergePolicy merge_policy = 5 [(validate.rules).enum = {defined_only: true}];
+}
+
+message RouterPlugin {
+  // The extension_plugin_options field is used to provide extension options for plugin.
+  google.protobuf.Struct extension_plugin_options = 1;
```
mattklein123, line 175: What is this? It it just a struct version of the Any below? If so can we get rid of it? Otherwise can you clarify?

contrib/golang/filters/http/source/cgo.cc
```diff
@@ -0,0 +1,135 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Golang {
+
+//
+// These functions may be invoked in another go thread,
+// which means may introduce race between go thread and envoy thread.
+//
```
mattklein123, line 11: I'm pretty sure to avoid problems here you are going to have to post to the dispatcher running the filter on the envoy thread, otherwise I don't see how this can work? Especially in the local reply function below and continuing, etc.. I think you should be using a dispatcher to post and checking the weak lock both before and after the post. This is done in other similar cases.

contrib/golang/filters/http/source/common/dso/dso.cc
```diff
@@ -0,0 +1,161 @@
...
+
+namespace Envoy {
+namespace Dso {
+
+std::map<std::string, DsoInstance*> DsoInstanceManager::dso_map_ = {};
+absl::Mutex DsoInstanceManager::mutex_ = {};
+
+bool DsoInstanceManager::pub(std::string dso_id, std::string dso_name) {
+  if (getDsoInstanceByID(dso_id) != nullptr) {
+    ENVOY_LOG_MISC(warn, "pub {} {} dso instance failed: already pub.", dso_id, dso_name);
```
mattklein123, line 13: What does this mean? Should this be an assert?

contrib/golang/filters/http/source/common/dso/dso.cc
```diff
@@ -0,0 +1,161 @@
...
+    return true;
+  }
+
+  absl::WriterMutexLock lock(&DsoInstanceManager::mutex_);
+
+  DsoInstance* dso = new DsoInstance(dso_name);
```
mattklein123, line 19: use unique_ptr here and elsewhere. Please no raw memory management.

contrib/golang/filters/http/source/common/dso/dso.h
```diff
@@ -0,0 +1,59 @@
...
+
+  void (*envoy_go_filter_on_http_destroy_)(httpRequest* p0, GoUint64 p1) = {nullptr};
+};
+
+class DsoInstanceManager {
+public:
+  static bool pub(std::string dso_id, std::string dso_name);
+  static bool unpub(std::string dso_id);
+  static DsoInstance* getDsoInstanceByID(std::string dso_id);
+  static std::string show();
```
mattklein123, line 51: Please add more comments. I don't know what this all means.

mattklein123 (changes requested):
Flushing some more comments, thanks.

/wait

contrib/golang/filters/http/source/cgo.cc
wangfakang: Yes, it is very important here, as you said, by using a envoy's dispatcher to post to make the some process only be executed in the envoy thread. As below code
https://github.com/envoyproxy/envoy/blob/a2f78efdf32feb169568f448acbbe05caf96dd3a/contrib/golang/filters/http/source/golang_filter.cc#L579-L596

mattklein123: OK I see you do it elsewhere. Then please update this comment to be more clear that you cover this in the filter.

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
wangfakang: This is the extension configuration option reserved for the go extension itself. To avoid changing the API when customizing some configurations. Such as the `disabled` option extended below
https://github.com/envoyproxy/envoy/blob/a2f78efdf32feb169568f448acbbe05caf96dd3a/api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto#L188-L197

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Entering this branch indicates that two go extensions use the same `DSO`, for example, different listener blocks. I will adjust the log level and content. Thanks.

wangfakang:
> Please never force push. It makes reviews more difficult. Just add commits and merge main. Thanks.

Ok. It was to trigger GitHub Actions and https://github.com/envoyproxy/envoy/issues/24156 it looks good for me.

contrib/golang/filters/http/source/common/dso/dso.h
wangfakang: Done

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Done

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
wangfakang: add example for extension_plugin_options.

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Done

contrib/golang/filters/http/source/cgo.cc
wangfakang: Done.

api/contrib/envoy/extensions/filters/http/golang/v3alpha/golang.proto
```diff
@@ -0,0 +1,217 @@
...
+  //     envoy.filters.http.golang:
+  //       "@type": type.googleapis.com/envoy.extensions.filters.http.golang.v3alpha.ConfigsPerRoute
+  //       plugins_config:
+  //         plugin1:
+  //          extension_plugin_options:
+  //            disabled: true
+  //          config:
+  //            "@type": type.googleapis.com/golang.http.plugin1
+  //            xxx: xxx
+  google.protobuf.Struct extension_plugin_options = 1;
```
mattklein123, line 189: Let's please just make this fully typed. It's much easier to document and understand. As they API is alpha we can always change it later. Thank you.

/wait

wangfakang: Done.

wangfakang:
> Flushing some more comments, thanks.
> 
> /wait

@mattklein123  Done. PTAL, Thanks.

contrib/golang/filters/http/source/common/dso/dso.h
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Get the full amount of dynamic library information.
+   * @return all so ids separated by commas.
+   */
+  static std::string show();
+
+private:
+  static absl::Mutex mutex_;
+  static std::map<std::string, DsoInstancePtr> dso_map_;
```
mattklein123, line 81: We don't use statics like this due to initialization fiasco. See CONSTRUCT_ON_FIRST_USE. Same elsewhere.

contrib/golang/filters/http/source/common/dso/dso.h
```diff
@@ -0,0 +1,85 @@
...
+
+class DsoInstanceManager {
+public:
+  /**
+   * Load the go plugin dynamic library.
+   * @param dso_id is unique ID for dynamic library.
+   * @param dso_name used to specify the absolute path of the dynamic library.
+   * @return false if pub are invalid. Otherwise, return true.
+   */
+  static bool pub(std::string dso_id, std::string dso_name);
```
mattklein123, line 57:
```suggestion
  static bool load(std::string dso_id, std::string dso_name);
```

contrib/golang/filters/http/source/common/dso/dso.h
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Unload the go plugin dynamic library.
+   * @param dso_id is unique ID for dynamic library.
+   * @return false if unpub are invalid. Otherwise, return true.
+   */
+  static bool unpub(std::string dso_id);
```
mattklein123, line 64:
```suggestion
  static bool unload(std::string dso_id);
```

contrib/golang/filters/http/source/go/pkg/http/filtermanager.go
```diff
@@ -0,0 +1,57 @@
...
+
+// TODO merge it to api.HttpFilterConfigFactory
+func RegisterHttpFilterConfigParser(parser api.StreamFilterConfigParser) {
+    httpFilterConfigParser = parser
+}
+
+func getOrCreateHttpFilterFactory(name string, configId uint64) api.StreamFilterFactory {
+    config, ok := configCache.Load(configId)
+    if !ok {
+        // TODO: panic
```
mattklein123, line 43: ? Please do something here and in other similar places.

contrib/golang/filters/http/source/golang_filter.h
```diff
@@ -0,0 +1,258 @@
...
+  uint64_t cost_time_encode_{0};
+  uint64_t cost_time_mem_{0};
+  uint64_t stream_id_{0};
+
+  httpRequestInternal* req_{nullptr};
+
+  // lock for has_destroyed_,
+  // to avoid race between envoy c thread and go thread (when calling back from go).
+  // it should also be okay without this lock in most cases, just for extreme case.
+  Thread::MutexBasicLockable mutex_{};
```
mattklein123, line 219: Document which variables are covered by the lock via ABSL_EXCLUSIVE_LOCKS_REQUIRED and similar. I believe various headers also have to be covered by this due to how you are not posting all modifications for headers gets/sets/etc.

mattklein123 (changes requested):
Flushing some more comments, thanks.

/wait

contrib/golang/filters/http/source/common/dso/dso.h
wangfakang: It's initialized in `dso.cc`, as below. The CONSTRUCT_ON_FIRST_USE looks more rigorous and I'll do it.
https://github.com/envoyproxy/envoy/blob/66258bbafd5f68ceff2a05e475b4b4f19352a29c/contrib/golang/filters/http/source/common/dso/dso.cc#L8-L10

wangfakang: Done

contrib/golang/filters/http/source/go/pkg/http/filtermanager.go
wangfakang: Done

contrib/golang/filters/http/source/golang_filter.h
wangfakang: Done.

contrib/golang/filters/http/source/common/dso/dso.h
```diff
@@ -0,0 +1,91 @@
...
+
+private:
+  static absl::Mutex& getMutex() { MUTABLE_CONSTRUCT_ON_FIRST_USE(absl::Mutex); }
+
+  using DsoTypeMap = std::map<std::string, DsoInstancePtr>;
+  static DsoTypeMap& getDsoMap() {
+    MUTABLE_CONSTRUCT_ON_FIRST_USE(DsoTypeMap, {
+                                                   {"", nullptr},
+                                               });
+  }
```
mattklein123, line 87: Instead of having each of these separate, have a single struct which you construct on first use. Then, you can use lock thread annotations to make sure the lock is acquired when you are using the map.

contrib/golang/filters/http/source/golang_filter.h
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  void continueStatus(GolangStatus status);
+
+  void sendLocalReply(Http::Code response_code, absl::string_view body_text,
+                      std::function<void(Http::ResponseHeaderMap& headers)> modify_headers,
+                      Grpc::Status::GrpcStatus grpc_status, absl::string_view details);
+
+  absl::optional<absl::string_view> getHeader(absl::string_view key);
+  void copyHeaders(GoString* go_strs, char* go_buf) ABSL_LOCKS_EXCLUDED(mutex_);
```
mattklein123, line 163: You do not want to be using `ABSL_LOCKS_EXCLUDED`. Remove all of this. You want to be using `ABSL_EXCLUSIVE_LOCKS_REQUIRED` on the variables that need to be locked, and then on methods where the lock is already acquired if applicable. Same elsewhere.

contrib/golang/filters/http/source/processor_state.h
```diff
@@ -0,0 +1,264 @@
...
+  StopAndBufferWatermark,
+  StopNoBuffer,
+};
+
+class ProcessorState : public Logger::Loggable<Logger::Id::http> {
+public:
+  explicit ProcessorState(Filter& filter) : filter_(filter) {}
+  ProcessorState(const ProcessorState&) = delete;
+  virtual ~ProcessorState() = default;
+  ProcessorState& operator=(const ProcessorState&) = delete;
```
mattklein123, line 98: nit: you can derive from NonCopyable.

mattklein123 (changes requested):
/wait

contrib/golang/filters/http/source/common/dso/dso.h
wangfakang: @mattklein123  Because most of the scenarios are reading the `DSO` map, so we separate this and use `ReaderMutexLock` and `WriterMutexLock`  to improve performance.

mattklein123: You can still use lock annotations, if you put them in a struct and have the annotations refer to the embedded mutex.

contrib/golang/filters/http/source/golang_filter.h
wangfakang: @mattklein123 You mean to delete the `lock` in the `copyHeaders` and add `ABSL_EXCLUSIVE_LOCKS_REQUIRED` annotation to it. Finally use lock before calling this function?
https://github.com/envoyproxy/envoy/blob/c063d61d52c81013e4620b4ddfedf7f196289de8/contrib/golang/filters/http/source/golang_filter.cc#L658-L659

contrib/golang/filters/http/source/processor_state.h
wangfakang: Done.

contrib/golang/filters/http/source/golang_filter.h
wangfakang: Done. add ABSL_EXCLUSIVE_LOCKS_REQUIRED for some methods.

contrib/golang/filters/http/source/common/dso/dso.h
wangfakang: Done. use a struct to encapsulate `dso` map and mutex.

contrib/golang/filters/http/source/golang_filter.h
```diff
@@ -193,6 +192,21 @@ class Filter : public Http::StreamFilter,
                               std::function<void(Http::ResponseHeaderMap& headers)> modify_headers,
                               Grpc::Status::GrpcStatus grpc_status, absl::string_view details);
 
+  absl::optional<absl::string_view> getHeaderInternal(absl::string_view key)
+      ABSL_EXCLUSIVE_LOCKS_REQUIRED(mutex_);
```
mattklein123, line 196: Again, it's most important to have the annotations on the data members themselves. Please do that.

/wait

wangfakang: > Again, it's most important to have the annotations on the data members themselves. Please do that.
> 
> /wait
@mattklein123 Because the member does not need to be locked in some scenarios,  the `ABSL_GUARDED_BY` annotation cannot be added to the member. So use `ABSL_EXCLUSIVE_LOCKS_REQUIRED` to annotate in concurrent methods. Such as below

https://github.com/envoyproxy/envoy/blob/1ca4641ccbb7edf0dec3da7ecf22d8400e52f5e5/contrib/golang/filters/http/source/golang_filter.cc#L586-L590

mattklein123: For right now instead of trying to over-optimize I would just lock in all cases with annotations. It's safer and after things are proven to be stable we can relax later if needed.

/wait

wangfakang:
@mattklein123 PTAL. Thsnks.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,903 @@
...
+}
+
+Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers, bool end_stream) {
+  ProcessorState& state = decoding_state_;
+
+  ENVOY_LOG(debug, "golang filter decodeHeaders, state: {}, phase: {}, end_stream: {}",
+            state.stateStr(), state.phaseStr(), end_stream);
+
+  if (dynamic_lib_ == nullptr) {
+    ENVOY_LOG(error, "dynamic_lib_ is nullPtr, maybe the instance already unpub.");
```
mattklein123, line 49: How can this happen? Shouldn't you fail to load the filter and once it's loaded the DSO should remain loaded? I think this should all be removed.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,903 @@
...
+    // get the state before changing it.
+    bool inGo = state.isProcessingInGo();
+
+    /*
+    // it's safe to reset phase_ and state_, since they are read/write in safe thread.
+    ENVOY_LOG(debug, "golang filter phase grow to EncodeHeader and state grow to WaitHeader since "
+                     "enter encodeHeaders early");
+    phase_ = Phase::EncodeHeader;
+    state_ = FilterState::WaitHeader;
+    */
```
mattklein123, line 140: Remove  commented out code. Here and elsewhere.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,903 @@
...
+    auto status = dynamic_lib_->envoyGoFilterOnHttpHeader(req_, end_stream ? 1 : 0, headers.size(),
+                                                          headers.byteSize());
+    return static_cast<GolangStatus>(status);
+
+  } catch (const EnvoyException& e) {
+    ENVOY_LOG(error, "golang filter doHeadersGo catch: {}.", e.what());
+
+  } catch (...) {
+    ENVOY_LOG(error, "golang filter doHeadersGo catch unknown exception.");
+  }
```
mattklein123, line 304: Please remove try/catch handling from data plane code. Why/how can this code throw? Also please never catch all exceptions. Let's just remove exception handling entirely from all of this.

If somehow go panics are translated to exceptions, let's have the go code catch/translate to error returns.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,903 @@
...
+  return result[0]->value().getStringView();
+}
+
+void copyHeaderMapToGo(Http::HeaderMap& m, GoString* go_strs, char* go_buf) {
+  auto i = 0;
+  m.iterate([&i, &go_strs, &go_buf](const Http::HeaderEntry& header) -> Http::HeaderMap::Iterate {
+    auto key = std::string(header.key().getStringView());
+    auto value = std::string(header.value().getStringView());
+
+    // std::cout << "idx: " << i << ", key: " << key << ", value: " << value << std::endl;
```
mattklein123, line 654: Remove all commented out code.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,903 @@
...
+
+    auto len = key.length();
+    go_strs[i].n = len;
+    go_strs[i].p = go_buf;
+    memcpy(go_buf, key.data(), len); // NOLINT(safe-memcpy)
```
mattklein123, line 659: Why are you not use safe-memcpy here and elsewhere?

mattklein123 (changes requested):
/wait

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done. add some comments for it. Because `go_buf` is the heap memory of go, and the length is the total length of all keys and values in the header. So here use `memcpy` is safe.

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done. Thanks.

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done. yeah, merge the DSO to filter self and throw error at the config parsing phase when load failed. Thanks.

wangfakang:
@mattklein123 PTAL. Thanks

contrib/golang/filters/http/source/common/dso/dso.cc
```diff
@@ -0,0 +1,159 @@
...
+                   dso_name, dlerror());
+  }
+}
+
+DsoInstance::~DsoInstance() {
+  envoy_go_filter_new_http_plugin_config_ = nullptr;
+  envoy_go_filter_merge_http_plugin_config_ = nullptr;
+  envoy_go_filter_on_http_header_ = nullptr;
+  envoy_go_filter_on_http_data_ = nullptr;
+  envoy_go_filter_on_http_destroy_ = nullptr;
```
mattklein123, line 123: Nit: setting to null in the destructor doesn't do anything. Same for line 127.

contrib/golang/filters/http/source/common/dso/dso.cc
```diff
@@ -0,0 +1,159 @@
...
+
+  auto func = dlsym(handler_, "envoyGoFilterNewHttpPluginConfig");
+  if (func) {
+    envoy_go_filter_new_http_plugin_config_ =
+        reinterpret_cast<GoUint64 (*)(GoUint64 p0, GoUint64 p1)>(func);
+  } else {
+    loaded_ = false;
+    ENVOY_LOG_MISC(error, "lib: {}, cannot find symbol: envoyGoFilterNewHttpPluginConfig, err: {}",
+                   dso_name, dlerror());
+  }
```
mattklein123, line 72: nit: I would add a helper for this repetition.

contrib/golang/filters/http/source/config.cc
```diff
@@ -0,0 +1,55 @@
...
+Http::FilterFactoryCb GolangFilterConfig::createFilterFactoryFromProtoTyped(
+    const envoy::extensions::filters::http::golang::v3alpha::Config& proto_config,
+    const std::string&, Server::Configuration::FactoryContext& context) {
+
+  FilterConfigSharedPtr config = std::make_shared<FilterConfig>(proto_config);
+
+  ENVOY_LOG_MISC(info, "load golang library at parse config: {} {}", config->soId(),
+                 config->soPath());
+
+  auto res = Envoy::Dso::DsoInstanceManager::load(config->soId(), config->soPath());
```
mattklein123, line 24: Shouldn't we be unloading when the last reference to this by a filter is dropped otherwise we might leak open handles? Maybe that doesn't matter in the first version but I would at least add a TODO/comment about this.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,821 @@
...
+
+  // NP: may enter encodeHeaders in any phase & any state_,
+  // since other filters or filtermanager could call encodeHeaders or sendLocalReply in any time.
+  // eg. filtermanager may invoke sendLocalReply, when scheme is invalid,
+  // with "Sending local reply with details // http1.invalid_scheme" details.
+  if (state.state() != FilterState::Done) {
+    ENVOY_LOG(warn,
+              "golang filter enter encodeHeaders early, maybe sendLocalReply or encodeHeaders "
+              "happened, current state: {}, phase: {}",
+              state.stateStr(), state.phaseStr());
```
mattklein123, line 99: What does this warning mean? It's not clear. In general we shouldn't warn in the data path as this is going to log spam. Should this be some kind of ENVOY_BUG or ASSERT or something?

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,821 @@
...
+
+    ENVOY_LOG(debug, "golang filter drain data buffer since enter encodeHeaders early");
+    // NP: is safe to overwrite it since go code won't read it directly
+    // need drain buffer to enable read when it's high watermark
+    state.drainBufferData();
+
+    // get the state before changing it.
+    bool inGo = state.isProcessingInGo();
```
mattklein123, line 107: nit: snake_cased here and elsewhere.

```suggestion
    bool in_go = state.isProcessingInGo();
```

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,821 @@
...
+    Grpc::Status::GrpcStatus grpc_status, absl::string_view details) {
+  ENVOY_LOG(debug, "sendLocalReply Internal, response code: {}", int(response_code));
+
+  ProcessorState& state = getProcessorState();
+
+  if (local_reply_waiting_go_) {
+    ENVOY_LOG(warn,
+              "other filter already invoked sendLocalReply or encodeHeaders, ignoring the local "
+              "reply from go, code: {}, body: {}, details: {}",
+              int(response_code), body_text, details);
```
mattklein123, line 491: Same question. Audit all warns in the filter code.

contrib/golang/filters/http/test/golang_filter_test.cc
```diff
@@ -0,0 +1,171 @@
...
+namespace HttpFilters {
+namespace Golang {
+namespace {
+
+class TestFilter : public Filter {
+public:
+  using Filter::Filter;
+
+  // TODO: for mock methods
+  // MOCK_METHOD;
```
mattklein123, line 44: ? remove. Audit for all commented out code.

mattklein123 (changes requested):
Thanks this is looking pretty close. 2 other comments:
1) Add a release note in current.yaml and reference new docs.
2) Contrib extensions do not have ASAN and TSAN run on them by default. Manually run all the new tests with ASAN and TSAN and post the results. Thank you.

/wait

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done, Updae the log level to info.

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Done. Thanks.

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Done. Thanks for your comments.

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done, Thanks.

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done.

contrib/golang/filters/http/test/golang_filter_test.cc
wangfakang: Done. Thanks for your comments.

contrib/golang/filters/http/source/config.cc
wangfakang: Thanks for your comments. Because the type of `DsoInstancePtr` is an `shared_ptr` and it will be destructed when there is no reference.  As below

https://github.com/envoyproxy/envoy/blob/00602640553c04f82cf6b8f78cf96f2563bc9479/contrib/golang/filters/http/source/common/dso/dso.cc#L82-L84

wangfakang:
> Thanks this is looking pretty close. 2 other comments:
> 
> 1. Add a release note in current.yaml and reference new docs.
> 2. Contrib extensions do not have ASAN and TSAN run on them by default. Manually run all the new tests with ASAN and TSAN and post the results. Thank you.
> 
> /wait


@mattklein123 Thank you very much for your careful review, which has taught me a lot. 
I used `clang-asan` and `clang-tsan` to successfully run all the test cases of the go extension. Some results are as below.

ASAN
======
```
$bazel test -c fastbuild //contrib/golang/filters/http/test/...  --config=clang-asan --test_verbose_timeout_warnings   --test_env=ENVOY_IP_TEST_VERSIONS=v4only  --verbose_failures  --test_arg="-l debug"  --define=wasm=disabled
DEBUG: /home/fakang.wfk/.cache/bazel/_bazel_fakang.wfk/05ab2d40a050525b20b2187bf5222296/external/rules_python/python/pip.bzl:46:10: pip_install is deprecated. Please switch to pip_parse. pip_install will be removed in a future release.
INFO: Analyzed 9 targets (0 packages loaded, 0 targets configured).
INFO: Found 5 targets and 4 test targets...
INFO: From Executing genrule @io_opentracing_cpp//:generate_version_h:
-- The C compiler identification is Clang 11.0.0
-- The CXX compiler identification is Clang 11.0.0
-- Check for working C compiler: /usr/local/bin/clang
-- Check for working C compiler: /usr/local/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/local/bin/clang++
-- Check for working CXX compiler: /usr/local/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- clang-tidy not found.
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/tmp.9MvmbsVAFz
-- Cache values
BUILD_DYNAMIC_LOADING:BOOL=ON
BUILD_MOCKTRACER:BOOL=OFF
BUILD_SHARED_LIBS:BOOL=ON
BUILD_STATIC_LIBS:BOOL=ON
BUILD_TESTING:BOOL=OFF
CLANG_TIDY_EXE:FILEPATH=CLANG_TIDY_EXE-NOTFOUND
CMAKE_BUILD_TYPE:STRING=
CMAKE_INSTALL_PREFIX:PATH=/usr/local
ENABLE_LINTING:BOOL=ON
INFO: From Generating proto_library @com_google_googleapis//google/devtools/cloudtrace/v2:cloudtrace_proto:
google/devtools/cloudtrace/v2/trace.proto:19:1: warning: Import google/api/annotations.proto is unused.
google/devtools/cloudtrace/v2/tracing.proto:26:1: warning: Import google/protobuf/timestamp.proto is unused.
INFO: From Compiling eval/public/structs/protobuf_descriptor_type_provider.cc:
In file included from external/com_google_cel_cpp/eval/public/structs/protobuf_descriptor_type_provider.cc:15:
external/com_google_cel_cpp/eval/public/structs/protobuf_descriptor_type_provider.h:53:27: warning: private field 'unboxing_option_' is not used [-Wunused-private-field]
  ProtoWrapperTypeOptions unboxing_option_;
                          ^
1 warning generated.
INFO: From Action external/opencensus_proto/opencensus/proto/agent/trace/v1/trace_service.grpc.pb.h:
bazel-out/k8-fastbuild/bin/external/opencensus_proto/external/opencensus_proto: warning: directory does not exist.
INFO: From Action external/com_google_googleapis/google/devtools/cloudtrace/v2/trace.grpc.pb.h:
bazel-out/k8-fastbuild/bin/external/com_google_googleapis/external/com_google_googleapis: warning: directory does not exist.
google/devtools/cloudtrace/v2/trace.proto:19:1: warning: Import google/api/annotations.proto is unused.
google/devtools/cloudtrace/v2/tracing.proto:26:1: warning: Import google/protobuf/timestamp.proto is unused.
INFO: Elapsed time: 2716.501s, Critical Path: 351.98s
INFO: 4591 processes: 1559 internal, 3032 linux-sandbox.
INFO: Build completed successfully, 4591 total actions

Executed 4 out of 4 tests: 4 tests pass.
```
TSAN
======
```
$bazel test -c fastbuild //contrib/golang/filters/http/test/...   --config=clang-tsan --test_verbose_timeout_warnings   --test_env=ENVOY_IP_TEST_VERSIONS=v4only --verbose_failures  --config=gcc  --test_arg="-l debug"  --define=wasm=disabled --build_tests_only
Starting local Bazel server and connecting to it...
WARNING: option '--build_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
WARNING: option '--test_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
DEBUG: /home/fakang.wfk/.cache/bazel/_bazel_fakang.wfk/05ab2d40a050525b20b2187bf5222296/external/rules_python/python/pip.bzl:46:10: pip_install is deprecated. Please switch to pip_parse. pip_install will be removed in a future release.
WARNING: option '--build_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
WARNING: option '--test_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
INFO: Analyzed 4 targets (930 packages loaded, 32893 targets configured).
INFO: Found 4 test targets...
INFO: Deleting stale sandbox base /home/fakang.wfk/.cache/bazel/_bazel_fakang.wfk/05ab2d40a050525b20b2187bf5222296/sandbox
INFO: Elapsed time: 566.694s, Critical Path: 471.52s
INFO: 22 processes: 1 internal, 21 linux-sandbox.
INFO: Build completed successfully, 22 total actions

Executed 3 out of 4 tests: 4 tests pass.
```

contrib/golang/filters/http/source/common/dso/dso.cc
```diff
@@ -61,70 +61,27 @@ DsoInstance::DsoInstance(const std::string dso_name) : dso_name_(dso_name) {
...
-  }
+#define _DLSYM(fn, name)                                                                           \
+  do {                                                                                             \
+    fn = reinterpret_cast<decltype(fn)>(dlsym(handler_, name));                                    \
+    if (!fn) {                                                                                     \
+      loaded_ = false;                                                                             \
+      ENVOY_LOG_MISC(error, "lib: {}, cannot find symbol: {}, err: {}", dso_name, name,            \
+                     dlerror());                                                                   \
+    }                                                                                              \
+  } while (0)
```
mattklein123, line 72: This can just be a regular function. It doesn't need to be a macro.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -93,7 +93,7 @@ Http::FilterHeadersStatus Filter::encodeHeaders(Http::ResponseHeaderMap& headers
   // eg. filtermanager may invoke sendLocalReply, when scheme is invalid,
   // with "Sending local reply with details // http1.invalid_scheme" details.
   if (state.state() != FilterState::Done) {
-    ENVOY_LOG(warn,
+    ENVOY_LOG(info,
```
mattklein123, line 96: info level logs are logs that show up for all users. This should be debug or removed. Same elsewhere.

contrib/golang/filters/http/source/config.cc
mattklein123: The problem is you don't capture res, so I don't understand how this stays alive?

mattklein123 (changes requested):
/wait

contrib/golang/filters/http/source/config.cc
wangfakang: > The problem is you don't capture res, so I don't understand how this stays alive?

@mattklein123. if the `load`  return failed then the DSO instance don't store the `dsoStore` map and close the handle in the destructor. otherwise store the DSO instance shared ptr to the `dsoStore` map for keepalive. As below

https://github.com/envoyproxy/envoy/blob/00602640553c04f82cf6b8f78cf96f2563bc9479/contrib/golang/filters/http/source/common/dso/dso.cc#L16-L21

mattklein123: Yes but when does it ever get deleted? That's my point.

wangfakang: > Yes but when does it ever get deleted? That's my point.

@mattklein123  The type of `DsoInstancePtr` is `std::shared_ptr<DsoInstance>` and it will be automatically destructed/deleted when there is no reference. 

https://github.com/envoyproxy/envoy/blob/00602640553c04f82cf6b8f78cf96f2563bc9479/contrib/golang/filters/http/source/common/dso/dso.h#L47

mattklein123: You store it in a static. When is it ever deleted? Also now that I look I don't see unload() ever called. AFAICT you are leaking DSO loads if a filter gets loaded and unloaded.

wangfakang: Yes. I see what you mean and i'll add some comments for it.

wangfakang: Done. thanks.

contrib/golang/filters/http/source/common/dso/dso.cc
wangfakang: Done. Use a function instead of a macro.
Thanks for your comments.

contrib/golang/filters/http/source/golang_filter.cc
wangfakang: Done, update the logleve to debug.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -0,0 +1,821 @@
...
+  }
+
+  merged_config_id_ = dlib->envoyGoFilterMergeHttpPluginConfig(parent_id, config_id_);
+  if (merged_config_id_ == 0) {
+    ENVOY_LOG(error, "invalid golang plugin config");
+  }
+  ENVOY_LOG(debug, "golang filter merge plugin config, from {} + {} to {}", parent_id, config_id_,
+            merged_config_id_);
+  return merged_config_id_;
+};
```
mattklein123, line 810: All of the errors in this function should I think be caught at config load time so should be throwing exceptions. I'm not sure what happens in the error path here. Please verify this and add appropriate config loading tests that fail.

mattklein123 (changes requested):
/wait

wangfakang:
> > Thanks this is looking pretty close. 2 other comments:
> > 
> > 1. Add a release note in current.yaml and reference new docs.
> > 2. Contrib extensions do not have ASAN and TSAN run on them by default. Manually run all the new tests with ASAN and TSAN and post the results. Thank you.
> > 
> > /wait
> 
> @mattklein123 Thank you very much for your careful review, which has taught me a lot. I used `clang-asan` and `clang-tsan` to successfully run all the test cases of the go extension. Some results are as below.
> 
> # ASAN
> ```
> $bazel test -c fastbuild //contrib/golang/filters/http/test/...  --config=clang-asan --test_verbose_timeout_warnings   --test_env=ENVOY_IP_TEST_VERSIONS=v4only  --verbose_failures  --test_arg="-l debug"  --define=wasm=disabled
> DEBUG: /home/fakang.wfk/.cache/bazel/_bazel_fakang.wfk/05ab2d40a050525b20b2187bf5222296/external/rules_python/python/pip.bzl:46:10: pip_install is deprecated. Please switch to pip_parse. pip_install will be removed in a future release.
> INFO: Analyzed 9 targets (0 packages loaded, 0 targets configured).
> INFO: Found 5 targets and 4 test targets...
> INFO: From Executing genrule @io_opentracing_cpp//:generate_version_h:
> -- The C compiler identification is Clang 11.0.0
> -- The CXX compiler identification is Clang 11.0.0
> -- Check for working C compiler: /usr/local/bin/clang
> -- Check for working C compiler: /usr/local/bin/clang -- works
> -- Detecting C compiler ABI info
> -- Detecting C compiler ABI info - done
> -- Detecting C compile features
> -- Detecting C compile features - done
> -- Check for working CXX compiler: /usr/local/bin/clang++
> -- Check for working CXX compiler: /usr/local/bin/clang++ -- works
> -- Detecting CXX compiler ABI info
> -- Detecting CXX compiler ABI info - done
> -- Detecting CXX compile features
> -- Detecting CXX compile features - done
> -- clang-tidy not found.
> -- Configuring done
> -- Generating done
> -- Build files have been written to: /tmp/tmp.9MvmbsVAFz
> -- Cache values
> BUILD_DYNAMIC_LOADING:BOOL=ON
> BUILD_MOCKTRACER:BOOL=OFF
> BUILD_SHARED_LIBS:BOOL=ON
> BUILD_STATIC_LIBS:BOOL=ON
> BUILD_TESTING:BOOL=OFF
> CLANG_TIDY_EXE:FILEPATH=CLANG_TIDY_EXE-NOTFOUND
> CMAKE_BUILD_TYPE:STRING=
> CMAKE_INSTALL_PREFIX:PATH=/usr/local
> ENABLE_LINTING:BOOL=ON
> INFO: From Generating proto_library @com_google_googleapis//google/devtools/cloudtrace/v2:cloudtrace_proto:
> google/devtools/cloudtrace/v2/trace.proto:19:1: warning: Import google/api/annotations.proto is unused.
> google/devtools/cloudtrace/v2/tracing.proto:26:1: warning: Import google/protobuf/timestamp.proto is unused.
> INFO: From Compiling eval/public/structs/protobuf_descriptor_type_provider.cc:
> In file included from external/com_google_cel_cpp/eval/public/structs/protobuf_descriptor_type_provider.cc:15:
> external/com_google_cel_cpp/eval/public/structs/protobuf_descriptor_type_provider.h:53:27: warning: private field 'unboxing_option_' is not used [-Wunused-private-field]
>   ProtoWrapperTypeOptions unboxing_option_;
>                           ^
> 1 warning generated.
> INFO: From Action external/opencensus_proto/opencensus/proto/agent/trace/v1/trace_service.grpc.pb.h:
> bazel-out/k8-fastbuild/bin/external/opencensus_proto/external/opencensus_proto: warning: directory does not exist.
> INFO: From Action external/com_google_googleapis/google/devtools/cloudtrace/v2/trace.grpc.pb.h:
> bazel-out/k8-fastbuild/bin/external/com_google_googleapis/external/com_google_googleapis: warning: directory does not exist.
> google/devtools/cloudtrace/v2/trace.proto:19:1: warning: Import google/api/annotations.proto is unused.
> google/devtools/cloudtrace/v2/tracing.proto:26:1: warning: Import google/protobuf/timestamp.proto is unused.
> INFO: Elapsed time: 2716.501s, Critical Path: 351.98s
> INFO: 4591 processes: 1559 internal, 3032 linux-sandbox.
> INFO: Build completed successfully, 4591 total actions
> 
> Executed 4 out of 4 tests: 4 tests pass.
> ```
> 
> # TSAN
> ```
> $bazel test -c fastbuild //contrib/golang/filters/http/test/...   --config=clang-tsan --test_verbose_timeout_warnings   --test_env=ENVOY_IP_TEST_VERSIONS=v4only --verbose_failures  --config=gcc  --test_arg="-l debug"  --define=wasm=disabled --build_tests_only
> Starting local Bazel server and connecting to it...
> WARNING: option '--build_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
> WARNING: option '--test_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
> DEBUG: /home/fakang.wfk/.cache/bazel/_bazel_fakang.wfk/05ab2d40a050525b20b2187bf5222296/external/rules_python/python/pip.bzl:46:10: pip_install is deprecated. Please switch to pip_parse. pip_install will be removed in a future release.
> WARNING: option '--build_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
> WARNING: option '--test_tag_filters' was expanded to from both option '--config=clang-tsan' (source command line options) and option '--config=clang-tsan' (source command line options)
> INFO: Analyzed 4 targets (930 packages loaded, 32893 targets configured).
> INFO: Found 4 test targets...
> INFO: Deleting stale sandbox base /home/fakang.wfk/.cache/bazel/_bazel_fakang.wfk/05ab2d40a050525b20b2187bf5222296/sandbox
> INFO: Elapsed time: 566.694s, Critical Path: 471.52s
> INFO: 22 processes: 1 internal, 21 linux-sandbox.
> INFO: Build completed successfully, 22 total actions
> 
> Executed 3 out of 4 tests: 4 tests pass.
> ```

@mattklein123 Thanks for your comments, all of them are addressed. Thanks

contrib/golang/filters/http/source/common/dso/dso.h
```diff
@@ -0,0 +1,85 @@
...
+   * @param dso_id is unique ID for dynamic library.
+   * @return nullptr if get failed. Otherwise, return the DSO instance.
+   */
+  static DsoInstancePtr getDsoInstanceByID(std::string dso_id);
+
+  /**
+   * Get the full amount of dynamic library information.
+   * @return all so ids separated by commas.
+   */
+  static std::string show();
```
mattklein123, line 70: Is this used anywhere?

contrib/golang/filters/http/source/golang_filter.h
```diff
@@ -0,0 +1,259 @@
...
+
+    func(std::forward<Args>(args)...);
+
+    auto duration = std::chrono::duration_cast<T>(
+        std::chrono::steady_clock::now() - // NO_CHECK_FORMAT(real_time)
+        start);
+
+    return duration.count();
+  }
+};
```
mattklein123, line 254: Where is this used?

mattklein123 (changes requested):
/wait

contrib/golang/filters/http/source/golang_filter.h
wangfakang: > Where is this used?

@mattklein123 This is a generic function to count the time consumed by function calls, I will add cgo time consumption statistics in the next version. So the implementation of the function is kept for now.

contrib/golang/filters/http/source/common/dso/dso.h
wangfakang: > Is this used anywhere?

@mattklein123 This is a generic function for the DSO class, I will add some information about DSO runtime statistics in the next version. So the implementation of the function is kept for now.

mattklein123: Please remove unused code. You can add it back in the next PRs. Same below. Thanks.

contrib/golang/filters/http/source/golang_filter.h
wangfakang: Done

contrib/golang/filters/http/source/common/dso/dso.h
wangfakang: @mattklein123  Done. Thanks for your comments.

mattklein123 (approved):
Thanks, LGTM to ship and iterate! Cool stuff.

PR #22586, opened 2022-08-05
gcp_authn: Stop the filter chain iteration for headers as well as data and trailers

tyxia:
`decodeHeader()` stop the filter chain iteration but the `decodeData()` resumes it because its default implementation in `pass_through_filter` returns `Http::FilterDataStatus::Continue`. This causes the problem that filter chain iteration resumes without waiting for the response from authentication server.
In order to fix it, we need to return `StopAllIterationAndWatermark` to stop the iteration for headers as well as data and trailers.

Risk Level: low

Testing: fix is confirmed from product team, unit test and integration test added in this PR


tyxia:
/unassign @htuch 

There is no dependency change in this PR

tyxia:
/assign @yanavlasov 

PTAL, Thanks!

nielsole:
Hi, thanks for creating this patch!
I cherry-picked this on 1.23.0 and can confirm that it works in our use-case.

As I understand it, `onComplete` is currently called for two different purposes:
* end of the original request stream
* the callback from `GcpAuthnClient`

The tricky part that lead to this bug seems to be that discerning these two calls and the order of the invocations. As we completely control `GcpAuthnClient` as well as `RequestCallbacks` class definition, wouldn't it be cleaner to have the GCP response call back as something like`onGcpClientComplete` and move the whole method that is currently `onComplete` to `onGcpClientComplete`? This would make state and error management around `state_`, `initiating_call_` and `response != nullptr` a bit easier, wouldn't it? This PR fixes the bug and somewhat serializes callbacks so this would would likely be a follow-up action if you concur.
I am unfamiliar with the codebase, nor am I too familiar with C++, so please take this as a good faith naive question :)

tyxia:
> onComplete
Hi, Thank you very much for confirmation with testing and your question! 

If I understand your points correctly, I think the name `onComplete` [here](https://github.com/envoyproxy/envoy/blob/eb352bb05d37a8ad4b6cc7012dc7fef74767b2ad/source/extensions/filters/http/gcp_authn/gcp_authn_impl.h#L31) confused you a bit. It is only used in `GcpAuthn filter` and it has nothing to do with other `onComplete` functions like stream callbacks [here](https://github.com/envoyproxy/envoy/blob/eb352bb05d37a8ad4b6cc7012dc7fef74767b2ad/envoy/http/async_client.h#L121). They just happen to be same name :)

IMO, Changing the name to `onGcpClientComplete` might be a bit verbose and not aligned with other conventions in the code base very well. Otherwise, we need give every `onComplete` in code base a prefix :) )

I feel as along as we are clear that this `RequestCallbacks` class along with its `onComplete` is a completely independent struct only in gcp_authn filter scope,  it equals to changing the name `onGcpClientComplete`. Basically, we use namespace/class scope to represent the meaning of the function.

BTW, the bug here is not related to `onComplete` callback. It is about the filter chain iteration that is stopped or resumed by `decodeHeader` and `decodeData`.

Please let me know if you have any other questions or thoughts.

nielsole:
> I think the name of onComplete function [here](https://github.com/envoyproxy/envoy/blob/eb352bb05d37a8ad4b6cc7012dc7fef74767b2ad/source/extensions/filters/http/gcp_authn/gcp_authn_impl.h#L31) confused you a bit. It is only used in GcpAuthn filter as the callback from GcpAuthnClientand it has nothing to do with other onComplete functions like stream callbacks [here](https://github.com/envoyproxy/envoy/blob/eb352bb05d37a8ad4b6cc7012dc7fef74767b2ad/envoy/http/async_client.h#L121). They just happen to be same name :)

thanks for clearing up my misunderstanding. 
In that case my suggestion indeed does not make much sense. :)

test/extensions/filters/http/gcp_authn/gcp_authn_filter_integration_test.cc
```diff
@@ -217,14 +223,34 @@ TEST_P(GcpAuthnFilterIntegrationTest, BasicflowWithoutAudience) {
...
+  initializeConfig(/*add_audience=*/true);
+  HttpIntegrationTest::initialize();
+  initiateClientConnection(/*send_request_body=*/true);
+  // Send the request with long wait time to intentionally delay the response from `gcp_authn`
+  // cluster.
+  EXPECT_TRUE(fake_upstreams_[1]->waitForHttpConnection(*dispatcher_, fake_gcp_authn_connection_,
+                                                        std::chrono::milliseconds(500000)));
+  // Send the request to `cluster_0` cluster.
+  AssertionResult assert_result = fake_upstreams_[0]->waitForHttpConnection(
+      *dispatcher_, fake_upstream_connection_, std::chrono::milliseconds(1));
```
yanavlasov, line 248: I think 1 millisecond is too short. Making it at least 1 second is more convincing that the request is not going through.

yanavlasov (changes requested):
/wait

test/extensions/filters/http/gcp_authn/gcp_authn_filter_integration_test.cc
tyxia: Yep.  Done

Also I have verified this test will fail without return status change and will pass consistently with this PR

yanavlasov (approved):

Issue #17601, opened 2021-08-05
dubbo tracing support


wbpcode:
After #16049 is completed, tracers can be used by various proxies, not just limited to HTTP. In envoy, now it is possible to trace dubbo/thrift streams as well as HTTP requests.

Dubbo Proxy would be a good start.

wbpcode:
cc @zyfjeff @lizan

yingjianjian:
> cc @zyfjeff @lizan

Hello, has Dubbo trace supported it? I'm looking forward to it @wbpcode

wbpcode:
> > cc @zyfjeff @lizan
> 
> Hello, has Dubbo trace supported it? I'm looking forward to it @wbpcode

sorry, it's not supported for now.

tedli:
Hi,

I'm trying to add a filter of dubbo, like the Router do, for learning purpose. Current only want to read from config and log something.

So, I created a filter, I want to read config using the `config` field of DubboFilter which is an `Any`. I give my filter a name, and register it with config field type `envoy::extensions::filters::network::http_connection_manager::v3::HttpConnectionManager::Tracing`, it built.
However when running, envoy complains about it don't know how to unmarsal yaml config to object.

```c++
DubboFilters::FilterFactoryCb LalalaFilterConfig::createFilterFactoryFromProtoTyped(
    [[maybe_unused]] const envoy::extensions::filters::network::http_connection_manager::v3::HttpConnectionManager::Tracing& proto,
    [[maybe_unused]] const std::string& stat_prefix,
    Server::Configuration::FactoryContext& context) {
    // log proto...
  return [&context](DubboFilters::FilterChainFactoryCallbacks& callbacks) -> void {
    callbacks.addFilter(std::make_shared<Lalala>(context.clusterManager()));
  };
}
```

```log
[2022-07-26 17:01:19.835][9004][critical][main] [source/server/server.cc:116] error initializing configuration '/path/envoy.yaml': Unable to parse JSON as proto (INVALID_ARGUMENT:(dubbo_filters[1].config): invalid value Invalid type URL, unknown type: extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.Tracing for type Any)
```

Is there some docs about this? Or there already some other filters, codes that did things like this, that I can imitate？

tedli:
Hi @wbpcode ,

[...]

PR #22594, opened 2022-08-07
WiP tracing: dubbo proxy tracing #17601

tedli:
Commit Message: Add tracing support for dubbo proxy
Additional Description: #17601 
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/network/dubbo_proxy/BUILD
+++ b/source/extensions/filters/network/dubbo_proxy/BUILD
@@ -110,6 +110,7 @@ envoy_cc_extension(
         "//source/extensions/filters/network/dubbo_proxy/router:rds_lib",
         "//source/extensions/filters/network/dubbo_proxy/router:route_matcher",
         "//source/extensions/filters/network/dubbo_proxy/router:router_lib",
+        "//source/extensions/filters/network/dubbo_proxy/tracer:config",
         "@envoy_api//envoy/extensions/filters/network/dubbo_proxy/v3:pkg_cc_proto",
     ],
 )
--- a/source/extensions/filters/network/dubbo_proxy/active_message.cc
+++ b/source/extensions/filters/network/dubbo_proxy/active_message.cc
@@ -102,6 +102,11 @@ const Network::Connection* ActiveMessageFilterBase::connection() const {
 
 Router::RouteConstSharedPtr ActiveMessageFilterBase::route() { return parent_.route(); }
 
+Tracer::Config& ActiveMessageFilterBase::tracerConfig() {
+  return parent_.tracerConfig();
+}
+
+
 SerializationType ActiveMessageFilterBase::serializationType() const {
   return parent_.serializationType();
 }
@@ -303,6 +308,10 @@ DubboProxy::Router::RouteConstSharedPtr ActiveMessage::route() {
   }
 
   return nullptr;
+}
+
+Tracer::Config& ActiveMessage::tracerConfig() {
+  return parent_.config().tracerConfig();
 }
 
 FilterStatus ActiveMessage::applyDecoderFilters(ActiveMessageDecoderFilter* filter,
--- a/source/extensions/filters/network/dubbo_proxy/active_message.h
+++ b/source/extensions/filters/network/dubbo_proxy/active_message.h
@@ -71,6 +71,7 @@ public:
   uint64_t streamId() const override;
   const Network::Connection* connection() const override;
   DubboProxy::Router::RouteConstSharedPtr route() override;
+  Tracer::Config& tracerConfig() override;
   SerializationType serializationType() const override;
   ProtocolType protocolType() const override;
   StreamInfo::StreamInfo& streamInfo() override;
@@ -164,6 +165,7 @@ public:
   ProtocolType protocolType() const;
   StreamInfo::StreamInfo& streamInfo();
   Router::RouteConstSharedPtr route();
+  Tracer::Config& tracerConfig();
   void sendLocalReply(const DubboFilters::DirectResponse& response, bool end_stream);
   void startUpstreamResponse();
   DubboFilters::UpstreamResponseStatus upstreamData(Buffer::Instance& buffer);
--- a/source/extensions/filters/network/dubbo_proxy/config.cc
+++ b/source/extensions/filters/network/dubbo_proxy/config.cc
@@ -2,9 +2,14 @@
 
 #include "envoy/extensions/filters/network/dubbo_proxy/v3/dubbo_proxy.pb.h"
 #include "envoy/extensions/filters/network/dubbo_proxy/v3/dubbo_proxy.pb.validate.h"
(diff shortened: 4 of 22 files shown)


tedli:
Up to now, my code confirmed works. It has to add code in router, because there could access decode invocation.
I wrote a [memo](https://lz8888.gitee.io/post/envoy-dubbo-tracing), describing the problem I'm facing.
I will correct span tags and refactor code the next step.

tedli:
**Demo:**

```yaml
node:
  id: dev
  cluster: dev
admin:
  address:
    socket_address:
      protocol: TCP
      address: 127.0.0.1
      port_value: 15000
static_resources:
  listeners:
  - name: dubbo
    address:
      socket_address:
        address: 0.0.0.0
        port_value: 30880
    filter_chains:
    - filters:
      - name: envoy.filters.network.dubbo_proxy
        typed_config:
          "@type": type.googleapis.com/envoy.extensions.filters.network.dubbo_proxy.v3.DubboProxy
          stat_prefix: dubbo
          protocol_type: Dubbo
          serialization_type: Hessian2
          route_config:
          - name: dubbo
            interface: "*"
            routes:
            - match:
                method:
                  name:
                    safe_regex:
                      google_re2:
                        max_program_size: 100
                      regex: ".+"
              route:
                cluster: dubbo
          dubbo_filters:
          - name: envoy.filters.dubbo.router
          - name: envoy.filters.dubbo.tracer
            config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.Tracing
              provider:
                name: envoy.tracers.skywalking
                typed_config:
                  "@type": type.googleapis.com/envoy.config.trace.v3.SkyWalkingConfig
                  grpc_service:
                    envoy_grpc:
                      cluster_name: oap
                    timeout: 0.250s
                  client_config:
                    service_name: sidecar
                    instance_name: lalala
  clusters:
  - name: dubbo
    connect_timeout: 0.25s
    type: STATIC
    load_assignment:
      cluster_name: dubbo
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: 172.16.29.218
                port_value: 20880
  - name: oap
    connect_timeout: 0.25s
    type: STATIC
    typed_extension_protocol_options:
      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
        explicit_http_config:
          http2_protocol_options: {}
    load_assignment:
      cluster_name: oap
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: 172.16.29.218
                port_value: 11800
```
There dubbo2 sample available:

```bash
docker pull sofastack/e2e-dubbo-consumer:0.2.1
docker pull sofastack/e2e-dubbo-provider:0.2.1
```

I copied `jars` from the above images, so I can run directly on my host.

```powershell
# I using powershell, if use bash or the args may need to addjust.

java -Xms1g -Xmx2g "-Ddubbo.protocol.port=20880" "-Ddubbo.provider.threadpool=fixed" "-Ddubbo.provider.threads=1000" "-Ddubbo.provider.iothreads=3" "-Dserver.port=65080" "-Dmanagement.server.port=65081" "-Dagent.service_name=provider" "-javaagent:..\..\Downloads\skywalking-agent\skywalking-agent.jar" -jar dubbo-provider.jar

java -Xms1g -Xmx2g "-Ddubbo.reference.com.alibaba.boot.dubbo.demo.consumer.DemoService.url=dubbo://172.28.216.18:30880" "-Dserver.port=65082" "-Dmanagement.server.port=65083" "-Dlogging.level.org.springframework.web=ERROR" "-Ddubbo.consumer.check=false" "-Dagent.service_name=consumer" "-javaagent:..\..\Downloads\skywalking-agent\skywalking-agent.jar" -jar dubbo-consumer.jar
```

And, it works.

image

wbpcode:
Thanks for you contribution. But if I miss something? Where is the new proto API for the dubbo tracing?

And there are some exist tracers which are protocol independent. We should reuse these tracers rather than to create new tracers for specific proxy.

tedli:
> Thanks for you contribution. But if I miss something? Where is the new proto API for the dubbo tracing?
> 
> And there are some exist tracers which are protocol independent. We should reuse these tracers rather than to create new tracers for specific proxy.
> 
> Thanks again.
> 
> /wait

Thanks for reply.
1. No proto API added, this `type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.Tracing` proto is reused.
2. No new tracer created. `Tracing::HttpTracerSharedPtr` is reused, so do the factory.

What I've done:
1. Empty tracer filter, this filter do nothing, only holds the proto config.
2. A `tracer_util` for create, manipulate span tags, and act as wrapper between header map and dubbo attachment.
3. The skywalking driver impl is not protocol independent, the ctor of sw span hardcoded span layer to http which is the default value, so I set span layer to rpc framework.

I'm glad to finish this PR. Please point out the inappropriate code in more detail, so I can get how to correct.

Thanks.

wbpcode:
1. IMO, it's OK to resue the `HttpConnectionManager.Tracing` for learning or private extension, because it just work. But if we want make this extension be part of core envoy, may be a more clear/common new proto API is necessary. Or may be we need move the `HttpConnectionManager.Tracing` to a common pkg.
2. It would be better to add a new tracing filed in the `DubboProxy` for dubbo tracing rather than create a special L7 dubbo filter.
3. If the skywalking driver impl is not protocol independent, we should try to fix it first.

tedli:
> 1. IMO, it's OK to resue the `HttpConnectionManager.Tracing` for learning or private extension, because it just work. But if we want make this extension be part of core envoy, may be a more clear/common new proto API is necessary. Or may be we need move the `HttpConnectionManager.Tracing` to a common pkg.
> 2. It would be better to add a new tracing filed in the `DubboProxy` for dubbo tracing rather than create a special L7 dubbo filter.
> 3. If the skywalking driver impl is not protocol independent, we should try to fix it first.

Yes, it make sense, thanks a lot. It will be a very rewarding exercise to me.

Then hold this PR. I will firstly open up a new one to add something like `setProtocol` to `Tracing::Span` interface, so other network filters can set to corresponding protocol, rather than the default value http.

jmarantz:
Looks like this still needs a format fix.

/wait

tedli:
> Looks like this still needs a format fix.
> 
> /wait

Yes. It's unfinished, I forget to convert to draft. As [wbpcode](https://github.com/wbpcode) said, it's better to introduce a tracing field to dubbo proxy proto, and refactor http tracing proto to common, rather than use it directly in other network filter.

I'm glad to finish this PR, I will keep on doing this.

PR #22601, opened 2022-08-08
network: try getOriginalDst from transparent socket's local_address

chrysan55:
Transparent socket receives packets with original destination address,
but getOriginalDst() would fall through when traffic is redirected by
iptables TPROXY target since SO_ORIGINAL_DST relies on nf_conntrack.

Signed-off-by: Gang Zhao <chrysan55@gmail.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


kyessenov:
Thanks for PR. Is using [transparent](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-transparent) flag not sufficient for TPROXY instead of using original_dst listener filter?

Please don't force push commits.

chrysan55:
> Please don't force push commits.

sorry for that , I would fix code style with next commit /XP

>  Is using [transparent](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-transparent) flag not sufficient for TPROXY instead of using original_dst listener filter?

Listener doc says:

> This flag should be used in combination with [an original_dst](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/listener_filters/original_dst_filter#config-listener-filters-original-dst) [listener filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-listener-filters) to mark the connections’ local addresses as “restored.”

From this line it seems like not sufficient using only transparent flag without original_dst listener filter. 

AFAIK listener with ([use_original_dst](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-use-original-dst) == True && [bind_to_port](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-bind-to-port) == False) works only in the condition of `localAddressRestored`, which relies on original_dst filter.

What this PR try to fix is original_dst filter doesnt work without nf_conntrack module. But nf_conntrack is unnecessary when we use iptables TPROXY for traffic redirect.

source/common/network/utility.cc
```diff
@@ -355,8 +355,19 @@ Address::InstanceConstSharedPtr Utility::getOriginalDst(Socket& sock) {
...
 
+#ifdef IP_TRANSPARENT
+  socklen_t flag_len = sizeof(int);
+  int is_transparent;
+  if (status != 0) {
+    status = sock.getSocketOption(SOL_IP, IP_TRANSPARENT, &is_transparent, &flag_len).return_value_;
+    if (status != 0 || !is_transparent) {
+      return nullptr;
+    }
+    safeMemcpy(reinterpret_cast<sockaddr*>(&orig_addr), sock.ioHandle().localAddress()->sockAddr());
```
kyessenov, line 366: It is not immediately obvious to me that `orig_addr` would be safe to use here. Is it possible that status on line 352 is error on IP_TRANSPARENT? Is there some other reason besides IP_TRANSPARENT that could make the address invalid?

chrysan55: > Is it possible that status on line 352 is error on IP_TRANSPARENT? 

Yes.  SO_ORIGINAL_DST is implemented in nf_conntrack, so status on line 352 would be -1 when there is no nf_conntrack module.

> Is there some other reason besides IP_TRANSPARENT that could make the address invalid?

I think original_dst filter implies listener socket is not `bind_to_port`,  in another word, traffic is redirected to listener socket.

Normally we use iptables NAT targets (including SNAT/DNAT/REDIRECT/MASQUERADE) to do that, which relies on nf_conntrack. So there is always nf_conntrack module in this "traffic-redirected-by-NAT" scenario, and getting SO_ORIGINAL_DST option works fine.

[tproxy](https://www.kernel.org/doc/html/latest/networking/tproxy.html#redirecting-traffic) is another way to do traffic redirect, which requires listener socket is IP_TRANSPARENT and could work without nf_conntrack. Then `getSocketOption` on line 352 returns -1 and func `Utility::getOriginalDst` returns nullptr. 

In this  "traffic-redirected-by-only-tproxy" scenario, the original_dst is actually the accepted socket's local address. This patch is trying to screen this branch out as a valid original_dst address.

This is my consideration, please let me know if it's wrong or confusing.

soulxu: It seems like you can return sock.ioHandle().localAddress() directly, needn't a memcopy to orig_addr

source/common/network/utility.cc
```diff
@@ -355,8 +355,19 @@ Address::InstanceConstSharedPtr Utility::getOriginalDst(Socket& sock) {
         sock.getSocketOption(SOL_IPV6, IP6T_SO_ORIGINAL_DST, &orig_addr, &addr_len).return_value_;
   }
 
+#ifdef IP_TRANSPARENT
```
soulxu, line 358: seems like you should check ip transparent available or not by https://github.com/envoyproxy/envoy/blob/768d0bdc25c19e0088e1af105b7d65507e47d1b9/source/common/api/posix/os_sys_calls_impl.h#L31, not only the macro

source/common/network/utility.cc
```diff
@@ -355,8 +355,19 @@ Address::InstanceConstSharedPtr Utility::getOriginalDst(Socket& sock) {
...
+  if (status != 0) {
+    status = sock.getSocketOption(SOL_IP, IP_TRANSPARENT, &is_transparent, &flag_len).return_value_;
+    if (status != 0 || !is_transparent) {
+      return nullptr;
+    }
+    safeMemcpy(reinterpret_cast<sockaddr*>(&orig_addr), sock.ioHandle().localAddress()->sockAddr());
+#else
   if (status != 0) {
     return nullptr;
+#endif
```
soulxu, line 370: if the IP transparent doesn't support on the os, and socket is transparent, should we log a warning here.

source/common/network/utility.cc
chrysan55: yes.. got it

source/common/network/utility.cc
chrysan55: Dont know why socket could be transparent when os doesnt support IP_TRANSPARENT...

soulxu: Haha, right, sorry for this stupid comment. Please ignore it.

source/common/network/utility.cc
chrysan55: have fixed this check, could you pls take a look again, thanks

source/common/network/utility.cc
```diff
@@ -356,6 +356,17 @@ Address::InstanceConstSharedPtr Utility::getOriginalDst(Socket& sock) {
   }
 
   if (status != 0) {
+#ifdef IP_TRANSPARENT
```
soulxu, line 359: I thought we needn't this since we already use `Api::OsSysCallsSingleton::get().supportsIpTransparent()`

source/common/network/utility.cc
```diff
@@ -356,6 +356,17 @@ Address::InstanceConstSharedPtr Utility::getOriginalDst(Socket& sock) {
...
+    if (Api::OsSysCallsSingleton::get().supportsIpTransparent()) {
+      socklen_t flag_len = sizeof(int);
+      int is_transparent;
+      status =
+          sock.getSocketOption(SOL_IP, IP_TRANSPARENT, &is_transparent, &flag_len).return_value_;
+      if (status == 0 && is_transparent) {
+        return sock.ioHandle().localAddress();
+      }
```
soulxu, line 367: would you able to add test for the logic you added?

soulxu:
thanks for the update, few comments

source/common/network/utility.cc
chrysan55: I've got compiling error of IP_TRANSPARENT not define in msvc ci building... Should I use another macro to set these lines apart from msvc platform ?

source/common/network/utility.cc
chrysan55: I'd like to but I'm not sure this logic could be unit tested. 
I checked  https://github.com/envoyproxy/envoy/blob/6b83db42c418abca6d21160cf6b1c439657d6ef0/test/common/network/utility_test.cc#L225
and have no idea what is this for ...

source/common/network/utility.cc
soulxu: Maybe worth a try on this
https://github.com/envoyproxy/envoy/blob/15d8b93608bc5e28569f8b042ae666a5b09b87e9/source/common/network/socket_option_impl.h#L14-L24

Also looks like you need to take care of different IP version address.

chrysan55: thanks for your help, I will improve this in next commit

soulxu:
thanks for the update! the code LGTM. The only left thing is the test, seems like there isn't any existing test for this filter yet. Maybe you can use `Api::MockOsSysCalls` to mock the getopt syscall to do the unittest.

chrysan55:
okay, i will complete this function's uint test

soulxu:
Sorry, forget mention you probably need a release note also, you can add here https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml

test/common/network/utility_test.cc
```diff
@@ -233,6 +244,47 @@ TEST(NetworkUtility, GetOriginalDst) {
...
+  sin.sin_family = AF_INET;
+  sin.sin_addr.s_addr = inet_addr("12.34.56.78");
+  socket.connection_info_provider_->setLocalAddress(
+      std::make_shared<Network::Address::Ipv4Instance>("1.2.3.4"));
+  EXPECT_CALL(socket, ipVersion()).WillRepeatedly(Return(Address::IpVersion::v4));
+  EXPECT_CALL(socket, getSocketOption(Eq(SOL_IP), Eq(SO_ORIGINAL_DST), _, _))
+      .WillOnce(DoAll(SetArg2Sockaddr(storage), Return(Api::SysCallIntResult{0, 0})))
+      .WillOnce(Return(Api::SysCallIntResult{-1, 0}));
+  EXPECT_CALL(socket, getSocketOption(Eq(SOL_IP), Eq(IP_TRANSPARENT), _, _))
+      .WillOnce(DoAll(SetArg2Int(1), Return(Api::SysCallIntResult{0, 0})));
```
soulxu, line 267: seems miss one more case for `getSocketOption` with `IP_TRANSPARENT` return failure.

soulxu (changes requested):
LGTM, last one comment, thanks!

soulxu:
@chrysan55 seems your latest commits using the wrong email to sign off https://github.com/envoyproxy/envoy/pull/22601/commits/604d4dc37798b37f1089927c5657c3b835f9fe53, then you have to fix those sign off on those commits, then do a force push, then CI will happy

chrysan55:
> @chrysan55 seems your latest commits using the wrong email to sign off [604d4dc](https://github.com/envoyproxy/envoy/commit/604d4dc37798b37f1089927c5657c3b835f9fe53), you have to fix those sign off on those commits, then do a force push, CI will happy

sorry i did stupid push .. will reset it immediately

soulxu:
> > @chrysan55 seems your latest commits using the wrong email to sign off [604d4dc](https://github.com/envoyproxy/envoy/commit/604d4dc37798b37f1089927c5657c3b835f9fe53), you have to fix those sign off on those commits, then do a force push, CI will happy
> 
> sorry i did stupid push .. will reset it immediately

no worries, free to ping me on slack (Alex Xu) if you need help.

test/common/network/utility_test.cc
chrysan55: ok ,  will fix it tonight

soulxu:
LGTM, thanks!

@kyessenov @lizan for maintainer review

changelogs/current.yaml
```diff
@@ -21,6 +21,9 @@ minor_behavior_changes:
 - area: prometheus_stats
   change: |
     removed blank line for being compatible with OpenMetrics.
+- area: original_dst
+  change: |
+    transparent listener could use original_dst filter without nf_conntrack enabled.
```
kyessenov, line 26: nit: could -> can

source/common/network/utility.cc
```diff
@@ -343,19 +344,32 @@ Address::InstanceConstSharedPtr Utility::getOriginalDst(Socket& sock) {
...
   }
 
+  sockaddr_storage orig_addr;
+  memset(&orig_addr, 0, sizeof(orig_addr));
+  socklen_t addr_len = sizeof(sockaddr_storage);
+  int status =
+      sock.getSocketOption(opt_dst.level(), opt_dst.option(), &orig_addr, &addr_len).return_value_;
+
   if (status != 0) {
+    if (Api::OsSysCallsSingleton::get().supportsIpTransparent()) {
```
kyessenov, line 364: Does this make a syscall? Would it be better to save the result once per filter instantiation, instead of calling it per connection?

changelogs/current.yaml
chrysan55: thanks, i will correct it

source/common/network/utility.cc
chrysan55: I think the socket syscall would be invoked only once since it's a singleton instance and the return is a static const.

https://github.com/envoyproxy/envoy/blob/5183dbf7f11b23f1f7013a3e42cbce71eb63252a/source/common/api/posix/os_sys_calls_impl.cc#L141

kyessenov (approved):
Thank you!

PR #22607, opened 2022-08-08
Add warning log when ThreadSynchronizer blocked for over 10 seconds

ravenblackx:
Commit Message: Add warning log when ThreadSynchronizer blocked for over 10 seconds
Additional Description: Before this change, if barrierOn blocks indefinitely, the test times out with no indication of what went wrong in the logs. After the change, the test will still time out, but the log will have a bunch of warnings telling the user why it's stuck.
Risk Level: None, test-only.
Testing: Test-only.
Docs Changes: Code-comments only.
Release Notes: n/a

### DIFF
--- a/source/common/common/thread_synchronizer.cc
+++ b/source/common/common/thread_synchronizer.cc
@@ -65,7 +65,7 @@ void ThreadSynchronizer::barrierOnWorker(absl::string_view event_name) {
   absl::MutexLock lock(&entry.mutex_);
   ENVOY_LOG(debug, "thread synchronizer: barrier on {}", event_name);
   while (!entry.mutex_.AwaitWithTimeout(absl::Condition(&entry.at_barrier_), absl::Seconds(10))) {
-    ENVOY_LOG(warning, "thread synchronizer: barrier on {} stuck for 10 seconds");
+    ENVOY_LOG(warning, "thread synchronizer: barrier on {} stuck for 10 seconds", event_name);
   }
   ENVOY_LOG(debug, "thread synchronizer: barrier complete {}", event_name);
 }
--- a/source/common/common/thread_synchronizer.cc
+++ b/source/common/common/thread_synchronizer.cc
@@ -60,17 +60,14 @@ void ThreadSynchronizer::syncPointWorker(absl::string_view event_name) {
   entry.signaled_ = false;
 }
 
-bool ThreadSynchronizer::barrierOnWorker(absl::string_view event_name, absl::Duration timeout) {
+void ThreadSynchronizer::barrierOnWorker(absl::string_view event_name) {
   SynchronizerEntry& entry = getOrCreateEntry(event_name);
   absl::MutexLock lock(&entry.mutex_);
   ENVOY_LOG(debug, "thread synchronizer: barrier on {}", event_name);
-  bool is_success = entry.mutex_.AwaitWithTimeout(absl::Condition(&entry.at_barrier_), timeout);
-  if (is_success) {
-    ENVOY_LOG(debug, "thread synchronizer: barrier complete {}", event_name);
-  } else {
-    ENVOY_LOG(debug, "thread synchronizer: barrier timed out {}", event_name);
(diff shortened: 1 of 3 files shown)


test/common/common/thread_test.cc
```diff
@@ -19,6 +19,30 @@ class ThreadAsyncPtrTest : public testing::Test {
...
+  sync.waitOn("triggered_quickly");
+
+  // Thread triggers the syncPoint.
+  auto thread1 = thread_factory_.createThread(
+      [&sync]() {
+        sync.syncPoint("triggered_quickly");
+        return true;
+      },
+      Options{"thread1"});
+  EXPECT_TRUE(sync.barrierOn("triggered_quickly", absl::Seconds(1)));
```
adisuissa, line 41: I'm not sure if it's a real concern, but this may be flaky if the thread creation takes too long.
I guess one way to handle this is to increase the timeout, but @jmarantz may have a better idea (e.g., if some TimeSystem can be used here).

adisuissa (commented):
Thanks for adding this!
Overall LGTM, one possible test issue that I'm not certain about.

test/common/common/thread_test.cc
jmarantz: II think this will be flake-prone, especially at 1 second. Also it will make debugging difficult.

I don't immediately see how to solve this problem with dispatch timers and simulated time, but maybe maybe @mattklein123 has some ideas.

But in any case I'd like to suggest:
 * not making the timeout an arg to barrierOn, but make it always apply, and be something more like 30s to reduce flakes.
 * enabling that timeout only when not compiled for debug.
 
I'm not 100% sure I love either of those suggestions myself but it might avoid the two issues I see.

If the goal is simply to have better error messages in test logs about issues at the barrier, then I do have another idea.  Don't exit the test, but do (say) log warnings every 10s while stuck at the barrier. Worst case that spins for 300s but you know where you got stuck, but don't introduce any flakes, and don't make debugging difficult.

ravenblackx: I'd be surprised if you could get 1 second to time out on such a simple test, given its runtime is typically <1ms, but okay, changed it to 30s since when it's passing a large timeout doesn't cost anything.

I think changing behavior for debug is probably more likely to be confusing than anything (and you could easily remove the specific timeout temporarily if you were unlucky enough to run into that issue). I didn't want to add a default timeout since that would change existing behaviors, versus a parameter can just change things people specifically want to change.

Making it a logging warning is a good idea though, that solves the issue well. Can't really add a test for that though.

jmarantz: Yeah the warning is better. You can test with EXPECT_LOG_CONTAINS and other macros in ./test_common/logging.h

You could make a test that blocks for (say) 20s before releasing the binary, and print warnings every 10s. You'll have to make sure the test isn't flaky though which might be a little tricky.

ravenblackx: Yeah, not keen on trying to test that, because A. the test would take *at least* 10 seconds to validate the behavior, which sucks, and B. it would be potentially flaky even if we found a good way to do it.

Counterpoint, it's essentially a test-only code path and does nothing new except log, maybe it just doesn't merit testing.

source/common/common/thread_synchronizer.cc
```diff
@@ -64,7 +64,9 @@ void ThreadSynchronizer::barrierOnWorker(absl::string_view event_name) {
   SynchronizerEntry& entry = getOrCreateEntry(event_name);
   absl::MutexLock lock(&entry.mutex_);
   ENVOY_LOG(debug, "thread synchronizer: barrier on {}", event_name);
-  entry.mutex_.Await(absl::Condition(&entry.at_barrier_));
+  while (!entry.mutex_.AwaitWithTimeout(absl::Condition(&entry.at_barrier_), absl::Seconds(10))) {
+    ENVOY_LOG(warning, "thread synchronizer: barrier on {} stuck for 10 seconds", event_name);
```
jmarantz, line 68: s/warning/warn/

jmarantz (commented):
lgtm to me modulo typo :)

/wait

jmarantz:
Probably worth @mattklein123 having a quick look -- it's a tiny change.

PR #22608, opened 2022-08-08
http: changing some Connection* interfaces to be OptRef

alyssawilk:
Risk Level: low
Testing: existing tests
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/filter_manager.cc
+++ b/source/common/http/filter_manager.cc
@@ -232,7 +232,9 @@ bool ActiveStreamFilterBase::commonHandleAfterTrailersCallback(FilterTrailersSta
   return true;
 }
 
-const Network::Connection* ActiveStreamFilterBase::connection() { return parent_.connection(); }
+OptRef<const Network::Connection> ActiveStreamFilterBase::connection() {
+  return parent_.connection();
+}
 
 Event::Dispatcher& ActiveStreamFilterBase::dispatcher() { return parent_.dispatcher_; }
 
@@ -1335,7 +1337,9 @@ void FilterManager::setBufferLimit(uint32_t new_limit) {
 }
 
 void FilterManager::contextOnContinue(ScopeTrackedObjectStack& tracked_object_stack) {
-  tracked_object_stack.add(connection_);
+  if (connection_.has_value()) {
+    tracked_object_stack.add(*connection_);
+  }
   tracked_object_stack.add(filter_manager_callbacks_.scope());
 }
 
--- a/source/common/router/router.cc
+++ b/source/common/router/router.cc
@@ -54,7 +54,7 @@ bool schemeIsHttp(const Http::RequestHeaderMap& downstream_headers,
   if (downstream_headers.getSchemeValue() == Http::Headers::get().SchemeValues.Http) {
     return true;
   }
-  if (connection && !connection->ssl()) {
+  if (connection.has_value() && !connection->ssl()) {
     return true;
   }
   return false;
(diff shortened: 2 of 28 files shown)


source/common/router/router.cc
```diff
@@ -50,11 +50,11 @@ constexpr char NumInternalRedirectsFilterStateName[] = "num_internal_redirects";
 uint32_t getLength(const Buffer::Instance* instance) { return instance ? instance->length() : 0; }
 
 bool schemeIsHttp(const Http::RequestHeaderMap& downstream_headers,
-                  const Network::Connection& connection) {
+                  OptRef<const Network::Connection> connection) {
   if (downstream_headers.getSchemeValue() == Http::Headers::get().SchemeValues.Http) {
     return true;
   }
-  if (!connection.ssl()) {
+  if (connection && !connection->ssl()) {
```
jmarantz, line 57: nit: we've mostly been using `connection.has_value()` here, though if this compiles I suppose this must be correct.

source/common/http/filter_manager.cc
```diff
@@ -1335,7 +1337,9 @@ void FilterManager::setBufferLimit(uint32_t new_limit) {
 }
 
 void FilterManager::contextOnContinue(ScopeTrackedObjectStack& tracked_object_stack) {
-  tracked_object_stack.add(connection_);
+  if (connection_.has_value()) {
```
jmarantz, line 1340: this appears to be a semantic change. Were we previously adding potentially null-pointers to the stack?

source/common/router/upstream_request.cc
```diff
@@ -213,16 +213,18 @@ void UpstreamRequest::decodeTrailers(Http::ResponseTrailerMapPtr&& trailers) {
 void UpstreamRequest::dumpState(std::ostream& os, int indent_level) const {
   const char* spaces = spacesForLevel(indent_level);
   os << spaces << "UpstreamRequest " << this << "\n";
-  const auto addressProvider = connection().connectionInfoProviderSharedPtr();
+  if (connection()) {
```
jmarantz, line 216: again this appears to be a semantic change; was this previously a potential crasher?

source/extensions/common/wasm/context.cc
```diff
@@ -1108,9 +1108,9 @@ StreamInfo::StreamInfo* Context::getRequestStreamInfo() const {
 
 const Network::Connection* Context::getConnection() const {
   if (encoder_callbacks_) {
-    return encoder_callbacks_->connection();
+    return encoder_callbacks_->connection().ptr();
```
jmarantz, line 1111: in a follow-up will you change getConnection() to also return an OptRef?

source/extensions/upstreams/http/tcp/upstream_request.cc
```diff
@@ -52,9 +52,9 @@ Envoy::Http::Status TcpUpstream::encodeHeaders(const Envoy::Http::RequestHeaderM
   if (route_entry->connectConfig().has_value()) {
     Buffer::OwnedImpl data;
     auto& connect_config = route_entry->connectConfig().value();
-    if (connect_config.has_proxy_protocol_config()) {
+    if (connect_config.has_proxy_protocol_config() && upstream_request_->connection()) {
```
jmarantz, line 55: nit (your choice): use connection().has_value()

jmarantz (commented):
lgtm basically with a couple of nits I'll leave up to you, and a couple of small questions on semantics.

source/common/http/filter_manager.cc
alyssawilk: are you asking if I'm fixing a crashing bug in public?  No :-P  the interface allowed for null pointers, but the implementation with a hard-coded Network::Connection& did not.  For upstream filters, there may actually not be a downstream connection in the case of async requests, so for that we have to handle this case.

source/extensions/common/wasm/context.cc
alyssawilk: maybe?  :-)

jmarantz (approved):

PR #22556, opened 2022-08-04
repo: Release `1.22.5`

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #22622, opened 2022-08-09
connection handler: correct the handling of listener update

### LINKED ISSUE #22449: Envoy crashes when IPv4-compat listeners are used

soulxu:

Commit Message: connection handler: correct the handling of listener update
Additional Description:
To balance the IPv4 mapped address request to IPv4-mapped IPv6 address listener, inserting a mapping item into `tcp_listener_map_by_address_` using IPv4 mapped address as the key.

There can be a chance to override the mapping with the same key(the IP address key). The override cases are:
1. a listener using the same IPv4 address with a listener which has the same IPv4 mapped address.
2. an update of IPv4 mapped address listener.

Currently, the override checks only allow the stopped listener can be overridden, since it prevents the IPv4 listener was override by the listener which has the same IPv4-mapped address.
https://github.com/envoyproxy/envoy/blob/a866dceb91d0ae0d2f1b45f116463bd33d1a9310/source/server/connection_handler_impl.cc#L117-L118

But this will lead to a memory leak when an update happened on an IPv4-mapped IPv6 listener.

When updating a listener, the old listener mapping will be left in the `tcp_listener_map_by_address_` since the old listener will be stopped after updated new listener. After draining the old listener, the old listener will be stopped and removed, then the old listener's left mapping item in the `tcp_listener_map_by_address_` still reference the old listener resource, which leads to segment fault.

This PR changes to check the listening address, it also can prevent the IPv4 listener was override by the listener which has the same IPv4-mapped address, and also allow the IPv4-mapped listener to update the old mapping item.

Risk Level: high
Testing: unittest added
Docs Changes: n/a
Release Notes: an item for bug fixes added.
Fixes #22449

### DIFF
--- a/test/server/connection_handler_test.cc
+++ b/test/server/connection_handler_test.cc
@@ -1786,10 +1786,9 @@ TEST_F(ConnectionHandlerTest, MatchhIpv4WhenBothIpv4AndIPv6WithIpv4CompatFlag2) 
   EXPECT_CALL(*access_log_, log(_, _, _, _));
 }
 
-// This test the case of there is stopped "0.0.0.0" listener, then add
-// "::" listener with ipv4_compat. Ensure the ipv6 listener will take over the
-// ipv4 listener.
-TEST_F(ConnectionHandlerTest, AddIpv4MappedListenerAfterIpv4ListenerStopped) {
+// This test the case of an update for listener which listening
+// on an Ipv4-mapped Ipv6 address.
+TEST_F(ConnectionHandlerTest, UpdateIpv4MappedListener) {
   // Listener1 is response for redirect the connection.
   Network::TcpListenerCallbacks* listener_callbacks1;
   auto listener1 = new NiceMock<Network::MockListener>();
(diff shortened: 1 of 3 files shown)


test/server/connection_handler_test.cc
```diff
@@ -1786,10 +1786,9 @@ TEST_F(ConnectionHandlerTest, MatchhIpv4WhenBothIpv4AndIPv6WithIpv4CompatFlag2)
   EXPECT_CALL(*access_log_, log(_, _, _, _));
 }
 
-// This test the case of there is stopped "0.0.0.0" listener, then add
-// "::" listener with ipv4_compat. Ensure the ipv6 listener will take over the
-// ipv4 listener.
-TEST_F(ConnectionHandlerTest, AddIpv4MappedListenerAfterIpv4ListenerStopped) {
```
soulxu, line 1792: Note for the reviewer: after the fix, this case will be dropped. But it seems fine, it doesn't seem someone needs this case in the real world.

soulxu:
cc @mattklein123 @ggreenway

soulxu:
/wait

need double check

soulxu:
This is ready for review

soulxu:
gently ping @mattklein123 @ggreenway since want to include this into v1.22.5 release https://github.com/envoyproxy/envoy/pull/22556

mattklein123 (approved):
Thanks!

PR #22628, opened 2022-08-09
[balsa] Reject NUL in first line, headers and trailers.

bencebeky:

Commit Message: Reject NUL in first line, headers and trailers.
Additional Description: Protected by default-false runtime flag.
Risk Level: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/http1/balsa_parser.cc
+++ b/source/common/http/http1/balsa_parser.cc
@@ -163,8 +163,19 @@ void BalsaParser::OnBodyChunkInput(absl::string_view input) {
   connection_->bufferBody(input.data(), input.size());
 }
 
-void BalsaParser::OnHeaderInput(absl::string_view /*input*/) {}
-void BalsaParser::OnTrailerInput(absl::string_view /*input*/) {}
+void BalsaParser::OnHeaderInput(absl::string_view input) {
+  // NUL character is not allowed anywhere in the start line or in the header section.
+  if (input.find(absl::string_view("\0", 1)) != absl::string_view::npos) {
+    HandleError(BalsaFrameEnums::INVALID_HEADER_CHARACTER);
+  }
+}
+
+void BalsaParser::OnTrailerInput(absl::string_view input) {
+  // NUL character is not allowed anywhere in the trailers section.
+  if (input.find(absl::string_view("\0", 1)) != absl::string_view::npos) {
+    HandleError(BalsaFrameEnums::INVALID_HEADER_CHARACTER);
+  }
+}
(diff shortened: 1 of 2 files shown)


bencebeky:
Dianna, Biren, PTAL.

/assign @diannahu @birenroy

source/common/http/http1/balsa_parser.cc
```diff
@@ -163,8 +163,19 @@ void BalsaParser::OnBodyChunkInput(absl::string_view input) {
   connection_->bufferBody(input.data(), input.size());
 }
 
-void BalsaParser::OnHeaderInput(absl::string_view /*input*/) {}
-void BalsaParser::OnTrailerInput(absl::string_view /*input*/) {}
+void BalsaParser::OnHeaderInput(absl::string_view input) {
+  // NUL character is not allowed anywhere in the start line or in the header section.
+  if (input.find(absl::string_view("\0", 1)) != absl::string_view::npos) {
```
diannahu, line 168: very optional, here and below: Can use absl::StrContains()?

test/common/http/http1/codec_impl_test.cc
```diff
@@ -1282,6 +1280,38 @@ TEST_P(Http1ServerConnectionImplTest, HeaderMutateEmbeddedNul) {
   }
 }
 
+// Mutate the trailers with an HTTP POST with embedded NULs.
+// This should always be rejected.
+TEST_P(Http1ServerConnectionImplTest, TrailerMutateEmbeddedNul) {
+  if (parser_impl_ == ParserImpl::HttpParser) {
```
diannahu, line 1286: Interesting! I'm +1 on BalsaParser's strict improvement.

diannahu (approved):
Nice! I also see https://github.com/envoyproxy/envoy/pull/22629 - IIUC, NUL is a special case then? A little surprised but supportive!

source/common/http/http1/balsa_parser.cc
bencebeky: Good idea, done.  Incidentally I used a suboptimal signature, searching for a substring of length 1 instead of searching for a char.  (Both absl::string_view::find() and absl::StrContains() support both call signatures.)

bencebeky:
> Nice! I also see #22629 - IIUC, NUL is a special case then? A little surprised but supportive!

Oops, great catch!  No, NUL is not a special case, so none of this StrContains is needed.  I'm abandoning this PR and moving the test changes to #22629.

PR #22629, opened 2022-08-09
[balsa] Reject invalid characters.

bencebeky:

Commit Message: [balsa] Reject invalid characters.
Additional Description: Protected by default-false runtime flag.
Risk Level: low
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/common/http/http1/codec_impl_test.cc
+++ b/test/common/http/http1/codec_impl_test.cc
@@ -1142,11 +1142,6 @@ TEST_P(Http1ServerConnectionImplTest, HostHeaderTranslation) {
 // Ensures that requests with invalid HTTP header values are properly rejected
 // when the runtime guard is enabled for the feature.
 TEST_P(Http1ServerConnectionImplTest, HeaderInvalidCharsRejection) {
-  if (parser_impl_ == ParserImpl::BalsaParser) {
-    // TODO(#21245): Re-enable this test for BalsaParser.
-    return;
-  }
-
   initialize();
 
   MockRequestDecoder decoder;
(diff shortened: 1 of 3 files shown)


bencebeky:
Dianna, Biren, PTAL.

/assign @diannahu @birenroy

diannahu (dismissed):
Nice!

bencebeky:
PTAL.  I added some more test changes.

diannahu (approved):

test/common/http/http1/codec_impl_test.cc
```diff
@@ -1257,31 +1255,61 @@ TEST_P(Http1ServerConnectionImplTest, HeaderInvalidAuthority) {
...
+    EXPECT_FALSE(status.ok()) << n;
+    EXPECT_TRUE(isCodecProtocolError(status));
+    EXPECT_THAT(status.message(), testing::HasSubstr("http/1.1 protocol error:"));
+  }
+}
+
+// Mutate the trailers with an HTTP POST with embedded NULs.
+// This should always be rejected.
+TEST_P(Http1ServerConnectionImplTest, TrailerMutateEmbeddedNul) {
+  if (parser_impl_ == ParserImpl::HttpParser) {
```
KBaichoo, line 1284: Would this test not also work for HttpParser? If not, what's the reason? Seems like it might be valuable for both.

KBaichoo (commented):
also is this still draft? Thanks

test/common/http/http1/codec_impl_test.cc
bencebeky: It passes for http-parser for most values of `n` except for 6, 7, 21, 22, when the NUL character is injected around the two : characters.  Not sure why.

bencebeky:
> also is this still draft? Thanks

Oops, it's ready for review, marking it as such...

test/common/http/http1/codec_impl_test.cc
KBaichoo: I think that might be due to trailers being off unless `codec_settings_.enable_trailers_ = true;` is configured. Does it work if that's added at the start of the test?

KBaichoo (approved):
lgtm, thanks

PR #22656, opened 2022-08-11
[WIP]: Provide custom responses based on upstream reply

### LINKED ISSUE #21441: Custom Responses based on upstream reply
*Title*: *Provide the ability to customise upstream responses*

*Description*:
The ability to customise responses from the upstream (particularly errors) allows Envoy to provide customised web pages to the user which is particularly useful in multi-tenant environments where the default page might be perceived as jarring to the end user. Some functionality does currently exist using [LocalReplyConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-localreplyconfig), but this has the following limitations:
* The custom responses are always locally defined. This can lead to memory bloat issues in multi-tenant environments.
* [ResponseMappers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-msg-extensions-filters-network-http-connection-manager-v3-responsemapper) allow for only one filter per custom response.
* Advanced matching/filtering logic is not easy to specify.
* One cannot customise  2xx responses.

[This proposal](https://docs.google.com/document/d/1LOjQqMpt_uce_yUrf4cACWKpa7bSkA8MAAKY_xHtOys) aims to solve the above limitations by defining a new filter to customise upstream responses.

pradeepcrao:
Fixes #21441

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:


api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,76 @@
...
+// policies.
+message CustomResponse {
+  // Definition of the custom response to be provided to the downstream.
+  // [#next-free-field: 7]
+  message Response {
+    message RemoteDataSource {
+      config.core.v3.HttpUri http_uri = 1;
+
+      config.core.v3.RetryPolicy retry_policy = 2;
+    }
```
wbpcode, line 40: Seems it would be better if we make this a downgrade mechanism. If the upstream reply match specified pattern, then we can send the original request to another remote service or just return a local response.

So, IMO, we should get the whole response from the `remote`, not just a body.

wbpcode (commented):
Great. Thanks for your contribution. This is feature I expected for a long time.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
pradeepcrao: Hi @wbpcode, I have modified the implementation in a separate PR (#23084) that does this.

PR #22666, opened 2022-08-12
access log: Add context interface with access to listener properties

yanavlasov:
Additional Description:
We have internal access loggers that need to get listener properties. This refactor adds creator methods for such access logger with a new factory context interface.

Risk Level: Low,
Testing: Unit Test
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/envoy/server/factory_context.h
+++ b/envoy/server/factory_context.h
@@ -180,11 +180,34 @@ public:
 };
 
 /**
+ * Factory context for access loggers that need access to listener properties.
+ */
+class AccessLogFactoryContext : public virtual CommonFactoryContext {
+public:
+  /**
+   * @return Stats::Scope& the listener's stats scope.
+   */
+  virtual Stats::Scope& listenerScope() PURE;
+
+  /**
+   * @return const envoy::config::core::v3::Metadata& the config metadata associated with this
+   * listener.
+   */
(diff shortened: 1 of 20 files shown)


envoy/server/factory_context.h
```diff
@@ -179,12 +179,35 @@ class ServerFactoryContext : public virtual CommonFactoryContext {
   virtual envoy::config::bootstrap::v3::Bootstrap& bootstrap() PURE;
 };
 
+/**
+ * Factory context for access loggers that need access to listener properties.
+ */
+class AccessLogFactoryContext : public virtual CommonFactoryContext {
```
mattklein123, line 185: Up to you but I might call this `ListenerAccessLogFactoryContext` since I guess the idea is this is only usable when in listener context? For example it might not work for upstream access logs?

Also, I'm a little confused on how this is going to be used? I assume you are invoking this somehow from a private filter somewhere? Can you either add more notes so someone doesn't delete this later because it doesn't seem used or add a small integration test that shows how it could be used with a test filter? Up to you.

mattklein123 (changes requested):
LGTM with a small comment, thank you.

/wait

envoy/server/factory_context.h
yanavlasov: Done

mattklein123 (approved):
Thanks!

PR #21199, opened 2022-05-09
skywalking: use request path as operation name

wbpcode:
Commit Message: skywalking: use request path as operation name
Additional Description:

To pick up #20452 .

Risk Level: Low.
Testing: Added.
Docs Changes: n/a.
Release Notes: Added.
Platform Specific Features: n/a.

PR #22667, opened 2022-08-12
tracing: Add common tag name, to specify app protocol

tedli:
Commit Message: Add common tag name, to specify app protocol
Additional Description: #22594 Span of skywalking has a field called `spanLayer`, indicate it's http, rpc, db and so on. Currently this field is hardcoded to http as it's default value in ctor. Other telemetry impls may also have similar concept, so it's better introduce a common tag key,  to specify app protocol through `setTag`, to achieve protocol independent.

Why not add something like `setSpanLayer` to `Tracing::Span` interface:
\- Because, the parameter type varies and coupling with impl, e.g. `skywalking::v3::SpanLayer`. And there are impls without this concept, currently among tracers that envoy supported, only skywalking comes with this. On the other hand, opentelemetry comes with `kind` indicating it's server or client, xray comes with `direction` stands for it's ingress or egress, both of kind and direction could be set using `operation_name`. Meanwhile operation name of skywalking is set to `path` due to #21199 .

Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/tracers/skywalking/tracer.h
+++ b/source/extensions/tracers/skywalking/tracer.h
@@ -62,8 +62,8 @@ public:
       // TraceContext.protocol of http is parsed from http message, which value could be HTTP/1.1,
       // etc.
       layer = skywalking::v3::SpanLayer::Http;
-    } else {
-      skywalking::v3::SpanLayer_Parse(std::string(protocol), &layer);
+    } else if (!skywalking::v3::SpanLayer_Parse(std::string(protocol), &layer)) {
+      layer = skywalking::v3::SpanLayer::Unknown;
     }
     span_entity_->setSpanLayer(layer);
   }
--- a/source/extensions/tracers/skywalking/tracer.h
+++ b/source/extensions/tracers/skywalking/tracer.h
@@ -41,7 +41,8 @@ public:
    *
    * @return The unique ptr to the newly created span.
    */
-  Tracing::SpanPtr startSpan(absl::string_view name, TracingContextPtr tracing_context);
+  Tracing::SpanPtr startSpan(absl::string_view name, absl::string_view protocol,
+                             TracingContextPtr tracing_context);
 
 private:
   TraceSegmentReporterPtr reporter_;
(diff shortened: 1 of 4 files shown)


wbpcode:
The tracers can get protocol from `protocol` method of TracingContext. So I don't think we need get the protocol value by setTag.

tedli:
> The tracers can get protocol from `protocol` method of TraceContext. So I don't think we need get the protocol value by setTag.

Thanks for reply.

I noticed there was a protocol method in TraceContext, StreamInfo also had a protocol method. But both the two stands for HTTP1, HTTP2, HTTP3, and finally set to tag with the key `Tracing::Tags::get().HttpProtocol`.

If reuse the protocol method, we have to write code like this at somewhere,

```c++
// pseudocode
SpanPtr startSpan(......, TraceContext& conetxt, ......) {
  const auto& protocol = context.protocol();
  if (protocol == "HTTP1" || protocol == "HTTP2" || protocol == "HTTP3") {
    setTag(Tracing::Tags::get().HttpProtocol, protocol);
  } else if (protocol == "Dubbo") {
    setSpanLayer(...... /* Dubbo */);
  } else if (protocol == "Thrift") {
     ......
  }
}
```

If this is acceptable, I will correct my code.

wbpcode:
> If reuse the protocol method, we have to write code like this at somewhere

Sorry, I didn't get it. Why we need these code? why couldn't we just set spanlayer of skywalking with the value of context.protocol()?

tedli:
> > If reuse the protocol method, we have to write code like this at somewhere
> 
> Sorry, I didn't get it. Why we need these code? why couldn't we just set spanlayer of skywalking with the value of context.protocol()?

Because common callee method like startSpan or something else dosen't know the caller wants to set http protocol or span layer. So the if/else is needed.
Or it have to implement new tracer for each protocol, then we can set span layer directly with context.protocol().

wbpcode:
I mean just always call the `setSpanLayer` here base on the protocol() value. Some `if/else` is necessary but we needn't iterate every possible protocol values. Except HTTP, we can **even** set all other protocols' span layer to Unknow.

tedli:
Hi @wbpcode ,

I took your advice, getting protocol from trace context to set skywalking span layer. Please take a look. Thanks.

source/extensions/tracers/skywalking/tracer.h
```diff
@@ -51,16 +52,27 @@ using TracerPtr = std::unique_ptr<Tracer>;
...
+      // etc.
+      layer = skywalking::v3::SpanLayer::Http;
+    } else {
+      skywalking::v3::SpanLayer_Parse(std::string(protocol), &layer);
+    }
+    span_entity_->setSpanLayer(layer);
   }
-  Span(absl::string_view name, Span& parent_span, TracingContextPtr tracing_context,
-       Tracer& parent_tracer)
+  Span(absl::string_view name, const skywalking::v3::SpanLayer& span_layer, Span& parent_span,
```
wbpcode, line 70: nit: const ref is unnecessary here for simple enum.

wbpcode (commented):
Thanks. It's looks good overall. Only one nit comment.

And please add a simple test to cover the case where protocol value is not 'HTTP'.

source/extensions/tracers/skywalking/tracer.h
tedli: I removed const ref, and added test cases, please have a look.

wbpcode (approved):
LGTM. Thanks for your contribution.

PR #22672, opened 2022-08-12
http: support ignoring host port on filter processing

### LINKED ISSUE #22664: Add support for matching VirtualHost.Domains without `port` considered
*Title*: Add support for matching VirtualHost.Domains without `port` considered

*Description*:
We would like a way for `domains` to be matched ignoring port. For example, `domains: [foo.com]` would match `Host: foo.com:1234`, if we configured some setting.

Alternatives considered:

* Set the match to include the port (e.g: `domains: [foo.com, foo.com:1234]`). The issue with this is we don't know all the ports. The most common case for this is when we have multiple different network paths (example: going through Kubernetes LoadBalancer vs NodePort vs direct, etc) - the set could be unbounded
* Use `strip_port_mode` in HCM. This is not viable because it mutates the request; we want to ignore the port in matching but keep it in upstream requests.
* Use a suffix match (e.g: `domains: [foo.com, foo.com:*]`). This *almost* works. The edge case is if we have other prefix matches.

For example:
```yaml
- domains: [*.example.com:443] # match wildcard
- domains: [foo.example.com, foo.example.com:*] # match concrete
```

In this case, `Host: foo.example.com` would correctly match `concrete`. However, due to precedence rules of wildcards, `Host: foo.example.com:443` would end up matching `wildcard`

zhxie:

Commit Message: http: support ignoring host port on filter processing
Additional Description:
Risk Level: Low
Testing: WIP
Docs Changes: N/A
Release Notes: WIP
Fixes #22664

### DIFF
--- a/source/common/http/conn_manager_config.h
+++ b/source/common/http/conn_manager_config.h
@@ -181,6 +181,8 @@ enum class StripPortType {
   MatchingHost,
   // Removes any port from host/authority header.
   Any,
+  // Ignores any port from host/authority header.
+  Ignore,
   // Keeps the port in host/authority header as is.
   None
 };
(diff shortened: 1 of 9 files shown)


source/common/http/conn_manager_config.h
```diff
@@ -181,6 +181,8 @@ enum class StripPortType {
   MatchingHost,
   // Removes any port from host/authority header.
   Any,
+  // Ignores any port from host/authority header.
```
kyessenov, line 184: This needs to be explicit about the scope: where is it ignored?

PR #22690, opened 2022-08-13
tls: plumb host name to cert validator

danzh2010:
Commit Message: add host name from `SSL_get_servername()` into struct `ExtraValidationContext` to be passed into cert validation interface.

host name is needed for platform specific cert validation interface in Envoy Mobile.

Risk Level: low, new code is not used in default cert validator
Testing: new unit test
Docs Changes: N/A
Release Notes: N/A


jmarantz:
Hi @danzh2010 can you fix CI and then I'll assign? Thanks!

/wait

RyanTheOptimist (commented):
One question for you...

RyanTheOptimist (dismissed):

source/extensions/transport_sockets/tls/cert_validator/cert_validator.h
```diff
@@ -43,7 +43,9 @@ struct ValidationResults {
 class CertValidator {
 public:
   // Wraps cert validation parameters added from time to time.
-  struct ExtraValidationContext {};
+  struct ExtraValidationContext {
+    absl::string_view host_name;
```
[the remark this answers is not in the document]
danzh2010: > Do we ever validate a certificate without a hostname being present? 

We have been doing that for ssl transport socket. But if we use auto_sni, there is a filter state keyed with `Network::UpstreamSubjectAltNames` to specify sni list (from `Host` header) to override the ones in the config. And the overriding values get verified against the leaf cert in DefaultCertValidator. 

> I wonder if it would make sense to pass this in to doVerifyChain directly as an explicit argument?

done

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -513,11 +513,12 @@ ValidationResults ContextImpl::customVerifyCertChain(
     return {ValidationResults::ValidationStatus::Failed, SSL_AD_INTERNAL_ERROR, absl::nullopt};
   }
   ASSERT(cert_validator_);
+  const char* host_name = SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name);
   // Do not provide async callback here, but defer its creation to extended_socket_info if the
   // validation is async.
   ValidationResults result = cert_validator_->doVerifyCertChain(
       *cert_chain, nullptr, extended_socket_info, transport_socket_options, *SSL_get_SSL_CTX(ssl),
-      {}, SSL_is_server(ssl));
+      {}, SSL_is_server(ssl), (host_name == nullptr ? "" : host_name));
```
ggreenway, line 521: `absl::NullSafeStringView()` will do this check and is used elsewhere to wrap boringssl values.

ggreenway (changes requested):
Just one minor nit. Also needs a merge of main to fix the build.

/wait

source/extensions/transport_sockets/tls/context_impl.cc
danzh2010: done

ggreenway (approved):

PR #22693, opened 2022-08-13
Propagate tracestate header in the OpenTelemetry tracer

AlexanderEllis:
Commit Message: Propagate tracestate header in the OpenTelemetry tracer
Additional Description: OTLP Spans include the tracestate from the `tracestate` header. We can propagate that information from the headers to the Span. For an initial cut, we can just store whatever incoming tracestate header is in the span. In the future, it may make sense to add more handling to update the tracestate, but I don't believe we currently do any tracestate updates in Envoy.
Risk Level: Low
Testing: Basic tests

Part of #9958


test/extensions/tracers/opentelemetry/opentelemetry_tracer_impl_test.cc
```diff
@@ -115,14 +114,54 @@ TEST_F(OpenTelemetryDriverTest, ParseSpanContextFromHeadersTest) {
...
+  EXPECT_EQ(sampled_tracestate_entry.size(), 1);
+  EXPECT_EQ(sampled_tracestate_entry[0]->value().getStringView(), "test=foo");
+  const std::string request_yaml = R"(
+resource_spans:
+  resource:
+    attributes:
+      key: "service.name"
+      value:
+        string_value: "unknown_service:envoy"
+  instrumentation_library_spans:
```
moderation, line 136: Should this be `resource_spans:` ?

AlexanderEllis: I think as written it's OK, since `instrumentation_library_spans` is the child field of `resource_spans`, but now that I think about it, this should really be `scope_spans` with that [other PR](https://github.com/envoyproxy/envoy/pull/22694). This is the classic dueling PRs though, where with name change, this test would only pass once the other is submitted...

Happy to fix once the other PR is in, or if this one is merged first I can update this code in the other PR.

AlexanderEllis: Speaking of that PR :) Updated this with `scope_spans` to match the new behavior. Thanks!

moderation: :+1: Not really related to this PR but interested in how to leverage the config at https://github.com/envoyproxy/envoy/blob/main/test/extensions/tracers/opentelemetry/opentelemetry_tracer_impl_test.cc#L337-L357. Not clear from the test code where this is configured.

AlexanderEllis: I definitely think it could use a little refactoring to be clearer (especially the bytes setting). I was currently using it to verify the final span that the Tracer exports, but so far I've only added the exact proto matching with this config for a few of the tests (you'll see `sendMessageRaw_(_, _)` in others instead of `sendMessageRaw_(Grpc::ProtoBufferEqIgnoreRepeatedFieldOrdering(request_proto), _)`).  For legibility, I think it would also be helpful to only test the specific fields for each test to help reduce the busyness of each expectation.

source/extensions/tracers/opentelemetry/span_context_extractor.cc
```diff
@@ -87,7 +91,20 @@ absl::StatusOr<SpanContext> SpanContextExtractor::extractSpanContext() {
...
   char decoded_trace_flags = absl::HexStringToBytes(trace_flags).front();
   bool sampled = (decoded_trace_flags & 1);
-  SpanContext span_context(version, trace_id, parent_id, sampled);
+
+  // If a tracestate header is received without an accompanying traceparent header,
+  // it is invalid and MUST be discarded.
+  // See https://www.w3.org/TR/trace-context/#processing-model-for-working-with-trace-context
+  absl::string_view tracestate = "";
+  // TODO: what if tracestate is in multiple headers? Spec says we should combine, but Envoy
+  // currently only surfaces the first seen.
```
htuch, line 100: See `get`, which returns a `GetResult` with all header values.

test/extensions/tracers/opentelemetry/span_context_extractor_test.cc
```diff
@@ -144,6 +144,27 @@ TEST(SpanContextExtractorTest, ThrowsExceptionWithAllZeroParentId) {
...
+  absl::StatusOr<SpanContext> span_context = span_context_extractor.extractSpanContext();
+
+  EXPECT_OK(span_context);
+  EXPECT_TRUE(span_context->tracestate().empty());
+}
+
+TEST(SpanContextExtractorTest, ExtractSpanContextWithTracestate) {
+  Http::TestRequestHeaderMapImpl request_headers{
+      {"traceparent", fmt::format("{}-{}-{}-{}", version, trace_id, parent_id, trace_flags)},
+      {"tracestate", "sample-tracestate"}};
```
htuch, line 160: Is there one for missing traceparent?

source/extensions/tracers/opentelemetry/tracer.cc
```diff
@@ -174,6 +176,9 @@ Tracing::SpanPtr Tracer::startSpan(const Tracing::Config& config, const std::str
   new_span.setId(Hex::uint64ToHex(span_id));
   // Respect the previous span's sampled flag.
   new_span.setSampled(previous_span_context.sampled());
+  if (!previous_span_context.tracestate().empty()) {
+    new_span.setTracestate(std::string{previous_span_context.tracestate()});
```
htuch, line 180: Is this tested?

htuch (changes requested):
Looks good, thanks, just a few minor feedback items.
/wait

source/extensions/tracers/opentelemetry/tracer.cc
AlexanderEllis: Yup, via `ParseSpanContextFromHeadersTest`. 

In `Driver::startSpan` in `opentelemetry_tracer_impl.cc`, if there is a propagation header present, it'll call this fn, `Tracer::startSpan` with the previous span context. Since we're manually adding the `traceparent` propagation header (and `tracestate`) in the  the test and then calling `driver_->startSpan`, the `tracestate` in the header is passed through in the `previous_span_context` here.

test/extensions/tracers/opentelemetry/span_context_extractor_test.cc
AlexanderEllis: Ah there isn't, good call! Technically it was already covered via the missing traceparent test above, but I think it's worthwhile to explicitly add one for it for legibility. I also fleshed out the comment in the extractor, mentioning that we are already checking for the existence of the traceparent header.

source/extensions/tracers/opentelemetry/span_context_extractor.cc
AlexanderEllis: Thanks for that pointer! Did a little digging, and it looks like `ConnectionManagerImpl::ActiveStream::traceRequest` calls `startSpan` with its `RequestHeaderMapPtr request_headers_` (the `RequestHeaderMap` containing the `get`).  `Driver::startSpan` is defined with just `Tracing::TraceContext` though, which only has `getByKey` and `forEach`.

With that in mind, I added some handling for this (and a test) with iteration over all the headers. Doesn't seem very efficient to me, but I think this works for the multiple headers case. Happy to improve if you have any suggestions 👍

htuch: Maybe fix `TraceContext` to deal with the appropriate header-specific aggregation behavior. I think any tracer interacting with tracestate will need this behavior. I think in most places that we see header get-by-key and the value is singular and not aggregated, there is some code smell that we haven't fully thought through the valid wire combinations.

moderation:
This ready to merge @htuch @AlexanderEllis ?

mattklein123 (approved):
Thanks will merge as it looks like comments have been addressed.

PR #22695, opened 2022-08-14
tracing: refactor tracing, tracer provider proto from http to common

tedli:
Commit Message: Refactor tracing, tracer provider proto from http to common
Additional Description: Tracing should be protocol independent. The tracing and tracer provider proto can be used by other proxy, so move from http to common. #22594
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jmarantz:
@tedli can you fix format? Thanks!

/wait

wbpcode:
Thanks for this contribution. It is great to create an common proto for tracing. 

**But we shouldn't edit existing API directly for the compatibility.**

Typically, we can create this new common proto for new extensions and may update the existing API in the next version.

jmarantz:
@tedli please don't force-push once a PR is in review. Please read https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#submitting-a-pr for details, especially the 14th (I think) bullet in that section.

/wait

mattklein123:
I'm on the fence on this. I agree it's better here, but at the same time I'm not sure duplicating the protos is worth it as if we are going to do this we should deprecate the HTTP ones and refer to this one.

For now can you just refer to the HTTP one inside dubbo? It's not great but it seems OK to me for now? I'm mostly just trying to weigh the churn vs. the benefit. Thoughts?

/wait

tedli:
> I'm on the fence on this. I agree it's better here, but at the same time I'm not sure duplicating the protos is worth it as if we are going to do this we should deprecate the HTTP ones and refer to this one.
> 
> For now can you just refer to the HTTP one inside dubbo? It's not great but it seems OK to me for now? I'm mostly just trying to weigh the churn vs. the benefit. Thoughts?
> 
> /wait

Thanks for reply.
I echo referring the HTTP one in dubbo. I will rollback, only keep a tracing field in dubbo proxy which type reuse the http one, and go to finish this https://github.com/envoyproxy/envoy/pull/22594

suniltheta:
This PR title & Commit Message: `Refactor tracing, tracer provider proto from http to common` can be updated now to something else more appropriate?

Also wondering if the `protodoc-title` [here](https://github.com/envoyproxy/envoy/blob/release/v1.23/api/envoy/config/trace/v3/http_tracer.proto#L20-L23) can also be updated to indicate that the dubbo proxy will also use it.

mattklein123:
Agree on updating the title and description, and at that point why not just do the code also? Thank you.

/wait

PR #22207, opened 2022-07-14
Full Set Of Changes For Pattern Match and Rewrite

silverstar194:
Commit Message:
Full Set Of Changes For Pattern Match and Rewrite

Additional Description:
This PR will implement issue detailed here and described below: https://github.com/envoyproxy/envoy/issues/7763

### Match Patterns and Templates
Wildcard support based on match patterns and templates.

- A match pattern matches an incoming URL path. 
Match patterns support glob operators to match URL text and variable definitions to bind matched text to names.

- Template patterns are used to re-write URLs.
Template patterns build new URLs and may reference variables bound by a match pattern.

### Match Examples

- /\*\*.m3u8 would match /foo.m3u8 and /foo/bar.m3u8.
- /{dir_name}/\*.ts would match /example/file.ts and bind dir_name="example" for a later template match to use.
- /{dir_name}/\*\*.ts would match /example/path/file.ts and bind dir_name="example" for a later template match to use. This would also match /example/.ts, which may or may not be a desired behavior.
- /{path=v1/\*}/{file=\*.ts} would match /v1/example/movie.ts (binding path="v1/example" and file="movie"), but would not match /v0/example/movie.ts.

See post for full details and example:
https://github.com/envoyproxy/envoy/issues/7763#issuecomment-825053497

Risk Level:
Testing:
Unit tests. Both both internal matching/rewrite library and config/data plane changes.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #22703, opened 2022-08-15
URL Rewrite: Match and Rewrite Library

silverstar194:
Commit Message:
URL Rewrite: Match and Rewrite Library

Additional Description:
This PR will add library to implement issue detailed here and described below: https://github.com/envoyproxy/envoy/issues/7763

Full set of changes to implement feature are found here: https://github.com/envoyproxy/envoy/pull/22207

Match Patterns and Templates
Wildcard support based on match patterns and templates.

A match pattern matches an incoming URL path.
Match patterns support glob operators to match URL text and variable definitions to bind matched text to names.

Template patterns are used to re-write URLs.
Template patterns build new URLs and may reference variables bound by a match pattern.

Match Examples
/**.m3u8 would match /foo.m3u8 and /foo/bar.m3u8.
/{dir_name}/*.ts would match /example/file.ts and bind dir_name="example" for a later template match to use.
/{dir_name}/**.ts would match /example/path/file.ts and bind dir_name="example" for a later template match to use. This would also match /example/.ts, which may or may not be a desired behavior.
/{path=v1/*}/{file=*.ts} would match /v1/example/movie.ts (binding path="v1/example" and file="movie"), but would not match /v0/example/movie.ts.
See post for full details and example:
https://github.com/envoyproxy/envoy/issues/7763#issuecomment-825053497

Risk Level:
Testing:
Unit tests. Both both internal matching/rewrite library and config/data plane changes.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


silverstar194:
Full set of changes to implement rewrite and match are here:
https://github.com/envoyproxy/envoy/pull/22207

tools/code_format/check_format.py
```diff
@@ -1149,10 +1150,11 @@ def check_visibility(error_messages):
             "':(exclude)source/extensions/filters/network/common/BUILD' "
             "':(exclude)source/extensions/transport_sockets/common/BUILD' "
             "':(exclude)source/extensions/udp_packet_writer/default/BUILD' "
-            "':(exclude)source/extensions/udp_packet_writer/gso/BUILD' ")
+            "':(exclude)source/extensions/udp_packet_writer/gso/BUILD' "
+            "':(exclude)source/extensions/path/match/pattern_template/BUILD' ")
         command = (
             "git diff $(tools/git/last_github_commit.sh) -- source/extensions/* %s |grep '+.*visibility ='"
```
silverstar194, line 1156: Fix bug in exclude list generation

tools/code_format/check_format.py
```diff
@@ -7,17 +7,21 @@
...
 # The way this script is currently used (ie no bazel) it relies on system deps.
 # As `pyyaml` is present in `envoy-build-ubuntu` it should be safe to use here.
 import yaml
+from functools import cached_property
+from typing import Callable
+from typing import Dict
+from typing import List
+from typing import Pattern
+from typing import Tuple
+from typing import Union
```
RyanTheOptimist, line 24: The previous version of these imports seems more concise. Did format fix this or, is this a design choice.

tools/code_format/check_format.py
RyanTheOptimist: Can you say more about what the bug here is? I think it'd probably be best to land this stand-alone since it's unrelated to this change.

tools/code_format/config.yaml
```diff
@@ -295,3 +295,4 @@ visibility_excludes:
 - source/extensions/transport_sockets/common/BUILD
 - source/extensions/udp_packet_writer/default/BUILD
 - source/extensions/udp_packet_writer/gso/BUILD
+- source/extensions/path/pattern_template_lib/BUILD
```
RyanTheOptimist, line 298: Presumably we don't need this in the PR since it's not yet being used elsewhere?

test/per_file_coverage.sh
```diff
@@ -70,6 +70,8 @@ declare -a KNOWN_LOW_COVERAGE=(
 "source/extensions/io_socket:96.2"
 "source/extensions/io_socket/user_space:96.2"
 "source/extensions/network/dns_resolver/getaddrinfo:96.3"
+"source/extensions/path/pattern_template_lib:93.2"
```
RyanTheOptimist, line 73: Normally we should be able to get ~100% code coverage from unit tests. Do you know which lines are not being covered?

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+#include "absl/types/variant.h"
+#include "re2/re2.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace PatternTemplate {
+
+namespace Internal {
+
+using Literal = absl::string_view;
```
RyanTheOptimist, line 22: nit: Comment, please. (though it's reasonable readable as is)

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Determines what operations to use on the input pattern segment
+ */
+enum class Operator { KPathGlob, KTextGlob };
```
RyanTheOptimist, line 27: nit: Remove the "K" prefix

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+struct RewriteSegment {
+
+  // A string that will be concatenated for rewrite.
+  absl::string_view literal_;
+
+  // Represents an index into the RE2 capture which value should be used
+  // to construct the rewritten URL. Note that the index should be greater
+  // than 0 as 0 index into the whole match RE2 pattern.
+  int capture_index_;
+};
```
RyanTheOptimist, line 42: This struct seems to represent a captured segment, hence the `capture_index`. Is it really Rewrite specific? If not, perhaps we could call this CapturedSegment.

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+struct ParsedUrlPattern {
+  std::vector<ParsedSegment> parsed_segments_;
+  absl::optional<absl::string_view> suffix_;
+  absl::flat_hash_set<absl::string_view> captured_variables_;
+
+  std::string debugString() const;
+};
+
+/**
+ * Check if literal is valid
```
RyanTheOptimist, line 71: What would make a literal invalid? (looking at the .cc file, I'm not sure). Do we have unit test coverage for this method?

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+ */
+bool isValidLiteral(absl::string_view literal);
+
+/**
+ * Check if rewrite literal is valid
+ */
+bool isValidRewriteLiteral(absl::string_view literal);
```
RyanTheOptimist, line 78: Same comment here. What would make an invalid rewrite literal? Do we have test coverage?

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Check if indent is valid
```
RyanTheOptimist, line 81: nit: I think this comment needs to be fixed

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+ */
+bool isValidVariableName(absl::string_view indent);
```
RyanTheOptimist, line 83: Do we have test coverage for this method?

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Used by the following Consume{Literal.Operator,Variable} functions
+ * in the return value. The functions would take the given pattern,
+ * parse what it can into |parsed_value| and return the unconsumed
+ *  portion of the pattern in |unconsumed_pattern|.
```
RyanTheOptimist, line 89: Nice comment!

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+      : parsed_value_(val), unconsumed_pattern_(pattern) {}
+
+  T parsed_value_;
+  absl::string_view unconsumed_pattern_;
+};
+
+/**
+ * Converts input pattern to ParsedResult<Literal>
+ */
+absl::StatusOr<ParsedResult<Literal>> consumeLiteral(absl::string_view pattern);
```
RyanTheOptimist, line 102: I'm not sure I understand why these methods are called "consume" instead of "covert" (to match the comment) or "parse" to match the return value. Looking at the implementation, I *think* it's attempting to parse a literal off of the front of `view_pattern`. Is that right? (Same question for the below).

Do we have test coverage?

source/extensions/path/pattern_template_lib/pattern_template_internal.h
```diff
@@ -0,0 +1,147 @@
...
+ */
+absl::StatusOr<ParsedResult<Variable>> consumeVariable(absl::string_view pattern);
+
+/**
+ * Converts input pattern to ParsedUrlPattern
+ */
+absl::StatusOr<ParsedUrlPattern> parseURLPatternSyntax(absl::string_view url_pattern);
+
+/**
+ * Converts Literal to std::string
```
RyanTheOptimist, line 120: I'm not sure I understand what this comment means. I *think* the idea is that these methods are providing a human readable version of the `pattern`? Or perhaps it's a regex syntax string representation? If the latter, perhaps the comment should say something like "Convert the Literal in `pattern` to a regex string representation?" (or some such)

source/extensions/path/pattern_template_lib/pattern_template.h
```diff
@@ -0,0 +1,58 @@
...
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace PatternTemplate {
+
+enum class RewriteStringKind { Variable, Literal };
+
+struct RewritePatternSegment {
```
RyanTheOptimist, line 17: What's the difference between this struct, RewriteSegment in pattern_template_internal.h and PatternTemplateRewriteSegments from the .proto? I'm mildly surprised that we need 3 such similarly name structs.

source/extensions/path/pattern_template_lib/pattern_template.h
```diff
@@ -0,0 +1,58 @@
...
+
+// Parses the specified pattern into a sequence of segments (which are
+// either literals or variable names).
+absl::StatusOr<std::vector<RewritePatternSegment>> parseRewritePattern(absl::string_view pattern);
+
+// Returns the parsed Url rewrite pattern to be used by
+// RewriteURLTemplatePattern()  |capture_regex| should
+// be the regex generated by ConvertURLPatternSyntaxToRegex().
+absl::StatusOr<envoy::extensions::pattern_template::PatternTemplateRewriteSegments>
+parseRewritePattern(absl::string_view pattern, absl::string_view capture_regex);
```
RyanTheOptimist, line 37: So I think I understand that the previous `parseRewritePattern` takes a string representing a template pattern, and splits it into literals and variables. But I don't think I understand what this what does. In particular, I don't think I understand what capture_regex represents.

source/extensions/path/pattern_template_lib/pattern_template.h
```diff
@@ -0,0 +1,58 @@
...
+
+// Returns if provided rewrite pattern is valid
+absl::Status isValidPathTemplateRewritePattern(const std::string& path_template_rewrite);
+
+// Returns if path_template and rewrite_template have valid variables.
+// Every variable in rewrite MUST be present in match
```
RyanTheOptimist, line 43: nit: add a trailing ".", please.

source/extensions/path/pattern_template_lib/pattern_template.h
```diff
@@ -0,0 +1,58 @@
...
+// For example:
+// Match: /foo/{bar}/{var} and Rewrite: /goo/{var} is valid
+// Match: /foo/{bar} and Rewrite: /goo/{bar}/{var} is invalid. Match is missing {var}.
+absl::Status isValidSharedVariableSet(const std::string& path_template_rewrite,
+                                      const std::string& capture_regex);
```
RyanTheOptimist, line 48: So the comment talks about "path_template" and "rewrite_template" but the arguments are "path_template_rewrite" and "capture_regex".

source/extensions/path/pattern_template_lib/pattern_template.h
```diff
@@ -0,0 +1,58 @@
...
+
+absl::Status isValidMatchPattern(const std::string match_pattern);
```
RyanTheOptimist, line 50: Comment these two methods, please.

RyanTheOptimist (commented):
Copying this piece of comment from #22207. One high level comment is about the "pattern template". If I'm reading all this code correctly I think it's really all about "URI Templates" from RFC 6570. If so, I think it would be a great idea to use that naming instead of "pattern template" which is a bit vague. Should we consider naming this "uri_template_lib" instead of "pattern_template_lib"?

silverstar194:
> Copying this piece of comment from #22207. One high level comment is about the "pattern template". If I'm reading all this code correctly I think it's really all about "URI Templates" from RFC 6570. If so, I think it would be a great idea to use that naming instead of "pattern template" which is a bit vague. Should we consider naming this "uri_template_lib" instead of "pattern_template_lib"?

This pattern matching may later be extended to headers or cookies so I would prefer to keep it more generic.

tools/code_format/check_format.py
silverstar194: It was format fixed. Happy to revert if previous was closer to Enovy's style

tools/code_format/config.yaml
silverstar194: Removed

test/per_file_coverage.sh
silverstar194: There are failure cases that are very difficult to cover, such as regex engine failure. This coverage is the same as internal google + a few extra I added as well.

source/extensions/path/pattern_template_lib/pattern_template_internal.h
silverstar194: Added

source/extensions/path/pattern_template_lib/pattern_template.h
silverstar194: done

source/extensions/path/pattern_template_lib/pattern_template.h
silverstar194: added

test/per_file_coverage.sh
RyanTheOptimist: If you look at the coverage report here:
https://storage.googleapis.com/envoy-pr/c8b8f01/coverage/source/extensions/path/pattern_template_lib/pattern_template.cc.gcov.html
and here:
https://storage.googleapis.com/envoy-pr/c8b8f01/coverage/source/extensions/path/pattern_template_lib/pattern_template_internal.cc.gcov.html
It looks like increasing the unit test coverage to get to 100% should be quite easy. For example, in pattern_template.cc there are no tests for:

isValidMatchPattern
isValidPathTemplateRewritePattern
isValidSharedVariableSet

And there is some missing coverage for some of the error cases. Please add unit tests for all methods.

source/extensions/path/pattern_template_lib/pattern_template.h
silverstar194: On further look this is duplicated. I've removed the struct completely

tools/code_format/check_format.py
silverstar194: Let me split this out.. It's a python concatenation issue. Currently the exclude list looks like `["item",  ... "itemN"]` it should be without `[` `]` like`"item",  ... "itemN"`

silverstar194: This was fixed in main. Merged in main changes.

tools/code_format/check_format.py
silverstar194: reverted

tools/code_format/check_format.py
silverstar194: PR for bug fix: https://github.com/envoyproxy/envoy/pull/22728

source/extensions/path/pattern_template_lib/pattern_template.h
```diff
@@ -30,25 +30,28 @@ absl::StatusOr<std::string> convertURLPatternSyntaxToRegex(absl::string_view url
...
-                                      const std::string& capture_regex);
+absl::Status isValidSharedVariableSet(absl::string_view path_template_rewrite,
+                                      absl::string_view capture_regex);
 
-absl::Status isValidMatchPattern(const std::string match_pattern);
+// Returns is the match_pattern is valid.
+// Validation attempts to parse pattern into literals and variables.
+absl::Status isValidMatchPattern(absl::string_view match_pattern);
 
+// Concatenates literals and extracts variable values to form the final rewritten url.
```
RyanTheOptimist, line 54: Adding an example here would be helpful. Presumably, `url` is a raw URL which has not had any substitutions. Like "https://example.com/foo/bar". Then `capture_regex` is something like "/foo/{bar}", and rewrite_pattern is ...? i'm not quite sure, but and example would be helpful.

RyanTheOptimist (commented):
Looks like there are still several unaddressed comments. In particular, I don't see unit tests for isValidRewriteLiteral, or really any of the methods in pattern_template_internal.h

RyanTheOptimist:
> > Copying this piece of comment from #22207. One high level comment is about the "pattern template". If I'm reading all this code correctly I think it's really all about "URI Templates" from RFC 6570. If so, I think it would be a great idea to use that naming instead of "pattern template" which is a bit vague. Should we consider naming this "uri_template_lib" instead of "pattern_template_lib"?
> 
> This pattern matching may later be extended to headers or cookies so I would prefer to keep it more generic.

The challenge is that "pattern_template" is not very descriptive. If it's really all about RFC 6570 at the moment, then uri_template_lib would be a good name. Alternatively, the PR description talks about "match and rewrite" which would also be a good name. Do we have a specific plan to use these methods for headers and cookies, or is this hypothetical at this point?

silverstar194:
> > > Copying this piece of comment from #22207. One high level comment is about the "pattern template". If I'm reading all this code correctly I think it's really all about "URI Templates" from RFC 6570. If so, I think it would be a great idea to use that naming instead of "pattern template" which is a bit vague. Should we consider naming this "uri_template_lib" instead of "pattern_template_lib"?
> > 
> > 
> > This pattern matching may later be extended to headers or cookies so I would prefer to keep it more generic.
> 
> The challenge is that "pattern_template" is not very descriptive. If it's really all about RFC 6570 at the moment, then uri_template_lib would be a good name. Alternatively, the PR description talks about "match and rewrite" which would also be a good name. Do we have a specific plan to use these methods for headers and cookies, or is this hypothetical at this point?

It's hypothetical at this point. I am ok with using uri_template_lib

RyanTheOptimist:
> > > > Copying this piece of comment from #22207. One high level comment is about the "pattern template". If I'm reading all this code correctly I think it's really all about "URI Templates" from RFC 6570. If so, I think it would be a great idea to use that naming instead of "pattern template" which is a bit vague. Should we consider naming this "uri_template_lib" instead of "pattern_template_lib"?
> > > 
> > > 
> > > This pattern matching may later be extended to headers or cookies so I would prefer to keep it more generic.
> > 
> > 
> > The challenge is that "pattern_template" is not very descriptive. If it's really all about RFC 6570 at the moment, then uri_template_lib would be a good name. Alternatively, the PR description talks about "match and rewrite" which would also be a good name. Do we have a specific plan to use these methods for headers and cookies, or is this hypothetical at this point?
> 
> It's hypothetical at this point. I am ok with using uri_template_lib

SGTM!

RyanTheOptimist:
Looks like format is failing because of:

```
ERROR: New directory extensions/path/uri_template_lib appears to not have owners in CODEOWNERS
ERROR: New directory extensions/path/uri_template_lib/proto appears to not have owners in CODEOWNERS
```

source/extensions/path/pattern_template_lib/pattern_template_internal.h
silverstar194: fixed

source/extensions/path/pattern_template_lib/pattern_template_internal.h
silverstar194: Updated comments. Unit tests are included

source/extensions/path/pattern_template_lib/pattern_template_internal.h
silverstar194: Yes.

source/extensions/path/pattern_template_lib/pattern_template_internal.h
silverstar194: You're correct it strips out the literal from pattern. It's consume but it's destructive and "consumes" the literal from string.

There is test coverage

source/extensions/path/pattern_template_lib/pattern_template_internal.h
silverstar194: Added to comment

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  }
+  std::vector<absl::string_view> parts = absl::StrSplit(pattern.substr(1), absl::MaxSplits('}', 1));
+  if (parts.size() != 2) {
+    return absl::InvalidArgumentError("Unmatched variable bracket");
+  }
+  absl::string_view unconsumed_pattern = parts[1];
+
+  // Parse the actual variable pattern, starting with the variable name.
+  std::vector<absl::string_view> var_parts = absl::StrSplit(parts[0], absl::MaxSplits('=', 1));
+  if (!isValidVariableName(var_parts[0])) {
```
RyanTheOptimist, line 174: nit: avoid abbreviation. Maybe "variable_parts" or just "parts"? (here and below for "var_")

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+      }
+      var_patt = var_patt.substr(1);
+    }
+  }
+
+  return ParsedResult<Variable>(var, unconsumed_pattern);
+}
+
+absl::StatusOr<absl::flat_hash_set<absl::string_view>>
+gatherCaptureNames(struct ParsedUrlPattern pattern) {
```
RyanTheOptimist, line 218: nit: should `pattern` be passed by const&, perhaps?

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  absl::flat_hash_set<absl::string_view> captured_variables;
+
+  for (const ParsedSegment& segment : pattern.parsed_segments_) {
+    if (!absl::holds_alternative<Variable>(segment)) {
+      continue;
+    }
+    if (captured_variables.size() >= kPatternMatchingMaxVariablesPerUrl) {
+      return absl::InvalidArgumentError("Exceeded variable count limit");
+    }
+    absl::string_view var_name = absl::get<Variable>(segment).var_name_;
```
RyanTheOptimist, line 228: nit: instead of "var_name" just "name"

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+
+    if (var_name.size() < kPatternMatchingMinVariableNameLen ||
+        var_name.size() > kPatternMatchingMaxVariableNameLen) {
+      return absl::InvalidArgumentError("Invalid variable length");
```
RyanTheOptimist, line 232: nit: "variable name length", I think

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+    if (captured_variables.contains(var_name)) {
+      return absl::InvalidArgumentError("Repeated variable name");
+    }
+    captured_variables.emplace(var_name);
+  }
+
+  return captured_variables;
+}
+
+absl::Status validateNoOperatorAfterTextGlob(struct ParsedUrlPattern pattern) {
```
RyanTheOptimist, line 243: I don't see this in a .h file (or the previous method). Should it be moved into an anon namespace?

nit: should pattern be passed by const&, perhaps?

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  }
+  return absl::OkStatus();
+}
+
+absl::StatusOr<ParsedUrlPattern> parseURLPatternSyntax(absl::string_view url_pattern) {
+  struct ParsedUrlPattern parsed_pattern;
+
+  static const LazyRE2 printable_regex = {"^/[[:graph:]]*$"};
+  if (!RE2::FullMatch(toStringPiece(url_pattern), *printable_regex)) {
+
```
RyanTheOptimist, line 279: nit: remove the empty line

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  // Consume the leading '/'
+  url_pattern = url_pattern.substr(1);
+
+  // Do the initial lexical parsing.
+  while (!url_pattern.empty()) {
+    ParsedSegment segment;
+    if (url_pattern[0] == '*') {
+
```
RyanTheOptimist, line 290: nit: remove all these empty lines after the ifs.

RyanTheOptimist (commented):
I think I managed to get through all of the non-test code.

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: Done

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: done

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: Added to header and used const&

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
+// Returns the regex pattern that is equivalent to the given url_pattern.
+// Used in the config pipeline to translate user given url pattern to
+// the safe regex Envoy can understand. Strips away any variable captures.
+absl::StatusOr<std::string> convertURLPatternSyntaxToRegex(absl::string_view url_pattern);
+
+// Parses the specified pattern into a sequence of segments (which are
+// either literals or variable names).
+absl::StatusOr<std::vector<RewritePatternSegment>> parseRewritePattern(absl::string_view pattern);
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+
+struct RewritePatternSegment {
+  RewritePatternSegment(absl::string_view value, RewriteStringKind kind)
+      : value_(value), kind_(kind) {}
+  absl::string_view value_;
+  RewriteStringKind kind_;
+};
+
+// Returns the regex pattern that is equivalent to the given url_pattern.
+// Used in the config pipeline to translate user given url pattern to
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+// the safe regex Envoy can understand. Strips away any variable captures.
```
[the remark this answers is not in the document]
silverstar194: Removed the misguided comment

source/extensions/path/uri_template_lib/proto/uri_template_rewrite_segments.proto
```diff
@@ -0,0 +1,29 @@
...
+  message RewriteSegment {
+    oneof segment_type {
+      // Represents a segment of the rewritten URL, including any path segments,
+      // slash and prefix.
+      string literal = 1;
+
+      // Represents an index into the RE2 capture which value should be used
+      // to construct the rewritten URL. Note that the index should be greater
+      // than 0 as 0 index into the whole match RE2 pattern.
+      int32 var_index = 2;
```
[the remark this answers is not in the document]
silverstar194: Renamed

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace UriTemplate {
+
+enum class RewriteStringKind { Variable, Literal };
+
+struct RewritePatternSegment {
```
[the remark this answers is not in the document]
silverstar194: RewritePatternSegment holds the actual string values for each segment.
UriTemplateRewriteSegments holds the indexes ONLY or literals.

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+// Used in the config pipeline to translate user given url pattern to
+// the safe regex Envoy can understand. Strips away any variable captures.
+absl::StatusOr<std::string> convertURLPatternSyntaxToRegex(absl::string_view url_pattern);
+
+// Parses the specified pattern into a sequence of segments (which are
+// either literals or variable names).
+absl::StatusOr<std::vector<RewritePatternSegment>> parseRewritePattern(absl::string_view pattern);
+
+// Returns the parsed Url rewrite pattern and processes variables to be used by
+// RewriteURLTemplatePattern() and in rewrite.
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+// |capture_regex| should be the regex generated by ConvertURLPatternSyntaxToRegex().
+absl::StatusOr<envoy::extensions::uri_template::UriTemplateRewriteSegments>
+parseRewritePattern(absl::string_view pattern, absl::string_view capture_regex);
```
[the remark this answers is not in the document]
silverstar194: I think we need the regex but I'm not 100% on the reason why. I didn't write the core of this and I'm hesitant to start rewriting large parts as there was a lot of thought but into the original design.

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+
+// Returns if provided rewrite pattern is valid
```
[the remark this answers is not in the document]
silverstar194: added

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+absl::Status isValidPathTemplateRewritePattern(absl::string_view path_template_rewrite);
```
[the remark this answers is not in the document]
silverstar194: Changed

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template.cc
```diff
@@ -0,0 +1,165 @@
...
+
+using Internal::ParsedUrlPattern;
+
+#ifndef SWIG
+// Silence warnings about missing initializers for members of LazyRE2.
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+#endif
+
+absl::StatusOr<std::string> convertURLPatternSyntaxToRegex(absl::string_view url_pattern) {
+  absl::StatusOr<ParsedUrlPattern> status = Internal::parseURLPatternSyntax(url_pattern);
```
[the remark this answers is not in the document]
silverstar194: I think I want to keep as-is. It provides some separation from the internal implementation

source/extensions/path/uri_template_lib/uri_template.cc
```diff
@@ -0,0 +1,165 @@
...
+  if (!status.ok()) {
+    return status.status();
+  }
+  return Internal::toRegexPattern(*status);
+}
+
+absl::StatusOr<std::vector<RewritePatternSegment>> parseRewritePattern(absl::string_view pattern) {
```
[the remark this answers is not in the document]
silverstar194: Yep! We do

source/extensions/path/uri_template_lib/uri_template.cc
```diff
@@ -0,0 +1,165 @@
...
+
+  // The pattern should start with a '/' and thus the first segment should
+  // always be a literal.
+  if (pattern.empty() || pattern[0] != '/') {
+    return absl::InvalidArgumentError("Invalid rewrite variable placement");
+  }
+
+  // Don't allow contiguous '/' patterns.
+  static const LazyRE2 invalid_regex = {"^.*//.*$"};
+  if (RE2::FullMatch(Internal::toStringPiece(pattern), *invalid_regex)) {
```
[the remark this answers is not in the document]
silverstar194: Likely but I am going to leave as it's been very battle tested as-is

source/extensions/path/uri_template_lib/uri_template.h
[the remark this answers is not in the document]
silverstar194: I want to return status to bubble up the error message.

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+
+// Returns if path_template_rewrite and capture_regex have valid variables.
+// Every variable in rewrite MUST be present in match.
```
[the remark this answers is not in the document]
silverstar194: Changed the name. Want to keep it more generic than passing the capture map.

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,66 @@
...
+// For example:
+// Match: /foo/{bar}/{var} and Rewrite: /goo/{var} is valid
+// Match: /foo/{bar} and Rewrite: /goo/{bar}/{var} is invalid. Match is missing {var}.
+absl::Status isValidSharedVariableSet(absl::string_view path_template_rewrite,
+                                      absl::string_view capture_regex);
+
+// Returns is the match_pattern is valid.
+// Validation attempts to parse pattern into literals and variables.
+absl::Status isValidMatchPattern(absl::string_view match_pattern);
```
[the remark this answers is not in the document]
silverstar194: Added

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+}
+
+bool isValidRewriteLiteral(absl::string_view literal) {
+  static const std::string* kValidLiteralRegex =
+      new std::string(absl::StrCat("^[", kLiteral, "/]+$"));
+  static const LazyRE2 literal_regex = {kValidLiteralRegex->data()};
+  return RE2::FullMatch(toStringPiece(literal), *literal_regex);
+}
+
+bool isValidVariableName(absl::string_view indent) {
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  static const LazyRE2 ident_regex = {"^[a-zA-Z][a-zA-Z0-9_]*$"};
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  return RE2::FullMatch(toStringPiece(indent), *ident_regex);
+}
+
+absl::StatusOr<ParsedResult<Literal>> consumeLiteral(absl::string_view pattern) {
+  absl::string_view lit =
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.cc
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.cc
```diff
@@ -0,0 +1,386 @@
...
+  if (!isValidVariableName(var_parts[0])) {
+    return absl::InvalidArgumentError("Invalid variable name");
+  }
+  Variable var = Variable(var_parts[0], {});
+
+  // Parse the variable match pattern (if any).
+  if (var_parts.size() < 2) {
+    return ParsedResult<Variable>(var, unconsumed_pattern);
+  }
+  absl::string_view var_patt = var_parts[1];
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Determines what operations to use on the input pattern segment
+ */
+enum class Operator { PathGlob, TextGlob };
+
+/**
+ * Represents a pattern variable. Variables are included in both path match and rewrite paths.
+ */
+struct Variable {
+  absl::string_view var_name_;
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+ */
+std::string toRegexPattern(const Variable& pattern);
+
+/**
+ * Converts ParsedUrlPattern to std::string
+ */
+std::string toRegexPattern(const struct ParsedUrlPattern& pattern);
+
+/**
+ * Converts string_view to be used in re2
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+absl::StatusOr<ParsedResult<Variable>> consumeVariable(absl::string_view pattern);
+
+/**
+ * Converts input pattern to ParsedUrlPattern.
+ * ParsedUrlPattern hold prefix, string literals, and variables for rewrite.
+ */
+absl::StatusOr<ParsedUrlPattern> parseURLPatternSyntax(absl::string_view url_pattern);
+
+/**
+ * Converts Literal to std::string
```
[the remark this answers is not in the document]
silverstar194: changed comment to be better

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Check if literal is valid
+ * Literals cannot hold wildcards or curl brackets.
+ */
+bool isValidLiteral(absl::string_view literal);
+
+/**
+ * Check if rewrite literal is valid
+ * Literals cannot hold wildcards or curl brackets.
```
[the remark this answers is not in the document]
silverstar194: added clarification that each is for a different pattern type

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+struct ParsedUrlPattern {
+  std::vector<ParsedSegment> parsed_segments_;
+  absl::optional<absl::string_view> suffix_;
+  absl::flat_hash_set<absl::string_view> captured_variables_;
+
+  std::string debugString() const;
+};
+
+/**
+ * Check if literal is valid
```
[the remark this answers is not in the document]
silverstar194: Added

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+
+  std::string debugString() const;
+};
+
+using ParsedSegment = absl::variant<Operator, Variable, Literal>;
+
+/**
+ * Represents the parsed path including literals and variables.
+ */
+struct ParsedUrlPattern {
```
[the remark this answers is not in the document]
silverstar194: changed to be consistent

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+  std::vector<ParsedSegment> parsed_segments_;
+  absl::optional<absl::string_view> suffix_;
```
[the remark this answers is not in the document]
silverstar194: The suffix is used in the case of a wildcard at the end of a match. Will update comment..Suffuix
For example:
Pattern: /foo/bar/**
Path: /foo/bar/some/more/stuff

Suffix: some/more/stuff

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Converts input pattern to ParsedResult<Variable>
+ */
+absl::StatusOr<ParsedResult<Variable>> consumeVariable(absl::string_view pattern);
+
+/**
+ * Converts input pattern to ParsedUrlPattern.
+ * ParsedUrlPattern hold prefix, string literals, and variables for rewrite.
+ */
+absl::StatusOr<ParsedUrlPattern> parseURLPatternSyntax(absl::string_view url_pattern);
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.h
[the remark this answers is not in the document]
silverstar194: updated

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+ * Determines what operations to use on the input pattern segment
+ */
+enum class Operator { PathGlob, TextGlob };
+
+/**
+ * Represents a pattern variable. Variables are included in both path match and rewrite paths.
+ */
+struct Variable {
+  absl::string_view var_name_;
+  std::vector<absl::variant<Operator, Literal>> var_match_;
```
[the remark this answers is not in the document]
silverstar194: Added comment

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,138 @@
...
+      : parsed_value_(val), unconsumed_pattern_(pattern) {}
+
+  T parsed_value_;
+  absl::string_view unconsumed_pattern_;
+};
+
+/**
+ * Converts input pattern to ParsedResult<Literal>
+ */
+absl::StatusOr<ParsedResult<Literal>> consumeLiteral(absl::string_view pattern);
```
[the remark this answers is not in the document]
silverstar194: Updated names and added comment

source/extensions/path/uri_template_lib/uri_template.h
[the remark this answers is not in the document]
RyanTheOptimist: Is the regex used in that function beyond the call to get the capture groups? If so, I don't see it. The only change I'm proposing is the data type of this function parameter, and calling NamedCapturingGroups() in the caller instead of the method. Is this a large rewrite?

RyanTheOptimist (commented):
I think we're getting close. I do think it's important to clarify if this code operates on paths vs URL.

source/extensions/path/uri_template_lib/uri_template_internal.h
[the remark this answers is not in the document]
silverstar194: It's just the path.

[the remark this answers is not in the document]
RyanTheOptimist: Ah, ok. So here and elsewhere, let's be clear with naming name this is a path. So here `ParsedPathPattern` and elsewhere `path_pattern` and `path` instead of `url_pattern` and url`.

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,150 @@
...
+bool isValidLiteral(absl::string_view literal);
+
+/**
+ * Returns true if `literal` is valid for pattern rewrite
+ * (does not contain wildcards or curly brackets).
+ */
+bool isValidRewriteLiteral(absl::string_view literal);
+
+/**
+ * Check if variable name is valid.
```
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.h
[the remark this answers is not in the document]
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template.h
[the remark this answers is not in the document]
silverstar194: Correct, but library may be used elsewhere in future and it's useful in test/dev to see the different issues with path formats through errors

RyanTheOptimist:
You'll need to do a main merge to get CI passing. In particular, you'll need to pick up: https://github.com/envoyproxy/envoy/pull/22815

source/extensions/path/uri_template_lib/uri_template_internal.h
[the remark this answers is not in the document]
RyanTheOptimist: What's the semantic difference between a missing suffix (the optional does not have a value) and a present suffix that is empty (optional has a value and it's the empty string)? Is there test coverage which covers this difference?

source/extensions/path/uri_template_lib/uri_template.h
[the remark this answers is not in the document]
RyanTheOptimist: Can you explain what you mean by large or functional chance? It's literally calling `NamedCapturingGroups()` in one place instead of another. 

In any case, since we've dramatically chanced this PR since it was started I don't think that drifting from the internal implementation is problematic.

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -47,91 +49,101 @@ using ParsedSegment = absl::variant<Operator, Variable, Literal>;
...
+ * Returns true if `literal` is valid for pattern match.
+ * (does not contain wildcards or curly brackets).
  */
 bool isValidLiteral(absl::string_view literal);
 
 /**
- * Check if rewrite literal is valid
- * Literals cannot hold wildcards or curl brackets.
+ * Returns true if `literal` is valid for pattern rewrite.
+ * (does not contain wildcards or curly brackets).
```
RyanTheOptimist, line 74: This comment appears identical to the previous, but presumably these two method are not the same so the comment should make clear what the difference is.

source/extensions/path/uri_template_lib/uri_template.cc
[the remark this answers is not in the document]
RyanTheOptimist: Ping

source/extensions/path/uri_template_lib/uri_template_internal.h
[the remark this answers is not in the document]
silverstar194: Ok. I tested it with eliminating `optional`. It doesn't cause any issues. I've changed it to just string_view

source/extensions/path/uri_template_lib/uri_template_internal.h
silverstar194: One comment says `pattern match` the other `pattern rewrite`

silverstar194:
> I think we're getting close. I do think it's important to clarify if this code operates on paths vs URL.

I've removed all the URLs references in the library as its really just paths. The feature is for urls so I left the file names and namespaces as-is.

source/extensions/path/uri_template_lib/uri_template.h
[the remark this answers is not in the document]
silverstar194: I don't really understand the proposed change. Could you explain more?

source/extensions/path/uri_template_lib/uri_template.cc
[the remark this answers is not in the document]
silverstar194: Looking through.. I believe the reason this was done for to allow easier testing and usage by letting you pass in a string_view instead of the ParsedUrlPattern.

source/extensions/path/uri_template_lib/uri_template_internal.h
RyanTheOptimist: Sorry, I meant the "(does not contain any wildcards or curly brackets)" part

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Represents a pattern variable. Variables are included in both path match and rewrite paths.
+ */
+struct Variable {
+  absl::string_view name_;
+
+  // replacement value for the rewrite
+  std::vector<absl::variant<Operator, Literal>> match_;
+
+  Variable(absl::string_view name, std::vector<absl::variant<Operator, Literal>> match)
```
RyanTheOptimist, line 41: nit: Constructors and methods should come before members, per the style guide.

https://google.github.io/styleguide/cppguide.html#Declaration_Order

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,150 @@
...
+enum class Operator { PathGlob, TextGlob };
+
+/**
+ * Represents a pattern variable. Variables are included in both path match and rewrite paths.
+ */
+struct Variable {
+  absl::string_view name_;
+
+  // replacement value for the rewrite
+  std::vector<absl::variant<Operator, Literal>> match_;
```
RyanTheOptimist, line 39: Is this the "replacement value", or the pattern which this variable matched? Since one of the options in this vector is "Operate" I would think that it's not actually the replacement value, but maybe I'm wrong?

(In any case capitalize and period please, per the style guide. https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar)

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,82 @@
...
+  RewriteStringKind kind_;
+};
+
+/**
+ * Returns the safe regex that Envoy understands that is equivalent to the given pattern.
+ */
+absl::StatusOr<std::string> convertPathPatternSyntaxToRegex(absl::string_view path_pattern);
+
+/** Parses the specified pattern into a sequence of segments (which are
+/ * either literals or variable names).
```
RyanTheOptimist, line 38: nit: I think `/**` should go on its own line and each line of the comment should start with ` *` (not `/ *`)/

RyanTheOptimist (commented):
Other than a few nits, this LGTM. Once they're fixed, I'll assign it to a non-Googler for cross-company review.

Thanks for doing all this work to get this code landed!

source/extensions/path/uri_template_lib/uri_template_internal.h
silverstar194: I think it's the actual value from looking at parseVariable. I believe the Operation is for wildcards (`*`) while Literal is for strings between `\` .

source/extensions/path/uri_template_lib/uri_template_internal.h
silverstar194: changed

source/extensions/path/uri_template_lib/uri_template_internal.h
RyanTheOptimist: Right. So as you point out, this `Variable` is produced by `parseVariable()` and stored in a `ParsedSegment`. It represents a parsed rewrite pattern, as I understand it. Looking at the unit tests, we see that `parseVariable()` is called with input like`"{v_1=*/abc/**}"` In this case, the `match_` is going to contain 3 elements, [glob, "abc", glob]. These are the elements of the parsed input pattern. I don't think they're they replacement value for the rewrite are they? The rewritten URL will not have "*" or "**" in it right? In other words, doesn't this represent a matched variable in a "match pattern" not a matched variable in a "rewrite pattern"? I think the latter is represented by `ParsedSegment`?

silverstar194: I see what you're saying. What if we said something like `holds matched segment from path used for rewrite`

RyanTheOptimist: I think it contains the matched *pattern* not the matched *path*. So how about:
```
// The pattern which this variable matches.
```

silverstar194: changed

RyanTheOptimist (dismissed):

RyanTheOptimist:
@zuercher for cross-company review

source/extensions/path/uri_template_lib/proto/rewrite_segments.proto
```diff
@@ -0,0 +1,29 @@
...
+import "udpa/annotations/status.proto";
+
+option java_package = "io.envoyproxy.envoy.extensions.uri_template";
+option java_outer_classname = "UriTemplateRewrite";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/url_template/url_template";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// Holds the segments for rewriting urls base on uri pattern templates
+message RewriteSegments {
```
zuercher, line 14: Does this end up getting included in the public API? If so, perhaps it should be tagged `[#not-implemented-hide:]` to hide it. I was going to suggest that perhaps it should be a C++ type, but I see where you use it to convert strings in test cases into objects for testing.

source/extensions/path/uri_template_lib/uri_template.h
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Returns true if the match_pattern is valid.
+ * Validation attempts to parse pattern into literals and variables.
+ */
+absl::Status isValidMatchPattern(absl::string_view match_pattern);
+
+/**
+ * Concatenates literals and extracts variable values to form the final rewritten path.
+ * For example:
+ * rewrite_pattern: [var_index=2, literal="cat"]
```
zuercher, line 72: nit: capture_index to match the proto

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Return true if variable name is valid.
+ */
+bool isValidVariableName(absl::string_view variable);
+
+/**
+ * Used by the following Parse{Literal.Operator,Variable} functions
+ * in the return value. The functions would take the given pattern,
+ * parse what it can into |parsed_value| and return the unparse
+ *  portion of the pattern in |unparsed_pattern|.
```
zuercher, line 87:
```suggestion
 * parse what it can into |parsed_value| and return the unparsed
 * portion of the pattern in |unparsed_pattern|.
```

source/extensions/path/uri_template_lib/uri_template_internal.h
```diff
@@ -0,0 +1,150 @@
...
+ */
+bool isValidRewriteLiteral(absl::string_view literal);
+
+/**
+ * Return true if variable name is valid.
+ */
+bool isValidVariableName(absl::string_view variable);
+
+/**
+ * Used by the following Parse{Literal.Operator,Variable} functions
```
zuercher, line 84:
```suggestion
 * Used by the following Parse{Literal,Operator,Variable} functions
```

test/extensions/path/uri_template_lib/uri_template_test.cc
```diff
@@ -0,0 +1,369 @@
...
+                                                      - literal: "/bucket-"
+                                                      - capture_index: 3
+                                                      - literal: ".suffix")EOF"},
+                         })));
+
+TEST_P(ParseRewriteSuccess, ParseRewriteSuccessTest) {
+  absl::StatusOr<envoy::extensions::uri_template::RewriteSegments> rewrite =
+      parseRewritePattern(rewritePattern(), kCaptureRegex);
+  ASSERT_OK(rewrite);
+  // EXPECT_THAT(rewrite.value(), testing::EqualsProto(expected_proto()));
```
zuercher, line 154: Should this be enabled or just removed?

zuercher (commented):
A few minor things, but otherwise looks good!

source/extensions/path/uri_template_lib/proto/rewrite_segments.proto
silverstar194: This should just be internal. No visibility from public API

source/extensions/path/uri_template_lib/uri_template.h
silverstar194: changed

test/extensions/path/uri_template_lib/uri_template_test.cc
silverstar194: removed it

source/extensions/path/uri_template_lib/uri_template_internal.h
silverstar194: changed

source/extensions/path/uri_template_lib/proto/rewrite_segments.proto
RyanTheOptimist: Hm. If this is internal only, then could it just be a C++ type instead of a proto? 
(+1 to `[#not-implemented-hide:]` if we need to keep it as a proto)

zuercher: I looked at the docs generated in the PR build and this doesn't seem to appear anywhere, so I think it's ok to leave as is.

zuercher (approved):
Thanks for sticking with this. I think it looks pretty good at this point.

source/extensions/path/uri_template_lib/proto/rewrite_segments.proto
silverstar194: Added todo to convert this to struct

RyanTheOptimist (approved):

KBaichoo:
We should add this to https://github.com/envoyproxy/envoy/blob/main/source/extensions/extensions_build_config.bzl and https://github.com/envoyproxy/envoy/blob/main/source/extensions/extensions_metadata.yaml which makes it easier for folks to exclude this extension if they're not using it.

PR #22537, opened 2022-08-03
uhv: header validators for H1 and H2 codecs

ameily:
Commit Message: uhv: header validators for H1 and H2 codecs
Additional Description: This PR builds off the foundation laid in #21974 to add validators for H1 and H2. H3 and path normalization are intentionally out of scope for this PR and will be completed later.
Risk Level: Low
Testing: Build Envoy with `--define uhv=enabled`, configure a service with UHV enabled, and then test a valid and invalid request.

```
# config.yml, in the HCM section
typed_header_validation_config:
  name: envoy.http.header_validators.envoy_default
  typed_config:
      "@type": type.googleapis.com/envoy.extensions.http.header_validators.envoy_default.v3.HeaderValidatorConfig
      #
      # UHV configuration
      # uncomment any of the following:
      #
      # restrict_http_methods: true
      # reject_headers_with_underscores: true
      # http1_protocol_options: {allow_chunked_length: true}
      #

# run tests
# valid request
http GET http://localhost:10000 X-Valid:ok
# invalid request
http GET http://localhost:10000 "X Invalid":bad
```

Fixes #19753
Docs Changes:
Release Notes:
Platform Specific Features:

Issue #6589, opened 2019-04-15
Path normalization controls


htuch:
The fix for CVE-2019-9900, provided a coarse grained ability to opt-in/out of path normalization. That is, you could either normalize, in which case both normalization was used for matching and transforming the path to the upstream, or not normalize.

Ideally we provide finer grained controls similar to Nginx, where users can opt to normalize for match independent of transforming the path to the upstream.

CC @PiotrSikora 

Action item for CVE-2019-9900

htuch:
Bumping priority, since today we still have normalization disabled by default.

htuch:
Once we have this, we should ensure that normalization occurs by default.

twghu:
Looking at this along with https://github.com/envoyproxy/envoy/issues/6588

twghu:
Work on finding a library for normalisation progessed on https://github.com/envoyproxy/envoy/issues/6588 seems tied to the solution for this.  Will monitor and action pending outcome of same.

htuch:
@twghu related but not dependent, I think we can add in the controls with the existing `PathUtil` interface.

jonnysgomes:
Hi @htuch , I'm using Istio and I'm having an issue related with path traversal. The Envoy version is 1.12. Istio brings the normalize path configuration enabled by default, as we can see [here](https://github.com/istio/istio/pull/14191/files). Even with that configuration enabled, I still can navigate through the url path, for instance:

`https://my.domain.net/myservice/../otherplace`

According with [this](https://github.com/envoyproxy/envoy/commit/7ed6d2187df94c4cb96f7dccb8643bf764af2ccb) link, Envoy already has fixed that. I've checked out the `normalize_path` flag and it's enabled. By my understanding, the url above should be resolved as `https://my.domain.net/myservice/`. So, it's could be a bad Envoy behavior or am I missing somethig?

PiotrSikora:
@jonnysgomes normalized `https://my.domain.net/myservice/../otherplace` is `https://my.domain.net/otherplace`, not `https://my.domain.net/myservice`.

Without normalization, `/myservice/../otherplace` is matched against `/myservice` prefix, which is potentially a security issue, since you're really accessing `/otherplace`.

With normalization, `/myservice/../otherplace` is matched against `/otherplace` prefix, which is correct.

twghu:
Implementation:

Requirements:
1. All existing behavior will be preserved.  
[...]

PR #22707, opened 2022-08-15
uhv: path normalization

ameily:
Commit Message: uhv: path normalization
Additional Description: Initial implementation of path normalization within UHV. This is based on PR #22537 and should be merged after the other is merged. For now this is up as a draft for initial review of the approach.
Risk Level: Low
Refs #6589


yanavlasov (changes requested):

yanavlasov (changes requested):

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+
+  *write = '\0';
+
+  auto normalized_length = static_cast<size_t>(write - path);
+  absl::string_view normalized_path{path, normalized_length};
+  header_map.setPath(normalized_path);
+
+  if (redirect) {
```
yanavlasov, line 256: The logic for redirect is slightly different. Redirect occurs if normalized path is different from the original path. I think here we skip the normalization changes due to slash or dot segment collapsing.

I think we can just check the length of the original path and normalized path and redirect if it is different.

yanavlasov (changes requested):

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,268 @@
...
+  auto path_ptr = std::make_unique<char[]>(length + 1); // auto free on return
+  char* path = path_ptr.get();
+
+  original_path.copy(path, length);
+
+  // We rely on the string being null terminated so that we can safely look forward 1 character.
+  path[length] = '\0';
+
+  // Start normalizing the path.
+  char* read = path;
```
yanavlasov, line 116: It would be preferable to use iterators instead of pointers, since in debug builds you get bounds validation and in release builds it works just as a pointer. You can allocate std::string for destination instead of raw char array, so it you get destination iterator and absl::string_view has iterator support as well.

yanavlasov (changes requested):

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: I'm slightly confused, can you explain this?

- When should redirect occur?
- How does the [`UNESCAPE_AND_REDIRECT `](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/http/header_validators/envoy_default/v3/header_validator.proto#L52) config option change this behavior?
  - If we only go based comparing the length/value of `path` before and after normalization, then I don't think `UNESCAPE_AND_REDIRECT` is needed.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,268 @@
...
+  ch = octet[1] >= 'A' ? (octet[1] - 'A' + 10) : (octet[1] - '0');
+  ch *= 16;
+  ch += octet[2] >= 'A' ? (octet[2] - 'A' + 10) : (octet[2] - '0');
+
+  if (testChar(kUnreservedCharTable, ch)) {
+    // Based on RFC, only decode characters in the UNRESERVED set.
+    return {PercentDecodeResult::Decoded, ch};
+  }
+
+  if (ch == '/' || ch == '\\') {
```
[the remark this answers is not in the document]
ameily: The documentation in `header_validator.proto` mentions decoding `%5c`. Can you confirm that we don't want to handle back slash characters before I update the proto documentation?

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,268 @@
...
+  // SPELLCHECKER(on)
+  //
+  while (read < end) {
+    char ch = *read;
+    char prev = *(write - 1);
+
+    switch (ch) {
+    case '%': {
+      // Potential percent-encoded octet
+      auto decode_result = normalizeAndDecodeOctet(read);
```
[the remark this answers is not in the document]
ameily: I fixed this using iterators and performing bounds checking.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,268 @@
...
+        // encoding (read[2]) so it will be processed in the next iteration.
+        read += 2;
+        *read = decode_result.octet();
+      }
+      break;
+    }
+
+    case '.': {
+      // Potential "/./" or "/../" sequence
+      if (*(read + 1) == '/' || (read + 1) == end) {
```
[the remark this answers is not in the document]
ameily: Done.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: Done. I switched to using iterators and `string_view` objects rather than raw pointers.

wbpcode:
friendly ping @yanavlasov

KBaichoo:
friendly ping @yanavlasov

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,348 @@
...
+  const auto original_uri = header_map.path();
+  if (original_uri == "*") {
+    // Asterisk form
+    return PathNormalizationResult::success();
+  }
+
+  const bool is_connect_method =
+      header_map.method() == ::Envoy::Http::Headers::get().MethodValues.Connect;
+  ::Envoy::Http::Utility::Url url;
+  const bool is_origin_form = !url.initialize(original_uri, is_connect_method);
```
yanavlasov, line 170: This will cause the URL to be parsed twice in H/1 case. Once here and then another one in the validateRequestHeaderMap when request target is in the absolute form. I think we should take care of the absolute form  request target before invoking normalizePathUri to avoid parsing twice.

yanavlasov (changes requested):
/wait

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,348 @@
...
+        *read = decode_result.octet();
+      }
+      break;
+    }
+
+    case '.': {
+      if (prev == '/') {
+        // attempt to read ahead 2 characters to see if we are in a "./" or "../" segment.
+        const auto dot_segment = path_view.substr(std::distance(begin, read), 3);
+        if (absl::StartsWith(dot_segment, "./") || dot_segment == ".") {
```
yanavlasov, line 262: The problem here is that it would miss sequences like `/..%2f` (when decoding of %2f is enabled) or /%2E/ (. is in the unreserved set, so %2E should be decoded).

My overall comment here, is that we should trade some efficiency for clarity and split this big loop that does everything into 3 smaller consecutive loops:

1. decode UNRESERVED and %2F if configured and normalize case for other %2F
2. merge adjacent slashes
3. collapse dot and dot segements.

In this way I think it is easier to reason about various edge cases.

yanavlasov (changes requested):

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: We still need the ability to pull out the `path` component of the `:path` header. So I added a couple methods that attempt to split `:path` into scheme/authority, path, and query/fragment. Then, after normalization has finished, only the path component is updated in the header, preserving the scheme/authority and query/fragment.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: Normalization is now 3 passes: decode, merge slashes (if configured), and collapse dot segments. The big change here is that the final path is not validated, which is fine because that is performed further on in the UHV process.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,446 @@
...
+
+  return std::make_tuple(pathAndQueryParams.substr(0, delim), pathAndQueryParams.substr(delim));
+}
+
+std::tuple<absl::string_view, absl::string_view>
+PathNormalizer::splitAuthorityAndPath(absl::string_view uri) const {
+  // We have to make an efficient best guess for the URL form here. The URL and path will be
+  // validated further on in the UHV stack, so we don't need to be precise, just enough to split
+  // the path component and everything before it.
+  if (absl::StartsWith(uri, "//")) {
```
yanavlasov, line 416: have dug deeper into this.
Envoy [expects](https://github.com/envoyproxy/envoy/blob/9ae348cc43e0341ec7eadd6a4addce587afe1b27/source/common/http/conn_manager_impl.cc#L1004) the `:path` header to always be in the `origin-form` (unless it is CONNECT, in which case the path is required to be empty). Also Envoy currently rejects asterisk-form for OPTIONS requests.

The comment in the code I have linked indicates that H/1 codec is expected to parse the absolute-form target and split into the `:authority` and `:path` which happens [here](https://github.com/envoyproxy/envoy/blob/9ae348cc43e0341ec7eadd6a4addce587afe1b27/source/common/http/http1/codec_impl.cc#L1067).

Given this, UHV should expect the `:path` to be in origin form, empty for CONNECT or * for OPTIONS (this is a breaking change actually, but we'll take care of this later). And reject all other forms. In this case path normalization only happens on the origin-form and there is no need to split authority.

yanavlasov (changes requested):
/wait

source/extensions/http/header_validators/envoy_default/http1_header_validator.cc
```diff
@@ -287,19 +287,48 @@ Http1HeaderValidator::validateRequestHeaderMap(RequestHeaderMap& header_map) {
...
+    // When a proxy receives a request with an absolute-form of request-target, the
+    // proxy MUST ignore the received Host header field (if any) and instead replace
+    // it with the host information of the request-target. A proxy that forwards
+    // such a request MUST generate a new Host field-value based on the received
+    // request-target rather than forward the received Host field-value.
+    // ...
+    // If the target URI includes an authority component, then a client MUST send a
+    // field-value for Host that is identical to that authority component,
+    // excluding any userinfo subcomponent and its "@" delimiter (Section 2.7.1).
+    ::Envoy::Http::Utility::Url url;
```
yanavlasov, line 318: In the current architecture this part is in the H/1 [codec](https://github.com/envoyproxy/envoy/blob/9ae348cc43e0341ec7eadd6a4addce587afe1b27/source/common/http/http1/codec_impl.cc#L1067).

I think it makes sense to move it here eventually, but we should do it in a follow up PR, to make it easier to review. So for now we should just codify that UHV expects the path to be in the origin-form unless it is a CONNECT with empty path or OPTIONS with asterisk-form.

In this case we can just check that the pat start with '/' before calling `normalizePathUri` and remove this code, since H/1 code takes care of overwriting the Host header for now.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: I believe I've addressed this. The path normalizer now validates that the path is in origin-form, asterisk-form (`OPTIONS` method), or empty (`CONNECT` method).

source/extensions/http/header_validators/envoy_default/http1_header_validator.cc
ameily: I've removed this.

ameily: I've removed this.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  path.resize(std::distance(begin, write));
+  return PathNormalizationResult::success();
+}
+
+std::tuple<absl::string_view, absl::string_view>
+PathNormalizer::splitPathAndQueryParams(absl::string_view path_and_query_params) const {
+  // Split on the query (?) or fragment (#) delimiter, whichever one is first.
+  auto delim = path_and_query_params.find_first_of("?#");
```
yanavlasov, line 386: Envoy's [current behavior](https://github.com/envoyproxy/envoy/blob/ce8129cdf640f66b57559ad888f4e6e69203686d/source/common/http/conn_manager_utility.cc#L540) is to reject URI paths with fragment in them. It can be changed using runtime flag.

This really should be a config option to accept, reject or strip fragment from URI path. A significant portions of requests in GCP have # in URI path.

Please add TODO to make behavior with respect to the # in URI configurable.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,400 @@
...
+
+  if (seen_segment_char) {
+    ++begin;
+  }
+
+  return begin;
+}
+
+PathNormalizer::PathNormalizationResult
+PathNormalizer::normalizePathUri(RequestHeaderMap& header_map) const {
```
yanavlasov, line 146: This introduces breaking change, although it is unclear to me if the impact is meaningful. In this implementation, slash merging and decoding of %2F is only done if path_normalization is enabled. In existing implementation these options are independent of each other. I.e. it is possible to only decode %2F or just merge slashes, without making other transformations. 

I think it is unlikely that someone would enable merging slashes without enabling path normalization. 

Please add a comment that the existing behavior has changed and we can evaluate if we need to do anything about it later.

yanavlasov (changes requested):
/wait

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: Done. I added a TODO referencing a new issue: #23886

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: Done. I added a TODO referencing a new issue: #23887

yanavlasov (dismissed):

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,403 @@
...
+ * Find the start index of the previous segment within the path. The previous segment starts at the
+ * first non-slash character after the preceding slash. For example:
+ *
+ *   path = "/hello/world/..";
+ *                  ^    ^-- iterator
+ *                  |-- start of previous segment
+ *
+ * The ``begin`` iterator is returned on error.
+ */
+std::string::iterator findStartOfPreviousSegment(std::string::iterator iter,
```
adisuissa, line 124: Does this function intentionally handle `/abc///def/..` to return the iterator to `a`?
If so I think the comment should clarify this use case.

It is unclear where begin should point to when this function is called.

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,403 @@
...
+  // segment       = *pchar
+  // segment-nz    = 1*pchar
+  // segment-nz-nc = 1*( unreserved / pct-encoded / sub-delims / "@" )
+  //               ; non-zero-length segment without any colon ":"
+  //
+  // pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
+  // SPELLCHECKER(on)
+  {
+    // pass 1: normalize and decode percent-encoded octets
+    auto result = decodePass(path);
```
adisuissa, line 217: nit: const

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,403 @@
...
+PathNormalizer::PathNormalizationResult PathNormalizer::decodePass(std::string& path) const {
+  auto begin = path.begin();
+  auto read = std::next(begin);
+  auto write = std::next(begin);
+  auto end = path.end();
+  bool redirect = false;
+
+  while (read != end) {
+    if (*read == '%') {
+      auto decode_result = normalizeAndDecodeOctet(read, end);
```
adisuissa, line 267: nit: const

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
```diff
@@ -0,0 +1,403 @@
...
+  auto end = path.end();
+
+  while (read != end) {
+    if (*read == '/') {
+      char prev = *std::prev(write);
+      if (prev == '/') {
+        // Duplicate slash, merge it
+        ++read;
+      } else {
+        // Not a duplicate slash or we aren't configured to merge slashes, copy it
```
adisuissa, line 323: I don't see the code that handles the "... we aren't configured to merge slashes". I guess the comment should be updated.

adisuissa (commented):
Thanks for working on this!
Now that #22537 is merged, can you please update the PR description?

Overall I think the code looks good.
I would suggest adding a fuzzer to this (maybe as a future PR).

source/extensions/http/header_validators/envoy_default/path_normalizer.cc
ameily: I've updated the comment to hopefully clarify the function's usage. In your example, the return result would be:

```cpp
path = "/abc///def/..";
//      ^      ^  ^-- current argument
//      |      |-- return value
//      |-- begin
```

yanavlasov (approved):

Issue #22609, opened 2022-08-08
printing non-utf8 string from metadata crashes Envoy (when trace enabled)


stevenzzzz:
*Title*: printing non-utf8 string from metadata crashes Envoy (when trace enabled)

*Description*:
envoy/http/metadata_interface.h seems assume that metadata K:V are in utf8 strings： 

```
class MetadataMap : public UnorderedStringMap {
public:
  using UnorderedStringMap::UnorderedStringMap;

  friend std::ostream& operator<<(std::ostream& out, const MetadataMap& metadata_map) {
    out << "metadata map:";
    for (const auto& metadata : metadata_map) {
      out << "\nkey: " << metadata.first << ", value: " << metadata.second << std::endl;
    }
    return out;
  }
};
```

But seems no such constraint in spec for metadata. 

We should at least eacape the K:V before passing them into spdglog which expects utf8 string. 
https://github.com/abseil/abseil-cpp/blob/master/absl/strings/escaping.h

*Repro steps*:
1. Enable trace logging in Envoy, 
2. put some protobuf::serializeAsString() as stream metadata value.

Envoy will crash when the serializeAsString() contains invalid utf8 string, while spdlog tries to interpret the value as an utf8 string.

stevenzzzz:
assign to @yanavlasov for triaging. 
/assign @yanavlasov

stevenzzzz:
cc @adisuissa

stevenzzzz:
the fix is in.

PR #22709, opened 2022-08-15
Escape before print MetadataMap which could contain invalid utf8 strings.

stevenzzzz:
Commit Message: Escape before print MetadataMap which could contain invalid utf8 strings.
Additional Description:
Risk Level: LOW 
Testing: unit test
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #22609
[Optional Deprecated:]


test/common/http/metadata_interface_test.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+using ::testing::HasSubstr;
+
+namespace Envoy {
+namespace Http {
+namespace {
+
+TEST(MetadataMapTest, KeyValueEscaped) {
```
adisuissa, line 19: IIUC this test only validates the usage of "operator<<".
Can you add a test to validate that the formatter works properly? (maybe something that uses [`LogLevelSetter`](https://github.com/envoyproxy/envoy/blob/972b33265f26ed206c047fd05d9427f4cc6408aa/test/test_common/logging.h#L31) and [`EXPECT_LOG_CONTAINS`](https://github.com/envoyproxy/envoy/blob/main/test/test_common/logging.h#L163) will do the trick).

adisuissa (changes requested):
Thanks for fixing this!

I think that escaping all strings should be better in the long-run. Adding the following formatter (to `source/common/common/fmt.h`) should work:
```
template <> struct formatter<std::string> : formatter<string_view> {
  auto format(const std::string& str_in, fmt::format_context& ctx) -> decltype(ctx.out()) {
    string_view fmt_string_view(absl::CEscape(str_in));
    return formatter<string_view>::format(fmt_string_view, ctx);
  }
};
```

/wait

stevenzzzz:
> auto format(const std::string& str_in, fmt::format_context& ctx) -> decltype(ctx.out()) {

just this change it breaks the unit test I created. 

I am not sure we should escaping all the strings, but maybe we should as this is used in just logging.
@alyssawilk

test/common/http/metadata_interface_test.cc
stevenzzzz: it also validates that the ENVOY_LOG_MISC is not crashing, which, without this PR would crash.

adisuissa:
cc @yanavlasov

dyfrgi:
Rather than create a new specialized formatter, we could use fmt's built-in ostream formatter to re-use the `operator<<()` we already have for `MetadataMap`. e.g.
```
template <> struct fmt::formatter<MetadataMap> : fmt::ostream_formatter {};
```
Idea from https://fmt.dev/latest/api.html#std-ostream-support. It's defined in fmt/ostream.h, so we'd need to include that in metadata_interface.h.

adisuissa:
> Rather than create a new specialized formatter, we could use fmt's built-in ostream formatter to re-use the `operator<<()` we already have for `MetadataMap`. e.g.
> 
> ```
> template <> struct fmt::formatter<MetadataMap> : fmt::ostream_formatter {};
> ```
> 
> Idea from https://fmt.dev/latest/api.html#std-ostream-support. It's defined in fmt/ostream.h, so we'd need to include that in metadata_interface.h.

Yes. The API you mentioned is from FMT-LIB v9, and we use the same default formatter (that uses ostream-formatter which is essentially the default_formatter) that calls operator<<. The include line [here](https://github.com/envoyproxy/envoy/blob/main/envoy/registry/registry.h#L22) creates a new formatter for the MetadataMap and overrides the ostream one.

stevenzzzz:
I dont think we can tell when the  fmt/ranges.h header will be included (either by Envoy or any lib Envoy depends on), which result in the map-specialization been expanded/used. In the write_range_entry call, the invalid utf8 string leads to ASAn error.  I think we can prevent it from happening before that's fixed in fmtlib.

envoy/http/metadata_interface.h
```diff
@@ -2,10 +2,13 @@
 
 #include <functional>
 #include <iostream>
+#include <sstream>
 #include <memory>
 #include <vector>
 
 #include "absl/container/node_hash_map.h"
+#include "absl/strings/escaping.h"
+#include "fmt/ranges.h"
```
adisuissa, line 11: nit: should this be changed to "fmt/format.h" or "fmt/stream.h"?

test/common/http/metadata_interface_test.cc
```diff
@@ -0,0 +1,53 @@
+#include <array>
+#include <cstdint>
+#include <string>
+
+#include "envoy/http/metadata_interface.h"
+
+#include "source/common/common/fmt.h"
+#include "source/common/common/logger.h"
+#include "source/common/http/utility.h"
```
adisuissa, line 9: This should be: "fmt/ranges.h", and please add a comment why it is added.

adisuissa (commented):
Thanks!
Left a couple small nits.

@mattklein123 PTAL for a non-googler review

envoy/http/metadata_interface.h
stevenzzzz: change to format.h.

test/common/http/metadata_interface_test.cc
stevenzzzz: done

adisuissa (approved):
LGTM, thanks!

mattklein123 (approved):

PR #22651, opened 2022-08-11
contrib: add ISA-L compressor

giantcroc:
Signed-off-by: giantcroc [changran.wang@intel.com](mailto:changran.wang@intel.com)

Commit Message:contrib: add ISA-L compressor
Additional Description:
[ISA-L](https://github.com/intel/isa-l) is a collection of optimized low-level functions targeting storage applications. The low-level functions of ISA-L are written in manual assembly code that many details are tuned to achieve optimal performance.

ISA-L uses efficient SIMD (Single Instruction, Multiple Data) instructions like AVX512 to maximize the use of CPU micro-architecture to accelerate the calculation process of compression algorithms. And ISA-L mainly optimized the implementation of CRC, DEFLATE and Huffman encoding for gzip compression algorithm.

We believe that ISA-L can help Envoy to improve gzip compression performance in case of **high network load**, details for [benchmark](https://github.com/envoyproxy/envoy/issues/22397#issue-1317778981).

ByteDance will be our sponsor for this extension, See [comment](https://github.com/envoyproxy/envoy/issues/22397#issuecomment-1207816495).

Risk Level: Low
Testing: Unit
Docs Changes: API
Release Notes: N/A
Platform Specific Features: modern x86 processor
Fix: #22397

PR #22717, opened 2022-08-16
build: update to qatlib 22.07

### LINKED ISSUE #22341: Newer release available `com_github_intel_qatlib`: 22.07.0 (current: 21.11.0)
Package Name: com_github_intel_qatlib@21.11.0
Current Version: 21.11.0@2021-11-10
Available Version: 22.07.0@2022-07-20
Upstream releases: https://github.com/intel/qatlib/releases

ipuustin:
Commit Message:

Update qatlib to 22.07.

Additional Description:

One new additional dependency is added: yasm assembler is needed for compiling assembler code for qatlib. Adding the yasm tool as build_data (it's not part of the resulting Envoy binary). An alternative would be to add yasm tool to the build container.

Risk Level: low
Testing: in-target testing, unit tests
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: qatlib requires a processor with Intel 4xxx series QAT device.
Fixes #22341


phlax:
> qatlib requires a processor with Intel 4xxx series QAT device.

one thing i noticed is that the `qat` tests seem to be running on arm - im wondering if this requires some bazel selects to prevent that

wrowe:
Is the requirement here to use libyasm, or is it to use the compiler? yasm seems to have paused in 2019, while nasm remains active and issuing incremental releases on 1.15.x. NASM is already on the windows build toolchain for openssl etc, and can be trivially added to the linux toolchain. Is there a reason for the preference?

ipuustin:
@phlax I can work on disabling ARM testing. Right now the tests use a mock library so that they can be run without QAT hardware (they just test the private key provider logic). So in a sense it doesn't matter too much on which platform the tests are run, but indeed the extension will not work on ARM, so testing on ARM is redundant.

@wrowe The requirement is to the compiler -- the library is not linked in. NASM indeed seems a lot more active project. I also see that https://github.com/envoyproxy/envoy/pull/22651 wants to add NASM so that it would seem a more prudent choice. I'll see if I can make libqat use NASM.

wrowe:
Thanks for looking into leveraging nasm!

/wait

wrowe:
Note, this is going to need to be addressed first before the update can be considered;

https://github.com/envoyproxy/envoy/pull/22651#discussion_r948011839

/wait

ipuustin:
I now have a NASM conversion patch. However, I talked with qatlib developers, and there is now a plan to drop the mandatory assembler dependency from the next qatlib release. I'll see if we can patch the 22.07 qatlib code already with that change, so that we could drop the new dependency from this PR.

ipuustin:
We're working now on removing the assembler dependency from qatlib, but this will still take a while. @VillePihlava

wrowe:
@ipuustin As this is still in motion, is it safe to defer this until either 22.08 ships or at least until after the envoy release around the 15th?

ipuustin:
@wrowe Yes, it's safe to defer.

ipuustin:
Replaced by https://github.com/envoyproxy/envoy/pull/24002.

PR #22722, opened 2022-08-16
allow additional network filters for QUIC listeners

pksohn:
Signed-off-by: Paul Sohn <paulsohn@google.com>

Commit Message: For QUIC listeners, allow network filters to be added before HTTP Connection Manager
Additional Description: Note that onData() is still never called on QUIC.
Risk Level: Low
Testing: Unit test
Docs Changes: N/A
Release Notes: N/A


danzh2010 (commented):
LGTM overall!
Can you document here https://github.com/envoyproxy/envoy/blob/02489bbf8990faddb6be1f2e0f22851b64c3fc85/api/envoy/config/listener/v3/listener_components.proto#L230 that the filters in UDP connections will be created but not process payload as TCP connections do?

danzh2010:
Plz also fix the clang-tidy CI

api/envoy/config/listener/v3/listener_components.proto
```diff
@@ -227,6 +227,10 @@ message FilterChain {
   // connections established with the listener. Order matters as the filters are
   // processed sequentially as connection events happen. Note: If the filter
   // list is empty, the connection will close by default.
+  //
+  // For UDP connections, network filters can be created but do not process
+  // incoming data as network filters on TCP connections do (i.e. the onData method
+  // will never be called).
```
mattklein123, line 233: What's the point of this? I assume just for things like RBAC where decisions can be made at the point of initial connection? If so can you clarify? Also, do you want to phrase it more like network filters do not currently process incoming network data? As we might fix this in the future?

Can you also add a release note about this?

test/server/listener_manager_impl_quic_only_test.cc
```diff
@@ -332,8 +332,72 @@ TEST_P(ListenerManagerImplQuicOnlyTest, QuicListenerFactoryWithWrongCodec) {
...
+      route_config:
+        name: local_route
+      http_filters:
+        - name: envoy.filters.http.router
+          typed_config:
+            "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+  - name: envoy.filters.network.rbac
+    typed_config:
+      "@type": type.googleapis.com/envoy.extensions.filters.network.rbac.v3.RBAC
+      stat_prefix: rbac
```
mattklein123, line 369: Shouldn't RBAC come before HCM?

Can you actually add a QUIC integration test that shows this working end to end? You could use a test filter which terminates the connection always for example. Thank you.

mattklein123 (changes requested):
Thanks generally makes sense with a few comments.

/wait

api/envoy/config/listener/v3/listener_components.proto
danzh2010: Probably we shouldn't generalize it to all UDP-based connections, but be explicit about QUIC. Today QUIC doesn't pass the connection payload to network filter interface `onData()` because the connection itself handles multiplexing and HTTP parsing. QUIC may support non-http protocols in the future, but multiplexing always happens at L4. So QUIC L4 filters are very likely to have a different set of interfaces. So far we don't really need any L4 filter interfaces, that's why we have only allowed HCM to be installed. But we have a use case now where L4 filters can change the connection transport properties, like cwnd or socket options at the beginning of the connection. This still won't use any L4 interfaces other than the constructor. So we want to relax the constraint here.

mattklein123: OK makes sense. Yes let's please be much more clear about what is supported and also add an integration test. Thank you.

test/server/listener_manager_impl_quic_only_test.cc
danzh2010: This is testing the failure case.

+1 for an integration test to verify a quic connection can actually work with multiple L4 filters. The filter constructor can do something like incrementing a stats counter.

mattklein123: Ah OK got it, makes sense. SGTM on the test.

api/envoy/config/listener/v3/listener_components.proto
pksohn: Thanks both. I've added a release note and tried to clarify the wording (let me know if you have further suggestions). I'll add an integration test next.

pksohn: I've added an integration test; please take a look.

api/envoy/config/listener/v3/listener_components.proto
```diff
@@ -227,6 +227,11 @@ message FilterChain {
   // connections established with the listener. Order matters as the filters are
   // processed sequentially as connection events happen. Note: If the filter
   // list is empty, the connection will close by default.
+  //
+  // For QUIC listeners, network filters can be created, but due to differences
+  // in the connection implementation compared to TCP, the onData() method will
+  // never be called. Therefore network filters for QUIC listeners should only
+  // expect to do work at the start of a new connection (i.e. in onNewConnection()).
```
soulxu, line 234: it would be good to mention, the HCM required to be the last filter for udp.

pksohn: Done, thanks

soulxu (approved):
one small comment

api/envoy/config/listener/v3/listener_components.proto
```diff
@@ -227,6 +227,13 @@ message FilterChain {
   // connections established with the listener. Order matters as the filters are
   // processed sequentially as connection events happen. Note: If the filter
   // list is empty, the connection will close by default.
+  //
+  // For QUIC listeners, network filters other than HTTP Connection Manager (HCM)
+  // can be created, but due to differences in the connection implementation compared
+  // to TCP, the onData() method will never be called. Therefore, network filters
+  // for QUIC listeners should only expect to do work at the start of a new connection
+  // (i.e. in onNewConnection()). If there are multiple network filters, HCM must be
+  // the last filter in the chain.
```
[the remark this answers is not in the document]
pksohn: Thanks, added another clarification.

mattklein123 (approved):
Thanks, LGTM!

PR #18246, opened 2021-09-23
proto: add a new field called append_action in the HeaderValueOption

agrawroh:
This PR adds a new field called `append_action` to the `HeaderValueOption`. This is just the `.proto` change being done in the preparation for #18130. We'll mark the existing `append` field as deprecated in #18310.

`append_action` extends the header matcher by describing what append action to take indicating whether:
1. The formatted header should be added only if it's not already present;
2. New value should be appended to the existing values if header already exists; or
3. New value should be overwritten by discarding any existing values if header already exists;

**Commit Message:** add a new field called append_action in the HeaderValueOption.
**Additional Description:** Adds a new field called `append_action` in `HeaderValueOption` which will ultimately deprecate the old `append` and will extend the existing matcher functionality.
**Risk Level:** -
**Testing:** -
**Docs Changes:** N/A (Will be done in #18130)
**Release Notes:** N/A (Will be done in #18130)
**Platform Specific Features:** N/A

**Signed-off-by:** Rohit Agrawal <rohit.agrawal@databricks.com>

PR #19204, opened 2021-12-06
code cleanup:  removals of GCOVR macros

alyssawilk:
This currently has 2 examples of what I hope will be GCOVR removal.
1) where we have non-proto enums where cases return, and a default-panic, remove the default and replace with PANIC_DUE_TO_CORRUPT_ENUM after the switch
2)  where we have proto enums, remove the default and add PANIC_IF_SENTINAL_VALUES_USED

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Part of https://github.com/envoyproxy/envoy/issues/19172

PR #22723, opened 2022-08-16
route: implementation of HeaderAppendAction

### LINKED ISSUE #22713: [Question] How can I insert a HTTP header only when that header does not exist?
For example, I want to add a HTTP header `foo: bar` into request when the original request does not have the HTTP header `foo`. But, if the original request have the HTTP header `foo`, envoy should do nothing header manipulation for this request.

The existing setting [HeaderValueOption](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-headervalueoption) can't resolve this issue. Please correct me if i am wrong.

CC @wbpcode

wbpcode:
Commit Message: route: implementation of HeaderAppendAction
Additional Description:

The HeaderAppendAction was added in the #18246. But the implementation never done. This PR try to complete this feature.

Further work of #18246. To close #22713.

Risk Level: Low.
Testing: Unit Test.
Docs Changes: n/a.
Release Notes: wait.


source/common/router/header_parser.cc
```diff
@@ -214,45 +211,59 @@ HeaderFormatterPtr parseInternal(const envoy::config::core::v3::HeaderValue& hea
...
+      // 'append' is set and ensure the 'append_action' value is equal to the default value.
+      if (header_value_option.append_action() != HeaderAppendOption::APPEND_IF_EXISTS_OR_ADD) {
+        throw EnvoyException("Both append and append_action are set and it's not allowed");
+      }
+
+      append_action = header_value_option.append().value()
+                          ? HeaderAppendOption::APPEND_IF_EXISTS_OR_ADD
+                          : HeaderAppendOption::OVERWRITE_IF_EXISTS_OR_ADD;
+    } else {
+      append_action = header_value_option.append_action();
```
htuch, line 236: if `append` is false, then I don't see how we can accept a default action here which is to essentially append (`APPEND_IF_EXISTS_OR_ADD`). Maybe we should only interpret `append_action` if `append` is true, in which case the semantics seem to make more sense.

wbpcode: `append` is optional bool. We will use `append_action` only when the we don't set `append`.

If `append` is set to false, the `OVERWRITE_IF_EXISTS_OR_ADD` will be used. 

So there should be no any behaviors change.

api/envoy/config/core/v3/base.proto
```diff
@@ -358,8 +358,10 @@ message HeaderValueOption {
   // existing values. Otherwise it replaces any existing values.
   google.protobuf.BoolValue append = 2;
```
soulxu, line 359: We should be able to mark this as deprecated.

source/common/router/header_parser.cc
```diff
@@ -214,45 +211,59 @@ HeaderFormatterPtr parseInternal(const envoy::config::core::v3::HeaderValue& hea
...
   HeaderParserPtr header_parser(new HeaderParser());
 
   for (const auto& header_value_option : headers_to_add) {
-    const bool append = PROTOBUF_GET_WRAPPED_OR_DEFAULT(header_value_option, append, true);
-    HeaderFormatterPtr header_formatter = parseInternal(header_value_option.header(), append);
+    HeaderAppendAction append_action;
+
+    if (header_value_option.has_append()) {
+      // 'append' is set and ensure the 'append_action' value is equal to the default value.
+      if (header_value_option.append_action() != HeaderAppendOption::APPEND_IF_EXISTS_OR_ADD) {
```
soulxu, line 228: This means if the user explicitly set `append` as true and `append_action` as APPEND_IF_EXISTS_OR_ADD is valid config.

But if we consider to deprecated `append` field, then this case should be invalid also. We encourage the user use the new field,  if both fields are specified, it should consider as invalid even if their value not conflict.

soulxu (changes requested):

test/common/router/header_formatter_test.cc
```diff
@@ -1252,14 +1248,14 @@ match: { prefix: "/new_endpoint" }
...
-    append: true
+    append_action: APPEND_IF_EXISTS_OR_ADD
   - header:
       key: "x-client-ip-port"
       value: "%DOWNSTREAM_REMOTE_ADDRESS%"
   - header:
       key: "x-client-port"
       value: "%DOWNSTREAM_REMOTE_PORT%"
-    append: true
+    append_action: APPEND_IF_EXISTS_OR_ADD
```
htuch, line 1258: There is a lot of test churn. Can you point to the tests that still provide coverage of the deprecated functionality? We should have some until the old field is retired.

api/envoy/config/core/v3/base.proto
htuch: +1

source/common/router/header_parser.cc
htuch: Got it, makes sense.

test/common/router/header_formatter_test.cc
wbpcode: Yeah, some more test for deprecated `append` is on the way.

api/envoy/config/core/v3/base.proto
wbpcode: of course. 😸

source/common/router/header_parser.cc
wbpcode: I think there is no way to distinguish a explicit APPEND_IF_EXISTS_OR_ADD or default APPEND_IF_EXISTS_OR_ADD. 🤔

soulxu: emm...I thought there should be an interferce `has_append_action`, seems not.

wbpcode:
It's ready for a formal review for now. cc @htuch @soulxu 😸

source/common/router/header_parser.h
```diff
@@ -38,16 +41,16 @@ class HeaderParser : public Http::HeaderEvaluator {
    */
   static HeaderParserPtr
   configure(const Protobuf::RepeatedPtrField<envoy::config::core::v3::HeaderValue>& headers_to_add,
-            bool append);
```
soulxu, line 41: Forget to remove the comment for `append` at https://github.com/envoyproxy/envoy/pull/22723/files#diff-b5669de72ffdab01ee2e54f54e9cb97f9d6f7bf4e053fb31c6af1ee56f2cf316R39

source/common/router/header_parser.cc
```diff
@@ -214,45 +211,59 @@ HeaderFormatterPtr parseInternal(const envoy::config::core::v3::HeaderValue& hea
...
 
   for (const auto& header_value_option : headers_to_add) {
-    const bool append = PROTOBUF_GET_WRAPPED_OR_DEFAULT(header_value_option, append, true);
-    HeaderFormatterPtr header_formatter = parseInternal(header_value_option.header(), append);
+    HeaderAppendAction append_action;
+
+    if (header_value_option.has_append()) {
+      // 'append' is set and ensure the 'append_action' value is equal to the default value.
+      if (header_value_option.append_action() != HeaderValueOption::APPEND_IF_EXISTS_OR_ADD) {
+        throw EnvoyException("Both append and append_action are set and it's not allowed");
```
soulxu, line 229: seems like lack unittest for this case.

soulxu (changes requested):
LGTM, two small comments

source/common/router/header_parser.h
wbpcode: right. Thanks.

soulxu (approved):
LGTM, thanks!

source/common/router/header_parser.cc
```diff
@@ -285,10 +296,20 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
...
+      case HeaderValueOption::ADD_IF_ABSENT:
+        if (auto header_entry = headers.get(key); header_entry.empty()) {
+          headers.addReferenceKey(key, value);
+        }
+        break;
+      case HeaderValueOption::OVERWRITE_IF_EXISTS_OR_ADD:
         headers.setReferenceKey(key, value);
+        break;
+      default:
+        PANIC_DUE_TO_CORRUPT_ENUM;
```
htuch, line 312: Looking back at https://github.com/envoyproxy/envoy/pull/19204, I think we want `PANIC_ON_PROTO_ENUM_SENTINEL_VALUES`.

test/common/router/header_formatter_test.cc
```diff
@@ -1639,6 +1640,132 @@ request_headers_to_remove: ["x-foo-header"]
...
+  {
+    Http::TestResponseHeaderMapImpl header_map{{"x-per-header", "exist-per"}};
+    resp_header_parser->evaluateHeaders(header_map, stream_info);
+    EXPECT_EQ("foo", header_map.get_("x-foo-header"));
+    EXPECT_EQ("bar", header_map.get_("x-bar-header"));
+    EXPECT_EQ("exist-per", header_map.get_("x-per-header"));
+  }
+}
+
+TEST(HeaderParserTest, DEPRECATED_FEATURE_TEST(EvaluateRequestHeadersAddWithDeprecatedAppend)) {
```
htuch, line 1701: Are these the only tests for the deprecated append?

api/envoy/config/core/v3/base.proto
```diff
@@ -334,7 +334,7 @@ message HeaderValueOption {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.api.v2.core.HeaderValueOption";
 
-  // [#not-implemented-hide:] Describes the supported actions types for header append action.
```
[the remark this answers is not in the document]
wbpcode: What do you mean by the `version history`? This `HeaderAppendAction` has no previous version.

source/common/router/header_parser.cc
wbpcode: get it.

wbpcode:
Because some special behavior in the `ext_authz`/`ext_proc` (#22845), most update in the ext_authz & ext_proc has been reverted.

api/envoy/config/core/v3/base.proto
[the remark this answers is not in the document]
htuch: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.24/v1.24.0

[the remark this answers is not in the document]
wbpcode: I see. The release note. Get it. Thanks.

source/common/router/header_parser.cc
```diff
@@ -55,7 +54,7 @@ HeaderFormatterPtr parseInternal(const envoy::config::core::v3::HeaderValue& hea
 
   absl::string_view format(header_value.value());
   if (format.empty()) {
-    return std::make_unique<PlainHeaderFormatter>("", append);
```
htuch, line 58: For my education, why did we ever have append here?

htuch: I see actually, you now propagate this independent of the header formatter, seems cleaner.

htuch (approved):
LGTM, thanks!

PR #22728, opened 2022-08-16
Fix exclude bug in check_visibility

silverstar194:
Commit Message:
Fix exclude bug in check_visibility

Additional Description:
It's a python concatenation issue. Currently the exclude list looks like ["item", ... "itemN"] it should be without [ ] like"item", ... "itemN"

Risk Level:
Low. Could only break test. No product changes.

Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


RyanTheOptimist:
@phlax can you take a look at this?

phlax (approved):
lgtm, thanks @silverstar194 

relatedly im wondering why it doesnt catch the error (or rather why the `1` returncode is ignored)

```python
 1095             if (e.returncode != 0 and e.returncode != 1):
```

silverstar194:
> lgtm, thanks @silverstar194
> 
> relatedly im wondering why it doesnt catch the error (or rather why the `1` returncode is ignored)
> 
> ```python
>  1095             if (e.returncode != 0 and e.returncode != 1):
> ```

It doesn't catch it because the `git diff` makes this to only run when there are changes to the extension directories. I think I'm the first one to do that after this change went through

Issue #22162, opened 2022-07-13
transaction mode redis


Jacobien1:
why does the redis transaction mode is not supported by envoy?

adisuissa:
cc  @weisisea @mattklein123

mattklein123:
Because in general there is no good way to perform a transaction across multiple sharded redis hosts using a stateless proxy like Envoy.

PR #22729, opened 2022-08-16
redis: added support for transactions (#22162)

asheryerm:
Signed-off-by: asheryer <asheryer@amazon.com>

Commit Message: redis: added support for transactions (#22162)
Additional Description:
This has been a feature requested by several large clients who would like to work with envoy+elasticache.
The lack of support for transactions in envoy has prevented them from migrating to envoy thus for.
For certain users, redis transactions are a vital part of their workflow/s.

I implemented transactions by opening a new connection to the primary host of the shard that handles the transaction.
A new connection is needed to prevent contamination of the transaction by other clients.
Once the transaction is over the connection is closed.

In more detail:
When a MULTI command is sent, we mark down that this client has started a transaction and we reply locally with an "OK". We then wait for a command with key KEY, and once it arrives we:
1. Establish a connection with the primary host of the shard that handles KEY.
2. Send a MULTI command on the new connection and ignore the response.
3. Only simple commands (that hash to one slot) can be used within a transaction.
4. Send all commands of the transaction on this connection.
5. Redirection is disabled.
6. When the transaction is terminated with EXEC or DISCARD we close the connection after sending the last command.

Edge cases such as nested MULTI, EXEC without MULTI, etc. are all handled.

As far as the user is concerned, the user experience is almost identical to working with a standalone Redis node.

Risk Level: Medium
Testing: Tested on standalone/cluster/elasticache
Docs Changes: changelog
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jmarantz:
looks like formatting is still broken.

/wait

asheryerm:
Working now on fixing the tests

bpo:
@asheryerm Glad to see you working on this. Did you look into accumulating the commands between `multi` and `exec` in a buffer before sending, thereby skipping the need to establish a dedicated connection per-transaction? Just curious if so and what the tradeoffs were versus this design.

asheryerm:
@bpo Hi Brian, yes, there are two viable options:
1. Accumulating the commands and sending them all when EXEC is called.
2. Opening a dedicated connection.

I chose option (2) for two reasons:
a. Redis detects certain  errors in the transaction before EXEC is called. Using a dedicated transaction allows you to receive feedback from Redis for each command in the transaction.
2. Assuming we accumulate all the commands and then pipeline them all to Redis, it is unclear if I can get Envoy to this atomically without other clients interleaving commands between the pipelined commands. This may be possible, but I wasn't 100% sure about it.

mattklein123:
@weisisea do you mind taking a first pass on this? Thank you.

weisisea:
Ack, will take a look today!

asheryerm:
@weisisea Do you know why the window release is failing?

asheryerm:
Dependency check is failing after syncing with main branch (but does succeed when I run it locally)

asheryerm:
@mattklein123 Hi Matt, can someone please review this PR? I have not received feedback on it yet.

weisisea:
Apologies, other things got busy last week. Will make sure to look today.

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
```diff
@@ -432,13 +446,98 @@ void SplitKeysSumResultRequest::onChildResponse(Common::Redis::RespValuePtr&& va
...
+    if (transaction.active_ == false) {
+      callbacks.onResponse(Common::Redis::Utility::makeError(
+          fmt::format("{} without MULTI", absl::AsciiStrToUpper(to_lower_string))));
+      return nullptr;
+    }
+
+    // Handle the case where the transaction is empty.
+    if (transaction.key_.empty()) {
+      if (to_lower_string == "exec") {
+        localResponse(callbacks, "(empty array)");
```
weisisea, line 490: I think "(empty array)" is a message printed from redis-cli. The actual RESP message returned after MULTI and EXEC is "*0\r\n":

```
> nc -q 1 localhost 6379
MULTI
+OK
EXEC
*0
```

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
```diff
@@ -432,13 +446,98 @@ void SplitKeysSumResultRequest::onChildResponse(Common::Redis::RespValuePtr&& va
...
+
 InstanceImpl::InstanceImpl(RouterPtr&& router, Stats::Scope& scope, const std::string& stat_prefix,
                            TimeSource& time_source, bool latency_in_micros,
                            Common::Redis::FaultManagerPtr&& fault_manager)
     : router_(std::move(router)), simple_command_handler_(*router_),
       eval_command_handler_(*router_), mget_handler_(*router_), mset_handler_(*router_),
       split_keys_sum_result_handler_(*router_),
-      stats_{ALL_COMMAND_SPLITTER_STATS(POOL_COUNTER_PREFIX(scope, stat_prefix + "splitter."))},
+      transaction_handler_(*router_), stats_{ALL_COMMAND_SPLITTER_STATS(
+                                          POOL_COUNTER_PREFIX(scope, stat_prefix + "splitter."))},
```
weisisea, line 540: nit: might be better to add `transaction_handler_(*router_),` in a separate line

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
```diff
@@ -432,13 +446,98 @@ void SplitKeysSumResultRequest::onChildResponse(Common::Redis::RespValuePtr&& va
...
+      return nullptr;
+    }
+
+    // In all other cases we will close the transaction connection after sending the last command.
+    transaction.should_close_ = true;
+  }
+
+  // When we receive the first command with a key we will set this key as our transaction
+  // key, and then send a MULTI command to the node that handles that key.
+  const auto route = router.upstreamPool(transaction.key_);
```
weisisea, line 504: When transaction.key_ is empty (for the first command), should we first assign transaction.key_ and then calculate the route?

What will happen if the keys of subsequent commands (not the first) hash to different shards? I guess the connected Redis node will return "MOVED" error and we will return the "MOVED" errors back to the client

weisisea:
This looks very cool. Thanks for working on the support for transactions!

Could you please help clarify on the following questions?

1. What will happen if the keys of subsequent commands (not the first) hash to different shards from the first key? I'm guessing that the connected Redis node will return "MOVED" error and we will return the "MOVED" errors back to the client
2. What will happen if the client closes the connection during the transaction (before sending EXEC or DISCARD)? Will the upstream connection be closed at some point?

Thanks!

weisisea:
> @weisisea Do you know why the window release is failing?

After clicking through the presubmit test links, reached a page https://dev.azure.com/cncf/envoy/_build/results?buildId=116184&view=logs&j=b840a642-5ff3-5357-2e4b-e06e40b0cffd&t=67965174-5100-5631-9dc0-68b9f0aacb53&l=1036 showing that:

```
source/common/network/connection_impl.cc(3): fatal error C1083: Cannot open include file: 'execinfo.h': No such file or directory
```

Actually is this file needed for this change?

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
asheryerm: Fixed.

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
asheryerm: You are correct, fixed.

asheryerm:
> Actually is this file needed for this change?

This is a good catch. I added this header previously for debugging and should have removed it. Fixed.

asheryerm:
> Could you please help clarify on the following questions?
> 
> 1. What will happen if the keys of subsequent commands (not the first) hash to different shards from the first key? I'm guessing that the connected Redis node will return "MOVED" error and we will return the "MOVED" errors back to the client
> 2. What will happen if the client closes the connection during the transaction (before sending EXEC or DISCARD)? Will the upstream connection be closed at some point?

1. Yes, once the first key is established you are pinned to the primary node that serves that key. And redirection is disabled. So any commands in the transaction that refer to a key in another slot will elicit an error from Redis.
2. If the downstream connection is closed it should destroy the ProxyFilter instance, which will call the destructor of Transaction, which will close the upstream connection.

asheryerm:
@weisisea Do you have any other comments about the PR?

weisisea:
Thanks for addressing my comments. I'll take another look by EOW.

source/extensions/filters/network/common/redis/client.h
```diff
@@ -203,13 +203,72 @@ class ClientFactory {
...
+    }
+    should_close_ = false;
+  }
+
+  bool active_{false};
+  bool connection_established_{false};
+  bool should_close_{false};
+  std::string key_;
+  ClientPtr client_;
+  Network::ConnectionCallbacks* connection_cb_;
```
weisisea, line 266: Could the variables be made private?

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
```diff
@@ -432,13 +446,102 @@ void SplitKeysSumResultRequest::onChildResponse(Common::Redis::RespValuePtr&& va
   }
 }
 
+SplitRequestPtr TransactionRequest::create(Router& router,
+                                           Common::Redis::RespValuePtr&& incoming_request,
+                                           SplitCallbacks& callbacks, CommandStats& command_stats,
+                                           TimeSource& time_source, bool delay_command_latency) {
+  Common::Redis::Client::Transaction& transaction = callbacks.transaction();
+  std::string to_lower_string = absl::AsciiStrToLower(incoming_request->asArray()[0].asString());
```
weisisea, line 454: nit: might be better to change variable name "to_lower_string" to "command_name". I don't think "to_lower_string" in existing code is straightforward to read.

source/extensions/filters/network/common/redis/client.h
```diff
@@ -203,13 +203,72 @@ class ClientFactory {
...
+    key_.clear();
+    if (connection_established_) {
+      client_->close();
+      connection_established_ = false;
+    }
+    should_close_ = false;
+  }
+
+  bool active_{false};
+  bool connection_established_{false};
```
weisisea, line 262: Could we check whether connection is established by checking whether client_ is null or not? Trying to see whether the variable can be saved.

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
```diff
@@ -432,13 +446,102 @@ void SplitKeysSumResultRequest::onChildResponse(Common::Redis::RespValuePtr&& va
...
+
 InstanceImpl::InstanceImpl(RouterPtr&& router, Stats::Scope& scope, const std::string& stat_prefix,
                            TimeSource& time_source, bool latency_in_micros,
                            Common::Redis::FaultManagerPtr&& fault_manager)
     : router_(std::move(router)), simple_command_handler_(*router_),
       eval_command_handler_(*router_), mget_handler_(*router_), mset_handler_(*router_),
       split_keys_sum_result_handler_(*router_),
-      stats_{ALL_COMMAND_SPLITTER_STATS(POOL_COUNTER_PREFIX(scope, stat_prefix + "splitter."))},
+      transaction_handler_(*router_), stats_{ALL_COMMAND_SPLITTER_STATS(
+                                          POOL_COUNTER_PREFIX(scope, stat_prefix + "splitter."))},
```
weisisea, line 544:
```suggestion
      transaction_handler_(*router_),
      stats_{ALL_COMMAND_SPLITTER_STATS(POOL_COUNTER_PREFIX(scope, stat_prefix + "splitter."))},
```

test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc
```diff
@@ -1205,5 +1205,98 @@ TEST_P(RedisProxyWithFaultInjectionIntegrationTest, DelayFault) {
...
+}
+
+// This tests a full redis transaction.
+
+TEST_P(RedisProxyIntegrationTest, FullTransaction) {
+  initialize();
+  IntegrationTcpClientPtr redis_client = makeTcpConnection(lookupPort("redis_proxy"));
+
+  proxyResponseStep(makeBulkStringArray({"multi"}), "+OK\r\n", redis_client);
+  proxyResponseStep(makeBulkStringArray({"set", "foo", "bar"}), "-upstream failure\r\n",
```
weisisea, line 1295: Why do we expect "upstream failure" in this test?

Should the list of commands be like this:
1. MULTI
2. command1
3. command2
4. EXEC

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
```diff
@@ -432,13 +446,102 @@ void SplitKeysSumResultRequest::onChildResponse(Common::Redis::RespValuePtr&& va
...
+  // When we receive the first command with a key we will set this key as our transaction
+  // key, and then send a MULTI command to the node that handles that key.
+  RouteSharedPtr route;
+  if (transaction.key_.empty()) {
+    transaction.key_ = incoming_request->asArray()[1].asString();
+    route = router.upstreamPool(transaction.key_);
+    Common::Redis::RespValueSharedPtr multi_request =
+        std::make_shared<Common::Redis::Client::MultiRequest>();
+    if (route) {
+      makeSingleServerRequest(route, "MULTI", transaction.key_, multi_request, null_pool_callbacks,
```
weisisea, line 512: For this MULTI command, the Redis node will return "OK" response. Will we forward this "OK" back to the client? If yes, it seems the client will be receive two "OK" responses (the first "OK" response is at line 476).

weisisea:
Hi @asheryerm, could you please check what will happen if the Redis node closes the connection and the client continues to send commands for the transaction? Wanted to make sure that we won't reference an object/pointer after it's destroyed.

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
asheryerm: Fixed

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
asheryerm: The "real" MULTI command is passed the null_pool_callbacks, which basically means that its response is discarded.

test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc
asheryerm: This test is very challenging because the transaction opens a connection and the unit test infrastructure doesn't support this at the moment. (Or I am not using it correctly). I have tried several different ways to do this and I haven't been successful so far. Do you have a suggestion for a test implementation that will work.

In the meantime, I've replaced this test with a test that verifies that a MULTI command is sent before the first simple command in the transaction.

source/extensions/filters/network/common/redis/client.h
asheryerm: I have tried this approach but it causes segmentation faults.
The reason seems to be to be that the client is not closed synchronously, so running:
`client_->close()`
`client = nullptr`
does not work.

asheryerm:
> Hi @asheryerm, could you please check what will happen if the Redis node closes the connection and the client continues to send commands for the transaction? Wanted to make sure that we won't reference an object/pointer after it's destroyed.

Hey @weisisea I tested this connecting envoy to a redis-server and then killing the server in the middle of the transaction:
I added printouts in the logs to see what is happening. Once the server is killed it leads to the ProxyFilter being destroyed, which destroys the transaction and closes any open connections
`proxy_1  | [2022-09-26 13:57:48.331][33][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:95] ASHER: Received closing event
proxy_1  | [2022-09-26 13:57:48.331][33][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:67] ASHER: In ProxyFilter::~ProxyFilter
proxy_1  | [2022-09-26 13:57:54.568][16][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:95] ASHER: Received closing event
proxy_1  | [2022-09-26 13:57:54.568][16][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:67] ASHER: In ProxyFilter::~ProxyFilter`

source/extensions/filters/network/common/redis/client.h
asheryerm: I thought this might be a bit redundant since I will have to add a getter and a setter for each var, so I decided to treat it more like a struct. What do you think?

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
asheryerm: The formatting check in the CI doesn't like this change

source/extensions/filters/network/common/redis/client.h
weisisea: Hmm.. I think Envoy code style seems to make variables private in a class. Maybe you could simply make it a struct?

source/extensions/filters/network/common/redis/client.h
weisisea: OK, sounds good

test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc
weisisea: Makes sense. There might need some more changes for the dedicated connection. Could you please add a TODO here for the full transaction test?

source/extensions/filters/network/redis_proxy/command_splitter_impl.cc
weisisea: Got it, thanks for checking!

weisisea:
> > Hi @asheryerm, could you please check what will happen if the Redis node closes the connection and the client continues to send commands for the transaction? Wanted to make sure that we won't reference an object/pointer after it's destroyed.
> 
> Hey @weisisea I tested this by connecting envoy to a redis-server and then killing the server in the middle of the transaction: I added printouts in the logs to see what is happening. Once the server is killed it leads to the ProxyFilter being destroyed, which destroys the transaction and closes any open connections `proxy_1 | [2022-09-26 13:57:48.331][33][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:95] ASHER: Received closing event proxy_1 | [2022-09-26 13:57:48.331][33][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:67] ASHER: In ProxyFilter::~ProxyFilter proxy_1 | [2022-09-26 13:57:54.568][16][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:95] ASHER: Received closing event proxy_1 | [2022-09-26 13:57:54.568][16][info][redis] [source/extensions/filters/network/redis_proxy/proxy_filter.cc:67] ASHER: In ProxyFilter::~ProxyFilter`
> 
> If the client continues sending commands past this point they will receive an `upstream failure`

Closing the downstream connection (ProxyFilter object) upon the dedicated connection being closed sounds good to me since we probably don't want the rest of the commands to be run outside of the transaction. Could you please elaborate more on how closing the downstream connection is achieved?

weisisea (approved):

weisisea:
A follow-up question, is there a plan on incorporating WATCH command support as well to complete the whole transaction story for Redis? Thanks!

weisisea:
Hi @mattklein123, the changes look good to me. Do you want to take a pass at it?

test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc
asheryerm: Yes, of course

asheryerm:
> Closing the downstream connection (ProxyFilter object) upon the dedicated connection being closed sounds good to me since we probably don't want the rest of the commands to be run outside of the transaction. Could you please elaborate more on how closing the downstream connection is achieved?

When the dedicated connection is closed the dispatcher sends a close event to the dedicated. However, the dedicated connection uses the ProxyFilter for its connection callbacks, so this leads to a termination of the ProxyFilter.

asheryerm:
> A follow-up question, is there a plan on incorporating WATCH command support as well to complete the whole transaction story for Redis? Thanks!

If this current PR is merged and you are more or less happy with the way I did it, then I can definitely look into supporting WATCH as well. It will probably be a smaller change once we have transaction support in place.

test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc
asheryerm: Done

weisisea:
> > A follow-up question, is there a plan on incorporating WATCH command support as well to complete the whole transaction story for Redis? Thanks!
> 
> If this current PR is merged and you are more or less happy with the way I did it, then I can definitely look into supporting WATCH as well. It will probably be a smaller change once we have transaction support in place.

Great, thanks!

changelogs/current.yaml
```diff
@@ -204,6 +204,9 @@ new_features:
 - area: access_log
   change: |
     updated command operator ``%GRPC_STATUS%`` to suppoprt the snake case.
+- area: redis
+  change: |
+    added support for redis transactions.
```
mattklein123, line 209: Can you modify https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/redis to talk about the transaction support briefly?

mattklein123 (changes requested):
Code LGTM. One small documentation request. Thank you!

/wait

changelogs/current.yaml
asheryerm: Done.

mattklein123 (approved):
Thanks!

PR #22731, opened 2022-08-16
json: avoid testing a 3-byte utf8 code-point that mismatches expectations in a future protobuf library upgrade

jmarantz:
Commit Message: The next protobuf upgrade will break a few JSON-related tests, mostly due to brittle dependence on error-message syntax. However this one is a consistency check for serializing 3-byte utf-8 code-points, and there is one that is serialized differently in the new version of the protobuf library than the previous one.
Additional Description:
Risk Level: low
Testing: test/common/json/...
Docs Changes: n/a
Release Notes: n/a


jmarantz:
This was fixed on the protobuf side.

Issue #19948, opened 2022-02-15
Could Envoy support regex PartialMatch?


clark1013:
*Description*:
I am trying to migrate from Nginx to Envoy. There are a lot of regex configs only match a subsequence of the uri. But Enovy's `safe_regex` forces to match the entire uri. So could envoy support this scene through `re2::RE2::PartialMatch`?

[optional *Relevant Links*:]
Nginx regex match use `pcre_exec`: https://www.pcre.org/original/doc/html/pcre_exec.html
It performs like this:
```
Source Code:
    pcre *re = pcre_compile("^/a", 0, &errstr, &erroff, NULL);
    if (re == NULL) {
        printf("regex compile failed: %s", errstr);
        return 1;
    }
    char *subjects[4] = {"/a", "/a/b/c", "/a/c", "/b/c"};
    for (int i=0; i<4; i++) {
        char *subject = subjects[i];
        int rc = pcre_exec(re, NULL, subject, strlen(subject), 0, 0, NULL, 0);
        printf("%s match result %d\n", subject, rc);
    }

Output:
/a match result 0
/a/b/c match result 0
/a/c match result 0
/b/c match result -1
```

daixiang0:
Nginx and Envoy are applicable to different fields, could you share the case you want to migrate?

clark1013:
> Nginx and Envoy are applicable to different fields, could you share the case you want to migrate?

@daixiang0 Currently our API gateway is based on Nginx, It's responsible for route a request to a service, The service discovery is based on Consul. Recently we are building our service mesh based on istio and Envoy, we want to directly use Envoy as the API gateway, so we need to migrate route configs from Nginx to Envoy.

KBaichoo:
cc @junr03 @snowp

KfreeZ:
@zhxie would hyperscan help with this?

jmarantz:
instead of PartialMatch could you use FullMatch, but modify the pattern so you don't anchor on the beginning/end? Sorry this may be a naive question. E.g. this test fragment works for me:
```
  RE2 re(".*abc.*");
  EXPECT_TRUE(RE2::FullMatch("xabcd", re));
```

clark1013:
> instead of PartialMatch could you use FullMatch, but modify the pattern so you don't anchor on the beginning/end? Sorry this may be a naive question. E.g. this test fragment works for me:
> 
> ```
>   RE2 re(".*abc.*");
[...]

PR #22733, opened 2022-08-17
regex: support partial matching in RE2

zhxie:

Commit Message: regex: support partial matching in RE2
Additional Description:

Earlier this year, a user asks to support partial match in RE2 in #19948. The user finds difficulty to migrate API gateway from Nginx to Envoy since they have many regex patterns in data processing, while the Nginx's regex engine (and most other proxies and WAFs using PCRE-like regex engine) matches any substring in the input against the regex pattern, which is the same behavior of partial matching in RE2.

Previously, we can modify the regex pattern `/A/` to `/.*A.*/` as a workaround, but since we have implemented regex engine interface, we can add an option to enable partial matching globally. It will be more friendly for our users.

Risk Level: Low
Testing: Integration
Docs Changes: N/A
Release Notes: Added

### DIFF
--- a/api/envoy/extensions/regex_engines/v3/google_re2.proto
+++ b/api/envoy/extensions/regex_engines/v3/google_re2.proto
@@ -21,4 +21,8 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // which records the program size, and the counter ``re2.exceeded_warn_level``, which is incremented
 // each time the program size exceeds the warn level threshold.
 message GoogleRE2 {
+  // Allow to return matched if a substring of the input matches the regex pattern. By default, the
+  // engine will return matched only if the input matches the regex pattern fully, from the
+  // beginning to the end of the input.
+  bool partial_match = 1;
 }
--- a/source/common/common/regex.h
+++ b/source/common/common/regex.h
@@ -57,7 +57,6 @@ public:
 
 class GoogleReEngine : public Engine {
 public:
-  GoogleReEngine();
   GoogleReEngine(bool partial_match);
 
   CompiledMatcherPtr matcher(const std::string& regex) const override;
--- a/source/common/common/regex.h
+++ b/source/common/common/regex.h
@@ -57,6 +57,7 @@ public:
 
 class GoogleReEngine : public Engine {
 public:
+  GoogleReEngine();
   GoogleReEngine(bool partial_match);
 
   CompiledMatcherPtr matcher(const std::string& regex) const override;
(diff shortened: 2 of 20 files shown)


api/envoy/extensions/regex_engines/v3/google_re2.proto
```diff
@@ -21,4 +21,8 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // which records the program size, and the counter ``re2.exceeded_warn_level``, which is incremented
 // each time the program size exceeds the warn level threshold.
 message GoogleRE2 {
+  // Allow to return matched if a substring of the input matches the regex pattern. By default, the
+  // engine will return matched only if the input matches the regex pattern fully, from the
+  // beginning to the end of the input.
+  bool partial_match = 1;
```
mattklein123, line 27: Please add a release note.

source/common/common/regex.h
```diff
@@ -40,13 +40,30 @@ class CompiledGoogleReMatcher : public CompiledMatcher {
...
+
+  // CompiledGoogleReMatcher
+  bool match(absl::string_view value) const override {
+    return re2::RE2::PartialMatch(re2::StringPiece(value.data(), value.size()), regex_);
+  }
+};
+
 class GoogleReEngine : public Engine {
 public:
+  GoogleReEngine();
```
mattklein123, line 60: Why is the default constructor needed?

mattklein123 (changes requested):
Thanks for working on this. This seems OK, though at a high level I'm wondering if partial match regex should actually be a property of the higher layer matcher config as I'm guessing all engines would support this? This would be more user friendly if that's ultimately why we are doing this. cc @jmarantz for thoughts.

/wait

source/common/common/regex.h
zhxie: Oh, it is not needed. I just attempted to pass the CI and see if there if further issues in the patch. I will fix that.

api/envoy/extensions/regex_engines/v3/google_re2.proto
zhxie: Added, thanks.

markdroth:
> though at a high level I'm wondering if partial match regex should actually be a property of the higher layer matcher config as I'm guessing all engines would support this? This would be more user friendly if that's ultimately why we are doing this.

FWIW, I agree with this.  I think this option would make a lot more sense as a field in the `RegexMatcher` proto.

mattklein123:
> FWIW, I agree with this. I think this option would make a lot more sense as a field in the RegexMatcher proto.

OK let's go with this. @zhxie can you rework this PR? Thank you.

/wait

tony612:
@zhxie Any change to merge this PR?

PR #22745, opened 2022-08-17
[v1.22] ci: disable non-RBE cache for release branches

lizan:
Signed-off-by: Lizan Zhou <lizan@tetrate.io>

### DIFF
--- a/.azure-pipelines/bazel.yml
+++ b/.azure-pipelines/bazel.yml
@@ -55,7 +55,6 @@ steps:
       GCP_SERVICE_ACCOUNT_KEY: $(GcpServiceAccountKey)
     ${{ if eq(parameters.rbe, false) }}:
       BAZEL_BUILD_EXTRA_OPTIONS: "${{ parameters.bazelBuildExtraOptions }}"
-      BAZEL_REMOTE_CACHE: $(LocalBuildCache)
 
   displayName: "Run CI script"
 


phlax (approved):
lgtm, thanks @lizan

.azure-pipelines/bazel.yml
```diff
@@ -55,7 +55,6 @@ steps:
       GCP_SERVICE_ACCOUNT_KEY: $(GcpServiceAccountKey)
     ${{ if eq(parameters.rbe, false) }}:
       BAZEL_BUILD_EXTRA_OPTIONS: "${{ parameters.bazelBuildExtraOptions }}"
-      BAZEL_REMOTE_CACHE: $(LocalBuildCache)
```
alyssawilk, line 58: aren't we going to have to remember to do this for every branch we create?  Can we make this conditional on being in the main branch in yaml on main, so as we create new rease branches they do the right thing, and/or update "create a release" instructions so we remember to follow up if automating is too tricky?

phlax: i think we need a better solution - but for now its a case of just keeping the branches working - ill open a ticket ...

phlax: #24001

alyssawilk: Awesome, thanks

Issue #22748, opened 2022-08-17
codec: support for half-open H2 streams


kyessenov:
Current implementation does not handle half-closed HTTP streams well. When a response is received, the stream is assumed to be destroyed (in particular, the underlying connection can be closed if it's drained/idle). This is especially problematic for HTTP CONNECT streams: when tcp_proxy attempts to write end_stream after response is complete for a draining cluster (e.g. via CDS update), a crash occurs.

This is a feature request to support an explicit closure of request stream and delay stream destruction until then.

buffalu:
we (mainly @segfaultdoctor) have been testing envoy with GRPC trying to make sure it'll work for our application. we have the following service file:
```
service ExampleService {
  rpc Echo (Message) returns (Message) {}
  rpc StreamToClient (Message) returns (stream Message) {}
  rpc StreamToServer (stream Message) returns (Message) {}
  rpc EchoStream (stream Message) returns (stream Message) {}
}
```

all of them work except StreamToServer. is this issue and #22754 supposed to address the issues we see with StreamToServer getting dropped on envoy but not directly?

kyessenov:
Yes, the issue is that Envoy considers the response completion as the stream completion, so if the client continues sending data after the upstream server ends its stream, the data is not sent upstream. This happens at the codec level, so there may be some variation in how gRPC SDKs encode streaming semantics in HTTP2.

Issue #2066, opened 2017-11-15
http1.1: support half closed clients


ldemailly:
In http1.1 a client can choose to add "Connection: close" to the header and half close their (write) end of the socket and wait for the answer

What happens is envoy does forward the request but then immediately closes both end (probably because it reads the EOF/readability of 0 as an error instead of normal)

Repro
```
$ fortio curl -loglevel=debug -keepalive=false -halfclose http://echosrv2:8080/debug
18:41:30 D http.go:290> Created client:
{IP:10.39.255.58 Port:8080 Zone:}
GET /debug HTTP/1.1
Host: echosrv2:8080
Connection: close
User-Agent: istio/fortio-0.3.6

18:41:30 D http.go:504> Half closed ok after sending request &{{0xc4200e25b0}} {10.39.255.58 8080 }
18:41:30 E http.go:545> EOF before reading anything on &{{0xc4200e25b0}} {10.39.255.58 8080 }
18:41:30 V fortio_main.go:141> Fetch result code -1, data len 0, headerlen 0
```
envoy log
```
[2017-11-16T18:41:30.081Z] "GET /debug HTTP/1.1" 0 - 0 0 0 - "-" "istio/fortio-0.3.6" "e16c359b-048a-47a5-b5ab-b5e917bd2f72" "echosrv2:8080" "10.36.2.8:8080"
```

Spec extract
> Servers SHOULD always respond to at least one request per connection, if at all possible. Servers SHOULD NOT close a connection in the middle of transmitting a response, unless a network or client failure is suspected.

ldemailly:
cc @ggreenway who mentioned he is fixing it for tcp

ggreenway:
Tracking tcp proxy issue at https://github.com/envoyproxy/envoy/issues/2067

fkowal:
I am tracking the same bug i think.
```
echo -e "GET /status/ping HTTP/1.1\r\nHost: example\r\n\r\n" | nc -v $IP $PORT
```
I this example the client send a tcp FIN, i guess this is what you mean as halfclose.
The Connection: close header is just a representation of that

kthommandra:
Does anyone know if mainstream browsers (Chrome, Firefox etc) can end up doing TCP half-close under certain scenarios ?

kthommandra:
It clearly shows that there are web apps that end up doing TCP half-close
(eg: see attachment Reviewboard app in Firefox)

[...]

PR #22754, opened 2022-08-18
codec: support half close

kyessenov:

Commit Message: Hold `ActiveRequest` destruction until a full close instead of doing immediately on response decode. 
Additional Description: This is needed to support half closed tcp_proxy CONNECT streams. Closing immediately on response triggers TCP connection close on draining upstream which breaks tcp_proxy continuing to send data. Fixes https://github.com/envoyproxy/envoy/issues/22748
 
Risk Level: medium / high
Testing: manually verified the crash is fixed, checking no regressions, will add dedicated tests once the approach is agreed
Docs Changes: none
Release Notes: yes


kyessenov:
cc @birenroy @mattklein123

kyessenov:
I think I have it close for h2 (and probably quic). I preserved the "response close = stream close" for HTTP/1.1 as before for tests to pass, since fixing that looks to me quite hard, and has been around for quite awhile (see https://github.com/envoyproxy/envoy/issues/2066).

envoy/http/codec.h
```diff
@@ -291,6 +291,11 @@ class StreamCallbacks {
    * watermark to under its low watermark.
    */
   virtual void onBelowWriteBufferLowWatermark() PURE;
+
+  /**
+   * Fires when a stream is ended (half-closed).
+   */
+  virtual void onStreamEnd() PURE;
```
alyssawilk, line 298: what this is doing isn't clear to me.   You can have end stream for request path and response path?

source/common/http/codec_client.cc
```diff
@@ -128,8 +128,25 @@ void CodecClient::responsePreDecodeComplete(ActiveRequest& request) {
...
 
+  request.decode_complete_ = true;
+  if (request.encode_complete_) {
+    completeRequest(request);
+  } else {
+    ENVOY_CONN_LOG(debug, "waiting for encode to complete", *connection_);
+  }
+}
+
+void CodecClient::onRequestStreamEnd(ActiveRequest& request) {
```
alyssawilk, line 140: Envoy tends towards Decode/Encode instead of Request/Response :-(

source/common/http/codec_client.cc
```diff
@@ -128,8 +128,25 @@ void CodecClient::responsePreDecodeComplete(ActiveRequest& request) {
   if (codec_client_callbacks_) {
     codec_client_callbacks_->onStreamPreDecodeComplete();
   }
-  deleteRequest(request);
 
+  request.decode_complete_ = true;
```
alyssawilk, line 132: if the DecodeComplete calls can handle the decode completion, should we use RequestEncoderWrapper/onEncodeComplete for the encode  side?

alyssawilk (commented):
If I'm reading this correctly, this is less about supporting half close (since we're not supporting HTTP/1.1 which actually uses TCP half close) and more about allowing either upstream or downstream sending end stream first?
Either way, thanks for tracking this down!
/wait

envoy/http/codec.h
kyessenov: Yes, this is a per-stream end, ES in diagram https://httpwg.org/specs/rfc9113.html#StreamStates.

source/common/http/codec_client.cc
kyessenov: Thanks, will update. It's quite confusing since codec is used upstream.

source/common/http/codec_client.cc
kyessenov: Oh, yeah, that looks like a good idea.

kyessenov:
More elegant suggestion https://github.com/envoyproxy/envoy/pull/22802.

PR #22761, opened 2022-08-18
[API]xds: Add additional metadata field to Resource wrapper

botengyao:

Added  a new metadata to Resources wrapper to unblock the possibility to carry more ad-hoc info, e.g. the trace data for each resources. 

Commit Message:
Additional Description:
Risk Level: Low
Testing: 
Docs Changes:

### DIFF
--- a/api/envoy/service/discovery/v3/discovery.proto
+++ b/api/envoy/service/discovery/v3/discovery.proto
@@ -386,6 +386,5 @@ message Resource {
 
   // The Metadata field can be used to provide additional information for the resource.
   // E.g. the trace data for debugging.
-  // [#not-implemented-hide:]
   config.core.v3.Metadata metadata = 9;
 }
--- a/api/envoy/service/discovery/v3/discovery.proto
+++ b/api/envoy/service/discovery/v3/discovery.proto
@@ -332,7 +332,7 @@ message DynamicParameterConstraints {
   }
 }
 
-// [#next-free-field: 9]
+// [#next-free-field: 10]
 message Resource {
   option (udpa.annotations.versioning).previous_message_type = "envoy.api.v2.Resource";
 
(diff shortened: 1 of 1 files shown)


api/envoy/service/discovery/v3/discovery.proto
```diff
@@ -383,4 +383,9 @@ message Resource {
   // Cache control properties for the resource.
   // [#not-implemented-hide:]
   CacheControl cache_control = 7;
+
+  // The Metadata field can be used to provide additional information for the resource.
+  // E.g. the trace data for debugging.
+  // [#not-implemented-hide:]
```
mattklein123, line 389: Are you planning on an Envoy side implementation? If not I would probably just remove this if it's a general property of the protocol. Otherwise this is simple enough do you want to just do the implementation?

It's not exactly clear to me if this is for protocol debugging or if you expect this to get merged into the actual physical resource that it's matched with? I assume the former is if the latter it would be part of that resource type?

mattklein123 (commented):
Thanks generally LGTM with a question.

/wait-any

api/envoy/service/discovery/v3/discovery.proto
botengyao: There could be an Envoy side implementation, like a new extension point, but the usage could be in external package. I think it is fine to remove the hide comment.  Right, I am not expecting to merge into the actual resource right now.

htuch: This work stream will necessitate Envoy side work; even if we're consuming in our extensions @botengyao we will need to have some record keeping as the resources are ingested. 

I agree with @mattklein123 though, since this is basically a transport protocol property and can be useful even without implementation we can go with this.

htuch (approved):
/lgtm api

mattklein123 (approved):
Thanks

PR #22764, opened 2022-08-18
config: refactoring opaque resource decoder to be a unique-ptr

adisuissa:
Commit Message: refactoring opaque resource decoder to be a unique-ptr
Additional Description:
Prior to this PR the OpaqueResourceDecoder was owned by each of the Subscriptions, and a reference to it was used by the GrpcMux's. This can cause an issue if the Mux outlives the subscription, as can happen in the newly added test ValidResourceDecoderAfterRemoval, and in the unified Mux implementation.
This PR converts the OpaqueResourceDecoder to a unique pointer, that can be used by both.
Followup by request in #22596.

Risk Level: Low/Medium - modifying many places in the config-plane codebase, but the refactor switches from a reference to a unique pointer.
Testing: Added a unit test that fails if non-unique pointer is used.
Docs Changes: N/A.
Release Notes: TBD.

### DIFF
--- a/envoy/config/grpc_mux.h
+++ b/envoy/config/grpc_mux.h
@@ -100,7 +100,7 @@ public:
   virtual GrpcMuxWatchPtr addWatch(const std::string& type_url,
                                    const absl::flat_hash_set<std::string>& resources,
                                    SubscriptionCallbacks& callbacks,
-                                   OpaqueResourceDecoder& resource_decoder,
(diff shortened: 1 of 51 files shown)


adisuissa:
cc @htuch as you requested in #22596, here's the implementation that uses unique-ptr.
To tell the truth, I feel both shared and unique are a bit risky as we may miss some edge case in the tests, but when looking at the changes again, I'm not certain if either is more riskier than the other.
Please let me know what you think.

PR #22743, opened 2022-08-17
deps: Update gRPC to version v1.49.0-pre1 

yanavlasov:
Commit message:
Update to gRPC v1.49.0-pre1 which is a little newer than 1.48.0 to accommodate refactor to xDS protobuf files in Envoy.

Risk Level: Low
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: N/A

Signed-off-by: Yan Avlasov <yavlasov@google.com>

PR #22781, opened 2022-08-19
bazel: update upb

keith:
This is required to build with newer versions of bazel. This isn't the
newest commit since that would also require a protobuf update.


moderation:
Dupe of https://github.com/envoyproxy/envoy/pull/22743 which is bringing in a newer version of `upb`?

keith:
ah yep nice, if that one lands we can close this one

wrowe:
As #22743 is more current passes the CI pipeline, let's pick up that change to 6-21-2022 release..

PR #22784, opened 2022-08-19
upstream: Use std::weak_ptr::lock() to construct shared pointers in EdfScheduler

pianiststickman:
Checking and then using std::weak_ptr::expired() is inherently thread-unsafe:
https://en.cppreference.com/w/cpp/memory/weak_ptr/expired

Signed-off-by: Eugene Chan <eugenechan@google.com>

Commit Message: upstream: Use std::weak_ptr::lock() to construct shared pointers in EdfScheduler
Additional Description:
Risk Level: low (should be no-op)
Testing: existing CI
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/upstream/edf_scheduler.h
+++ b/source/common/upstream/edf_scheduler.h
@@ -29,34 +29,28 @@ template <class C> class EdfScheduler : public Scheduler<C> {
 public:
   // See scheduler.h for an explanation of each public method.
   std::shared_ptr<C> peekAgain(std::function<double(const C&)> calculate_weight) override {
-    if (hasEntry()) {
-      prepick_list_.push_back(std::move(queue_.top().entry_));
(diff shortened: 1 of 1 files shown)


htuch (approved):
LGTM, would be nice to have a test but kind of tricky for this kind of race condition. @alyssawilk can you take a look at this as well to verify as you spent some time here on the prepick logic? Thanks.

alyssawilk:
linked doc says "This function is inherently racy if the managed object is shared among threads."
Sorry I'm still digging out from vacation so going on 4 year old memory but aren't all the schedulers in Envoy per-thread? If so I assume this would slightly decrease perf and wouldn't improve thread saftey?

pianiststickman:
The schedulers are per-thread but from what I can tell the Hosts that they schedule can be shared across threads.

alyssawilk (approved):

Issue #22591, opened 2022-08-06
stats: need tag-extractor support for multiple regexes targeting same tag name


zirain:
Istio use ` "regex": "(response_code=\\.=(.+?);\\.;)|_rq(_(\\.d{3}))$"` for response_code stats tag, [here](https://github.com/istio/istio/blob/f0bb371da14af5229cd297c76fc7145771aec6db/tools/packaging/common/envoy_bootstrap.json#L48) is the source code.

## step 1
I test with the following stats, which work as expected:
```yaml
stats_config:
  use_all_default_tags: false
  stats_tags:
    - tag_name: response_code
      regex: "_rq(_(\\d{3}))$"
      #regex: "(response_code=\\.=(.+?);\\.;)|_rq(_(\\d{3}))$"
      #regex: "_rq(_(\\d{3}))$|(response_code=\\.=(.+?);\\.;)"
admin:
  access_log_path: /tmp/admin_access.log
  address:
    socket_address:
      protocol: TCP
      address: 127.0.0.1
      port_value: 9901
static_resources:
  listeners:
    - name: listener_0
      address:
        socket_address:
          protocol: TCP
          address: 0.0.0.0
          port_value: 10000
      filter_chains:
        - filters:
            - name: envoy.filters.network.http_connection_manager
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
                stat_prefix: ingress_http
                codec_type: auto
                route_config:
                  name: local_route
                  virtual_hosts:
                    - name: local_service
                      domains:
                        - "*"
                      routes:
                        - match:
                            prefix: "/"
                          route:
                            cluster: default_service
                http_filters:
                  - name: envoy.filters.http.router
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router

    - name: staticreply-8080
      address:
        socket_address:
          address: 127.0.0.1
          port_value: 8080
      filter_chains:
        - filters:
            - name: envoy.http_connection_manager
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
                stat_prefix: ingress_http
                codec_type: auto
                route_config:
                  name: local_route
                  virtual_hosts:
                    - name: local_service
                      domains:
                        - "*"
                      routes:
                        - match:
                            prefix: "/"
                          direct_response:
                            status: 200
                            body:
[...]

Issue #14071, opened 2020-11-17
stats: allow pre-defined tags


kyessenov:
The internal API to create tags has two versions:
1. Create from qualified stat name.
2. Create from stat name plus a set of tags. 

We'd like to have a third version:
1. Create from stat name plus a set of tag names. Tag names match against the content of the stat to obtain values. For example, stat "a.b.c" with tag names "d.e" produces "a{d=b,e=c}". Here the separator is a dot, but could be any special character.

The benefit is two-fold:
1. Save on space, instead of "a.d.b.e.c" we can share "d" and "e" symbols between stat names with different tag values.
2. Simplify user experience by not required regexes to achieve the same effect.

mattklein123:
cc @jmarantz to work with you on this hopefully.

jmarantz:
Questions:

1. what's the potential space advantage? Note that given that symbol tables are already integrated it might be less than you expect.
2. when using the interface to create from stat name plus a set of tags, do we need a regex? I didn't think we did.

kyessenov:
My issue is not the space, but the requirement to supply regexes to post-factum extract the tags.
As for 2), Wasm ABI does all or nothing, and I think today it's not using the set of tags option. We may need to re-design the stats parts of ABI to expose the alternative version to create a metic.

jmarantz:
Why do you need the RE if you use `counterFromStatNameWithTags`?

kyessenov:
@jmarantz REs are needed for Wasm extensions because the ABI only has a simple by-name stat: https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/master/proxy_wasm_externs.h#L142. I agree that we could fix it via an ABI change and Wasm filter improvements. Do you think the ABI should just expose `counterFromStatNameWithTags`? I think the challenge is to make the lifecycle easy to use from extensions. As an example, here's what opentelemetry looks like today: https://github.com/open-telemetry/opentelemetry-cpp/blob/master/examples/metrics_simple/main.cc#L35

jmarantz:
Honestly I feel like we should deep-dive on the wasm stats ABI. I have been concerned about it for some time because I think by going for maximum simplicity it is going to potentially impact performance and contention in particular. 

Why don't we set up some time to talk after thanksgiving?

As background, it's probably worth going over Splunk's KubeCon slides from last year where severe stats scalability issues were uncovered in Prometheus:

[...]

PR #22791, opened 2022-08-22
stats: fix response_code tag-extractor not working in istio

zirain:
Signed-off-by: hejianpeng <hejianpeng2@huawei.com>

In short, in istio we use `"regex": "(response_code=\\.=(.+?);\\.;)|_rq(_(\\d{3}))$"` which should working for both wasm stats and envoy stats.

the original issue is [here](https://github.com/istio/istio/issues/39903), more details can be found [here](https://github.com/envoyproxy/envoy/issues/22591)

cc @kyessenov @jmarantz PTAL

Commit Message: stats: fix response_code tag-extractor not working in istio
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] https://github.com/envoyproxy/envoy/issues/22591
[Optional Deprecated:]


test/common/stats/tag_extractor_impl_test.cc
```diff
@@ -31,6 +31,22 @@ TEST(TagExtractorTest, TwoSubexpressions) {
   EXPECT_EQ("cluster_name", tags.at(0).name_);
 }
 
+TEST(TagExtractorTest, IstioResponseCode) {
+  TagExtractorStdRegexImpl tag_extractor("response_code",
+                                         "(response_code=\\.=(.+?);\\.;)|_rq(_(\\d{3}))$");
```
kyessenov, line 36: Does the non-capturing group `(?: ...)` not work for this use-case? I'm not sure what std::regex produces when you have this OR.

zirain: this regex is copied from [here](https://github.com/istio/istio/blob/d7d55360475e1c0ef0150f006dddea7e5eeb3c62/tools/packaging/common/envoy_bootstrap.json#L48) with removing of `.`

istio use this from the beginning, but seems it's not working as expected?

mattklein123:
@kyessenov ping on review on this?

test/common/stats/tag_extractor_impl_test.cc
kyessenov: It depends on the regex engine I suppose: re2 vs std::regex. The issue is that there are three groups here, and it matters how they are numbered as well as how they are populated when optional:

1: outer removal groups
2: left option: response_code=.=
3. right option: _rq.

That means, to answer my question, non-capturing doesn't help. Does this issue happen with re2 or just regex? AFAIR re2 supports optional capturing groups, which can be utilized in envoy. I  hesitate to treat empty value as missing value (it might break for cases when the regex matches empty). 

@jmarantz for regex expertise.

zirain: I think stats run on `StdRegex`, correct me if I'm wrong.

jmarantz: The built-in stats run purely on RE2. However, for legacy reasons, we left the std::regex impl for API-driven tag extraction, despite the fact that it's significantly slower.

Maybe we should add a parallel path for RE2-based API-driven tag extraction.

jmarantz: also, with respect to the capture-groups above, I'm not sure of the answer. I wouldn't say I'm a regex expert -- my knowledge is more around how to make systems run fast despite the presence of regexes, particularly the incredibly slow std::regex. I'll take a look at the PR.

source/common/stats/tag_extractor_impl.cc
```diff
@@ -114,13 +114,19 @@ bool TagExtractorStdRegexImpl::extractTag(TagExtractionContext& context, std::ve
                                                      regex_) &&
       match.size() > 1) {
     // remove_subexpr is the first submatch. It represents the portion of the string to be removed.
-    const auto& remove_subexpr = match[1];
+    uint idx;
+    for (idx = 1; idx < match.size(); idx++) {
+      if (match[idx].str() != "") {
```
kyessenov: I think this should be instead:

```
-      if (!match[idx].str().empty()) {
+      if (match[idx].matched) {
```

This way we continue to allow empty capturing groups, which would be a regression (maybe add a test for empty group since it appears there is none). Istio does allow setting empty string as tag values, for example.

source/common/stats/tag_extractor_impl.cc
```diff
@@ -114,13 +114,19 @@ bool TagExtractorStdRegexImpl::extractTag(TagExtractionContext& context, std::ve
...
+      }
+    }
+    const auto& remove_subexpr = match[idx];
 
     // value_subexpr is the optional second submatch. It is usually inside the first submatch
     // (remove_subexpr) to allow the expression to strip off extra characters that should be removed
     // from the string but also not necessary in the tag value ("." for example). If there is no
     // second submatch, then the value_subexpr is the same as the remove_subexpr.
-    const auto& value_subexpr = match.size() > 2 ? match[2] : remove_subexpr;
+    const auto& value_subexpr = idx + 1 < match.size() ? match[idx + 1] : remove_subexpr;
```
kyessenov, line 129: If we want to be consistent, this should be another `for` loop looking for a `match[idx].matched` past idx.

kyessenov (changes requested):
I think this is a good change, but I have a suggestion before LGTM.

test/common/stats/tag_extractor_impl_test.cc
kyessenov: I didn't realize we cannot set RE2 for bootstrap tag extractors. That's probably a separate change, and I agree, we do need to support optional capturing groups in RE2, too.

jmarantz: Can you make a parallel test for RE2? If there's a functional but for std::regex there might be one for re2 also.

source/common/stats/tag_extractor_impl.cc
```diff
@@ -114,13 +114,19 @@ bool TagExtractorStdRegexImpl::extractTag(TagExtractionContext& context, std::ve
                                                      regex_) &&
       match.size() > 1) {
     // remove_subexpr is the first submatch. It represents the portion of the string to be removed.
-    const auto& remove_subexpr = match[1];
+    uint32_t idx;
+    for (idx = 1; idx < match.size(); idx++) {
+      if (!match[idx].str().empty()) {
```
jmarantz, line 119: just to be pedantic, how do we know that the empty string isn't the correct matched tag value?

I don't necessarily have a better solution but I'm just thinking about possible corner-cases with this.

jmarantz (commented):
Can you change the description from simply referencing istio, to describing what's specfically broken in the current implementation and how this PR fixes it?

source/common/stats/tag_extractor_impl.cc
```diff
@@ -114,13 +114,27 @@ bool TagExtractorStdRegexImpl::extractTag(TagExtractionContext& context, std::ve
...
       match.size() > 1) {
     // remove_subexpr is the first submatch. It represents the portion of the string to be removed.
-    const auto& remove_subexpr = match[1];
+    uint32_t idx;
+    for (idx = 1; idx < match.size(); idx++) {
+      if (match[idx].matched) {
+        break;
+      }
+    }
+    const auto& remove_subexpr = match[idx];
```
jmarantz, line 123: If there is no match then idx==match.size() and this statement will (I believe) access memory outside the bounds of the array.

source/common/stats/tag_extractor_impl.cc
```diff
@@ -114,13 +114,27 @@ bool TagExtractorStdRegexImpl::extractTag(TagExtractionContext& context, std::ve
                                                      regex_) &&
       match.size() > 1) {
     // remove_subexpr is the first submatch. It represents the portion of the string to be removed.
-    const auto& remove_subexpr = match[1];
+    uint32_t idx;
```
jmarantz, line 117: rather than doing all of this, WDYT of changing the Istio regexes to only match one or two things?

You could also add something like:
```
if (match.size() > 2) {
  ENVOY_LOG_MISC(error, "tag extraction matched too many regexes");
  return false;
}
```

jmarantz (commented):
I understand that Istio used this from the beginning, but it's also the case that since early 2018, the semantics of tag-matching regexes was changed in a way that likely impacted Istio  stats functionally. 

Previously to that, regexes were evaluated in some well-defined order and iteratively mutated the stat name between each tag regex. I think that was implemented in late 2017 so it really only worked that way for a couple of months.

Now the system in a sense evaluates the regexes in parallel, and the all the tag names/values are only removed at the very end. This enabled critical Envoy optimizations for Istio's usage, which was the first time I heard of an Envoy configuration that needed to initialize with 10k clusters and not have the startup take minutes (due to these incredibly slow regexes).

It was probably a testing gap that this didn't catch the Istio dependency on the late 2017 behavior in early 2018, but I think it's too late now to change back. Can we just change Istio's usage instead of making this code more complex?

zirain:
> I understand that Istio used this from the beginning, but it's also the case that since early 2018, the semantics of tag-matching regexes was changed in a way that likely impacted Istio stats functionally.
> 
> Previously to that, regexes were evaluated in some well-defined order and iteratively mutated the stat name between each tag regex. I think that was implemented in late 2017 so it really only worked that way for a couple of months.
> 
> Now the system in a sense evaluates the regexes in parallel, and the all the tag names/values are only removed at the very end. This enabled critical Envoy optimizations for Istio's usage, which was the first time I heard of an Envoy configuration that needed to initialize with 10k clusters and not have the startup take minutes (due to these incredibly slow regexes).
> 
> It was probably a testing gap that this didn't catch the Istio dependency on the late 2017 behavior in early 2018, but I think it's too late now to change back. Can we just change Istio's usage instead of making this code more complex?

I agree with that do not `making this code more complex`, I'm OK to close this.

what's your suggestion for this case? I'm not very good at regex, can we support both the original envoy stats(`cluster.xds-grpc.upstream_rq_200`) and istio wasm stats(`wasmcustom.reporter=.=destination;.;source_workload=.=unknown;.;source_workload_namespace=.=unknown;.;source_principal=.=spiffe://cluster.local/ns/echo-2-62304/sa/default;.;source_app=.=unknown;.;source_version=.=unknown;.;source_canonical_service=.=unknown;.;source_canonical_revision=.=latest;.;source_cluster=.=unknown;.;destination_workload=.=server-v1;.;destination_workload_namespace=.=echo-2-62304;.;destination_principal=.=spiffe://cluster.local/ns/echo-2-62304/sa/default;.;destination_app=.=server;.;destination_version=.=v1;.;destination_service=.=server.echo-2-62304.svc.cluster.local;.;destination_service_name=.=server;.;destination_service_namespace=.=echo-2-62304;.;destination_canonical_service=.=server;.;destination_canonical_revision=.=v1;.;destination_cluster=.=Kubernetes;.;request_protocol=.=http;.;response_flags=.=NFCF;.;connection_security_policy=.=mutual_tls;.;response_code=.=500;.;grpc_response_status=.=;.;istio_response_bytes: P0(nan,60) P25(nan,60.25) P50(nan,60.5) P75(nan,60.75) P90(nan,60.9) P95(nan,60.95) P99(nan,60.99) P99.5(nan,60.995) P99.9(nan,60.999) P100(nan,61)`)  with one regex?

kyessenov:
@zirain I think the suggestion is to add _two_ tag extractors for the same tag instead of _one_ regex for two extractors. It seems like a more composable solution as long as it works.

zirain:
> @zirain I think the suggestion is to add _two_ tag extractors for the same tag instead of _one_ regex for two extractors. It seems like a more composable solution as long as it works.

IIRC, you can not set two tag extractors on same tag name.
if so, do you means keep `response_code: (response_code=\\.=(.+?);\\.;)` for wasm stats, add `envoy_response_code: _rq(_(\\d{3}))$` for envoy stats? correct me if I'm wrong?

this seems not friendly to end-users.

jmarantz:
I haven't directly used regexes specified outside the core system (source/common/config/well_known_names.cc) but that has two distinct ways to populate `envoy.http_conn_manager_prefix`. See https://github.com/envoyproxy/envoy/blob/main/source/common/config/well_known_names.cc#L133

What is it that prevents you from having two regexes populating the same tag?

zirain:
> I haven't directly used regexes specified outside the core system (source/common/config/well_known_names.cc) but that has two distinct ways to populate `envoy.http_conn_manager_prefix`. See https://github.com/envoyproxy/envoy/blob/main/source/common/config/well_known_names.cc#L133
> 
> What is it that prevents you from having two regexes populating the same tag?

got following error:
```
2022-09-02T01:55:10.419374Z	critical	envoy main	error initializing configuration 'etc/istio/proxy/envoy-rev.json': Tag name 'response_code' specified twice.
Tag name 'response_code' specified twice.
```
with stats config like:
```json
  "stats_config": {
    "use_all_default_tags": false,
    "stats_tags": [
      {
        "tag_name": "cluster_name",
        "regex": "^cluster\\.((.+?(\\..+?\\.svc\\.cluster\\.local)?)\\.)"
      },
      {
        "tag_name": "tcp_prefix",
        "regex": "^tcp\\.((.*?)\\.)\\w+?$"
      },
      {
        "regex": "(response_code=\\.=(.+?);\\.;)",
        "tag_name": "response_code"
      },
      {
        "regex": "_rq(_(\\d{3}))$",
        "tag_name": "response_code"
      },
      {
        "tag_name": "response_code_class",
        "regex": "_rq(_(\\dxx))$"
      },
```

jmarantz:
I see.  Maybe we could just fix that restriction? I don't know of a reason that restriction needs to exist anymore.

I see it enforced here: https://github.com/envoyproxy/envoy/blob/c5e7d7a30e4255b201302af9393f11494f03ebd3/source/common/stats/tag_producer_impl.cc#L24 

but only for the the CLI-specified and config-specified extractors, and not for the built-in extractors. Additionally it's enforced by populating a set and checking for already-inserted, but that set is then dropped without using it for anything. I think we could just drop the population and the exception, and add some tests to ensure this works as expected.

@mattklein123 @ggreenway WDYT?

zirain:
> I see. Maybe we could just fix that restriction? I don't know of a reason that restriction needs to exist anymore.
> 
> I see it enforced here:
> 
> https://github.com/envoyproxy/envoy/blob/c5e7d7a30e4255b201302af9393f11494f03ebd3/source/common/stats/tag_producer_impl.cc#L24
> 
> but only for the the CLI-specified and config-specified extractors, and not for the built-in extractors. Additionally it's enforced by populating a set and checking for already-inserted, but that set is then dropped without using it for anything. I think we could just drop the set population and the exception, and add some tests to ensure this works as expected.
> 
> @mattklein123 @ggreenway WDYT?
> 
> Also after having re-looked at your testcase, I'm not convinced they _ever_ worked as the current methodology of parallel extraction pre-dated wasm development in Istio/Envoy. I can see the argument to allow ORed match expressions but I think it's easier and probably more optimizable to allow multiple distinct patterns to match the same tag.
> 
> In particular, whenever a regex begins with `"^foo\."` there's a really important performance optimization that is made in the matching infrastructure to evaluate the regex onll against stat-names that match the first token. That's a lot easier to do if we use multiple highly specific regexes, rather than very complex regexes.

yes, I agree that.  Complex regex is hard to understand, we should keep every regex simple and short.
I will be glad to see envoy allow `allow multiple distinct patterns to match the same tag`.

ggreenway:
To confirm I understand this correctly: the proposal is to allow the same tag to be extracted with multiple rules, which would apply to different stats, correct? What would be the behavior if the multiple rules matched on the same stat? Would one overwrite the other? If so, which one wins?

jmarantz:
Note this situation exists in the built-in stats today: envoy.http_conn_manager_prefix. See https://github.com/envoyproxy/envoy/blob/main/source/common/config/well_known_names.cc#L133

I believe what happens is that the array of tag name/values would be populated with two different values for the same name. It would then be up to prometheus or whatever how to interpret that.

zirain:
> Note this situation exists in the built-in stats today: envoy.http_conn_manager_prefix. See https://github.com/envoyproxy/envoy/blob/main/source/common/config/well_known_names.cc#L133
> 
> I believe what happens is that the array of tag name/values would be populated with two different values for the same name. It would then be up to prometheus or whatever how to interpret that.

@jmarantz would you like to send a PR for this, I have no bandwidth in Sep.

jmarantz:
OK I can work on this in the background; as Greg gave the thumbs up :)

test/common/stats/tag_extractor_impl_test.cc
jmarantz: I'm working on trying to approach this by using 2 regex tag matchers rather than 1 with an OR.

But the regex in this testcase looks weird `(response_code=\\.=(.+?);\\.;)` . I really am not that good at regexes, and I don't understand what's with the 2 semi-colons, and why the response-code pattern doesn't looks like a 3-digit number; is that supposed to represent something other than an http response code?

Also there's nothing in this test that matches the first clause in the ORed regex, afaict.

jmarantz: Also this specific case seems odd as a testcase.   Envoy's built-in regexes already includes:
```
addRe2(RESPONSE_CODE_CLASS, R"(_rq_((\d))xx$)", "_rq_");
```

and as it stands that cannot be overridden, but once I remove the dup-check you could just add in response_code=.... form.

ggreenway: It kind of looks like `(response_code=\\.=(.+?);\\.;)` is trying to match on the prometheus formatted output, which doesn't make much sense. But the two `=` with any character in between doesn't quite fit.

kyessenov: This is the Wasm artifact. It generates concatenated metrics like `key=.=value;.;` literally and then relies on tag extraction to get keys and values back. This is far from optimal, but it's difficult to create free-form metrics any other way in Envoy.

jmarantz: This is why Envoy's API supports specifying tag name/value pairs explicitly during tag construction.

Would it make sense to iterate down that path? I think that might not be all it needs to for Wasm's use case but that could be changed, particularly if this is going to wind up sinking a lot of CPU encoding/decoding strings with regexes.

jmarantz: In any case the first part of that ORed regex is redundant with what's built into Envoy.

kyessenov: It would be great to stop creating mangled stat names, but we have looked twice there before, and I think the API doesn't fully support creating arbitrary key/value pairs explicitly as needed by Wasm. @douglas-reid I think was exploring it a few months ago.

jmarantz: yes but that might be a better solution if we can get there, then going down the path of more mangling

What do you perceive is missing?

I think what's missing from my perspective is that if you create a stat "foo" with tags x:1, y:2, and another stat "foo" with x:1, y:3, I think it will not really work. The names would have to be uniquified. I didn't try to do this myself.

But what are you seeing as the missing link to utilize specifying tag name/value pairs during construction?

If that's what it is we could have that API append the tag names/values to the stat name so each instance is unique.

jmarantz:
Anyway I have this working and could submit the PR for review, but while writing the test I began questioning the strategy. LMK what you think.

test/common/stats/tag_extractor_impl_test.cc
douglas-reid: @jmarantz iirc, the issue is that we don't have _all_ tag values at stats creation time. i believe the underlying envoy apis support dynamic tags, but that is not exposed (iirc) for wasm extensions. i agree that adding that capability to the Wasm ABI would be awesome (as would related features discussed in https://github.com/envoyproxy/envoy/issues/14071#issuecomment-1185084414).

jmarantz: If you don't have all the tag-values at creation time, how can it work today? All the tag names and values must be encoded in the stat name when it is created.

The only question is whether we use a regex to extract tag names and values from the fully elaborated stat name, or whether we start with the names and values given explicitly,  and construct a unique tag name for them.

kyessenov: The Wasm ABI just uses a string for defining a metric, see https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/wasm/context.cc#L1247. So the key is to make it accept a list of tags and also do better about reusing symbols in those tags.

kyessenov: I can try to take another stab at this issue. So my understanding is that we have to plug through tag list in several places:
1. Wasm ABI should take a list of tags, requires ABI bump (@PiotrSikora);
2. Envoy should preserve this list of tags for a custom metric.
3. Envoy needs to output the tags in prometheus-compatible format.

jmarantz: This approach makes sense to me from a performance perspective, both because of the lack of the tag-list in the API, and the usage of a string_view. And there's another issue with the API (which I pointed out years ago when it was being built) -- that using a string_view for the name argument means you'll need to take a symbol-table lock every time a stat is created from wasm.

That's OK if it happens when a module is loaded, but should probably not be considered acceptable if running in the data-plane. Or has wasm evolved to where you can run some init code during loading to establish counters and gauges in wasm space that can be accessed by data threads afterward?


From a purely functional perspective, though, it should be OK to compose stat names that encode the tag names/values. But why not use the tag formats for `rq_XXX` that's already built into wasm? Why is the `response_code=XXX; ` needed as well?

kyessenov: @jmarantz Re: locking. I don't think we can avoid adding a new key-value metric on the data path, that's just how the telemetry works in Istio mesh where we just don't know a-priori the clients or the servers we speak to during configuration. What we know is a list of key names, but not complete metrics. We are aware of the performance problems, and we don't require using this extension.

Re: format: The prometheus metric specification was ported from a different system so we have to keep the symbols the same.

jmarantz: RE locking: OK you need to re-create the symbols in the data path; that's survivable. But you need to have an API that takes a StatName (or array of StatNames) so that they can be held on the wasm side (and maybe replicated per core) and passed into the API. Otherwise wasm stat creation will never scale to large numbers of cores.

RE "prometheus metric" OK I don't fully understand that but I'm OK moving forward with https://github.com/envoyproxy/envoy/pull/23024 if you think that helps. It's just that the test regex copied from this PR doesn't appear to make sense to me. We can just have the test case be a single alternate regex to populate `envoy.response_code`.

kyessenov: I'm not deeply familiar with this Wasm code. Are you saying that Wasm context should hold on to storage better and be smarter about re-using already allocated symbols (per core)? Yes, I agree, and right now this is handled by the client code which avoids making ABI calls by caching those metric IDs.

jmarantz: Leaving this to clients might be OK and it might not, depending on how long-lived the client's context is, typically. If it's a long time then the system will eventually settle into reasonable performance.

Consider a test case -- I'm not sure if this is realistic. Let's say there's 100 threads, and each of those has a separate wasm context. Now say there are 100k stats, each of which are comprised of different permutations of fixed keywords such as "upstream" or "http", names of specific clusters, which are quite common, and perhaps a dozen different response-code segments such as rq_200 and rq_2xx.

Now let's say (perhaps pessimistically) that the wasm module gets reloaded every minute for some reason related to the control plane.

This means the number of symbol table locks that will be needed before all the wasm-side caches are warm is 100 * 10k = 1M. I think there is a risk of a burst of contention every minute, which would affect long-tail latency.

We tried very hard to do in the stats API internal to C++, is to make it easy to save the common segments that comprise the stat names in pools or sets that are initialized once at startup. When a counter is created by concatenating together a bunch of pre-symbolized stat name segments, no locks are needed.

I think it would be beneficial to wasm long-tail performance to expose this API (StatNamePool, StatNameSet, createCounterFromStatNames, et al). But I also understand that my scenario above may be pessimistic in a few dimensions, and maybe in practice it's OK.

kyessenov: Yes, our model does not handle xDS bursts very well. We use ECDS because we want to have exactly two copies of wasm contexts and the associated caches. I think the real reason why storage for us should be better exposed is because we have the opposite problem of metric expiry. We need to free symbols which I think doesn't happen because Wasm context uses one big arena for all symbols. Every new stat stays resident forever which has been noticed by some users that spin a lot of workloads with generated names.

Our team (and validated externally) opinion is that we should just rewrite our stats extension as a native extension, but we don't have available resources to do that.

jmarantz:
I'll close this for now in favor of #23024 -- let's see how the commentary on that goes.

PR #22796, opened 2022-08-22
quic: make server connection defer sending packets during processing incoming packets

danzh2010:
Commit Message: enable defer sending in QuicConnection::ProcessUdpPacket() on server side. This bundles packet writes for each connection at the end of the event loop instead of each packet read. It is for improving CPU cost of sendmsg() for QUIC listener, especially with UDP GSO batch writing.

Also change how closing connections are deferred during L4 filters creation. Without this change, the connection close is deferred till handling all types of delayed close in `QuicSession::OnCanWrite()` which `ProcessUdpPacket()` would call into upon handling the 1st incoming packet. With this change, connection close caused by L4 filter creation is handled right before calling ProcessUdpPacket() because OnCanWrite() is deferred.

Risk Level: medium, behavior change of quic write pipeline  
Testing: existing tests
Docs Changes: N/A
Release Notes: N/A
Runtime guard: envoy_reloadable_features_quic_defer_send_in_response_to_packet


RyanTheOptimist:
Looks like //test/integration:quic_protocol_integration_test is failing. Is that expected?

source/common/quic/envoy_quic_server_connection.cc
```diff
@@ -17,7 +17,13 @@ EnvoyQuicServerConnection::EnvoyQuicServerConnection(
     : quic::QuicConnection(server_connection_id, initial_self_address, initial_peer_address,
                            &helper, &alarm_factory, writer, owns_writer,
                            quic::Perspective::IS_SERVER, supported_versions),
-      QuicNetworkConnection(std::move(connection_socket)) {}
+      QuicNetworkConnection(std::move(connection_socket)) {
+  if (Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.quic_defer_send_in_response_to_packet")) {
+    set_defer_send_in_response_to_packets(GetQuicFlag(FLAGS_quic_defer_send_in_response));
+    defer_send_ = true;
```
RyanTheOptimist, line 24: out of curiosity, instead of adding a new envoy-specific member, would it make sense to expose `defer_send_in_response_to_packets_` via a getter on QuicConnection?

source/common/quic/envoy_quic_server_session.cc
```diff
@@ -191,5 +191,16 @@ quic::QuicSSLConfig EnvoyQuicServerSession::GetSSLConfig() const {
   return config;
 }
 
+void EnvoyQuicServerSession::ProcessUdpPacket(const quic::QuicSocketAddress& self_address,
```
RyanTheOptimist, line 194: Out of curiosity, would we want the Client session to have the same logic that the server session adds in this PR?

source/common/quic/quic_filter_manager_connection_impl.cc
```diff
@@ -146,10 +146,6 @@ void QuicFilterManagerConnectionImpl::updateBytesBuffered(size_t old_buffered_by
 
 void QuicFilterManagerConnectionImpl::maybeApplyDelayClosePolicy() {
   if (!inDelayedClose()) {
-    if (close_type_during_initialize_.has_value()) {
-      close(close_type_during_initialize_.value());
-      close_type_during_initialize_ = absl::nullopt;
-    }
```
RyanTheOptimist, line 152: Do you think there's any need for a feature flag to control this part of the change? Presumably you're confident that it's a safe change? Or should we guard it with the same flag?

RyanTheOptimist (commented):
I might be missing it, but do we have test which verifies the intended behavior for this change? Specifically that this "bundles packet writes for each connection at the end of the event loop instead of each packet read. " I was expecting to see a test were multiple packets were read and the flag controlled when packets were sent?

source/common/quic/quic_filter_manager_connection_impl.cc
danzh2010: It's guarded by close_type_during_initialize_. If the current feature flag is on, close_type_during_initialize_ wont' have value here, as OnCanWrite() happens outside of ProcessUdpPacket().

source/common/quic/envoy_quic_server_session.cc
danzh2010: We are not doing this in any existing QUICHE client. Maybe client CPU usage hasn't been a problem.

source/common/quic/envoy_quic_server_connection.cc
danzh2010: That's also possible.

danzh2010:
> I might be missing it, but do we have test which verifies the intended behavior for this change? Specifically that this "bundles packet writes for each connection at the end of the event loop instead of each packet read. " I was expecting to see a test were multiple packets were read and the flag controlled when packets were sent?

I modified test/common/quic/active_quic_listener_test.cc to check that send alarm is armed after receiving CHLO.

danzh2010:
> Looks like //test/integration:quic_protocol_integration_test is failing. Is that expected?

Nope, I'm looking into it. They are all upstream disconnect tests, but it's weird that they fail only in windows build.

source/common/quic/quic_filter_manager_connection_impl.cc
RyanTheOptimist: I don't think I understand. `maybeApplyDelayClosePolicy()` is called from a couple of different places. Before this PR, it's possible for `maybeApplyDelayClosePolicy()` to actually call `close()` if not in delayed close (on line 150). But with this PR, if we're not in delayed close then `maybeApplyDelayClosePolicy()` will not call `close()`.

But I guess you're saying that before this PR, `close_type_during_initialize_` will never have a value here. Do I have that right?

source/common/quic/envoy_quic_server_connection.cc
RyanTheOptimist: No need to block this PR, but care to cook up a quiche CL and eventually switch over to using it?

RyanTheOptimist:
> > I might be missing it, but do we have test which verifies the intended behavior for this change? Specifically that this "bundles packet writes for each connection at the end of the event loop instead of each packet read. " I was expecting to see a test were multiple packets were read and the flag controlled when packets were sent?
> 
> I modified test/common/quic/active_quic_listener_test.cc to check that send alarm is armed after receiving CHLO.

*nod* And presumably the send alarm being set leads to the behavior mentioned in the PR description. But it's not obvious to me that this is the case. (Though I certainly believe you that it is!) Would it be hard to have a test which actually waited to send something until multiple packets are received? For example, the existing test in active_quic_listener_test.cc doesn't seem to send any packets after receiving the CHLO. Is there a way to make it do that?

source/common/quic/envoy_quic_server_connection.cc
RyanTheOptimist: Oh, you already landed this CL! Nice!

source/common/quic/quic_filter_manager_connection_impl.cc
danzh2010: maybeApplyDelayClosePolicy() is only called in OnCanWrite() which was called at the end of ProcessUdpPacket() and various other places before this PR. But in this PR, we check `close_type_during_initialize_` at the beginning of ProcessUdpPacket() and close connection there. So OnCanWrite() won't be called at all in ProcessUdpPacket(). This check only takes effect at processing the 1st packet, `close_type_during_initialize_` will always be nullptr afterwards.

RyanTheOptimist: Ok, I see. Thanks for the explanation.

source/common/quic/quic_filter_manager_connection_impl.cc
```diff
@@ -97,6 +99,7 @@ void QuicFilterManagerConnectionImpl::close(Network::ConnectionCloseType type) {
     } else {
       delayed_close_state_ = DelayedCloseState::CloseAfterFlush;
     }
+    std::cerr << "========= danzh: delay close\n";
```
RyanTheOptimist, line 102: Presumably you're still debugging the CI failures?

danzh2010: Yes, this is the only way I could think of to debug windows.

danzh2010:
> _nod_ And presumably the send alarm being set leads to the behavior mentioned in the PR description. But it's not obvious to me that this is the case. (Though I certainly believe you that it is!) Would it be hard to have a test which actually waited to send something until multiple packets are received? For example, the existing test in active_quic_listener_test.cc doesn't seem to send any packets after receiving the CHLO. Is there a way to make it do that?

It was harder than I expected. CHLO doesn't always elicit sending packet due to packet coalescing, so there might be no behavioral change in this test with and without defer sending. But we are not testing QUICHE code, so to make the test more obvious, I used the getter of defer_send_in_response_to_packets_ to check its value in the test.

source/common/quic/envoy_quic_server_connection.cc
```diff
@@ -17,12 +17,17 @@ EnvoyQuicServerConnection::EnvoyQuicServerConnection(
...
-      QuicNetworkConnection(std::move(connection_socket)) {
-  if (Runtime::runtimeFeatureEnabled(
+      QuicNetworkConnection(std::move(connection_socket)),
+      defer_send_(Runtime::runtimeFeatureEnabled(
           "envoy.reloadable_features.quic_defer_send_in_response_to_packet")) {
+#ifndef WIN32
+  if (defer_send_) {
+    // Defer sending while processing UDP packets till the end of the current event loop to optimize
+    // UDP GSO sendmsg efficiency. But this optimization causes some test failures under Windows,
+    // and Windows doesn't support GSO, do not apply this optimization on Windows.
```
RyanTheOptimist, line 27: nit: Mind filing a github issue to track this and adding a TODO to get to the bottom of this?

RyanTheOptimist (dismissed):
LGTM modulo a TODO

source/common/quic/envoy_quic_server_connection.cc
danzh2010: done

RyanTheOptimist (approved):

PR #21370, opened 2022-05-19
tracing: remove Lightstep tracer

jmacd:
Commit Message: This PR entirely removes the Lightstep tracer. Lightstep is looking forward to the OpenTelemetry tracer integration.
Additional Description: The owner of this code left Lightstep and we have no plans to maintain this code now that OpenTelemetry is ready.
Risk Level: Low
Testing: I am not familiar with the Envoy code base, but did the best I could to remove Lightstep everywhere. I am hoping to let the CI pipeline catch any errors.
Docs Changes:
Release Notes: Remove Lightstep tracer.
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #22804, opened 2022-08-22
Partial Revert "tracing: remove Lightstep tracer (#21370)"

howardjohn:
This partially reverts commit eb521f42e760b9e5cb6ca544e5c87523f0592fa9.

See
https://github.com/envoyproxy/envoy/pull/21370#issuecomment-1210127481

Signed-off-by: John Howard <howardjohn@google.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/config/trace/v3/lightstep.proto
+++ b/api/envoy/config/trace/v3/lightstep.proto
@@ -22,6 +22,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 
 // Configuration for the LightStep tracer.
 // [#extension: envoy.tracers.lightstep]
+// [#not-implemented-hide:]
 message LightstepConfig {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.config.trace.v2.LightstepConfig";
--- a/api/envoy/config/trace/v3/lightstep.proto
+++ b/api/envoy/config/trace/v3/lightstep.proto
@@ -1,0 +1,58 @@
+syntax = "proto3";
+
+package envoy.config.trace.v3;
+
+import "envoy/config/core/v3/base.proto";
+
+import "envoy/annotations/deprecation.proto";
+import "udpa/annotations/migrate.proto";
+import "udpa/annotations/status.proto";
+import "udpa/annotations/versioning.proto";
+import "validate/validate.proto";
+
+option java_package = "io.envoyproxy.envoy.config.trace.v3";
+option java_outer_classname = "LightstepProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/config/trace/v3;tracev3";
+option (udpa.annotations.file_migrate).move_to_package =
+    "envoy.extensions.tracers.lightstep.v4alpha";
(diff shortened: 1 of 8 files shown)


htuch:
LGTM, but looks like a genuine issue with the docs build.

howardjohn:
/retest

failure is downloading bazel, I don't think its related?

howardjohn:
I am a bit unsure how to do this. If I remove it from metadata.yaml it fails with 'Did you forget to add 'envoy.tracers.lightstep''. If I add it there it fails with 'CodeChecker ERROR [extensions_registered] Metadata for unused extension found: envoy.tracers.lightstep'. Do I need to somehow use it in a NOP sort of way?

htuch:
@phlax can you advise here? We basically had an API + extension + metadata. Then we removed all 3, which was a bit aggressive, so we want to restore the API + metadata only.

phlax:
> can you advise here? We basically had an API + extension + metadata. Then we removed all 3, which was a bit aggressive, so we want to restore the API + metadata only.

iiuc this is a novel situation that we didnt plan for when checking for the correct extension setup to generate docs (i think this would have always failed - even before metadata was separated out)

probably we can just add an ignore list for the code.checker - assuming that the rest of the code is happy with this setup

api/envoy/config/trace/v3/lightstep.proto
```diff
@@ -0,0 +1,58 @@
...
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/config/trace/v3;tracev3";
+option (udpa.annotations.file_migrate).move_to_package =
+    "envoy.extensions.tracers.lightstep.v4alpha";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: LightStep tracer]
+
+// Configuration for the LightStep tracer.
+// [#extension: envoy.tracers.lightstep]
```
phlax, line 24:
```suggestion
// [#extension: envoy.tracers.lightstep]
// [#not-implemented-hide:]
```

phlax (commented):
debugging this further locally - i think the key thing is we dont want this included in docs as its no longer supported (similar i guess to the v2 protos)

the problem we then have is that, unlike the v2 protos that are not sent to protodoc, the `api/envoy/config/trace/v3` target is which now includes the lightstep.proto

the workaround i found is to make this proto/message `not-implemented-hide` which is actually accurate even its not `never-been-implemented-hide`

for some reason (not sure exactly why) this only worked for me if i rebased to current main

api/envoy/config/trace/v3/lightstep.proto
howardjohn: Thank you! A huge help

phlax:
thinking about this further - i guess my one concern is that other machinery will read this - eg validation and perhaps the envoy binary would accept the config as valid but not work

phlax (approved):
landing this (and removing api tags) on the basis of @htuch 's approval above

PR #22806, opened 2022-08-23
http: add max response headers size config

agrawroh:
### Background

Currently, there seems to be no easy way to configure the max limit on the response headers and it's hard-coded to **80 KiB** for HTTP/1.1 and **60K** for HTTP/2 & HTTP/3.

### Changes

This PR adds a new field called `max_response_headers_kb` to make this limit configurable.

---
**Commit Message:** add a new config to override the max response headers size.
**Risk Level:** Low
**Testing:** Unit Tests
**Docs Changes:** Added the description of the new proto field. 
**Release Notes:** Added

### DIFF
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -173,7 +173,7 @@ message AlternateProtocolsCacheOptions {
   repeated AlternateProtocolsCacheEntry prepopulated_entries = 4;
 }
 
-// [#next-free-field: 7]
+// [#next-free-field: 8]
 message HttpProtocolOptions {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.api.v2.core.HttpProtocolOptions";
@@ -223,6 +223,11 @@ message HttpProtocolOptions {
   // and the connection will be force-closed after the drain period. See :ref:`drain_timeout
   // <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.drain_timeout>`.
   google.protobuf.Duration max_connection_duration = 3;
+
+  // The maximum response headers size allowed from the upstream server. If not set, the default max response headers
+  // size allowed is 80 KiB. Response that exceed this limit would fail with a protocol error.
+  google.protobuf.UInt32Value max_response_headers_kb = 7
+      [(validate.rules).uint32 = {lte: 8192 gt: 0}];
 
   // The maximum number of headers. If unconfigured, the default
   // maximum number of request headers allowed is 100. Requests that exceed this limit will receive
--- a/test/common/http/http1/codec_impl_test.cc
+++ b/test/common/http/http1/codec_impl_test.cc
@@ -2191,8 +2191,8 @@ class Http1ClientConnectionImplTest : public Http1CodecTestBase {
 public:
   void initialize() {
     codec_ = std::make_unique<Http1::ClientConnectionImpl>(
-        connection_, http1CodecStats(), callbacks_, codec_settings_, max_response_headers_count_,
-        /* passing_through_proxy=*/false);
+        connection_, http1CodecStats(), callbacks_, codec_settings_, max_response_headers_kb_,
+        max_response_headers_count_, /* passing_through_proxy=*/false);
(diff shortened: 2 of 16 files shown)


api/envoy/config/core/v3/protocol.proto
```diff
@@ -224,6 +224,11 @@ message HttpProtocolOptions {
   // <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.drain_timeout>`.
   google.protobuf.Duration max_connection_duration = 3;
 
+  // The maximum response headers size allowed from the upstream server. If not set, the default max response headers
+  // size allowed is 80 KiB. Response that exceed this limit would fail with a protocol error.
```
adisuissa, line 228: As you note the default value is different for H1 (60K) and H2/H3 (80K).
IIUC this is intentional (lower limits of the total size of the headers when multiple streams are used).
@yanavlasov do you think the value should be the same for all HTTP versions or should it be different (as in the current implementation)?

api/envoy/config/core/v3/protocol.proto
```diff
@@ -224,6 +224,11 @@ message HttpProtocolOptions {
...
+  google.protobuf.UInt32Value max_response_headers_kb = 7
```
adisuissa, line 229: @htuch I'm wondering if this field should be here or in the HCM (similar to [`max_request_headers_kb`](https://github.com/envoyproxy/envoy/blob/4632a68618322c482e3b750adf0805ed9c01460a/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L439)). I tend to add it to HttpProtocolOptions as suggested, but would like to know if you think otherwise.

test/common/http/http1/codec_impl_test.cc
```diff
@@ -3438,6 +3440,24 @@ TEST_P(Http1ClientConnectionImplTest, ManyResponseHeadersRejected) {
   EXPECT_EQ(status.message(), "http/1.1 protocol error: headers count exceeds limit");
 }
 
+// Tests that the number of response headers is configurable.
+TEST_P(Http1ClientConnectionImplTest, LargeResponseHeadersConfigurable) {
+  max_response_headers_kb_ = 8192;
```
adisuissa, line 3445: I think the test should have a smaller value, just in case the underlying codec rejects larger values.

test/integration/multiplexed_upstream_integration_test.cc
```diff
@@ -534,13 +534,13 @@ TEST_P(MultiplexedUpstreamIntegrationTest, ManyResponseHeadersAccepted) {
   EXPECT_TRUE(response->complete());
 }
 
-// Tests that HTTP/2 response headers over 60 kB are rejected and result in a stream reset.
+// Tests that HTTP/2 response headers over 80KB are rejected and result in a stream reset.
```
adisuissa, line 537: Please add a test that overrides the default value in the H2/H3 case.

adisuissa (commented):
Thanks, left some minor comments.

api/envoy/config/core/v3/protocol.proto
htuch: I think it's fine here - it's arguably a property of the upstream so not HCM. That said, maybe HCM should migrate its request header sizes here for consistency.

test/integration/multiplexed_upstream_integration_test.cc
agrawroh: Added, Thanks!

test/common/http/http1/codec_impl_test.cc
agrawroh: Changed it to **4096** instead.

api/envoy/config/core/v3/protocol.proto
agrawroh: Do we need to make it lower by default or is it okay to keep it consistent and set it to **80K** by default?

adisuissa: It depends on the reasoning behind the 2 values.
@yanavlasov WDYT?

PR #22823, opened 2022-08-24
xds: Adds integration tests for xDS corner cases

abeyad:
(1) An integration test for two resources being sent over two separate
DiscoveryResponses from two separate authorities but with the same
resource name and type url.

(2) An integration test for ADS sending a DiscoveryResponse update
containing only the updated resources (which may be a subset of the
total resources sent for that type by the ADS server).


test/integration/ads_integration_test.cc
```diff
@@ -55,6 +57,56 @@ TEST_P(AdsIntegrationTest, BasicClusterInitialWarming) {
...
+
+  auto cluster_0 = buildCluster("cluster_0");
+  auto cluster_1 = buildCluster("cluster_1");
+  EXPECT_TRUE(compareDiscoveryRequest(cds_type_url, "", {}, {}, {}, true));
+  sendDiscoveryResponse<envoy::config::cluster::v3::Cluster>(cds_type_url, {cluster_0, cluster_1},
+                                                             {cluster_0, cluster_1}, {}, "1");
+  test_server_->waitForGaugeEq("cluster_manager.warming_clusters", 2);
+  EXPECT_TRUE(compareDiscoveryRequest(eds_type_url, "", {cluster_0.name(), cluster_1.name()},
+                                      {cluster_0.name(), cluster_1.name()}, {}));
+  auto cluster_la_0 = buildClusterLoadAssignment(cluster_0.name());
```
adisuissa, line 75: nit: `cluster_la` resembles a cluster name. I suggest changing to `cla` or `load_assignment`.

test/integration/ads_integration_test.cc
```diff
@@ -55,6 +57,56 @@ TEST_P(AdsIntegrationTest, BasicClusterInitialWarming) {
...
+        "", downstreamProtocol(), version_);
+    EXPECT_TRUE(response->complete());
+    EXPECT_EQ("200", response->headers().getStatusValue());
+    Json::ObjectSharedPtr loader = TestEnvironment::jsonLoadFromString(response->body());
+    envoy::admin::v3::ConfigDump config_dump;
+    TestUtility::loadFromJson(loader->asJsonString(), config_dump);
+    return config_dump.configs_size() == 2;
+  };
+  EXPECT_TRUE(timeSystem().waitFor(lock, absl::Condition(&load_assignment_updated),
+                                   TestUtility::DefaultTimeout));
```
adisuissa, line 107: This can probably be simplified by validating a stat like [`membership_healthy`](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html?highlight=membership#general) on both clusters.

test/integration/xds_integration_test.cc
```diff
@@ -796,5 +801,209 @@ TEST_P(LdsStsIntegrationTest, TcpListenerRemoveFilterChainCalledAfterListenerIsR
...
+      // SDS for the first cluster.
+      initXdsStream(getXdsUpstream1(), xds_connection_1_, xds_stream_1_);
+      EXPECT_TRUE(compareSotwDiscoveryRequest(
+          xds_stream_1_, /*expected_type_url=*/Config::TypeUrl::get().Secret,
+          /*expected_version=*/"",
+          /*expected_resource_names=*/{cert_name}, /*expect_node=*/true));
+      auto sds_resource = getClientSecret(cert_name);
+      sendSotwDiscoveryResponse<envoy::extensions::transport_sockets::tls::v3::Secret>(
+          xds_stream_1_, Config::TypeUrl::get().Secret, {sds_resource}, "1");
+    }
```
adisuissa, line 974: Sorry, I think I've mislead you when we discussed this IRL.
If there are 2 different xDS clusters, each with a different backing server, and each connection to an xDS cluster registers for a specific resource (1:1 mapping), then the test can assume what will be the requested resource (the above is not true when a single xDS cluster is used, and the order of requested resources could be different).

test/integration/xds_integration_test.cc
```diff
@@ -796,5 +801,209 @@ TEST_P(LdsStsIntegrationTest, TcpListenerRemoveFilterChainCalledAfterListenerIsR
...
+
+  initialize();
+
+  // Wait until the discovery responses have been processed.
+  test_server_->waitForCounterGe(
+      "cluster.cluster_0.client_ssl_socket_factory.ssl_context_update_by_sds", 1);
+  test_server_->waitForCounterGe(
+      "cluster.cluster_1.client_ssl_socket_factory.ssl_context_update_by_sds", 1);
+
+  auto config_dump = getSecretsFromConfigDump();
```
adisuissa, line 996: Could this be achieved by looking at counters or gauges instead of config-dump (decoupling xDS tests from config-dump)?

adisuissa (commented):
Thanks!
Left a few comments suggesting some simplifications.

test/integration/xds_integration_test.cc
abeyad: Chatted offline with Adi, leaving this as is

test/integration/ads_integration_test.cc
abeyad: Thanks for the suggestion, done.

test/integration/base_integration_test.h
```diff
@@ -208,10 +208,19 @@ class BaseIntegrationTest : protected Logger::Loggable<Logger::Id::testing> {
       const std::vector<std::string>& expected_resource_names, bool expect_node = false,
       const Protobuf::int32 expected_error_code = Grpc::Status::WellKnownGrpcStatus::Ok,
       const std::string& expected_error_message = "");
+  // For SotW tests, provides the same functionality as the function above, except it also takes in
+  // the FakeStream instead of using xds_stream_. Continue to use the above function for SotW tests
+  // unless the xDS test is not using xds_stream_.
+  AssertionResult compareSotwDiscoveryRequest(
```
alyssawilk, line 214: could we just extend the function above to take stream as a last argument and default it to xds_stream_ ?  Ditto below

alyssawilk (commented):
Largely leaving the test review to Adi but one question/nit on framework changes

test/integration/base_integration_test.h
abeyad: Good point.  Non-static members can't be used as default arguments, so I can't directly assign the default to xds_stream_, but I made the default nullptr and if it's nullptr, use xds_stream_.

test/integration/ads_integration_test.cc
```diff
@@ -55,6 +56,42 @@ TEST_P(AdsIntegrationTest, BasicClusterInitialWarming) {
...
+  test_server_->waitForGaugeEq("cluster_manager.warming_clusters", 2);
+  EXPECT_TRUE(compareDiscoveryRequest(eds_type_url, "", {cluster_0.name(), cluster_1.name()},
+                                      {cluster_0.name(), cluster_1.name()}, {}));
+  auto cla_0 = buildClusterLoadAssignment(cluster_0.name());
+  auto cla_1 = buildClusterLoadAssignment(cluster_1.name());
+  sendDiscoveryResponse<envoy::config::endpoint::v3::ClusterLoadAssignment>(
+      eds_type_url, {cla_0, cla_1}, {cla_0, cla_1}, {}, "1");
+
+  test_server_->waitForGaugeEq("cluster_manager.warming_clusters", 0);
+  test_server_->waitForGaugeGe("cluster_manager.active_clusters", 2);
```
adisuissa, line 80: should probably be 4 (including the dummy_cluster and ads_cluster)

test/integration/ads_integration_test.cc
```diff
@@ -55,6 +56,42 @@ TEST_P(AdsIntegrationTest, BasicClusterInitialWarming) {
...
+
+  // Send an update for one of the ClusterLoadAssignments only.
+  cla_0.mutable_endpoints(0)->mutable_lb_endpoints(0)->mutable_load_balancing_weight()->set_value(
+      2);
+  sendDiscoveryResponse<envoy::config::endpoint::v3::ClusterLoadAssignment>(eds_type_url, {cla_0},
+                                                                            {cla_0}, {}, "2");
+
+  // Verify that getting an update for only one of the ClusterLoadAssignment resources does not
+  // delete the other. We use cluster membership health as a proxy for this.
+  test_server_->waitForCounterEq("cluster_manager.cluster_updated", 1);
```
adisuissa, line 90: IIUC the [`cluster_updated` docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats.html?highlight=cluster_updated#general) correctly, this should be updated during the CDS and not EDS.
Will `update_success` work here?

test/integration/xds_integration_test.cc
```diff
@@ -796,5 +801,213 @@ TEST_P(LdsStsIntegrationTest, TcpListenerRemoveFilterChainCalledAfterListenerIsR
...
+  FakeHttpConnectionPtr xds_connection_1_;
+  FakeStreamPtr xds_stream_1_;
+  FakeHttpConnectionPtr xds_connection_2_;
+  FakeStreamPtr xds_stream_2_;
+};
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, XdsSotwMultipleAuthoritiesTest,
+                         GRPC_CLIENT_INTEGRATION_PARAMS);
+
+TEST_P(XdsSotwMultipleAuthoritiesTest, SameResourceNameAndTypeFromMultipleAuthorities) {
```
adisuissa, line 960: Please add a comment describing the test.

test/integration/ads_integration_test.cc
```diff
@@ -55,6 +56,42 @@ TEST_P(AdsIntegrationTest, BasicClusterInitialWarming) {
   test_server_->waitForGaugeGe("cluster_manager.active_clusters", 2);
 }
 
+TEST_P(AdsIntegrationTest, UpdateToSubsetOfResources) {
```
adisuissa, line 59: Please add a comment describing the test.

adisuissa (commented):
Overall looks good, left minor comments.
Thanks!

test/integration/BUILD
```diff
@@ -62,6 +62,7 @@ envoy_cc_test(
         "//test/test_common:network_utility_lib",
         "//test/test_common:resources_lib",
         "//test/test_common:utility_lib",
+        "@envoy_api//envoy/admin/v3:pkg_cc_proto",
```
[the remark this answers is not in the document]
abeyad: done

test/integration/ads_integration_test.cc
```diff
@@ -1,3 +1,4 @@
+#include "envoy/admin/v3/config_dump.pb.h"
```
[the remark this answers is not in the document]
abeyad: done

test/integration/ads_integration_test.cc
abeyad: done

test/integration/ads_integration_test.cc
abeyad: done

test/integration/ads_integration_test.cc
abeyad: Changed to use update_success

test/integration/xds_integration_test.cc
abeyad: done

abeyad (commented):
Thanks Adi! Addressed the feedback, PTAL.

adisuissa (approved):
LGTM, thanks!

Issue #18209, opened 2021-09-22
generic proxy support


wbpcode:
Now, Envoy already supports  proxy traffic of many different protocol. But if we want to support a new protocol, we alway **need to implement a complete L4 filter and all related L7 filters. But in fact, many protocols have similarities. So, may be we can try to implement a generic proxy and then, by extending different encoders and decoders to extend different protocol support. And all these protocol can reuse the L7 filters and route of the generic proxy.**

Perhaps, the generic proxy cannot achieve the best performance, but it can reduce the complexity and workload of extending protocol support in Envoy. In this way, everyone can simply add their own private protocol support to Envoy and the scope of use of Envoy should be further expanded.

Design docs: https://docs.google.com/document/d/1y_ALyjGp_H1_TRrrRxZxnqENISqeZI6eLn71FAFsZPk/edit?usp=sharing

**meta protocol API**: #18823 
**meta protocol core framework**: #18944 

other related issue: 
Generic Access Log: #13085
Generic Routing & Rds: #9656

wbpcode:
![generic_proxy](https://user-images.githubusercontent.com/12389633/134347472-834f1518-533c-4a47-996b-40faf2284d45.png)

ref MOSN, brpc and aeraki metadata protocol etc.

wbpcode:
> It would be interesting to try making these binary encoders/decoders optionally asynchronous, so that they could be accelerated in HW (e.g. FPGA).

Sounds great. Although my current design and implementation do not take this into consideration, I can give it a try. 🤔

PR #22826, opened 2022-08-24
extension: added generic proxy network filter

wbpcode:
Commit Message: extension: added generic proxy framework
Additional Description:

This is very long story. Please check #18209 for more detailed discussions and background. 
And there is the design proposol https://docs.google.com/document/d/1y_ALyjGp_H1_TRrrRxZxnqENISqeZI6eLn71FAFsZPk/edit#heading=h.a0godkb0g8bc
for why we want the extension.

Due to my personal reasons, I consistently did not finish the final testing of this extension, so it was delayed until now.
I picked this up finally and will complete it in this time.

It's sad the @rojkov cannot sponsor it. So I will self sponsor this extension.

Risk Level: Low. New extension.
Testing: Add.
Docs Changes: Wait.
Release Notes: Wait.
Platform Specific Features: Wait.


wbpcode:
- [ ] integration test.
- [ ] docs & release note.

CODEOWNERS
```diff
@@ -265,6 +265,8 @@ extensions/filters/http/oauth2 @derekargueta @snowp
 /*/extensions/filters/http/header_to_metadata @zuercher @JuniorHsu
 # zookeeper
 /*/extensions/filters/network/zookeeper_proxy @snowp @JuniorHsu
+# Mata protocol filter
```
lizan, line 268:
```suggestion
# Meta protocol filter
```

lizan (commented):
/lgtm api

wbpcode:
cc @mattklein123 Hi, this PR is ready for review. Could you take a look when you have free time? Thanks.

wbpcode:
Because it's long time after last dicussion. Here is some simple context about the extention.

### Why we need this extensions?

Third-party RPC or private protocol support is a common requirement for users of Envoy/Mesh. However it's very complex to developed a new Envoy extension to proxy a third-party RPC or private protocol for now.
This generic proxy is L4 network filter which will provides common route, L7 filter chain, stream management, etc. The developers could add new protocol support to the Envoy by just implementing the core codec of this specific protocol. This reduce much of the complexity of new protocol support.

### About this extension

The design of this extension is come from many others contribution. The initial idea of implement generic proxy in the Envoy is come from @zhaohuabing (aeraki-mesh). The xds matcher based route table is come from @htuch. The async codec API is come from @rojkov. And now, we finally have a implementation for the Envoy community. (The aeraki-mesh will keep their extension because by this way they can add new features more quickly and freely. So we renamed this extension to simple `generic_proxy` to distinguish this extension from the aeraki-mesh's extension and avoid the name conflict in the aeraki-mesh's building.)

### Next step of this extension?

1. Because the async codec API and xds matcher based route table, implementation in this PR is almost complete different with our internal implementation or aeraki-mesh's implementation. So more tests are necessary to ensure everything is work. (The current coverage rate is 95%.)
2. Dubbo codec support. Dubbo proxy has very simple stream model and could be easily to migrate it to this new extension. After that, we can avoid the cost of maintaining the dubbo proxy. Of course, we will have a long, long time to gradually deprecate the old dubbo proxy.
3. More features, such rds support, observability, route level config, etc.

mattklein123:
@wbpcode this is too much code for me to reasonably review. I'm not sure the best way of moving this forward. I think rewriting Dubbo to use this would give me a lot more confidence. How would you suggest incrementally moving this forward?

wbpcode:
@mattklein123 As I originally planned, after this PR is merged, I would first move dubbo decoder and encoder to extensions/common and then implement dubbo codec for generic proxy based on these encoder/decoder.

wbpcode:
Yeah, this is a super big PR that hard to review. 🤣 But this extension still be marked as WIP, so we still have chance to correct it in the future if there are something wrong.

IMO, the core should be the message API and codec API, you can check these code in the `genetic_proxy/interface`.

wbpcode:
This extension was moved to contrib as our offline discussion. @mattklein123

mattklein123 (approved):
Thanks. I'm OK merging this into contrib. Once Dubbo has been rewritten to use this and vetted in production I think we can deprecate/remove the existing Dubbo filter and move this all into core. Thank you!

PR #22827, opened 2022-08-24
tls: expose some private members in Defaultcertvalidator

danzh2010:
Commit Message: these interfaces are likely to be used by subclasses.

Risk Level: low, no behavioral change
Testing: N/A
Docs Changes: N/A
Release Notes: N/A


source/extensions/transport_sockets/tls/cert_validator/default_validator.h
```diff
@@ -107,18 +108,21 @@ class DefaultCertValidator : public CertValidator, Logger::Loggable<Logger::Id::
   static bool matchSubjectAltName(X509* cert,
                                   const std::vector<SanMatcherPtr>& subject_alt_name_matchers);
 
+protected:
+  // Increment failure stats and update validation state in ssl_extended_info if not nullptr.
+  void onVerifyError(Ssl::SslExtendedSocketInfo* ssl_extended_info, absl::string_view error);
+
+  bool allow_untrusted_certificate_{false};
```
RyanTheOptimist, line 115: The style guide requires data members to be private, so we probably want an accessor here instead of making the member itself protected.

RyanTheOptimist (commented):
Is the idea that we will be subclassing the default verifier, and we need to expose onVerifyError for a subclass to call?

ggreenway:
Can we pause to discuss and think about if and how it would be appropriate to have subclasses of the default verifier? It may make more sense to have some helper functions/classes and have a completely separate verifiers that use those helpers. Code re-use through inheritance in c++ is often non-ideal.

danzh2010:
> Can we pause to discuss and think about if and how it would be appropriate to have subclasses of the default verifier? It may make more sense to have some helper functions/classes and have a completely separate verifiers that use those helpers. Code re-use through inheritance in c++ is often non-ideal.

That's a good point. After all only error stats and allow_untrusted_certificate_ is reused. So might as well just making the new verifier handle those by itself.

PR #18997, opened 2021-11-12
Extend StatefulHeaderFormatter to allow forwarding HTTP1 reason phrase

syhpoon:
Commit Message: http: extend StatefulHeaderFormatter to allow forwarding HTTP1 reason phrase
Additional Description: In some cases the upstream can produce a non-default reason phrase when using HTTP1. For example AWS S3 can return `503 Slow Down` in certain cases. StatefuleHeaderFormatter can now take an optional flag to enable forwarding non-default HTTP1 reason phrases.
Risk Level: Medium
Testing: unit testing, integration testing
Docs Changes: release notes
Release Notes: http: added support for forwarding HTTP1 reason phrase
Platform Specific Features: N/A

PR #22847, opened 2022-08-25
http.lua: allow setting HTTP1 reason phrase

syhpoon:
:wave: 

Note:
This is a PoC PR to ensure the validity of the approach and the feasibility of the eventual upstream merge.
All the missing parts (docs, tests) will be added later if maintainers find the approach acceptable. 

This PR is somewhat complementary to https://github.com/envoyproxy/envoy/pull/18997/files and uses some machinery established in it.
Essentially, we'd need to be able to customize HTTP1 reason phrase based on the responses from `ext_authz` filter. 
Currently it's possible to override the code but the reason phrase (being HTTP1-specific) is set to the default value.

The approach in this PR adds a lua function to override reason phrase. It is done by injecting a stateful header formatter into the response headers map if it hasn't been set yet and setting the reason phrase, which will be used by the encoder.

### DIFF
--- a/docs/root/configuration/http/http_filters/lua_filter.rst
+++ b/docs/root/configuration/http/http_filters/lua_filter.rst
@@ -629,8 +629,9 @@ setHttp1ReasonPhrase()
 
   headers:setHttp1ReasonPhrase(reasonPhrase)
 
-Sets a custom HTTP1 response reason phrase. This call is *only valid in the response flow*.
-*reasonPhrase* is a string that supplies the reason phrase value.
+Sets a custom HTTP/1 response reason phrase. This call is *only valid in the response flow*.
+*reasonPhrase* is a string that supplies the reason phrase value. Additionally this call only
+effects HTTP/1 connections. It will have no effect if the client is HTTP/2 or HTTP/3.
 
 .. _config_http_filters_lua_buffer_wrapper:
 
--- a/docs/root/configuration/http/http_filters/lua_filter.rst
+++ b/docs/root/configuration/http/http_filters/lua_filter.rst
@@ -622,6 +622,16 @@ replace()
 Replaces a header. *key* is a string that supplies the header key. *value* is a string that supplies
 the header value. If the header does not exist, it is added as per the *add()* function.
 
+setHttp1ReasonPhrase()
+^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: lua
+
+  headers:setHttp1ReasonPhrase(reasonPhrase)
+
+Sets a custom HTTP1 response reason phrase. This call is *only valid in the response flow*.
+*reasonPhrase* is a string that supplies the reason phrase value.
+
 .. _config_http_filters_lua_buffer_wrapper:
 
 Buffer API
--- a/source/extensions/filters/http/lua/wrappers.cc
+++ b/source/extensions/filters/http/lua/wrappers.cc
@@ -1,8 +1,10 @@
 #include "source/extensions/filters/http/lua/wrappers.h"
 
+#include "source/common/http/header_map_impl.h"
 #include "source/common/http/header_utility.h"
 #include "source/common/http/utility.h"
 #include "source/extensions/filters/common/lua/wrappers.h"
+#include "source/extensions/http/header_formatters/preserve_case/preserve_case_formatter.h"
 
 namespace Envoy {
 namespace Extensions {
@@ -11,10 +13,11 @@ namespace Lua {
 
 HeaderMapIterator::HeaderMapIterator(HeaderMapWrapper& parent) : parent_(parent) {
   entries_.reserve(parent_.headers_.size());
-  parent_.headers_.iterate([this](const Http::HeaderEntry& header) -> Http::HeaderMap::Iterate {
-    entries_.push_back(&header);
-    return Http::HeaderMap::Iterate::Continue;
-  });
+  parent_.headers_.iterate(
+      [this](const ::Envoy::Http::HeaderEntry& header) -> ::Envoy::Http::HeaderMap::Iterate {
+        entries_.push_back(&header);
+        return ::Envoy::Http::HeaderMap::Iterate::Continue;
+      });
 }
 
 int HeaderMapIterator::luaPairsIterator(lua_State* state) {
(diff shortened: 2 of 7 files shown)


wbpcode:
Hmm, this make sense to me. cc @mattklein123

source/extensions/filters/http/lua/wrappers_reason_phrase.cc
```diff
@@ -0,0 +1,43 @@
...
+#include "source/extensions/filters/common/lua/wrappers.h"
+#include "source/extensions/http/header_formatters/preserve_case/preserve_case_formatter.h"
+#include "envoy/http/header_formatter.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Lua {
+
+int HeaderMapWrapper::luaSetHttp1ReasonPhrase(lua_State* state) {
```
mattklein123, line 15: nit: can you just add this to the other file? Not sure this needs a new file when it's part of the header map wrapper?

mattklein123 (changes requested):
In general this change makes sense to me also. Needs release notes, docs, tests, etc. Thank you!

source/extensions/filters/http/lua/wrappers_reason_phrase.cc
syhpoon: Yup, will do! The separate file was mostly to reduce upstream merge conflicts if we had to maintain this feature in our own branch.

syhpoon:
@wbpcode @mattklein123 should be ready for review. Thanks!

source/extensions/filters/http/lua/wrappers.cc
```diff
@@ -118,8 +124,38 @@ void HeaderMapWrapper::checkModifiable(lua_State* state) {
   }
 }
 
+int HeaderMapWrapper::luaSetHttp1ReasonPhrase(lua_State* state) {
+  checkModifiable(state);
+
+  const char* phrase = luaL_checkstring(state, 2);
```
wbpcode, line 130: nit: use `luaL_checklstring` here

source/extensions/filters/http/lua/wrappers.cc
```diff
@@ -118,8 +124,38 @@ void HeaderMapWrapper::checkModifiable(lua_State* state) {
...
+    using Http::HeaderFormatters::PreserveCase::PreserveCaseHeaderFormatter;
+
+    try {
+      ResponseHeaderMapImpl& map = dynamic_cast<ResponseHeaderMapImpl&>(headers_);
+      std::unique_ptr<StatefulHeaderKeyFormatter> fmt =
+          std::make_unique<PreserveCaseHeaderFormatter>(true, PreserveCaseFormatterConfig::DEFAULT);
+      fmt->setReasonPhrase(absl::string_view(std::string(phrase)));
+      map.setFormatter(std::move(fmt));
+    } catch (const std::bad_cast&) {
+    }
```
wbpcode, line 148: May be just cast this to `ResponseHeaderMapImpl*` to avoid the `try catch`.

wbpcode: And I am not sure, can we just add a method to the `ResponseHeaderMap` to set reason phrase? cc @mattklein123

wbpcode (changes requested):
Thanks for you contribution. And some comments were added.

source/extensions/filters/http/lua/wrappers.cc
syhpoon: It's `Http::HeaderMap` type that's being used in the `HeaderMapWrapper` class here, so unless I add the setter on it I'd still need to cast, I think (c++ is not a tool I use often, so I may be wrong).

source/extensions/filters/http/lua/wrappers.cc
syhpoon: done!

source/extensions/filters/http/lua/wrappers.cc
syhpoon: Switched to direct pointer cast for now.

syhpoon:
@wbpcode should be ready for the follow-up.

source/extensions/filters/http/lua/wrappers.cc
```diff
@@ -118,8 +124,38 @@ void HeaderMapWrapper::checkModifiable(lua_State* state) {
...
+        PreserveCaseFormatterConfig;
+    using Envoy::Http::ResponseHeaderMapImpl;
+    using Envoy::Http::StatefulHeaderKeyFormatter;
+    using Http::HeaderFormatters::PreserveCase::PreserveCaseHeaderFormatter;
+
+    ResponseHeaderMapImpl* map = dynamic_cast<ResponseHeaderMapImpl*>(&headers_);
+    if (map) {
+      std::unique_ptr<StatefulHeaderKeyFormatter> fmt =
+          std::make_unique<PreserveCaseHeaderFormatter>(true, PreserveCaseFormatterConfig::DEFAULT);
+      fmt->setReasonPhrase(absl::string_view(std::string(phrase)));
```
wbpcode, line 146: May be `setReasonPhrase(absl::string_view(phrase, input_size))`

source/extensions/filters/http/lua/wrappers.cc
```diff
@@ -118,8 +124,38 @@ void HeaderMapWrapper::checkModifiable(lua_State* state) {
...
+      map->setFormatter(std::move(fmt));
+    }
+  } else {
+    formatter->setReasonPhrase(absl::string_view(std::string(phrase, input_size)));
```
wbpcode, line 150: ditto.

wbpcode (changes requested):
Thanks. It's looks good overall. Some minor comments are added. 😄

syhpoon:
Thanks for your feedback, @wbpcode !
Hopefully now it should be all addressed.

wbpcode (dismissed):
Thanks.

wbpcode:
Although the formatter will result in an additional copy for every header name. But it's nothing to do with this PR.

wbpcode:
@mattklein123 for final review.

docs/root/configuration/http/http_filters/lua_filter.rst
```diff
@@ -622,6 +622,16 @@ replace()
...
 
+setHttp1ReasonPhrase()
+^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: lua
+
+  headers:setHttp1ReasonPhrase(reasonPhrase)
+
+Sets a custom HTTP1 response reason phrase. This call is *only valid in the response flow*.
+*reasonPhrase* is a string that supplies the reason phrase value.
```
mattklein123, line 633:
```suggestion
Sets a custom HTTP/1 response reason phrase. This call is *only valid in the response flow*.
*reasonPhrase* is a string that supplies the reason phrase value. Additionally this call only
effects HTTP/1 connections. It will have no effect if the client is HTTP/2 or HTTP/3.
```

source/extensions/filters/http/lua/wrappers.cc
```diff
@@ -118,8 +124,38 @@ void HeaderMapWrapper::checkModifiable(lua_State* state) {
...
+
+  if (!formatter.has_value()) {
+    using envoy::extensions::http::header_formatters::preserve_case::v3::
+        PreserveCaseFormatterConfig;
+    using Envoy::Http::ResponseHeaderMapImpl;
+    using Envoy::Http::StatefulHeaderKeyFormatter;
+    using Http::HeaderFormatters::PreserveCase::PreserveCaseHeaderFormatter;
+
+    ResponseHeaderMapImpl* map = dynamic_cast<ResponseHeaderMapImpl*>(&headers_);
+    if (map) {
```
mattklein123, line 143: This dynamic cast is necessary because this could be called on request headers also? Can you add a comment? It would be nice to avoid the cast but probably not worth the effort to track request/response some other way.

mattklein123 (changes requested):
Thanks LGTM with some small comments.

/wait

source/extensions/filters/http/lua/wrappers.cc
syhpoon: Yes, pretty much. Also since I'm not that familiar with Envoy code yet and I wasn't sure there could be other (current or future) `headers_` implementations.

mattklein123 (approved):
Thanks!

Issue #4552, opened 2018-09-27
Migrate to python3


htuch:
What do folks think about moving to Python 3 across the board? We're going to need this at Google in the short-medium term. I'm not an expert on what the best practices are here, how we can best maintain Python 2/3 mixed compatibility etc. @danielhochman do you have any thoughts?

mattklein123:
:popcorn:

cclauss:
Todos:
* [ ] Remove Python 3 syntax errors and undefined names (#5599)
* [ ] Modify __.circleci/config.yml__ to start testing in Python 3 in allow_failures mode
* [ ] Ensure requirements are at versions that support both Python 2 and Python 3
* [ ] Test, test, test
* [ ] Remove allow_failures mode from CircleCI
* [ ] Modify README.md, Docs, etc. where required.

keith:
This has come up again with some bazel updates, more context https://github.com/envoyproxy/envoy/pull/7337

kylebevans:
An Ubuntu 18.04 system with only python3 threw this error while building:

```
ERROR: /home/kylebe/envoy/source/extensions/filters/network/kafka/BUILD:163:1: Executing genrule //source/extensions/filters/network/kafka:kafka_response_generated_source failed (Exit 1) bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Error occurred while attempting to use the default Python toolchain (@rules_python//python:autodetecting_toolchain).
According to '/usr/bin/python -V', version is 'Python 3.6.7', but we need version 2. PATH is:

/home/kylebe/envoy/clang+llvm-9.0.0-x86_64-linux-gnu-ubuntu-18.04/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin

Please ensure an interpreter with version 2 is available on this platform as 'python2' or 'python', or else register an appropriate Python toolchain as per the documentation for py_runtime_pair (https://github.com/bazelbuild/rules_python/blob/master/docs/python.md#py_runtime_pair).

[...]

PR #22865, opened 2022-08-26
Execute hot-restarter using python3

GhostLyrics:
Commit Message:
Additional Description:
Risk Level:
Testing: None (research how envoy is currently running on modern systems reveals people are using hacks like the package [python-is-python3](https://askubuntu.com/questions/1296790/python-is-python3-package-in-ubuntu-20-04-what-is-it-and-what-does-it-actually))
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

As brought up in https://github.com/envoyproxy/envoy/issues/4552#issuecomment-1224654557 I've submitted a PR. Pinging @phlax as suggested.

Note that this MR *will* break envoy on Platforms that are still using Python2 and don't offer Python3 at all. 

I've done a quick test on what I consider the most up-2-date linux (archlinux) to see what the result was when calling `/usr/bin/env python` and `/usr/bin/env python3`. There, they point to the same file. This PR is primarily to correct the behaviour for Debian-based platforms. If you consider this PR to be wrong, you likely want to change your Debian package to require `python-is-python3` for Bullseye (and up?).

### DIFF
--- a/restarter/hot-restarter.py
+++ b/restarter/hot-restarter.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 from __future__ import print_function
 
 import os


mattklein123:
Seems fine. Can you fix DCO please? Thank you.

/wait

phlax (approved):
lgtm, thanks @GhostLyrics - altho im not totally clear if there is anyone currently loading this with python2 - either way i think its reasonable now to just enforce python3

we can probably remove/update some other parts of the file - like the `__future__` statement beneath - but i think this is fine to land as is

mattklein123 (approved):

Issue #21535, opened 2022-06-01
Allow filtering of headers sent in CheckRequest to gRPC ext authz


petedmarsh:
*Title*: Allow filtering of headers sent in CheckRequest to gRPC ext authz

*Description*:

The HTTP ext authz allows for the filtering of headers from the request sent to the ext authz service (`allowed_headers`: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-authorizationrequest), but the gRPC version does not.

In my case a lot of non-required headers are being passed along to my ext authz, I would prefer if they were not. I could of course move to straight HTTP but I don't really want to do that either.

rulex123:
👋 @KBaichoo I have a [patch](https://github.com/envoyproxy/envoy/pull/22877) that addresses this issue. Any chance we can get somebody to help with a first pass of code review? Thanks!

PR #22877, opened 2022-08-29
Support allowlisting request headers for both gRPC and HTTP authorization servers

rulex123:
Commit Message: add support for allowlisting request headers included in the check request to the authorization server, be it HTTP or gRPC (currently, this is supported for HTTP only). This patch deprecates the current `allowed_headers` config field (on the `authorization_request` message) and introduces a new one on the `ext_authz` message. To note is that the default behaviour (i.e. no config specified) differs based on the type of authorization server (this is to maintain backwards compatibility).

Additional Description: n/a

Risk Level: low

Testing: updated unit and integration tests

Docs Changes: updated API protos

Release Notes: updated release notes

[Optional Fixes #Issue]: https://github.com/envoyproxy/envoy/issues/21535

### DIFF
--- a/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
+++ b/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
@@ -25,7 +25,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // External Authorization :ref:`configuration overview <config_http_filters_ext_authz>`.
 // [#extension: envoy.filters.http.ext_authz]
 
-// [#next-free-field: 17]
+// [#next-free-field: 18]
 message ExtAuthz {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.config.filter.http.ext_authz.v2.ExtAuthz";
@@ -153,6 +153,22 @@ message ExtAuthz {
   // :ref:`destination<envoy_v3_api_field_service.auth.v3.AttributeContext.destination>`.
   // The labels will be read from :ref:`metadata<envoy_v3_api_msg_config.core.v3.Node>` with the specified key.
   string bootstrap_metadata_labels_key = 15;
+
+  // Check request to authorization server will include the client request headers that have a correspondent match
+  // in the :ref:`list <envoy_api_msg_type.matcher.ListStringMatcher>`. Note that in addition to the
+  // user's supplied matchers:
+  //
+  // 1. In addition to the the user's supplied matchers, ``Host``, ``Method``, ``Path``,
+  //   ``Content-Length``, and ``Authorization`` are **automatically included** to the list.
+  //
+  // 2. For requestst to an HTTP authorization server: *Content-Length* will be set to 0 and the request to the
+  // authorization server will not have a message body. However, the check request can include the buffered 
+  // client request body (controlled by :ref:`with_request_body
+  // <envoy_api_field_config.filter.http.ext_authz.v2.ExtAuthz.with_request_body>` setting),
+  // consequently the value of *Content-Length* of the authorization request reflects the size of
+  // its payload size.
+  //
+  type.matcher.v3.ListStringMatcher allowed_headers = 17;
 }
 
 // Configuration for buffering the request data.
@@ -246,7 +262,9 @@ message AuthorizationRequest {
   //   <envoy_v3_api_field_extensions.filters.http.ext_authz.v3.ExtAuthz.with_request_body>`
   //   setting) hence the value of its ``Content-Length`` reflects the size of its payload size.
   //
-  type.matcher.v3.ListStringMatcher allowed_headers = 1;
+  //   This field has been deprecated in favor of :ref:`allowed_headers
+  //   <envoy_v3_api_field_extensions.filters.http.ext_authz.v3.ExtAuthz.allowed_headers>`.
+  type.matcher.v3.ListStringMatcher allowed_headers = 1 [deprecated = true];
 
   // Sets a list of headers that will be included to the request to authorization service. Note that
   // client request of the same key will be overridden.
--- a/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
+++ b/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
@@ -162,7 +162,7 @@ message ExtAuthz {
   // 1. In addition to the the user's supplied matchers, ``Host``, ``Method``, ``Path``,
   //   ``Content-Length``, and ``Authorization`` are **automatically included** to the list.
   //
-  // 2. For requestst to an HTTP authorization server: *Content-Length* will be set to 0 and the request to the
+  // 2. For requests to an HTTP authorization server: *Content-Length* will be set to 0 and the request to the
   // authorization server will not have a message body. However, the check request can include the buffered
   // client request body (controlled by :ref:`with_request_body
   // <envoy_api_field_config.filter.http.ext_authz.v2.ExtAuthz.with_request_body>` setting),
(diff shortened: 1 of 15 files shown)


rulex123:
CI fails with some unrelated error (HTTP 404 on downloading some gz file), not sure what to do about that 🤷‍♀️

api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
```diff
@@ -153,6 +154,24 @@ message ExtAuthz {
   // :ref:`destination<envoy_v3_api_field_service.auth.v3.AttributeContext.destination>`.
   // The labels will be read from :ref:`metadata<envoy_v3_api_msg_config.core.v3.Node>` with the specified key.
   string bootstrap_metadata_labels_key = 15;
+
+  // Check request to authorization server will include the client request headers that have a correspondent match
```
ggreenway, line 158: I don't think we should change the existing behavior. I think if this new message isn't set, then the previous behavior of sending all headers should apply.

ggreenway (changes requested):
This change makes a lot of sense, but I think we should stay backwards-compatible with the previous behavior when the new config option isn't set, even though that makes it more complicated to describe document the behavior.

/wait

rulex123:
OK, I can rework the code to preserve the current behaviour.

rulex123:
@ggreenway Before I change the code, I'd like to make sure we agree on the new API. To recap, the current default behaviour is that all client request headers are relayed to a gRPC auth server, but only a handful (i.e. Authorization, Host, Path and Method) to an HTTP auth server. For HTTP, one can add to these default headers by configuring `allowed_headers` on the authorization request message.

Moving forward, we want to support configuring which headers are relayed to the auth server (no matter the protocol): the (new) default behaviour will be to relay all client request headers, and a newly introduced field on the ext_authz message - called `headers_to_relay` - is used to control this behaviour. In addition, the pre-existing `allowed_headers` is deprecated. 
 
One implication of these changes is that HTTP auth servers that rely on the (previous) default behaviour, will have to configure the new `headers_to_relay` field if they wish to mimic it. Otherwise, with these changes they will start receiving all client request headers. 

WDYT?

ggreenway:
If you want to change existing behavior, you're going to need to go through a deprecation cycle of the existing behavior, which roughly looks like adding a warning of using the existing behavior, wait a few releases, flip to the new behavior by default but allow a runtime override, wait a few releases, then delete code for the old behavior.

It will be simpler/easier to just maintain the current behavior, even though it doesn't really make sense to have different defaults between grpc and http modes.

rulex123:
@ggreenway I reworked the code to preserve backwards compatibility, as you suggested. I _think_ this patch might be ready for a first pass of code review. Could you take a look, or loop in someone that can help with that? Thanks!

KBaichoo:
/assign @pradeepcrao 

Pradeep can you take a look as extension owner as Greg is OOO (sick). Thank you!

changelogs/current.yaml
```diff
@@ -34,4 +34,8 @@ new_features:
   change: |
     allowing the dynamic forward proxy cluster to :ref:`allow_coalesced_connections <envoy_v3_api_field_extensions.clusters.dynamic_forward_proxy.v3.ClusterConfig.allow_coalesced_connections>`  for HTTP/2 and HTTP/3 connections.
 
+- area: ext_authz
+  change: |
+    added support to allowlist headers included in the check request to gRPC authorization server (previously only available for HTTP authorization server).
+
```
pradeepcrao, line 40: Please also mention what is deprecated as part of this change.

phlax:
@pradeepcrao would you mind giving this another pass

pradeepcrao:
Will take a look tomorrow. Sorry for the delay.

source/extensions/filters/common/ext_authz/check_request_utils.cc
```diff
@@ -119,15 +134,28 @@ void CheckRequestUtils::setHttpRequest(
...
+        }
       } else {
-        // Merge duplicate headers.
-        (*mutable_headers)[key].append(",").append(std::string(e.value().getStringView()));
+        if (mutable_headers->find(key) == mutable_headers->end()) {
+          (*mutable_headers)[key] = std::string(e.value().getStringView());
+        } else {
+          // Merge duplicate headers.
+          (*mutable_headers)[key].append(",").append(std::string(e.value().getStringView()));
+        }
```
ggreenway, line 158: This block is a duplicate of the above. Please combine them; I think the condition can be `if (request_header_matchers == nullptr || request_header_matchers->matches(key))`

source/extensions/filters/common/ext_authz/ext_authz_http_impl.cc
```diff
@@ -245,18 +201,23 @@ void RawHttpClientImpl::check(RequestCallbacks& callbacks,
...
+    if (key == Http::Headers::get().Path) {
+      if (!config_->pathPrefix().empty()) {
         headers->addCopy(key, absl::StrCat(config_->pathPrefix(), header.second));
       } else {
         headers->addCopy(key, header.second);
       }
+      continue;
     }
+
+    headers->addCopy(key, header.second);
```
ggreenway, line 220: Revert to original.

source/extensions/filters/common/ext_authz/ext_authz_http_impl.h
```diff
@@ -140,7 +95,7 @@ class ClientConfig {
   static MatcherSharedPtr
   toUpstreamMatchers(const envoy::type::matcher::v3::ListStringMatcher& list);
 
-  const MatcherSharedPtr request_header_matchers_;
+  MatcherSharedPtr request_header_matchers_;
```
ggreenway, line 98: Why did this become non-const?

source/extensions/filters/http/ext_authz/ext_authz.h
```diff
@@ -94,6 +96,24 @@ class FilterConfig {
...
       }
     }
+
+    if (config.has_allowed_headers() &&
+        config.http_service().authorization_request().has_allowed_headers()) {
+      ExceptionUtil::throwEnvoyException("Invalid duplicate configuration for allowed_headers.");
+    }
+
+    if (config.has_grpc_service() && config.has_allowed_headers()) {
+      request_header_matchers_ = Filters::Common::ExtAuthz::CheckRequestUtils::toRequestMatchers(
```
ggreenway, line 106: Please add comment that unset request_header_matchers_ means to allow all, and that this is done for backwards compatibility.

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
```diff
@@ -632,27 +711,30 @@ TEST_P(ExtAuthzGrpcIntegrationTest, HTTP1DownstreamRequestWithBody) {
   expectCheckRequestWithBody(Http::CodecType::HTTP1, 4);
 }
 
-// Verifies that the request body is included in the CheckRequest when the downstream protocol is
-// HTTP/1.1 and the size of the request body is larger than max_request_bytes.
+// // // Verifies that the request body is included in the CheckRequest when the downstream protocol
```
ggreenway, line 714: Formatting: why so many `//`?

ggreenway (changes requested):
/wait

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
rulex123: not sure what I did there.. reverted

source/extensions/filters/common/ext_authz/ext_authz_http_impl.h
rulex123: reverted

ggreenway:
Apologies, but I haven't had time to review this yet, and I'll be out until 11/28.

ggreenway (approved):
This looks great. Thanks!

PR #22884, opened 2022-08-29
Add optional ttl to key_value_store_base

caschoener:
This addresses [#22303 ](https://github.com/envoyproxy/envoy/issues/22303).  Use the existing TTLManager class to register a timer for each ttl as it's received with a callback to remove the associated key.

Testing: Added a unit test for new behavior.


source/common/common/key_value_store_base.cc
```diff
@@ -35,12 +37,20 @@ KeyValueStoreBase::KeyValueStoreBase(Event::Dispatcher& dispatcher,
...
+      ttl_manager([this](const auto& expired) { ttlCallback(expired); }, dispatcher,
+                  dispatcher.timeSource()) {
   if (flush_interval.count() > 0) {
     flush_timer_->enableTimer(flush_interval);
   }
 }
 
+void KeyValueStoreBase::ttlCallback(const std::vector<std::string>& expired) {
+  for (const auto& key : expired) {
+    remove(key);
```
caschoener, line 50: This is inefficient if there's a large number of keys.  Perhaps I should implement a remove_batch(keys) function for the class

source/common/common/key_value_store_base.cc
```diff
@@ -72,6 +83,9 @@ void KeyValueStoreBase::addOrUpdate(absl::string_view key_view, absl::string_vie
   if (max_entries_ && store_.size() > max_entries_) {
     store_.pop_front();
   }
+  if (ttl)
+    ttl_manager.add(ttl.value(), key);
```
RyanTheOptimist, line 87: nit: Braces are required for all control statements include single line if, while, etc. statements.

https://github.com/envoyproxy/envoy/blob/main/STYLE.md

source/common/common/key_value_store_base.cc
RyanTheOptimist: That seems like a wise idea. I also wonder if there is an opportunity to store less state (by avoiding the need to replicate the key value) for the timeouts but given how the TTL manager works, that seems non-trivial. So `remove(vector)` seems like a fine idea.

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -46,7 +51,20 @@ TEST_F(KeyValueStoreTest, Basic) {
   store_->remove("foo");
   EXPECT_EQ(absl::nullopt, store_->get("foo"));
 }
+TEST_F(KeyValueStoreTest, TTL) {
+  EXPECT_EQ(absl::nullopt, store_->get("foo"));
```
RyanTheOptimist, line 55: nit: This is already tested in "Basic" so we can skip it here.

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -46,7 +51,20 @@ TEST_F(KeyValueStoreTest, Basic) {
...
   EXPECT_EQ(absl::nullopt, store_->get("foo"));
 }
+TEST_F(KeyValueStoreTest, TTL) {
+  EXPECT_EQ(absl::nullopt, store_->get("foo"));  
 
+  store_->addOrUpdate("foo", "bar", std::chrono::milliseconds(5));
+  EXPECT_EQ("bar", store_->get("foo").value());
+  // Advance timer, but not enough to trigger expiry
+  test_time_.setSystemTime(std::chrono::milliseconds(1));
+  ttl_timer_->invokeCallback();
```
RyanTheOptimist, line 61: In production, would we expect the timer to fire at this point? I would have expected that it would be armed to fire after 5ms at which point it would do the eviction. Perhaps we can add an EXPECT_CALL like this:
`EXPECT_CALL(*ttl_timer_, enableTimer(std::chrono::milliseconds(timeout), _));`
just before we call `addOrUpdate()`.

RyanTheOptimist (commented):
Nice! I think this need to have the format fixed, though as there are some rather long lines.

source/common/common/key_value_store_base.cc
caschoener: I actually just looked into it and QuicheLinkedHashMap doesn't support batch deletion... guess this will have to be good enough for now unless we want to change the store_ type

source/common/common/key_value_store_base.cc
```diff
@@ -58,20 +69,32 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+    ttl = absl::nullopt;
+  }
+
   // Attempt to insert the entry into the store. If it already exists, remove
   // the old entry and insert the new one so it will be in the proper place in
   // the linked list.
   if (!store_.emplace(key, value).second) {
     store_.erase(key);
     store_.emplace(key, value);
+    ttl_manager_.clear(key);
```
caschoener, line 89: Changed the behavior to clear any existing ttl if the key gets updated with ttl=null

source/common/common/key_value_store_base.cc
```diff
@@ -1,5 +1,9 @@
 #include "source/common/common/key_value_store_base.h"
 
+#include <algorithm>
+#include <chrono>
+#include <optional>
```
abeyad, line 5: you'll want to include absl/types/optional.h, not the optional in the standard library

RyanTheOptimist:
FYI: //test/common/http:http_server_properties_cache_impl_test       FAILED TO BUILD

source/common/common/key_value_store_base.cc
```diff
@@ -35,12 +39,20 @@ KeyValueStoreBase::KeyValueStoreBase(Event::Dispatcher& dispatcher,
     : max_entries_(max_entries), flush_timer_(dispatcher.createTimer([this, flush_interval]() {
         flush();
         flush_timer_->enableTimer(flush_interval);
-      })) {
+      })),
+      ttl_manager_([this](const auto& expired) { ttlCallback(expired); }, dispatcher,
```
abeyad, line 43: nit: const std::vector<std::string>& expired might be more clear than using auto

source/common/common/key_value_store_base.cc
```diff
@@ -58,20 +70,31 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
 
-void KeyValueStoreBase::addOrUpdate(absl::string_view key_view, absl::string_view value_view) {
+void KeyValueStoreBase::addOrUpdate(absl::string_view key_view, absl::string_view value_view,
+                                    absl::optional<std::chrono::milliseconds> ttl) {
   ENVOY_BUG(!under_iterate_, "addOrUpdate under the stack of iterate");
   std::string key(key_view);
   std::string value(value_view);
+  // Remove ttl if <= 0
+  if (ttl && ttl <= std::chrono::milliseconds(0)) {
+    ASSERT(false);
```
abeyad, line 80: remove?

source/common/common/key_value_store_base.cc
RyanTheOptimist: It's true that linked hash map doesn't have a bulk delete, but the remove() method calls flush after each remove. So this current implementation of ttlCallback will result in N flushes if N keys are removed. It seems like it might be nice to avoid the extra flushes if we could. WDYT?

source/common/common/key_value_store_base.cc
```diff
@@ -58,20 +70,31 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+    ttl = absl::nullopt;
+  }
```
RyanTheOptimist, line 82: If a TTL of 0 is an error, then we should probably not store the entry in the KV store at all. In other words, I'm suggest a return here, not setting the ttl to nullopts, I also wonder if we should enforce a min TTL.

source/common/common/key_value_store_base.cc
```diff
@@ -58,20 +70,31 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
   }
   if (max_entries_ && store_.size() > max_entries_) {
     store_.pop_front();
   }
+  if (ttl) {
+    ttl_manager_.add(ttl.value(), key);
+  }
   if (!flush_timer_->enabled()) {
     flush();
   }
```
abeyad, line 100: I think we should also refresh the TTL (e.g. clear() followed by add()) when `get()` is called for a key.

caschoener: Oh!  I didn't know that's the behavior we were going for.

source/common/common/key_value_store_base.cc
caschoener: This is Ryan's suggestion (hopefully) to use `ASSERT(false);` to throw an error in debug mode if an invalid ttl is passed in.

test/extensions/key_value/file_based/key_value_store_test.cc
RyanTheOptimist: Maybe I missed this but did you add an EXPECT_CALL?

envoy/common/key_value_store.h
```diff
@@ -21,8 +23,11 @@ class KeyValueStore {
    * Adds or updates a key:value pair in the store.
    * @param key supplies a key to add or update.
    * @param value supplies the value to set for that key.
+   * @param ttl optionally give a ttl>0. If it's an update, ttl will be updated as well. If no ttl,
```
RyanTheOptimist, line 26: How about something like: `@param ttl optionally specifies a lifetime after which this entry will be removed. ttl must be greater than 0.` or some such?

source/common/common/key_value_store_base.cc
RyanTheOptimist: Do we have test coverage for updating the TTL from a valid TTL to null? Or from one ttl to another before the timer fires? If not, we should add it.

source/common/common/key_value_store_base.cc
RyanTheOptimist: Interesting idea, @abeyad. For alt-svc we definitely don't want to extend the ttl when we read from the store. Do you think we want this behavior for xDS?

caschoener: Talked to Ali offline, we think the current behavior will work for both the xds and alt-svc use-cases.

abeyad: Chatted w/ Chris offline just now as well, I agree we don't want to do this in the KV store.  I was thinking of the scenario where a resource doesn't have a TTL, so we won't have any eviction and if the resource stops getting referenced, we wouldn't get updates for it, so we'd just have an orphaned entry laying around.  But I think we can and should solve for this at a layer above the KV store APIs (in our custom xDS store that uses the KV store).

So in conclusion: ignore what I said :)

RyanTheOptimist: Heh. Yeah, I think doing that at a layer above the KV store sounds like the right approach, too!

envoy/common/key_value_store.h
```diff
@@ -21,8 +23,11 @@ class KeyValueStore {
    * Adds or updates a key:value pair in the store.
    * @param key supplies a key to add or update.
    * @param value supplies the value to set for that key.
+   * @param ttl @param ttl optionally specifies a lifetime after which this entry will be removed.
```
RyanTheOptimist, line 26: nit: remove the extra "@param ttl"

source/common/common/key_value_store_base.h
```diff
@@ -27,14 +30,18 @@ class KeyValueStoreBase : public KeyValueStore,
 
   std::string error;
   // KeyValueStore
-  void addOrUpdate(absl::string_view key, absl::string_view value) override;
+  void addOrUpdate(absl::string_view key, absl::string_view value,
+                   absl::optional<std::chrono::seconds> ttl) override;
+  // Removes all keys before calling flush()
```
RyanTheOptimist, line 35: nit: this method should be moved out from under the `// KeyValueStore` comment, since it's specific to this class, not to the interface.

source/common/common/key_value_store_base.cc
```diff
@@ -53,25 +58,46 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+  }
+  if (!flush_timer_->enabled()) {
+    flush();
+  }
+}
+
+void KeyValueStoreBase::removeBatch(const std::vector<std::string>& keys) {
+  ENVOY_BUG(!under_iterate_, "remove_batch under the stack of iterate");
+  for (const auto& key : keys) {
+    store_.erase(std::string(key));
```
RyanTheOptimist, line 99: Should we remove the keys from the ttl manager too?

(Do we need test coverage for this case?)

test/common/http/http_server_properties_cache_impl_test.cc
```diff
@@ -76,9 +76,9 @@ TEST_F(HttpServerPropertiesCacheImplTest, SetAlternativesThenSrtt) {
   initialize();
   EXPECT_EQ(0, protocols_->size());
   EXPECT_EQ(std::chrono::microseconds(0), protocols_->getSrtt(origin1_));
-  EXPECT_CALL(*store_, addOrUpdate("https://hostname1:1", "alpn1=\"hostname1:1\"; ma=5|0|0"));
+  EXPECT_CALL(*store_, addOrUpdate("https://hostname1:1", "alpn1=\"hostname1:1\"; ma=5|0|0", _));
```
RyanTheOptimist, line 79: Can we use nullopt here like we do in the other tests? (or kNoTtl, or somesuch?)

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -34,42 +37,74 @@ class KeyValueStoreTest : public testing::Test {
...
   EXPECT_EQ(absl::nullopt, store_->get("foo"));
 }
 
+TEST_F(KeyValueStoreTest, TTL) {
+  test_time_.setSystemTime(std::chrono::milliseconds(0));
+  EXPECT_CALL(*ttl_timer_, enableTimer(std::chrono::milliseconds(5000), _));
+  store_->addOrUpdate("foo", "bar", std::chrono::seconds(5));
+  EXPECT_EQ("bar", store_->get("foo").value());
+  // Advance timer to trigger expiry
+  test_time_.setSystemTime(std::chrono::milliseconds(10000));
```
RyanTheOptimist, line 61: nit: I would expect the timer to go off at the time it is scheduled to on line 57, namely after 5s not after 10s.

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -34,42 +37,74 @@ class KeyValueStoreTest : public testing::Test {
...
+  ttl_timer_->invokeCallback();
+  EXPECT_EQ(absl::nullopt, store_->get("foo"));
+}
+
+TEST_F(KeyValueStoreTest, TTLUpdate) {
+  test_time_.setSystemTime(std::chrono::milliseconds(0));
+
+  store_->addOrUpdate("foo", "bar", std::chrono::seconds(10));
```
RyanTheOptimist, line 69: Please add the EXPECT_CALLs on the timer here. And while we're at it, let's advance the time so we get:
t=0: add(ttl=10)
t=1: add(ttl=5)
t=6: timeFires

The important thing about the EXPECT_CALL is that it guarantees that the expiration is scheduled at the correct time.

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -34,42 +37,74 @@ class KeyValueStoreTest : public testing::Test {
...
+  store_->addOrUpdate("foo", "bar", std::chrono::seconds(5));
+  // Advance timer to trigger expiry
+  test_time_.setSystemTime(std::chrono::milliseconds(8000));
+  ttl_timer_->invokeCallback();
+  EXPECT_EQ(absl::nullopt, store_->get("foo"));
+}
+
+TEST_F(KeyValueStoreTest, TTLUpdateNull) {
```
RyanTheOptimist, line 77: So this does an update of the ttl from non-null to null. We should also test null to non-null.

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -34,42 +37,74 @@ class KeyValueStoreTest : public testing::Test {
...
+  store_->addOrUpdate("foo", "bar", absl::nullopt);
   EXPECT_EQ("bar", store_->get("foo").value());
-  store_->addOrUpdate("foo", "eep");
+  store_->addOrUpdate("foo", "eep", absl::nullopt);
   EXPECT_EQ("eep", store_->get("foo").value());
   store_->remove("foo");
   EXPECT_EQ(absl::nullopt, store_->get("foo"));
 }
 
+TEST_F(KeyValueStoreTest, TTL) {
```
RyanTheOptimist, line 55: nit: Here and below "TTL" => "Ttl" per the style guide.

test/extensions/key_value/file_based/key_value_store_test.cc
```diff
@@ -34,42 +37,74 @@ class KeyValueStoreTest : public testing::Test {
...
+  EXPECT_EQ(absl::nullopt, store_->get("foo"));
+}
+
+TEST_F(KeyValueStoreTest, TTLUpdateNull) {
+  test_time_.setSystemTime(std::chrono::milliseconds(0));
+  store_->addOrUpdate("foo", "bar", std::chrono::seconds(5));
+  store_->addOrUpdate("foo", "bar", absl::nullopt);
+  // Expect timer is disabled
+  EXPECT_FALSE(ttl_timer_->enabled());
+}
```
RyanTheOptimist, line 83: Can we also have a test of `remove()` to see what it does to the timer?

RyanTheOptimist (commented):
Looking good!

source/common/common/key_value_store_base.cc
caschoener: Not here, exactly, but yes good catch we do need this in the remove() call.  And I'll rename this back to ttlcallback()

test/extensions/key_value/file_based/key_value_store_test.cc
caschoener: Maybe this comment is based on an old revision? or I'm missing something

RyanTheOptimist: Here is the code I see in the current version of the PR:

```
  test_time_.setSystemTime(std::chrono::milliseconds(0));
  EXPECT_CALL(*ttl_timer_, enableTimer(std::chrono::milliseconds(5000), _));
  store_->addOrUpdate("foo", "bar", std::chrono::seconds(5));
  EXPECT_EQ("bar", store_->get("foo").value());
  // Advance timer to trigger expiry
  test_time_.setSystemTime(std::chrono::milliseconds(10000));
  ttl_timer_->invokeCallback();
```

The callback seems to be invoked when the time is 10s. (The timer is *enabled* to fire when the time is 5s but that's not when the callback is invoked.)

And actually, we might want to tweak the comment from "Advance timer to trigger expiry" to something like "Advance time to when the timer is scheduled to fire".

caschoener: Oh I understand now. I was mixing up expect and `EXPECT`, if that makes sense.  I can change to the precise numbers throughout.  

And sure can do

source/common/common/key_value_store_base.h
```diff
@@ -24,17 +27,20 @@ class KeyValueStoreBase : public KeyValueStore,
   // parses key value pairs from |contents| and inserts into store_.
   // Returns true on success and false on failure.
   bool parseContents(absl::string_view contents);
+  void ttlCallback(const std::vector<std::string>& keys);
 
   std::string error;
```
RyanTheOptimist, line 32: Admittedly, this seems to be pre-existing, but as long as you're here, can you remove this?

source/common/common/key_value_store_base.h
```diff
@@ -24,17 +27,20 @@ class KeyValueStoreBase : public KeyValueStore,
   // parses key value pairs from |contents| and inserts into store_.
   // Returns true on success and false on failure.
   bool parseContents(absl::string_view contents);
+  void ttlCallback(const std::vector<std::string>& keys);
```
RyanTheOptimist, line 30: nit: Please comment. Instead of `ttlCallback()` how about something like `onExpiredKeys()`, or `onTtlExpired()`?

RyanTheOptimist:
It looks like the most recent changes were force pushed. Was that to fix DCO? If not, please don't force push as per the guidance in

https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md

source/common/common/key_value_store_base.cc
```diff
@@ -53,32 +58,54 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+  if (ttl) {
+    ttl_manager_.add(std::chrono::milliseconds(ttl.value()), key);
+  }
+  if (!flush_timer_->enabled()) {
+    flush();
+  }
+}
+
+void KeyValueStoreBase::ttlCallback(const std::vector<std::string>& keys) {
+  ENVOY_BUG(!under_iterate_, "remove_batch under the stack of iterate");
```
RyanTheOptimist, line 97: nit: the message here says "remove_batch" but that's not the name of the method.

RyanTheOptimist (commented):
Almost there

caschoener:
Not to fix DCO.  But envoy presubmit is reporting dependency failures due to `...antlr4_runtimes/temp5394868970196215811/v4.10.1.tar.gz: GET returned 404 Not Found` which seemed weird and unrelated.

RyanTheOptimist:
> Not to fix DCO. But envoy presubmit is reporting dependency failures due to `...antlr4_runtimes/temp5394868970196215811/v4.10.1.tar.gz: GET returned 404 Not Found` which seemed weird and unrelated.
> 
> So I tried rebasing.

I see. If that happens again, we can probably fix it. Either by doing a `git merge main` or by doing something else to clean things up. I'm aways happy to help sort out git issues. It's kinda fun torture :)

test/common/http/http_server_properties_cache_impl_test.cc
```diff
@@ -13,6 +13,8 @@ namespace Envoy {
 namespace Http {
 
 namespace {
+
+static const absl::optional<std::chrono::seconds> kNoTtl = absl::nullopt;
```
RyanTheOptimist, line 17: nit: newline after.

RyanTheOptimist (approved):

RyanTheOptimist:
@snowp PTAL for cross-company review.

source/common/common/key_value_store_base.cc
```diff
@@ -53,32 +58,54 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
   }
   if (max_entries_ && store_.size() > max_entries_) {
     store_.pop_front();
   }
+  if (ttl) {
+    ttl_manager_.add(std::chrono::milliseconds(ttl.value()), key);
+  }
+  if (!flush_timer_->enabled()) {
+    flush();
+  }
```
snowp, line 93: Why are we now flushing during add? Is this just a bug fix?

snowp (commented):
LGTM, just one question

source/common/common/key_value_store_base.cc
caschoener: This is just github messing up the diff.  The addOrUpdate function already called `flush()` (to write out the changes to disk).  I added a new function `onExpiredKeys()` with the same 3 lines for flush, and github put the new lines in the wrong place

PR #22887, opened 2022-08-29
HTTPs in HTTP CONNECT tunnel integration test.

yanjunxiang-google:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/extensions/transport_sockets/internal_upstream/BUILD
+++ b/test/extensions/transport_sockets/internal_upstream/BUILD
@@ -43,6 +43,9 @@ envoy_extension_cc_test(
         "//source/extensions/access_loggers/file:config",
         "//source/extensions/bootstrap/internal_listener:config",
         "//source/extensions/filters/network/tcp_proxy:config",
+        "//source/extensions/clusters/dynamic_forward_proxy:cluster",
+        "//source/extensions/filters/http/dynamic_forward_proxy:config",
+        "//source/extensions/filters/http/router:config",
         "//source/extensions/io_socket/user_space:config",
         "//source/extensions/transport_sockets/internal_upstream:config",
         "//source/extensions/transport_sockets/raw_buffer:config",
@@ -54,5 +57,6 @@ envoy_extension_cc_test(
         "@envoy_api//envoy/config/bootstrap/v3:pkg_cc_proto",
         "@envoy_api//envoy/extensions/filters/network/tcp_proxy/v3:pkg_cc_proto",
         "@envoy_api//envoy/extensions/transport_sockets/internal_upstream/v3:pkg_cc_proto",
+        "@envoy_api//envoy/extensions/transport_sockets/raw_buffer/v3:pkg_cc_proto",
     ],
 )
--- a/test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
+++ b/test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
@@ -232,6 +232,8 @@ public:
                            dns_cache_config);
   void setupBootstrapExtension();
   void setupTransportSocket(envoy::config::listener::v3::FilterChain& filter_chain);
+  void setupTcpProxyFilter(envoy::config::listener::v3::Filter& filter, const std::string& filter_name,
+                           const std::string& cluster_name, const std::string& stats_prefix);
   void setupExternalListenerConfig();
   void setupInternalUpstreamConfig();
   void setupInternalListenerCofnig();
@@ -251,9 +253,11 @@ public:
 
   std::unique_ptr<ClientTestConnection> conn_;
   std::shared_ptr<WaitForPayloadReader> payload_reader_;
-  bool clear_http_test_{true};
-  bool tcp_proxy_filter_{true};
-  bool hcm_filter_{false};
+  bool clear_http_test_{false};
+  bool external_tcp_proxy_filter_{false};
+  bool external_hcm_filter_{false};
+  bool internal_tcp_proxy_filter_{false};
+  bool internal_hcm_filter_{false};
   std::string request_data_{"foo"};
   std::string response_data_{"bar"};
 };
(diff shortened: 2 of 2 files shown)


test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
```diff
@@ -182,5 +191,392 @@ TEST_F(InternalUpstreamIntegrationTest, BasicFlowWithoutTransportSocket) {
...
+  auto port_str = std::to_string(fake_upstreams_[0]->localAddress()->ip()->port());
+  request_msg.clear();
+  request_msg =
+    "GET / HTTP/1.1\r\n"
+    "Host: localhost:"+ port_str  + "\r\n"
+    "\r\n\r\n";
+  buffer.add(request_msg);
+  conn_->write(buffer, false);
+  dispatcher_->run(Event::Dispatcher::RunType::NonBlock);
+  conn_->close(Network::ConnectionCloseType::FlushWrite);
```
kyessenov, line 522: I think it would be good to process response here, to make sure the roundtrip works.

test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
```diff
@@ -182,5 +191,392 @@ TEST_F(InternalUpstreamIntegrationTest, BasicFlowWithoutTransportSocket) {
...
+
+// Setup the internal upstream cluster which bridge the traffic back to internal listener.
+void HttpsInspectionIntegrationTest::setupInternalUpstreamConfig() {
+  config_helper_.addConfigModifier([&](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
+    auto* static_resources = bootstrap.mutable_static_resources();
+    auto* cluster = static_resources->mutable_clusters()->Add();
+    cluster->set_name("internal_upstream");
+    // Insert internal upstream transport.
+    TestUtility::loadFromYaml(R"EOF(
+        name: envoy.transport_sockets.internal_upstream
```
kyessenov, line 388: Do you really need to pass metadata for this test? I think it would be good to keep test contained to the data path first, before making it more complicated.

test/extensions/transport_sockets/internal_upstream/BUILD
```diff
@@ -43,6 +43,9 @@ envoy_extension_cc_test(
         "//source/extensions/access_loggers/file:config",
         "//source/extensions/bootstrap/internal_listener:config",
         "//source/extensions/filters/network/tcp_proxy:config",
+        "//source/extensions/clusters/dynamic_forward_proxy:cluster",
+        "//source/extensions/filters/http/dynamic_forward_proxy:config",
```
kyessenov, line 47: General comment: we're generally trying to limit dependencies from the integration test on real filters. Maybe create another test file to keep it separate (and I think this could live in internal_listener test suite).

test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
yanjunxiang-google: done

test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
yanjunxiang-google: done

test/extensions/transport_sockets/internal_upstream/BUILD
yanavlasov: +1 integration tests should only rely on "core" extensions. For these tests DFP does not really make any difference. Please change the test to not use DFP.

test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
```diff
@@ -182,5 +191,399 @@ TEST_F(InternalUpstreamIntegrationTest, BasicFlowWithoutTransportSocket) {
...
+  config_helper_.addBootstrapExtension(R"EOF(
+    name: envoy.bootstrap.internal_listener
+    typed_config:
+      "@type": type.googleapis.com/envoy.extensions.bootstrap.internal_listener.v3.InternalListener
+    )EOF");
+}
+
+void HttpsInspectionIntegrationTest::setupTransportSocket(envoy::config::listener::v3::FilterChain& filter_chain) {
+  if (!clear_http_test_) {
+    auto* transport_socket =  filter_chain.mutable_transport_socket();
```
yanavlasov, line 334: Is it possible to express all these changes in yaml? It is much easier to read than proto modifications in C++

test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
```diff
@@ -182,5 +191,399 @@ TEST_F(InternalUpstreamIntegrationTest, BasicFlowWithoutTransportSocket) {
...
+         get_msg.length()) {
+    dispatcher_->run(Event::Dispatcher::RunType::NonBlock);
+  }
+
+  FakeRawConnectionPtr fake_upstream_connection;
+  ASSERT_TRUE(fake_upstreams_[0]->waitForRawConnection(fake_upstream_connection));
+  ASSERT_TRUE(fake_upstream_connection->waitForData(
+      FakeRawConnection::waitForInexactMatch("GET / HTTP/1.1")));
+  // Sends a response back to client.
+  payload_reader_->set_data_to_wait_for("HTTP/1.1 503 Service Unavailable");
```
yanavlasov, line 524: Why do you get 503? I think you can make it work such that fake upstream responds with 200. It would be good to make sure that routing on internal listener works.

test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc
```diff
@@ -182,5 +191,399 @@ TEST_F(InternalUpstreamIntegrationTest, BasicFlowWithoutTransportSocket) {
...
+  skip_tag_extraction_rule_check_ = true;
+  initialize();
+  // Open clear-text connection.
+  conn_->connect();
+  Buffer::OwnedImpl buffer;
+  std::string request_msg =
+    "CONNECT www.cnn.com:80 HTTP/1.1\r\n"
+    "Host: www.cnn.com:80\r\n"
+    "\r\n\r\n";
+  payload_reader_->set_data_to_wait_for("HTTP/1.1 200 OK");
```
yanavlasov, line 554: It is possible that not all response data to CONNECT request was read before trying to switch to TLS. As a result TLS client is trying to interpret leftover bytes from CONNECT response as TLS. It is unlikely but could be source of flakiness down the road.

Strictly speaking you need to wait for `\r\n\r\n` before switching to TLS. What happens if you wait for that sequence. If you want to verify both `HTTP/1.1 200 OK` and wait for `\r\n\r\n` you may need to modify the WaitForPayloadReader to wait until multiple strings match.

yanavlasov (changes requested):
/wait

PR #22888, opened 2022-08-30
docs: list dependencies' license

### LINKED ISSUE #22874: Publishing the copyrights and licenses of the third party libraries
Some of the third party libraries that Envoy uses have license terms that demand that the copyright or the license itself is included in the software and/or documentation.

For example, [the nghttp2 library uses the MIT license](https://github.com/nghttp2/nghttp2/blob/master/COPYING). It contains this sentence: "Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to [...], subject to the following conditions: The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

How does Envoy comply with this legal requirement?

I think the copyright notice and the license of nghttp2 (and other third-party libraries) should appear somewhere in Envoy's documentation. There is already a [table that lists all third-party libraries](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/external_deps). A similar table that contains a list of the external dependencies, their licenses and (if necessary) also the complete license texts would solve this problem.

zhxie:

Commit Message: docs: list dependencies' license
Additional Description:

TODO:

- BoringSSL (FIPS): unknown repo
- liburing: dual license needs to be chosen by code owners
- skywalking-data-collect-protocol: license in README only
- zstd: dual license needs to be chosen by code owners
- wavm: version not found
- V8: unknown repo
- Chromium's trace event headers: license not found
- Common Expression Language (CEL) C++ library: version not found
- Kafka (server binary): unknown repo
- Intel Dlb: unknown repo

Risk Level: Low
Testing: N/A
Docs Changes: Yes
Release Notes: N/A
Partial fix for #22874

### DIFF
--- a/tools/docs/generate_external_deps_rst.py
+++ b/tools/docs/generate_external_deps_rst.py
@@ -127,8 +127,6 @@ def main():
         version = rst_link(render_version(v['version']), get_version_url(v))
         release_date = v['release_date']
         license = v.get('license', '')
-        if license == 'N/A':
-            license = ''
         if license:
             license_url = v.get('license_url', '')
             if license_url:
--- a/tools/docs/generate_external_deps_rst.py
+++ b/tools/docs/generate_external_deps_rst.py
@@ -101,7 +101,7 @@ def get_version_url(metadata):
 
 
 def csv_row(dep):
-    return [dep.name, dep.version, dep.release_date, dep.cpe]
+    return [dep.name, dep.version, dep.release_date, dep.cpe, dep.license]
 
 
 def main():
@@ -112,7 +112,7 @@ def main():
 
     pathlib.Path(security_rst_root).mkdir(parents=True, exist_ok=True)
 
-    Dep = namedtuple('Dep', ['name', 'sort_name', 'version', 'cpe', 'release_date'])
+    Dep = namedtuple('Dep', ['name', 'sort_name', 'version', 'cpe', 'release_date', 'license'])
     use_categories = defaultdict(lambda: defaultdict(list))
     # Bin rendered dependencies into per-use category lists.
     for k, v in repository_locations.items():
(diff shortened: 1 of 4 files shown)


zhxie:
@phlax can you check in advance if it satisfies.

Link to external deps: https://storage.googleapis.com/envoy-pr/d3a7e84/docs/intro/arch_overview/security/external_deps.html

tools/docs/generate_external_deps_rst.py
```diff
@@ -126,7 +126,14 @@ def main():
         name = rst_link(project_name, project_url)
         version = rst_link(render_version(v['version']), get_version_url(v))
         release_date = v['release_date']
-        dep = Dep(name, project_name.lower(), version, cpe, release_date)
+        license = v.get('license', '')
+        if license == 'N/A':
```
phlax, line 130: im wondering if we need to accomodate N/A if the license is not compulsory (perhaps it should be)

phlax (commented):
@zhxie looks great, thanks for picking this up so quickly

i dont see `license_path` used so im guessing we didnt need it - which i think is also simpler

phlax:
cc @moderation

tools/docs/generate_external_deps_rst.py
zhxie: I am not sure if license is compulsory or not, but in my opinion, I think license should have the same priority as CVE. In fact, CVE is also not compulsory at the moment. It is a simple copy-paste, and I can change it if needed.

phlax: yep - lets go with not making it compulsory, at least for now - but lets drop the N/A - rather if its not filled in just dont add it

zhxie: Fixed.

moderation:
Wondering what is the driver for this. Is there there is an end-user asking for this or is this a nice to have?

Before - https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/external_deps
After - https://storage.googleapis.com/envoy-pr/4c0cb9d/docs/intro/arch_overview/security/external_deps.html

phlax:
> Wondering what is the driver for this. Is there there is an end-user asking for this or is this a nice to have?

see https://github.com/envoyproxy/envoy/issues/22874 for related discussion - but the issue has come up recently on slack also

mostly license compliance - ie - do whatever you want as long as you "include" this license

phlax:
> Before/After ...

some rows are a bit wide on non-breaking cols  - perhaps we could do some truncation or similar - altho probs outside scope of this PR

https://storage.googleapis.com/envoy-pr/a07d03d/docs/intro/arch_overview/security/external_deps.html?highlight=external#envoy-wasm-runtime-wavm

phlax (approved):
lgtm, thanks @zhxie 

lets land this and follow up with the missing licenses etc

PR #19828, opened 2022-02-04
cache filter: Create an abstract test fixture for HttpCache implementations.

mkbehr:
Commit Message: cache filter: Create an abstract test fixture for HttpCache implementations.
Additional Description:
Move SimpleHttpCacheTest into a new HttpCacheImplementationTest filter. Other
HttpCache implementations can reuse the tests in HttpCacheImplementationTest by
defining an HttpCacheTestDelegate, which wraps the cache.

HttpCacheTestDelegate allows caches to specify whether or not they support
validating stale cache entries and updating their headers. The relevant tests
from SimpleHttpCacheTest are split into different versions to run depending on
whether or not the cache supports validation.

This also moves operator definitions out of
test/extensions/filters/http/cache/common.h to avoid linker errors.

Risk Level: Low
Testing: bazel test //test/extensions/filters/http/cache/...
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: N/A

PR #22901, opened 2022-08-30
Add a completion callback to insertTrailers in cache_filter

ravenblackx:
Commit Message: Add a completion callback to insertTrailers in cache_filter
Additional Description: Without this additional callback, an asynchronous cache implementation has no way to signal completion of an insert operation involving trailers. This also changes the cache test implementation to require that the completion callback has been called, before continuing to other parts of the test, where previously it would assume completion as soon as insertTrailers was called.
Risk Level: Low (change to a WIP filter)
Testing: Tests updated
Docs Changes: n/a
Release Notes: added a completion callback to insertTrailers in cache interface. Any external cache implementation extensions will need to also add this callback, and call it on completion.

### DIFF
--- a/test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
+++ b/test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
@@ -103,30 +103,45 @@ absl::Status HttpCacheImplementationTest::insert(
     return absl::OkStatus();
   }
   if (body.empty() && trailers.has_value()) {
-    inserter->insertTrailers(trailers.value());
+    inserter->insertTrailers(trailers.value(), nullptr);
     return absl::OkStatus();
   }
 
   // For responses with body and trailers, wait for insertBody's callback before
-  // calling insertTrailers.
+  // calling insertTrailers. Note, in a multipart body test this would need to
+  // check for the callback having been called for *every* body part, but since
+  // the test only uses single-part bodies, inserting trailers in direct response
+  // to the callback works.
   auto insert_mutex = std::make_shared<absl::Mutex>();
   bool insert_done = false;
   auto insert_cancelled = std::make_shared<bool>(false);
   absl::Status insert_status;
   InsertCallback insert_callback = [&trailers, &inserter, insert_mutex, &insert_done,
                                     insert_cancelled, &insert_status](bool success_ready_for_more) {
-    absl::MutexLock lock(insert_mutex.get());
-    if (*insert_cancelled) {
-      return;
+    {
+      absl::MutexLock lock(insert_mutex.get());
+      if (*insert_cancelled) {
+        return;
+      }
+      if (!success_ready_for_more) {
+        insert_status = absl::UnknownError("Insert was aborted by cache");
+        insert_done = true;
+        return;
+      }
     }
-    if (!success_ready_for_more) {
-      insert_status = absl::UnknownError("Insert was aborted by cache");
+    inserter->insertTrailers(trailers.value(), [&insert_done, insert_mutex, insert_cancelled,
+                                                &insert_status](bool success) {
+      absl::MutexLock lock(insert_mutex.get());
+      if (*insert_cancelled) {
+        return;
+      }
+      if (!success) {
+        insert_status = absl::UnknownError("Insert was aborted by cache in insertTrailers");
+      } else {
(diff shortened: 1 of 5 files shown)


test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
```diff
@@ -68,21 +69,16 @@ void HttpCacheImplementationTest::updateHeaders(
...
-    if (*cancelled) {
-      return;
-    }
-    lookup_result_ = std::move(result);
-    done = true;
-  });
-  absl::MutexLock lock(mu.get());
-  EXPECT_TRUE(mu->AwaitWithTimeout(absl::Condition(&done_condition), absl::Seconds(3)))
-      << "timed out in lookup " << request_path;
+  auto headers_promise = std::make_shared<std::promise<LookupResult>>();
```
jmarantz, line 72: Potential use of std::promise came up in another review I did a while back: https://github.com/envoyproxy/envoy/pull/19828

I pushed back on that primarily because it won't work with simulated time, and I have the same concerns about the use here.

However I also will note that the current code doesn't work with simulated time either, but I am concerned about effort spent rewriting/reviewing code in a way that doesn't fix that.

ravenblackx: The alternative nod to asynchrony would have the converse problem - if it works with simulated time then it doesn't work with a thread that takes some real time to do the work, because it advanced fake time and then immediately "times out" before the other thread completed the work. The only way to support both at the same time is to just advance the fake clock before any waitFor, which would be no different between promises or mutexes.

I can't see any reason one would want a cache implementation that explicitly injects delays, which would be the only context in which simulated time would be meaningful for this test. Real-time timeouts, on the other hand, are useful for testing any cache that works with non-instant data sources, which I think would be most cache implementations.

Also, I see test_time_system.h's waitFor is doing the exact same thing as this anyway - not changing simulated time, and blocking for up to [some amount of real time].

The amount of extra code to do this with absl::Mutexes in all the necessary contexts is significantly larger and harder to read than the amount of code changed to use promises in place of (no support at all | absl::Mutex), and would still, in the end, behave the exact same way.

jmarantz: The way that code gets to wake up and do work before simulated time expires is via dispatcher timers. Advancing simulated time runs all expired timers.

One reason to utilize simulated time when testing caches is to robustly and quickly (in real time) test expiry of cached items with max-age.

ravenblackx: Sure, but using these real-time timeouts on promises in tests doesn't in any way interfere with that use of simulated time. It doesn't prevent advancing simulated time for a test case where that's useful, and it doesn't wait for simulated time to advance for these test cases where simulated time is [almost certainly] not being used. I mean these timeouts aren't really time-related at all, they're just acting as test failure deadlines (rather than waiting 300 seconds and delivering no useful error message, if we did the same wait without a timeout). And, again, the std::promise variant isn't behaving differently in any meaningful way than the pre-existing absl::Mutex variant was, it's just much simpler code.

I totally get why simulated time is useful (while at Google I even created a cross-process shared simulated clock class for some complex integration tests), but I don't see how it's relevant to test promise timeouts on operations that almost certainly don't even involve simulated time.

ravenblackx: ... If the timeout is considered too controversial I could just remove it and have the promises wait *without* a timeout, for, again, the exact same behavior in practice when the test is passing - the only difference would be that one possible subset of failing tests would be harder to debug. (It would also make running in a debugger easier to control, so it has that going for it.)

jmarantz: This makes sense. I've been a bit slammed (perf season) and I will dig into this.

The idea of adding std::promise into the Envoy codebase came up before and I re-awakened the thread on envoy-dev and invited you also. This makes more sense the way you are using it, as a test-timeout mechanism only. But I'd like to hear from others.

Also since I started the original thread, another std::promise usage was added to Envoy but I think it's in contrib, so maybe there's a lower bar there.

mattklein123 (approved):

PR #22907, opened 2022-08-30
quiche: Use absl::Flag for QUICHE flags

RyanTheOptimist:
quiche: Use absl::Flag for QUICHE flags

Risk Level: Low
Testing: Unit tests
Docs Changes: N/A
Release Notes: N/A


source/common/quic/platform/quiche_flags_impl.cc
```diff
@@ -9,19 +9,89 @@
...
+QUIC_FLAG(quic_reloadable_flag_spdy_testonly_default_true, true)
+QUIC_FLAG(quic_restart_flag_spdy_testonly_default_false, false)
+QUIC_FLAG(quic_restart_flag_spdy_testonly_default_true, true)
+QUIC_FLAG(quic_reloadable_flag_http2_testonly_default_false, false)
+QUIC_FLAG(quic_reloadable_flag_http2_testonly_default_true, true)
+QUIC_FLAG(quic_restart_flag_http2_testonly_default_false, false)
+QUIC_FLAG(quic_restart_flag_http2_testonly_default_true, true)
+#undef QUIC_FLAG
+
+#define QUIC_PROTOCOL_FLAG(type, flag, value, help)                                                \
```
danzh2010, line 76: For protocol flags, we may have extra arg `external_value` in these macros. So I think the QUIC_PROTOCOL_FLAG_MACRO_CHOOSER handling is still needed.

RyanTheOptimist: It doesn't look like we actually do this any longer, since none of the flags have such a value. But I guess it could come back later. Done.

danzh2010:
cc @birenroy

alyssawilk (approved):
Very cool.  
Should we update Envoy::Runtime::refreshReloadableFlags and
quiche::FlagRegistry::getInstance().updateReloadableFlags(
to look more like the standard runtime guard update?

RyanTheOptimist:
> Very cool. Should we update Envoy::Runtime::refreshReloadableFlags and quiche::FlagRegistry::getInstance().updateReloadableFlags( to look more like the standard runtime guard update?

Sounds like a good idea! I filed #22963 to address in a follow up.

PR #22908, opened 2022-08-30
custom_matcher: Test the custom_match effectively

tyxia:
The existing test is using `NeverMatch` matcher which always return false for match() method. So, we didn't really test the matching behavior for custom_match field/mechanism. In this PR:
- Implement a new custom_matcher and factory to do simple string comparison matching.
- Test custom_match with this newly added custom matcher/factory for the expected matching result that `result.on_match_.has_value()` is true

### DIFF
--- a/test/common/matcher/matcher_test.cc
+++ b/test/common/matcher/matcher_test.cc
@@ -208,9 +208,10 @@ matcher_list:
           typed_config:
             "@type": type.googleapis.com/google.protobuf.BoolValue
         custom_match:
-          name: never_match
+          name: custom_match
           typed_config:
             "@type": type.googleapis.com/google.protobuf.StringValue
+            value: custom_foo
   )EOF";
 
   envoy::config::common::matcher::v3::Matcher matcher;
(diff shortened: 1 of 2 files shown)


test/common/matcher/matcher_test.cc
```diff
@@ -208,26 +208,31 @@ TEST_F(MatcherTest, CustomMatcher) {
...
 
   envoy::config::common::matcher::v3::Matcher matcher;
   MessageUtil::loadFromYaml(yaml, matcher, ProtobufMessage::getStrictValidationVisitor());
 
   TestUtility::validate(matcher);
 
-  auto inner_factory = TestDataInputBoolFactory("foo");
-  NeverMatchFactory match_factory;
+  // Build the input data that is to be matched.
+  std::string value = "custom_foo";
```
daixiang0, line 223: Could make it const?

daixiang0 (approved):
LGTM

mattklein123 (approved):

Issue #3058, opened 2018-04-12
Outlier Detection for TCP


bplotnick:
*Title*: *Support outlier detection for TCP mode*

*Description*:
TCP is currently supported for active health checking, but is not supported for passive health checking. There are a few ways that this could be implemented. I'm thinking the simplest check would be TCP connection failures (this is what HAProxy does for layer4 health [observing](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#5.2-observe)). But we might also be able to check on connection failures or something.

ggreenway:
tcp_proxy currently does support outlier detection. See bb069d5a8ff2620c1fb8bd0391491582908f276b. 

Did you observe this not working as expected? Or is the documentation wrong somewhere?

bplotnick:
Ah sorry about that. I did not observe this, it was based purely on documentation. It looks like consecutive 5xx will map to tcp connection failures and do what I want.

I think the part of the documentation that says `...or an event that would cause the HTTP router to return one on the upstream’s behalf (reset, connection failure, etc.)` covers this. Though I may submit a PR to explicitly call out TCP in the documentation.

ggreenway:
Documentation PRs are always greatly appreciated! Sorry the docs weren't clear on this.

PR #22909, opened 2022-08-31
Update documentation for how 5xxs map to TCP traffic

chusteven:
Updating documentation to make clearer this issue: https://github.com/envoyproxy/envoy/issues/3058

Risk Level: Low since it's just documentation?
Testing: Didn't do any -- happy to but was too lazy to set up my environment 😅 
Docs Changes: Specifically for [outlier detection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier), I was confused that 5xx mapped -- for TCP traffic -- to connection failures
Release Notes: Updated documentation on outlier detection
[Optional Runtime guard:] ?
[Optional Fixes #Issue] https://github.com/envoyproxy/envoy/issues/3058
[Optional Fixes commit #PR or SHA] N/A
[Optional Deprecated:] N/A
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):] N/A


api/envoy/api/v2/cluster/outlier_detection.proto
```diff
@@ -24,9 +24,9 @@ option (udpa.annotations.file_status).package_version_status = FROZEN;
 // more information on outlier detection.
 // [#next-free-field: 21]
 message OutlierDetection {
-  // The number of consecutive 5xx responses or local origin errors that are mapped
-  // to 5xx error codes before a consecutive 5xx ejection
-  // occurs. Defaults to 5.
+  // The number of consecutive 5xx responses or local origin errors (e.g. for TCP traffic,
```
htuch, line 27: I think the doc callouts in RST are fine, but the API comment changes are misleading, as they over-index on the TCP case.

chusteven: That makes sense to me. But from my perspective, part of the issue in the first place was that this was named `5xx` which I immediately assumed meant HTTP traffic only. Is there other wording that we can compromise on?

Maybe:
```
The number of consecutive server-side error responses (for HTTP traffic,
5xx responses; for TCP traffic, connection failures; for Redis, failure to respond
PONG, etc.) before a consecutive 5xx ejection occurs. Defaults to 5.
```

I think also L3 traffic is supported but I don't have any experience with that to know the correct wording to use for when that fails 😓

htuch: Let's go with that.

adisuissa:
Please fix DCO.
/wait

yanavlasov (approved):

yanavlasov:
@adisuissa for API stamp

adisuissa (approved):
/lgtm api

htuch (approved):
LGTM, thanks!

PR #22918, opened 2022-08-31
http: cleaning upstream interfaces for upstream filters.

alyssawilk:
This changes the Upstream Filter code to creating the codec filter via factory which 
1) requires adding all UpstreamRequest/CodecFilter interactions to an UpstreamCallback interface accessible through the filter manager
2) requires unhiding the configuration because the presubmit cross-checks for the registered factory require all the config be unhidden
3) allows configuring the codec filter in configuration, in case of eventual alternate terminal filter
4) allows fixing up the filter dependency validator to validate the entire upstream filter chain.

Upstream filters flipped on for CI, SHOULD BE FLIPPED OFF BEFORE SUBMITTING

Risk Level: low assuming it's flipped back off
Testing: updated unit tests, covered by filter integration tests
Docs Changes: n/a
Release Notes: n/a (off by default)


api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -150,7 +150,11 @@ message HttpProtocolOptions {
     AutoHttpConfig auto_config = 5;
   }
 
-  // [#not-implemented-hide:]
+  // .. warning::
```
alyssawilk, line 153: FWIW I don't like unhiding this before it's on, but now that I'm registering the codec filter, all our cross-checks require the field with extension-category be visible :-(

alyssawilk:
@mattklein123 if you have any thoughts about / improvement suggestions for envoy/http/filter.h please let me know

alyssawilk:
cc @paul-r-gall

source/common/http/filter_chain_helper.cc
```diff
@@ -76,7 +76,7 @@ FilterChainUtility::createSingletonDownstreamFilterConfigProviderManager(
   std::shared_ptr<Http::DownstreamFilterConfigProviderManager>
       downstream_filter_config_provider_manager =
           context.singletonManager().getTyped<Http::DownstreamFilterConfigProviderManager>(
-              SINGLETON_MANAGER_REGISTERED_NAME(upstream_filter_config_provider_manager),
+              SINGLETON_MANAGER_REGISTERED_NAME(downstream_filter_config_provider_manager),
```
RyanTheOptimist, line 79: Is this a bugfix?

RyanTheOptimist (commented):
Any chance you could flesh out the PR description? "cleaning" sounds a bit more innocuous than the code in this PR suggests :) There's enough code here that it would be helpful to have a bit of a roadmap.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
htuch: As long as we document as done here and it's default disable I'm OK, @phlax for the challenge here.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -150,7 +150,11 @@ message HttpProtocolOptions {
...
+  //   HTTP filters are not supported by default.
+  //   This warning will be removed as support moves beyond alpha.
+  //
   // Optional HTTP filters for the upstream filter chain.
+  // [#extension-category: envoy.filters.http.upstream]
```
htuch, line 158: For my education, why did we decide to put the filter chain here vs. how this is done for downstream? Curious about the API symmetry/consistency. @envoyproxy/api-shepherds

source/common/http/filter_chain_helper.cc
alyssawilk: yep.  It wasn't a problem before because because there were no pure upstream filters - everything was accidentally using the downstream factory.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
paul-r-gall: I see, so rather than having the set of upstream filters specified once in the HCM with config overrides set in the cluster, we just specify the set of upstream filters in every cluster?

alyssawilk: right now we set them in each cluster.  I think we want to have a default in the bootstrap config, and overrides in every cluster, but we can add that on later (and again I'm hoping you can pick up that part)

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -150,7 +150,11 @@ message HttpProtocolOptions {
     AutoHttpConfig auto_config = 5;
   }
 
-  // [#not-implemented-hide:]
+  // .. warning::
+  //   HTTP filters are not supported by default.
+  //   This warning will be removed as support moves beyond alpha.
+  //
```
mattklein123, line 156: IMO this is kind of confusing. This is upstream only I guess? Or is this going to be downstream as well? If it's upstream only shouldn't this be in the upstream options above? Also per my comment below are we going to allow per-route upstream filter chains? I guess stepping back can we enumerate all of the different ways we eventually want to configure this and then we can sort out the best API?

docs/root/configuration/http/http_filters/codec_filter.rst
```diff
@@ -0,0 +1,8 @@
+.. _config_http_filters_codec:
+
+Codec
+======
+
+The codec filter is the only supported terminal filter for upstream filters.
+
+There currently are no supported configuration options for the codec filter.
```
mattklein123, line 8: Nit: If we are going to publicly expose this, can we make it more clear what this is with the name? "upstream_codec" or possible even "upstream_terminal_filter" or something?

alyssawilk:
@snowp ping?

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
alyssawilk: sorry yeah, this is where we configure the filter chain for upstream filters.  As commented below I had to unhide for now because of the presubmit checks.  

This is the moral equivalent of the filter list in the upstream HCM, only instead of these filters being applied to traffic on that HCM/listener, they're applied to streams in that cluter.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
alyssawilk: I thought for downstream we had one filter chain, and per route filter configs. Did we add more enhancements when I wasn't looking?

source/common/router/upstream_request.cc
```diff
@@ -41,98 +41,108 @@
...
+  callbacks_->upstreamCallbacks()->upstream()->readDisable(false);
+}
+
+void UpstreamCodecFilter::onAboveWriteBufferHighWatermark() {
+  callbacks_->clusterInfo()->stats().upstream_flow_control_paused_reading_total_.inc();
+  callbacks_->upstreamCallbacks()->upstream()->readDisable(true);
+}
+
+void UpstreamCodecFilter::onUpstreamConnectionEstablished() {
+  if (latched_end_stream_.has_value()) {
```
snowp, line 55: Would you expect this to get multiple times? Does the lifetime of the upstream filters live between multiple attempts? If not expected, should this ASSERT?

source/common/router/upstream_request.cc
```diff
@@ -204,6 +227,7 @@ class UpstreamFilterManager : public Http::FilterManager {
     state().encoder_filter_chain_aborted_ = true;
     state().remote_encode_complete_ = true;
     state().local_complete_ = true;
+    // TODO(alyssawilk) this should be done through the router to play well with hedging.
```
snowp, line 230: Is this because we're immediately issuing a local reply here, skipping the retry/hedge handling done in the router?

snowp (changes requested):
Few comments but overall this LGTM!

source/common/router/upstream_request.cc
alyssawilk: (currently) it can only be called once but shouldn't assert because it's totally possible to have the upstream connection established before the full stream is proxied.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
mattklein123: Ah yeah sorry I was confusing route config vs. filter chain config. I agree though that if we could support it or at least future proof the API for it it would be nice.

mattklein123: Also I still have the question about upstream vs. downstream. Do we expect this to work for downstream also? Should this be part of the upstream options? I think there was another comment thread on this but I can't find it.

alyssawilk: this proto is in the namespace envoy.upstreams.http.http_protocol_options

It's the HTTP protocol options that go in cluster config for configuring if an upstream is doing explicit HTTP1/HTTP2/HTTP3, auto, etc.

I think this is the correct place to add http_filters which are a per-cluster upstream filter chain as it's http-specific config, cluster-specific config, and upstream only.

mattklein123: Ah OK sorry I missed that. I though this was generic protocol options. SGTM.

source/common/router/upstream_request.cc
snowp: Yeah I think there is a lot of ways this could work, so probably easiest to just do something simple and then add in different mechanisms down the line if people need to also control what the retry/hedge/etc behavior should be

snowp (approved):
Thanks!

htuch (approved):
/lgtm api

alyssawilk:
whelp I forgot to flip the flag back.  retroactively this is higher risk than it's tagged as.
We had already fixed the blocking bug, so I AFIK it's good to turn up, I'd just planned on doing that in a solo follow-up PR

PR #22921, opened 2022-08-31
xds: Add a KeyValueStore-backed xDS delegate extension to contrib

abeyad:
An xDS delegate extension point was added in
https://github.com/envoyproxy/envoy/pull/22473 to enable custom behavior upon
receiving and loading xDS resources. This change creates an implementation of
the XdsResourcesDelegate interface that is backed by a KeyValueStore.

The intended use case is to enable persisting xDS resources and loading them on
startup in Envoy Mobile, in the event that the xDS control plane is unreachable.

Risk Level: Low
Testing: Unit & Integration tests
Docs Changes: N/A (hidden experimental feature)
Release Notes: N/A


alyssawilk:
@adisuissa you game for first pass on this?

source/common/config/grpc_mux_impl.cc
```diff
@@ -90,6 +90,15 @@ void GrpcMuxImpl::sendDiscoveryRequest(absl::string_view type_url) {
     }
   }
 
+  if (starting_) {
+    // On the initialization of the gRPC mux, load the persisted config, if available. If the xDS
```
adisuissa, line 94: This seems like it could "waste" memory and cpu resources, especially if 1) the save data is stale, 2) the server is ready and can easily serve config.
Shouldn't this be part of a fall-back option?

adisuissa (commented):
Thanks!
One high-level comment, instead of adding the extension under the "xds" directory, it should probably be under the path name "config".

contrib/xds/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,130 @@
...
+
+} // namespace
+
+KeyValueStoreXdsDelegate::KeyValueStoreXdsDelegate(KeyValueStorePtr&& xds_config_store)
+    : xds_config_store_(std::move(xds_config_store)) {}
+
+std::vector<envoy::service::discovery::v3::Resource>
+KeyValueStoreXdsDelegate::getResources(const XdsSourceId& source_id,
+                                       const std::vector<std::string>& resource_names) const {
+  if (resource_names.empty()) {
```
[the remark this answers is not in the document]
abeyad: Done

abeyad:
> Thanks! One high-level comment, instead of adding the extension under the "xds" directory, it should probably be under the path name "config".

Done, thanks for the review!  Addressed the comments and pushed up some commits.  PTAL!

alyssawilk:
still letting Adi drive here but you'll want to sync past the merge conflicts when you have a few :-)

abeyad:
> still letting Adi drive here but you'll want to sync past the merge conflicts when you have a few :-)

PR has been updated with the latest from upstream/main

source/common/config/grpc_mux_impl.cc
```diff
@@ -321,7 +380,17 @@ void GrpcMuxImpl::onEstablishmentFailure() {
       watch->callbacks_.onConfigUpdateFailed(
           Envoy::Config::ConfigUpdateFailureReason::ConnectionFailure, nullptr);
     }
+    if (starting_) {
+      // On the initialization of the gRPC mux, if connection to the xDS server fails, load the
+      // persisted config, if available. The locally persisted config will be used until
+      // connectivity is established with the xDS server.
```
adisuissa, line 386: Can you add an integration tests that emulates the "connectivity is re-established" scenario?

contrib/config/test/kv_store_xds_delegate_integration_test.cc
```diff
@@ -0,0 +1,340 @@
...
+    registerTestServerPorts(named_ports, test_server_);
+  }
+
+  FakeHttpConnectionPtr sds_connection_;
+  FakeStreamPtr sds_stream_;
+  FakeHttpConnectionPtr rtds_connection_;
+  FakeStreamPtr rtds_stream_;
+};
+
+INSTANTIATE_TEST_SUITE_P(IpVersions, KeyValueStoreXdsDelegateIntegrationTest,
```
adisuissa, line 254: This should also be tested against unified-SotW

contrib/config/test/kv_store_xds_delegate_integration_test.cc
```diff
@@ -0,0 +1,340 @@
...
+  test_server_->waitForCounterGe(
+      "cluster.cluster_0.client_ssl_socket_factory.ssl_context_update_by_sds", 1);
+  // Two runtime loads are expected, one for the admin layer and one for the RTDS layer.
+  test_server_->waitForCounterGe("runtime.load_success", 2);
+
+  // Verify that the latest resource values are used by Envoy.
+  checkSecretExists(std::string(CLIENT_CERT_NAME), /*version_info=*/"1");
+  EXPECT_EQ("whatevs", getRuntimeKey("foo"));
+  EXPECT_EQ("yar", getRuntimeKey("bar"));
+  EXPECT_EQ("saz", getRuntimeKey("baz"));
```
adisuissa, line 336: I guess this is the indication that the data was loaded from the KV-store.
To enhance observability I suggest adding a counter whenever data is loaded from the KV-store (for each type).

api/contrib/envoy/extensions/config/v3alpha/kv_store_xds_delegate_config.proto
```diff
@@ -0,0 +1,23 @@
...
+
+option java_package = "io.envoyproxy.envoy.extensions.config.v3alpha";
+option java_outer_classname = "KvStoreXdsDelegateConfigProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/config/v3alpha";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#extension: envoy.xds_delegates.kv_store]
+//
+// Configuration for a KeyValueStore-based XdsResourcesDelegate implementation.
```
alyssawilk, line 17: Comment what it should do?  (update KV store with latest config, etc)

contrib/config/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,127 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace Config {
+namespace {
+
+using ::Envoy::Config::XdsSourceId;
+using ::envoy::extensions::config::v3alpha::KeyValueStoreXdsDelegateConfig;
+
+// TODO(abeyad): Add a per-Envoy instance prefix to the storage key, to allow xDS resources to be
+// persisted per-running-Envoy when Envoy is embedded as a library (e.g. Envoy Mobile).
```
alyssawilk, line 25: Hm, this is an interesting one.  I wonder if we can get away with a winner-takes-all solution here - I'm interested in @RyanTheOptimist thoughts

contrib/config/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,127 @@
...
+} // namespace
+
+KeyValueStoreXdsDelegate::KeyValueStoreXdsDelegate(KeyValueStorePtr&& xds_config_store)
+    : xds_config_store_(std::move(xds_config_store)) {}
+
+std::vector<envoy::service::discovery::v3::Resource>
+KeyValueStoreXdsDelegate::getResources(const XdsSourceId& source_id,
+                                       const std::vector<std::string>& resource_names) const {
+  if (resource_names.empty() || (resource_names.size() == 1 && resource_names[0] == "*")) {
+    // Empty names or one entry with "*" means wildcard.
```
alyssawilk, line 42: optional nit: I mildly dislike having two wildcard options.  Could we validate that the resource name not be empty?

source/common/config/grpc_mux_impl.h
```diff
@@ -178,6 +185,7 @@ class GrpcMuxImpl : public GrpcMux,
   XdsResourcesDelegateOptRef xds_resources_delegate_;
   const std::string target_xds_authority_;
   bool first_stream_request_;
+  bool starting_{true};
```
adisuissa, line 188: nit: starting is too generic. Maybe something in the lines of previously fetched data, or something that will convey the meaning of this

adisuissa (commented):
Overall LGTM.

I wonder how this is impacting warmed objects.
For example, if there was a CDS request that returned some EDS-type clusters, and then a disconnect happened; would you expect the endpoints to be loaded from the KV-store?
I think in the current implementation (assuming ADS is used) this will not happen, but it would be better to validate using an integration test.

alyssawilk (commented):
Very cool!

Are proto format failures real?
Also I'd definitely check coverage for this one as I don't think CI does coverage checks for contrib

contrib/config/source/kv_store_xds_delegate.cc
RyanTheOptimist: Hm. I'm not quite sure what is meant by winner-takes-all. Is the thinking that multiple Envoy instances would share the same KV store and whoever gets there first owns it? Or something else?

api/contrib/envoy/extensions/config/v3alpha/kv_store_xds_delegate_config.proto
abeyad: Done.

contrib/config/source/kv_store_xds_delegate.cc
abeyad: Actually, I think we can ensure different Envoy engines can have different key prefixes to avoid collisions by setting different keys in https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/extensions/key_value/platform/platform.proto#L11

contrib/config/source/kv_store_xds_delegate.cc
abeyad: Unfortunately, Envoy allows these two different ways to specify wildcard (empty resource names or one resource name with "*").  Or do you mean, when invoking getResources, if resource_names is empty, add one `*` resource name to the list before passing in?

contrib/config/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,127 @@
...
+      resources.push_back(std::move(r));
+    }
+  }
+  return resources;
+}
+
+std::vector<envoy::service::discovery::v3::Resource>
+KeyValueStoreXdsDelegate::getAllResources(const XdsSourceId& source_id) const {
+  std::vector<envoy::service::discovery::v3::Resource> resources;
+  // TODO(abeyad): This is slow as we are iterating over all entries in the KV store; the
```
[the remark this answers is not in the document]
abeyad: I like that and was hoping to do something like this.  Only try solving the problem if/when it becomes an issue down the road.

contrib/config/test/kv_store_xds_delegate_integration_test.cc
abeyad: This PR doesn't handle unified-SotW yet.  I wanted to get it working for regular Sotw and then do the same for unified-Sotw in a follow up PR.

source/common/config/grpc_mux_impl.h
abeyad: Done.

contrib/config/source/kv_store_xds_delegate.cc
alyssawilk: Ok in that case do we need a per-instance prefix?

contrib/config/source/kv_store_xds_delegate.cc
alyssawilk: no I didn't know this was prior art.  Don't mind me then :-P

contrib/config/source/kv_store_xds_delegate.cc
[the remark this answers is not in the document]
abeyad: I added a comment to this effect.

contrib/config/source/kv_store_xds_delegate.cc
abeyad: We don't, I pushed a commit removing this comment.

source/common/config/grpc_mux_impl.cc
abeyad: Added a test for this

contrib/config/test/kv_store_xds_delegate_integration_test.cc
abeyad: I added some stats

api/contrib/envoy/extensions/config/v3alpha/kv_store_xds_delegate_config.proto
```diff
@@ -0,0 +1,26 @@
...
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#extension: envoy.xds_delegates.kv_store]
+//
+// Configuration for a KeyValueStore-based XdsResourcesDelegate implementation. This implementation
+// updates the underlying KV store with xDS resources received from the configured management
+// servers, enabling configuration to be persisted locally and used on startup in case connectivity
+// with the xDS management servers could not be established.
+//
+// [#not-implemented-hide:]
```
alyssawilk, line 22: I don't think we need to hide this given it's in contrib

alyssawilk, line 22: did we want to comment the size "limitations"?

alyssawilk (commented):
LG modulo one nit!

api/contrib/envoy/extensions/config/v3alpha/kv_store_xds_delegate_config.proto
abeyad: Done

abeyad: Removed

contrib/config/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,139 @@
...
+std::vector<envoy::service::discovery::v3::Resource>
+KeyValueStoreXdsDelegate::getResources(const XdsSourceId& source_id,
+                                       const std::vector<std::string>& resource_names) const {
+  std::vector<envoy::service::discovery::v3::Resource> resources;
+  if (resource_names.empty() || (resource_names.size() == 1 && resource_names[0] == "*")) {
+    // Empty names or one entry with "*" means wildcard.
+    resources = getAllResources(source_id);
+  } else {
+    for (const std::string& resource_name : resource_names) {
+      if (auto existing_resource = xds_config_store_->get(constructKey(source_id, resource_name))) {
```
adisuissa, line 50: bit: const

contrib/config/test/kv_store_xds_delegate_integration_test.cc
```diff
@@ -0,0 +1,367 @@
...
+namespace Envoy {
+namespace {
+
+constexpr char SDS_CLUSTER_NAME[] = "sds_cluster.lyft.com";
+constexpr char RTDS_CLUSTER_NAME[] = "rtds_cluster";
+constexpr char CLIENT_CERT_NAME[] = "client_cert";
+
+std::string kvStoreDelegateConfig() {
+  const std::string filename = TestEnvironment::temporaryPath("xds_kv_store.txt");
+  ::unlink(filename.c_str());
```
adisuissa, line 21: Can this use [this unlink](https://github.com/envoyproxy/envoy/blob/main/envoy/api/os_sys_calls.h#L212)?

contrib/xds/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,130 @@
...
+  for (const auto& resource_ref : resources) {
+    const auto& decoded_resource = resource_ref.get();
+    if (decoded_resource.hasResource()) {
+      envoy::service::discovery::v3::Resource r;
+      // TODO(abeyad): Support dynamic parameter constraints.
+      r.set_name(decoded_resource.name());
+      r.set_version(decoded_resource.version());
+      r.mutable_resource()->PackFrom(decoded_resource.resource());
+      if (decoded_resource.ttl()) {
+        r.mutable_ttl()->CopyFrom(Protobuf::util::TimeUtil::MillisecondsToDuration(
```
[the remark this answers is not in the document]
adisuissa: Can you add a test for this?
I think there's a high-level requirements question here. If a resource has TTL of 1 minute, and it is stored in the KV store with that value, what should happen if it is read the day after? Should it still be 1 minute, or should it be discarded.

contrib/config/test/kv_store_xds_delegate_integration_test.cc
abeyad: done, and for the one below

abeyad:
The proto format failures have disappeared, with adding the allow-fully-qualified-name annotation introduced in https://github.com/envoyproxy/envoy/pull/23163

contrib/xds/source/kv_store_xds_delegate.cc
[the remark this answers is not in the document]
abeyad: I think it should be discarded (e.g. imagine cached secrets that have expired... we probably don't want to use the expired certs).  I'm open to hearing other ideas, cc @alyssawilk 

RE: a test, I'm actually adding support for setting TTLs on the KV store in a follow-up PR, so I was going to add a test then.  I wanted to keep this PR more manageable, hence adding the TTL support in a follow up PR.

contrib/config/source/kv_store_xds_delegate.cc
```diff
@@ -0,0 +1,146 @@
...
+  std::vector<envoy::service::discovery::v3::Resource> resources;
+  if (resource_names.empty() || (resource_names.size() == 1 && resource_names[0] == "*")) {
+    // Empty names or one entry with "*" means wildcard.
+    resources = getAllResources(source_id);
+  } else {
+    for (const std::string& resource_name : resource_names) {
+      const std::string resource_key = constructKey(source_id, resource_name);
+      if (const auto existing_resource = xds_config_store_->get(resource_key)) {
+        envoy::service::discovery::v3::Resource r;
+        if (!r.ParseFromString(std::string(*existing_resource))) {
```
abeyad, line 53: @adisuissa I added a check here if the resource fails to parse, including maintaining stats about it and removing the resource from the KV store if it's no longer parseable.

alyssawilk (approved):
LGTM, deferring to Adi for merge

contrib/config/source/kv_store_xds_delegate.cc
abeyad: I pushed a commit that shows via a test that we handle invalid protos (e.g. proto enum value no longer valid) appropriately.

abeyad:
@adisuissa friendly ping; I pushed up a commit containing an integration test that shows we handle invalid proto enums properly.

source/common/config/grpc_mux_impl.cc
```diff
@@ -107,6 +107,53 @@ void GrpcMuxImpl::sendDiscoveryRequest(absl::string_view type_url) {
...
+          std::make_unique<DecodedResourceImpl>(resource_decoder, resource));
+    }
+
+    processDiscoveryResources(decoded_resources, api_state, type_url, version_info,
+                              /*call_delegate=*/false);
+  }
+  END_TRY
+  catch (const EnvoyException& e) {
+    // TODO(abeyad): do something more than just logging the error and incrementing a counter?
+    control_plane_stats.xds_local_load_failed_.inc();
```
adisuissa, line 151: This can probably be a call to delegate signaling that the config was rejected.
This way the delegate can free resources, if needed.

adisuissa (commented):
Thanks, minor comment, but looks good otherwise.

envoy/config/grpc_mux.h
```diff
@@ -19,6 +19,7 @@ using ScopedResume = std::unique_ptr<Cleanup>;
  */
 #define ALL_CONTROL_PLANE_STATS(COUNTER, GAUGE, TEXT_READOUT)                                      \
   COUNTER(rate_limit_enforced)                                                                     \
+  COUNTER(xds_local_load_failed)                                                                   \
```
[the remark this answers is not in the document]
abeyad: Thanks for the idea, totally agree it's a much better way.  Done!

source/common/config/grpc_mux_impl.cc
abeyad: Done

abeyad (commented):
PTAL

adisuissa (approved):
LGTM, thanks!

lizan (approved):

abeyad:
Thanks for the reviews, everyone!

PR #22924, opened 2022-08-31
upstream filters: developer docs

alyssawilk:
Adding developer docs for upstream filters.

This includes an example config, and fixing up the DualInfo to contain stats (otherwise xDS reloads cause stats leaks by not being associated with listener/cluster scope).

Risk Level: low
Testing: n/a (no upstream filters use stats)
Docs Changes: developer docs
Release Notes: n/a


alyssawilk:
@lizan ping!

alyssawilk:
ah @lizan is out.  Tossing over to @ggreenway

configs/upstream-filters.yaml
```diff
@@ -0,0 +1,76 @@
...
+      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+        explicit_http_config:
+          http2_protocol_options:
+            {}
+        http_filters:
+          - name: buffer
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.http.buffer.v3.Buffer
+              max_request_bytes: 5242880
```
soulxu, line 65: There should be codec filter as termination filter, since you add check at https://github.com/envoyproxy/envoy/pull/22918/files#diff-70f33a41a52fb3e38be32857512e0de967bf016ee43479ca1530aa86c9843790R1036-R1038  :)

soulxu: Or... you just wait for another PR merged

alyssawilk: right now this is correct, but yeah, whichever lands first the second will need a merge fix

envoy/server/factory_context.h
```diff
@@ -327,6 +327,12 @@ class UpstreamHttpFactoryContext {
    *         the server will start listening.
    */
   virtual Init::Manager& initManager() PURE;
+
+  /*
+   * @return the stats scope of the particulra context. This will last as long
```
ggreenway, line 332:
```suggestion
   * @return the stats scope of the cluster. This will last as long
```

source/docs/upstream_filters.md
```diff
@@ -0,0 +1,32 @@
+# Converting a downstream filter to be a dual filter
+
+Before coverting a filter to be an upstream filter you should do some basic
+functionality analysis. Make sure
+
+  * The filter does not use any downstream-only functionality, such as clearing
+    cached routes, or performing internal redirects.
```
ggreenway, line 7: Or anything related to downstream connection? Or things like ext_authz, which I'm not sure how to succinctly explain why it doesn't make sense on upstream. I guess the difference between incoming vs outgoing.

ggreenway (changes requested):
/wait

source/docs/upstream_filters.md
alyssawilk: you can do things related to the downstream connection.  I could imagine having different auth filters for differed clusters depending on which clusters had access to PII or some such.  One initial use case has to do with doing different header sanitization for "internal" vs" cloud" clusters, where mirroring means you can't configure the correct sanitization per-route, but you might do different sanitization based on "is this a trusted IP" on either path.

source/docs/upstream_filters.md
```diff
@@ -0,0 +1,32 @@
...
+  * Either the filter does not sendLocalReply, or you thoroughly document how
+    sendLocalReply code paths will play with hedging / retries (cut off the
+    hedge attempt, and local-reply failures won't trigger retries)
+
+Once you've done this, you're ready to convert. An example converted filter is the Envoy
+[Buffer]
+(https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/buffer/config.cc)
+[Filter](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/buffer/config.h)
+
```
suniltheta, line 16: I think the markdown might need some slight update https://github.com/alyssawilk/envoy/blob/docs/source/docs/upstream_filters.md
Rendered markdown page looks like below 👇

---
Once you've done this, you're ready to convert. An example converted filter is the Envoy
[Buffer]
(https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/buffer/config.cc)
[Filter](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/buffer/config.h)

ggreenway (approved):

PR #22927, opened 2022-08-31
main: declare MainThread in MainCommonBase in addition to MainCommon.

jmarantz:
Commit Message: Currently, the MainThread is declared in the MainCommon constructor, but not MainCommonBase. Validation mode occurs in MainCommonBase though, without MainThread, so declare it in both.

MainThread can be declared at multiple points in the same thread, so this is not a problem.

We can't *move* the declaration out of MainCommon because it needs to be declared before OptionsImpl is instantiated, so it needs to stay in both places.
Additional Description:
Risk Level: low
Testing: //test/....
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features:

### DIFF
--- a/source/exe/main_common.h
+++ b/source/exe/main_common.h
@@ -85,6 +85,8 @@ protected:
 private:
   void configureComponentLogLevels();
   void configureHotRestarter(Random::RandomGenerator& random_generator);
+
+  Thread::MainThread main_thread_;
 };
 
 // TODO(jmarantz): consider removing this class; I think it'd be more useful to


jmarantz:
@seventhscream FYI

test/exe/main_common_test.cc
```diff
@@ -104,6 +104,53 @@ TEST_P(MainCommonTest, ConstructDestructHotRestartDisabledNoInit) {
...
+          route_config:
+            name: local_route
+)";
+
+    static const char* argv[] = {"envoy-static",  "--mode",     "validate",
+                                 "--config-yaml", yaml.c_str(), nullptr};
+    Envoy::OptionsImpl options(ARRAY_SIZE(argv) - 1, argv, &MainCommon::hotRestartVersion,
+                               spdlog::level::info);
+    std::unique_ptr<Thread::Thread> thread = Thread::threadFactoryForTest().createThread([&]() {
+      MainCommonBase base(options, real_time_system, default_listener_hooks, prod_component_factory,
```
KBaichoo, line 144: Doesn't `MainCommon::run` just invoke `MainCommonBase::run` which could do the validation since it defers to `MainCommonBase`? Is there a reason I'm missing for not doing that?

KBaichoo (commented):
thanks for working on this!

You'll want to merge main to avoid the dependency failure see https://github.com/envoyproxy/envoy/pull/22935

test/exe/main_common_test.cc
```diff
@@ -104,6 +104,31 @@ TEST_P(MainCommonTest, ConstructDestructHotRestartDisabledNoInit) {
...
+// This test verifies that validation can run from MainCommonBase, from a thread
+// other than the test thread. This is a desirable calling sequence in some
+// contexts. To make this work we must declare MainThread in MainCommonBase, in
+// addition to declaring it in MainCommon. There is no harm in double-declaring.
+TEST_P(MainCommonTest, ValidateUsingMainCommonBaseOutsideTestThread) {
+  EXPECT_FALSE(Thread::MainThread::isMainThreadActive());
+  {
+    Event::TestRealTimeSystem real_time_system;
+    DefaultListenerHooks default_listener_hooks;
+    ProdComponentFactory prod_component_factory;
```
KBaichoo, line 116: can we push these down to where they're used e.g. just above the `MainCommonBase` ctor?

test/exe/main_common_test.cc
```diff
@@ -104,6 +104,31 @@ TEST_P(MainCommonTest, ConstructDestructHotRestartDisabledNoInit) {
...
+    const char* argv[] = {"envoy-static",       "--mode", "validate", "--config-path",
```
KBaichoo, line 117: remove whitespace before `"--mode"`

test/exe/main_common_test.cc
KBaichoo: Makes sense, thanks

source/exe/main_common.h
```diff
@@ -85,6 +85,8 @@ class MainCommonBase {
 private:
   void configureComponentLogLevels();
   void configureHotRestarter(Random::RandomGenerator& random_generator);
+
+  Thread::MainThread main_thread_;
```
KBaichoo, line 89: can you add a comment here why we declare it here? I think it's somewhat non-obvious

KBaichoo (commented):
looks good otherwise, thanks!

test/exe/main_common_test.cc
jmarantz: done

source/exe/main_common.h
jmarantz: done

test/exe/main_common_test.cc
jmarantz: Well, clang-format really wants those spaces in there so the `"--mode"`` lines up over the `nullptr`.

KBaichoo (approved):
lgtm, thanks

Issue #22914, opened 2022-08-31
Fine grain logging does not work with JSON log message (%j)


mkauf:
*Description*:
`%j` in the log format string is not evaluated if `--enable-fine-grain-logging` is used.

*Repro steps*:

Start envoy like this:
```
envoy --log-format %j --enable-fine-grain-logging
```

Output:
```
[...]
%j
%j
%j
%j
At least one of --config-path or --config-yaml or Options::configProto() should be non-empty
```

For comparison, logging with `%j` works if fine grain logging is not used.

Start envoy like this:
```
envoy --log-format %j
```

Output:
```
[...]
error initializing configuration '': At least one of --config-path or --config-yaml or Options::configProto() should be non-empty
exiting
At least one of --config-path or --config-yaml or Options::configProto() should be non-empty
```

mattklein123:
cc @botengyao @KBaichoo

botengyao:
I see, %j is the Envoy custom flag added by #15226. Fine-grain logger is only using the default format pattern from spd::log. Will release a PR to make them consistent.

PR #15226, opened 2021-02-27
log: Add "%j" to print message as JSON escaped string

dio:
Commit Message: This patch adds a new custom flag `%j` to the log pattern to print the actual message to log as JSON escaped string.

Additional description: This produces JSON escaped string for the `msg.payload` part only, not the whole logline. Further description of the use case: https://github.com/envoyproxy/envoy/issues/13636.
Risk Level: Low
Testing: Unit
Docs Changes: Updated
Release Notes: Added
Fixes https://github.com/envoyproxy/envoy/issues/15020

Signed-off-by: Dhi Aurrahman <dio@rockybars.com>

PR #22931, opened 2022-08-31
logger: add %j and %_ support for fine-grain logger

botengyao:

`%j` and `%_` are the Envoy custom formatter added by #15226. Fine-grain logger is only using the default format pattern from spdlog. This PR makes the default and fine-grain logger consistent.

Issue: #22914

Commit Message:
Additional Description:
Risk Level: Low
Testing: bazel-bin/source/exe/envoy-static --log-format %j --enable-fine-grain-logging
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/common/fine_grain_logger.cc
```diff
@@ -165,12 +165,28 @@ spdlog::logger* FineGrainLogContext::createLogger(const std::string& key)
...
 
   new_logger->set_level(getLogLevel(key));
-  new_logger->set_pattern(Logger::Context::getFineGrainLogFormat());
+  setLogFormat(new_logger.get(), Logger::Context::getFineGrainLogFormat());
   new_logger->flush_on(level_enum::critical);
   fine_grain_log_map_->insert(std::make_pair(key, new_logger));
   return new_logger.get();
 }
 
+void FineGrainLogContext::setLogFormat(spdlog::logger* const logger,
```
KBaichoo, line 174: Can we have this just leverage shared code with the registry? https://github.com/envoyproxy/envoy/blob/6647c3f7d22bf8ab080f3b3d6ab4fb42cbb14629/source/common/common/logger.cc#L251-L266

That way they won't get out of sync.

test/common/common/log_macros_test.cc
```diff
@@ -330,6 +330,25 @@ TEST(FineGrainLog, SetLevel) {
   EXPECT_EQ(getFineGrainLogContext().getFineGrainLogEntry(__FILE__)->level(), spdlog::level::info);
 }
 
+TEST(FineGrainLog, SetJsonEscapedFormat) {
```
KBaichoo, line 333: we should test the contents logged for escaped or have https://github.com/envoyproxy/envoy/blob/6647c3f7d22bf8ab080f3b3d6ab4fb42cbb14629/test/common/common/logger_test.cc#L117 run parameterized with a logger created from the fine grain logger. I think the second approach is better.

KBaichoo (commented):
thanks for tackling this!

I think this is worth adding a release note since this is user-facing

source/common/common/fine_grain_logger.cc
botengyao: Done! Thank you.

test/common/common/log_macros_test.cc
botengyao: Done. Thank you!

source/common/common/logger.cc
```diff
@@ -240,31 +240,35 @@ std::vector<Logger>& Registry::allLoggers() {
   return *all_loggers;
 }
 
-spdlog::logger& Registry::getLog(Id id) { return *allLoggers()[static_cast<int>(id)].logger_; }
+spdlog::logger& Registry::getLog(Id id) { return *allLoggers()[static_cast<int>(id)].getLogger(); }
```
KBaichoo, line 243: With the old interface we didn't create a temporary shared pointer (e.g. atomic ref count increment), but now we do :( and we end up throwing it away. Can we implement this without adding this overhead?

test/common/common/logger_test.cc
```diff
@@ -130,29 +130,33 @@ class LoggerCustomFlagsTest : public testing::Test {
...
               testing::internal::GetCapturedStderr());
   }
 
 protected:
-  DelegatingLogSinkSharedPtr logger_;
+  spdlog::logger* logger_;
 };
 
-TEST_F(LoggerCustomFlagsTest, LogMessageAsIs) {
+INSTANTIATE_TEST_SUITE_P(EnvoyAndFineGrainLoggerFormatTest, LoggerCustomFlagsTest,
```
KBaichoo, line 145: nice!

source/common/common/logger.h
```diff
@@ -343,6 +343,11 @@ class Registry {
    */
   static void setLogFormat(const std::string& log_format);
 
+  /**
+   * Sets the log format for a specific logger.
+   */
+  static void setLogFormatForLogger(spdlog::logger* const logger, const std::string& log_format);
```
KBaichoo, line 349: can we either update these to be `setLogFormatForComponentLoggers`, same for setLogLevel or move `setLogFormatForLogger` elsewhere since that object focuses more on Component level logging, preferring the latter

KBaichoo (commented):
Thanks for working on this, looks much better

source/common/common/logger.cc
botengyao: Good point. I am trying to return the reference itself, can you take a look now? Thank you!

source/common/common/logger.h
botengyao: This is like a util function, move it out.

source/common/common/logger.cc
```diff
@@ -276,6 +265,22 @@ Logger* Registry::logger(const std::string& log_name) {
   return logger_to_return;
 }
 
+void LoggerUtil::setLogFormatForLogger(spdlog::logger* const logger,
```
KBaichoo, line 268: if the loggers are going to be non-null, thoughts on making this `spdlog::logger&`?

KBaichoo (commented):
looks good otherwise, thanks!

source/common/common/logger.h
```diff
@@ -362,6 +362,17 @@ class Registry {
   static std::vector<Logger>& allLoggers();
 };
 
+/**
+ * Defines a set of functions shared by Envoy and Find-grain loggers.
+ */
+class LoggerUtil {
```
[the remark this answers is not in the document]
botengyao: Done.

KBaichoo (approved):
lgtm pending CI, thanks

botengyao:
Can we get this merged? Thank you!

ggreenway (approved):

PR #22932, opened 2022-09-01
Add translation of the legacy config into default header validator config

yanavlasov:
Additional Description:
This enables the default header validator for all tests that use HCM (i.e. integration tests).

After merging this PR, this change can then be merged into #22537 to verify default HV effect on integration tests.

Risk Level: Low, UHV enabled build only
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A


test/extensions/filters/network/http_connection_manager/config_test.cc
```diff
@@ -2596,6 +2598,43 @@ class TestHeaderValidatorFactoryConfig : public Http::HeaderValidatorFactoryConf
   }
 };
 
+// Override the default config factory for the proto that is expected to be created when
+// header validator config was not specified.
```
alyssawilk, line 2602: wait, why is the default factory not the default?

source/extensions/filters/network/http_connection_manager/config.cc
```diff
@@ -133,27 +134,51 @@ createHeaderValidatorFactory([[maybe_unused]] const envoy::extensions::filters::
...
-  if (config.has_typed_header_validation_config()) {
 #ifdef ENVOY_ENABLE_UHV
-    auto* factory = Envoy::Config::Utility::getFactory<Http::HeaderValidatorFactoryConfig>(
-        config.typed_header_validation_config());
-    if (!factory) {
-      throw EnvoyException(fmt::format("Header validator extension not found: '{}'",
-                                       config.typed_header_validation_config().name()));
-    }
+  ::envoy::config::core::v3::TypedExtensionConfig legacy_header_validator_config;
+  if (!config.has_typed_header_validation_config()) {
```
alyssawilk, line 139: comment here as well?
something along the lines of "if using UHV with no configuration, ensure the defaults match Envoy's original behavior"

alyssawilk (commented):
sweet, thanks!
/wait

source/extensions/filters/network/http_connection_manager/config.cc
yanavlasov: Done

test/extensions/filters/network/http_connection_manager/config_test.cc
yanavlasov: This override is for the test to see what UHV config proto is created by HCM factory by default.

alyssawilk (approved):

PR #22934, opened 2022-09-01
gcp_authn: Override any existing value in the request header that includes authn: Bearer id token

tyxia:
Issue: If an incoming request includes an existing Authorization token, we should override instead of appending to the existing value.
Fix: Use `setCopy` instead of `addCopy` for the override behavior.


tyxia:
/assign @yanavlasov 

Could you please take a look? Thanks!

tyxia:
@yanavlasov  The test coverage is added and  PTAL. Thanks!

yanavlasov (approved):

PR #22948, opened 2022-09-01
http: not buffering data for CONNECT when mirror polices are specified

### LINKED ISSUE #22736: Does request_mirror_policies work for CONNECT requests?
The title says it... i was wondering if this might work since i didnt find anything in the docs..
In my setup it seems like when i enable it on a route with connect_matcher in debug logs it only makes upstream request to the primary cluster but nothing to the secondary cluster under request_mirror_policies . 
( I tried both with terminating the connect or proxying the connect on that route. )

soulxu:

Commit Message: http: not buffering data for CONNECT when mirror polices are specified
Additional Description:
Risk Level: low
Testing: unittest
Docs Changes: n/a
Release Notes: n/a
Fixes #22736

### DIFF
--- a/api/envoy/config/route/v3/route_components.proto
+++ b/api/envoy/config/route/v3/route_components.proto
@@ -730,6 +730,10 @@ message RouteAction {
   // .. note::
   //
   //   Shadowing will not be triggered if the primary cluster does not exist.
+  //
+  // .. note::
+  //
+  //   Shadowing doesn't support Http CONNECT and upgrades.
   // [#next-free-field: 6]
   message RequestMirrorPolicy {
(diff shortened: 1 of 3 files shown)


soulxu:
> Fine by me, but I think it's also worth commenting in the API docs that CONNECT and upgrades aren't supported as that's more likely to prevent future confusion, WDYT? /wait

yea, agree that it is worth to update the API doc, let me update, thanks!

alyssawilk (approved):

soulxu:
/retest

it doesn't seems like related failure

mattklein123 (approved):

PR #22960, opened 2022-09-01
tools: Enable the proto format script to use bazelisk

abeyad:
If bazelisk is installed, tools/proto_format/proto_format.sh will use
it; if bazelisk is not installed, bazel will be used.

### DIFF
--- a/tools/proto_format/proto_format.sh
+++ b/tools/proto_format/proto_format.sh
@@ -31,9 +31,12 @@ if [[ "$1" == "freeze" ]]; then
     PROTO_SYNC_CMD="fix"
 fi
 
+bazel_or_isk=bazelisk
+command -v bazelisk &> /dev/null || bazel_or_isk=bazel
+
 # Copy back the FileDescriptorProtos that protoxform emitted to the source tree. This involves
 # pretty-printing to format with protoprint.
-bazel run "${BAZEL_BUILD_OPTIONS[@]}" \
+$bazel_or_isk run "${BAZEL_BUILD_OPTIONS[@]}" \
     --//tools/api_proto_plugin:default_type_db_target=@envoy_api//:all_protos \
     ${FREEZE_ARG} \
     //tools/proto_format:proto_sync \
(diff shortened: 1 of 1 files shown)


phlax:
@abeyad i guess the motivation here is at least partly as it is triggered by the git hooks - im wondering if we should make any change there

abeyad:
> @abeyad i guess the motivation here is at least partly as it is triggered by the git hooks - im wondering if we should make any change there

@phlax I actually ran into this issue trying to run the script on my local machine, which is why I thought this change would be convenient for whichever environment the script runs in.  Which git hooks are you referring to?

phlax:
> Which git hooks are you referring to?

the ones in the Envoy repo - partly the reason for keeping this is a script is that it is used there - the other is that it is used in CI

i guess my reluctance is that this script itself might be made to a bazel target so it doesnt feel like the right place to be checking for bazel/isk 

the common use case outside of running it directly is with the build container

mostly i run it directly - but in an env with some stuff set up already

i guess the other possibility is having something set up locally to check for bazel/isk as that seems to be more of an issue with the local env, given that so much else of the repo requires the use of bazel - that wouldnt go through this script

abeyad:
> i guess my reluctance is that this script itself might be made to a bazel target so it doesnt feel like the right place to be checking for bazel/isk

Ah ok, so I think what you're saying is that we should just run the bazel target that the script invokes directly if we're running it locally (and then we can use bazelisk instead of bazel if we want)?

> i guess the other possibility is having something set up locally to check for bazel/isk as that seems to be more of an issue with the local env, given that so much else of the repo requires the use of bazel - that wouldnt go through this script

But if we set up something locally to check for bazelisk, how would we get the scripts to use that without changing the scripts too?

phlax:
i was thinking just doing what envoy-build-tools does and making the bazelisk script available at `/usr/local/bin/bazel` or similar

abeyad:
> i was thinking just doing what envoy-build-tools does and making the bazelisk script available at `/usr/local/bin/bazel` or similar

Ah ok, that's an interesting idea.  I can give that a try.  I believe bazelisk installs the different versions of bazel that it uses in different directories than /usr/local/bin so hopefully it won't cause a conflict.

phlax:
see here https://github.com/envoyproxy/envoy-build-tools/blob/0f17a22c7902e21e5e003f5296adb628d2f16b2f/build_container/build_container_common.sh#L62-L66

abeyad:
> see here https://github.com/envoyproxy/envoy-build-tools/blob/0f17a22c7902e21e5e003f5296adb628d2f16b2f/build_container/build_container_common.sh#L62-L66

Yup, I did something similar in my environment and it works well :) 
I'm good with closing this PR if you think that's best.

PR #22965, opened 2022-09-02
grpc_transcoder: support case_insensitive_enum_parsing

qiwzhang:

Add a new config `case_insensitive_enum_parsing` to enable enum value case insensitive parsing.

Normally,  only all upper case enum values are supported. 
With this flag,  other non all upper cases will be supported too.  

To fix: https://github.com/GoogleCloudPlatform/esp-v2/issues/726

Actually implementation is https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/pull/75

Risk Level: None, guided by the flag
Testing:  Unit-test and integration test
Docs Changes:  No
Release Notes:  Yes,


phlax:
somewhat related to this - iirc the Envoy binary will accept lower case enums whereas the protobuf validator doesnt - would be good to get some more consistency in how these are handled

qiwzhang:
When envoy loading json config,  it is already setting `options.case_insensitive_enum_parsing = true;`  [here](https://github.com/envoyproxy/envoy/blob/main/source/common/protobuf/utility.cc#L263)

I did not find the protobuf validator code, it may not set it.

For grpc_transcoder, we could just default it to be on to support case insensitive.  But it may break existing customer traffic,  lower case json was rejected before, with newer versions, it starts to work.   I feel it is better to have the config, let users to turn it on based on their cases.

phlax:
seems sensible to me @markdroth could you take a look

qiwzhang:
@envoyproxy/api-shepherds  could you review this pr

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
```diff
@@ -425,6 +425,71 @@ TEST_P(GrpcJsonTranscoderIntegrationTest, QueryParams) {
       respBody);
 }
 
+TEST_P(GrpcJsonTranscoderIntegrationTest, TestEnumValueDefault) {
```
adisuissa, line 428: Please add a comment describing the tests.
Also is the "Default" in the name of the test implies default enum value, or should the test name be changed to `TestEnumValueCaseMatch` or something similar?

adisuissa (commented):
Thanks for adding the feature!
/lgtm api

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
qiwzhang: Done. Thanks

qiwzhang:
Hi @envoyproxy/dependency-shepherds  could you approve this dependency update?  Thanks

phlax (approved):
/lgtm deps

thanks for addresssing this @qiwzhang

PR #22987, opened 2022-09-05
zipkin tracing: support to make the envoy as the independent hop for tracing

### LINKED ISSUE #22944: Could I custom upstream child span's kind to CLIENT?
*Title*: Could I custom upstream child span's kind to CLIENT?

*Description*:
I'm using zipkin and enable child span with `start_child_span` in router config. I found the `kind` is `SERVER` when traffic_direction is `INBOUND`. I'm using Envoy as a gateway, so I want the inbound span to be `SERVER` and the child span is `CLIENT`. It seems it follows the http filter trace config from this PR https://github.com/envoyproxy/envoy/pull/2170/files#diff-e3584a650478334f7373c0df344c85d499b5123f5237ce2bbabaad5ee53b899eL751

Is there any way to customize this? Or maybe we should always set it to `CLIENT`(egress)?

[optional *Relevant Links*:]
This can be reproduced with the [zipkin example](https://github.com/envoyproxy/envoy/tree/main/examples/zipkin). And this is the json body for envoy-1:
```
[
  {
    "localEndpoint": {
      "port": 0,
      "serviceName": "envoy-proxy-1",
      "ipv4": "172.18.0.6"
    },
    "parentId": "4865575b60739ef9",
    "shared": true,
    "id": "99a2e42466ec967f",
    "traceId": "4865575b60739ef9",
    "annotations": [
      {
        "timestamp": 1662026117696389,
        "value": "ss"
      }
    ],
    "tags": {
      "response_flags": "-",
      "upstream_cluster": "service_cluster1",
      "peer.address": "172.18.0.4:8080",
      "upstream_cluster.name": "service_cluster1",
      "upstream_address": "172.18.0.4:8080",
      "http.protocol": "HTTP/1.1",
      "http.status_code": "200",
      "component": "proxy"
    },
    "duration": 1637,
    "name": "router service_cluster1 egress",
    "timestamp": 1662026117694751,
    // I want this to be CLIENT for egress span
    "kind": "SERVER"
  },
  {
    "duration": 1702,
    "id": "99a2e42466ec967f",
    "localEndpoint": {
      "ipv4": "172.18.0.6",
      "serviceName": "envoy-proxy-1",
      "port": 0
    },
    "annotations": [
      {
        "timestamp": 1662026117696419,
        "value": "ss"
      }
    ],
    "traceId": "4865575b60739ef9",
    "shared": true,
    "timestamp": 1662026117694651,
    "name": "routeToEnvoy1",
    "kind": "SERVER",
    "parentId": "4865575b60739ef9",
    "tags": {
      "user_agent": "curl/7.61.1",
      "http.method": "GET",
      "request_size": "0",
      "http.protocol": "HTTP/1.1",
      "upstream_cluster": "service_cluster1",
      "http.status_code": "200",
      "response_size": "37",
      "http.url": "http://127.0.0.1:10000/trace/1",
      "downstream_cluster": "-",
      "peer.address": "172.18.0.7",
      "response_flags": "-",
      "upstream_cluster.name": "service_cluster1",
      "guid:x-request-id": "79698dd0-2c31-90cc-b452-fa1c1a217ea6",
      "component": "proxy"
    }
  }
]
```
btw, If I change traffic_direction of envoy-proxy-1:10000 to `OUTBOUND`, both above spans will be CLIENT.

wbpcode:
Commit Message: zipkin tracing: support to make the envoy as the independent hop for tracing

Additional Description:

In the previous implementation of zipkin tracer, the inbound envoy will generates server span and the outbound envoy will generates client span. By this way, in the sidecar mode, envoy is treated as part of the app. But in the gateway mode, the zipkin couldn't work correctly.

This PR added a new option to make the envoy worked as independent hop for zipkin tracing. In this new implementation, envoy can work correctly in both sidecar mode and gateway mode.

And this PR will close #22944.

Risk Level: Low, zipkin only updates.
Testing: Unit tests.
Docs Changes: n/a.
Release Notes: Added


api/envoy/config/trace/v3/zipkin.proto
```diff
@@ -68,4 +68,11 @@ message ZipkinConfig {
   // Optional hostname to use when sending spans to the collector_cluster. Useful for collectors
   // that require a specific hostname. Defaults to :ref:`collector_cluster <envoy_v3_api_field_config.trace.v3.ZipkinConfig.collector_cluster>` above.
   string collector_hostname = 6;
+
+  // If this is set to true, then Envoy will be treated as independent service. Server span will be created
+  // for downstream request and client span will be created for upstream request.
+  // If this is set to true, then the
+  // :ref:`start_child_span of router <envoy_v3_api_field_extensions.filters.http.router.v3.Router.start_child_span>`
+  // SHOULD be set to true also.
```
adisuissa, line 76: Maybe add what happens if `start_child_span` is set to false?

api/envoy/config/trace/v3/zipkin.proto
```diff
@@ -68,4 +68,11 @@ message ZipkinConfig {
   // Optional hostname to use when sending spans to the collector_cluster. Useful for collectors
   // that require a specific hostname. Defaults to :ref:`collector_cluster <envoy_v3_api_field_config.trace.v3.ZipkinConfig.collector_cluster>` above.
   string collector_hostname = 6;
+
+  // If this is set to true, then Envoy will be treated as independent service. Server span will be created
```
adisuissa, line 72: Can you clarify what "independent" means in this context?

source/extensions/tracers/zipkin/tracer.cc
```diff
@@ -66,14 +68,25 @@ SpanPtr Tracer::startSpan(const Tracing::Config& config, const std::string& span
   span_ptr->setName(span_name);
 
   // Set the span's kind (client or server)
-  if (config.operationName() == Tracing::OperationName::Egress) {
-    annotation.setValue(CLIENT_SEND);
+  if (independent_proxy_) {
+    // If previous context is inner context then this span must be span created for upstream
+    // request. Client span will created for upstream request when the independent_proxy is
```
adisuissa, line 73:
```suggestion
    // request. Client span will be created for upstream request when independent_proxy is
```

source/extensions/tracers/zipkin/tracer.cc
```diff
@@ -20,12 +20,14 @@ SpanPtr Tracer::startSpan(const Tracing::Config& config, const std::string& span
   // Build the CS annotation
   Annotation cs;
   cs.setEndpoint(std::move(ep));
-  if (config.operationName() == Tracing::OperationName::Egress) {
-    cs.setValue(CLIENT_SEND);
-  } else {
+  if (independent_proxy_) {
+    // No previous context then this must be span created for downstream request. Server span will
+    // created for downstream request when the independent_proxy is set to true
```
adisuissa, line 25:
```suggestion
    // be created for downstream request when independent_proxy is set to true
```

adisuissa (commented):
Thanks for fixing this! Left a few minor comments.
I'll defer the code-review to someone that understands the zipkin tracer.

api/envoy/config/trace/v3/zipkin.proto
wbpcode: > Can you clarify what "independent" means in this context?

In the previous implementation of zipkin, sidecar envoy will be treated as part of the App. Following trace chain will be generated.

```
server span (inbound sidecar) -> APP (May generates span or not) -> client span (outbound sidecar).
```

And if the `start_child_span` is set to true, the following trace chain will be generated, **and which is wrong**.

```
server span (inbound sidecar) -> server span (router of inbound sidecar) -> APP -> client span (outbound sidecar) -> client span (router of outbound sidecar).

or

server span (gateway) -> server span (router of gateway) -> APP
```

And in our new implementation, if the `independent_proxy` is set to true, **the envoy (inbound or outbound) will be treated as independent hop** in the trace chain. Server span will be created for downstream request and client span will be created for upstream request.

If the `start_child_span` is set to true, the following trace chain will be generated.


```
server span (inbound sidecar) -> client span (router of inbound sidecar) -> APP -> server span (outbound sidecar) -> client span (router of outbound sidecar).

or

server span (gateway) -> client span (router of gateway) -> APP

```

api/envoy/config/trace/v3/zipkin.proto
wbpcode: > Maybe add what happens if `start_child_span` is set to false?

If the `independent_proxy` is set to true and  `start_child_span` is set to false, then error trace chain will be generated in the sidecar mode.

```
server span (inbound sidecar) -> APP -> server span (outbound sidecar).
```

api/envoy/config/trace/v3/zipkin.proto
```diff
@@ -68,4 +68,11 @@ message ZipkinConfig {
   // Optional hostname to use when sending spans to the collector_cluster. Useful for collectors
   // that require a specific hostname. Defaults to :ref:`collector_cluster <envoy_v3_api_field_config.trace.v3.ZipkinConfig.collector_cluster>` above.
   string collector_hostname = 6;
+
+  // If this is set to true, then Envoy will be treated as independent hop in trace chain. Server span will be created
+  // for downstream request and client span will be created for upstream request.
+  // If this is set to true, then the
+  // :ref:`start_child_span of router <envoy_v3_api_field_extensions.filters.http.router.v3.Router.start_child_span>`
+  // SHOULD be set to true also to ensure the correctness of trace chain.
+  bool independent_proxy = 7;
```
kyessenov, line 77: This flag is pretty confusing when used with operation_name and shared_span_context. Would it be better to add an e-num option to https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-enum-config-core-v3-trafficdirection instead, which I think is the source of truth for the tracer behavior?

Also, regarding naming: I think gateway or a middle proxy would be a better term here.

kyessenov (commented):
Code looks reasonable, but maybe we can simplify...

api/envoy/config/trace/v3/zipkin.proto
wbpcode: > Would it be better to add an e-num option to https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-enum-config-core-v3-trafficdirection instead, which I think is the source of truth for the tracer behavior?

If we add third value to the enum and set it, we will lose the original direction info of inbound or outbound. It may breaks some other extensions that depend on the traffic direction.

wbpcode: > Also, regarding naming: I think gateway or a middle proxy would be a better term here.

`independent_gateway` or `independent_middle_proxy`?

kyessenov: We'll probably need to do the same thing for other tracers, no? Other tracers can treat a new enum option as `UNSPECIFIED` which shouldn't break code because traffic direction was never required.

As for the naming, I'll leave it to the API reviewer. The end result is that the proxy becomes visible in the trace graph, so maybe making it imply that would be easier to understand.

api/envoy/config/trace/v3/zipkin.proto
```diff
@@ -68,4 +68,11 @@ message ZipkinConfig {
   // Optional hostname to use when sending spans to the collector_cluster. Useful for collectors
   // that require a specific hostname. Defaults to :ref:`collector_cluster <envoy_v3_api_field_config.trace.v3.ZipkinConfig.collector_cluster>` above.
   string collector_hostname = 6;
+
+  // If this is set to true, then Envoy will be treated as independent hop in trace chain. Server span will be created
+  // for downstream request and client span will be created for upstream request.
```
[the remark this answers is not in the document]
wbpcode: The server span and client span have different timestamp and meaning.

In the Zipkin tracing, server span is created when the application accepts a new request from downstream client (application acted as a server). And when the application sends a request to upstream server, a client span will be created (application acted as a client).
```
--- request ---> | server span ------ client span | --- request --->
                             application
```


In the previous implementation, in the sidecar mode, envoy sidecar is treated as part of application. So, the inbound sidecar will accept requests from downstream clients and create server spans. The outbound sidecar will send requests to upstream servers and create client spans.

```
--- request ---> | server span ------ client span | --- request --->
                 inbound sidecar      outbound sidecar
```

Here are some assumptions to ensure everything is OK:
- The inbound sidecar or outbound sidecar only create a span for a request (`start_child_span` must be set false).
- The application itself will not create other spans.


So, if the `start_child_span` is set to true, or in the gateway mode, the assumptions will be broken, and the envoy couldn't generate correct trace chain.

api/envoy/config/trace/v3/zipkin.proto
wbpcode:
```
  // Specifies the intended direction of the traffic relative to the local Envoy.
  // This property is required on Windows for listeners using the original destination filter,
  // see :ref:`Original Destination <config_listener_filters_original_dst>`.
  core.TrafficDirection traffic_direction = 16;
```

> Other tracers can treat a new enum option as UNSPECIFIED which shouldn't break code because traffic direction was never required.

I am not mean other tracers. The traffic direction is common attributes of listener, so may affects may other extensions. For example, original dst filter on windows.


> We'll probably need to do the same thing for other tracers, no?

Only tracers that have span type of client/server need this. For example, the skywalking tracer. But different from zipkin, skywalking always treats envoy as independent hop in the trace chain.

wbpcode: If it's necessary, it's ok for me to create a common proto API that could be used by all tracers for this situation. But IMO, I think we shouldn't mixed it with `traffic_direction`. 🤔

kyessenov: Ok, the original motivation for `traffic_direction` was because every telemetry extension needed this flag, and it caused a lot of redundancy in xDS. We can always revisit it later if it becomes a problem again, so it's fine to add a custom flag just for zipkin.

wbpcode:
friendly ping @adisuissa

api/envoy/config/trace/v3/zipkin.proto
```diff
@@ -68,4 +68,11 @@ message ZipkinConfig {
...
+  // If this is set to true, then the
+  // :ref:`start_child_span of router <envoy_v3_api_field_extensions.filters.http.router.v3.Router.start_child_span>`
+  // SHOULD be set to true also to ensure the correctness of trace chain.
```
adisuissa, line 76: I still don't fully grok this, I guess this is because I don't have a good understanding of tracing and zipkin.
If this is highly recommended to set the `start_child_span` to true, then let's add a note here, so a user can easily configure their proxy.

changelogs/current.yaml
```diff
@@ -160,6 +160,10 @@ new_features:
     added an alternative function signature to ``httpCall()`` with ``options`` as an argument. This allows to skip sampling the produced trace span
     by setting ``{["trace_sampled"] = false}`` as the ``options``. And this allows to return multiple header values for same header name by setting
     ``{["return_duplicate_headers"] = true}`` as the ``options``.
+- area: zipkin
+  change: |
+    added :ref:`independent_proxy <envoy_v3_api_field_config.trace.v3.ZipkinConfig.independent_proxy>` to make Envoy worked
+    as independent hop for zipkin tracing.
```
adisuissa, line 166:
```suggestion
    added :ref:`independent_proxy <envoy_v3_api_field_config.trace.v3.ZipkinConfig.independent_proxy>` to make Envoy appear
    as an independent hop for zipkin tracing.
```

api/envoy/config/trace/v3/zipkin.proto
adisuissa: It is still somewhat not clear to me and could be due to my misunderstanding of zipkin and tracing.
Is this field separating server and client spans?

wbpcode:
> High-level question: is it correct to think that this flag should be true by default? (i.e., if we were to have a runtime guard around it, will it make sense to have it set to true eventually)

IMO, it should be true. But I am afraid there will be many users will be affected. I'm not sure if many people have gotten used to the original model.

api/envoy/config/trace/v3/zipkin.proto
wbpcode: Got it. It's a little like the skywaling tracer. Becasuse spans of zipkin have type (client or server), and span should appear in pairs. So, a server span shoud be `root span` or `child span of one client span`. And `start_child_span` is necessary in that case.

wbpcode:
> It is still somewhat not clear to me and could be due to my misunderstanding of zipkin and tracing.
Is this field separating server and client spans?

In the privious implementation, the zinkin tracer create only one span (server span for inbound, client span for outbound) for a request. **The inbound proxy and outbond proxy constitute a complete span pair.**


This implementation couldn't work properly in gateway mode. And if inbound traffic routing or outbound traffic routing is disabled, the trace chain couldn't be generated correctly.


And in the new implementaion, if the `independent_proxy` is set to true (and start child span should be set to true also), the inbound proxy will create a complete span pair for a request independently. The outbound proxy will create a complete span pair indepentdently also.
Inbound proxy and outbound proxy will no longer be bound.

wbpcode:
friendly ping @adisuissa

wbpcode:
the previous assign is failed  🤣

wbpcode:
the previous assign is failed  🤣

adisuissa:
> In the privious implementation, the zinkin tracer create only one span (server span for inbound, client span for outbound) for a request. **The inbound proxy and outbond proxy constitute a complete span pair.**
> 
> This implementation couldn't work properly in gateway mode. And if inbound traffic routing or outbound traffic routing is disabled, the trace chain couldn't be generated correctly.
> 
> And in the new implementaion, if the `independent_proxy` is set to true (and start child span should be set to true also), the inbound proxy will create a complete span pair for a request independently. The outbound proxy will create a complete span pair indepentdently also. Inbound proxy and outbound proxy will no longer be bound.

Thanks for the explanation. In this case, the name of the field could probably be improved to explain this behavior.
Something in the lines of `split_in_out_spans` or something that explains what this function does would improve the usability of this feature.
I would suggest adding when the value should be set (i.e., when the proxy is used as a gateway) in the comments.

wbpcode:
> > In the privious implementation, the zinkin tracer create only one span (server span for inbound, client span for outbound) for a request. **The inbound proxy and outbond proxy constitute a complete span pair.**
> > This implementation couldn't work properly in gateway mode. And if inbound traffic routing or outbound traffic routing is disabled, the trace chain couldn't be generated correctly.
> > And in the new implementaion, if the `independent_proxy` is set to true (and start child span should be set to true also), the inbound proxy will create a complete span pair for a request independently. The outbound proxy will create a complete span pair indepentdently also. Inbound proxy and outbound proxy will no longer be bound.
> 
> Thanks for the explanation. In this case, the name of the field could probably be improved to explain this behavior. Something in the lines of `split_in_out_spans` or something that explains what this function does would improve the usability of this feature. I would suggest adding when the value should be set (i.e., when the proxy is used as a gateway) in the comments.

Got it. I will update it tomorrow.

api/envoy/config/trace/v3/zipkin.proto
```diff
@@ -68,4 +68,19 @@ message ZipkinConfig {
   // Optional hostname to use when sending spans to the collector_cluster. Useful for collectors
   // that require a specific hostname. Defaults to :ref:`collector_cluster <envoy_v3_api_field_config.trace.v3.ZipkinConfig.collector_cluster>` above.
   string collector_hostname = 6;
+
+  // If this is set to true, then Envoy will be treated as an independent hop in trace chain. An complete span pair will be created for single
```
adisuissa, line 72:
```suggestion
  // If this is set to true, then Envoy will be treated as an independent hop in trace chain. A complete span pair will be created for single
```

adisuissa (commented):
API looks good other than a minor comment.
/assign-from @envoyproxy/envoy-maintainers

api/envoy/config/trace/v3/zipkin.proto
adisuissa, line 72:
```suggestion
  // If this is set to true, then Envoy will be treated as an independent hop in trace chain. A complete span pair will be created for a single
```

source/extensions/tracers/zipkin/tracer.cc
```diff
@@ -66,14 +68,25 @@ SpanPtr Tracer::startSpan(const Tracing::Config& config, const std::string& span
   span_ptr->setName(span_name);
 
   // Set the span's kind (client or server)
-  if (config.operationName() == Tracing::OperationName::Egress) {
-    annotation.setValue(CLIENT_SEND);
+  if (split_spans_for_request_) {
+    // If previous context is inner context then this span must be span created for upstream
+    // request. Client span will be created for upstream request when split_spans_for_request is
+    // set to true.
```
adisuissa, line 74: Can you please clarify this comment?

adisuissa (commented):
Thanks, left a minor comment.
My understanding of the zipkin tracer is limited, but code LGTM.

source/extensions/tracers/zipkin/tracer.cc
wbpcode: It's a little hard to make it clearer. 🤣 

For a single client request, envoy accept it (downstream request phase), process it (by filters), and send it to server (upstream request phase).

If the `previous_context` is inner context. Then it means the current span is **not first span** for a single request. Because the first span only has external `previous_context` or has no `previous_context`.

If current span is not first span, then this span must be a span that created for an upstream request that created by router and will be send to upstream server.

And if `split_spans_for_request` is set to true, the upstream requests' span should have type of `CLIENT_SEND`.

adisuissa, line 74:
```suggestion
    // If the previous context is an inner context then this span must be a span created for an upstream
    // request. A client span will be created for an upstream request when split_spans_for_request is
    // set to true.
```

This clause is inside the `if (split_spans_for_request_)` part, so it is a bit unclear when reading "Client span will be created for upstream request when split_spans_for_request is set to true", because it is already set to true.

wbpcode: Got it. Thanks. :)

adisuissa (approved):
LGTM, thanks!

Issue #17013, opened 2021-06-16
Load balancer: Pluggable subsetting for load balancers


conqerAtapple:
*Title*: *Load balancer: Pluggable subsetting for load balancers*

*Description*:
We would like to have a pluggable component for feeding the subset of endpoints that feeds into the  load balancer system such that it allows different subsetting schemes like deterministic subsetting. This will keep the load balancing scheme decoupled from the subsetting scheme. Without a pluggable subsetting scheme, looks like we currently only have a  cluster priority based subset.

[*Relevant Links*:]
https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/daperture-load-balancer

mattklein123:
I agree that having some type of pluggable subsetting capability would be nice. Design proposals appreciated.

conqerAtapple:
@mattklein123 i will try to write up a basic design by next week so that we can iterate over the details.

conqerAtapple:
High Level Design For Generic Subsetting 
=======

Structural
===========

Inheritance 
----------

```

                                      SubsetLoadBalancer (Abstract)
                                                   ^
					           |	
                        ---------------------------------------------------
                        |                                                  |
MetadataBasedSubsetLoadBalancer                                    DeterministicApertureSubsetLoadBalancer

```

Other Data Structures
====

RingInfo (Details TBD)
---
Contains information about metadata to calculate the position of the Envoy instance in the hash ring: 
 - Aperture size
 - Weight distribution




Flow
====

Initialization
---------
	

- thread local cluster initialization
- Initialize subset load balancer
     - MetadataBasedSubsetLoadBalancer initialization will be same as existing SubsetLoadBalancer
     - DeterministicApertureSubsetLoadBalancer will be initialized with EndpointRingInfo (TBD)


		

Refresh/Updates
-------

*MetadataBasedSubsetLoadBalancer*: No changes from existing update logic for "SubsetLoadBalancer"

*DeterministicApertureSubsetLoadBalancer*:

- On updated with set of new/updated/deleted hosts will calculate the priority/weighted subsets based on the EndpointRingInfo

conqerAtapple:
@mattklein123 @alyssawilk Added a basic design and flow design. Need to iterate on the details.

alyssawilk:
[...]

Issue #20634, opened 2022-04-01
xds: Envoy should support its built-in LB policies as extensions


markdroth:
LB extensibility was originally described in #5598.  The infrastructure for this has been implemented, but there are a few missing pieces.  Here's the relevant history:
- A new field for LB policy extensibility was added to xDS in #7744.
- The extension mechanism was added to Envoy in #17400.
- The semantics of the xDS field were changed slightly in #18419.
- xDS extension protos were added for ROUND_ROBIN and RING_HASH in #19517 and #20423, but Envoy does not yet actually implement configuring those policies as extensions.

I think the remaining work here is:
- Add extension protos for all built-in LB policies.
- Implement actually configuring these policies via the extension protos.

Once this is done, we can consider whether we want to deprecate the existing enum for configuring built-in policies.

daixiang0:
Compared with enum, I prefer LB policies.

wbpcode:
Great. This can make the envoy lb policies to be easy extended.

markdroth:
Looks like a config for the least_request policy is being added in #21210.

markdroth:
It looks like Envoy currently requires that the `common_lb_config` field is not set if the new `load_balancing_policy` field is used:

https://cs.github.com/envoyproxy/envoy/blob/6c75133dac08e6db4988d398479dc02a5dd8f11c/source/common/upstream/upstream_impl.cc#L990

I think this needs to be changed such that if the `load_balancing_policy` field is used, Envoy simply ignores the `common_lb_config` field.  Otherwise, there's no way to cleanly switch from the old field to the new field as intended in https://github.com/envoyproxy/envoy/pull/18419.

markdroth:
I don't think this is done yet.  We still need to actually implement the extensions, and figure out how to handle `CommonLbConfig`.

wbpcode:
Seems like because of this. 🤣 

![image](https://user-images.githubusercontent.com/12389633/196468536-ed007215-078b-4259-8fd7-65a4564e765c.png)

wbpcode:
Next we should make the subset lb as an extension.

alyssawilk:
curious, is the plan here to port subset lb to be a thread aware load balancer, or extend TypedLoadBalancerFactory to be able to create non-thread-aware lbs? if the latter I'd be game for helping out!

wbpcode:
@alyssawilk It would be great if you can help it. 😸 

I think both of these are out targes. The first one is a mid-term goal.
[...]

PR #17400, opened 2021-07-19
Enable load balancing policy extensions

pianiststickman:
Signed-off-by: Eugene Chan <eugenechan@google.com>

Commit Message: Enable load balancing policy extensions
Additional Description:

Enables `LOAD_BALANCING_POLICY_CONFIG` enum value in `LbPolicy` and supports typed load balancers specified in `load_balancing_policy`. Continues work done by Charlie Getzen <charliegetzenlc@gmail.com> in #15827.

Custom load balancers specified by `load_balancing_policy` are created as implementations of `ThreadAwareLoadBalancer`. Thread-local load balancers can be implemented as thread-aware load balancers that contain no logic at the thread-aware level, i.e. the purpose of the thread-aware LB is solely to contain the factory used to instantiate the thread-local LBs. (In the future it might be appropriate to provide a construct that abstracts away thread-aware aspects of `ThreadAwareLoadBalancer` for LBs that don't need to be thread-aware.)

A cluster that uses `LOAD_BALANCING_POLICY_CONFIG` may not also set a subset LB configuration. If the load balancer type makes use of subsetting, it should include a subset configuration in its own configuration message. Future work on load balancing extensions should include moving the subset LB to use load balancing extensions.

Similarly, a cluster that uses `LOAD_BALANCING_POLICY_CONFIG` may not set the `CommonLbConfig`, and it is not passed into load balancer creation (mostly owing to its dubious applicability as a top level configuration message to hierarchical load balancing policy). If the load balancer type makes use of the `CommonLbConfig`, it should include a `CommonLbConfig` in the configuration message for the load balancing policy.

Considerations for migration of existing load balancers:

- pieces of the `ThreadAwareLoadBalancerBase` implementation are specific to the built-in hashing load balancers and should be moved into a base class specifically for hashing load balancers. As it stands, custom load balancing policies are required to implement a `createLoadBalancer()` method even if the architecture of the LB policy does not require a hashing load balancer. I think we would also benefit from disentangling `ThreadAwareLoadBalancerBase` from `LoadBalancerBase`, as the former never actually does any host picking.

- as we convert existing thread-local load balancers to thread-aware load balancers, new local LBs will be re-created upon membership changes. We should provide a mechanism allowing load balancers to control whether this rebuild should occur, e.g. a callback that calls `create()` for thread-aware LBs by default, which can be overridden to do nothing for thread-local LBs.

Risk Level: low
Testing: brought up a cluster with a custom load balancer specified by `load_balancing_policy`; new unit tests included
[...]

Issue #5598, opened 2019-01-14
Load balancer extensibility


htuch:
Load balancers seems to be a natural first class extension point in Envoy, but we don't support this today. As load balancer behaviors become more complicated, e.g. witness issues such as https://github.com/envoyproxy/envoy/issues/4685, and the need for custom locality handling in Istio (CC @rshriram @costinm), it would be great to allow for LB extensions and even CDS delivery of LB behaviors via Lua/WASM.

Complications that make this challenging include the tight integration between LB and various ClusterManager data structures such as host, priority and locality data structures. We would neat a tighter, more stable and better defined API here. Also, allowing reuse of existing LBs and just minor customization, without having to reimplement the entire LB, would be useful. 

I'm opening this issue for discussion and long-term evolution of the LB implementation, help wanted.

CC @mattklein123 @snowp @cpakulski @rshriram @costinm

mattklein123:
@htuch heads up that I was discussing this with @HenryYYang today and I don't think we are going to end up needing this for the redis cluster work so Lyft won't be implementing.

htuch:
Ack; it's still on our backlog but not a super high priority.

snowp:
@markdroth I see you landed https://github.com/envoyproxy/envoy/pull/7744 a while ago that seems to tie nicely into this issue. I was gonna spend some time thinking of how to do this over the next few months, and the config approach in your PR is very similar to what I was thinking. Is this something you're actively working on?

markdroth:
@snowp My impetus for #7744 was for gRPC clients being configured via xDS, not for Envoy, so I am not personally working on the Envoy-side changes to support this.  But I do agree that this should be implemented in Envoy, and we had discussed the possibility of @htuch taking this on at some point.  But if this is something that you need sooner, I suspect he would not mind if you take this on.

No matter who does the Envoy-side work, I would be happy to consult on functionality and semantics, to make sure that things stay consistent between Envoy and gRPC.

htuch:
@markdroth yeah, @snowp and I chatted at EnvoyCon. He is the Envoy-side domain expert in this area, so it would be awesome if he can own this for us.

snowp:
[...]

PR #22991, opened 2022-09-06
[load balancer] Introducing deterministic aperture load balancing.

conqerAtapple:
Based on the discussions in issue:https://github.com/envoyproxy/envoy/issues/17013, this change introduces implementation of deterministic aperture load balancing algorithm. The implementation matches the original finagle implementation (`https://github.com/twitter/finagle/tree/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer`).
    

Risk Level: Medium
Testing: Unit/Integration, Mannual
Docs Changes: TBD
Release Notes: TBD
API Changes: TBD


conqerAtapple:
This is a draft based on https://github.com/envoyproxy/envoy/issues/17013.

conqerAtapple:
@markdroth @zuercher  PTAL. This is an initial draft based on discussions in https://github.com/envoyproxy/envoy/issues/17013. 

Also, looks like the.broken build is unrelated to the PR?

conqerAtapple:
@zuercher I think at this point the CI errors are coverage related which is expected as this is WIP/Draft. Could use eyes on the PR. Thanks!

ggreenway:
@zuercher assigning to you as you have expertise in related areas. Feel free to re-assign if needed. But this has sat unassigned for awhile so trying to get it some attention.

source/common/upstream/deterministic_aperture_lb.h
```diff
@@ -0,0 +1,161 @@
...
+    Ring(double offset, double width, const NormalizedHostWeightVector& normalized_host_weights,
+         double min_normalized_weight, uint64_t min_ring_size, uint64_t max_ring_size,
+         HashFunction hash_function, bool use_hostname_for_hashing,
+         RingHashLoadBalancerStats ring_hash_stats, DeterministicApertureLoadBalancerStats& stats);
+
+    // ThreadAwareLoadBalancerBase::HashingLoadBalancer
+    HostConstSharedPtr chooseHost(uint64_t hash, uint32_t attempt) const override;
+
+    //
+    // Utility Ring methods. Note that they are public so that we can add unit tests easily.
```
zuercher, line 86: I think we normally prefer not to expose methods just for testing. We have cases where we mark a test as a friend class for this kind of direct testing of class internals, but I wonder if it's not sufficient to provide a mocked RNG to test the different paths.

source/common/upstream/ring_hash_lb.h
```diff
@@ -48,7 +48,9 @@ class RingHashLoadBalancer : public ThreadAwareLoadBalancerBase,
 
   const RingHashLoadBalancerStats& stats() const { return stats_; }
 
-private:
+  static RingHashLoadBalancerStats generateStats(Stats::Scope& scope);
```
zuercher, line 51: Can this be moved into the protected section?

source/common/upstream/upstream_impl.h
```diff
@@ -850,6 +854,8 @@ class ClusterInfoImpl : public ClusterInfo,
   absl::optional<envoy::config::cluster::v3::Cluster::RingHashLbConfig> lb_ring_hash_config_;
   absl::optional<envoy::config::cluster::v3::Cluster::MaglevLbConfig> lb_maglev_config_;
   absl::optional<envoy::config::cluster::v3::Cluster::OriginalDstLbConfig> lb_original_dst_config_;
+  absl::optional<envoy::config::cluster::v3::Cluster::DeterministicApertureLbConfig>
+      lb_daperture_config_;
```
zuercher, line 858: Maybe just `lb_aperture_config_`? "daperture" reads like a typo everytime I see it.

source/common/upstream/deterministic_aperture_lb.cc
```diff
@@ -0,0 +1,162 @@
...
+      offset_(offset), width_(width), rng_(random_dev_()), random_distribution_(0, 1),
+      stats_(stats) {
+  unit_width_ = (1.0 / ring_size_);
+}
+
+absl::optional<size_t> DeterministicApertureLoadBalancer::Ring::getIndex(double offset) const {
+  if (offset < 0 || (offset >= unit_width_ * ring_size_)) {
+    return absl::nullopt;
+  }
+  return offset / unit_width_;
```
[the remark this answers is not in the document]
conqerAtapple: We could. But i think we will have to do a modulus operation also (`(offset * ring_size_)%ring_size`)

source/common/upstream/ring_hash_lb.h
conqerAtapple: Unfortunately `RingHashLoadBalancer::Ring` accepts a reference to `RingHashLoadBalancerStats` in its constructor. So to be able to create `DeterministicApertureLoadBalancer::Ring` from a outside context like test, we will need public access for `RingHashLoadBalancer::generateStats`

conqerAtapple:
@zuercher Thanks for reviewing and thoughtful comment. I hav addressed most of the points you raised in the comments.

source/common/upstream/deterministic_aperture_lb.cc
```diff
@@ -0,0 +1,150 @@
...
+    double start = std::fmod((offset + width), 1.0);
+
+    return 1.0 - (intersect(index_begin, index_end, start, offset)) / unit_width_;
+  }
+
+  return intersect(index_begin, index_end, offset, offset + width) / unit_width_;
+}
+
+size_t DeterministicApertureLoadBalancer::Ring::getIndex(double offset) const {
+  RELEASE_ASSERT(offset >= 0 && offset <= 1, "valid offset");
```
zuercher, line 92: I think RELEASE_ASSERT is probably a little aggressive. We normally reserve it for cases where Envoy cannot continue: out of memory, for instance.

My opinion is that we should use a normal ASSERT here and force offset into the expected range if it's out of bounds due to some corner case of rounding or floating point precision error.

source/common/upstream/deterministic_aperture_lb.cc
```diff
@@ -0,0 +1,150 @@
...
+                                                          double e1) const {
+  ENVOY_LOG(trace, "Overlap for (b0: {}, e0: {}, b1: {}, e1: {})", b0, e0, b1, e1);
+  return std::max(0.0, std::min(e0, e1) - std::max(b0, b1));
+}
+
+size_t DeterministicApertureLoadBalancer::Ring::pick() const {
+  return getIndex(std::fmod((offset_ + width_ * nextRandom()), 1.0));
+}
+
+size_t DeterministicApertureLoadBalancer::Ring::tryPickSecond(size_t first) const {
```
zuercher, line 106: Should this be `pickSecond` now that it always returns a pick?

zuercher (commented):
Generally looks good. Have you looked at the code coverage report to see that you have good coverage of the corner cases in weight and tryPickSecond?

source/common/upstream/deterministic_aperture_lb.cc
conqerAtapple: Good point @zuercher . I have updated the PR.

conqerAtapple:
@zuercher Looks like CI is happy now. Not sure why it was unhappy with coverage earlier :). Thanks for review again!

zuercher:
@markdroth I think this is ready for API review.

ggreenway:
ping @markdroth

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -117,6 +117,8 @@ message Cluster {
     // <envoy_v3_api_field_config.cluster.v3.Cluster.load_balancing_policy>` field without
     // setting any value in :ref:`lb_policy<envoy_v3_api_field_config.cluster.v3.Cluster.lb_policy>`.
     LOAD_BALANCING_POLICY_CONFIG = 7;
+
+    DETERMINISTIC_APERTURE = 8;
```
adisuissa, line 121: Please add a comment explaining this value.

adisuissa (commented):
Thanks for working on this!
From API perspective, following #5598 and #17400, I think this may better be an extension policy rather than something that is highly coupled with Envoy.
cc @htuch and @markdroth for a second opinion.

conqerAtapple:
> Thanks for working on this! From API perspective, following #5598 and #17400, I think this may better be an extension policy rather than something that is highly coupled with Envoy. cc @htuch and @markdroth for a second opinion.

Thanks for  reviewing @adisuissa . 
Couple of thoughts:
- I agree that we could add extension policy to enable adding something like this (aka custom threadaware load balancing policy). it would require more work than the scope of this PR IMHO. 
- The LB mechanism here is generic enough, especially for large scale mesh

Please let me know.

htuch:
Given the specialized nature of this and complexity, I would be in favor of LB extension. That said, this is weakly held opinion, if there are overriding practical reasons to put it in core I'll defer to other shepherds/maintainers.

conqerAtapple:
> Given the specialized nature of this and complexity, I would be in favor of LB extension. That said, this is weakly held opinion, if there are overriding practical reasons to put it in core I'll defer to other shepherds/maintainers.

Thanks @htuch. I understand.
 The thought process behind  having the the LB in this PR  as part of the core algorithms is that it compliments the existing ones like Maglev and is especially valuable in highly scaled mesh.

markdroth:
I think this should definitely be an extension.  Eventually, we want to migrate all of the built-in LB policies to extensions, as per https://github.com/envoyproxy/envoy/issues/20634.  I don't think we should introduce any new built-in policies at this point.

conqerAtapple:
> I think this should definitely be an extension. Eventually, we want to migrate all of the built-in LB policies to extensions, as per #20634. I don't think we should introduce any new built-in policies at this point.

Thanks @markdroth for review.

Let me know if there is anything I can do to get the `LB as extension` in Envoy.

conqerAtapple:
ping @markdroth . What is the next step for this? The back story is that as we are moving to Envoy, we want to have parity in our current mesh solution's LB algorithms and Envoy. At the same time since this is a generic algorithm, it would be useful to be available for all. 
I am more than happy to add this as an extension within the new API framework. Please let me know.

markdroth:
Yes, I think you should restructure this as an LB policy extension.  Envoy already has support for LB policy extensions, so all you need to do is change the API here such that this new policy is configured as an extension instead of using the legacy enum.  See some of the PRs linked in https://github.com/envoyproxy/envoy/issues/20634 for examples.

conqerAtapple:
> Yes, I think you should restructure this as an LB policy extension. Envoy already has support for LB policy extensions, so all you need to do is change the API here such that this new policy is configured as an extension instead of using the legacy enum. See some of the PRs linked in #20634 for examples.

Got it!  Was curious how would the the inbuilt LB policies be consumed by Envoy once they are converted as LB extensions policies? Where would they reside? Is there a public repository?

conqerAtapple:
> The extension can still live in this repo, just in a different path. The config will want to be in the "extensions" tree, similar to the other LB extensions described in #20634.

Thanks @markdroth . Will do.

wbpcode:
If we want make this load balancing a extension. then this will be blockered by the #23474.

Some basic method/interface updates are necessary for lb extension to make everything just work correctly.

/wait

conqerAtapple:
> If we want make this load balancing a extension. then this will be blockered by the #23474.
> 
> Some basic method/interface updates are necessary for lb extension to make everything just work correctly.
> 
> /wait

Sounds good. Although i am trying to understand why is there a limitation on LB subsetting for the extension model? This load balancer fits naturally with the existing Ring hash load balancer infrastructure (which is supported by the subsetting system).

wbpcode:
> Sounds good. Although i am trying to understand why is there a limitation on LB subsetting for the extension model? This load balancer fits naturally with the existing Ring hash load balancer infrastructure (which is supported by the subsetting system).

It's a problem that we should solve in the future.

conqerAtapple:
@markdroth @wbpcode waiting on https://github.com/envoyproxy/envoy/pull/23474.

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -1066,6 +1085,9 @@ message Cluster {
 
     // Optional configuration for the RoundRobin load balancing policy.
     RoundRobinLbConfig round_robin_lb_config = 56;
+
+    // Optional configuration for the Deterministic Aperture load balancing policy.
+    DeterministicApertureLbConfig deterministic_aperture_lb_config = 57;
```
wbpcode, line 1090: All these API should be added as load balancing policy extension configuration.

conqerAtapple: Thanks for looking @wbpcode . The existing ones (least request, maglev etc) seem to be present at both places (`api/envoy/lextensions/oad_balancing_policies/..` as well as in `api/envoy/config/cluster/v3/cluster.proto`). Is that the intent for this also? If so, how do I include `RingHash` message from `api/envoy/lextensions/oad_balancing_policies/deterministic_aperture.proto`? If I do, then in the code, I will have to pass the extension API message to `https://github.com/envoyproxy/envoy/blob/main/source/common/upstream/ring_hash_lb.h#L46` which currently does not accept the extension API.

conqerAtapple: ping @wbpcode

wbpcode: > Is that the intent for this also?
It' just for back-compatibility.

wbpcode: > which currently does not accept the extension API.

Some minor refactors may be necessary to construct `RingHashLoadBalancer` from the extension API.

conqerAtapple: @wbpcode I am updating the PR to use the extension API. I have added a conversion method from  extension RingHash configuration proto to cluster RingHash proto to localize the change.

conqerAtapple:
@adisuissa @wbpcode @markdroth would need your advice on `docs`.

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -1066,6 +1067,10 @@ message Cluster {
 
     // Optional configuration for the RoundRobin load balancing policy.
     RoundRobinLbConfig round_robin_lb_config = 56;
+
+    // Optional configuration for the Deterministic Aperture load balancing policy.
+    envoy.extensions.load_balancing_policies.deterministic_aperture.v3.DeterministicApertureLbConfig
+        deterministic_aperture_lb_config = 57;
```
wbpcode, line 1073: I don't think this is the correct way to use load balancing extension. 

See https://github.com/envoyproxy/envoy/pull/23472 as a example.

We won't added new field to `Cluster` for specific load balancing policy. All load balancing extension should be configured by the `load_balancing_policy` (wrapper of `TypedExtensionConfig`).

conqerAtapple: @wbpcode  I understand what you are trying to say. Unfortunately currently the only way to get the load balancer configuration in the `Factory` is via the `ClusterInfo` object. So we will have to convert the `loadBalancingPolicy` to `DeterministicApertureLbConfig`. Is there an example for that?

conqerAtapple: @wbpcode I have made the suggested changes (from #23472). `doc` build still fails :(

conqerAtapple:
@wbpcode I think the load balancer now follows the extension pattern. `docs` build fails for some reason. It is not very obvious what the reason might be.

api/envoy/extensions/load_balancing_policies/deterministic_aperture/v3/deterministic_aperture.proto
```diff
@@ -0,0 +1,36 @@
...
+// This way, we avoid a complete mesh of connections between Envoy and backends. The participating
+// Envoy peers are placed on a ring that is conceptually laid on top of the Ring Hash of backends
+// to determine their overlaps. By using the ring overlaps as a mechanism to divide the backends,
+// the backends get divided as a fraction of their overlaps with each peer. The algorithm uses a
+// uniform random distribution to select a backend from the range of overlapping backends. This
+// along with P2C ensures uniform load distribution.
+// [#extension: envoy.load_balancing_policies]
+message DeterministicApertureLbConfig {
+  // Hash ring configuration for the internal ring for endpoints.
+  ring_hash.v3.RingHash ring_config = 1;
```
markdroth, line 29: Is this new policy ring-hash-specific?  If so, maybe instead of implementing this as a separate policy, it should just be a new config field in the existing ring-hash policy?

That having been said, I think a better approach would be to change this such that it is not actually specific to the ring_hash policy in the first place.  I haven't looked at the implementation to see how tightly coupled this is with ring_hash, but if the overall goal here is to allow the LB policy to choose from only a subset of the endpoints, then presumably the child policy here could be any policy, even something other than ring_hash.  In that case, this field could be replaced with:

```
config.cluster.v3.LoadBalancingPolicy child_policy = 1;
```

markdroth: But why is the implementation of this new subsetting policy dependent upon implementation details of the child policy in the first place?  It seems to me that this policy should be self-contained: it should determine which subset of hosts the child policy should be looking at and then delegate to the child policy to pick from that set of hosts.  It should not care how the child policy is actually doing that pick.

If we agree that this is the long-term approach that we want here, then we should definitely not bake any dependency on ring_hash into the API itself.  So I would definitely prefer to see a generic child policy config here.

I'll defer to @wbpcode in terms of the implementation, but in principle, it seems wrong to me to implement this in a way that works only with ring_hash, regardless of the API representation.  It seems like this is just unnecessarily breaking abstractions by exposing inner workings of the ring_hash policy, and it is unnecessarily constraining to provide this functionality only in conjunction with the ring_hash policy.

I would really prefer to see this implemented in a way that is completely agnostic to the details of the underlying child policy.

conqerAtapple: I think this might need some explanation :).  The `deterministic aperture` lb uses a `ring` to place its endpoints but then overlays the Envoy peers ring on top to split the endpoints `equally` among the peers. This is well explained here : https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/daperture-load-balancer.

markdroth: I understand that the algorithm here is using a ring, and I agree that it makes sense to reuse the existing ring implementation in Envoy.  The thing I'm objecting to is the way we're accomplishing that reuse.  This PR is currently doing that by tying the deterministic_aperature policy to the ring_hash policy, which (a) prevents flexibility by not allowing deterministic_aperature to be used with other child policies, even though in principle its functionality seems orthogonal to that of the child policy, and (b) breaks the abstraction of the ring_hash policy by tying another policy to its internals, which will make it harder to make changes to ring_hash in the future.

I think a cleaner way to structure this would be to refactor the ring implementation out of the ring_hash policy and into its own library, and then use that ring library in both the ring_hash policy and in the deterministic_aperature policy.  That way, you can still reuse the existing ring implementation, but nothing in the deterministic_aperature policy will have to reach inside of the ring_hash policy, and it will allow using deterministic_aperature with any child policy.

conqerAtapple: I think it is a good suggestion - refactoring the `ring` implementation to a library which can be reused. I am not sure if we want to represent the internal structure as an abstraction , atleast in its initial implementation. IMHO although theoretically it is a good idea, i think the extra complexity might not be used much. Let me know what you think.

Thanks a lot for your input!

conqerAtapple: Also, I think the Ring configuration for Deterministic Aperture would look the same as for Ring Hash balancer's Ring. So not sure how to structure this.

markdroth: > I am not sure if we want to represent the internal structure as an abstraction

I'm not sure what you mean by that.  I think the ring can be represented using an interface that looks something like this:

```
class Ring {
 public:
  struct RingEntry {
    uint64_t hash;
    Host* host;
  };

  Ring(size_t min_size, size_t max_size, HashFunc hash_func);

  // Returns the size of the ring.
  size_t size() const;

  // Returns the entry at the specified index.
  RingEntry& entry(size_t index);

  // Returns the index to use for the specified hash.
  size_t FindIndex(uint64_t hash);
};
```

> I think the Ring configuration for Deterministic Aperture would look the same as for Ring Hash balancer's Ring

I think it would be fine to just duplicate the relevant config fields from the ring_hash config into the deterministic_aperature config.

conqerAtapple: What I meant was

markdroth: > What I meant was that I am not sure if we need an abstraction of LoadBalancingPolicy to represent the `Ring` as I am not sure if we will be implementing another `LoadBalancingPolicy` for all practical purposes.

Sorry, I don't understand that part.

I agree that the ring should not be represented as a `LoadBalancingPolicy` object, if that's what you mean.  But it sounds like we are already agreeing on the rough shape of the API for the ring object.

I am also not sure what you mean by not implementing another `LoadBalancingPolicy`.  I think the deterministic_aperature LB policy will still need to be a `LoadBalancingPolicy`; it will just be a policy that filters the set of hosts to pass to the child policy and then delegates to the child policy to do the pick.

conqerAtapple: > I  agree that the ring should not be represented as a LoadBalancingPolicy object, if that's what you mean. 

Yes thats what i meant :)

I am a bit concerned that we will have two messages (one each for Deterministic Aperture and Ring Hash load balancers) to represent the same underlying configuration for Ring.  What do you think?

markdroth: I think it's fine to have two instances of the ring config fields, since there will be two ring instances.  There isn't actually any requirement that they need to be configured with the same values.  In fact, given that the two rings will actually not have the same set of hosts in them (the one in the ring_hash child policy will have only the subset of hosts selected by the parent policy), I can imagine cases where users would actually want to tune the ring sizes differently for the two instances.

conqerAtapple: What i meant was that the proto message for both would most likely be same. Since we were discussing having each load balancer having its own Ring message, I was not sure if that is ideal.

markdroth: I think it's fine to just duplicate the necessary fields in both of the LB policy configs.  From the perspective of the configs, the fact that the underlying policies happen to share ring implementations is just an implementation detail, not something that should be baked into the API.

conqerAtapple: Sounds good.  Thanks for the suggestions. 

So in essence, here is what we have concluded:
1. Refactor Ring as a lib so that it could be used by both load balancers
2. Have separate RingConfig for both load balancer (RingHash already has one)

Did i miss anything?

markdroth: I don't think we need a separate `RingConfig` message.  I think it would be fine to just copy the relevant fields from the `RingHash`  message to the new `DeterministicApertureLbConfig` message.

But otherwise, yes, this sounds good!

conqerAtapple: Thanks @markdroth .  Since the changes discussed here implies we change the existing RingHash load balancer to use the new refactored Ring implementation, I am wondering if that will increase the scope of this PR. One way to do this could be to not change the existing RingHash load balancer in this PR but have only DeterministicAperture load balancer use the new the Ring library. Then when we move the RingHash load balancer to the new extension model, change it to use the RIng library. I can give a shot at moving the RingHash load balancer to the extension model in another PR.

markdroth: I'll let @wbpcode weigh in here, but IMHO, refactoring the ring functionality out of the ring_hash LB policy really has nothing to do with converting the ring_hash LB policy to be an extension.  I think if you're going to be moving that functionality into its own library, it makes sense to change ring_hash to use that library in the same PR, rather than just duplicating the functionality and then removing the duplication later.

conqerAtapple: Got it. Thanks for the clarification 👍

conqerAtapple: @markdroth I have refactored out the `Ring` concept.  PTAL.
The `docs` build is still broken unfortunately.

api/envoy/extensions/load_balancing_policies/deterministic_aperture/v3/deterministic_aperture.proto
```diff
@@ -0,0 +1,61 @@
...
+// Deterministic Aperture load balancing is designed to conserve connections between Envoy and
+// backend instances. This is achieved by dividing the backends uniformly among participating peers.
+// This way, we avoid a complete mesh of connections between Envoy and backends. The participating
+// Envoy peers are placed on a ring that is conceptually laid on top of the Ring Hash of backends
+// to determine their overlaps. By using the ring overlaps as a mechanism to divide the backends,
+// the backends get divided as a fraction of their overlaps with each peer. The algorithm uses a
+// uniform random distribution to select a backend from the range of overlapping backends. This
+// along with P2C ensures uniform load distribution.
+// [#next-free-field: 7]
+message DeterministicApertureLbConfig {
```
markdroth, line 28: This config still needs a field for the child policy config, or else there's no way to tell it what policy to actually use to pick from amongst the selected subset of endpoints.  I think you need to add something like this:

```
config.cluster.v3.LoadBalancingPolicy child_policy = 7
    [(validate.rules).message = {required: true}];
```

markdroth (commented):
The overall structure here looks good!

I'll let @wbpcode and the Envoy maintainers review the actual code.  I have just one small comment on the API.

api/envoy/extensions/load_balancing_policies/deterministic_aperture/v3/deterministic_aperture.proto
conqerAtapple: @markdroth Thanks for looking. Not sure what you mean. The DeterministicAperture load balancer policy is itself a `LbPolicy` of type `LOAD_BALANCING_POLICY_CONFIG`.  The integration test in the PR shows the usage.

conqerAtapple: @markdroth The `policy` here is at the same level as other load balancers in Envoy. So `deterministic aperture` is able to `choose` host from the subset of endpoints. This subset is derived by the `deterministic aperture` load balancer by overlapping the `ring` of endpoints with the `ring` of participating Envoys. So in short the load balancer here is self contained (like others - aka Maglev, Ring etc). Please let me know if that makes sense.

markdroth: I understand that this is intended to be used as a top-level LB policy.  But that doesn't mean that it can't delegate to another policy for the actual pick from among the filtered subset of hosts.

What is the actual picking algorithm used here?  Is there any reason that that algorithm needs to be tied to the subsetting behavior?  If not, it seems more flexible to delegate to a child policy.

conqerAtapple: The actual picking algorithm is described in the `load_balancer.h` header. Copying the comments here:
```
Implementation:
 *      - Uses a `Hash Ring` for the endpoints.
 *      - Envoy peers are placed on a ring that is conceptually laid on top of the Ring Hash of
 * backends to determine their overlaps.
 *      - By using the ring overlaps as a mechanism to divide the
 * backends, the backends get divided as a fraction of their overlaps with each peer.
 *      - The algorithm uses a uniform random distribution to select a backend from the range of
 * overlapping backends. This along with P2C ensures uniform load distribution.
 */


```

The implementation is in `LoadBalancer::chooseHost` method. The implementation was already reviewed by @zuercher .

markdroth: Thanks, I understand the algorithm now.  But you didn't respond to my question about why that algorithm needs to be hard-coded along with the subsetting behavior.

Wouldn't the behavior be strictly more flexible if we split it up into two policies, so that users could choose whether to use the subsetting behavior independently from choosing which pick algorithm to use?  Couldn't there be cases where someone might want the uniform random distribution picking algorithm you describe without the subsetting?  And couldn't there be cases where someone might want the subsetting behavior with a picking policy other than this uniform random distribution algorithm?

Unless there is some reason that these two steps cannot work independently of each other, it seems unnecessarily inflexible to forcibly tie them together.  I don't understand why we wouldn't want to split them up.

conqerAtapple: That is a possibility but currently not in the scope. I might add that flexibility in the future but for now, simplicity is the goal. As you said, we could imagine in the future something that allows `pick` strategy as an option. Currently it is strictly p2c.

markdroth: I understand that you don't currently need the flexibility of independently selecting the subsetting behavior from the picking behavior.  But implementing this in a monolithic way now means that when we do need the flexibility in the future, Envoy will be stuck supporting this as both a monolithic policy and as separate and composable policies.  In contrast, if we just implement this as separate and composable policies from the get-go, we can address both use-cases without taking on that tech debt.

I'm not asking you to change the actual algorithm here; the behavior will still match what is described in that twitter blog post.  All I'm asking is that we to do this in a cleaner way so that we don't take on that tech debt.  I don't think this should actually be that much more work.

conqerAtapple: I could add an Enum like :

enum PickPolicy {
   P2C
}
So that we can choose the `pick` policy. Or I could add more complex structure/abstraction. Please let me know.
In reality,  I think we wont be expanding on the P2C policy mostly because this algorithm is an optimization of P2C.

markdroth: I think that using an enum and adding a more complex abstraction basically amounts to introducing a new custom extension API within the subsetting policy, which means that anyone who wants to plugin their own picking algorithm would have to write a new plugin for that extension point, and it would not be usable outside of the subsetting policy.  And the new extension API would basically have the same functionality that the existing LB policy API does, so it's not clear to me why it would be beneficial to use a different extension API here.

I think we should just use the existing LB policy API.  Writing a custom extension API seems like more work, more duplication of effort, and ultimately gives us less flexibility.

conqerAtapple: Just to make sure I understand - DeterministicAperture would be a LB policy which will host another LB policy ? Since the policy here is the picking algorithm, I was hoping that that can be encapsulated via an `Enum`.  Having the picking algorithm as a policy is also ok but i am worried it would imply a separate `typed_config` , no? Please let me know. Happy to add what works best.

markdroth: Yes, it will be a separate typed config.  Specifically, as I mentioned in my original comment, I think we want to add the following field here:

```
config.cluster.v3.LoadBalancingPolicy child_policy = 7
    [(validate.rules).message = {required: true}];
```

The typed config is inside the `LoadBalancingPolicy` message.

I think the key point here is that the top-level policy should do the subsetting behavior, and it should then delegate to a child policy to pick from among the selected subset of hosts.  I don't really care that much about the names of the two policies; if you'd rather use the name "deterministic_aperature" for the lower-level picking policy instead of the top-level subsetting policy, that's fine.

conqerAtapple: Got it. Had a question -  how do I represent "P2C" as the default policy? Do i create a factory for P2C and a separate `typed config` for it. Then i can iterate through the `policies` list in the `child_policy` in the load balancer constructor?

markdroth: Yes, I think you should create a separate factory and a separate config proto for the child policy.

I don't think the child policy should default to any particular policy; I think it makes more sense to have the config explicitly specify the child policy.

conqerAtapple: I started refactoring the code but I am finding it a bit convoluted to fit the concept of `picking` algorithm into the LoadBalancerPolicy abstraction. The problem is that there is currently `TypedLoadBalancerFactory` which is meant to create load balancers defined by `LoadBalancingPolicy`.  But the create method for this interface has parameters that are good fit for a policies who need priority set etc. All we need in the case of the deterministic aperture is a pick algorithm. In the case of deterministic aperture, the endpoints and priority sets etc are already consumed by parent Ring to create the subsets. The picking algorithm would just need the `Ring` to get the subset. This forces me to  to create a new factory for creating the child policy that `create` a child given a Ring. 
Not sure if this is what you had in mind but i am finding this a bit too complex just to achieve which piking algorithm to use.  
Also, I dont think we can use existing load balancing policies (such as Maglev or RinhHash) as child policy since they  would need their own Ring etc.

conqerAtapple: Just to clarify my proposal: Here is a `ChildPolicy` definition: 
```
// [#protodoc-title: Deterministic Aperture Child Policy]
message ChildPolicy {
    reserved 1,2,3;

    reserved "config", "name", "typed_config";

    core.v3.TypedExtensionConfig typed_extension_config = 4;
}
```
And the `ChildPolicy` interface and factory:
```
class ChildPolicy {
 public:
  virtual ~ChildPolicy() = default;
  virtual Upstream::HostConstSharedPtr chooseHost() PURE;
};

using std::unique_ptr<ChildPolicy> ChildPolicyPtr;

/**
 * Factory config for deterministic aperture load balancer's child policies.
 */
class TypedChildPolicyFactory : public Config::TypedFactory {
public:
  ~TypedChildPolicyFactory() override = default;

  std::string category() const override {
    return "envoy.load_balancing_policies.deterministic_aperture.child_policy";
  }

  /**
   * @return LoadBalancerPtr a child load balancer policy implementation.
   */
  virtual ChildPolicyPtr create(Ring* ring, Stats::ScopeSharedPtr) PURE;
};
```

`P2C` specific implementation:

```
class P2CPolicy : public ChildPolicy {
public:
  P2CPolicy(Ring*, Stats::ScopeSharedPtr);

  Upstream::HostConstSharedPtr chooseHost() override;

private:
  
  /*
   * Pick an index as part of the `p2c` algorithm.
   * Uses uniform random distribution within to pick a random index in the peer's `offset` and
   * `width`.
   */
  size_t pick() const;

  /*
   * Pick another index in the peer's `offset` and `width` range.
   * @param first Index that was already picked. The new pick cannot overlap the first pick's
   * region.
   */
  size_t pickSecond(size_t first) const;

  /*
   * Picks two indexes as per the `p2c` algorithm.
   */
  std::pair<size_t, size_t> pick2() const;

  Ring* parent_;
 
};
```

markdroth: I understand what you are proposing, but I don't think it makes any sense to create a new interface here.  The whole point of an LB policy is to provide the endpoint picking behavior across a set of endpoints, so creating a new interface to do the same thing makes no sense.  That would add needless complexity and prevent people from using any of the existing policies as child policies.

To be clear, what I am suggesting here is that the P2C child policy should create its own ring rather than reusing the ring from the parent subsetting policy.  There should not be any need to share any state between the parent and child policy except for the information that can be passed through the existing LB policy API.

conqerAtapple: I am missing something here. The P2C Policy with  its own Ring is the current state of the PR.  The current PR exposes the P2C Policy + Ring as a load balancing policy which can be used as shown in the integration test.

markdroth: What I am saying is that there should be two different rings here: one in the parent policy that does subsetting, and one in the child policy that picks the endpoint from the selected subset.  You should not need to pass the ring through the LB policy API when creating the child.

conqerAtapple: The child (P2C) would need access to the parent Ring (to ask what are my subsets). I am not sure in the case of P2C, if it needs a Ring for itself since all it needs is the section of the Ring it can pick from. The parent Ring provides API for child to ask `give me the ring subsection(subset)` so that I can apply my policy. Some  other policy might decide to create its own Ring with the subset information but that depends on that policy.

markdroth: Let's step back for a moment here.  Let's set aside the details of the specific algorithm used in the two policies and instead talk about what each of the two policies is responsible for and how they interact with each other.

The LB policy API basically gives the policy a set of hosts and then asks it to pick from that set of hosts for each request.  We want both the parent and the child policy to use that same API.

The goal of the parent policy is to decide what subset of hosts the child should pick from.  In effect, the parent policy does not actually do any picking; all it does is filter the set of hosts that it is given so that it passes only a subset of those hosts to the child policy.

The goal of the child policy is to pick from the set of hosts that it is given by its parent.  It should not matter to the child whether its parent is Envoy (i.e., it is the top-level policy) or whether its parent is another LB policy that did subsetting before delegating to it; either way, it is just responsible for picking from amongst the hosts it's been given.

I think the important properties here are:
1. Both of these policies are implemented using the existing LB policy API, not creating some new API to do the same thing.
2. There should be no communication between the two policies that cannot be expressed via the existing LB policy API.

So getting back to this specific situation, I don't think we want an API for the child to refer back to the ring in the parent policy, because that would tie the child to this parent; the child would no longer be able to be used by itself without the parent.  Instead, I think we want the parent policy to build the ring, decide which subset to use, and then construct a list of hosts in that subset.  The parent policy will then pass that list of hosts to the child in the same format as would be expected by any LB policy.

If the child policy doesn't actually need a ring for its algorithm, then it doesn't need to construct a ring.  It can just do a uniform random distribution pick from among the set of hosts that are passed to it.

KBaichoo:
/wait

For merge conflicts, API changes discussion

conqerAtapple:
@zuercher Sorry i got pulled into another task so did not get any time for this PR.

Need understanding of what @markdroth is proposing.

Thanks for the suggestion @markdroth 

I am not sure if i completely understand the two layer policy. I am not sure why subsetting LB is part of this LB policy at all. The idea of this algorithm is to provide another LB algorithm (not a subsetting algorithm). This LB algorithm serves the same purpose as of all other core algorithms (example RingHash or RoundRobin). So maybe @markdroth could help me by explaining how RoundRobin algorithm could be implemented (assuming it was not already part of the current core implementations) in terms of his proposal. This will help me grasp the proposal.

PR #23000, opened 2022-09-06
health_check: Allow external health check for upstream

hudayou:
Signed-off-by: Jacky Hu <jacky.hu@walmart.com>

health_check: Allow external health check for upstream

Added custom health check address via `<envoy_v3_api_msg_config.endpoint.v3.endpoint.healthcheckconfig>`.

Commit Message: health_check: Allow external health check for upstream
Additional Description:
Risk Level: L
Testing: unit test, integration test, manual testing
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


hudayou:
The format patch to fix is empty
Any suggestions on how to fix it?
Thanks

```
INFO: Elapsed time: 0.151s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Running command line: /build/tmp/_bazel_envoybuild/b570b5ccd0454dc9af9f65ab1833764d/external/com_github_bufbuild_buf/bin/buf lint
Failure: buf.build/cncf/xds:0492166643854ea08814f4349286f9cb: does not exist
Format check failed, try apply following patch to fix:
```

changelogs/current.yaml
```diff
@@ -36,6 +36,9 @@ minor_behavior_changes:
 - area: cache_filter
   change: |
     added a completion callback to insertHeaders and insertTrailers in cache interface. Any external cache implementation extensions will need to also add this callback, and call it on completion.
+- area: hds
```
mattklein123, line 39: nit: this should be health_check, it's not related to hds

api/envoy/config/endpoint/v3/endpoint_components.proto
```diff
@@ -44,6 +44,17 @@ message Endpoint {
...
+
+    // Optional alternative health check host address.
+    //
+    // .. attention::
+    //
+    //   The form of host address depends on the given cluster type. For STATIC or EDS,
+    //   it is expected to be a direct IP address (or something resolvable by the
+    //   specified :ref:`resolver <envoy_v3_api_field_config.core.v3.SocketAddress.resolver_name>`
+    //   in the Address). For LOGICAL or STRICT DNS, it is expected to be hostname,
+    //   and will be resolved via DNS.
```
mattklein123, line 56: This should have more testing. Please add an actual integration test in https://github.com/envoyproxy/envoy/blob/main/test/integration/health_check_integration_test.cc, and also test logical/strict DNS, etc.

source/common/upstream/upstream_impl.cc
```diff
@@ -298,16 +298,22 @@ HostDescriptionImpl::HostDescriptionImpl(
...
-          : Network::Utility::getAddressWithPort(*dest_address, health_check_config.port_value());
+  if (health_check_config.has_address()) {
+    auto address = Network::Address::resolveProtoAddress(health_check_config.address());
+    health_check_address_ =
+        port_value == 0 ? address : Network::Utility::getAddressWithPort(*address, port_value);
+  } else {
+    health_check_address_ = port_value == 0
+                                ? dest_address
+                                : Network::Utility::getAddressWithPort(*dest_address, port_value);
+  }
```
mattklein123, line 316: I think this could be factored into a utility function and shared with the logical host code above?

mattklein123 (changes requested):
Thanks for working on this. Some comments to get started.

/wait

api/envoy/config/endpoint/v3/endpoint_components.proto
hudayou: The other review comments I have addressed and tested locally.

While I was trying to add the tests for this comment, I found that in order to support hostname that resolvable via DNS
We might need to add a health check address resolve target here and similarly in logic_dns_cluster.cc
https://github.com/envoyproxy/envoy/blob/main/source/common/upstream/strict_dns_cluster.cc#L34

If the resolution comes back with multiple addresses, we ends up with multiple endpoints with different health check addresse  or the endpoints equals to a matrix of endpoint addresses with endpoint health check addresses.

Our team's use case is to allow the integration of envoy health check with our centralized health checker which has a fixed ip address.

Can we limit the address to be just a direct IP address?

Or if you prefer it to be generalized and it can be both hostname and direct IP address?

mattklein123: I would limit and it can always be relaxed later. IMO a fixed IP is not super user friendly, so you might want to support a hostname which will use only a single address or something like that, but up to you as long as it is clearly documented.

hudayou: Thanks, I chose to limit it.
I'll do a PR for a custom health check extension which requires this PR.
Then after that I'll do another PR to relax it to be both hostname and IP.

Hopefully this is acceptable for you.

health_check_integration_test is using ConfigHelper::buildStaticCluster for building the test cluster
https://github.com/envoyproxy/envoy/blob/main/test/integration/health_check_integration_test.cc#L82

So I did the change below, then the new code gets tested.
https://github.com/envoyproxy/envoy/pull/23000/files#diff-ce04c82dd2b166c6df41e514e7c253ceeb276f7db1d186ddc16f8f1841ad9496R498

hudayou:
CI failed to download required blobs, rebased to main branch
https://dev.azure.com/cncf/envoy/_build/results?buildId=116532&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec&l=96

api/envoy/config/endpoint/v3/endpoint_components.proto
```diff
@@ -44,6 +44,13 @@ message Endpoint {
     // to a non-empty value allows overriding the cluster level configuration for a specific
     // endpoint.
     string hostname = 2;
+
+    // Optional alternative health check host address.
+    //
+    // .. attention::
+    //
+    //   The form of the health check host address is expected to be a direct IP address.
```
mattklein123, line 52: Please add a test for the case where this is not a direct IP address and make sure the failure message is actionable for the user.

mattklein123 (changes requested):
Thanks generally LGTM. Can you please add an integration test for this here? https://github.com/envoyproxy/envoy/blob/main/test/integration/health_check_integration_test.cc

Thank you.

/wait

mattklein123 (approved):
Awesome, thanks!

Issue #21068, opened 2022-04-28
Priority order of Compressors (Brotli, Gzip)


barnhill:
*Title*: *Priority order of Compressors (Brotli, Gzip)*

*Description*:
When configuring envoy we would like to have Brotli prioritized over Gzip if its in the Accept-Encoding header in any order.  Is there a way to prioritize Brotli even if the header comes through with the order `gzip,br`.  Currently it will see that gzip is supported and it will just use it.  This comes into play with iOS devices as from apple their default value for this header is `gzip,deflate,br` and it will never use Brotli due to this ordering.  Any help would be appreciated.

KBaichoo:
Not yet an expert in this area, from https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/compressor_filter#config-http-filters-compressor the compression filter knows the other compression libraries available in the filter chain. 

Moreover, it seems like the filter works with quality values, so I __think__ if you added a simple filter than replaces the accept-encoding on the request side with quality value weighting, priority of compression should be applied e.g. with `br quality value > gzip quality value`

barnhill:
But if you request it with "gzip,deflate,br" as in the case of the default on iOS ... it will never use Brotli as it uses the first one in the list that it supports.  Maybe there is a way to do this with composite filters?

KBaichoo:
A cleaner way to achieve this is adding additional tie breaking logic to the filter. It would probably be added here as well as the proto: https://github.com/envoyproxy/envoy/blob/5afc6faf0b2986581f8e7d88b46b3dbf1881f57c/source/extensions/filters/http/compressor/compressor_filter.cc#L390-L398

For us to override the choice even if the quality values are equal if we prefer one compressor over the other.

barnhill:
Any ideas on moving this forward?

KBaichoo:
Hi, @barnhill I don't have bandwidth to work on this, but hopefully the pointers above should be enough for someone who wants this to implement it. Happy to review any such implementations.

barnhill:
It would be nice to label this as help wanted

giantcroc:
@barnhill Hi, I can help with it.

KBaichoo:
Fixed by https://github.com/envoyproxy/envoy/pull/23005

PR #23005, opened 2022-09-07
compression: support choosing compressor first when the q-values are same

giantcroc:
Signed-off-by: giantcroc [changran.wang@intel.com](mailto:changran.wang@intel.com)

Commit Message:
As #21068, we need to add an option to let user not order to choose specific compressor when the q-values in `Accept-Encoding` are same.
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #21068
[Optional Deprecated:]

### DIFF
--- a/api/envoy/extensions/filters/http/compressor/v3/compressor.proto
+++ b/api/envoy/extensions/filters/http/compressor/v3/compressor.proto
@@ -124,6 +124,6 @@ message Compressor {
   //    ``<stat_prefix>.compressor.<compressor_library.name>.<compressor_library_stat_prefix>.*``.
   ResponseDirectionConfig response_direction_config = 8;
 
-  // If true, chooses this compressor first to do compression when the q-values are same. 
+  // If true, chooses this compressor first to do compression when the q-values are same.
   bool choose_first = 9;
 }
--- a/api/envoy/extensions/filters/http/compressor/v3/compressor.proto
+++ b/api/envoy/extensions/filters/http/compressor/v3/compressor.proto
@@ -124,6 +124,7 @@ message Compressor {
   //    ``<stat_prefix>.compressor.<compressor_library.name>.<compressor_library_stat_prefix>.*``.
   ResponseDirectionConfig response_direction_config = 8;
 
-  // If true, chooses this compressor first to do compression when the q-values are same.
+  // If true, chooses this compressor first to do compression when the q-values in `Accept-Encoding` are same.
+  // The last compressor which enables choose_first will be chosen if multiple compressor filters in the chain have choose_first as true.
   bool choose_first = 9;
 }
--- a/api/envoy/extensions/filters/http/compressor/v3/compressor.proto
+++ b/api/envoy/extensions/filters/http/compressor/v3/compressor.proto
@@ -22,7 +22,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // Compressor :ref:`configuration overview <config_http_filters_compressor>`.
 // [#extension: envoy.filters.http.compressor]
 
-// [#next-free-field: 9]
+// [#next-free-field: 10]
 message Compressor {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.config.filter.http.compressor.v2.Compressor";
@@ -123,4 +123,7 @@ message Compressor {
   //    instead of
   //    ``<stat_prefix>.compressor.<compressor_library.name>.<compressor_library_stat_prefix>.*``.
   ResponseDirectionConfig response_direction_config = 8;
+
+  // If true, chooses this compressor first to do compression when the q-values are same. 
+  bool choose_first = 9;
 }
--- a/source/extensions/filters/http/compressor/compressor_filter.cc
+++ b/source/extensions/filters/http/compressor/compressor_filter.cc
@@ -91,7 +91,8 @@ CompressorFilterConfig::CompressorFilterConfig(
       request_direction_config_(proto_config, common_stats_prefix_, scope, runtime),
       response_direction_config_(proto_config, common_stats_prefix_, scope, runtime),
       content_encoding_(compressor_factory->contentEncoding()),
-      compressor_factory_(std::move(compressor_factory)) {}
+      compressor_factory_(std::move(compressor_factory)),
+      choose_first_(proto_config.choose_first()) {}
 
 StringUtil::CaseUnorderedSet CompressorFilterConfig::DirectionConfig::contentTypeSet(
     const Protobuf::RepeatedPtrField<std::string>& types) {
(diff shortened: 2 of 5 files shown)


api/envoy/extensions/filters/http/compressor/v3/compressor.proto
```diff
@@ -123,4 +123,7 @@ message Compressor {
   //    instead of
   //    ``<stat_prefix>.compressor.<compressor_library.name>.<compressor_library_stat_prefix>.*``.
   ResponseDirectionConfig response_direction_config = 8;
+
+  // If true, chooses this compressor first to do compression when the q-values are same.
+  bool choose_first = 9;
```
soulxu, line 128: Not expert, but few questions here:

* What happened if multiple compressor filter has choose_first as true?
* Is there a case people want to specify different `choose_first` for request and response direction?
* How about adding weight for each compressor filter instead of just first choose?

giantcroc: 1. will choose the last compressor which enables `choose_first` if multiple compressor filter has choose_first as true.
2. `choose_first` affects the decision of `Accept-Encoding` that means only response direction needs this.

giantcroc: 3.I think it is not common case because users usually prefer to use better unique compressor instead of "one gzip one brotli".

soulxu: Thanks for the reply

> 1. will choose the last compressor which enables `choose_first` if multiple compressor filter has choose_first as true.
Could you describe this behavior in the API comment?

> 2. `choose_first` affects the decision of `Accept-Encoding` that means only response direction needs this.
Got it, thanks, sorry for the silly question.
> 3.I think it is not common case because users usually prefer to use better unique compressor instead of "one gzip one brotli".
Yea, agree that weight doesn't make sense here.

test/extensions/filters/http/compressor/compressor_filter_test.cc
```diff
@@ -812,6 +812,135 @@ TEST_F(MultipleFiltersTest, UseFirstRegisteredFilterWhenWildcard) {
   EXPECT_EQ(1, stats2_.counter("test2.compressor.test2.test.header_wildcard").value());
 }
 
+class ChooseFirstTest
+    : public MultipleFiltersTest,
```
soulxu, line 816: Is there any chance we can share some code between MultipleFiltersTest or CompressorFilterTest? there are a lot of method duplicated with CompressorFilterTest. But it also seems ok to test those case like MultipleFiltersTest way, only ensure which filter is used.

test/extensions/filters/http/compressor/compressor_filter_test.cc
```diff
@@ -812,6 +812,135 @@ TEST_F(MultipleFiltersTest, UseFirstRegisteredFilterWhenWildcard) {
...
+      public testing::WithParamInterface<std::tuple<std::string, bool, std::string>> {
+protected:
+  void SetUp() override {
+    envoy::extensions::filters::http::compressor::v3::Compressor compressor;
+    TestUtility::loadFromJson(R"EOF(
+{
+  "choose_first": "true",
```
soulxu, line 823: It would be great to test multiple filter as choose_first as true case also

api/envoy/extensions/filters/http/compressor/v3/compressor.proto
giantcroc: > Thanks for the reply
> 
> > 1. will choose the last compressor which enables `choose_first` if multiple compressor filter has choose_first as true.
> 
> Could you describe this behavior in the API comment?
> 

OK, thanks!

test/extensions/filters/http/compressor/compressor_filter_test.cc
giantcroc: Yeah, I also think so, but it seems a lot of work to rewrite this test. maybe I can have another PR to do this work.

test/extensions/filters/http/compressor/compressor_filter_test.cc
giantcroc: OK, I will add this next.

test/extensions/filters/http/compressor/compressor_filter_test.cc
soulxu: Could you add a TODO comment here?

test/extensions/filters/http/compressor/compressor_filter_test.cc
soulxu: I guess you forget to add test in the update, otherwise this PR LGTM

test/extensions/filters/http/compressor/compressor_filter_test.cc
giantcroc: OK, done, thanks!

test/extensions/filters/http/compressor/compressor_filter_test.cc
giantcroc: Sorry I was on vacation before, I have added more test case, thanks!

test/extensions/filters/http/compressor/compressor_filter_test.cc
```diff
@@ -812,6 +812,147 @@ TEST_F(MultipleFiltersTest, UseFirstRegisteredFilterWhenWildcard) {
...
+  std::string response_stats_prefix_{};
+  NiceMock<Http::MockStreamEncoderFilterCallbacks> encoder_callbacks_;
+};
+
+INSTANTIATE_TEST_SUITE_P(
+    ChooseFirstTestSuite, ChooseFirstTest,
+    testing::Values(std::make_tuple("true", "false", "test1", true, "test1"),
+                    std::make_tuple("true", "false", "test2, test1", true, "test1"),
+                    std::make_tuple("true", "false", "test2;q=1, test1;q=1", true, "test1"),
+                    std::make_tuple("true", "false", "test2;q=1, test1;q=0.5", false, ""),
```
soulxu, line 934: Sorry, I missed this case from the previous review. Why we didn't let the test2 as the choice here?

Also worth to add a case for `std::make_tuple("true", "false", "test2;q=0.5, test1;q=1", false, "")`

soulxu:
LGTM, thanks!

@mattklein123 for maintainer review

source/extensions/filters/http/compressor/compressor_filter.cc
```diff
@@ -322,7 +323,7 @@ CompressorFilter::chooseEncoding(const Http::ResponseHeaderMap& headers) const {
   }
 
   // Find all compressors enabled for the filter chain.
-  std::map<std::string, uint32_t> allowed_compressors;
+  std::map<std::string, std::pair<uint32_t, bool>> allowed_compressors;
```
mattklein123, line 326: nit: can you factor out the pair into a struct with named members? It's really hard to read the code and understand what is going on. I think it would be much more readable that way. Thank you.

mattklein123 (changes requested):

/wait

source/extensions/filters/http/compressor/compressor_filter.cc
giantcroc: OK, done, thanks!

mattklein123 (approved):
Thanks!

PR #23024, opened 2022-09-07
stats: Eliminate duplicate tag-name check to allow multiple regexes to populate the same tag name.

jmarantz:
Commit Message: Currently, there can be multiple built-in regexes targeting the same tag name, and in fact there's at least one case where this occurs: 

https://github.com/envoyproxy/envoy/blob/466e78586afaeb8ecb5f92351cd5ffdee3871f49/source/common/config/well_known_names.cc#L133 
https://github.com/envoyproxy/envoy/blob/466e78586afaeb8ecb5f92351cd5ffdee3871f49/source/common/config/well_known_names.cc#L136

This change prevents a second tag value for a given name being from being extracted, to meet Prometheus' requirements.

Having two alternate ways of generating the same tag value allows them to be expressed using two distinct regexes, which are easier to understand, and possible for the infrastructure to optimize with the prefix-map. This situation also occurs with Istio/Wasm, which for reasons that elude me, generate stats with two very different syntaxes both meaning HTTP Response Code, and adds those extractors using configuration.

An alternate approach is to add complexity to the regex processing to allow matches in an ORed regex, which is a bit confusing, and results in regexes that cannot be optimized well by our current system. There is no one prefix that can be used to reduce the set of regexes that need to be evaluated against every stat, and the long regexes with captures are hard for humans to read. See https://github.com/envoyproxy/envoy/pull/22791

The disadvantage of allowing multiple regexes to generate the same tag, is that it may create more scenarios where a stats sink like Prometheus may be given multiple tags with the same name, and it would be good to get some notion that this is OK. Currently such cases would be rejected during process startup (for CLI-based tags) or during config processing.

I opened this up for review to initiate this discussion, but want to make sure various stakeholders have a chance to weigh in. Though no protobufs were changed structurally in this PR, it's kind of an API change (with .proto comments) and should probably be approved as one.

Additional Description:
Risk Level: medium
Testing: //test/...
Docs Changes: changed comments in proto file that previously indicated dups were not allowed
Release Notes:
Platform Specific Features:
Fixes: https://github.com/envoyproxy/envoy/issues/22591


jmarantz:
@zirain FYI

kyessenov:
Code LGTM. I think we need to differentiate the output with duplicate tags from having duplicate tag extractors. Collisions only happen if some metric happens to be extracted twice with the same tag (for the Istio case above, that should never happen). So it's OK to permit extractors, but we may need to harden the exporter if it the other side fails non-gracefully (e.g. prom crashes when seeing duplicate labels).

jmarantz:
I think if we need to defend Prom et al from duplicate tags we should keep a set during tag extraction. It is a bit performance-sensitive so I'd prefer not to if it's not needed.

jmarantz:
I've thought of a way to add duplicate removals that shouldn't be too expensive in terms of CPU. It would be great to hear from Prometheus experts about whether this is the right thing, or whether redundant or conflicting values would be OK to leave in.

zirain:
lgtm, tanks @jmarantz

ggreenway:
The labels in a prometheus metric MUST be unique, according to the spec.

jmarantz:
Thanks -- in that case I think the existing state risks miss-populating Prom in one case with two ways to populate HTTP_CONN_MANAGER_PREFIX and also needs to be fixed. I'll revert to draft, and make that part of this PR before re-opening.

What do you think the behavior should be? First value wins? Last value? I'm thinking last-value where we evaluate config-based tag specifiers and command-line based tag-specifiers after the built-in ones so users can override built-in behavior.

jmarantz:
Sounds good; thanks Greg.

jmarantz:
@ggreenway can you take a look at the approach? It needs a doc update and another test or two, but I wanted to see if you think this is the right path before doing all that. I think the CI failure is spurious.

In this impl, the first time we do a successful tag-extraction of a stat, for a particular tag name, that will prevent the other tag extractors for that name from even running. That means that if a stat-name matches both extractors, the value will be left in the stat name.

I expect this to generally never happen, so I think it doesn't matter that much, but I made the test sort of agnostic to the order of tag extractions, and that means several possible tag-values and tag-extracted-stat-names.

source/common/stats/tag_producer_impl.cc
```diff
@@ -73,6 +66,13 @@ int TagProducerImpl::addExtractorsMatching(absl::string_view name) {
 }
 
 void TagProducerImpl::addExtractor(TagExtractorPtr extractor) {
+  auto insertion = extractor_map_.insert(std::make_pair(extractor->name(), std::ref(*extractor)));
+  if (!insertion.second) {
+    extractor->setOtherExtractorWithSameNameExists(true);
```
jmarantz, line 71: I did a quick scale test and I think I may drop this hack -- capturing the `set<string_view>` unconditionally during tag-extraction only costs about 2% when initializing 100k clusters, and that may be measurement error.

source/common/stats/tag_producer_impl.cc
```diff
@@ -101,36 +101,41 @@ void TagProducerImpl::forEachExtractorMatching(
...
-  forEachExtractorMatching(metric_name, [&remove_characters, &tags, &tag_extraction_context](
-                                            const TagExtractorPtr& tag_extractor) {
-    tag_extractor->extractTag(tag_extraction_context, tags, remove_characters);
+  absl::flat_hash_set<absl::string_view> dup_set;
+  forEachExtractorMatching(metric_name, [&remove_characters, &tags, &tag_extraction_context,
+                                         &dup_set](const TagExtractorPtr& tag_extractor) {
+    bool other_extractor_with_same_name_exists = tag_extractor->otherExtractorWithSameNameExists();
+    if (other_extractor_with_same_name_exists &&
+        dup_set.find(tag_extractor->name()) != dup_set.end()) {
+      ENVOY_LOG_MISC(error, "Skipping duplicate tag for ", tag_extractor->name());
```
ggreenway, line 114: This could get extremely noisy/verbose. Maybe this should be one of the every-so-often log macros?

source/common/stats/tag_producer_impl.cc
```diff
@@ -101,36 +101,41 @@ void TagProducerImpl::forEachExtractorMatching(
...
   std::vector<absl::string_view> tokens;
-  forEachExtractorMatching(metric_name, [&remove_characters, &tags, &tag_extraction_context](
-                                            const TagExtractorPtr& tag_extractor) {
-    tag_extractor->extractTag(tag_extraction_context, tags, remove_characters);
+  absl::flat_hash_set<absl::string_view> dup_set;
+  forEachExtractorMatching(metric_name, [&remove_characters, &tags, &tag_extraction_context,
+                                         &dup_set](const TagExtractorPtr& tag_extractor) {
+    bool other_extractor_with_same_name_exists = tag_extractor->otherExtractorWithSameNameExists();
+    if (other_extractor_with_same_name_exists &&
+        dup_set.find(tag_extractor->name()) != dup_set.end()) {
```
ggreenway, line 113: I think this approach is good. The number of tags on a metric should always be pretty small, so searching them here shouldn't be too expensive. 

This may be faster in practice as a std::vector. IIRC for sizes less than roughly 25 the linear access pattern makes it faster.

source/common/stats/tag_extractor_impl.cc
```diff
@@ -294,5 +294,14 @@ bool TagExtractorTokensImpl::searchTags(const std::vector<absl::string_view>& in
   return pattern_index == tokens_.size() && input_index == input_tokens.size();
 }
 
+TagExtractorFixedImpl::TagExtractorFixedImpl(absl::string_view name, absl::string_view value)
+    : TagExtractorImplBase(name, value), value_(std::string(value)) {}
+
+bool TagExtractorFixedImpl::extractTag(TagExtractionContext&, std::vector<Tag>& tags,
+                                       IntervalSet<size_t>&) const {
+  addTag(tags) = value_;
```
ggreenway, line 302: the name `addTag()` makes me think of a function with a side effect. It looks weird that it returns a mutable reference that is then assigned to.

ggreenway (commented):
/wait

source/common/stats/tag_extractor_impl.cc
jmarantz: Agreed -- this was pre-existing but I'm renaming to addTagReturningValueRef.

source/common/stats/tag_producer_impl.cc
jmarantz: Actually I'm leaning toward leaving it in. It's not *that* bad a hack to save a verified 2%. I ran a bunch of "envoy-static --mode init_only on a 10k cluster static config file and it saves an average of 2%.

If you think that's not worth it I can remove it and the population of TagExtractor::otherExtractorWithSameNameExists().

source/common/stats/tag_producer_impl.cc
jmarantz: I'm not sure if find on a vector is better than absl::hash_set, but it's certainly fastest to not check at all when the extractor does not have mate of the same name :)

source/common/stats/tag_producer_impl.cc
jmarantz: I think it won't happen that often because once a stat is populated it won't be re-extracted. However I switched to the power-of-2 logger just in case.

api/envoy/config/metrics/v3/stats.proto
```diff
@@ -220,7 +220,11 @@ message TagSpecifier {
   //
   // .. note::
   //
-  //   It is invalid to specify the same tag name twice in a config.
+  //   A stat name may be spelled in such a way that it matches two different
+  //   tag extractors for the same tag name. In that case, all but one of tag
+  //   values will be dropped. It is not specified which tag value will be
+  //   retained. The extraction will only occur for one of extractors, and only
```
ggreenway, line 226: grammar: one of extractors

ggreenway (changes requested):
Mostly LGTM, aside from missing word in a comment

/wait

api/envoy/config/metrics/v3/stats.proto
jmarantz: thanks, caught & fixed another one also.

ggreenway (approved):

zirain:
ping @markdroth can you take a look at this?

jmarantz:
@adisuissa can you do an API stamp? This is just clarifying ambiguous behavior with an edited comment.

adisuissa (commented):
/lgtm api

PR #23028, opened 2022-09-07
http: changing filter interfaces to make downstream access explicit.

alyssawilk:
as a follow-up to https://github.com/envoyproxy/envoy/pull/22918, splitting downstream-only filter calls into an explicit API, so it's clear which functions are not safe for upstream filters to call.

When merging this PR, folks implementing C++ filters may have to update their filter code.   Any filter calling the following functions:
route(const Router::RouteCallback& cb)
setRoute(Router::RouteConstSharedPtr route)
clearRouteCache()
requestRouteConfigUpdate(RouteConfigUpdatedCallbackSharedPtr route_config_updated_cb)

will need to add downstreamCallbacks()-> in front of the relevant function call.

for unit tests, the same work will need to be done for direct function calls, or downstream_callbacks_ when interacting with mocks.

Risk Level: low
Testing: updated tests
Docs Changes: n/a
Release Notes: n/a?


source/common/router/router.cc
```diff
@@ -1727,7 +1727,9 @@ bool Filter::convertRequestHeadersForInternalRedirect(Http::RequestHeaderMap& do
   }
   downstream_headers.setPath(path_and_query);
 
-  callbacks_->clearRouteCache();
+  if (callbacks_->downstreamCallbacks()) {
+    callbacks_->downstreamCallbacks()->clearRouteCache();
+  }
```
mattklein123, line 1732: Isn't this guaranteed to be non-null? If not can you add a comment on why?

mattklein123 (changes requested):
Thanks makes sense. Can you also add a release note? Nice!

/wait

alyssawilk:
done.  I'd initially gone to add a release note and there really isn't an appropriate area - it's not a behavioral change but it's the closest category I guess?

mattklein123 (approved):
Thanks!

PR #23035, opened 2022-09-08
[balsa] Ignore long trailers.

bencebeky:
Ignore trailers with long key or value if
codec_settings_.enable_trailers_ is false.

This also causes trailers not to be validated if
codec_settings_.enable_trailers_ is false, which is different from http-parser behavior.

Tracking issue: #21245

Commit Message: [balsa] Ignore long trailers.
Additional Description: Ignore trailers with long key or value if codec_settings_.enable_trailers_ is false.
Risk Level: low
Docs Changes: n/a
Release Notes: n/a


test/common/http/http1/codec_impl_test.cc
```diff
@@ -928,6 +928,11 @@ TEST_P(Http1ServerConnectionImplTest, Http11InvalidRequest) {
 }
 
 TEST_P(Http1ServerConnectionImplTest, Http11InvalidTrailerPost) {
+  if (parser_impl_ == ParserImpl::BalsaParser) {
+    // BalsaParser only validates trailers if `enable_trailers_` is set.
+    codec_settings_.enable_trailers_ = true;
+  }
```
diannahu, line 934: This sounds reasonable to me and indeed seems to be more in the spirit of https://github.com/envoyproxy/envoy/blob/main/source/common/http/http1/codec_impl.cc#L765-L768.

diannahu (dismissed):
Nice! A short change to ignore long trailers :)

bencebeky:
Dianna: Thank you for the review.
Kevin: PTAL

/assign @KBaichoo

KBaichoo (commented):
Seems reasonable that we'd not validate trailers if they're not enabled. How does balsa end up cleaning up the buffer with invalid trailers?

bencebeky:
> Seems reasonable that we'd not validate trailers if they're not enabled. How does balsa end up cleaning up the buffer with invalid trailers?

BalsaFrame does not buffer the raw input, instead it copies headers (and trailers, if `trailer_` is set) into BalsaHeaders objects, and calls BalsaVisitorInterface methods with absl::string_view arguments that point into the raw input.

KBaichoo:
Adding wait as discussed out of band for a test that has balsa trailers disabled but invalid trailers, we can recover and process a subsequent well formed request

KBaichoo (approved):
lgtm, thanks

bencebeky:
Thank you.

Issue #16503, opened 2021-05-14
ORIGINAL_DST for upstream IP only, with custom port


howardjohn:
*Title*: ORIGINAL_DST for upstream IP only, with custom port

*Description*:
It would be great to allow an ORIGINAL_DST cluster that only takes the original IP address, not the port. Example use cases:
* In a mesh, I have a cluster for google.com. I want ORIGINAL_DST, not STRICT_DNS, because doing DNS lookup 2x (client and envoy) can add some overhead. However, I want the client to send requests to port 80 in plaintext, and have Envoy add TLS and direct to port 443.

Strawman API proposal:
In OriginalDstLbConfig, add `upstream_port_override int`. Usage like
```
original_dst_lb_config:
  upstream_port_override: 443
```

I could imagine more complex use cases (ex map 80->443, but port 8080->8080), but I don't think that level of complexity is needed, as it can be solved with two clusters, unlike the above use case.

cc @stono @lambdai

lambdai:
I can understand this is useful regarding the http port 80 and 443. But it doesn' make sense for tcp modulo http.

Ideally, the destination port is determined by the application layer, e.g. a [http filter rewrite the original dest](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/original_dst) including ip and port.

I am think about a powerful http original dst to support `X-forwarded-Port/X-Forwarded-Host`. 

What do you think? @howardjohn

howardjohn:
I think this is still useful. Also I do think TCP makes sense, this is protocol agnostic. Its basically just a rewrite of port... we can do it for all cluster types except ORIGINAL_DST

kfaseela:
@howardjohn shall we reopen the issue, to continue further discussion? @phlax would it be possible to get some help in triaging or assigning someone for the issue?

phlax:
i can reopen and mark as help wanted

wrt assigning the issue - that would need to be someone who wants to take it on

it also has to be agreed that the change is desirable, and this is not really within my area/remit

kfaseela:
> i can reopen and mark as help wanted
> 
> wrt assigning the issue - that would need to be someone who wants to take it on
> 
> it also has to be agreed that the change is desirable, and this is not really within my area/remit

thanks @phlax

lambdai:
Would below extended field be good? There might be security concern but it's no worse than the existing `use_http_header`
CC @htuch 
```
 message OriginalDstLbConfig {
[...]

PR #23040, opened 2022-09-08
orig dst: add support for port override

briansonnenberg:
Cherrypick of [this commit](https://github.com/istio/envoy/commit/cc9bfc0b956cf2d53bc00675fdb11e4a505d3dd2) with some slight modifications.  Adds a couple test cases and uses WKT.

Co-authored-by: Brian Sonnenberg <bsonnenberg@google.com>
Signed-off-by: Steven Landow <landow@google.com>
Signed-off-by: Brian Sonnenberg <bsonnenberg@google.com>
Signed-off-by: John Howard <howardjohn@google.com>

Commit Message:
Added an option in OriginalDstCluster type to override the original dst port with a defined port.
Additional Description:
Risk Level: Low
Testing: Added tests + all tests passing.
Docs Changes:
Release Notes: TBD
Platform Specific Features:
[Optional Fixes #Issue] [#16503](https://github.com/envoyproxy/envoy/issues/16503)


api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -546,6 +546,8 @@ message Cluster {
     // The http header to override destination address if :ref:`use_http_header <envoy_v3_api_field_config.cluster.v3.Cluster.OriginalDstLbConfig.use_http_header>`.
     // is set to true. If the value is empty, :ref:`x-envoy-original-dst-host <config_http_conn_man_headers_x-envoy-original-dst-host>` will be used.
     string http_header_name = 2;
+
+    uint32 upstream_port_override = 3;
```
howardjohn, line 550: This could probably use a bit of documentation describing what it does

source/common/upstream/original_dst_cluster.cc
```diff
@@ -39,6 +39,9 @@ HostConstSharedPtr OriginalDstCluster::LoadBalancer::chooseHost(LoadBalancerCont
         dst_host = connection->connectionInfoProvider().localAddress();
       }
     }
+    if (dst_host && port_override_ > 0) {
+      dst_host = Network::Utility::getAddressWithPort(*dst_host.get(), port_override_);
```
howardjohn, line 43: if `getAddressWithPort` returns `null` then dst_host will be set to null as well I think - maybe want to only override if its non-null?

howardjohn (dismissed):
Looks great!

source/common/upstream/original_dst_cluster.cc
briansonnenberg: If getAddressWithPort returns null then something has gone seriously wrong and I think we should fall through to line 86.

source/common/network/utility.cc
```diff
@@ -328,8 +328,9 @@ Address::InstanceConstSharedPtr Utility::getAddressWithPort(const Address::Insta
     return std::make_shared<Address::Ipv4Instance>(address.ip()->addressAsString(), port);
   case Address::IpVersion::v6:
     return std::make_shared<Address::Ipv6Instance>(address.ip()->addressAsString(), port);
+  default:
+    return nullptr;
   }
-  PANIC("not handled");
 }
```
mattklein123, line 334: Why was this changed? What case causes this to be hit now?

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -546,6 +546,10 @@ message Cluster {
     // The http header to override destination address if :ref:`use_http_header <envoy_v3_api_field_config.cluster.v3.Cluster.OriginalDstLbConfig.use_http_header>`.
     // is set to true. If the value is empty, :ref:`x-envoy-original-dst-host <config_http_conn_man_headers_x-envoy-original-dst-host>` will be used.
     string http_header_name = 2;
+
+    // The port to override for the original dst address. This port
+    // will take precedence over filter state and header override ports
+    uint32 upstream_port_override = 3;
```
mattklein123, line 552: This should be a WKT to avoid relying on 0 being not set.

mattklein123 (changes requested):
Thanks. This makes sense at a high level. A few questions to get started.

/wait

source/common/network/utility.cc
briansonnenberg: There was [some concern](https://github.com/howardjohn/envoy/commit/53a23d5a3df705338c776eb6c500d793218a0367#r74719909) that we might panic in the original PR.  I figured it's better to allow the callers to handle a nullptr than to just give up.  Fine with reverting this if we don't care.

mattklein123: Let's just revert for now.

source/common/upstream/original_dst_cluster.h
```diff
@@ -77,10 +78,14 @@ class OriginalDstCluster : public ClusterImplBase {
     const std::shared_ptr<OriginalDstCluster> parent_;
     // The optional original host provider that extracts the address from HTTP header map.
     const absl::optional<Http::LowerCaseString>& http_header_name_;
+    const bool use_port_override_;
+    const uint32_t port_override_;
```
mattklein123, line 82: use absl::optional here to avoid 2 variables.

mattklein123 (changes requested):
LGTM with small comment.

/wait

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -546,6 +546,10 @@ message Cluster {
     // The http header to override destination address if :ref:`use_http_header <envoy_v3_api_field_config.cluster.v3.Cluster.OriginalDstLbConfig.use_http_header>`.
     // is set to true. If the value is empty, :ref:`x-envoy-original-dst-host <config_http_conn_man_headers_x-envoy-original-dst-host>` will be used.
     string http_header_name = 2;
+
+    // The port to override for the original dst address. This port
+    // will take precedence over filter state and header override ports
+    google.protobuf.UInt32Value upstream_port_override = 3;
```
mattklein123, line 552: Please add a release note for this.

source/common/upstream/original_dst_cluster.h
```diff
@@ -128,6 +130,7 @@ class OriginalDstCluster : public ClusterImplBase {
   absl::Mutex host_map_lock_;
   HostMapConstSharedPtr host_map_ ABSL_GUARDED_BY(host_map_lock_);
   absl::optional<Http::LowerCaseString> http_header_name_;
+  absl::optional<uint32_t> port_override_{};
```
mattklein123, line 133: nit

```suggestion
  absl::optional<uint32_t> port_override_;
```

mattklein123 (changes requested):
Thanks LGTM with small remaining comments.

/wait

api/envoy/config/cluster/v3/cluster.proto
soulxu, line 552: We can add validation `[(validate.rules).uint32 = {lte: 65535}]` just like https://github.com/envoyproxy/envoy/blob/667f0600743104857d59f56da4ab6aeea57fc60e/api/envoy/config/core/v3/address.proto#L78

soulxu (changes requested):

mattklein123 (changes requested):
LGTM modulo remaining PGV comment, thank you.

/wait

soulxu:
thanks for the update! LGTM

soulxu (dismissed):

mattklein123 (dismissed):
Thanks!

mattklein123:
Oops sorry looks like you need a main merge.

/wait

mattklein123 (approved):
Thanks!

Issue #22555, opened 2022-08-04
Thrift Health Checker support


JuniorHsu:
This is a design doc for feedback.

We want to introduce a thrift health checker so thrift only cluster could reflect the health-ness more correctly.

It will be a custom health checker. It sends a thrift request with configed method_name and expect a success response. 
The Thrift server can respond an exception to cause an immediate health check failure.

Here's the proposed thrift.proto

```
message Thrift {
  reserved 1, 5;

  reserved "service_name", "request_headers_to_add";

  // Specifies the method name should be set to each request that is sent to
  // the local thrift host.
  string method_name = 2 [(validate.rules).string = {min_len: 1}];

  // Supplies the type of transport that the Thrift health checker should use. Note that
  // :ref:`AUTO_TRANSPORT<envoy_v3_api_enum_value_extensions.filters.network.thrift_proxy.v3.TransportType.AUTO_TRANSPORT>`.
  // is not supported.
  // [#extension-category: envoy.filters.network]
  filters.network.thrift_proxy.v3.TransportType transport = 3
      [(validate.rules).enum = {defined_only: true}];

  // Supplies the type of protocol that the Thrift health checker should use. Note that
  // :ref:`AUTO_PROTOCOL<envoy_v3_api_enum_value_extensions.filters.network.thrift_proxy.v3.ProtocolType.AUTO_PROTOCOL>`.
  // is not supported.
  // [#extension-category: envoy.filters.network]
  filters.network.thrift_proxy.v3.ProtocolType protocol = 4
      [(validate.rules).enum = {defined_only: true}];
}
```
Note that we need to specify transport and protocol since we don't have
the request body to inspect. AUTO is not a valid option for the same reason.

Unfortunately, we can't use `not_in` for the validator since there's a bug
for foreign enum which triggers the compile error
```

bazel-out/k8-dbg/bin/external/envoy_api/envoy/extensions/health_checkers/thrift/v3/thrift.pb.validate.cc:62:17: error: unknown type name 'TransportType'; did you mean 'filters::network::thrift_proxy::v3::TransportType'?
        const std::set<TransportType> _Thrift_Transport_NotInLookup = {
                       ^~~~~~~~~~~~~
                       filters::network::thrift_proxy::v3::TransportType
```

An example setting for custom_health_check as a Thrift health checker is shown below:

```
  custom_health_check:
    name: envoy.health_checkers.thrift
      typed_config:
[...]

PR #23043, opened 2022-09-08
Thrift health checker support 

JuniorHsu:
Commit Message:
Introduce a thrift health checker so thrift only cluster could reflect the health-ness more correctly.

It will be a custom health checker. It sends a thrift request with configed method_name and expect a success response.
The thrift server can respond an exception to cause an immediate health check failure.

An example setting for custom_health_check as a Thrift health checker is shown below:

```
  custom_health_check:
    name: envoy.health_checkers.thrift
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.health_checkers.thrift.v3.Thrift
      method_name: "ping",
      transport: FRAMED
      protocol: BINARY
```

Risk Level: low
Testing: unit test
Docs Changes:  docs/root/configuration/upstream/health_checkers/thrift.rst
[Optional Fixes #Issue] #22555


JuniorHsu:
The basic flow is verified by unit tests. While I worked on testing corner cases, could I have some initial feedback?

cc @zuercher @rgs1 @fishcakez @PeterL328

JuniorHsu:
Some code decisions:
(a) `UpstreamRequest` is not reusable since HC (Health Checker) creates `ClientConnection` directly but `UpstreamRequest` uses `ConnectionPool`. Therefore we implement a simple thrift client.
(b) Extract `ThriftHealthChecker::Client` for the sake of testing and possible future reuse.
(c) Inject `ClientFactory` so we can split the test to `ThriftHealthChecker::Client` and
    mock it to test `ThriftHealthChecker`
(d) Introduce` passthrough_decoder_event_handler` for DIY propose
(e) Now we don't have retry semantic response.

JuniorHsu:
@markdroth heads up that the reason we don't use `not_in` to exclude AUTO and TWITTER is a bug
for foreign enum which triggers the compile error

```
bazel-out/k8-dbg/bin/external/envoy_api/envoy/extensions/health_checkers/thrift/v3/thrift.pb.validate.cc:62:17: error: unknown type name 'TransportType'; did you mean 'filters::network::thrift_proxy::v3::TransportType'?
        const std::set<TransportType> _Thrift_Transport_NotInLookup = {
                       ^~~~~~~~~~~~~
                       filters::network::thrift_proxy::v3::TransportType
```

JuniorHsu:
Code coverage for source/extensions/health_checkers/thrift is lower than limit of 96.6 (89.3)

more boundary cases will help

source/extensions/health_checkers/thrift/client_impl.cc
```diff
@@ -0,0 +1,162 @@
...
+  parent_.onBelowWriteBufferLowWatermark();
+}
+
+Network::FilterStatus ThriftSessionCallbacks::onData(Buffer::Instance& data, bool) {
+  // Response data flow to ClientImpl and then SimpleResponseDecoder.
+  parent_.onData(data);
+  return Network::FilterStatus::StopIteration;
+}
+
+// ClientImpl
```
rgs1, line 77: drop

source/extensions/health_checkers/thrift/client_impl.h
```diff
@@ -0,0 +1,156 @@
...
+      : transport_(std::move(transport)), protocol_(std::move(protocol)),
+        decoder_(std::make_unique<Decoder>(*transport_, *protocol_, *this)) {}
+
+  // Return if the response is complete.
+  bool onData(Buffer::Instance& data);
+
+  // Check if it is a success response or not.
+  bool responseSuccess();
+
+  FilterStatus messageBegin(MessageMetadataSharedPtr metadata) override;
```
rgs1, line 47: // PassThroughDecoderEventHandler

source/extensions/health_checkers/thrift/client_impl.h
```diff
@@ -0,0 +1,156 @@
...
+class ClientImpl : public Client,
+                   public Network::ConnectionCallbacks,
+                   protected Logger::Loggable<Logger::Id::hc> {
+public:
+  ClientImpl(ClientCallback& callback, TransportType transport, ProtocolType protocol,
+             const std::string& method_name, Upstream::HostSharedPtr host, int32_t seq_id)
+      : parent_(callback), transport_(transport), protocol_(protocol), method_name_(method_name),
+        host_(host), seq_id_(seq_id) {}
+
+  void start() override;
```
rgs1, line 101: // Client

source/extensions/health_checkers/thrift/thrift.cc
```diff
@@ -0,0 +1,119 @@
...
+namespace {
+
+// Helper functions to get the correct hostname for an L7 health check.
+const std::string& getHostname(const Upstream::HostSharedPtr& host,
+                               const Upstream::ClusterInfoConstSharedPtr& cluster) {
+  if (!host->hostnameForHealthChecks().empty()) {
+    return host->hostnameForHealthChecks();
+  }
+
+  return cluster->name();
```
rgs1, line 25: nit:

```
return !host->hostnameForHealthChecks().empty() ? host->hostnameForHealthChecks() : cluster->name();
```

source/extensions/health_checkers/thrift/thrift.cc
```diff
@@ -0,0 +1,119 @@
...
+  return host_->createHealthCheckConnection(parent_.dispatcher_, parent_.transportSocketOptions(),
+                                            parent_.transportSocketMatchMetadata().get());
+}
+// Network::ConnectionCallbacks
+void ThriftHealthChecker::ThriftActiveHealthCheckSession::onEvent(Network::ConnectionEvent event) {
+  if (event == Network::ConnectionEvent::RemoteClose ||
+      event == Network::ConnectionEvent::LocalClose) {
+    ENVOY_LOG(trace, "on event close, is_local_close={} expect_close={}",
+              event == Network::ConnectionEvent::LocalClose, expect_close_);
+    // TODO: check if response is partial or complete
```
rgs1, line 107: TODO(kuochung)

rgs1 (changes requested):
Made a first pass on docs and style stuff, I'll make a 2nd pass once those are handled. Thanks KC!

docs/root/configuration/upstream/health_checkers/thrift.rst
```diff
@@ -0,0 +1,27 @@
...
+Thrift Health Checker
+=====================
+
+The Thrift health checker is a custom health checker (with :code:`envoy.health_checkers.thrift` as name)
+which checks Thrift upstream hosts. It sends a thrift request with
+:ref:`method_name <envoy_v3_api_field_extensions.health_checkers.thrift.v3.Thrift.method_name>` and expect
+a success response. The Thrift server can respond an exception to cause an immediate health check failure.
+Must specify :ref:`transport <envoy_v3_api_field_extensions.health_checkers.thrift.v3.Thrift.transport>` and
+:ref:`protocol <envoy_v3_api_field_extensions.health_checkers.thrift.v3.Thrift.protocol>` for health checker
+to generate the expected thrift request.
```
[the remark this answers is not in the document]
JuniorHsu: I keep "with :code:`envoy.health_checkers.thrift` as name)" since the doc may need some link

source/extensions/health_checkers/thrift/client.h
```diff
@@ -0,0 +1,81 @@
...
+public:
+  /**
+   * Initialize the connection.
+   */
+  virtual void start() PURE;
+
+  /**
+   * Make a simple thrift request.
+   */
+  virtual bool makeRequest() PURE;
```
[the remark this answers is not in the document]
JuniorHsu: i do sendRequest

source/extensions/health_checkers/thrift/thrift.cc
JuniorHsu: i remove this since time_out check should be enough

source/extensions/filters/network/thrift_proxy/thrift.h
```diff
@@ -22,6 +24,7 @@ enum class TransportType {
 
 /**
  * Names of available Transport implementations.
+ * TODO(kuochunghsu): rename class name.
```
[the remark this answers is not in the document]
JuniorHsu: my apologies. this is note to myself so i'll address it for your second pass.

source/extensions/filters/network/thrift_proxy/thrift.h
```diff
@@ -50,6 +53,25 @@ class TransportNameValues {
...
+private:
+  using TransportTypeMap =
+      std::map<envoy::extensions::filters::network::thrift_proxy::v3::TransportType, TransportType>;
+
+  TransportTypeMap transport_map_{
+      {envoy::extensions::filters::network::thrift_proxy::v3::AUTO_TRANSPORT, TransportType::Auto},
+      {envoy::extensions::filters::network::thrift_proxy::v3::FRAMED, TransportType::Framed},
+      {envoy::extensions::filters::network::thrift_proxy::v3::UNFRAMED, TransportType::Unframed},
+      {envoy::extensions::filters::network::thrift_proxy::v3::HEADER, TransportType::Header},
+  };
```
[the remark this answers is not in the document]
JuniorHsu: this is a `ConstSingleton` so we will break the `const-ness` if we want the use fiasco

source/extensions/filters/network/thrift_proxy/thrift.h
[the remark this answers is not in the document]
JuniorHsu: I end up introduce `ProtoUtils` so `ProtoUtils::getTransportType(thrift_config.transport())` could
do the query. Therefore `TransportNames::fromType` is not extented and don't need renamed.
Also use `switch` to avoid additional fiasco memory.

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
```diff
@@ -0,0 +1,46 @@
...
+// Thrift health checker :ref:`configuration overview <config_health_checkers_thrift>`.
+// [#extension: envoy.health_checkers.thrift]
+
+// [#next-free-field: 6]
+message Thrift {
+  reserved 1, 5;
+
+  reserved "service_name", "request_headers_to_add";
+
+  // Specifies the method name should be set to each request that is sent to
```
rgs1, line 26: Specifies the method that will be set on each request... to an upstream host.

source/extensions/health_checkers/thrift/client.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+class Client : public Event::DeferredDeletable {
+public:
+  /**
+   * Initialize the connection.
+   */
+  virtual void start() PURE;
+
+  /**
+   * Make a simple thrift request.
```
rgs1, line 42: Let's just go with: Sends the health check request.

Things could get less simple down the line :-)

source/extensions/health_checkers/thrift/client_impl.cc
```diff
@@ -0,0 +1,144 @@
...
+  return FilterStatus::Continue;
+}
+
+FilterStatus SimpleResponseDecoder::messageEnd() {
+  ENVOY_LOG(trace, "SimpleResponseDecoder messageEnd");
+  complete_ = true;
+  return FilterStatus::Continue;
+}
+
+// ThriftSessionCallbacks
```
rgs1, line 48: i don't think we need this comment

source/extensions/health_checkers/thrift/client_impl.cc
```diff
@@ -0,0 +1,144 @@
...
+void ThriftSessionCallbacks::onAboveWriteBufferHighWatermark() {
+  parent_.onAboveWriteBufferHighWatermark();
+}
+
+void ThriftSessionCallbacks::onBelowWriteBufferLowWatermark() {
+  parent_.onBelowWriteBufferLowWatermark();
+}
+
+Network::FilterStatus ThriftSessionCallbacks::onData(Buffer::Instance& data, bool) {
+  // Response data flow to ClientImpl and then SimpleResponseDecoder.
```
rgs1, line 60: ditto

source/extensions/health_checkers/thrift/client_impl.h
```diff
@@ -0,0 +1,145 @@
...
+namespace Extensions {
+namespace HealthCheckers {
+namespace ThriftHealthChecker {
+
+using namespace Envoy::Extensions::NetworkFilters;
+using namespace Envoy::Extensions::NetworkFilters::ThriftProxy;
+
+// The simple response decoder decodes the response and inform the health
+// check session if it's a success response or not.
+class SimpleResponseDecoder : public DecoderCallbacks,
```
rgs1, line 21: informs

source/extensions/health_checkers/thrift/client_impl.h
```diff
@@ -0,0 +1,145 @@
...
+  bool complete_{};
+};
+
+using SimpleResponseDecoderPtr = std::unique_ptr<SimpleResponseDecoder>;
+
+class ClientImpl;
+
+// Network::ClientConnection takes a shared pointer callback but we need a
+// unique DeferredDeletable pointer for connection management. Therefore we
+// need additional wrapper class.
```
rgs1, line 60: an additional

source/extensions/health_checkers/thrift/thrift.cc
```diff
@@ -0,0 +1,115 @@
...
+      protocol_(ProtoUtils::getProtocolType(thrift_config.protocol())),
+      client_factory_(client_factory) {
+  if (transport_ == TransportType::Auto || protocol_ == ProtocolType::Auto ||
+      protocol_ == ProtocolType::Twitter) {
+    throw EnvoyException(
+        fmt::format("Invalid thrift health check configuration: {}", thrift_config.DebugString()));
+  }
+}
+
+// ThriftActiveHealthCheckSession:
```
rgs1, line 40: i don't get why we need these comments on the method definitions, it should be enough to reference the interface on the declarations i think (and inline with the style seen on other files/modules).

rgs1 (changes requested):
Thanks for the follow-up, left a few more comments. One question:

Do we need some more docs and comments as to any potential issues if the sequence id of the HC are overlapped with data sequence ids?

cc: @fishcakez

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
JuniorHsu: adding "health check"

`Specifies the method name should be set to each health check request to an upstream host`

source/extensions/health_checkers/thrift/client.h
JuniorHsu: sure. i put "simple" since i expect this could be extend to different decoder, i.e., more usage than just health checker. but i agree "simple" is too vague. Let's go with your suggestion for now since hc is the only  use case

source/extensions/health_checkers/thrift/thrift.cc
JuniorHsu: Right this coding section style isn't popular in envoy repo so i'll remove all of them.

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
JuniorHsu: s/should/will/

JuniorHsu:
With some discussion with  @fishcakez, we should use fixed sequence id 0.

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
```diff
@@ -0,0 +1,46 @@
...
+// Thrift health checker :ref:`configuration overview <config_health_checkers_thrift>`.
+// [#extension: envoy.health_checkers.thrift]
+
+// [#next-free-field: 6]
+message Thrift {
+  reserved 1, 5;
+
+  reserved "service_name", "request_headers_to_add";
+
+  // Specifies the method name will be set to each health check request to an upstream host.
```
rgs1, line 26: Specifies the method name that will be set on each health check request dispatched to an upstream host.

docs/root/configuration/upstream/health_checkers/thrift.rst
```diff
@@ -0,0 +1,29 @@
...
+Thrift Health Checker
+=====================
+
+The Thrift Health Checker (with :code:`envoy.health_checkers.thrift` as name) uses Thrift requests,
+responses and exceptions to check upstream hosts. It sends a request with
+:ref:`method_name <envoy_v3_api_field_extensions.health_checkers.thrift.v3.Thrift.method_name>`
+and expects a success response in exchange. The upstream host can also respond with an exception to cause the check to fail.
+The :ref:`transport <envoy_v3_api_field_extensions.health_checkers.thrift.v3.Thrift.transport>` and
+:ref:`protocol <envoy_v3_api_field_extensions.health_checkers.thrift.v3.Thrift.protocol>` types to be set for each health
+check request must be configured to enable Thrift health checks. The sequence id is always 0 for each health check request.
```
rgs1, line 12: > The sequence id is always 0 for each health check request.

Is this true when reusing connections? I also don't see anything in the docs about using a new connection on every check vs reusing the connection (but maybe this is configured at an upper level)....

rgs1 (changes requested):
Looking good, one more question and a another nit.

docs/root/configuration/upstream/health_checkers/thrift.rst
JuniorHsu: > Is this true when reusing connections? 

Yes, we force the seq id to 0 and not increasing it. btw default is *reusing* connection.
```
void ThriftHealthChecker::ThriftActiveHealthCheckSession::onInterval() {
  ENVOY_LOG(trace, "ThriftActiveHealthCheckSession onInterval");
  if (!client_) {
    ENVOY_LOG(trace, "ThriftActiveHealthCheckSession construct client");
    client_ = parent_.client_factory_.create(
        *this, parent_.transport_, parent_.protocol_, parent_.method_name_, host_,
        /* health checker seq id */ 0, /* fixed_seq_id */ true);
    client_->start();
    expect_close_ = false;
  }

  client_->sendRequest();
}
```

> I also don't see anything in the docs about using a new connection on every check vs reusing the connection (but maybe this is configured at an upper level)....

Right, It's in the upper level. 
https://github.com/envoyproxy/envoy/blob/0b10b055a6e1dbdd600f9a80f8b8d803bdd8ec9c/api/envoy/config/core/v3/health_check.proto#L304

And we close the connection for each check response if we don't want to reuse
```
void ThriftHealthChecker::ThriftActiveHealthCheckSession::onResponseResult(bool is_success) {
  if (is_success) {
    handleSuccess();
  } else {
    // TODO(kuochunghsu): We might want to define retriable response.
    handleFailure(envoy::data::core::v3::ACTIVE, /* retriable */ false);
  }

  if (client_ && !parent_.reuse_connection_) {
    expect_close_ = true;
    client_->close();
  }
}
```

See also `ThriftHealthCheckerTest, NoConnectionReuse` for testing in `thrift_test.cc`, which is the only test case for *not* reusing connection

JuniorHsu:
@markdroth the implantation review is converging. would you have a chance to review api part? thanks.

JuniorHsu:
@fishcakez noticed that we don't specify transport and protocol in thrift.proto to be required since protoc-gen doesn't support `(validate.rules).enum.required`.

Fortunately we throw for AUTO_PROTOCOL and AUTO_TRANSPORT, which is 0 and the default value as well, which means that we accidentally make them required.
https://github.com/envoyproxy/envoy/blob/1fabd6decb8de8ec97d9d632c60ffb648a4c917a/api/envoy/extensions/filters/network/thrift_proxy/v3/thrift_proxy.proto#L33 

Another feedback from him was that we might able to honor `ThriftProtocolOptions` when we have a good precedence design, so transport/protocol in thrift.proto could be optional, but we could discuss it when we have a need later.

adisuissa:
@markdroth can you PTAL at the API changes?
Assigning @zuercher as code-owner.
/assign @zuercher

JuniorHsu:
Thanks Adi.
Leave a note that we've verified thrift HC in our environment.

rgs1 (dismissed):
LGTM, thanks!

Over to @zuercher for final pass & merging.

JuniorHsu:
@envoyproxy/api-shepherds any chance to move this forward? Thanks.

JuniorHsu:
Hello @htuch 
This PR has @.rgs1 and @.fishcakez reviewed the api but need one more stamp from @envoyproxy/api-shepherds 
We conduct several bike-shedding as we can see in the comment so we believe it's good to go.
We also verified the the implementation in our environment as well. 
Could you help to move this forward as we have you several thrift proto reviews?
Thank you.

source/extensions/health_checkers/thrift/utility.h
```diff
@@ -0,0 +1,33 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace HealthCheckers {
+namespace ThriftHealthChecker {
+
+namespace {
+
+static const envoy::extensions::health_checkers::thrift::v3::Thrift
+getThriftHealthCheckConfig(const envoy::config::core::v3::HealthCheck& health_check_config,
+                           ProtobufMessage::ValidationVisitor& validation_visitor) {
```
zuercher, line 20: nit: feels like this just belongs in config.cc?

source/extensions/health_checkers/thrift/thrift.cc
```diff
@@ -0,0 +1,119 @@
...
+    : HealthCheckerImplBase(cluster, config, dispatcher, runtime, api.randomGenerator(),
+                            std::move(event_logger)),
+      method_name_(thrift_config.method_name()),
+      transport_(ProtoUtils::getTransportType(thrift_config.transport())),
+      protocol_(ProtoUtils::getProtocolType(thrift_config.protocol())),
+      client_factory_(client_factory) {
+  if (transport_ == TransportType::Auto || protocol_ == ProtocolType::Auto ||
+      protocol_ == ProtocolType::Twitter) {
+    throw EnvoyException(
+        fmt::format("Invalid thrift health check configuration: {}", thrift_config.DebugString()));
```
zuercher, line 36: nit: should the error be more specific about what's invalid? Something like "unsupported transport or protocol in thrift health check configuration"?

source/extensions/health_checkers/thrift/thrift.cc
```diff
@@ -0,0 +1,119 @@
...
+                                            parent_.transportSocketMatchMetadata().get());
+}
+
+void ThriftHealthChecker::ThriftActiveHealthCheckSession::onEvent(Network::ConnectionEvent event) {
+  if (event == Network::ConnectionEvent::RemoteClose ||
+      event == Network::ConnectionEvent::LocalClose) {
+    ENVOY_LOG(trace, "on event close, is_local_close={} expect_close={}",
+              event == Network::ConnectionEvent::LocalClose, expect_close_);
+    if (!expect_close_) {
+      handleFailure(envoy::data::core::v3::NETWORK);
```
zuercher, line 106: Can this happen some time after a full response if the server has some kind of idle timeout? And if so, is that really a failure? The HTTP health checker might have a useful pattern to follow here.

zuercher (dismissed):
I had a few small comments, but nothing I consider a blocker.

source/extensions/health_checkers/thrift/thrift.cc
JuniorHsu: Do you mean race condition like `onTimeout` called first and a full response is on the fly?
If so, onTimeout will close the connection and no longer to receive connection callback.
Other health checkers use the same trick: a `expect_*` flag to prevent race condition.

source/extensions/health_checkers/thrift/thrift.cc
JuniorHsu: Good point. Will address it in separating PR if api is happy to move forward. Thanks!

source/extensions/health_checkers/thrift/utility.h
JuniorHsu: This is also used in thrift_test.cc. Extracting it to a separated file can minimize the link overhead and dependency.

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
```diff
@@ -0,0 +1,46 @@
...
+
+// [#protodoc-title: Thrift]
+// Thrift health checker :ref:`configuration overview <config_health_checkers_thrift>`.
+// [#extension: envoy.health_checkers.thrift]
+
+// [#next-free-field: 6]
+message Thrift {
+  reserved 1, 5;
+
+  reserved "service_name", "request_headers_to_add";
```
mattklein123, line 24: This can be removed, it's a new message.

api/envoy/extensions/health_checkers/thrift/v3/README.md
```diff
@@ -0,0 +1 @@
+Protocol buffer definitions for the Thrift health checker.
```
mattklein123, line 1: Nit, delete this file. It's obvious.

mattklein123 (changes requested):
API LGTM with a small nit, thanks.

/wait

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
JuniorHsu: Sure. Let me know if I need to reorder field id.

source/extensions/health_checkers/thrift/thrift.cc
JuniorHsu: I'll address this now since matt requested review

api/envoy/extensions/health_checkers/thrift/v3/README.md
JuniorHsu: sure. Thanks for the review.

api/envoy/extensions/health_checkers/thrift/v3/thrift.proto
JuniorHsu: ci failed. let's reorder it.

JuniorHsu:
Code coverage for source/common/config is lower than limit of 96.5 (96.0)

This is irrelevant since I didn't touch the code and looks like it happens on top of the tree.

mattklein123 (approved):

PR #23045, opened 2022-09-08
Convert ipv4 mapped ipv6 address back to ipv4 when ipv6 is forced.

RenjieTang:

Commit Message: Convert ipv4 mapped ipv6 address back to ipv4 when ipv6 is forced.
Additional Description: On Android clients, a IPV4 address will be mapped to a IPV6 address for the Envoy socket. But since the QUIC connection holds the original IPV4 address, it wouldn't be able to verify the mapped server address.
Risk Level: Low
Testing: Unit tests
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]: https://github.com/envoyproxy/envoy-mobile/issues/2515

### DIFF
--- a/source/common/quic/envoy_quic_utils.cc
+++ b/source/common/quic/envoy_quic_utils.cc
@@ -42,14 +42,6 @@ quic::QuicSocketAddress envoyIpAddressToQuicSocketAddress(const Network::Address
     ipv4_addr->sin_family = AF_INET;
     ipv4_addr->sin_port = htons(port);
     ipv4_addr->sin_addr.s_addr = envoy_ip->ipv4()->address();
-  } else if (envoy_ip->ipv6()->forceV4CompatibleAddress() != nullptr) {
-    // Convert ipv4 mapped ipv6 address for QUIC.
-    auto ipv4_addr = reinterpret_cast<sockaddr_in*>(&ss);
-    memset(ipv4_addr, 0, sizeof(sockaddr_in));
-    ipv4_addr->sin_family = AF_INET;
-    ipv4_addr->sin_port = htons(port);
-    ipv4_addr->sin_addr.s_addr =
-        envoy_ip->ipv6()->forceV4CompatibleAddress()->ip()->ipv4()->address();
   } else {
     // Create and return quic ipv6 address
     auto ipv6_addr = reinterpret_cast<sockaddr_in6*>(&ss);
--- a/source/common/quic/envoy_quic_utils.cc
+++ b/source/common/quic/envoy_quic_utils.cc
@@ -42,25 +42,22 @@ quic::QuicSocketAddress envoyIpAddressToQuicSocketAddress(const Network::Address
     ipv4_addr->sin_family = AF_INET;
     ipv4_addr->sin_port = htons(port);
     ipv4_addr->sin_addr.s_addr = envoy_ip->ipv4()->address();
+  } else if (envoy_ip->ipv6()->forceV4CompatibleAddress() != nullptr) {
+    // Convert ipv4 mapped ipv6 address for QUIC.
+    auto ipv4_addr = reinterpret_cast<sockaddr_in*>(&ss);
+    memset(ipv4_addr, 0, sizeof(sockaddr_in));
+    ipv4_addr->sin_family = AF_INET;
+    ipv4_addr->sin_port = htons(port);
+    ipv4_addr->sin_addr.s_addr =
+        envoy_ip->ipv6()->forceV4CompatibleAddress()->ip()->ipv4()->address();
   } else {
-    absl::uint128 ipv4_prefix_in_network_order{0xFFFF};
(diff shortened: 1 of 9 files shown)


source/common/quic/envoy_quic_utils.cc
```diff
@@ -43,13 +43,24 @@ quic::QuicSocketAddress envoyIpAddressToQuicSocketAddress(const Network::Address
...
-    *reinterpret_cast<absl::uint128*>(ipv6_addr->sin6_addr.s6_addr) = envoy_ip->ipv6()->address();
+    absl::uint128 ipv4_prefix_in_network_order{0xFFFF};
+    ipv4_prefix_in_network_order <<= 112;
+    if ((envoy_ip->ipv6()->address() << 32) == ipv4_prefix_in_network_order) {
+      // Convert a ipv4 mapped ipv6 address back to ipv4.
+      auto ipv4_addr = reinterpret_cast<sockaddr_in*>(&ss);
+      memset(ipv4_addr, 0, sizeof(sockaddr_in));
+      ipv4_addr->sin_family = AF_INET;
+      ipv4_addr->sin_port = htons(port);
+      ipv4_addr->sin_addr.s_addr = static_cast<uint32_t>(envoy_ip->ipv6()->address() >> 96);
```
soulxu, line 54: We have existing code help to convert ipv4-mapped ipv6 address to ipv4 address https://github.com/envoyproxy/envoy/blob/7a56d178771ba267aa2a8fd863a4d8b139fc1caf/source/common/network/address_impl.h#L237, is it help here?

RenjieTang: Thanks for the pointer!
But this led me to a dilemma.
In real world, the address comes from the socket, and is constructed by [addressFromSockAddrOrDie](https://github.com/envoyproxy/envoy/blob/main/source/common/network/address_impl.h#L44). By default, the |v6_only| parameter is true and thus v4CompatibleAddress() wouldn't work.
If we pass in **v6_only=true**, then the address will be converted v4 during construction and we don't really need to call v4CompatibleAddress() anymore. I'm not sure if we should make this change, since it will affect all kinds of connections, not just QUIC which needs the fix.
I added a forceV4CompatibleAddress() method to demonstrate. But maybe there's a better way?

RyanTheOptimist: It's really curious that v4CompatibleAddress requires v6_only_ to be true. :/ It does seem like this would otherwise be a good solution. But I agree that changing this globally could be dicey.

Both of the approaches you've take in this PR seem fine to me.

soulxu: Thanks for the update!

The v6_only flag is coming from `Network::Address::resolveProtoAddress`, and `v4CompatibleAddress` is used for check the ipv4-mapped address is valid or not. And the checking not depend v6only flag, it depends on v4_compat flag from proto config
https://github.com/envoyproxy/envoy/blob/667f0600743104857d59f56da4ab6aeea57fc60e/source/server/listener_impl.cc#L490

So it should be safe to remove the v6_only_ check inside v4CompatibleAddress().

If we still think that is risky, another option is using
https://github.com/envoyproxy/envoy/blob/667f0600743104857d59f56da4ab6aeea57fc60e/source/common/network/address_impl.cc#L49
directly.

The only point is that both `forceV4CompatibleAddress` and `ipv6ToIpv4CompatibleAddress` can take care of multiple platforms. I know this is a fix for android, but seems this code should run on other platform also? If this is totally no effect other platform, I'm fine with the both of approaches you have in this PR also.

RenjieTang: Ryan helped me find https://github.com/envoyproxy/envoy/pull/21803/files.
This is actually a broader issue for Android.

I completed Matt's TODO and basically force v6->v4 conversion when we parse socket address on Android.

RyanTheOptimist: This approach looks great! Any chance it would be possible to add a unit test in address_impl_test.cc?

RyanTheOptimist:
FYI Looks like the macos failure is a real failure:

```
source/common/network/io_socket_handle_impl.cc:474:48: error: use of undeclared identifier 'forceV6'; did you mean 'Address::forceV6'?
  if (sockaddr_to_use->sa_family == AF_INET && forceV6()) {
                                               ^~~~~~~
                                               Address::forceV6
./source/common/network/address_impl.h:25:6: note: 'Address::forceV6' declared here
bool forceV6();
     ^
1 error generated.

```

source/common/quic/envoy_quic_utils.cc
RenjieTang: done.

RyanTheOptimist (approved):

RyanTheOptimist:
I think this is ready to merge, but it would be nice to revise the PR description to match the new implementation.

PR #23049, opened 2022-09-09
fix sendLocalResponse in wasm

### LINKED ISSUE #23047: Calling `proxy_send_local_response` twice in wasm code results in access to freed memory space
*Description*:

I have two filters in my config, cors filter and wasm filter. 

```yaml
node:
  cluster: test-cluster
  id: test-idn

admin:
  address:
    socket_address: { address: 0.0.0.0, port_value: 9903 }

static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address: { address: 127.0.0.1, port_value: 10000 }
    filter_chains:
    - filters:
        typed_config:
          stat_prefix: ingress_http
          codec_type: AUTO
          route_config:
            name: local_route
            virtual_hosts:
              routes:
              - match: { path: "/" }
                name: test
                route:
                  cluster: some_service
                  cors:
                    allow_methods: "GET,POST,PUT,DELETE,HEAD,OPTIONS,PATCH"
                    allow_headers: "*"
                    expose_headers: "*"
                    max_age: "86400"
                    allow_credentials: true
                    filter_enabled:
                      default_value:
                        numerator: 100
                    allow_origin_string_match:
                      safe_regex:
                        google_re2: {}
                        regex: ".*"
          http_filters:
          - name: envoy.filters.http.cors
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.cors.v3.Cors
          - name: wasm
            typed_config:
              "@type": type.googleapis.com/udpa.type.v1.TypedStruct
              type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
              value:
                config:
                  name: wasm
                  fail_open: true
                  vm_config:
                    runtime: envoy.wasm.runtime.v8
                    code:
                      local:
                        filename: /tmp/crash.wasm
  clusters:
  - name: some_service
    connect_timeout: 0.25s
    type: STATIC
    lb_policy: ROUND_ROBIN
    load_assignment:
      cluster_name: some_service
      endpoints:
      - lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: 127.0.0.1
                port_value: 12345
```

The wasm code is like this:
```go
package main

import (
        "github.com/tetratelabs/proxy-wasm-go-sdk/proxywasm"
        "github.com/tetratelabs/proxy-wasm-go-sdk/proxywasm/types"
)

func main() {
        proxywasm.SetVMContext(&vmContext{})
}
type vmContext struct {
        types.DefaultVMContext
}
func (*vmContext) NewPluginContext(contextID uint32) types.PluginContext {
        return &pluginContext{}
}
type pluginContext struct {
        types.DefaultPluginContext
}
func (*pluginContext) NewHttpContext(contextID uint32) types.HttpContext {
        return &httpHeaders{contextID: contextID}
}
type httpHeaders struct {
        types.DefaultHttpContext
}

func (ctx *httpHeaders) OnHttpRequestBody(bodySize int, endOfStream bool) types.Action {
       proxywasm.SendHttpResponse(510, nil, nil, -1)
       proxywasm.SendHttpResponse(509, nil, nil, -1)
}
```
When I do a request like this, envoy crashes:
```bash
curl localhost:10000 -H 'origin: test'
```
Trace logs:
image

ASAN found `heap-use-after-free`.

I think the problem is : 
1. In wasm  function `proxy_on_request_body` calls host function `proxy_send_local_response` twice, registers the `sendLocalReply` by [addAfterVmCallAction](https://github.com/envoyproxy/envoy/blob/7a56d178771ba267aa2a8fd863a4d8b139fc1caf/source/extensions/common/wasm/context.cc#L1661)
2. this [doAfterVmCallActions](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/4fcf895fa2433a1cdf20704926b8b7e4039a6a04/include/proxy-wasm/wasm.h#L147) only pop the first function in the queue.
3. The first `sendLocalReply`  happens before [onRequestBody](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/4fcf895fa2433a1cdf20704926b8b7e4039a6a04/src/context.cc#L336) return, an it will trigger the wasm' filter's `encodeHeaders` and enter the wasm function `proxy_on_response_header`
4. The second `sendLocalReply` happens before [onResponseHeaders](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/4fcf895fa2433a1cdf20704926b8b7e4039a6a04/src/context.cc#L372) return, and it will re-enter the `decoder_callbacks_->sendLocalReply`.
5.  The second call of `sendLocalReply` will trigger [sendDirectLocalReply](https://github.com/envoyproxy/envoy/blob/7a56d178771ba267aa2a8fd863a4d8b139fc1caf/source/common/http/filter_manager.cc#L889), it will override the response headers, and the [reference](https://github.com/envoyproxy/envoy/blob/7a56d178771ba267aa2a8fd863a4d8b139fc1caf/source/common/http/filter_manager.cc#L934)  in the first `sendLocalReply` will be released.
6. The cors filter's `encodeHeaders` will be called after the wasm filter's `encodeHeaders` during the first `sendLocalReply`, and it will access the freed memory by using the reference of the old response headers.

johnlanni:
fix #23047 

Commit Message:
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1638,6 +1638,9 @@ void Context::failStream(WasmStreamType stream_type) {
 WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view body_text,
                                       Pairs additional_headers, uint32_t grpc_status,
                                       std::string_view details) {
+  // This flag is used to avoid calling sendLocalReply() twice, even if wasm code has this
+  // logic. We can't reuse "local_reply_sent_" here because it can't avoid calling nested
+  // sendLocalReply() during encodeHeaders().
   if (local_reply_hold_) {
     return WasmResult::BadArgument;
   }
@@ -1665,6 +1668,8 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view b
                                   modify_headers = std::move(modify_headers), grpc_status,
                                   details = StringUtil::replaceAllEmptySpace(
                                       absl::string_view(details.data(), details.size()))] {
+      // When the wasm vm fails, failStream() is called if the plugin is fail-closed, we need
+      // this flag to avoid calling sendLocalReply() twice.
       if (local_reply_sent_) {
         return;
       }
--- a/source/extensions/common/wasm/context.cc
+++ b/source/extensions/common/wasm/context.cc
@@ -1638,6 +1638,9 @@ void Context::failStream(WasmStreamType stream_type) {
 WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view body_text,
                                       Pairs additional_headers, uint32_t grpc_status,
                                       std::string_view details) {
+  if (local_reply_hold_) {
+    return WasmResult::BadArgument;
+  }
   // "additional_headers" is a collection of string_views. These will no longer
   // be valid when "modify_headers" is finally called below, so we must
   // make copies of all the headers.
@@ -1662,10 +1665,15 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view b
                                   modify_headers = std::move(modify_headers), grpc_status,
                                   details = StringUtil::replaceAllEmptySpace(
                                       absl::string_view(details.data(), details.size()))] {
+      if (local_reply_sent_) {
+        return;
+      }
       decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(response_code), body_text,
                                          modify_headers, grpc_status, details);
+      local_reply_sent_ = true;
     });
   }
+  local_reply_hold_ = true;
   return WasmResult::Ok;
 }
 
--- a/test/extensions/common/wasm/test_data/test_context_cpp.cc
+++ b/test/extensions/common/wasm/test_data/test_context_cpp.cc
@@ -63,7 +63,7 @@ FilterHeadersStatus TestContext::onResponseHeaders(uint32_t, bool) {
 
 bool TestRootContext::onStart(size_t configuration_size) {
   test_ = getBufferBytes(WasmBufferType::VmConfiguration, 0, configuration_size)->toString();
-  if (test_ == "dns_resolve") {
+  if (test_ == "resolve dns") {
     envoy_resolve_dns("example.com", sizeof("example.com") - 1, &dns_token_);
   }
   return true;
@@ -85,7 +85,7 @@ bool TestRootContext::onDone() {
 
 // Null VM fails on nullptr.
 void TestRootContext::onTick() {
-  if (test_ == "dns_resolve") {
+  if (test_ == "resolve dns") {
     if (envoy_resolve_dns(nullptr, 1, &dns_token_) != WasmResult::InvalidMemoryAccess) {
       logInfo("resolve_dns should report invalid memory access");
     }
@@ -98,7 +98,7 @@ void TestRootContext::onTick() {
 
 // V8 fails on pointer too large.
 void TestRootContext::onQueueReady(uint32_t) {
-  if (test_ == "dns_resolve") {
+  if (test_ == "resolve dns") {
     if (envoy_resolve_dns(reinterpret_cast<char*>(INT_MAX), 0, &dns_token_) !=
         WasmResult::InvalidMemoryAccess) {
       logInfo("resolve_dns should report invalid memory access");
--- a/test/extensions/common/wasm/test_data/test_context_cpp.cc
+++ b/test/extensions/common/wasm/test_data/test_context_cpp.cc
@@ -13,11 +13,6 @@
 
 START_WASM_PLUGIN(CommonWasmTestContextCpp)
 
-class TestContext : public EnvoyContext {
-public:
-  explicit TestContext(uint32_t id, RootContext* root) : EnvoyContext(id, root) {}
-};
-
 class TestRootContext : public EnvoyRootContext {
 public:
   explicit TestRootContext(uint32_t id, std::string_view root_id) : EnvoyRootContext(id, root_id) {}
@@ -28,8 +23,20 @@ public:
   void onQueueReady(uint32_t) override;
   void onResolveDns(uint32_t token, uint32_t results_size) override;
 
+  std::string test_;
+
 private:
   uint32_t dns_token_;
+};
+
+class TestContext : public EnvoyContext {
+public:
+  explicit TestContext(uint32_t id, RootContext* root) : EnvoyContext(id, root) {}
+  FilterDataStatus onRequestBody(size_t body_buffer_length, bool end_of_stream) override;
+  FilterHeadersStatus onResponseHeaders(uint32_t, bool) override;
+
+private:
+  TestRootContext* root() { return static_cast<TestRootContext*>(Context::root()); }
 };
 
 static RegisterContextFactory register_TestContext(CONTEXT_FACTORY(TestContext),
@@ -37,8 +44,24 @@ static RegisterContextFactory register_TestContext(CONTEXT_FACTORY(TestContext),
 static RegisterContextFactory register_EmptyTestContext(CONTEXT_FACTORY(EnvoyContext),
                                                         ROOT_FACTORY(EnvoyRootContext), "empty");
 
-bool TestRootContext::onStart(size_t) {
-  envoy_resolve_dns("example.com", sizeof("example.com") - 1, &dns_token_);
+FilterDataStatus TestContext::onRequestBody(size_t, bool) {
+  auto test = root()->test_;
+  if (test == "duplicate_local_reply") {
+    sendLocalResponse(200, "ok", "body", {});
+    sendLocalResponse(200, "fail", "body", {});
+  }
+  return FilterDataStatus::Continue;
+}
+
+FilterHeadersStatus TestContext::onResponseHeaders(uint32_t, bool) {
+  return FilterHeadersStatus::Continue;
+}
+
+bool TestRootContext::onStart(size_t configuration_size) {
+  test_ = getBufferBytes(WasmBufferType::VmConfiguration, 0, configuration_size)->toString();
+  if (test_ == "dns_resolve") {
+    envoy_resolve_dns("example.com", sizeof("example.com") - 1, &dns_token_);
+  }
   return true;
 }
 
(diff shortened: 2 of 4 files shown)


source/extensions/common/wasm/context.cc
```diff
@@ -1638,6 +1638,9 @@ void Context::failStream(WasmStreamType stream_type) {
 WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view body_text,
                                       Pairs additional_headers, uint32_t grpc_status,
                                       std::string_view details) {
+  if (local_reply_hold_) {
```
htuch, line 1641: Needs tests

johnlanni: Sorry for the delay, I've added a test that covers this line.

test/extensions/common/wasm/wasm_test.cc
```diff
@@ -1399,6 +1399,63 @@ TEST_P(WasmCommonContextTest, EmptyContext) {
...
+
+  TestContext& rootContext() { return *static_cast<TestContext*>(root_context_); }
+
+  TestFilter& filter() { return *static_cast<TestFilter*>(context_.get()); }
+};
+
+INSTANTIATE_TEST_SUITE_P(Runtimes, WasmCommonHttpContextTest,
+                         Envoy::Extensions::Common::Wasm::runtime_and_cpp_values);
+
+TEST_P(WasmCommonHttpContextTest, DuplicateLocalReply) {
```
htuch, line 1437: Please document in comments what this is validating.

test/extensions/common/wasm/wasm_test.cc
```diff
@@ -1399,6 +1399,63 @@ TEST_P(WasmCommonContextTest, EmptyContext) {
...
+  if (std::get<0>(GetParam()) != "null") {
+    code = TestEnvironment::readFileToStringForTest(TestEnvironment::substitute(absl::StrCat(
+        "{{ test_rundir }}/test/extensions/common/wasm/test_data/test_context_cpp.wasm")));
+  } else {
+    // The name of the Null VM plugin.
+    code = "CommonWasmTestContextCpp";
+  }
+  EXPECT_FALSE(code.empty());
+
+  setup(code, "duplicate_local_reply");
```
htuch, line 1448: This kind of conditioning of setup on test name is a bit of an anti-pattern. I would prefer whatever setup criteria parameters there are to reflect the semantic intent, rather than the caller test's name.

source/extensions/common/wasm/context.cc
```diff
@@ -1662,10 +1665,15 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view b
...
                                       absl::string_view(details.data(), details.size()))] {
+      if (local_reply_sent_) {
+        return;
+      }
       decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(response_code), body_text,
                                          modify_headers, grpc_status, details);
+      local_reply_sent_ = true;
     });
   }
+  local_reply_hold_ = true;
```
htuch, line 1676: Why do we need both hold and sent? Doesn't your "hold" case already cover the sent case?

test/extensions/common/wasm/wasm_test.cc
johnlanni: Thanks for your advice

test/extensions/common/wasm/wasm_test.cc
johnlanni: done

source/extensions/common/wasm/context.cc
johnlanni: https://github.com/envoyproxy/envoy/blob/f09ed3646b88fae777a060be37f3224dfa67f428/source/extensions/common/wasm/context.cc#L1609

https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/4fcf895fa2433a1cdf20704926b8b7e4039a6a04/src/context.cc#L34

This is to avoid sending the local reply twice if the vm fails. I added another test for this case.

johnlanni:
https://dev.azure.com/cncf/envoy/_build/results?buildId=117861&view=logs&j=b7634614-24f3-5416-e791-4f3affaaed6c&t=21e6aa7d-f369-5abd-5e4e-e888cac18e9c

https://github.com/envoyproxy/envoy/blob/f09ed3646b88fae777a060be37f3224dfa67f428/source/extensions/common/wasm/context.h#L30

There seems to be a problem with this clang tidy check, this using will be used in the context.cc like:

https://github.com/envoyproxy/envoy/blob/f09ed3646b88fae777a060be37f3224dfa67f428/source/extensions/common/wasm/context.cc#L152

test/extensions/common/wasm/test_data/test_context_cpp.cc
```diff
@@ -58,24 +85,29 @@ bool TestRootContext::onDone() {
 
 // Null VM fails on nullptr.
 void TestRootContext::onTick() {
-  if (envoy_resolve_dns(nullptr, 1, &dns_token_) != WasmResult::InvalidMemoryAccess) {
-    logInfo("resolve_dns should report invalid memory access");
-  }
-  if (envoy_resolve_dns("example.com", sizeof("example.com") - 1, nullptr) !=
-      WasmResult::InvalidMemoryAccess) {
-    logInfo("resolve_dns should report invalid memory access");
+  if (test_ == "resolve dns") {
```
htuch, line 88: The anti-pattern I'm suggesting to not do here is thread around the code a bunch of logic conditioning on strings from the caller. Please create meaningful options struct or other state variables/parameters to reflect what you want `setup` and the test fixture to do.

source/extensions/common/wasm/context.cc
htuch: I'll defer to @mathetake @lizan here, can you folks take a pass as CODEOWNERS? Thanks.

mathetake: yeah unfortunately I think it is inevitable to have both of these two booleans.

test/extensions/common/wasm/test_data/test_context_cpp.cc
mathetake: agreed cc @PiotrSikora

johnlanni: done

source/extensions/common/wasm/context.cc
```diff
@@ -1662,10 +1665,15 @@ WasmResult Context::sendLocalResponse(uint32_t response_code, std::string_view b
                                   modify_headers = std::move(modify_headers), grpc_status,
                                   details = StringUtil::replaceAllEmptySpace(
                                       absl::string_view(details.data(), details.size()))] {
+      if (local_reply_sent_) {
+        return;
+      }
       decoder_callbacks_->sendLocalReply(static_cast<Envoy::Http::Code>(response_code), body_text,
                                          modify_headers, grpc_status, details);
+      local_reply_sent_ = true;
```
PiotrSikora, line 1673: Is this issue limited to Wasm or does the same problem occur when a non-Wasm extension calls `sendLocalReply` twice?

test/extensions/common/wasm/wasm_test.cc
```diff
@@ -1399,6 +1399,91 @@ TEST_P(WasmCommonContextTest, EmptyContext) {
...
+  // In the case of VM failure, failStream is called, so we need to make sure that we don't send the
+  // local reply twice.
+  EXPECT_CALL(decoder_callbacks_,
+              sendLocalReply(Envoy::Http::Code::ServiceUnavailable, testing::Eq(""), _,
+                             testing::Eq(Grpc::Status::WellKnownGrpcStatus::Unavailable),
+                             testing::Eq("wasm_fail_stream")));
+
+  // Create in-VM context.
+  filter().onCreate();
+  EXPECT_EQ(proxy_wasm::FilterDataStatus::StopIterationNoBuffer, filter().onRequestBody(0, false));
```
mathetake, line 1484: I think these are redundant for this test

```suggestion
```

test/extensions/common/wasm/wasm_test.cc
```diff
@@ -1399,6 +1399,91 @@ TEST_P(WasmCommonContextTest, EmptyContext) {
   root_context_->validateConfiguration("", plugin_);
 }
 
+class TestFilter : public Envoy::Extensions::Common::Wasm::Context {
```
mathetake, line 1402: Don't we already have this kind of class for tests? eg shouldn't what we have for `ROOT_FACTORY(EnvoyRootContext), "empty")` be reused?

test/extensions/common/wasm/wasm_test.cc
johnlanni: `filter().onRequestBody` is used to trigger the crash, did you mean like this?
```suggestion
  // Create in-VM context.
  filter().onCreate();
  filter().onRequestBody(0, false);
```

test/extensions/common/wasm/wasm_test.cc
johnlanni: Did you mean here?  I have changed to reuse the `EnvoyRootContext`:
https://github.com/envoyproxy/envoy/pull/23049/files#diff-fb64400ce23fde0ad1cad5c6c7fb205733c5b39c540e4687547fcda36971941aR114

test/extensions/common/wasm/wasm_test.cc
mathetake: oh sorry never mind my comment!

test/extensions/common/wasm/wasm_test.cc
mathetake: no I mean I think you shouldn't need to definite this class in the first place. You should be able to reuse `WasmCommonContextTest` and control the in-Wasm root context by simply passing "send local reply twice" to the setUp method. See https://github.com/envoyproxy/envoy/blob/c94ee6c23b75591ebd54ef29b48aa82792be41ba/test/extensions/common/wasm/wasm_test.cc#L1382-L1400

mathetake: or at least modify that `WasmCommonContextTest` class to reuse it here and "empty" test as well

johnlanni: done

mathetake: Cool looking good

mathetake (dismissed):
LGTM

source/extensions/common/wasm/context.cc
htuch: @johnlanni this is fine I think, can you add some documentation in the form of comments for posterity, as the reasons we need this is subtle as per your above comment? Everything else LGTM.

htuch (approved):
LGTM, thanks!

Issue #1182, opened 2017-06-28
gRPC/JSON transcoding using reflection


mattklein123:
In https://github.com/lyft/envoy/pull/1079 @lizan has implemented transcoding that requires the proto definitions to be available to envoy and up to date. This is a great start, but as discussed, the real killer feature here IMO is to do it transparently using reflection. Theoretically this should be possible.

cc @louiscryan

jwfang:
great work +1

what happen if different upstream instances have different versions of the same protocol ? or even complete different service due to misconfigure.

currently the *protobuf descriptor* is shared for all request for the same listener, right?

mattbailey:
While figuring out #6608, adding `OPTIONS` method reflection might be something to consider for the transcoder, since either with reflection or possibly the descriptor, the transcoder is in a unique position to determine the availability of http methods on an endpoint.

Richard1ybb:
Are there any updates now ? I am very interested in this.

luoxiner:
I'm interested in that, too !

jspeiser:
I'm also very interested in this!

juliaogris:
In a Kubernetes and Istio context we have built the Kubernetes Operator [transflect](https://github.com/cashapp/transflect) to automate gRPC/JSON transcoding updates via the gRPC server reflection API. For annotated deployments transflect creates [Istio EnvoyFilter resources](https://istio.io/latest/docs/reference/config/networking/envoy-filter/), which update the [gRPC JSON transcoder filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_json_transcoder_filter) configuration with the new FileDescriptorSet and service list.

hq6:
I'm interested in possibly working on this, if nobody else has already started work on it.

However, I haven't yet read the contribution guidelines, so it'll take me a bit of time to get up to speed on procedures.

qiwzhang:
@hq6  you can start to work on it.  I can help to review the change.  But I like to hear your design first.  

the key is to build this [descriptor_pool_](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L152) by calling backend grpc server reflection.  

I assume you will do that in the main thread when JsonTranscoderConfig is constructing, since there are multiple async calls, you may need to register with init_manager to wait for them.

Thanks

hq6:
[...]

Issue #12074, opened 2020-07-14
Using envoy with multiple grpc endpoints


c10b10:
Is it possible to tweak the `envoy.yaml` file to make it proxy multiple gRPC endpoints? If so, how?

qiwzhang:
It is possible.  You can create a route config to route the request by service name.  Create a RouteMatch based in path prefix which is the gRPC service name,  each service will route to different gRPC backend defined in different clusters.

c10b10:
@qiwzhang Can you point me to some example code / documentation on this topic?

povesteam:
The gRPC services are on TCP layer, not HTTP, therefore it is not possible to differentiate multiple endpoints by path.

Instead, header-based separation is possible, using route configs like this:

```yml
        # ...
        - match:
            prefix: "/"
            headers:
              - name: x-service
                exact_match: my-awesome-service
        # ...
```

Please share if you find a way to differentiate gRPC by path. Thank you!

kundansingh86:
> It is possible.  You can create a route config to route the request by service name.  Create a RouteMatch based in path prefix which is the gRPC service name,  each service will route to different gRPC backend defined in different clusters.

Can you please share an example?

fr33kvanderwand:
> The gRPC services are on TCP layer, not HTTP, therefore it is not possible to differentiate multiple endpoints by path.
> 
> Instead, header-based separation is possible, using route configs like this:
> 
> ```yaml
>         # ...
>         - match:
>             prefix: "/"
>             headers:
>               - name: x-service
>                 exact_match: my-awesome-service
>         # ...
> ```
> 
@povesteam thanks, but can you please provide a bit more detailed example?

povesteam:
Meanwhile we changed our approach, because envoy was overly complex to configure (from my point of view as a developer; maybe an experienced devops would have done better).
Now gRPC calls are only made server-to-server, without any proxy. So we ditched envoy, at least for now.
I don't have an example anymore, @fr33kvanderwand. Sorry about that.

fr33kvanderwand:
Thanks for the reply, @povesteam  :D
I, for now, after reading a lot of envoy documentation, ended up with something like that:
```yaml
# ...
virtual_hosts:
- name gRPC-web-proxy
  domains: ["*"]
  routes:
  - name: my-super-service
    match:
      prefix: "/"
      headers:
        - name: my-service-header
          exact_match: "super-service"
    route:
      cluster: super_service
      cors: /* ... */
[...]

PR #23061, opened 2022-09-11
Add grpc reflection to grpc_json_transcoder.

hq6:
Previously, the grpc_json_transcoder filter required a statically
generated set of protobuf descriptors to perform translation.

This commit adds the ability to fetch protobuf descriptors via grpc
reflection.

Note that, per discussion on #1182, this change entails a fairly major
restructuring of the filter's code.

In addition to unit tests, this PR was tested manually using `greeter_server` from grpc examples in the grpc repository.

Signed-off-by: Henry Qin <root@hq6.me>

### DIFF
--- a/api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
+++ b/api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
@@ -16,7 +16,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // gRPC-JSON transcoder :ref:`configuration overview <config_http_filters_grpc_json_transcoder>`.
 // [#extension: envoy.filters.http.grpc_json_transcoder]
 
-// [#next-free-field: 14]
+// [#next-free-field: 15]
 // GrpcJsonTranscoder filter configuration.
 // The filter itself can be used per route / per virtual host or on the general level. The most
 // specific one is being used for a given route. If the list of services is empty - filter
@@ -103,6 +103,17 @@ message GrpcJsonTranscoder {
     bool reject_binding_body_field_collisions = 3;
   }
 
+  message ReflectionConfig {
+    // Supplies the cluster name of the gRPC cluster to use for reflection.
+    // This will typically be the same gRPC cluster that Envoy is forwarding
+    // gRPC traffic to, but it does not need to be.
+    string cluster_name = 1;
+
+    // The amount of time in ms that Envoy should wait for gRPC reflection
+    // requests to return.
+    uint64 request_timeout = 2;
+  }
+
   oneof descriptor_set {
     option (validate.required) = true;
 
@@ -115,6 +126,11 @@ message GrpcJsonTranscoder {
     // :ref:`the proto descriptor set <config_grpc_json_generate_proto_descriptor_set>` for the gRPC
     // services.
     bytes proto_descriptor_bin = 4;
+
+    // Supplies the grpc reflection configuration to fetching the content of
+    // :ref:`the proto descriptor set <config_grpc_json_generate_proto_descriptor_set>` for the gRPC
+    // services.
+    ReflectionConfig reflection_cluster_config = 14;
   }
 
   // A list of strings that
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -21,6 +21,9 @@ behavior_changes:
 
 minor_behavior_changes:
 # *Changes that may cause incompatibilities for some users, but should not for most*
+- area: grpc_json_transcoder
+  change: |
+    Support grpc reflection for building json transcoding.
 - area: resource_monitors
   change: |
     changed behavior of the fixed heap monitor to count pages allocated to TCMalloc as free memory if it's not used by Envoy. This change can be reverted temporarily by setting the runtime guard ``envoy.reloadable_features.do_not_count_mapped_pages_as_free`` to true.
--- a/docs/root/configuration/http/http_filters/grpc_json_transcoder_filter.rst
+++ b/docs/root/configuration/http/http_filters/grpc_json_transcoder_filter.rst
@@ -104,3 +104,18 @@ gRPC or RESTful JSON requests to localhost:51051.
 
 .. literalinclude:: _include/grpc-transcoder-filter.yaml
     :language: yaml
+
+
+Modified Envoy configuration to use gRPC reflection
+---------------------------------------------------
+Rather than using a statically generated configuration, Envoy can fetch service
+definitions via grpc reflection. This can be done by replacing
+``proto_descriptor`` in the filter's config with the following:
+
+.. code-block:: yaml
+
+  reflection_cluster_config:
+    cluster_name: "<INSERT_GRPC_CLUSTER_NAME_HERE"
+
+Note that reflection must be configured on the target grpc service for this to
+work; the method for doing this is language-specific.
--- a/source/extensions/filters/http/grpc_json_transcoder/BUILD
+++ b/source/extensions/filters/http/grpc_json_transcoder/BUILD
@@ -14,10 +14,20 @@ envoy_extension_package()
 
 envoy_cc_library(
     name = "json_transcoder_filter_lib",
-    srcs = ["json_transcoder_filter.cc"],
-    hdrs = ["json_transcoder_filter.h"],
+    srcs = [
+        "async_reflection_fetcher.cc",
+        "descriptor_pool_builder.cc",
+        "json_transcoder_filter.cc",
+    ],
+    hdrs = [
+        "async_reflection_fetcher.h",
+        "async_reflection_fetcher_callbacks.h",
+        "descriptor_pool_builder.h",
+        "json_transcoder_filter.h",
+    ],
     external_deps = [
         "path_matcher",
+        "grpc",
         "grpc_transcoding",
         "http_api_protos",
         "api_httpbody_protos",
@@ -26,12 +36,15 @@ envoy_cc_library(
         ":http_body_utils_lib",
         ":transcoder_input_stream_lib",
         "//envoy/http:filter_interface",
+        "//source/common/grpc:async_client_lib",
         "//source/common/grpc:codec_lib",
         "//source/common/grpc:common_lib",
         "//source/common/http:headers_lib",
         "//source/common/protobuf",
         "//source/common/runtime:runtime_features_lib",
+        "@com_github_grpc_grpc//:grpc++_reflection",
         "@com_google_googleapis//google/api:http_cc_proto",
+        "@envoy_api//envoy/config/core/v3:pkg_cc_proto",
         "@envoy_api//envoy/extensions/filters/http/grpc_json_transcoder/v3:pkg_cc_proto",
     ],
 )
--- a/source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
+++ b/source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
@@ -1,0 +1,92 @@
+#include "source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.h"
+
+#include "envoy/config/core/v3/grpc_service.pb.h"
+
+#include "source/common/common/lock_guard.h"
+#include "source/common/grpc/async_client_impl.h"
+
+using grpc::reflection::v1alpha::ServerReflection;
+using grpc::reflection::v1alpha::ServerReflectionRequest;
+using grpc::reflection::v1alpha::ServerReflectionResponse;
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace GrpcJsonTranscoder {
+
+void AsyncReflectionFetcher::requestFileDescriptors(
+    std::function<void(Envoy::Protobuf::FileDescriptorSet)> file_descriptors_available) {
+  file_descriptors_available_ = file_descriptors_available;
+  // Register to init_manager, force the listener to wait for completion of the
+  // reflection Rpcs.
+  init_target_ = std::make_unique<Init::TargetImpl>("JsonGrpcFilter: Grpc Reflection Rpcs",
+                                                    [this]() -> void { startReflectionRpcs(); });
+  init_manager_.add(*init_target_);
+}
+
+void AsyncReflectionFetcher::startReflectionRpcs() {
+  envoy::config::core::v3::GrpcService config;
+  config.mutable_envoy_grpc()->set_cluster_name(reflection_cluster_config_.cluster_name());
+
+  const Envoy::Protobuf::MethodDescriptor* server_reflection_method =
+      Envoy::Protobuf::DescriptorPool::generated_pool()->FindMethodByName(
+          "grpc.reflection.v1alpha.ServerReflection.ServerReflectionInfo");
+  if (server_reflection_method == nullptr) {
+    ENVOY_LOG(error, "Failed to find MethodDescriptor for ServerReflectionInfo");
+    return;
+  }
+
+  for (size_t i = 0; i < services_.size(); i++) {
+    std::string service = services_[i];
+    Envoy::Grpc::AsyncClient<ServerReflectionRequest, ServerReflectionResponse>& async_client =
+        async_clients_[i];
+
+    AsyncReflectionFetcherCallbacks* callbacks = new AsyncReflectionFetcherCallbacks(
+        [this](AsyncReflectionFetcherCallbacks* callback, ServerReflectionResponsePtr&& message) {
+          this->onRpcCompleted(callback, std::move(message));
+        });
+    callbacks_.emplace(callbacks);
+    remaining_callbacks_.insert(callbacks);
+    Http::AsyncClient::StreamOptions options;
+
+    // Default of 5s is chosen arbitarily.
+    options.setTimeout(std::chrono::milliseconds(reflection_cluster_config_.request_timeout()
+                                                     ? reflection_cluster_config_.request_timeout()
+                                                     : 5000));
+    Envoy::Grpc::AsyncStream<ServerReflectionRequest> async_stream =
+        async_client.start(*server_reflection_method, *callbacks, options);
+
+    ServerReflectionRequest request;
+    request.set_file_containing_symbol(service);
+    async_stream.sendMessage(request, true);
+  }
+}
+
+void AsyncReflectionFetcher::onRpcCompleted(AsyncReflectionFetcherCallbacks* callbacks,
+                                            std::unique_ptr<ServerReflectionResponse>&& message) {
+  Thread::LockGuard guard_(receive_message_mutex_);
+
+  for (const std::string& file_descriptor_bytes :
+       message->file_descriptor_response().file_descriptor_proto()) {
+    if (!file_descriptor_set_.add_file()->ParseFromString(file_descriptor_bytes)) {
+      ENVOY_LOG(error,
+                "transcoding_filter: Unable to parse proto descriptor returned from reflection");
+      return;
+    }
+  }
+
+  // Record that we have already received the response for this Rpc.
+  remaining_callbacks_.erase(callbacks);
+
+  // We have received all the protos we have asked for, so inject them into the
+  // proto config and signal to InitManager that loading is complete.
+  if (remaining_callbacks_.empty()) {
+    file_descriptors_available_(file_descriptor_set_);
+    init_target_->ready();
+  }
+}
+
+} // namespace GrpcJsonTranscoder
+} // namespace HttpFilters
+} // namespace Extensions
+} // namespace Envoy
--- a/source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
+++ b/source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
@@ -49,7 +49,7 @@ void AsyncReflectionFetcher::startReflectionRpcs() {
     remaining_callbacks_.insert(callbacks);
     Http::AsyncClient::StreamOptions options;
 
-    // Default of 5s is chosen arbitarily.
+    // Default of 5s is chosen arbitrarily.
     options.setTimeout(std::chrono::milliseconds(reflection_cluster_config_.request_timeout()
                                                      ? reflection_cluster_config_.request_timeout()
                                                      : 5000));
--- a/source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.h
+++ b/source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.h
@@ -1,0 +1,157 @@
+#pragma once
+
+#include <google/protobuf/descriptor.h>
+#include <google/protobuf/repeated_field.h>
+
+#include <memory>
+
+#include "envoy/api/api.h"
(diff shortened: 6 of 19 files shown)


source/extensions/filters/http/grpc_json_transcoder/config.cc
```diff
@@ -28,8 +29,49 @@ GrpcJsonTranscoderFilterConfig::createRouteSpecificFilterConfigTyped(
...
+    async_clients.reserve(proto_config.services_size());
+
+    for (int i = 0; i < proto_config.services_size(); i++) {
+      async_clients.push_back(std::make_shared<Envoy::Grpc::AsyncClientImpl>(
+          context.clusterManager(), config, context.api().timeSource()));
+    }
+
+    AsyncReflectionFetcherSharedPtr async_reflection_fetcher =
+        std::make_shared<AsyncReflectionFetcher>(proto_config.reflection_cluster_config(),
+                                                 proto_config.services(), async_clients,
```
qiwzhang, line 62: async_clients should be implementation details of AsyncReflectionFetcher.  It should be inside the class

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.h
```diff
@@ -0,0 +1,157 @@
...
+   * The file descriptors we have received from reflection requests thus far.
+   */
+  Envoy::Protobuf::FileDescriptorSet file_descriptor_set_ ABSL_GUARDED_BY(receive_message_mutex_);
+
+  /**
+   * This mutex protects the data structures manipulated in the
+   * onReceiveMessage callback, since the same object is used for handling all
+   * the callbacks.
+   */
+  Envoy::Thread::MutexBasicLockable receive_message_mutex_;
```
qiwzhang, line 149: Since you are using envoy grpc,  you don''t need thread protection,   its callbacks are called in the same thread when the requests are started.  They will be all in the envoy main thread.

source/extensions/filters/http/grpc_json_transcoder/BUILD
```diff
@@ -14,10 +14,20 @@ envoy_extension_package()
...
-    hdrs = ["json_transcoder_filter.h"],
+    srcs = [
+        "async_reflection_fetcher.cc",
+        "descriptor_pool_builder.cc",
+        "json_transcoder_filter.cc",
+    ],
+    hdrs = [
+        "async_reflection_fetcher.h",
+        "async_reflection_fetcher_callbacks.h",
+        "descriptor_pool_builder.h",
```
qiwzhang, line 25: making two separate libraries  for descriptor_pool_builder and asyn_reflection_fetcher

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
```diff
@@ -0,0 +1,93 @@
...
+       message->file_descriptor_response().file_descriptor_proto()) {
+    if (!file_descriptor_set_.add_file()->ParseFromString(file_descriptor_bytes)) {
+      ENVOY_LOG(error,
+                "transcoding_filter: Unable to parse proto descriptor returned from reflection");
+      return;
+    }
+  }
+
+  // Record that we have already received the response for this Rpc.
+  remaining_callbacks_.erase(callbacks);
```
qiwzhang, line 80: need to handle error case,   maybe move line 80 at top of the function?

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
```diff
@@ -0,0 +1,156 @@
...
+  FileDescriptorSet descriptor_set;
+
+  switch (proto_config_.descriptor_set_case()) {
+  case envoy::extensions::filters::http::grpc_json_transcoder::v3::GrpcJsonTranscoder::
+      DescriptorSetCase::kProtoDescriptor:
+    if (!descriptor_set.ParseFromString(
+            api_->fileSystem().fileReadToEnd(proto_config_.proto_descriptor()))) {
+      throw EnvoyException("transcoding_filter: Unable to parse proto descriptor");
+    }
+    loadFileDescriptorProtos(descriptor_set);
```
qiwzhang, line 23: Do we need to sort dependency in this case.  maybe we can call descriptor_pool_available directly?

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
```diff
@@ -0,0 +1,156 @@
...
+      Protobuf::FileDescriptorProto file_proto;
+      pool_database.FindFileContainingSymbol(symbol_name, &file_proto);
+      if (!descriptor_pool->BuildFile(file_proto)) {
+        ENVOY_LOG(error, "transcoding_filter: Unable to build proto descriptor pool");
+        return;
+      }
+    }
+  }
+
+  descriptor_pool_available_(std::move(descriptor_pool));
```
qiwzhang, line 150: maybe call "on_descriptor_pool_ready"?

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
```diff
@@ -0,0 +1,156 @@
...
+    if (!descriptor_set.ParseFromString(proto_config_.proto_descriptor_bin())) {
+      throw EnvoyException("transcoding_filter: Unable to parse proto descriptor");
+    }
+    loadFileDescriptorProtos(descriptor_set);
+    break;
+  case envoy::extensions::filters::http::grpc_json_transcoder::v3::GrpcJsonTranscoder::
+      DescriptorSetCase::kReflectionClusterConfig:
+    // The constructor of this object is responsible for ensuring the async_reflection_fetcher_ is
+    // not null.
+    async_reflection_fetcher_->requestFileDescriptors(
```
qiwzhang, line 36: Ideally,  async_reflection_fetcher should be created here, as it is implementation detail of DescriptorPoolBuilder.   For easy unit-test to pass in a mock async_fetcher,  you can use a asyn_fetch_factory, so it can be mocked.

hq6 (commented):
Thank you for the review!

Please see comments inline and let me know if 
a) you want to insist on any of the points I've expressed a different opinion on, or 
b) whether you're willing to approve if I don't change those parts of this PR

I will hold off on modifying this PR further until you've expressed an opinion in either direction, so as to avoid wasted work until we've reached agreement on the target structure of the code.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
qiwzhang: I am fine with the policy of if fetching fail, not to ready envoy.    But we need to make the error bold and clear, not require users to turn on debug log to debug it.   We could either crash the envoy by throw exception, or log error.

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
qiwzhang: If the sorting is not needed,  please don't do it.   Even it is on the filter loading time in the main thread, it will delay the envoy to be ready, it will use memory and cpu.

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
qiwzhang: number of lines doesn't matter here,  it is where the logic belongs matter.    A module design with clear logic in separate modules is much better than codes lump all logic into one function.

qiwzhang:
Open source code, at lease Envoy project,  requires code quality. It is NOT me that insisting on the way I want to do,  It is the code quality. 

There is a big problem that I have commented and you did not reply,  the freeing of the callbacks.   Let me try to find that comment or re-comment it again.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
qiwzhang: I see.  logging error is fine here for parsing the response.   But what about fetching failed cases, did we log errors?

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
```diff
@@ -0,0 +1,93 @@
...
+    ENVOY_LOG(error, "Failed to find MethodDescriptor for ServerReflectionInfo");
+    return;
+  }
+
+  for (size_t i = 0; i < services_.size(); i++) {
+    std::string service = services_[i];
+    Envoy::Grpc::AsyncClient<ServerReflectionRequest, ServerReflectionResponse>& async_client =
+        async_clients_[i];
+
+    AsyncReflectionFetcherCallbacks* callbacks = new AsyncReflectionFetcherCallbacks(
```
qiwzhang, line 45: Where is the `callbacks` freed? ???

qiwzhang:
BTW,  I like to see separate unit-test xyz_test.cc for the two new classes,  ProtoDescriptorBuilder and AsyncReflectionFetcher.   If they have dependency, mock their dependency.  

Thanks

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -103,6 +103,17 @@ message GrpcJsonTranscoder {
     bool reject_binding_body_field_collisions = 3;
   }
 
+  message ReflectionConfig {
+    // Supplies the cluster name of the gRPC cluster to use for reflection.
+    // This will typically be the same gRPC cluster that Envoy is forwarding
+    // gRPC traffic to, but it does not need to be.
```
adisuissa, line 109: Is the gRPC cluster that Envoy forwards to is the default value for this field? If so, please explicitly note so. If not, what is the default value?

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -103,6 +103,17 @@ message GrpcJsonTranscoder {
...
+    string cluster_name = 1;
+
+    // The amount of time in ms that Envoy should wait for gRPC reflection
+    // requests to return.
+    uint64 request_timeout = 2;
```
adisuissa, line 114: Could this be of type `google.protobuf.Duration`?
What is the default value?

changelogs/current.yaml
```diff
@@ -21,6 +21,9 @@ behavior_changes:
 
 minor_behavior_changes:
 # *Changes that may cause incompatibilities for some users, but should not for most*
+- area: grpc_json_transcoder
+  change: |
+    Support grpc reflection for building json transcoding.
```
adisuissa, line 26: Is this a new feature or a modified behavior?

docs/root/configuration/http/http_filters/grpc_json_transcoder_filter.rst
```diff
@@ -104,3 +104,18 @@ gRPC or RESTful JSON requests to localhost:51051.
 
 .. literalinclude:: _include/grpc-transcoder-filter.yaml
     :language: yaml
+
+
+Modified Envoy configuration to use gRPC reflection
+---------------------------------------------------
+Rather than using a statically generated configuration, Envoy can fetch service
+definitions via grpc reflection. This can be done by replacing
+``proto_descriptor`` in the filter's config with the following:
```
adisuissa, line 113: Does this imply that the `proto_descriptor` and `reflection_cluster_config` are mutually exclusive?
if so this should be clarified in the API (proto comments).

adisuissa (commented):
Thanks!
Left a few API comments.

source/extensions/filters/http/grpc_json_transcoder/BUILD
hq6: Will do.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
hq6: This following is the only failing fetch case I'm aware of. Are there others that I am not?

```
  void onRemoteClose(Envoy::Grpc::Status::GrpcStatus status, const std::string& message) override {
    if (status != Envoy::Grpc::Status::Ok) {
      ENVOY_LOG(error, "transcoding_filter: server reflection request failed with status {}: {}",
                status, message);
    }
  }
```

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
hq6: Will separate and remove the extra sorting.

source/extensions/filters/http/grpc_json_transcoder/descriptor_pool_builder.cc
hq6: For the record, I believe this is a matter of software design philosophy, but I always conform to local style, so I'll update this since it sounds like you feel strongly about it.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
hq6: Sorry I missed this earlier.

This is a good catch, that they are not freed!

However, it does leave an interesting question:
When do you think it's safe to free them in case of failure?
For success, I think we can free them after we set init_target_->ready().
For failure, there are approaches I can currently think of:
a. We can keep a count of any kind of return, success or failure, and free when that count is reached.
b. We can record that an error happened, and erase from `remaining_callbacks_` just like in success. 
When `remaining_callbacks_` reach zero, we release regardless of success or failure, but only call ready() when the error flag is not set.

Thoughts on which is preferable, or do you have other ideas for deciding when to free?

hq6:
> Open source code, at lease Envoy project, requires code quality. It is NOT me that insisting on the way I want to do, It is the code quality.
> 

For the record, I disagree with the Envoy project's definition of code quality. 
I believe understandability is more important than modularity, and subscribe to John Ousterhout's philosophy of deep abstractions to reduce complexity.

However, I will update the code to conform for consistency's sake.

> There is a big problem that I have commented and you did not reply, the freeing of the callbacks. Let me try to find that comment or re-comment it again.

Thanks for finding it again and re-commenting. I also did not find the original one, maybe GitHub lost it?

hq6:
> BTW, I like to see separate unit-test xyz_test.cc for the two new classes, ProtoDescriptorBuilder and AsyncReflectionFetcher. If they have dependency, mock their dependency.
> 
> Thanks

Yes, I think you pointed this out in your comments on my original draft PR, so I separated it out in this one.
If there's something else I didn't separate out that you asked for, please remind!

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
hq6: I don't think there is a default value for this field in the PR as written (beyond the protobuf default).  
This was an oversight on my part, thanks for catching!
Based on [this post](https://github.com/envoyproxy/envoy/issues/12074), it looks like Envoy can forward to multiple gRPC clusters.

So I have three questions.
1.Do you think we should specify a default value?
2. If we do specify a default value and there are multiple gRPC clusters configured, should we pick one arbitrarily or throw an error or log an error and do not initialize?
3. If we do not specify a default value, do you have thoughts on which of the following is preferable?
a) Logging an error and not initializing when no value is specified.
b) Throwing an exception and crashing Envoy when no value is specified.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
hq6: Yes, I can make it  a Duration. I'll document the default value regardless.

changelogs/current.yaml
hq6: It depends on the Envoy project's definition of a feature, so I'm not sure.
If I had to guess, I would guess it's a new feature because it enables a new capability that didn't exist before.

docs/root/configuration/http/http_filters/grpc_json_transcoder_filter.rst
hq6: Yes. I will make sure to make it clear in the comments.

hq6:
@qiwzhang @adisuissa It'll probably take me quite a while to address all the review comments. 
Should I close this PR or make it a draft or keep it open in the interim?

changelogs/current.yaml
adisuissa: In this case this update should probably be under a different category.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
adisuissa: To follow up on the 3rd option you suggested, in the API we have ability to denote a string field as mandatory by setting its length ([example](https://github.com/envoyproxy/envoy/blob/0202852eb0d1423f86f72cb5ef735d459ffd9183/api/envoy/config/cluster/v3/cluster.proto#L807)).
The config will be rejected in case this is not populated.
If the value of this field needs to be declared somewhere else, please clarify the docs accordingly.

adisuissa:
> @qiwzhang @adisuissa It'll probably take me quite a while to address all the review comments, since some of the changes @qiwzhang requested are structural (Note that this is a personal project so I have limited time to work on it).
> 
> Should I close this PR or make it a draft or keep it open in the interim while I address the comments?

As long as the PR keeps the same spirit, and is not completely different, then we can keep it open in "waiting" state for a while. This will keep the previous comments.
If you think it is going to be significantly different, then you can open a new PR.

/wait

hq6:
@qiwzhang I am still waiting on your opinion on [this question](https://github.com/envoyproxy/envoy/pull/23061#discussion_r970354609) before this work on this PR can proceed further.

I agree with you that it is an important problem, as you mentioned.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
qiwzhang: Actually, you do have `callbacks_` to hold the unique_ptr for them,  they are freed eventually when `AsyncReflectionFetcher` is freed.

Just a suggestion,  `remaining_callbacks_` can be just a integer to keep track of number of outstanding requests.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
qiwzhang: This should be fine then.

qiwzhang:
@hq6 I replied it.  Thanks.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
hq6: I made it a set because I was not sure if internal retries in the Envoy gRPC library could cause duplicate responses.

Do you happen to know if it duplicate responses are guaranteed to be handled and de-duped?

qiwzhang:
@hq6 Overall this pr is good. Thanks for the hard work.  It just needs a bit of restructure in classes `ProtoDescriptorBuilder` and `AsyncReflectionFetcher` with proper unit-tests.   I don't see a big changes from this pr, so we can continue to use this pr.

I can sense some defensive tone in your comment.  If my comment somehow offended you, I am sorry, it was not my intention. 

Anyway,  this is a good feature to have.  Thanks for the hard work.

source/extensions/filters/http/grpc_json_transcoder/async_reflection_fetcher.cc
qiwzhang: All envoy async callbacks are guaranteed to be called once.  Not need to worry about been called twice.  In your code, if onCloseStream() is called on network failure,  it will not call your callback.  But that is fine since we decide not to "ready" envoy in failed cases.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
qiwzhang: How about adding `config.core.v3.RetryPolicy retry_policy`.  Since network failures are so bad,  it is better to have retry.  Can add it in a later pr.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
qiwzhang: How about using `envoy::config::core::v3::GrpcService`?    `cluster_name` is only for envoy_grpc.  GrpcService can support other grpc.

But on the other hand, this grpc server is the upstream server, normal grpc requests should be routed by envoy upstream.   So it is nature to use envoy_grpc.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
qiwzhang: For `request_timeout`,  you can do

```
google.protobuf.Duration timeout = 3 [(validate.rules).duration = {
    required: true
    gte {}
  }];
```

changelogs/current.yaml
qiwzhang: It is a new feature

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
hq6: I'll defer to @adisuissa here, as the API shepherd.

changelogs/current.yaml
hq6: Thanks for confirming! Will move.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
hq6: It's a fair point. I'll add it in this PR if it's easy.

hq6:
> I can sense some defensive tone in your comment. If my comment somehow offended you, I am sorry, it was not my intention.

No offense taken, although I do object to any claims that there's one globally accepted definition of code quality today. 
I currently believe that a majority of software developers today hold as a basic tenet that modularity is more important than ease of understanding, and I am part of the minority that disagrees.

I simply wanted to make it clear that this contribution to Envoy does not mean I endorse Envoy's school of thought about what code quality means. :)

route33:
Is this merge stuck? when is going to be released?

hq6:
This PR cannot be merged until comments are addressed, and code coverage is up to the required level. 
I've been busy with my day job, and haven't yet had time to address the comments.

jeffsawatzky:
@hq6 are you still working on this? Or has this been cancelled?

hq6:
@jeffsawatzky I'm not actively working on this due to other commitments.
I was planning to try to get to it after tax season, but if someone else wants to take over the PR by addressing the remaining comments and fixing code coverage, I am not opposed.

For coordination reasons, it'd be good for such a person to comment on this PR or the related issue to avoid duplicated work.

the-duke:
@hq6 awesome work keep it up. This would be an extremely useful feature!

Issue #21615, opened 2022-06-07
router:  performance improvement around route matching: remove per-route cost of repeative work on request path/headers.


stevenzzzz:
*Title*: performance improvement around route matching

*Description*:
Seems like some low hanging fruits for route matching performance improvement: 

// In bool RouteEntryImplBase::matchRoute:

```
//  we are initializing parsing QueryParams out of request path per route (if query parameters configured in the route): 
  if (!config_query_parameters_.empty()) {
    Http::Utility::QueryParams query_parameters =
        Http::Utility::parseQueryString(headers.getPathValue());
    matches &= ConfigUtility::matchQueryParams(query_parameters, config_query_parameters_);
    if (!matches) {
      return false;
    }
  }

```

In PathMatcher::match(const absl::string_view path) there is a per-route cost of stripping the query/fragment from path:
I think there might be an issue for this already? We found this with a benchmark test.

``` 
bool PathMatcher::match(const absl::string_view path) const {
  return matcher_.match(Http::PathUtil::removeQueryAndFragment(path));
}
```

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

daixiang0:
I think `matchRoute` is for HeaderMatcher and QueryParameterMatcher, not PathMatcher, they match different fields. The path may include query params but no header.

Could you share the benchmark result here?

stevenzzzz:
you are right they match different fields, but I am not sure why that's relevant in this bug. 

The path_matcher actually only matches on <path>, not including query parameters or fragments.

The benchmark we did was internal, basically, caching the result of removeQueryAndFragment would end up with a performance improvement of nearly 5x difference.  

@AdiJohn could probably provide more details on the benchmark.  He actually filed an issue about the "PathMatcher::match" : http://go/envoy-upstream/issues/20352

stevenzzzz:
@adisuissa do you think this worth a nostale tag?

sschepens:
@stevenzzzz sure, I can do that. The only thing i'm a bit weary about query/parameters is that nowadays if someone isn't matching, we're not even parsing them, tough we can probably attach it the first time it's needed, does that seem right?

stevenzzzz:
attaching on request sounds good to me. Tho I am wondering if this worth the extra complexity. 
Path matching is like 99% of routes we have seens, where by default the path_matcher strips the query parameters (if any) by default anyway.

bryanwux:
[...]

PR #23064, opened 2022-09-11
router: Imrpove router matching by caching sanitized_path in stream_info

bryanwux:
Remove repeative work on sanitizing request headers when routing by caching `sanitized_path` in stream_info.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #21615
[Optional Deprecated:]


source/common/router/config_impl.cc
```diff
@@ -1404,10 +1404,11 @@ absl::optional<std::string> PathTemplateRouteEntryImpl::currentUrlPathAfterRewri
...
 }
 
-RouteConstSharedPtr PathTemplateRouteEntryImpl::matches(const Http::RequestHeaderMap& headers,
-                                                        const StreamInfo::StreamInfo& stream_info,
-                                                        uint64_t random_value) const {
+RouteConstSharedPtr PathTemplateRouteEntryImpl::matches(
+    const Http::RequestHeaderMap& headers, const StreamInfo::StreamInfo& stream_info,
+    uint64_t random_value, bool ignore_path_parameters_in_path_matching) const {
   throw absl::UnimplementedError("Path template matcher not implemented");
+  (void)ignore_path_parameters_in_path_matching;
```
daixiang0, line 1411: void a bool value?

bryanwux: void silences compiler unused warning. I will comment on it.

source/common/stream_info/stream_info_impl.h
```diff
@@ -312,6 +314,16 @@ struct StreamInfoImpl : public StreamInfo {
     ASSERT(downstream_bytes_meter_.get() == downstream_bytes_meter.get());
   }
 
+  void setSanitizedPath(const Http::RequestHeaderMap& headers) override {
+    sanitized_path_ = Http::PathUtil::removeQueryAndFragment(headers.getPathValue());
+    auto pos = sanitized_path_.find_first_of(";");
+    if (pos != absl::string_view::npos) {
+      sanitized_path_.remove_suffix(sanitized_path_.length() - pos);
+    }
```
giantcroc, line 322: Hi, why we need to remove suffix after `;` directly?

source/common/router/config_impl.cc
daixiang0: What about using an anonymous parameter?

source/common/stream_info/stream_info_impl.h
bryanwux: If `ignore_port_in_host_matching` is set to true, the suffix after `;` will be removed. Like `"/path-bluh;env=prod"` will be sanitized to `"/path-bluh"`.

source/common/router/config_impl.cc
bryanwux: Probably an UNREFERENCED_PARAMETER wrapper on the variable?  WDYT? https://github.com/envoyproxy/envoy/blob/0b10b055a6e1dbdd600f9a80f8b8d803bdd8ec9c/source/common/common/macros.h#L21-L26

zhxie: You can just simply omit the unused variable name in the function signature to suppress the warning/error.

bryanwux: Sounds good.

zhxie:
I guess the patch only caches sanitized path (the path without path-parameters), but the performance concern mentioned in the issue is about `removeQueryAndFragment`?

bryanwux:
> I guess the patch only caches sanitized path (the path without path-parameters), but the performance concern mentioned in the issue is about `removeQueryAndFragment`?

True. I will cache the result of `removeQueryAndFragment` beforehand.

source/common/stream_info/stream_info_impl.h
giantcroc: OK, got it, thanks!

envoy/stream_info/stream_info.h
```diff
@@ -732,6 +732,10 @@ class StreamInfo {
    */
   virtual void setDownstreamBytesMeter(const BytesMeterSharedPtr& downstream_bytes_meter) PURE;
 
+  virtual void setPathWithoutQueryAndFragment(const Http::RequestHeaderMap& headers) PURE;
```
adisuissa, line 735: Please add comments for the interface.

A different question, why not set the entire path in the stream info, and allow fetching the path without the query and fragment parts?

adisuissa (commented):
Left a high-level comment.

Assigning to Yan for further review as it is related to router.
/assign @yanavlasov

envoy/stream_info/stream_info.h
bryanwux: It's actually the result of `removeQueryAndFragment` will affect the performance. I think set the entire path in the stream info is unnecessary?

source/common/stream_info/stream_info_impl.h
giantcroc: > If `ignore_port_in_host_matching` in route configuration is set to true, the suffix after `;` will be removed. Like `"/path-bluh;env=prod"` will be sanitized to `"/path-bluh"`.

OK, got it. Thanks!

yanavlasov:
I think the problem with this change is that if a HTTP filter changes the `:path` header the router will not see this change and will still route based on the original `:path`.

Was there a performance issue found with the finding the start of query in the existing code? I understand that theoretically it can be expensive if the `:path` is long. If this needs to be addressed the only viable way is to add helper methods to the RequestHeaderMap to store string_view of the `:path` without query that can be updated any time `:path` header is changed.

bryanwux:
> Was there a performance issue found with the finding the start of query in the existing code? I understand that theoretically it can be expensive if the `:path` is long. If this needs to be addressed the only viable way is to add helper methods to the RequestHeaderMap to store string_view of the `:path` without query that can be updated any time `:path` header is changed.

Yes, caching the result of `removeQueryAndFragment` will end up with a 5x performance improvement by benchmark test. The problem is there is per-route repeative work here: 
https://github.com/envoyproxy/envoy/blob/f6f56150b119e58f671401fae4cf31c44f93bec0/source/common/common/matchers.cc#L133-L135

yanavlasov:
Are you talking about 5x improvement for a microbenchmark, or a full Envoy system benchmark? I'd expect this change to make almost no difference when measuring request throughput of a complete system.

I do not think this can be done with the `HEADER_FUNC`. This will need to be custom code that overrides the setPath method and adds additional method to access query.

PR #23067, opened 2022-09-12
http: improving shadowing with upstream filters

alyssawilk:
Fixing a bug where the router was not properly shut down for shadowing with upstream filters.
Fixing a second bug where headers were latched too late, and request headers were not isolated from shadow headers

Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/async_client_impl.h
+++ b/source/common/http/async_client_impl.h
@@ -93,6 +93,7 @@ class AsyncStreamImpl : public AsyncClient::Stream,
 public:
   AsyncStreamImpl(AsyncClientImpl& parent, AsyncClient::StreamCallbacks& callbacks,
                   const AsyncClient::StreamOptions& options);
+  ~AsyncStreamImpl() override { router_.onDestroy(); }
 
   // Http::AsyncClient::Stream
   void sendHeaders(RequestHeaderMap& headers, bool end_stream) override;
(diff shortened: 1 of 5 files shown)


alyssawilk:
cc @paul-r-gall this is a precursor to fixing up headers for upstream filters + shadowing.

alyssawilk:
cc @paul-r-gall ok, now has the header fixes in.
Also you note I had to tweak the add body filter to no longer reference the downstream connection - it was nullptr for the mirrored requests.  you possibly leverage that to not run encode functions for your mirrored requests?

mattklein123 (approved):
Thanks, LGTM!

source/common/http/async_client_impl.h
```diff
@@ -93,6 +93,7 @@ class AsyncStreamImpl : public AsyncClient::Stream,
 public:
   AsyncStreamImpl(AsyncClientImpl& parent, AsyncClient::StreamCallbacks& callbacks,
                   const AsyncClient::StreamOptions& options);
+  ~AsyncStreamImpl() override { router_.onDestroy(); }
```
mergeconflict, line 96: Sorry for the late drive-by here, but quick question: would it make sense for `Router::ProdFilter` or `Router::Filter`'s destructor to call its own `onDestroy`? (I'm looking at the comments on `StreamFilterBase::onDestroy` but I don't really understand whether they're applicable here.)

PR #23070, opened 2022-09-12
Make URL query in unit test standard compliant

yanavlasov:
Commit Message:
`[]` characters are not allowed in URL query per https://datatracker.ietf.org/doc/html/rfc3986#section-3.4

Risk Level: Low
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
+++ b/test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
@@ -280,18 +280,9 @@ TEST_P(GrpcJsonTranscoderIntegrationTest, QueryParamsDecodedName) {
       },
       R"({"id":"20","theme":"Children"})");
 
-  // json_name = "search[decoded]", "search[decoded]" should work
-  testTranscoding<bookstore::CreateShelfRequest, bookstore::Shelf>(
-      Http::TestRequestHeaderMapImpl{{":method", "POST"},
-                                     {":path", "/shelf?shelf.search[decoded]=Google"},
-                                     {":authority", "host"},
-                                     {"content-type", "application/json"}},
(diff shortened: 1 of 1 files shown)


yanavlasov:
@nareddyt to review

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
```diff
@@ -280,18 +280,9 @@ TEST_P(GrpcJsonTranscoderIntegrationTest, QueryParamsDecodedName) {
...
-          {":status", "200"},
-          {"content-type", "application/json"},
-      },
-      R"({"id":"20","theme":"Children"})");
+  // The [] characters are not valid in URL query per
+  // https://datatracker.ietf.org/doc/html/rfc3986#section-3.4 As a result the case json_name =
+  // "search[decoded]", "search[decoded]" is not tested
 
   // json_name = "search%5Bencoded%5D", "search[encode]" should fail.
   // It is tested in test case "DecodedQueryParameterWithEncodedJsonName"
```
nareddyt, line 288: This test should case should also never occur, right? Maybe just add similar comment to this file but don't remove the test `DecodedQueryParameterWithEncodedJsonName`.

nareddyt (approved):

nareddyt:
@qiwzhang FYI, since you wrote this test

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
htuch: Shouldn't we have some coverage of what behavior we expect here?

yanavlasov:
This should wait until UHV is commited

PR #23077, opened 2022-09-13
http: Improving HTTP/1.1 proxy support

alyssawilk:
When the proxy filter was implemented it did direct string checks to ensure a 200 response.  This addresses a TODO to accept any well formed 200 response.

Risk Level: low
Testing: new unit tests, integration tests
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/extensions/transport_sockets/http_11_proxy/connect.cc
+++ b/source/extensions/transport_sockets/http_11_proxy/connect.cc
@@ -7,8 +7,6 @@
 #include "source/common/buffer/buffer_impl.h"
 #include "source/common/common/scalar_to_byte_vector.h"
 #include "source/common/common/utility.h"
-#include "source/common/http/http1/balsa_parser.h"
-#include "source/common/http/http1/legacy_parser_impl.h"
 #include "source/common/network/address_impl.h"
 #include "source/common/runtime/runtime_features.h"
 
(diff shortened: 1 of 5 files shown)


alyssawilk:
cc @bencebeky I'm just straight up using balsa here if you're good with it.  It turns out http_parser stops one character early and I'd rather just use balsa.  This filter is being rolled out for Envoy Mobile and all this does is validate a 200 OK response for CONNECT requests

bencebeky:
> cc @bencebeky I'm just straight up using balsa here if you're good with it. It turns out http_parser stops one character early and I'd rather just use balsa. This filter is being rolled out for Envoy Mobile and all this does is validate a 200 OK response for CONNECT requests

Sounds good to me.

alyssawilk:
@snowp ping!

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -45,8 +58,8 @@ Network::IoResult UpstreamHttp11ConnectSocket::doWrite(Buffer::Instance& buffer,
 
 Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer) {
   if (need_to_strip_connect_response_) {
-    // Limit the CONNECT response headers to an arbitrary 200 bytes.
-    constexpr uint32_t MAX_RESPONSE_HEADER_SIZE = 200;
+    // Limit the CONNECT response headers to an arbitrary 2000 bytes.
```
snowp, line 61: I know this is arbitrary but why the increase?

PR #23002, opened 2022-09-06
http: making http11 connect proxy work with HTTP/3 auto_config

alyssawilk:
the connect proxy code was originally only designed to work with explicit HTTP/1.1 upstreams.
It happened to work for auto_config with TCP, and this allows it to work with auto_config and TCP or QUIC.

Risk Level: low
Testing: new unit, integration tests
Docs Changes: n/a
Release Notes: n/a

PR #23079, opened 2022-09-13
proxy: fix alpn proxy with HTTP3

alyssawilk:
As a follow-on to https://github.com/envoyproxy/envoy/pull/23002 make sure the proxy transport socket can be configured for auto_upstream with http/3

This changes the validation from "the transport socket is an HTTP/3 socket" to "the transport socket wraps an HTTP/3 socket" allowing the HTTP/3 socket to be wrapped in an http_11_proxy socket.  As previously tested, if a proxy address is present, we will force the request to be served over TCP, where the proxy code will be applied.

Risk Level: low
Testing: yes
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/upstream/upstream_impl.cc
+++ b/source/common/upstream/upstream_impl.cc
@@ -1200,7 +1200,8 @@ ClusterImplBase::ClusterImplBase(
 
   if (info_->features() & ClusterInfoImpl::Features::HTTP3) {
 #if defined(ENVOY_ENABLE_QUIC)
-    if (cluster.transport_socket().name() != "envoy.transport_sockets.quic") {
+    if (cluster.transport_socket().DebugString().find("envoy.transport_sockets.quic") ==
+        std::string::npos) {
       throw EnvoyException(
           fmt::format("HTTP3 requires a QuicUpstreamTransport transport socket: {}", cluster.name(),
                       cluster.DebugString()));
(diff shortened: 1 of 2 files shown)


source/common/upstream/upstream_impl.cc
```diff
@@ -1200,7 +1200,8 @@ ClusterImplBase::ClusterImplBase(
 
   if (info_->features() & ClusterInfoImpl::Features::HTTP3) {
 #if defined(ENVOY_ENABLE_QUIC)
-    if (cluster.transport_socket().name() != "envoy.transport_sockets.quic") {
+    if (cluster.transport_socket().DebugString().find("envoy.transport_sockets.quic") ==
```
RyanTheOptimist, line 1203: Out of curiosity, how does this differ from the existing code? In particular, looking at this change I don't understand how it "make(s) sure the proxy transport socket can be configured for auto_upstream with http/3", but I'm sure you do :)

RyanTheOptimist (commented):
I suspect this looks good, because the test looks good. But I don't think I actually understand the change. Care to flesh out the PR description with a discussion of the "how this PR works"? (Or just reply to my comment)

source/common/upstream/upstream_impl.cc
alyssawilk: In the previous PR, we verified that for HTTP/3 we'd downgrade requests going through a proxy to TCP. We didn't actually have an e2e test for HTTP/3 with proxy config, becasue the HTTP/3 auto tests are all in the alt-svc extension directory, and the proxy tests are in the http_11_proxy directory, and thou shalt not mix and match extensions.  Unfortunately by not e2e testing we failed to verify that you could actually _configure_ Envoy with a http_11_proxy wrapping a quic transport socket.

Previously you couldn't because we checked for an explicit HTTP/3 socket.  This change allows the HTTP/3 socket to be wrapped by an outer http_11_proxy transport socket.

RyanTheOptimist (approved):

PR #23081, opened 2022-09-13
stats: fix tag extraction for `worker_id` to avoid double counting

### LINKED ISSUE #22705: Prometheus stats with label `envoy_worker_id` are double-counted with `sum`
In #21756, a tag-extraction rule for `envoy_worker_id` was added. However, the stats with a label for the worker-id (the per-worker stats) and the global stat have the same basename, resulting in the following output from /stats/prometheus:

```
envoy_listener_downstream_cx_total{envoy_listener_address="vip"} 2
envoy_listener_downstream_cx_total{envoy_worker_id="0",envoy_listener_address="vip"} 1
envoy_listener_downstream_cx_total{envoy_worker_id="1",envoy_listener_address="vip"} 1
```

The problematic case is the following:
`sum(envoy_listener_downstream_cx_total)` will produce `4`, which is misleading because there are only 2 connections. 

Workaround: `sum(envoy_listener_downstream_cx_total{envoy_worker_id=""})` will produce `2` as expected.

ggreenway:

Commit Message:
stats: fix tag extraction for `worker_id` to avoid double counting

Default tag extraction rules were changed for `worker_id` extraction. Previously, `worker_` was removed from the original name during the extraction. This led to the same base stat name for both the per-worker and overall stat. For instance, in prometheus stats, the following stats were produced:

      envoy_listener_downstream_cx_total{} 2
      envoy_listener_downstream_cx_total{envoy_worker_id="0"} 1
      envoy_listener_downstream_cx_total{envoy_worker_id="1"} 1

This resulted in ``sum(envoy_listener_downstream_cx_total)`` producing 4, even though there are only 2 connections.
The new behavior results in stats such as this:

      envoy_listener_downstream_cx_total{} 2
      envoy_listener_worker_downstream_cx_total{envoy_worker_id="0"} 1
      envoy_listener_worker_downstream_cx_total{envoy_worker_id="1"} 1

Additional Description:
Risk Level: Low
Testing: adjusted
Docs Changes:
Release Notes: added (and verified generated formatting is correct)
Platform Specific Features:
[Optional Runtime guard:]
Fixes #22705
[Optional Deprecated:]

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -41,7 +41,7 @@ minor_behavior_changes:
     changed the filter callback interfaces to make sure that downstream-only functionality is explicit.
 - area: stats
   change: |
-    Default tag extraction rules were changed for `worker_id` extraction. Previously, `worker_` was removed from the original name during the extraction. This
+    Default tag extraction rules were changed for ``worker_id`` extraction. Previously, ``worker_`` was removed from the original name during the extraction. This
     led to the same base stat name for both the per-worker and overall stat. For instance, in prometheus stats, the following stats were produced: ::
 
       envoy_listener_downstream_cx_total{envoy_listener_address="vip"} 2
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -44,16 +44,16 @@ minor_behavior_changes:
     Default tag extraction rules were changed for ``worker_id`` extraction. Previously, ``worker_`` was removed from the original name during the extraction. This
     led to the same base stat name for both the per-worker and overall stat. For instance, in prometheus stats, the following stats were produced: ::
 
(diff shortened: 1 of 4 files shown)


jmarantz (approved):
code lgtm; I'm wondering though if this might be considered a breaking change for folks incorporating this in monitoring.

Maybe the release note is sufficient. But it's Greg who I'd be asking for advice on this.

ggreenway:
Yes this is a breaking change for some specific monitoring, but arguably this is much better than leaving it as-is, where the "obvious" query results in double-counting a number of stats.

There are many things wrong with prometheus stats right now, and all of the fixes are technically breaking, but there's not a good way to fix them without making the changes. We made a large number of these breaking changes in the previous version; I think this is the path forward.

PR #21763, opened 2022-06-17
docs: remove guidance to use names

kyessenov:
Signed-off-by: Kuat Yessenov <kuat@google.com>
Commit Message: make a pass to remove stale suggestion to use filter names.
Risk Level: low
Testing: none
Docs Changes: yes
Release Notes: none
Issue: Fix #21759

PR #23084, opened 2022-09-13
Provide custom responses based on upstream reply using redirection

### LINKED ISSUE #21441: Custom Responses based on upstream reply

pradeepcrao:
Fixes #21441

Commit Message:
Additional Description:
Risk Level:
Testing: unit and integration tests
Docs Changes: added doc
Release Notes: added release note


pradeepcrao:
wait/

pradeepcrao:
wait/

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+  message Response {
+    message RemoteDataSource {
+      // The Http URI to fetch the custom response from.
+      // It should be a full FQDN with protocol, host and path.
+      //
+      // Example:
+      //
+      // .. code-block:: yaml
+      //
+      //    uri: https://www.googleapis.com/oauth2/v1/certs
```
adisuissa, line 42: Certs is probable not the best example.
Can this be an actual example of contents sent as a custom response?
If not, just state `https://www.mydomain.com/path/to/404.txt` or something

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+      //
+      string uri = 1 [(validate.rules).string = {min_len: 1}];
+    }
+
+    message LocalDataSource {
+      // The new local reply body text if specified. It will be used
+      // in the `%LOCAL_REPLY_BODY%` command operator in the `body_format`.
+      config.core.v3.DataSource body = 1;
```
adisuissa, line 50: Are there any requirements about the `body` (e.g., is it required)?

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+
+      // Optional body format to be used for this response.
+      config.core.v3.SubstitutionFormatString body_format = 2;
+    }
+
+    // The new response status code if specified.
+    google.protobuf.UInt32Value status_code = 1 [(validate.rules).uint32 = {lt: 600 gte: 200}];
+
+    // The custom response generated body either locally or retrieved
+    // from a remote source.
```
adisuissa, line 60: What will be the outcome if the remote data source is unavailable?

adisuissa (commented):
Thanks for working on this!
Added a few API comments to start with.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
pradeepcrao: If we cannot route to it, we return the original response.

If remote data source returns an error code, we serve the error response from the remote data source to the downstream.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
pradeepcrao: This is an optional field will update the comments.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+      string uri = 1 [(validate.rules).string = {min_len: 1}];
+    }
+
+    message LocalDataSource {
+      // The new local reply body text if specified. It will be used
+      // in the `%LOCAL_REPLY_BODY%` command operator in the `body_format`.
+      config.core.v3.DataSource body = 1;
+
+      // Optional body format to be used for this response.
+      config.core.v3.SubstitutionFormatString body_format = 2;
```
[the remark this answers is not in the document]
pradeepcrao: There is no default. It's not used if not specified in the config.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+      // Optional new local reply body text. It will be used
+      // in the `%LOCAL_REPLY_BODY%` command operator in the `body_format`.
+      config.core.v3.DataSource body = 1;
+
+      // Optional body format to be used for this response.
+      config.core.v3.SubstitutionFormatString body_format = 2;
+    }
+
+    // The new response status code if specified.
+    google.protobuf.UInt32Value status_code = 1 [(validate.rules).uint32 = {lt: 600 gte: 200}];
```
wbpcode, line 57: Would it be better if the `status_code` be part of `LocalDataSource`?

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+
+    // The custom response generated body either locally or retrieved
+    // from a remote source.
+    oneof body {
+      LocalDataSource local = 2;
+
+      RemoteDataSource remote = 3;
```
wbpcode, line 64: IMO, it would be better if the data source could be extendable. It could be redis cache, a grpc service, or even a downgrade service that could accept the original request and generate a new response.

wbpcode (commented):
Thanks. This is a great work. As previous comment, if a remote data source is supported, then, IMO, it would be better to make this a `downgrade` filter that will send request to a `downgrade` handler.

The `downgrade` handler could be return a simple local response or send the request to remote service.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+    // remote body, before it is sent to a downstream client.
+    repeated config.core.v3.HeaderValueOption headers_to_add = 4
+        [(validate.rules).repeated = {max_items: 1000}];
+
+    // Name needs to be unique for selection using the unified matcher api.
+    string name = 5 [(validate.rules).string = {min_len: 1}];
+  }
+
+  // A list of custom responses.
+  repeated Response custom_responses = 1;
```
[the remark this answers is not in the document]
wbpcode: There is a name field in the `Response` as index.

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,170 @@
...
+    }
+    const auto route = decoder_callbacks_->route();
+    // Don't allow a redirect to a non existing route.
+    if (!route) {
+      config_->stats().custom_response_redirect_no_route_.inc();
+      ENVOY_LOG(trace, "Redirect for custom response failed: no route found");
+      return Http::FilterHeadersStatus::Continue;
+    }
+    downstream_headers_->setMethod(Http::Headers::get().MethodValues.Get);
+    downstream_headers_->remove(Http::Headers::get().ContentLength);
```
yanavlasov, line 141: I think the issue is that this will leak any headers added by the current filter chain to the redirected origin. This could be a problem for GFE3 or RXLB.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+// The Custom Response Filter allows for replacing upstream responses.
+
+// The filter configuration is a collection of custom response
+// policies.
+message CustomResponse {
+  // Definition of the custom response to be provided to the downstream.
+  // [#next-free-field: 6]
+  message Response {
+    message RemoteDataSource {
+      // The Http URI to fetch the custom response from.
```
yanavlasov, line 35: I think this comment is misleading, since Envoy does not fetch the response, but rather than redirects the client to this URL and handles redirect internally. There are some significant differences:

1. Extensions for the redirected URL will run
2. request header, included these added by filter chain during request processing will be sent to the redirected origin.
3. There is a comment that it does not work for internally generated responses (i.e. sendLocalReply). Is this restriction ok? I think this feature was designed for cases where control plane had failed to provide any endpoints for clusters.

yanavlasov (changes requested):
/wait-any

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+
+    // Name needs to be unique for selection using the unified matcher api.
+    string name = 5 [(validate.rules).string = {min_len: 1}];
+  }
+
+  // A list of custom responses.
+  repeated Response custom_responses = 1;
+
+  // Matcher to match against the original response using filters of type
+  // config.accesslog.v3.AccessLogFilter to select a custom_response
```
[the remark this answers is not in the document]
pradeepcrao: Sorry, need to clean up the comments.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+  // by name.
```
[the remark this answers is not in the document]
pradeepcrao: I assumed that a single policy could match against multiple matching criteria, and this way would reduce the size of the config. I guess we could just use an `or_matcher`. I will make changes accordingly.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
pradeepcrao: Sorry about that, this was true when the implementation used the Http async client. I will update it.

Regarding 3, as implemented, this feature is incompatible with sendLocalReply for `remote` responses (and not `local`), if sendLocalReply is sent after the redirection as well. 

If the control plane fails to provide clusters for the original request, the redirection should succeed if the cluster is found for the redirected route. 

Further if a custom response redirection is triggered by a sendLocalReply for a route-specific config (used for authentication for eg.) which is not exercised for the redirected route, the redirection will succeed.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,86 @@
...
+      LocalDataSource local = 2;
+
+      RemoteDataSource remote = 3;
+    }
+
+    // HTTP headers to add to the response. This allows the
+    // response policy to append, to add or to override headers of
+    // the original response for local body, or the custom response from the
+    // remote body, before it is sent to a downstream client.
+    repeated config.core.v3.HeaderValueOption headers_to_add = 4
```
paul-r-gall, line 71: rename headers_to_add -> response_headers_to_add.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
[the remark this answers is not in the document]
wbpcode: +1. And by this way, we could easily extend new implementation in the future.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
```diff
@@ -0,0 +1,103 @@
...
+};
+
+TEST_F(CustomResponseFilterTest, LocalData) {
+  setupMockObjects();
+  createConfig();
+  setupFilterAndCallback();
+
+  setServerName("server1.example.foo");
+  Http::TestResponseHeaderMapImpl headers{{":status", "499"}};
+  EXPECT_CALL(encoder_callbacks_, sendLocalReply(_, _, _, _, _));
```
yanavlasov, line 85: I think you could check the expected status and even call the callbacks that sendLocalReply provides in parameters and check the headers and body based on the config.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
```diff
@@ -0,0 +1,103 @@
...
+  EXPECT_EQ(filter_->encodeHeaders(headers, true), Http::FilterHeadersStatus::StopIteration);
+}
+
+TEST_F(CustomResponseFilterTest, DISABLED_RemoteData) {
```
yanavlasov, line 89: Should this test be removed?

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
```diff
@@ -0,0 +1,103 @@
...
+
+TEST_F(CustomResponseFilterTest, LocalData) {
+  setupMockObjects();
+  createConfig();
+  setupFilterAndCallback();
+
+  setServerName("server1.example.foo");
+  Http::TestResponseHeaderMapImpl headers{{":status", "499"}};
+  EXPECT_CALL(encoder_callbacks_, sendLocalReply(_, _, _, _, _));
+  EXPECT_EQ(filter_->encodeHeaders(headers, true), Http::FilterHeadersStatus::StopIteration);
```
yanavlasov, line 86: Would it make sense to call the `decodeHeaders` too so it can set-up the filter state and test that part as well?

yanavlasov (changes requested):
/wait

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Verify we get the correct local custom response.
+TEST_P(CustomResponseIntegrationTest, LocalReply) {
+  initialize();
+
+  codec_client_ = makeHttpConnection(lookupPort("http"));
+  default_request_headers_.setHost("some.route");
+  auto response =
+      sendRequestAndWaitForResponse(default_request_headers_, 0, unauthorized_response_, 0);
```
yanavlasov, line 145: The test name implies that Envoy's local reply is tested. But it looks like upstream fake is responding in this case.

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
```diff
@@ -0,0 +1,353 @@
...
+  default_request_headers_.setHost("some.route");
+  auto response = codec_client_->makeRequestWithBody(default_request_headers_, 10);
+  ASSERT_TRUE(response->waitForEndStream());
+  EXPECT_TRUE(response->complete());
+  // Verify we don't get a modified status value.
+  EXPECT_EQ("500", response->headers().getStatusValue());
+}
+
+// Verify that we can NOT intercept local replies sent during encode
+TEST_P(CustomResponseIntegrationTest, EncodeLocalReplyBeforeCER) {
```
yanavlasov, line 316: Is it true in all cases? Can local reply on the encode path be intercepted if it happens before the CER filter?

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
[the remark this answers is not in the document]
pradeepcrao: Removed field as the policies are match actions in the matchers.

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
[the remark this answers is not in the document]
pradeepcrao: Implemented this. PTAL.

source/extensions/filters/http/custom_response/response.h
```diff
@@ -0,0 +1,78 @@
...
+
+  Http::Code getStatusCodeForLocalReply(const Http::ResponseHeaderMap& response_headers) const;
+
+private:
+  // Unique name for this policy.
+  const std::string name_;
+
+  // Body read from local data source.
+  // Note that one of local_body_ or remote_data_source_ needs to be populated.
+  absl::optional<std::string> local_body_;
```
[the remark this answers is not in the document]
pradeepcrao: No longer applicable as the policies no longer share code.

source/extensions/filters/http/custom_response/custom_response_filter.cc
pradeepcrao: Will run integration tests in coordination with GFE team.

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
pradeepcrao: If the filter that sent the local reply sends the same local reply again on redirection by the `RedirectPolicy` of the CustomResponse filter, the downstream will see the local reply, and not the response from the upstream after redirection.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
pradeepcrao: Fixed the test.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
pradeepcrao: Done.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
pradeepcrao: Done.

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,170 @@
...
+                                                      : Http::Headers::get().SchemeValues.Https);
+      }
+    });
+
+    // Replace the original host, scheme and path.
+    downstream_headers_->setScheme(absolute_url.scheme());
+    downstream_headers_->setHost(absolute_url.hostAndPort());
+
+    auto path_and_query = absolute_url.pathAndQueryParams();
+    if (Runtime::runtimeFeatureEnabled(
```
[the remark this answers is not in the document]
pradeepcrao: Added a check in config creation, and added a testpoint in config_test.cc

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
pradeepcrao: done.

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
pradeepcrao: Actually, currently we cannot call `recreateStream` from within `sendLocalReply`. This could be fixed by leveraging the `onLocalReply` callback to return a new status which will cause the sendLocalReply to return early. I will send a PR for review following this one.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
pradeepcrao: I modified the test to check the status and body. I have a test that checks the headers in the integration test.

pradeepcrao:
@yanavlasov, @adisuissa, @wbpcode could you folks please take another pass at this? Thanks!

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
```diff
@@ -0,0 +1,90 @@
...
+message RedirectPolicy {
+  // The host that will serve the custom response.
+  //
+  // Example:
+  //
+  // .. code-block:: yaml
+  //
+  //    uri: https://www.mydomain.com
+  //
+  string host = 1 [(validate.rules).string = {min_len: 1}];
```
yanavlasov, line 38: What was the reason for splitting redirect URI into host and path?

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
```diff
@@ -0,0 +1,90 @@
...
+option java_outer_classname = "PoliciesProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/custom_response/v3;custom_responsev3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: Custom Response Policies]
+// [#extension: envoy.filters.http.custom_response]
+
+// Custom response policy to redirect the original response to a different
```
yanavlasov, line 26: It may be worthwhile to explain that Envoy will redirect internally rather then sending redirect to the client.

changelogs/current.yaml
```diff
@@ -70,6 +70,28 @@ new_features:
     :ref:`generic_proxy filter <envoy_v3_api_msg_extensions.filters.network.generic_proxy.v3.GenericProxy>`.
 - area: upstream
   change: |
+    added support for quit command to the redis proxy.
+- area: custom response http filter
+  change: |
+    added :ref:`custom response http filter <config_http_filters_custom_response>` which adds the ability to customize responses sent to downstreams using local or remote sources.
+- area: tcp_proxy
```
yanavlasov, line 77: Looks like a bad merge in this file.

yanavlasov (changes requested):
/wait

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: This to mirror how we expect customers to specify custom response policies. They will have a mapping of one backend service/bucket to several different paths for custom responses. Also having this separation is also helpful if the route is matched based on custom headers (with the information in host field) for backend buckets.

pradeepcrao: @paul-r-gall

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: Done

changelogs/current.yaml
pradeepcrao: Fixed

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,112 @@
...
+  //                 name: "504_response"
+  //                 typed_config:
+  //                   "@type": type.googleapis.com/envoy.type.matcher.v3.HttpResponseStatusCodeMatchInput
+  //               value_match:
+  //                 exact: "504"
+  //       on_match:
+  //         action:
+  //           name: action
+  //           typed_config:
+  //             "@type": type.googleapis.com/envoy.config.core.v3.TypedExtensionConfig
```
yanavlasov, line 94: I do not think you need this double wrapping of policy config into the TypedExtensionConfig. You can just include the RedirectPolicy directly. This will simplify your factory code for policies.

yanavlasov (changes requested):
/wait

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
pradeepcrao: As I understand it, ActionBase takes a proto Type as a template parameter. The action for the custom response matcher can be one of several policies, and the policy definitions need to be wrapped in a single proto definition which will be that template parameter.

pradeepcrao: On second thoughts, I don't see Action::typeUrl being used or enforced anywhere in the matcher infrastructure. Let me take a stab at implementing your suggestion, as it will make the interface and implementation a whole lot cleaner.

pradeepcrao: Thanks so much for your suggestion. I was quite irked by my initial wrapping of TypedExtensionConfigs, but clearly not enough to push through for a better solution.

PTAL

test/extensions/filters/http/custom_response/utility.h
```diff
@@ -0,0 +1,206 @@
...
+##              name: modify-request-headers-action
+##              typed_config:
+##                "@type": type.googleapis.com/google.protobuf.Struct
+  )EOF";
+
+// Simulate filters that send local reply during either encode or decode based
+// on route specific config.
+class LocalReplyDuringDecodeIfNotCER : public Http::PassThroughFilter {
+public:
+  constexpr static char name[] = "local-reply-during-decode-if-not-cer";
```
adisuissa, line 138: Why are these strings (also in `LocalReplyDuringEncodeIfNotCER` and `LocalReplyDuringDecode`) needed?
If it's for a unique id for the filter, it should be discouraged (per #21763).

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,105 @@
...
+  //             status_code: 299
+  //             uri: "https://foo.example/gateway_error"
+  //             response_headers_to_add:
+  //             - header:
+  //                 key: "foo2"
+  //                 value: "x-bar2"
+  //
+  // -- attention::
+  //  The first matched policy wins. Once the response is matched, matcher
+  //  evaluations ends.
```
adisuissa, line 100: nit:
```suggestion
  //  evaluations end.
```

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
```diff
@@ -0,0 +1,90 @@
...
+  repeated config.core.v3.HeaderValueOption response_headers_to_add = 4
+      [(validate.rules).repeated = {max_items: 1000}];
+
+  // HTTP headers to add to the request before it is internally redirected.
+  repeated config.core.v3.HeaderValueOption request_headers_to_add = 5
+      [(validate.rules).repeated = {max_items: 1000}];
+
+  // Custom action to modify request headers before selection of the
+  // redirected route.
+  config.core.v3.TypedExtensionConfig modify_request_headers_action = 6;
```
adisuissa, line 68: Can this have an `[#extension-category: ...]` to note which extensions can be used here?

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
```diff
@@ -0,0 +1,90 @@
...
+message LocalResponsePolicy {
+  // Optional new local reply body text. It will be used
+  // in the `%LOCAL_REPLY_BODY%` command operator in the `body_format`.
+  config.core.v3.DataSource body = 1;
+
+  // Optional body format to be used for this response.
+  config.core.v3.SubstitutionFormatString body_format = 2;
+
+  // The new response status code if specified.
+  google.protobuf.UInt32Value status_code = 3 [(validate.rules).uint32 = {lt: 600 gte: 200}];
```
adisuissa, line 82: If this field isn't included, does the response include the original status code?

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
```diff
@@ -0,0 +1,90 @@
...
+
+option java_package = "io.envoyproxy.envoy.extensions.filters.http.custom_response.v3";
+option java_outer_classname = "PoliciesProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/custom_response/v3;custom_responsev3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: Custom Response Policies]
+// [#extension: envoy.filters.http.custom_response]
```
adisuissa, line 24: This seems to be the same name as the CER filter. Probably should be different.

source/extensions/filters/http/custom_response/policies/local_response_policy.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "source/common/common/enum_to_int.h"
+#include "source/common/config/datasource.h"
+#include "source/common/formatter/substitution_format_string.h"
+#include "source/common/formatter/substitution_formatter.h"
+#include "source/common/http/header_map_impl.h"
+#include "source/common/http/message_impl.h"
+#include "source/common/http/utility.h"
+#include "source/common/router/header_parser.h"
+#include "source/extensions/filters/http/custom_response/custom_response_filter.h"
+
```
adisuissa, line 14: Some of the code below looks similar to [`BodyFormatter`](https://github.com/envoyproxy/envoy/blob/7805e46b0a5a27bd390c151b6ba03862331a39cf/source/common/local_reply/local_reply.cc#L19) and `LocalReplyImpl`. Should/could the code be refactored to ensure consistent behavior?

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
```diff
@@ -0,0 +1,169 @@
...
+// [#protodoc-title: Custom Response Filter]
+// [#extension: envoy.filters.http.custom_response]
+
+// The Custom Response Filter allows for replacing upstream responses.
+
+// The filter configuration is a collection of custom response
+// policies in a matcher tree. The configuration can be defined at the filter,
+// virtual host or route level. The most specific configuration will apply.
+message CustomResponse {
+  // [#next-free-field: 2]
```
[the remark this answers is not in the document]
pradeepcrao: sorry, it's not clear what needs to be moved before the message declaration.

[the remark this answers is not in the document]
adisuissa:
```suggestion
// [#next-free-field: 2]
message CustomResponse {
```

test/extensions/filters/http/custom_response/utility.h
pradeepcrao: The names are required for specifying the filter. See for example `RouteSpecificDecodeLocalReplyBeforeRedirectedCER`  in custom_response_integration_test.cc

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: Yes.

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: The CER filter has the name `Custom Response Filter`

test/extensions/filters/http/custom_response/utility.h
```diff
@@ -0,0 +1,206 @@
...
+              host: "https://global/storage"
+              path: "/internal_server_error"
+              response_headers_to_add:
+              - header:
+                  key: "foo3"
+                  value: "x-bar3"
+##            modify_request_headers_action:
+##              name: modify-request-headers-action
+##              typed_config:
+##                "@type": type.googleapis.com/google.protobuf.Struct
```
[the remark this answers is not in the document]
pradeepcrao: This is needed for some tests that replace the `##` with `  `

api/envoy/extensions/filters/http/custom_response/v3/custom_response.proto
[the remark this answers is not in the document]
pradeepcrao: done.

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: Most of the heavy lifting is done by `Envoy::Formatter::Formatter` and `LocalReplyPolicy` does not use `content_type` like `BodyFormatter`. I'll add a comment to investigate this in the future.

test/extensions/filters/http/custom_response/utility.h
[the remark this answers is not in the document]
pradeepcrao: Implemented a better way for doing this.

test/extensions/filters/http/custom_response/utility.h
pradeepcrao: Added typed config.

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: Done

pradeepcrao: On second thoughts, I'll mark it as as a TODO. Currently the use case for this is very specific to Google, and I can't think of a generic extension for this extension point that will be useful for non-Google use cases. I will take a second pass at this once this is accepted.

source/extensions/filters/http/custom_response/policies/local_response_policy.h
```diff
@@ -0,0 +1,46 @@
...
+  Http::Code getStatusCodeForLocalReply(const Http::ResponseHeaderMap& response_headers) const;
+
+private:
+  // Rewrite the response body for locally specified bodies.
+  void formatBody(const Http::RequestHeaderMap& request_headers,
+                  const Http::ResponseHeaderMap& response_headers,
+                  const StreamInfo::StreamInfo& stream_info, std::string& body) const;
+
+  // Body read from local data source.
+  absl::optional<std::string> local_body_;
```
yanavlasov, line 35: Can you mark members of the objects that are created from config `const` please? Since you are sharing config object among many filter instances in different worker threads, it would be good to express that the state in these objects is immutable.

yanavlasov (changes requested):
/wait

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,54 @@
...
+namespace Extensions {
+namespace HttpFilters {
+namespace CustomResponse {
+
+Http::FilterHeadersStatus CustomResponseFilter::decodeHeaders(Http::RequestHeaderMap& header_map,
+                                                              bool) {
+  auto filter_state = encoder_callbacks_->streamInfo().filterState()->getDataReadOnly<Policy>(
+      "envoy.filters.http.custom_response");
+  if (!filter_state) {
+    downstream_headers_ = &header_map;
```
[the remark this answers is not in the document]
pradeepcrao: The original request headermap is NOT used after the internal redirect, and is not relied upon by the filter.

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,54 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace CustomResponse {
+
+Http::FilterHeadersStatus CustomResponseFilter::decodeHeaders(Http::RequestHeaderMap& header_map,
+                                                              bool) {
+  auto filter_state = encoder_callbacks_->streamInfo().filterState()->getDataReadOnly<Policy>(
+      "envoy.filters.http.custom_response");
+  if (!filter_state) {
```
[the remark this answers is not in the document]
pradeepcrao: Added comment.

source/extensions/filters/http/custom_response/policies/local_response_policy.cc
```diff
@@ -0,0 +1,90 @@
...
+namespace HttpFilters {
+namespace CustomResponse {
+LocalResponsePolicy::LocalResponsePolicy(
+    const envoy::extensions::filters::http::custom_response::v3::LocalResponsePolicy& config,
+    Server::Configuration::CommonFactoryContext& context)
+    : formatter_(config.has_body_format()
+                     ? Formatter::SubstitutionFormatStringUtils::fromProtoConfig(
+                           config.body_format(), context)
+                     : nullptr) {
+  // TODO: is true right here?
```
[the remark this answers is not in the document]
pradeepcrao: removed. This was added when I was prototyping the code and forgot to remove this.

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
```diff
@@ -0,0 +1,485 @@
...
+    config_helper_.addVirtualHost(some_route);
+
+    auto foo = config_helper_.createVirtualHost("foo.example");
+    foo.mutable_routes(0)->set_name("foo");
+    foo.mutable_routes(0)->mutable_direct_response()->set_status(221);
+    foo.mutable_routes(0)->mutable_direct_response()->mutable_body()->set_inline_string("foo");
+    foo.mutable_routes(0)->mutable_match()->set_prefix("/");
+    config_helper_.addVirtualHost(foo);
+
+    config_helper_.addConfigModifier(
```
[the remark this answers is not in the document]
pradeepcrao: Added comments. Please let me know if this is clearer..

source/extensions/filters/http/custom_response/policies/local_response_policy.cc
```diff
@@ -0,0 +1,90 @@
...
+LocalResponsePolicy::encodeHeaders(Http::ResponseHeaderMap& headers, bool,
+                                   CustomResponseFilter& custom_response_filter) const {
+  auto encoder_callbacks = custom_response_filter.encoderCallbacks();
+  ENVOY_BUG(encoder_callbacks->streamInfo().filterState()->getDataReadOnly<Policy>(
+                "envoy.filters.http.custom_response") == nullptr,
+            "Filter State should not be set when using the LocalResponse policy.");
+  // Handle local body
+  std::string body;
+  Http::Code code = getStatusCodeForLocalReply(headers);
+  if (encoder_callbacks->streamInfo().getRequestHeaders() != nullptr) {
```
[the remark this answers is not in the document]
pradeepcrao: I was guarding against nullptr. Modified to get static empty headers in that case.

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
```diff
@@ -0,0 +1,485 @@
...
+          header_value_option->mutable_header()->set_value("direct-response-enabled");
+          header_value_option->mutable_header()->set_key("x-direct-response-header");
+
+          auto* filter = hcm.mutable_http_filters()->Add();
+          filter->set_name("envoy.filters.http.custom_response");
+          filter->mutable_typed_config()->PackFrom(custom_response_filter_config_);
+          hcm.mutable_http_filters()->SwapElements(0, 1);
+          int cer_position = 0;
+
+          for (const auto& config : filters_before_cer_) {
```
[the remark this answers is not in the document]
pradeepcrao: added comment.

source/extensions/filters/http/custom_response/policies/local_response_policy.h
pradeepcrao: Great catch. FWIW, the policies themselves are immutable and the filter only has const access to them.

source/extensions/filters/http/custom_response/policies/local_response_policy.cc
```diff
@@ -0,0 +1,90 @@
...
+    formatBody(*encoder_callbacks->streamInfo().getRequestHeaders(), headers,
+               encoder_callbacks->streamInfo(), body);
+  }
+
+  const auto mutate_headers = [this, encoder_callbacks](Http::ResponseHeaderMap& headers) {
+    header_parser_->evaluateHeaders(headers, encoder_callbacks->streamInfo());
+  };
+  encoder_callbacks->sendLocalReply(code, body, mutate_headers, absl::nullopt, "");
```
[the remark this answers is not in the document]
pradeepcrao: added, PTAL

source/extensions/filters/http/custom_response/policies/redirect_policy.cc
```diff
@@ -0,0 +1,173 @@
...
+                                             ? absl::optional<Http::Code>(static_cast<Http::Code>(
+                                                   config.status_code().value()))
+                                             : absl::optional<Http::Code>{}},
+      response_header_parser_(
+          Envoy::Router::HeaderParser::configure(config.response_headers_to_add())),
+      request_header_parser_(
+          Envoy::Router::HeaderParser::configure(config.request_headers_to_add())),
+      modify_request_headers_action_(createModifyRequestHeadersAction(config, context)) {
+  Http::Utility::Url absolute_url;
+  std::string uri(absl::StrCat(host_, path_));
```
adisuissa, line 44: nit: const

source/extensions/filters/http/custom_response/policies/redirect_policy.cc
```diff
@@ -0,0 +1,173 @@
...
+}
+
+std::unique_ptr<ModifyRequestHeadersAction> RedirectPolicy::createModifyRequestHeadersAction(
+    const envoy::extensions::filters::http::custom_response::v3::RedirectPolicy& config,
+    Envoy::Server::Configuration::ServerFactoryContext& context) {
+  std::unique_ptr<ModifyRequestHeadersAction> action;
+  if (config.has_modify_request_headers_action()) {
+    auto& factory = Envoy::Config::Utility::getAndCheckFactory<ModifyRequestHeadersActionFactory>(
+        config.modify_request_headers_action());
+    auto action_config = Envoy::Config::Utility::translateAnyToFactoryConfig(
```
adisuissa, line 58: nit: const

source/extensions/filters/http/custom_response/policies/redirect_policy.cc
```diff
@@ -0,0 +1,173 @@
...
+  // Modify the request headers & recreate stream.
+  // TODO(pradeepcrao): Currently Custom Response is not compatible with sendLocalReply.
+  if (custom_response_filter.onLocalRepplyCalled()) {
+    return Http::FilterHeadersStatus::Continue;
+  }
+  auto downstream_headers = custom_response_filter.downstreamHeaders();
+  RELEASE_ASSERT(downstream_headers != nullptr, "downstream_headers cannot be nullptr");
+
+  Http::Utility::Url absolute_url;
+  std::string uri(absl::StrCat(host_, path_));
```
adisuissa, line 103: nit: const

adisuissa (commented):
Thanks for working on this feature!
Left mostly nits.

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
pradeepcrao: Do you think the source and apis for policies should live under `<source/api>/extensions/custom_response/policies` or `<source/api>extensions/http/custom_response/policies`

source/extensions/filters/http/custom_response/policies/local_response_policy.cc
```diff
@@ -0,0 +1,90 @@
...
+                     : nullptr) {
+  // TODO: is true right here?
+  if (config.has_body()) {
+    local_body_.emplace(Config::DataSource::read(config.body(), true, context.api()));
+  }
+
+  if (config.has_status_code()) {
+    status_code_ = static_cast<Http::Code>(config.status_code().value());
+  }
+  header_parser_ = Envoy::Router::HeaderParser::configure(config.response_headers_to_add());
```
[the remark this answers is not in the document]
pradeepcrao: moved initialisation to constructor list. PTAL.

api/envoy/extensions/filters/http/custom_response/v3/policies.proto
adisuissa: I think your second suggestion (`<source/api>extensions/http/custom_response/policies`) is better, as these policies are only for HTTP responses.

In addition to moving the location, I suggest having a new extension category for the policies (e.g., `envoy.http.custom_response.policies`) and update the CER filter documentation and reference this extension category, so one can know which policies can be used in `on_match` actions.

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,63 @@
...
+  // Note that the original request header map is NOT carried over to the
+  // redirected response. The redirected request header map does NOT participate
+  // in the custom response framework.
+  auto filter_state = encoder_callbacks_->streamInfo().filterState()->getDataReadOnly<Policy>(
+      "envoy.filters.http.custom_response");
+  if (!filter_state) {
+    downstream_headers_ = &header_map;
+    const auto* per_route_settings =
+        Http::Utility::resolveMostSpecificPerFilterConfig<FilterConfig>(decoder_callbacks_);
+    config_to_use_ = per_route_settings ? static_cast<const FilterConfig*>(per_route_settings)
```
yanavlasov, line 31: Would it be possible to move resolution of most specific config into the encodeHeaders and avoid using the `config_to_use_` member?

I'm trying to see if the state that this filter encapsulates can be simplified, since it is quite difficult to reason about its validity.

yanavlasov (changes requested):
/wait-any

pradeepcrao:
Hi @adisuissa, @yanavlasov and @wbpcode, could you please take another look and see if you have any further comments regarding this PR?

source/extensions/filters/http/custom_response/config.cc
```diff
@@ -0,0 +1,65 @@
...
+
+} // namespace
+
+FilterConfig::FilterConfig(
+    const envoy::extensions::filters::http::custom_response::v3::CustomResponse& config,
+    Server::Configuration::ServerFactoryContext& context, Stats::StatName stats_prefix)
+    : stats_prefix_(stats_prefix) {
+  // Allow matcher to not be set, to allow for cases where we only have route or
+  // virtual host specific configurations.
+  if (config.has_custom_response_matcher()) {
```
yanavlasov, line 38: Suggestion: move this `if` statement to a function in anonymous namespace that returns `Matcher::MatchTreePtr<::Envoy::Http::HttpMatchingData>` and then you can make matcher_ const. Consteness makes it easier to reason about immutability of the config after it was created.

It is too bad that the `MatchTree` does not enforce const correctness, but this is a story for another day.

source/extensions/filters/http/custom_response/policy.h
```diff
@@ -0,0 +1,80 @@
...
+  ~Policy() override = default;
+
+  virtual Http::FilterHeadersStatus encodeHeaders(Http::ResponseHeaderMap&, bool,
+                                                  CustomResponseFilter&) const PURE;
+
+protected:
+  Policy() = default;
+};
+
+using PolicySharedPtr = std::shared_ptr<Policy>;
```
yanavlasov, line 33: Do you need Policy based objects to be mutable after they were created? If not, it would be better to make shared pointer of `const Policy`. This is again to make it easier to reason about config immutability.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -0,0 +1,175 @@
...
+                                            ? ::Envoy::Http::Headers::get().SchemeValues.Http
+                                            : ::Envoy::Http::Headers::get().SchemeValues.Https);
+        }
+      });
+
+  downstream_headers->setScheme(absolute_url.scheme());
+  downstream_headers->setHost(absolute_url.hostAndPort());
+
+  auto path_and_query = path_;
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.http_reject_path_with_fragment")) {
```
yanavlasov, line 133: This should be expressed in the configuration. Given that it is an internal redirect the fragment is meaningless, there is no way it can make it to the client. So we can either just always strip it, or fail configuration with fragment in the path.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -0,0 +1,175 @@
...
+  if (modify_request_headers_action_) {
+    modify_request_headers_action_->modifyRequestHeaders(*downstream_headers,
+                                                         decoder_callbacks->streamInfo(), *this);
+  }
+  const auto route = decoder_callbacks->route();
+  // Don't allow a redirect to a non existing route.
+  if (!route) {
+    stats_.custom_response_redirect_no_route_.inc();
+    ENVOY_LOG(trace, "Redirect for custom response failed: no route found");
+    return ::Envoy::Http::FilterHeadersStatus::Continue;
```
yanavlasov, line 157: The access log will log request headers mutated by the CER policy. There is not much we can do in this case, but maybe a dd a comment here.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -0,0 +1,175 @@
...
+      auto const code = *status_code_;
+      headers.setStatus(std::to_string(enumToInt(code)));
+      encoder_callbacks->streamInfo().setResponseCode(static_cast<uint32_t>(code));
+    }
+    return ::Envoy::Http::FilterHeadersStatus::Continue;
+  }
+
+  // Modify the request headers & recreate stream.
+  // TODO(pradeepcrao): Currently Custom Response is not compatible with sendLocalReply.
+  if (custom_response_filter.onLocalRepplyCalled()) {
```
yanavlasov, line 97: I still need to process in what conditions this will trigger. We can discuss this in-person.

yanavlasov (commented):
/wait

source/extensions/filters/http/custom_response/config.cc
pradeepcrao: done. Missed this when making the policy config classes have only const attributes.

source/extensions/filters/http/custom_response/policy.h
pradeepcrao: Done.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: Changed to always strip it.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: Added a comment.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: Added a comment.

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::FilterHeadersStatus CustomResponseFilter::encodeHeaders(Http::ResponseHeaderMap& headers,
+                                                              bool end_stream) {
+  // If filter state for custom response exists, it means this response is a
+  // custom response. Apply the custom response mutations to the response from
+  // the remote source and return.
+  auto filter_state = encoder_callbacks_->streamInfo().filterState()->getDataReadOnly<Policy>(
```
adisuissa, line 38: Should the return type be `const auto*`? (couldn't really figure out the return type)
nit: if the return type isn't easy to deduct, I suggest explicitly writing it).

source/extensions/filters/http/custom_response/custom_response_filter.cc
```diff
@@ -0,0 +1,65 @@
...
+      "envoy.filters.http.custom_response");
+  if (filter_state) {
+    return filter_state->encodeHeaders(headers, end_stream, *this);
+  }
+
+  // Check for route specific config.
+  const auto* per_route_settings =
+      Http::Utility::resolveMostSpecificPerFilterConfig<FilterConfig>(decoder_callbacks_);
+  const auto config_to_use = per_route_settings
```
adisuissa, line 47: Should the return type be `const auto*`?

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -0,0 +1,177 @@
...
+  // the remote source and return.
+  auto encoder_callbacks = custom_response_filter.encoderCallbacks();
+  auto decoder_callbacks = custom_response_filter.decoderCallbacks();
+  auto filter_state = encoder_callbacks->streamInfo().filterState()->getDataReadOnly<Policy>(
+      "envoy.filters.http.custom_response");
+  if (filter_state) {
+    ENVOY_BUG(filter_state == this, "Policy filter state should be this policy.");
+    //  Apply mutations if this is a non-error response. Else leave be.
+    auto const cr_code = ::Envoy::Http::Utility::getResponseStatusOrNullopt(headers);
+    if (!cr_code.has_value() || (*cr_code < 100 || *cr_code > 299)) {
```
adisuissa, line 81: Is 3xx considered an error response?

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -0,0 +1,177 @@
...
+                                            : ::Envoy::Http::Headers::get().SchemeValues.Https);
+        }
+      });
+
+  downstream_headers->setScheme(absolute_url.scheme());
+  downstream_headers->setHost(absolute_url.hostAndPort());
+
+  auto path_and_query = path_;
+  // Strip the #fragment from Location URI if it is present.
+  auto fragment_pos = path_.find('#');
```
adisuissa, line 137: nit: const

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -0,0 +1,177 @@
...
+  path_and_query = path_.substr(0, fragment_pos);
```
adisuissa, line 138: Is this used somewhere?

adisuissa (commented):
Thanks, overall LGTM.
Left a few nits and some API comments.

yanavlasov (commented):
/wait

test/extensions/filters/http/custom_response/custom_response_integration_test.cc
```diff
@@ -0,0 +1,575 @@
...
+
+// Test ModifyRequestHeadersAction
+TEST_P(CustomResponseIntegrationTest, ModifyRequestHeaders) {
+  // Add modify_request_headers_action to the config.
+  modifyPolicy<RedirectPolicyProto>(custom_response_filter_config_, "520_action",
+                                    [](RedirectPolicyProto& policy) {
+                                      auto action = policy.mutable_modify_request_headers_action();
+                                      action->set_name("modify-request-headers-action");
+                                      action->mutable_typed_config()->mutable_type_url()->assign(
+                                          "type.googleapis.com/google.protobuf.Struct");
```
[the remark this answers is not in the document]
pradeepcrao: This adds the `TestModifyRequestHeadersAction` extension to add the `x-envoy-cer-backend` header as required by the route.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: Made the comment clearer. Envoy will reject the request if the `#fragment` is present for the internal redirect.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: Clarified the comment.

api/envoy/extensions/http/custom_response/local_response_policy/v3/local_response_policy.proto
```diff
@@ -0,0 +1,44 @@
...
+
+// Custom response policy to serve a locally stored response to the
+// downstream.
+message LocalResponsePolicy {
+  // Optional new local reply body text. It will be used
+  // in the `%LOCAL_REPLY_BODY%` command operator in the `body_format`.
+  config.core.v3.DataSource body = 1;
+
+  // Optional body format to be used for this response.
+  config.core.v3.SubstitutionFormatString body_format = 2;
```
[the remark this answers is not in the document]
pradeepcrao: Clarified comment.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -0,0 +1,72 @@
...
+  // Example:
+  //
+  //  .. code-block:: yaml
+  //
+  //  path: /path/to/503_response.txt
+  string path = 2 [(validate.rules).string = {min_len: 1}];
+
+  // The new response status code if specified. This is used to override the
+  // status code of the response from the new upstream if it is not an error status.
+  google.protobuf.UInt32Value status_code = 3 [(validate.rules).uint32 = {lte: 999 gte: 100}];
```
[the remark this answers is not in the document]
pradeepcrao: This is the status code to use to overwrite the status code for the response received after internal redirect before it is sent to the downstream. We do not overwrite it if the the response received after internal redirect is >= 300.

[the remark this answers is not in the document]
pradeepcrao: Note that overriding the status is NOT dependent on the status value in the config itself. It is dependent on the current status value in the response headers after the internal redirect.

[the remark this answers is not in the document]
adisuissa: I see, thanks for clarifying.

adisuissa (commented):
/lgtm api

source/extensions/filters/http/custom_response/custom_response_filter.h
```diff
@@ -0,0 +1,58 @@
...
+  ::Envoy::Http::StreamEncoderFilterCallbacks* encoderCallbacks() { return encoder_callbacks_; }
+  ::Envoy::Http::StreamDecoderFilterCallbacks* decoderCallbacks() { return decoder_callbacks_; }
+  bool onLocalReplyCalled() const { return on_local_reply_called_; }
+
+  ~CustomResponseFilter() override = default;
+
+  CustomResponseFilter(std::shared_ptr<FilterConfig> config) : config_{std::move(config)} {}
+
+private:
+  std::shared_ptr<const FilterConfig> config_;
```
yanavlasov, line 50: nit: the `config_` can be const as well. You can do it in a follow up PR.

yanavlasov (approved):

source/common/http/matching/inputs.cc
```diff
@@ -2,13 +2,18 @@
...
 namespace Envoy {
 namespace Http {
 namespace Matching {
 REGISTER_FACTORY(HttpRequestHeadersDataInputFactory, Matcher::DataInputFactory<HttpMatchingData>);
 REGISTER_FACTORY(HttpResponseHeadersDataInputFactory, Matcher::DataInputFactory<HttpMatchingData>);
 REGISTER_FACTORY(HttpRequestTrailersDataInputFactory, Matcher::DataInputFactory<HttpMatchingData>);
 REGISTER_FACTORY(HttpResponseTrailersDataInputFactory, Matcher::DataInputFactory<HttpMatchingData>);
+REGISTER_FACTORY(HttpResponseStatusCodeInputFactory, Matcher::DataInputFactory<HttpMatchingData>);
+REGISTER_FACTORY(HttpResponseStatusCodeClassInputFactory,
+                 Matcher::DataInputFactory<HttpMatchingData>);
```
alyssawilk, line 16: drive by, this is LEGACY_ALWAYSLINK which means it'll be linked into all binaries even if the extensions using it aren't linked in.
Can we have them be in their own files and libraries to avoid that?

source/extensions/filters/http/custom_response/custom_response_filter.h
pradeepcrao: done.

source/common/http/matching/inputs.cc
pradeepcrao: done.

alyssawilk:
cc @yanavlasov what's the state on this one? Looking for !googler pass?

Issue #22335, opened 2022-07-20
DOWNSTREAM_WIRE_BYTES_SENT in listener access logs doesn't work


howardjohn:
*Title*: DOWNSTREAM_WIRE_BYTES_SENT in listener access logs doesn't work

*Description*:
Config:
```yaml
admin:
  access_log_path: "/dev/null"
  address:
    socket_address:
      address: 0.0.0.0
      port_value: 15000
node:
  id: id
  cluster: test
static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address:
        protocol: TCP
        address: 0.0.0.0
        port_value: 10000
    filter_chains:
    - filters: []
    accessLog:
    - name: envoy.access_loggers.file
      typedConfig:
        '@type': type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog
        logFormat:
          jsonFormat:
            # Either one triggers it
            downstream_header_bytes_sent: '%DOWNSTREAM_HEADER_BYTES_SENT%'
            downstream_wire_bytes_sent: '%DOWNSTREAM_WIRE_BYTES_SENT%'
        path: /dev/stdout

  clusters: []
```

Any request will then fail:
```
[2022-06-17 09:33:19.008][2181325][critical][backtrace] [external/envoy/source/server/backtrace.h:104] Caught Segmentation fault, suspect faulting address 0x0
[2022-06-17 09:33:19.008][2181325][critical][backtrace] [external/envoy/source/server/backtrace.h:91] Backtrace (use tools/stack_decode.py to get line numbers):
[2022-06-17 09:33:19.008][2181325][critical][backtrace] [external/envoy/source/server/backtrace.h:92] Envoy version: 4a5f2c7e624e84f131cd69127f16919b80265d8a/1.23.0-dev/Clean/RELEASE/BoringSSL
[2022-06-17 09:33:19.008][2181325][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #0: __restore_rt [0x7fb860423200]
[2022-06-17 09:33:19.014][2181325][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #1: Envoy::Formatter::StreamInfoUInt64FieldExtractor::extractValue() [0x55a3c347845c]
[2022-06-17 09:33:19.023][2181325][critical][backtrace] [external/envoy/source/server/backtrace.h:96] #2: Envoy::Formatter::StreamInfoFormatter::formatValue() [0x55a3c34716e6]
[...]

PR #23092, opened 2022-09-13
access_logs: Fix crash due to uninitialized bytes meter

briansonnenberg:
Opening this to get others' thoughts.  Not sure if this is how it should be handled or if there should be safety checks in the callbacks themselves.

Signed-off-by: Brian Sonnenberg <bsonnenberg@google.com>

Commit Message:  call setDownstreamBytesMeter in the case that no filter chain matched to avoid crashing due to uninitialized bytes meter.
Additional Description:
Risk Level:  Low
Testing: Manual only
Docs Changes:
Release Notes: TBD
Platform Specific Features:
[Optional Fixes #Issue] [#22335](https://github.com/envoyproxy/envoy/issues/22335)

### DIFF
--- a/source/server/active_stream_listener_base.cc
+++ b/source/server/active_stream_listener_base.cc
@@ -33,6 +33,7 @@ void ActiveStreamListenerBase::newConnection(Network::ConnectionSocketPtr&& sock
     ENVOY_LOG(debug, "closing connection from {}: no matching filter chain found",
               socket->connectionInfoProvider().remoteAddress()->asString());
     stats_.no_filter_chain_match_.inc();
+    stream_info->setDownstreamBytesMeter(std::make_shared<StreamInfo::BytesMeter>());
     stream_info->setResponseFlag(StreamInfo::ResponseFlag::NoRouteFound);
     stream_info->setResponseCodeDetails(StreamInfo::ResponseCodeDetails::get().FilterChainNotFound);
     emitLogs(*config_, *stream_info);


source/server/active_stream_listener_base.cc
```diff
@@ -33,6 +33,7 @@ void ActiveStreamListenerBase::newConnection(Network::ConnectionSocketPtr&& sock
     ENVOY_LOG(debug, "closing connection from {}: no matching filter chain found",
               socket->connectionInfoProvider().remoteAddress()->asString());
     stats_.no_filter_chain_match_.inc();
+    stream_info->setDownstreamBytesMeter(std::make_shared<StreamInfo::BytesMeter>());
```
kyessenov, line 36: BytesMeter is HTTP specific so not available for TCP and other custom protocols. Why not fix it at the point of use? It seems like a mis-configuration to apply HTTP access log to non-HTTP traffic?

briansonnenberg: Looking at [this documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage), I didn't think DOWNSTREAM_WIRE_BYTES_SENT is HTTP specific?

briansonnenberg: Okay, so to summarize there's two separate issues here: first that DOWNSTREAM_WIRE_BYTES_SENT is accessing the wrong meter for TCP specifically, and second that DOWNSTREAM_HEADER_BYTES_SENT (and _WIRE until we fix) crashes when applied to a tcp proxy that does not match a filter chain.  It does not crash when we do match a filter chain.  You're suggesting fixing the first and rejecting the second as misconfiguration?

kyessenov: I am suggesting:
1) TCP meters should work. Does it work today using `bytesMeter`?
2) Access log should not crash when mis-configured. That means either configuration is rejected or that it defensively checks for `null` `bytesMeter` before reading it.

briansonnenberg: Maybe I'm misunderstanding, but I thought bytesMeter is the meter for upstream bytes info (the downstream bytes that we send/receive on the upstream connection).  downstreamBytesMeter is the meter we use to track bytes we send/receive from downstream.  We use both for TCP.  We end up initializing the downstreamBytesMeter when we match a filter_chain anyway.  All that this change does is initialize it in the case that we don't match a filter chain.

```
// Happy path with filter chain match
#0  0x0000561ecf36eea4 in Envoy::StreamInfo::StreamInfoImpl::setDownstreamBytesMeter(std::shared_ptr<Envoy::StreamInfo::BytesMeter> const&) ()
#1  0x0000561ecf62b875 in Envoy::TcpProxy::Filter::initialize(Envoy::Network::ReadFilterCallbacks&, bool) ()
#2  0x0000561ecf62a752 in Envoy::TcpProxy::Filter::initializeReadFilterCallbacks(Envoy::Network::ReadFilterCallbacks&) ()
#3  0x0000561ed49f5ef7 in Envoy::Network::FilterManagerImpl::addReadFilter(std::shared_ptr<Envoy::Network::ReadFilter>) ()
#4  0x0000561ed49d6ab2 in Envoy::Network::ConnectionImpl::addReadFilter(std::shared_ptr<Envoy::Network::ReadFilter>) ()
#5  0x0000561ecf5a2714 in Envoy::Extensions::NetworkFilters::TcpProxy::ConfigFactory::createFilterFactoryFromProtoTyped(envoy::extensions::filters::network::tcp_proxy::v3::TcpProxy const&, Envoy::Server::Configuration::FactoryContext&)::$_0::operator()(Envoy::Network::FilterManager&) const ()
#6  0x0000561ecf5a2522 in std::_Function_handler<void (Envoy::Network::FilterManager&), Envoy::Extensions::NetworkFilters::TcpProxy::ConfigFactory::createFilterFactoryFromProtoTyped(envoy::extensions::filters::network::tcp_proxy::v3::TcpProxy const&, Envoy::Server::Configuration::FactoryContext&)::$_0>::_M_invoke(std::_Any_data const&, Envoy::Network::FilterManager&) ()
#7  0x0000561ed3b84835 in std::function<void (Envoy::Network::FilterManager&)>::operator()(Envoy::Network::FilterManager&) const ()
#8  0x0000561ed3fa8826 in Envoy::Server::Configuration::FilterChainUtility::buildFilterChain(Envoy::Network::FilterManager&, std::vector<std::function<void (Envoy::Network::FilterManager&)>, std::allocator<std::function<void (Envoy::Network::FilterManager&)> > > const&) ()
#9  0x0000561ed3e64625 in Envoy::Server::ListenerImpl::createNetworkFilterChain(Envoy::Network::Connection&, std::vector<std::function<void (Envoy::Network::FilterManager&)>, std::allocator<std::function<void (Envoy::Network::FilterManager&)> > > const&) ()
#10 0x0000561ed474b438 in Envoy::Server::ActiveStreamListenerBase::newConnection(std::unique_ptr<Envoy::Network::ConnectionSocket, std::default_delete<Envoy::Network::ConnectionSocket> >&&, std::unique_ptr<Envoy::StreamInfo::StreamInfo, std::default_delete<Envoy::StreamInfo::StreamInfo> >) ()
#11 0x0000561ed475477f in Envoy::Server::ActiveTcpSocket::newConnection() ()
#12 0x0000561ed475557c in Envoy::Server::ActiveTcpSocket::continueFilterChain(bool) ()
#13 0x0000561ed3116d1a in Envoy::Server::ActiveTcpSocket::startFilterChain() ()
#14 0x0000561ed31143f5 in Envoy::Server::ActiveStreamListenerBase::onSocketAccepted(std::unique_ptr<Envoy::Server::ActiveTcpSocket, std::default_delete<Envoy::Server::ActiveTcpSocket> >) ()
```

In the happy path, we see this in the logs:

```{"bytes_sent":174,"downstream_wire_bytes_sent":174,"downstream_header_bytes_sent":0}```

kyessenov: There's some contradiction in the code -- it sounds like byte meters are used in `tcp_proxy` even though they are for HTTP. I still think the access loggers should just check for `nullptr` because that will also fix it for other network filters. For example, I think it probably also crashes for redis.

briansonnenberg: Yeah, looks like it was done [here](https://github.com/envoyproxy/envoy/commit/26329cc6e3e87e061a37b08cc79f63ad257bd837).  Wanted to avoid the null check, but agree that it's probably best.

kyessenov: @jstraceski Can you explain why your PR only covers `tcp_proxy` for byte metering instead of all network connections like it was done for the old `bytesSent` [here](https://github.com/envoyproxy/envoy/commit/26329cc6e3e87e061a37b08cc79f63ad257bd837)? I think they probably don't match for filter chains that mutate the streams AFAICT.

kyessenov:
We had a brief offline discussion with @KBaichoo and @jstraceski and decided to move the bytes meter to the network filter chain manager, so that we fix it for all network filters, and not just missing filter chain.

Issue #4540, opened 2018-09-26
FilterChain Discovery Service (FDS)


htuch:
In order to support filter chain updates without forcing a full listener drain, it's been suggested we add a FilterChain Discovery Service. This would allow individual FilterChains belonging to a Listener to be independently discovered and updated.

This would also lend support to finer grained Listener resource naming for incremental and on-demand xDS (generalizing https://github.com/envoyproxy/envoy/issues/2500).

Opening this issue to track any design work and ownership.

@ggreenway @andraxylia @lizan @PiotrSikora @mattklein123

andraxylia:
We need this fix for Istio for long-lived connections with SNI @rshriram @costinm .

ggreenway:
Related add-on feature: lazy-loading of FilterChains, triggered by a listener-filter that is watching for unknown SNI values.

andraxylia:
@htuch does the drain happen when a particular FilterChain is updated, and also when any FilterChain is added or deleted in the listener?

mattklein123:
> We need this fix for Istio for long-lived connections with SNI

@andraxylia the intention here is to only effect new connections, right? Obviously previous connections will use the old filter chain? I just want to make sure we are on the same page. If so this sounds like a useful and straightforward feature to me.

andraxylia:
@mattklein123 All is good if a change in the FilterChain  does not affect existing TCP and GRPC connections - for instance the addition of a host in server_names in the FilterMatch.

@rshriram @costinm told me that because of issues with full drain we cannot move Istio from multiple listeners to 1 listener with multiple FilterChains.

ggreenway:
I brainstormed with @htuch and here's what we propose for now:

* Listeners will specify a list of FilterChains by name to discover. That means that adding a new static FilterChain will involve modifying the listener. This model is the best way to support lazy-loading of FilterChains.
* We will modify the current listener implementation so that if a listener has filter chains added, modified, or removed, draining only happens to connections on the affected filter chain. Other connections do not need to drain. This will be done as a separate PR, as it does not require filterchain discovery.
* FilterChain Discovery Service is geared towards the lazy-load use case. Statically defined/loaded filter chains can be added/modified/removed with the existing LDS.

Any comments or concerns with this plan?

mattklein123:
I don't understand how filter chains interact with draining? Can't old connections use the old filter chain and new ones use the new filter chain? If we want to add draining on filter chain change I think that's totally orthogonal?

[...]

PR #23096, opened 2022-09-13
Feature : Listener Filter Chain Discovery Service

rakeshdatta:
**Commit Message**: Feature: Listener Filter Chain Discovery Service (SoTW)
**Additional Description:** 
Representing tenants as filter-chains in envoy, its essential to be able to dynamically and independently load the tenant (filter chain) configs. To achieve that, this feature introduces anothe xDS called Filter Chain Discovery Service, which allows the filter chains inside a listener to be discovered dynamically. 

This allows a tenant config to be added, deleted and modified on the fly, without impacting other tenant configs.

This is also an Implementation for the ask here:  (https://github.com/envoyproxy/envoy/issues/4540)
This feature would allow dynamic config update of filter chains.

**Main envoy config yaml:**

```
admin:
  address:
    socket_address:
      protocol: TCP
      address: 0.0.0.0
      port_value: 9090
node:
  id: envoy_01
  cluster: talon_01
static_resources:
  listeners:
  - name: https_listener
    address:
      socket_address:
        address: 0.0.0.0
        port_value: 9443
    listener_filters:
      - name: envoy.filters.listener.original_dst
      - name: envoy.filters.listener.tls_inspector
      - name: envoy.filters.listener.tenant_inspector
    fcds_cfg:
      name: "filter_chains_config_01"
      config_source:
        path: ./fcds.yaml

  clusters:
  - name: dynamic_forward_proxy_cluster
    connect_timeout: 1s
    lb_policy: CLUSTER_PROVIDED
    cluster_type:
      name: envoy.clusters.dynamic_forward_proxy
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.clusters.dynamic_forward_proxy.v3.ClusterConfig
        dns_cache_config:
          name: dynamic_forward_proxy_cache_config
          dns_lookup_family: V4_ONLY
```

**fcds (dynamic) config yaml:**

```
---
version_info: "0"
resources:
- "@type": type.googleapis.com/envoy.config.listener.v3.FilterChain
  name: "tenant_id_01"
  filter_chain_match:
    tenant_id: "tenant_id_01"
  filters:
    - name: envoy.filters.network.sni_filter
      <custom filter logic goes here>
    - name: envoy.filters.network.sni_dynamic_forward_proxy
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.filters.network.sni_dynamic_forward_proxy.v3.FilterConfig
        port_value: 443
        dns_cache_config:
          name: dynamic_forward_proxy_cache_config
          dns_lookup_family: V4_ONLY
    - name: envoy.tcp_proxy
      typed_config:
        stat_prefix: tcp
        cluster: dynamic_forward_proxy_cluster
- "@type": type.googleapis.com/envoy.config.listener.v3.FilterChain
  name: "tenant_id_02"
  filter_chain_match:
    tenant_id: "tenant_id_02"
  filters:
    - name: envoy.filters.network.sni_filter
      <custom filter config goes here>
    - name: envoy.filters.network.sni_dynamic_forward_proxy
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.filters.network.sni_dynamic_forward_proxy.v3.FilterConfig
        port_value: 443
        dns_cache_config:
          name: dynamic_forward_proxy_cache_config
          dns_lookup_family: V4_ONLY
    - name: envoy.tcp_proxy
      typed_config:
        stat_prefix: tcp
        cluster: dynamic_forward_proxy_cluster
```

**Risk Level: High**

**Testing**: Locally tested with Inotify-based config updates.

**Pending work**:  Test and implement filter-chain level draining

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


mattklein123:
Assigning over to @htuch @adisuissa for first pass on API and general feature overview. Thank you!

api/envoy/config/listener/v3/listener.proto
```diff
@@ -358,3 +362,8 @@ message Listener {
   // :ref:`global_downstream_max_connections <config_overload_manager_limiting_connections>`.
   bool ignore_global_conn_limit = 31;
 }
+
+message Fcds{
+  string fcds_name = 1 [(validate.rules).string.min_len = 1];
+  envoy.config.core.v3.ConfigSource config_source = 2 [(validate.rules).message.required = true];
+}
```
adisuissa, line 369: This should probably be called ListenerFCDS, to differentiate between listener and http filter chains.

api/envoy/config/listener/v3/listener.proto
```diff
@@ -138,6 +139,9 @@ message Listener {
   // :ref:`FAQ entry <faq_how_to_setup_sni>`.
   repeated FilterChain filter_chains = 3;
 
+  // FCDS: Filter Chain Discovery Service config block. filter_chains and fcds are mutually exclusive
+  Fcds fcds = 34;
```
adisuissa, line 143: I think this needs to be promoted to oneof with `filter_chains`, and clearly comment that they are mutual exclusive.

Also, is this going to return a single filter-chain? Maybe it should return a glob collection of filter-chains.

api/envoy/service/filter_chain/fcds.proto
```diff
@@ -0,0 +1,46 @@
...
+// configurations. Each listener will bind its HTTP connection manager filter to
+// a route table via this identifier.
+service FcdsDiscoveryService {
+  option (envoy.annotations.resource).type = "envoy.config.litener.v3.FilterChain";
+
+  rpc StreamFilterChains(stream discovery.v3.DiscoveryRequest)
+      returns (stream discovery.v3.DiscoveryResponse) {
+  }
+
+  rpc DeltaFilterChains(stream discovery.v3.DeltaDiscoveryRequest)
```
adisuissa, line 33: Not yet implemented in this PR.

adisuissa (commented):
Thanks for working on this!

I think DCO needs to be fixed.

High level question - what are the warming constraints this xDS service require?

api/envoy/config/listener/v3/listener.proto
```diff
@@ -358,3 +362,8 @@ message Listener {
   // :ref:`global_downstream_max_connections <config_overload_manager_limiting_connections>`.
   bool ignore_global_conn_limit = 31;
 }
+
+message Fcds{
+  string fcds_name = 1 [(validate.rules).string.min_len = 1];
```
kyessenov, line 367: Every listener filter chain has a name so one can think of a listener as a collection of filter chains. Is this name referring to the collection name?

api/envoy/service/filter_chain/fcds.proto
rakeshdatta: Yes @adisuissa Delta is not implemented in this PR. I am working on it and will submit using a subsequent PR. Hope that works.

api/envoy/config/listener/v3/listener.proto
rakeshdatta: I tried oneof when I started implementing this. Had some issues that I can't recall. Will give it another try and update this PR as needed.

api/envoy/config/listener/v3/listener.proto
rakeshdatta: That's right @kyessenov. It is referring to the name of the collection of listener filter chains under a listener. It is not used in the code anywhere today though. It's just a name to refer to.

Can I have multiple of those such collections of listener filter-chains under the same listener? No.

kyessenov: Thanks for explaining. If two collections share a filter chain, then at the transport level, do we still have to get duplicate resources `<fcds_name/filter_chain.name>` if two listeners share a filter chain? I wonder if there is a way to have a global subscription instead of a per-listener subscription model.

adisuissa: > Thanks for explaining. If two collections share a filter chain, then at the transport level, do we still have to get duplicate resources `<fcds_name/filter_chain.name>` if two listeners share a filter chain? I wonder if there is a way to have a global subscription instead of a per-listener subscription model.

Somewhat dependent on the use-case, but if you think of ADS, then there should be a single subscription to the management server, and it will notify the multiple "watchers".

api/envoy/config/listener/v3/listener.proto
adisuissa: > I tried oneof when I started implementing this. Had some issues that I can't recall. Will give it another try and update this PR as needed.
> 
Changing to oneof breaks backwards compatibility. This should be promoted in future major version, and clearly stated in the comments.


> FCDS block would encompass a collection of filter chains., which are under a listener Plz refer to the yaml samples in the PR description.

Looking [here](https://github.com/envoyproxy/envoy/pull/23096/files#diff-cc79aca7dddd943b7bb9979041523e3a1ffd52c36a6f51b5265493b828b48572R27) it implies that the returned type is a single FilterChain object for each request, not multiple ones. I think this should be a [Glob collection](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md#glob) and return a list of such elements.

rakeshdatta:
> /wait

I am working on the comments plus the draining logic. Will be back shortly. Thanks for reviewing the PR.

api/envoy/service/filter_chain/fcds.proto
```diff
@@ -0,0 +1,46 @@
...
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: FCDS]
+
+// The resource_names field in DiscoveryRequest specifies a route configuration.
+// This allows an Envoy configuration with multiple HTTP listeners (and
+// associated HTTP connection manager filters) to use different route
+// configurations. Each listener will bind its HTTP connection manager filter to
+// a route table via this identifier.
+service FcdsDiscoveryService {
```
markdroth, line 26: Do we actually need this RPC service?  I think we've agreed to move the xDS API to a model where new resource types can be added and used via ADS without having to add a new RPC service for each one.

rakeshdatta: Hi, @markdroth Thanks for the comment. This is new to me; I was not aware. As I am not sure about this, tagging @adisuissa to comment. 

Also, wondering what is the downside of having new RPC APIs for this new XDS service. In case we don't have this, can u help me understand how we hook the FCDS back-end logic to the ADS RPC APIs?

rakeshdatta: Hi @markdroth @adisuissa Plz suggest if it is still acceptable to have these separate RPC endpoint for FCDS.

P.S. I resolved couple of bugs and handling another one on the draining (only the filter chains, without recreating a new listener) sequence now. I am testing off the file inotify way now, and should get to the grpc transport next.

rakeshdatta:
> 

Hi @jmarantz, Thank you for sharing! I am aware of the rebasing requirement; however, I was actually trying to fix the DCO in the very first commit by rebasing the technique. That is not helping though. In case there is an easier way to the fix the DCO if the very first commit, could u plz suggest? Thanks again!

jmarantz:
I know that it's possible to "repair" a PR with git magic (and losing PR comments in the process). But usually I find it easiest to simply open a new PR, and reference the old one from it. That way any comments in the old PR are not lost.

Maybe the most important thing, though, as you start to contribute to Envoy, is to set up a github hook to automatically add the signed-by line. There are instructions in https://github.com/envoyproxy/envoy/blob/main/DEVELOPER.md

api/envoy/service/filter_chain/fcds.proto
markdroth: From the wire protocol perspective, ADS already supports this; all you need to do is use the new resource type in the ADS stream.

I don't know how this would work from the perspective of Envoy's existing ADS implementation, although I would expect that this would get much easier once Envoy implements the xDS caching layer described in #13009.  @adisuissa can advise further as to how to actually implement this today.

api/envoy/config/listener/v3/listener.proto
rakeshdatta: Hi @kyessenov, @adisuissa 
I gave it a thought. I want to represent a single resource as a filter chain, not a list of filter chains. I think the name that I added in the proto, created confusion, which I could remove as it's not used in any logic.

- For file inotify, much like LDS, I wanted a single listener to point to one FCDS file which has a list of filter chains resources
- Also, for GRPC, I want a listener to be able to subscribe to all(*) or a specific list of filter chains. The Control plane would respond back with a list of resources, where each resource is a filter chain.

In that case, much like LDS proto, I believe the resource type not being a list is fine. 

However, the challenge here is, how could a listener subscribe to a specific list of filter chain updates. I dont see a way set the resource names in the config_source proto. Is there any example of xDS that I can refer to?

ssripadham: > Changing to oneof breaks backwards compatibility. This should be promoted in future major version, and clearly stated in the comments.
> 
> >

@adisuissa  - Can you please show an example how to promote a proto change to a "Future major version"?

PR #20090, opened 2022-02-23
update tracing error tag for grpc status codes

bryanwux:
Signed-off-by: Jiayu Wu [jiayu1.wu@intel.com](jiayu1.wu@intel.com)

Commit Message: Change tracing error tag behaviour for grpc codes [#18877](https://github.com/envoyproxy/envoy/issues/18877)
Additional Description:

Below is the proposed change for tracing error tag behaviour.  A brief description about whether each response code should be indicated as an error has also been listed.
```
OK | 0 | error=false 
CANCELLED | 1 |  error=false Not a server error. The operation is cancelled by the client.
UNKNOWN | 2 |  error=true   Server error. Status value received from unknown address space. 
INVALID_ARGUMENT | 3 | error=false  Not a server error. The client enters an invalid argument. 
DEADLINE_EXCEEDED | 4 | error=true  Server error. Response from the server has been delayed long, even if it may be successful.
NOT_FOUND | 5 | error=false   Not a server error. The requested file or directory is not found.
ALREADY_EXISTS | 6 | error=false   Not a server error. The file or directory the client attempts to create already exists.
PERMISSION_DENIED | 7 | error=false   Not a server error. The client doesn't have permission to execute the operation. 
RESOURCE_EXHAUSTED | 8 | error=false   Not a server error. Some resource has been exhausted like the file system is out of space.
FAILED_PRECONDITION | 9 | error=false   Not a server error. The requested operation is rejected because it doesn't meet preconditions. 
ABORTED | 10 | error=false   Not a server error. The operation is aborted, typically due to concurrency issues.
OUT_OF_RANGE | 11 | error=false   Not a server error. The operation was attempted to read past the valid range.
UNIMPLEMENTED | 12 | error=true   Server error. The service is not implemented or not enabled.
INTERNAL | 13 | error=true   Server error. The system is broken, typically reserved for serious cases. 
UNAVAILABLE | 14 | error=true   Server error. The service is currently not available.
DATA_LOSS | 15 | error=true   Server error. Unrecoverable data loss or corruption.
UNAUTHENTICATED | 16 | error=false   Not a server error. The request does not have valid credentials for the operation.

```
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]: [#18877](https://github.com/envoyproxy/envoy/issues/18877)
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[...]

PR #23098, opened 2022-09-14
tracing: envoy.reloadable_features.update_grpc_response_error_tag deprecation

bryanwux:
https://github.com/envoyproxy/envoy/pull/20090 introduced a runtime guarded feature. It has been 6 months since the new code has been exercised by default, so it's time to remove the old code path.

Signed-off-by: Jiayu Wu <jiayu1.wu@intel.com>

Commit Message:
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes: 
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


test/common/tracing/BUILD
```diff
@@ -8,47 +8,11 @@ licenses(["notice"])  # Apache 2
 
 envoy_package()
 
-envoy_cc_test(
-    name = "http_tracer_impl_legacy_test",
```
daixiang0, line 12: What is the difference between `http_tracer_impl_test` and `http_tracer_impl_legacy_test`?

bryanwux: runtime feature is enabled in  `http_tracer_impl_test `while is not enabled in `http_tracer_impl_legacy_test`. Since runtime guard is now removed, I think we should also remove the legacy test.

source/common/tracing/http_tracer_impl.cc
```diff
@@ -87,42 +87,36 @@ template <class T> static void addGrpcResponseTags(Span& span, const T& headers)
   // Set error tag when Grpc status code represents an upstream error. See
   // https://github.com/envoyproxy/envoy/issues/18877.
   absl::optional<Grpc::Status::GrpcStatus> grpc_status_code = Grpc::Common::getGrpcStatus(headers);
-  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.update_grpc_response_error_tag")) {
```
daixiang0, line 90: The old logic is that
```
if feature
  if
else
```

Now we have enabled the feature, still need the `else` on the top level?

giantcroc: I think we don't need `else` any more, because the feature is always enabled.

bryanwux: True. The logic in `else` is actually the old code path which we don't need anymore.

PR #23108, opened 2022-09-14
http: fixing the EM build broken by a namespacing problem

alyssawilk:


alyssawilk:
Given I tested this locally I'm inclined to force merge once I pass format checks.  Speak now if you object :-P

RyanTheOptimist:
> Given I tested this locally I'm inclined to force merge once I pass format checks. Speak now if you object :-P

SGTM. I might suggest fleshing out the PR description though as it's currently empty.

Any idea why this is broken in EM but fine in Envoy?

RyanTheOptimist (approved):

alyssawilk:
I think EM was pulling this file from Envoy::Extensions::Http and then it looked for Envoy::Extensions::Http::Callbacks instead of Envoy::Http::Callbacks ?

RyanTheOptimist:
> I think EM was pulling this file from Envoy::Extensions::Http and then it looked for Envoy::Extensions::Http::Callbacks instead of Envoy::Http::Callbacks ?

Oh! Yeah, that makes sense. Thanks!

PR #23114, opened 2022-09-14
Make TCP_KEEPALIVE socket options only apply to TCP sockets

### LINKED ISSUE #22663: TCP socket options from upstream cluster config are incorrectly sent down to QUIC
I ran into this issue when I tried speaking QUIC in Envoy-mobile but I think the root cause is from Envoy.

When the [ClusterInfoImpl](https://github.com/envoyproxy/envoy/blob/9141f23079988dc96985d7d5a4992f1849e1e091/source/common/upstream/upstream_impl.cc#L844https://github.com/envoyproxy/envoy/blob/9141f23079988dc96985d7d5a4992f1849e1e091/source/common/upstream/upstream_impl.cc#L844) for HTTP/3 is created, it would fetch the cluster config and potentially parse some [TCP socket options](https://github.com/envoyproxy/envoy/blob/9141f23079988dc96985d7d5a4992f1849e1e091/source/common/upstream/upstream_impl.cc#L106).

Later when we try to create a [QUIC client connection](https://github.com/envoyproxy/envoy/blob/8d513a670006e1ecd06df6c4ef89d4389e102ebb/source/common/quic/envoy_quic_client_connection.cc#L24), the HTTP/3 connection pool would get socket options [from the cluster](https://github.com/envoyproxy/envoy/blob/095f4ca336d3d705e629b207fb2cbbc22d29db8f/source/common/http/http3/conn_pool.cc#L132) and apply them to the underlying UDP socket.

Apparently, socket options like TCPKEEPALIVE are specific to TCP and will fail in QUIC. 

I wonder how we should solve this. Should we move socket option configs out of the cluster config? Or maybe we can add a parameter to specify whether an option can only be applied to a specific socket type?

RyanTheOptimist:
Make TCP_KEEPALIVE socket options only apply to TCP sockets

Fixes #22663

Risk Level: Low
Testing: New integration tests
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/network/socket_impl.cc
+++ b/source/common/network/socket_impl.cc
@@ -23,7 +23,8 @@ SocketImpl::SocketImpl(IoHandlePtr&& io_handle,
                        const Address::InstanceConstSharedPtr& remote_address)
     : io_handle_(std::move(io_handle)),
       connection_info_provider_(
-          std::make_shared<ConnectionInfoSetterImpl>(local_address, remote_address)) {
+          std::make_shared<ConnectionInfoSetterImpl>(local_address, remote_address)),
+      sock_type_(Network::Socket::Type::Stream) {
 
   if (connection_info_provider_->localAddress() != nullptr) {
     addr_type_ = connection_info_provider_->localAddress()->type();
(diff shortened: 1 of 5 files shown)


RyanTheOptimist:
@RenjieTang FYI

alyssawilk (approved):
Nice clean fix!

RyanTheOptimist:
> Nice clean fix!

Thanks! Though annoyingly the initial version failed MSAN because the sock_type_ field of the socket is actually un-initialized in one of the constructors! So I had to fix that. (Thankfully, it's trivial) PTAL.

alyssawilk (approved):

PR #23115, opened 2022-09-14
[QUIC]Change byte counting data type in watermark accounting to explicitly use uint64_t

### LINKED ISSUE #23097: [QUIC]Use of size_t causes arithmetic error
The issue happens [here](https://github.com/envoyproxy/envoy/blob/main/source/common/quic/quic_filter_manager_connection_impl.cc#L136) where we subtract 2 size_t and store the result into a int64_t variable.

When size_t is mapped to uint64_t which is the case for most modern system, the subtraction works ok.
But in 32-bit system, for example x86, size_t is mapped to uint32_t, and the subtraction could result in a large overflow number.

These size_t parameters are actually implicitly casted from [uint64_t](https://github.com/envoyproxy/envoy/blob/main/source/common/quic/send_buffer_monitor.cc#L16), which doesn't seem right.

I think we need to change all the relevant code to use uint64_t instead.

RenjieTang:

Commit Message: Change byte counting data type in watermark accounting to explicitly use uint64_t
Additional Description: when size_t is mapped to uint32_t, int64_t = uint32_t - uint32_t could potentially result in an unsigned wrap around, and causes error.
Risk Level: Low
Testing: unit test
Docs Changes: n/a
Release Notes: n/a
Fixes #23097

### DIFF
--- a/source/common/quic/envoy_quic_stream.h
+++ b/source/common/quic/envoy_quic_stream.h
@@ -25,7 +25,7 @@ class EnvoyQuicStream : public virtual Http::StreamEncoder,
 public:
   // |buffer_limit| is the high watermark of the stream send buffer, and the low
   // watermark will be half of it.
-  EnvoyQuicStream(uint32_t buffer_limit, QuicFilterManagerConnectionImpl& filter_manager_connection,
+  EnvoyQuicStream(uint64_t buffer_limit, QuicFilterManagerConnectionImpl& filter_manager_connection,
                   std::function<void()> below_low_watermark,
                   std::function<void()> above_high_watermark, Http::Http3::CodecStats& stats,
                   const envoy::config::core::v3::Http3ProtocolOptions& http3_options)
@@ -87,7 +87,7 @@ public:
   }
 
   // SendBufferMonitor
-  void updateBytesBuffered(size_t old_buffered_bytes, size_t new_buffered_bytes) override {
+  void updateBytesBuffered(uint64_t old_buffered_bytes, uint64_t new_buffered_bytes) override {
     if (new_buffered_bytes == old_buffered_bytes) {
       return;
     }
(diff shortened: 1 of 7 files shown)


source/common/quic/envoy_quic_stream.h
```diff
@@ -25,7 +25,7 @@ class EnvoyQuicStream : public virtual Http::StreamEncoder,
 public:
   // |buffer_limit| is the high watermark of the stream send buffer, and the low
   // watermark will be half of it.
-  EnvoyQuicStream(uint32_t buffer_limit, QuicFilterManagerConnectionImpl& filter_manager_connection,
+  EnvoyQuicStream(uint64_t buffer_limit, QuicFilterManagerConnectionImpl& filter_manager_connection,
```
danzh2010, line 28: Do we really need this change to fix updateBytesBuffered()? I think in some Envoy interface the buffer_limit is returned as uint32_t.

RenjieTang: No they aren't necessary for this issue.
I've removed them so this change is cleaner.

danzh2010 (approved):
Thanks for the fix!

adisuissa (commented):
/assign @alyssawilk

alyssawilk (commented):
Code LGTM but two questions
How could we end up with negative bytes buffered and can we get a unit test?
/retest

RenjieTang:
> Code LGTM but two questions How could we end up with negative bytes buffered and can we get a unit test? /retest

We don't get negative numbers for buffered bytes. It's `int64_t delta = new_buffered_bytes - old_buffered_bytes;` that has problem. When `new_buffered_bytes` and `old_buffered_bytes` are both uint32_t and `new_buffered_bytes` < `old_buffered_bytes`, then the delta will be a very large number.

PR #23117, opened 2022-09-15
Thrift: Do not reset downstream for partial response

JuniorHsu:

Commit Message:
We've seen upstream connection is closed between response start and response end, which closes downstream connection and leads SR drop.

This disconnection is an intentional behavior from the service since thrift doesn't have a RST semantic.

In this diff, we do local reply to downstream instead of resetting it since we haven't sent any byte to downstream. 

Risk Level: low
Testing: unit

### DIFF
--- a/test/extensions/filters/network/thrift_proxy/router_test.cc
+++ b/test/extensions/filters/network/thrift_proxy/router_test.cc
@@ -729,7 +729,8 @@ TEST_P(ThriftRouterRainidayTest, PoolRemoteConnectionFailure) {
         auto& app_ex = dynamic_cast<const AppException&>(response);
         EXPECT_EQ(AppExceptionType::InternalError, app_ex.type_);
         EXPECT_THAT(app_ex.what(),
-                    ContainsRegex(".*connection failure: remote connection failure.*"));
+                    ContainsRegex(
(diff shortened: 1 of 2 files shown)


test/extensions/filters/network/thrift_proxy/router_test.cc
```diff
@@ -1751,7 +1756,17 @@ TEST_F(ThriftRouterTest, UpstreamDraining) {
 TEST_F(ThriftRouterTest, UpstreamPartialResponse) {
   initializeRouter();
 
-  EXPECT_CALL(callbacks_, resetDownstreamConnection());
+  // EXPECT_CALL(callbacks_, resetDownstreamConnection());
```
zuercher, line 1759: Is there a test that invokes the underflow path?

JuniorHsu: thanks for catch the leftover. yes there is. 

https://github.com/envoyproxy/envoy/blob/c91181872158622de0dadd154d7131c815cdd81d/test/extensions/filters/network/thrift_proxy/router_test.cc#L954

zuercher (approved):

PR #23118, opened 2022-09-15
tcp tunneling: optionally move response header to downstream info

### LINKED ISSUE #23116: tcp_proxy: save response headers for CONNECT upstreams
We have a network telemetry extension that wants to use some metadata passed as a response header for CONNECT tunneling initiated by `tcp_proxy`. We want an ability to expose those headers to network extensions beyond just checking for 200. I have a couple of ideas:
* copy headers to the downstream L4 filter state; maybe require explicit names and mapping;
* copy headers as dynamic metadata to the downstream L4 filter state;
* create a mechanism to make upstream L7 filters somehow access downstream L4 stream info;
* rewrite the telemetry to be an upstream L7 filter instead for CONNECT.

Thoughts? @alyssawilk

kyessenov:

Commit Message: Add support to save response headers from CONNECT tunnels in `tcp_proxy`. Fixes #23116 
Additional Description: The use case is saving "baggage" header which provides additional metadata about the upstream endpoint for telemetry purposes.
Risk Level: low (opt-in)
Testing: unit, integration
Docs Changes: none
Release Notes: yes


source/common/tcp_proxy/upstream.h
```diff
@@ -166,6 +166,7 @@ class HttpUpstream : public GenericUpstream, protected Http::StreamCallbacks {
     // Http::ResponseDecoder
     void decode1xxHeaders(Http::ResponseHeaderMapPtr&&) override {}
     void decodeHeaders(Http::ResponseHeaderMapPtr&& headers, bool end_stream) override {
+      parent_.config_.copyResponseHeaders(*headers, parent_.downstream_info_.filterState());
```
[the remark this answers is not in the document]
kyessenov: Yeah, I think we could do that with little performance impact. One annoying issue is that we don't have a generic way to traverse the filter state that is a map of string to string. That latter issue has been a long standing blocker for replacing dynamic metadata with filter state so maybe we should just fix that as well.

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
```diff
@@ -68,6 +68,16 @@ message TcpProxy {
     option (udpa.annotations.versioning).previous_message_type =
         "envoy.config.filter.network.tcp_proxy.v2.TcpProxy.TunnelingConfig";
 
+    // Defines the mapping from the header value to a filter state object key.
+    message HeaderToFilterState {
+      // Filter state object key, prefixed with ``envoy.tcp_proxy.tunnel_response_header.``.
+      string key = 1 [(validate.rules).string = {min_len: 1}];
```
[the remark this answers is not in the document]
kyessenov: I'll try a complete bag copy, and in that case we can use one key for all of them.

source/common/tcp_proxy/upstream.h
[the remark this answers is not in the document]
alyssawilk: do we need a map of string to string?  we could just create a filter state object "connect_headers" which we std::move the headers into?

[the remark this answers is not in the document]
kyessenov: Yes, that's what I want to do. But then, for example, access log printers can only print it as a whole. It's not a problem for internal filter consumption, just generic inspectors like access log, etc.

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
```diff
@@ -68,6 +68,16 @@ message TcpProxy {
     option (udpa.annotations.versioning).previous_message_type =
         "envoy.config.filter.network.tcp_proxy.v2.TcpProxy.TunnelingConfig";
 
+    // Defines the mapping from the header value to a filter state object key.
+    message HeaderToFilterState {
```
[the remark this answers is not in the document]
kyessenov: Done

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
```diff
@@ -99,6 +99,10 @@ message TcpProxy {
     // Neither ``:-prefixed`` pseudo-headers nor the Host: header can be overridden.
     repeated config.core.v3.HeaderValueOption headers_to_add = 3
         [(validate.rules).repeated = {max_items: 1000}];
+
+    // Emit response headers to the downstream info filter state for consumption
+    // by the network filters. The filter state key is ``envoy.tcp_proxy.tunnel_response_headers``.
+    bool emit_response_headers = 4;
```
alyssawilk, line 105: I'd lean towards something akin to 
save_connect_headers but I'll defer to api shephards on this one :-)

source/common/router/router.h
```diff
@@ -384,6 +384,7 @@ class Filter : Logger::Loggable<Logger::Id::router>,
   const Network::Connection* downstreamConnection() const override {
     return callbacks_->connection().ptr();
   }
+  StreamInfo::StreamInfo* downstreamInfo() const override { return &callbacks_->streamInfo(); }
```
alyssawilk, line 387: This definitely makes me twitchy.
For L7, we have a patterm of when we need to of get upstream information to the downstream stream, we do by putting it in the upstream stream info and copying or moving to the downstream stream info when we associate the connections. I'm concerned if we add this to the API we're going to end up with lifetime issues.
Can instead pass the downstream info as a constructor to HttpConnPool so the addition is limited to TCP proxying?

alyssawilk (commented):
/wait

source/common/router/router.h
kyessenov: Yes, I wasn't happy about it either. Let me explore a better way. FWIW it's already available as immutable via `downstreamConnection` (so the issue is already present if downstream dies before upstream).

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
kyessenov: @markdroth For API review and deciding on naming.

alyssawilk (dismissed):
LGTM (though I still don't love emit - @markdroth ?)
Passing to Greg for !google pass

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
markdroth: How about calling this `propagate_response_headers`?

kyessenov: Renamed.

changelogs/current.yaml
```diff
@@ -223,6 +223,11 @@ new_features:
 - area: redis
   change: |
     added support for quit command to the redis proxy.
+- area: tcp_proxy
+  change: |
+    added support for emitting the response headers in :ref:`TunnelingConfig
```
alyssawilk, line 228: emitting -> propagating ?

alyssawilk (commented):
Yeah I like this better - thanks for your patience with naming
CI looks unhappy?
/wait

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
```diff
@@ -99,6 +99,10 @@ message TcpProxy {
     // Neither ``:-prefixed`` pseudo-headers nor the Host: header can be overridden.
     repeated config.core.v3.HeaderValueOption headers_to_add = 3
         [(validate.rules).repeated = {max_items: 1000}];
+
+    // Save the response headers to the downstream info filter state for consumption
+    // by the network filters. The filter state key is ``envoy.tcp_proxy.tunnel_response_headers``.
```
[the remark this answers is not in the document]
kyessenov: Done.

alyssawilk (approved):
over to @ggreenway !

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
```diff
@@ -99,6 +99,10 @@ message TcpProxy {
     // Neither ``:-prefixed`` pseudo-headers nor the Host: header can be overridden.
     repeated config.core.v3.HeaderValueOption headers_to_add = 3
         [(validate.rules).repeated = {max_items: 1000}];
+
+    // Save the response headers to the downstream info filter state for consumption
+    // by the network filters. The filter state key is ``envoy.tcp_proxy.propagate_response_headers``.
+    bool propagate_response_headers = 4;
```
ggreenway, line 105: Why not do this unconditionally? Performance?

ggreenway (changes requested):
Please add a test validating that headers are not propagated when the new setting is disabled.

/wait

api/envoy/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto
kyessenov: Yes, for long-living streams, we're adding extra memory per stream since the filter state lives till connection closes.

ggreenway (approved):

PR #12586, opened 2020-08-11
udp: Add use_original_src_ip feature on udp proxy

chadr123:
Commit Message:
There is a similar feature for no snat but it only works for tcp case.
The envoy supports filter structure so that we can add or remove the filter dynamically.
But the udp load banalcer has a limitation that can have only one filter.

So, we cannot add more filters on udp load banalcer.
So, the new option is introduced that name is use_original_src_ip on udp_proxy filter.
If it is set as true, all packets that start from envoy can have original source ip address that
same as sender's ip address.

Fixes #12513, #12277

Signed-off-by: DongRyeol Cha <dr83.cha@samsung.com>
Additional Description: None
Risk Level: Low
Testing: bazel test
Docs Changes: Add a new option that name is use_original_src_ip to UdpProxyConfig.
Release Notes: changed current.rst in this PR.

PR #23121, opened 2022-09-15
udp: avoid keeping UDP sockets unconnected

### LINKED ISSUE #21712: Handling ephemeral port exhaustion when using UDP protocol
*Title*: *Handling ephemeral port exhaustion when using UDP protocol*

*Description*:
Hi! I'm using envoy to load balance UDP traffic for service similar to DNS (1 packet with query, 1 packet with response). Everything works fine until traffic burst causes ephemeral port exhaustion and no new UDP sockets can be created. I'm able to delay the problem with `idle_timeout` setting for [UdpProxyConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/udp/udp_proxy/v3/udp_proxy.proto#extensions-filters-udp-udp-proxy-v3-udpproxyconfig) however it's not a long term solution.

*Questions*:
1. Why envoy keeps outgoing UDP sockets in UNCONN state? This causes that SO_REUSEADDR cannot by used and local ports cannot be reused for different upstream IP/PORT tuples.
```
udp   UNCONN      0       0               0.0.0.0:2057           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2062           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2067           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2068           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2070           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2073           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2080           0.0.0.0:*
udp   UNCONN      0       0               0.0.0.0:2081           0.0.0.0:*
[...]
```
2. Is it possible to implement a list of source addresses in [BindConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-bindconfig) to use multiple source ip? This is probably a proper long term solution recomended by HAProxy, Nginx and others.

Thanks for any hints! :)

zhxie:

Commit Message: udp: avoid keeping UDP sockets unconnected
Additional Description:

The patch fixes keeping UDP sockets unconnected and avoids exhausting UDP ports according to the issue.

```sh
# Previous
UNCONN  0       0                     0.0.0.0:10000          0.0.0.0:*       users:(("envoy-static",pid=875931,fd=17))
ESTAB   0       0                   127.0.0.1:52082        127.0.0.1:10000   users:(("nc",pid=875966,fd=3))
UNCONN  0       0                     0.0.0.0:56624          0.0.0.0:*       users:(("envoy-static",pid=875931,fd=16))
ESTAB   0       0                   127.0.0.1:5005         127.0.0.1:56624   users:(("nc",pid=875954,fd=3))

# Current
UNCONN  0       0                     0.0.0.0:10000          0.0.0.0:*       users:(("envoy-static",pid=884658,fd=17))
ESTAB   0       0                   127.0.0.1:47803        127.0.0.1:10000   users:(("nc",pid=884693,fd=3))
ESTAB   0       0                   127.0.0.1:53490        127.0.0.1:5005    users:(("envoy-static",pid=884658,fd=16))
ESTAB   0       0                   127.0.0.1:5005         127.0.0.1:53490   users:(("nc",pid=884681,fd=3))
```

Risk Level: Low
Testing: manual (have no idea to test with CI)
Docs Changes: N/A
Release Notes: N/A
Fixes #21712

### DIFF
--- a/source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
+++ b/source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
@@ -359,7 +359,7 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
     }
   }
   Api::IoCallUint64Result rc =
-      Network::Utility::writeToSocket(socket_->ioHandle(), buffer, local_ip);
+      Network::Utility::writeToSocket(socket_->ioHandle(), buffer, local_ip, *host_->address());
   if (!rc.ok()) {
     cluster_.cluster_stats_.sess_tx_errors_.inc();
   } else {
--- a/source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
+++ b/source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
@@ -351,8 +351,15 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
   //       set. We allow the OS to select the right IP based on outbound routing rules if
   //       use_original_src_ip_ is not set, else use downstream peer IP as local IP.
   const Network::Address::Ip* local_ip = use_original_src_ip_ ? addresses_.peer_->ip() : nullptr;
+  if (!connected_) {
+    Api::SysCallIntResult rc = socket_->connect(host_->address());
+    if (SOCKET_FAILURE(rc.return_value_)) {
+      cluster_.cluster_stats_.sess_tx_errors_.inc();
+      return;
+    }
+  }
   Api::IoCallUint64Result rc =
-      Network::Utility::writeToSocket(socket_->ioHandle(), buffer, local_ip, *host_->address());
+      Network::Utility::writeToSocket(socket_->ioHandle(), buffer, local_ip);
   if (!rc.ok()) {
     cluster_.cluster_stats_.sess_tx_errors_.inc();
   } else {
--- a/source/extensions/filters/udp/udp_proxy/udp_proxy_filter.h
+++ b/source/extensions/filters/udp/udp_proxy/udp_proxy_filter.h
@@ -236,6 +236,7 @@ private:
     // packets from the upstream host. Note that a a local ephemeral port is bound on the first
     // write to the upstream host.
     const Network::SocketPtr socket_;
+    bool connected_{};
 
     UdpProxySessionStats session_stats_{};
     absl::optional<StreamInfo::StreamInfoImpl> udp_sess_stats_;
--- a/source/common/network/utility.cc
+++ b/source/common/network/utility.cc
@@ -509,6 +509,11 @@ Utility::protobufAddressSocketType(const envoy::config::core::v3::Address& proto
 }
 
 Api::IoCallUint64Result Utility::writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
+                                               const Address::Ip* local_ip) {
+  return writeToSocket(handle, buffer, local_ip, *handle.peerAddress());
+}
+
+Api::IoCallUint64Result Utility::writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
                                                const Address::Ip* local_ip,
                                                const Address::Instance& peer_address) {
   Buffer::RawSliceVector slices = buffer.getRawSlices();
--- a/source/common/network/utility.cc
+++ b/source/common/network/utility.cc
@@ -509,11 +509,6 @@ Utility::protobufAddressSocketType(const envoy::config::core::v3::Address& proto
 }
 
 Api::IoCallUint64Result Utility::writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
-                                               const Address::Ip* local_ip) {
-  return writeToSocket(handle, buffer, local_ip, *handle.peerAddress());
-}
-
-Api::IoCallUint64Result Utility::writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
                                                const Address::Ip* local_ip,
                                                const Address::Instance& peer_address) {
   Buffer::RawSliceVector slices = buffer.getRawSlices();
--- a/source/common/network/utility.h
+++ b/source/common/network/utility.h
@@ -334,6 +334,8 @@ public:
   static Api::IoCallUint64Result writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
                                                const Address::Ip* local_ip,
                                                const Address::Instance& peer_address);
+  static Api::IoCallUint64Result writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
+                                               const Address::Ip* local_ip);
 
   /**
    * Read a packet from a given UDP socket and pass the packet to given UdpPacketProcessor.
--- a/source/common/network/utility.h
+++ b/source/common/network/utility.h
@@ -334,8 +334,6 @@ public:
   static Api::IoCallUint64Result writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
                                                const Address::Ip* local_ip,
                                                const Address::Instance& peer_address);
-  static Api::IoCallUint64Result writeToSocket(IoHandle& handle, const Buffer::Instance& buffer,
-                                               const Address::Ip* local_ip);
 
   /**
    * Read a packet from a given UDP socket and pass the packet to given UdpPacketProcessor.


source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
```diff
@@ -351,6 +351,13 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
   //       set. We allow the OS to select the right IP based on outbound routing rules if
   //       use_original_src_ip_ is not set, else use downstream peer IP as local IP.
   const Network::Address::Ip* local_ip = use_original_src_ip_ ? addresses_.peer_->ip() : nullptr;
+  if (!connected_) {
```
soulxu, line 354: seems you never set the `connected_` as true

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
```diff
@@ -351,6 +351,13 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
   //       set. We allow the OS to select the right IP based on outbound routing rules if
   //       use_original_src_ip_ is not set, else use downstream peer IP as local IP.
   const Network::Address::Ip* local_ip = use_original_src_ip_ ? addresses_.peer_->ip() : nullptr;
```
soulxu, line 353: What happened when use original src ip? Since after `connect()`, the socket will a full 4-tuple associated `{source ip, source port, destination ip, destination port}` that source IP can be the different IP than original src one?

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: Oops, fixed.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: The [integration test](https://github.com/envoyproxy/envoy/blob/main/test/extensions/filters/udp/udp_proxy/udp_proxy_integration_test.cc#L286) can guarantee that after `connect()`, `use_original_src_ip` can continue to work.

soulxu: Emm...I didn't see that test set the `use_original_src_ip`. Actually, I didn't find any test in the file set that flag. But correct me if missed something.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
soulxu: Would you be able to add test for it?

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
```diff
@@ -351,6 +351,14 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
...
+  if (!connected_) {
+    Api::SysCallIntResult rc = socket_->connect(host_->address());
+    if (SOCKET_FAILURE(rc.return_value_)) {
+      cluster_.cluster_stats_.sess_tx_errors_.inc();
+      return;
```
soulxu, line 358: nit, should we add a log here?

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: Update unit test to ensure that `connect` will happen once and only once.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: Added, thanks.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: Yes, you are right. The integration test does not really test `use_original_src_ip`. I found that in the original PR, `use_original_src_ip` is tested in the [unit test](https://github.com/envoyproxy/envoy/pull/12586/files#diff-c73fc65a80218e8cb67b41e20b10ec5a2770e66e67a3714a979eb2dee3f1f281).

zhxie: After manual testing, `use_original_src_ip` is conflict with `connect`. I will escape connecting if `use_original_src_ip` is set.

soulxu: thanks for the update, also could you add a test for it?

zhxie: The [unit test](https://github.com/envoyproxy/envoy/blob/6e31c37f19c3fc5566edc4133380a681c96ab6d9/test/extensions/filters/udp/udp_proxy/udp_proxy_filter_test.cc#L792) can guarantee. BTW, we cannot make integration test since `use_original_src_ip` requires `CAP_NET_ADMIN` or root privilege.

zhxie:
I think it will be better to add a runtime guard.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
soulxu: great, thanks, my last request is just add comment here explain why use the `connect()` and why didn't use `connect()` when `use_original_src_ip_`

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
soulxu: great, thanks, my last request is just add comment here explain why use the connect() and why didn't use connect() when use_original_src_ip_

soulxu:
LGTM, except adding more comments in the code

@RyanTheOptimist for maintainer review

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: Added, thanks.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.h
```diff
@@ -236,6 +236,7 @@ class UdpProxyFilter : public Network::UdpListenerReadFilter,
     // packets from the upstream host. Note that a a local ephemeral port is bound on the first
     // write to the upstream host.
     const Network::SocketPtr socket_;
+    bool skip_connecting_{};
```
RyanTheOptimist, line 239: nit: Please add a comment which explains why this would be true vs false and what it controls (if true, does not call `connect()` on the socket).

Also, if it skips `connect()` should it be called `skip_connect_`

RyanTheOptimist (dismissed):
Other than a nit, I think this LGTM. It's too bad, though, that there is no good way to test that we're not exhausting sockets in some integration test :(

RyanTheOptimist:
Looks like this needs a main merge.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.h
zhxie: Fixed, thanks.

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
```diff
@@ -346,11 +346,26 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
...
+  //       socket should not be connected since the source IP will be changed.
   // NOTE: We do not specify the local IP to use for the sendmsg call if use_original_src_ip_ is not
   //       set. We allow the OS to select the right IP based on outbound routing rules if
   //       use_original_src_ip_ is not set, else use downstream peer IP as local IP.
   const Network::Address::Ip* local_ip = use_original_src_ip_ ? addresses_.peer_->ip() : nullptr;
+  if (!use_original_src_ip_ && !skip_connect_) {
+    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.udp_proxy_connect")) {
+      Api::SysCallIntResult rc = socket_->ioHandle().connect(host_->address());
+      if (SOCKET_FAILURE(rc.return_value_)) {
+        ENVOY_LOG(error, "cannot connect: ({}) {}", rc.errno_, errorDetails(rc.errno_));
```
mattklein123, line 361: This could theoretically log spam very badly. I would either switch to debug or use the backoff logger?

mattklein123 (changes requested):

/wait

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
zhxie: Switched to debug, thanks.

soulxu:
/retest

ci fixed

source/extensions/filters/udp/udp_proxy/udp_proxy_filter.cc
```diff
@@ -346,11 +346,26 @@ void UdpProxyFilter::ActiveSession::write(const Buffer::Instance& buffer) {
...
   //       set. We allow the OS to select the right IP based on outbound routing rules if
   //       use_original_src_ip_ is not set, else use downstream peer IP as local IP.
   const Network::Address::Ip* local_ip = use_original_src_ip_ ? addresses_.peer_->ip() : nullptr;
+  if (!use_original_src_ip_ && !skip_connect_) {
+    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.udp_proxy_connect")) {
+      Api::SysCallIntResult rc = socket_->ioHandle().connect(host_->address());
+      if (SOCKET_FAILURE(rc.return_value_)) {
+        ENVOY_LOG(debug, "cannot connect: ({}) {}", rc.errno_, errorDetails(rc.errno_));
+        cluster_.cluster_stats_.sess_tx_errors_.inc();
+        return;
```
mattklein123, line 363: Sorry forgot this before. Can you add test coverage for this case? I don't think it exists? You should be able to mock the syscall failure.

/wait

zhxie: Done, added a unit test for connect error handling.

mattklein123 (approved):
Thanks!

Issue #22758, opened 2022-08-18
Please regenerate RBE toolchain configs


Wyverald:
**Problem**: Envoy RBE is failing on Bazel@HEAD+Downstream pipeline due to https://github.com/bazelbuild/bazel/commit/3469784a4935c91b4ac22bcfed6be52e6dfec878 ([example failure](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2586#0182a995-9720-43d4-85be-7883b6c255a2))

**What needs to be done**: [envoy-build-tools](https://github.com/envoyproxy/envoy-build-tools) needs to be updated to use the latest rbe_config_gen from [bazel-toolchains](https://github.com/bazelbuild/bazel-toolchains) (at least version 5.1.2 so that it contains https://github.com/bazelbuild/bazel-toolchains/commit/8c717f8258cd5f6c7a45b97d974292755852b658). Then the toolchain configs need to be regenerated, and [this file in Envoy](https://github.com/envoyproxy/envoy/blob/835dc5021aead1989a6e72ff165db91b58b127ff/bazel/repository_locations.bzl#L79-L89) needs to be updated so that Envoy uses the latest envoy-build-tools.

phlax:
@Wyverald can you raise a PR in envoy-build-tools to update

mattklein123:
cc @lizan @keith

keith:
working on this here https://github.com/envoyproxy/envoy-build-tools/pull/181

Wyverald:
gentle ping. Envoy remains broken in the downstream pipeline, and if we can't resolve that soon, we'd need to disable Envoy on Bazel CI.

keith:
I submitted https://github.com/envoyproxy/envoy/pull/22822 to unblock bazel CI

keith:
@Wyverald can you confirm that fixes the next run?

Wyverald:
The next run that includes that fix will be tonight CET, so I'll report back tomorrow. Thanks!

Wyverald:
For some reason, Envoy is still broken: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2601#0182d854-183d-4bcd-b15e-f60f99173f16

@aranguyen do you know why that may be? On line 279 of the logs, it says `Inherited 'build' options: [...] --incompatible_use_platforms_repo_for_constraints=false`, yet later the build still fails on `@bazel_tools//platforms:X` usages (line 292).

[...]

PR #23140, opened 2022-09-15
bazel: Add temporary patch for googletest

keith:
This has landed upstream, but updating googletest proved harder than expected. This fixes https://github.com/envoyproxy/envoy/issues/22758 in the meantime.


mattklein123 (approved):
LGTM, thank you.

PR #23145, opened 2022-09-16
stateful_session: Add Header based session state

### LINKED ISSUE #22475: Support for header based stateful session
Currently stateful session supports cookie based [session management](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/stateful_session_filter).

It would be useful to extend this to header based session management as cookies have some assumed default behaviour with browsers and not work well with tabs.

The proposal is to add header based stateful session support in the following ways.
1. Similar to cookie, Envoy can generate the header and encode the host ip:port where caller is expected to pass it on every request. This works similar to cookies except that it is not a special header and hence no default behaviour assumed.
2. Implement an in memory cache (backed by external cache like redis) that Envoy can lookup. 
 
#2 needs more thought on what key should be used and how to provide integration with external cache etc. But #1 is simpler to implement and people who are willing to build client side solution to keep track of the header and pass on every request can benefit from this (This is actually our main use case as we have client specific code that can send the header on every request).

Pawan-Bishnoi:
Commit Message: "stateful_session: Add Header based session state"

Additional Description: 
Envoy already supports [cookie based state management](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto) for sticky sessions. 
This PR adds one more option - header based state management (as discussed [here](https://github.com/envoyproxy/envoy/issues/22475)). 
The header name is configurable via API. 
The responsibility to pass the said header with every request falls on the client.
~~The rest of the behaviour is consistent with the cookie based approach.~~ 
Path match is only available via StatefulSessionPerRoute.

(Reviewers may refer the existing cookie based implementation while reviewing this)

Risk Level: Low
Testing: Unit & Integration tests added.
Docs Changes: Added.
Release Notes: Waiting.
Fixes #22475

API Considerations: 
1. Looking for guidance on [this](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md#where-to-put-new-protos) point. 
I have kept the new proto files in this repo to keep them consistent with the cookie based approach.
2. Could not find any guidance about updating the codeowners file, please see if that needs to be corrected.


api/envoy/type/http/v3/header.proto
```diff
@@ -0,0 +1,25 @@
...
+option java_package = "io.envoyproxy.envoy.type.http.v3";
+option java_outer_classname = "HeaderProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/type/http/v3;httpv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Http header API]
+
+// Header defines an API for obtaining HTTP Header.
+message Header {
```
markdroth, line 17: This seems like a very odd type.  It's called "Header", but it includes both a header name and a path.

This doesn't seem like the sort of thing that will ever be used in any other context.  I think these fields should just be inlined into the `HeaderBasedSessionState` message.

Pawan-Bishnoi: I agree. The idea was to keep it consistent with the `CookieBasedSessionState` [message](https://github.com/envoyproxy/envoy/blob/0d0596315a40f0da502af1a9a885d6def0e5e02e/api/envoy/extensions/http/stateful_session/cookie/v3/cookie.proto#L39).
I can inline this and the updated config would look like this ->
```
http_filters:
 - name: envoy.filters.http.stateful_session
   typed_config:
     "@type": type.googleapis.com/envoy.extensions.filters.http.stateful_session.v3.StatefulSession
     session_state:
       name: envoy.http.stateful_session.header
       typed_config:
         "@type": type.googleapis.com/envoy.extensions.http.stateful_session.header.v3.HeaderBasedSessionState
         name: x-sticky-session-custom-header
         path: /session/v1/
```
Which is alright I guess 🤔 
@wbpcode wdyt?

wbpcode: I think this should be just a special config for this new extension. 🤔 So, we shouldn't add this `Header` to the `type.http.v3`.

And is the path necessary for header based stateful session?

Pawan-Bishnoi: > And is the path necessary for header based stateful session?

Yes, because:
- It keeps the client simple. It can just pass the header every time.
- More importantly, Envoy knows when it is expected to send session info.
    If not path information is present with Envoy then it won't be able to differentiate between:
          1. client not sending the header because it doesn't want stickiness for this call vs 
          2. client not sending the session header because this is the first call.

lizan: > * More importantly, Envoy knows when it is expected to send session info.
>   If not path information is present with Envoy then it won't be able to differentiate between:
>   1. client not sending the header because it doesn't want stickiness for this call vs
>   2. client not sending the session header because this is the first call.

I'm confused, this doesn't seem to be related with `path` here.

source/extensions/http/stateful_session/header/header.h
```diff
@@ -0,0 +1,77 @@
...
+    std::string address = Envoy::Base64::decode(header_value);
+    return !address.empty() ? absl::make_optional(std::move(address)) : absl::nullopt;
+  }
+
+  std::string getHeaderName() const { return name_; }
+
+  const std::string name_;
+  const std::string path_;
+
+  std::function<bool(absl::string_view)> path_matcher_;
```
lizan, line 70: Let's not reinvent the patch_matcher in a filter. This should be achieved by route specific configuration or composite filter.

Pawan-Bishnoi: Composite filter looks exciting! A minor issue though:

The Stateful Session Filter supports both listener level (see [here](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/stateful_session_filter#one-example)) and route level configuration ([here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/stateful_session/v3/stateful_session.proto#extensions-filters-http-stateful-session-v3-statefulsessionperroute))
And this PR is only adding another way to store session state and doesn't aim to change how StatefulSession filter is configured.
Do you think your comment still applies?

api/envoy/type/http/v3/header.proto
Pawan-Bishnoi: For `StatefulSessionPerRoute`, the path isn't needed. Since that is already [per route](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/stateful_session/v3/stateful_session.proto#extensions-filters-http-stateful-session-v3-statefulsessionperroute).
But `StatefulSession` [config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/stateful_session/v3/stateful_session.proto#extensions-filters-http-stateful-session-v3-statefulsession) is much larger in scope, so we need a way to convey when stickiness is expected.

source/extensions/http/stateful_session/header/header.h
lizan: So route level configuration is already there, why do you need a `path` here? You can disable stateful session by that.

Pawan-Bishnoi: That's right. For `StatefulSessionPerRoute`, the path is not needed.
But for `StatefulSession` configuration at Listener level, it is.

My understanding is that:
Stateful Session filter _requires_ two configuration options for any _session storage mechanism_ (such as CookieBasedSessionState)
1. Listener level
2. Route level

Since this PR is only adding another session state storage mechanism (i.e. HeaderBasedSessionState), it has to abide by the laws of StatefulSession filter. 

In short - `HeaderBasedSessionState` must support the Listener level config since `StatefulSession` filter's config dictates that.
@wbpcode to correct me if I misunderstood the API.

Pawan-Bishnoi: discussed offline and it makes total sense. I will make the change.

api/envoy/type/http/v3/header.proto
Pawan-Bishnoi: Done.

Pawan-Bishnoi: Done. Removed the `path`

lizan (commented):
Can you fix CI failures?

/wait

Pawan-Bishnoi:
done @lizan 
Ptal.

lizan (commented):
LGTM, @wbpcode for final merge.

/lgtm api

Pawan-Bishnoi:
@wbpcode ptal.

source/extensions/http/stateful_session/header/header.cc
```diff
@@ -0,0 +1,31 @@
...
+namespace StatefulSession {
+namespace Header {
+
+void HeaderBasedSessionStateFactory::SessionStateImpl::onUpdate(
+    const Upstream::HostDescription& host, Envoy::Http::ResponseHeaderMap& headers) {
+  absl::string_view host_address = host.address()->asStringView();
+  if (!upstream_address_.has_value() || host_address != upstream_address_.value()) {
+    const std::string encoded_address =
+        Envoy::Base64::encode(host_address.data(), host_address.length());
+    headers.addCopy(Envoy::Http::LowerCaseString(factory_.getHeaderName()), encoded_address);
```
wbpcode, line 15: May be `SetCopy` here.

wbpcode (commented):
LGTM overall. Thanks. Only a minor comment.

source/extensions/http/stateful_session/header/header.h
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    auto header_value = hdr[0]->value().getStringView();
+    std::string address = Envoy::Base64::decode(header_value);
+    return !address.empty() ? absl::make_optional(std::move(address)) : absl::nullopt;
+  }
+
+  std::string getHeaderName() const { return name_; }
+
+  const std::string name_;
```
wbpcode, line 59: Sorry. A new minor comment is here. Could we use `Http::LowerCaseString` here directly to avoid creating lower case string repeatedly at runtime.

source/extensions/http/stateful_session/header/header.cc
Pawan-Bishnoi: thank you! That makes much more sense. Changed.

source/extensions/http/stateful_session/header/header.h
Pawan-Bishnoi: done

source/extensions/http/stateful_session/header/header.h
```diff
@@ -54,9 +54,9 @@ class HeaderBasedSessionStateFactory : public Envoy::Http::SessionStateFactory {
     return !address.empty() ? absl::make_optional(std::move(address)) : absl::nullopt;
   }
 
-  std::string getHeaderName() const { return name_; }
+  Envoy::Http::LowerCaseString getHeaderName() const { return name_; }
```
wbpcode, line 57: `const Envoy::Http::LowerCaseString&` to avoid unnecessary copy here.

Pawan-Bishnoi: const won't allow that. Reaching out to discuss offline.

wbpcode: `const Envoy::Http::LowerCaseString& getHeaderName() const { return name_; }`

Pawan-Bishnoi: done

wbpcode (approved):
LGTM. Thanks.

wbpcode (changes requested):
LGTM for almost everything. Thanks. 🌹 
could you also add a release note for this change?

changelogs/current.yaml
```diff
@@ -151,6 +151,10 @@ new_features:
     added an alternative function signature to ``httpCall()`` with ``options`` as an argument. This allows to skip sampling the produced trace span
     by setting ``{["trace_sampled"] = false}`` as the ``options``. And this allows to return multiple header values for same header name by setting
     ``{["return_duplicate_headers"] = true}`` as the ``options``.
+- area: http
+  change: |
+    Added :ref:`HeaderBasedSessionState <envoy_v3_api_msg_extensions.http.stateful_session.header.v3.HeaderBasedSessionState>` to manage
+    :ref:`StatefulSession State <envoy_v3_api_msg_extensions.filters.http.stateful_session.v3.StatefulSession>` via Request Header.
```
wbpcode, line 282: Request Header -> request/response header

wbpcode (approved):
LGTM. Thanks.

Issue #22064, opened 2022-07-07
double continue on data after `inject*DataToFilterChain()`


nareddyt:
### Bug
Currently the filter manager has a bug where it iterates data on a single filter multiple times. This occurs when a filter:

1) Stops iteration of body and trailers.
2) Calls `injectDecodedDataToFilterChain()` to inject body to the next filter (implicitly continuing iteration of the body).
3) Calls `continueDecoding()` to continue iteration of the trailers.

After further debugging, I found the filter manager is calling `decodeData` on the _next_ filter twice, once in step 2 and again in step 3. This leads to data duplication and messes up internal filter state. Same bug exists on encoder path via `injectEncodedDataToFilterChain()`.

### Reproduction
See the new test filter and integration test added in https://github.com/envoyproxy/envoy/pull/22056. This reproduces the error with the test `EndStreamAtTrailersRealisticFilters`. With the original code (please revert changes to `filter_manager.cc` to test), it fails with `upstream_reset_before_response_started{connection_termination}`.

### Attempt to fix
I tried to fix it in https://github.com/envoyproxy/envoy/pull/22056 by introducing simple boolean variables to the StreamDecoderFilter state. It fixes these issues, but it causes issues for other filters. See https://github.com/envoyproxy/envoy/pull/22056#issuecomment-1178254727 for all the other tests that fail when this is introduced.

Frankly, I am not sure why those other tests are failing. I tried stepping through the debugger but I do not have enough knowledge on the other filters to optimally debug.

### Proposal

I am worried the current fix will break other users. **Instead, let's add a new HTTP Decoder Filter API.** This guarantees to not break any pre-existing users and makes the use of the `inject*DataToFilterChain()` clearer. This new API is meant to complement the current `continueDecoding()` API: https://github.com/envoyproxy/envoy/blob/516bc2b0621daffbc783e6038be64a8a0f9ef3d1/envoy/http/filter.h#L344-L355

New API:
```
/**
   * The same as `continueDecoding()`, but only iterates on filter trailers. This routine
   * does not handle headers, metadata, or body data.
   *
   * This routine can only be called
   * if the filter has previously returned StopIteration from decodeTrailers().It
   * is meant to replace `continueDecoding()` -- it is an error to call both routines
   * in an individual filter.
   *
   * This routine should only be used in advanced cases in which a filter calls `injectDecodedDataToFilterChain()`.
   * Since data injection indicates that a filter does not wish to participate in standard HTTP connection manager
[...]

PR #23147, opened 2022-09-16
grpc_json_transcoder: bug fix to allow async continuation of trailers

nareddyt:
Commit Message: 
grpc_json_transcoder: bug fix to allow async continuation of trailers

Additional Description:

We discovered a bug in the encoder path when gRPC JSON transcoder filter is chained with any other filter that does async continuation of trailers. See https://github.com/envoyproxy/envoy/issues/22064 for details.

Root cause: gRPC transcoder uses the `FilterManager` buffer to wait for a full response in the unary gRPC response path. But unfortunately Envoy does not correctly track which filter the buffered data belongs to (@KBaichoo can explain this). This results in the internal buffer being posted to `encodeData` for the wrong filter (i.e. filters before the transcoder in the encode path).

Fix: Modify the transcoder to not use `FilterManager` buffer. It now:
- Uses its own internal buffer
- Returns `StopIterationNoBuffer`
- Adds data from internal buffers during `encodeTrailers`.

Risk Level: Low

Testing: New integration test, updated unit tests.

Docs Changes: None. This is a bug that only occurs when used with other custom extensions.

Release Notes: None. This is a bug that only occurs when used with other custom extensions.

Fixes https://github.com/envoyproxy/envoy/issues/22064

### DIFF
--- a/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
+++ b/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
@@ -647,7 +647,7 @@ Http::FilterDataStatus JsonTranscoderFilter::encodeData(Buffer::Instance& data, 
   }
 
   response_in_.move(data);
-  if (encoderBufferLimitReached(response_in_.bytesStored())) {
+  if (encoderBufferLimitReached(response_in_.bytesStored() + response_out_.length())) {
     return Http::FilterDataStatus::StopIterationNoBuffer;
   }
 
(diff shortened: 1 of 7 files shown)


nareddyt:
@qiwzhang  and @KBaichoo PTAL

qiwzhang:
@nareddyt could you split it into two pr,   one for the actual fix, another for the debug log changes.  There are too many debug log changes, the actual change is buried, very hard to read.

nareddyt:
> @nareddyt could you split it into two pr, one for the actual fix, another for the debug log changes. There are too many debug log changes, the actual change is buried, very hard to read.

Good point, done. PTAL

nareddyt:
Split debug log updates to https://github.com/envoyproxy/envoy/pull/23150

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -647,24 +647,31 @@ Http::FilterDataStatus JsonTranscoderFilter::encodeData(Buffer::Instance& data,
...
+  if (encoderBufferLimitReached(response_in_.bytesStored() + response_out_.length())) {
     return Http::FilterDataStatus::StopIterationNoBuffer;
   }
 
   if (end_stream) {
     response_in_.finish();
   }
 
-  readToBuffer(*transcoder_->ResponseOutput(), data);
+  readToBuffer(*transcoder_->ResponseOutput(), response_out_);
```
qiwzhang, line 658: should we assert response_out_ is empty before readToBuffer()

same for line 699

qiwzhang (approved):
LGTM

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
nareddyt: No, it does not need to be empty, `readToBuffer()` just appends data to the end.

It is not guaranteed to be empty either. If we get multiple `encodeData()` calls, only first call will be empty. Next calls will append data.

lizan (approved):

PR #23160, opened 2022-09-19
Add path_matcher + prefix_rewrite validation

silverstar194:
Commit Message:
Add path_matcher + prefix_rewrite validation

Additional Description:
You should not be able to perform a prefix_rewrite when matching with a path_matcher.

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]

[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/router/config_impl.cc
```diff
@@ -669,6 +669,10 @@ RouteEntryImplBase::RouteEntryImplBase(const VirtualHostImpl& vhost,
         "Specify only one of prefix_rewrite, regex_rewrite or path_rewrite_policy");
   }
 
+  if (!route.route().prefix_rewrite().empty() && path_matcher_ != nullptr) {
```
tyxia, line 672: We should be able to directly use `prefix_rewrite_` that has already been initialized, right? This avoids redundant function call to route() and prefix_rewrite().

PR #23161, opened 2022-09-19
tls: exposing interfaces to tls sockets to remove HTTP/3 dynamic casts

alyssawilk:
Adding TLS and QUIC interfaces UpstreamTransportSocketFactory to avoid dynamic casts for HTTP/3
Adding e2e tests of using HTTP/3 with a http_11_proxy transport socket now the dynamic casts don't break it.

Risk Level: low: minor refactor to data plane
Docs Changes: n/a
Release Notes: n/a


danzh2010:
Is it ready for review? Or still WIP?

envoy/network/transport_socket.h
```diff
@@ -310,6 +318,21 @@ class UpstreamTransportSocketFactory : public virtual TransportSocketFactoryBase
...
+
+  /*
+   * @return the ClientContextConfig, or absl::nullopt for non-TLS factories.
+   */
+  virtual OptRef<const Ssl::ClientContextConfig> clientContextConfig() const { return {}; }
+
+  /*
+   * @return the QuicCryptoClientConfig or nullptr for non-QUIC factories.
+   */
+  virtual std::shared_ptr<quic::QuicCryptoClientConfig> getCryptoConfig() { return nullptr; }
```
danzh2010, line 335: How does this work for FIPS build?

alyssawilk:
good to go - just needed a unit test fix for CI

envoy/network/transport_socket.h
ggreenway: I think it works because of the forward declaration on line 19 of this file. But I'm a little worried about pulling quiche types into the interface definitions. But it's probably fine, given that it's just a forward declaration and not a real type.

ggreenway (approved):

envoy/network/transport_socket.h
alyssawilk: yeah, the compile time options build verifies that the !HTTP3 build is fine :-)

alyssawilk:
I think the new kotlin tests passed, and most of the failures look like fetch failures, so the code is ready for at least a first pass when you have a few!

alyssawilk:
source/extensions/transport_sockets/common is lower than limit of 96.6 (94.4)
/wait

source/common/http/http3/conn_pool.cc
```diff
@@ -155,8 +154,7 @@ allocateConnPool(Event::Dispatcher& dispatcher, Random::RandomGenerator& random_
                             "Creating Http/3 client");
         // If there's no ssl context, the secrets are not loaded. Fast-fail by returning null.
         auto factory = &pool->host()->transportSocketFactory();
-        ASSERT(dynamic_cast<Quic::QuicClientTransportSocketFactory*>(factory) != nullptr);
-        if (static_cast<Quic::QuicClientTransportSocketFactory*>(factory)->sslCtx() == nullptr) {
+        if (factory->sslCtx() == nullptr) {
```
danzh2010, line 157: There is a warning log following this line. Does it still indicate an unexpected behavior now?

source/common/http/http3/conn_pool.cc
```diff
@@ -119,8 +119,7 @@ Http3ConnPoolImpl::createClientConnection(Quic::QuicStatNames& quic_stat_names,
                                           OptRef<Http::HttpServerPropertiesCache> rtt_cache,
                                           Stats::Scope& scope) {
   std::shared_ptr<quic::QuicCryptoClientConfig> crypto_config =
-      dynamic_cast<Quic::QuicClientTransportSocketFactory&>(host_->transportSocketFactory())
-          .getCryptoConfig();
+      host_->transportSocketFactory().getCryptoConfig();
```
danzh2010, line 122: For my own knowledge, the proxy transport socket factory returns null here, then how would the conn pool act upon this return value?

source/common/http/http3/conn_pool.cc
alyssawilk: I assume later code will hard fail if the sslCtx is null - I just changed how we accessed it.

source/common/http/http3/conn_pool.cc
alyssawilk: isn't it a reference, not a pointer?

danzh2010 (commented):
LGTM, along aside fixing the "e3e" test in the PR description.

ggreenway (approved):

Issue #23135, opened 2022-09-15
tools: protoprint.py doesn't handle fully qualified names properly


abeyad:
While working on https://github.com/envoyproxy/envoy/pull/22921, I encountered an issue with the `proto_format` presubmit.  The presubmit check fails in https://github.com/envoyproxy/envoy/pull/22921/files#diff-3aec35493d987a047143be6535109226e876cedb1bb62858b099596b61d267a9 because the proto's fully-qualified name is `.envoy.extensions.config.v3alpha.KeyValueStoreXdsDelegateConfig`, while it contains a field of type `.envoy.config.common.key_value.v3.KeyValueStoreConfig`.  

The `protoprint.py` tool causes the presubmit check to fail because it says the FQN `.envoy.config.common.key_value.v3.KeyValueStoreConfig` should be changed to `config.common.key_value.v3.KeyValueStoreConfig`.  However, if I change it to `config.common.key_value.v3.KeyValueStoreConfig`, compilation fails because the `config` namespace is interpreted relative to the top-level context's namespace of `envoy.extensions.config`, and there is no such thing as `envoy.extensions.config.common.key_value.v3.KeyValueStoreConfig`.

I believe this is a bug in the `protoprint.py` tool, and the tool itself acknowledges that it's tricky to get the FQN normalization correct: https://github.com/envoyproxy/envoy/blob/main/tools/protoprint/protoprint.py#L360-L381.

One option is to fix this bug in the tool.  I'm not sure if there's a good rule we can add to the tool that fixes this issue and doesn't introduce new ones.

Another option is to just allow FQN names.  In chatting with @adisuissa , it seems such conflicts will arise more and more with more code going into `envoy.extensions` and `envoy.contrib`.

Another option is to allow a field to be ignored by the FQN normalizer with some annotation in the comments.

What do you all think?

adisuissa:
I think that what we've seen in the PR falls under "... its root package has a name that collides with one of the packages of the context type", because there's a conflict between "envoy/config" and "envoy/contrib/config", and that might be a place where FQN may be allowed.

If "extensions" and "contrib" were kept in the namespace then FQN could be avoided, but as they are not, there should be a way to differentiate between the packages/types.

abeyad:
> I think that what we've seen in the PR falls under "... its root package has a name that collides with one of the packages of the context type", because there's a conflict between "envoy/config" and "envoy/contrib/config", and that might be a place where FQN may be allowed.
> 
[...]

PR #23163, opened 2022-09-19
tools: Enable FQN in proto fields through an annotation

abeyad:
https://github.com/envoyproxy/envoy/issues/23135 discussed some situations where the proto_format.sh tool normalizes the fully-qualified-name (FQN) of a field, when such normalization actually results in a build error from the compiler.

This PR introduces a new annotation, allow-fully-qualified-name, that can be used to annotate the comments of a field for which the proto_format.sh tool should permit a FQN to be used and not try to normalize the field type.

Risk Level: Low
Testing: Unit test


htuch (approved):
LGTM, thanks!

PR #23168, opened 2022-09-19
http: fixing a bug where internal redirects aren't served on drained connections

alyssawilk:
Previously, if performing an internal redirect on a drained connection, the HCM would checkForDrainedConnections and for HTTP/1.1 downstream, seeing no streams between the close and the recreate, would close the connection before serving the redirect.

Risk Level: low
Testing: new integration test
Docs Changes: n/a
Release Notes: inline


KBaichoo (approved):
lgtm, thanks

mattklein123 (approved):

PR #23169, opened 2022-09-19
router: Fix invalid ASSERT for regex path matcher

yanavlasov:
Additional Description:
The expression for the ASSERT that is firing is incorrect and does not correspond to how Envoy evaluates regex path match:

`ASSERT(path_matcher_->match(path));` in the RegexRouteEntryImpl::rewritePathHeader(...) vs `path_matcher_->match(sanitizePathBeforePathMatching(headers.getPathValue()))` in RegexRouteEntryImpl::matches(...)

Note missing sanitizePathBeforePathMatching() call in the ASSERT statement.

Risk Level: Low, Test Only
Testing: Unit Tests
Docs Changes:
Release Notes:
Platform Specific Features:


adisuissa (approved):
LGTM, thanks!

PR #23175, opened 2022-09-19
Add filesystem cache implementation

ravenblackx:
Commit Message: Add filesystem cache implementation
Additional Description: This is a work in progress; it functions as a cache in this state, but lacks LRU removal.

Doc: https://docs.google.com/document/d/1DWy1TE8RfV_0_fWs28DARS-GblUNuSRBqFxagOi6QPQ/edit
Risk Level: None - new extension so it doesn't do anything if not activated.
Testing: Passes the standard cache test. 100% coverage.
Docs Changes: Adds docs for configuring the new cache implementation.
Release Notes: n/a, still WIP.
Platform Specific Features: Uses the AsyncFile library, therefore no Windows support yet.


source/extensions/extensions_build_config.bzl
```diff
@@ -229,7 +229,8 @@ EXTENSIONS = {
     #
     # CacheFilter plugins
     #
-    "envoy.extensions.http.cache.simple":               "//source/extensions/filters/http/cache/simple_http_cache:config",
+    "envoy.extensions.filters.http.cache.file_system_http_cache": "//source/extensions/filters/http/cache/file_system_http_cache:config",
```
htuch, line 232: Why is this in the filter namespace? Isn't this analogous to simple_http_cache and belongs in the `envoy.extensoins.http.cache` tree?

ravenblackx: There was some discussion about this, and the conclusion as I understood it was that simple http cache was done wrong, because its namespace doesn't correspond to its filesystem location.

I don't know if changing the namespace for simple http cache would make sense (would potentially be a config-breaking change if anyone is using that cache, but seems cleanest), or moving its files to match its namespace and then having this one move correspondingly, or if it's preferred that I just follow suit with the ill-conceived layout now that it's already happened. Or maybe just leaving simple http cache wrong and doing future ones right is best.

htuch: Can you point to this discussion? I don't see this as a filter, more a plugin to a filter, which is really its own extension type.

htuch: I prefer the new structure you propose going forward. If we will grow the number of extensions it will pay off. 

We can fix the source layout to be fully consistent across the two. I'm not sure about the API, the package name is material in the protobug type, so the only way we could migrate simple_http_cache is with a copy+paste+deprecate. I'm not sure that is worth it, so I would just leave it there and document for now.

@mattklein123 to sanity check my reasoning here.

ravenblackx: Sounds good to me. (Though if it was up to me I would just do a breaking change to clean up simple_http_cache, because it's marked WIP [and also terrible], so should be okay to make breaking changes to!)
I'll make that change when I've finished with the test I'm currently writing.

ravenblackx: The practical advantage of the structure I already had is that it makes it naturally live in the same namespace as all the HttpCache stuff it depends on. *Ideally* all that stuff would *also* be in an Envoy::Extensions::HttpCache namespace. I'm not sure how to strike the balance here - for now I'm moving the paths and api but leaving the mismatched namespace.

ravenblackx:
Sorry about force-push - I have this as a two-branch chain in my workspace (add the API, add the impl), so moving the API in the first branch was an insertion of a commit from the perspective of this branch.

source/extensions/extensions_build_config.bzl
mattklein123: I'm +1 on the new structure you propose. Given that simple cache is not actually useful in production, IMO it's fine to just break it if you would be willing to do that work. We would definitely appreciate it! Thank you!

htuch: Yeah, didn't realize it was still WiP. In that case I'm also +1 on the breaking change to clean this up.

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -0,0 +1,69 @@
...
+  // Path at which the cache files will be stored.
+  //
+  // This also doubles as the unique identifier for a cache, so a cache can be shared
+  // between different routes, or separate paths can be used to specify separate caches.
+  //
+  // If the same ``cache_path`` is used in more than one ``CacheConfig``, the rest of the
+  // ``FileSystemHttpCacheConfig`` must also match.
+  //
+  // The path must end with a path separator character (/ or \ per platform.)
+  string cache_path = 2 [(validate.rules).string = {pattern: "[/\\\\]$"}];
```
htuch, line 44: I'm not sure about this path validation criteria. Basically, it feels very specific to this feature, whereas we reference paths elsewhere in the API such as `PathConfigSource` and don't do this. I would advocate for trying to harmonize the path semantics and syntax.

htuch: Well, there is `WatchedDirectory` in there as well. I'm not suggesting to literally use the `PathConfigSource` or `WatchedDirectory`, just that we should have consistent treatment of path validation and constraints.

ravenblackx: After looking into this for a while, I gather that even Windows now supports forward-slash separators (or mixed in the same path), so I guess I can just insert a forward slash for joining if there wasn't one already present. I'll make that change.

ravenblackx: So that's done, and tested, and all the other CI issues are fixed at last.

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -0,0 +1,68 @@
...
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: FileSystemHttpCacheConfig]
+// [#extension: envoy.extensions.http.cache.file_system_http_cache]
+
+// Configuration for a cache implementation that caches in the local file system.
+//
+// By default this cache uses a least-recently-used eviction strategy.
+//
+// Used in the ``typed_config`` field of ``envoy.extensions.filters.http.cache.v3.CacheConfig``.
```
htuch, line 28: Not sure if we need this comment, since the #extension annotation should join with the extension category and provide an experience like https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/cache/v3/cache.proto.html

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -0,0 +1,68 @@
...
+      [(validate.rules).message = {required: true}];
+
+  // Path at which the cache files will be stored.
+  //
+  // This also doubles as the unique identifier for a cache, so a cache can be shared
+  // between different routes, or separate paths can be used to specify separate caches.
+  //
+  // If the same ``cache_path`` is used in more than one ``CacheConfig``, the rest of the
+  // ``FileSystemHttpCacheConfig`` must also match, and will refer to the same cache
+  // instance.
```
htuch, line 42: Is this validated and config rejected if this is not the case?

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -0,0 +1,68 @@
...
+  // If unset there is no limit.
+  google.protobuf.UInt64Value max_cache_entry_count = 5;
+
+  // A number of folders into which to subdivide the cache.
+  //
+  // Setting this can help with performance in file systems where a large number of inodes
+  // in a single branch degrades performance. The optimal value in that case would be
+  // ``sqrt(expected_cache_entry_count)``.
+  //
+  // On file systems that perform well with many inodes, the default value of 1 should be used.
```
htuch, line 66: But if not specified this will be zero. What does that mean?

htuch (commented):
API makes sense modulo comments. @jmarantz can you take an implementation pass? Thanks.

source/extensions/extensions_build_config.bzl
htuch: @ravenblackx are you going to move the simple cache in a separate PR?

jmarantz (commented):
I would love to help review the file-system cache.

However I would like to review a design doc first and then look at code. I have spent considerable time on file-system cache behavior and performance at various times in my career and would like to help here.

However I'd like to understand the goals and strategy of this cache before I dive into the code.

References of note:
 * pagespeed file-system cache: user doc https://www.modpagespeed.com/doc/system#file_cache and source 
 *  example issue with that cache (easy to find on web): https://serverfault.com/questions/592276/pagespeed-cleanup-performance-issue
 *  source https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/kernel/cache/file_cache.cc (and .h next to it, also https://github.com/apache/incubator-pagespeed-mod/blob/master/test/pagespeed/kernel/cache/file_cache_test.cc)
 * @oschaaf 's proposal on how to build a better cache for pagespeed: https://docs.google.com/presentation/d/1frzkktcsGUO9V3pBU0xKSCng94ycSDoMsEbMNCFbqYM/edit#slide=id.ge3afd3915_1_48

In particular, your choice to punt on eviction for now makes me wonder who would use this instead of the simple cache? I assume you can't use a non-evicting cache in production either.

I'm not wedded to any of the design decisions we made in 2010 when building the pagespeed file cache but I (and Otto) might be able to help evaluate some of the choices being made here.

Happy to chat further.

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
ravenblackx: Yes. Should that be explained in the comment, or are you just checking that my usage of 'must' is correct? :)

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
ravenblackx: Same thing - "default value" means what you get if it's not specified, right?
(Zero would be meaningless, so it doesn't seem necessary to force the protocol to make it explicitly 1, or to reject 0.)

Also, the implementation at this stage doesn't use this field at all yet. Not sure if it's necessary to say that explicitly when an extension is still WIP.

ravenblackx:
> However I would like to review a design doc first and then look at code. 

I don't think there's really all that much to design at this stage, it's essentially a minimum viable product of a filesystem cache at this point, but I do have a half-assed design doc lying around anyway! https://docs.google.com/document/d/1DWy1TE8RfV_0_fWs28DARS-GblUNuSRBqFxagOi6QPQ/edit?usp=sharing

> In particular, your choice to punt on eviction for now makes me wonder who would use this instead of the simple cache?
> I assume you can't use a non-evicting cache in production either.

It's not "punt on eviction and start using it", it's "punt on eviction so that we can review the part of it that's implemented and make the eviction part a separate review because this is already a pretty huge thing for review."
The eviction implementation plan (already in progress) is to start with a very simple (maybe too simple) LRU eviction that just does a linear scan over the cache hash table to find the least recently used entry when an eviction is necessary. If it turns out that provokes a bottleneck there are several other options I have in mind, but there's a decent chance that something more complicated would actually be slower than the very simple option in practice, and it certainly would be harder to test and fully understand.

> who would use this instead of the simple cache?

Someone with an instance of a proxy that's more memory constrained than disk-space constrained.

> I'm not wedded to any of the design decisions we made in 2010 when building the pagespeed file cache

Good news, it seems like there's a significant amount in common - e.g. cache metadata in memory (and partially duplicated to the disk cache).

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
ravenblackx: Removed.

jmarantz:
Thanks for the doc; I feel that a lot of the decisions made in the non-evicting version need to be highly informed about how you will do eviction. I'll try to read in the next day or so.

ravenblackx:
> Thanks for the doc; I feel that a lot of the decisions made in the non-evicting version need to be highly informed about how you will do eviction. I'll try to read in the next day or so.

There are additional structures that would be necessary to accommodate different eviction methods, but I don't think any of them would really change this core, only build on top of it. It's gonna want those O(1) lookups regardless of any secondary indexing method.

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
htuch: If you don't plan on using it in this PR, I'd suggest just skipping it in the API or an explicit `[#not-implemented-hide:]`.

jmarantz:
Thanks I read through the doc and left comments and would like to iterate on that a bit before code. There really are a lot of options in front of you even before you actually implement eviction :)

jmarantz:
I'm still hoping to converge fully on the doc before diving into the code :)

/wait

source/extensions/http/cache/file_system_http_cache/cache_entry_file.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+class FileSystemHttpCache;
+
+class CacheEntryFile {
```
jmarantz, line 17: is this class used now? If so, where? (couldn't find refs to it in code in the PR).

jmarantz (commented):
what's up with the "CODEOWNERS" file?

Is there a way to break this up into smaller PRs, maybe bottom up cover each class and its tests?

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
ravenblackx: Tagged things not-implemented-hide.

source/extensions/http/cache/file_system_http_cache/cache_entry_file.h
ravenblackx: Cleaned that up, along with some other cruft, sorry.

ravenblackx:
> what's up with the "CODEOWNERS" file?

Hopefully it's just that I got my name wrong (in most places someone else hasn't taken my name first!) Updated.

> Is there a way to break this up into smaller PRs, maybe bottom up cover each class and its tests?

Not really, not in a way that makes any sense. Unit-testing helper classes like the sub-classes here would only create a bunch of churn with no value. e.g. when I rewrote this from using a memory cache for lookups to not, it would have required rewriting a whole mess of tests... and also the first two PRs would have been like "what is this even for?" to the point that the rest of the chain would still have to be there to make sense of them, and then they would have ended up discarded only *after* the whole chain was written, and then a whole bunch of new zero-value tests would have to be written for the new helper-abstractions, and then all the git-wrangling of trying to do stuff with a chain of PRs where the first ones got replaced. And any such tests would essentially just be "did the code do what the code did" instead of "did the code do what it is supposed to do", like the test for the example cache that ended up incorrectly asserting that a cache must do the incomplete thing. The tests at the filter level are the lowest-level tests that have any value.

I would have preferred to have broken it into smaller pieces, but the first piece (just API) can't pass CI because there's validation that any API must have an implementation, and then you can't work bottom-up from there because the implementation of the API is the top piece. I am not a fan of this model precisely because it essentially demands either monolithic PRs, or small PRs that are initially deadcode and make no sense on their own... but we must work with the tools we have.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,78 @@
...
+uint64_t htonl64(uint64_t n) { return ntohl64(n); }
+
+// in-place byte order swap for a uint32.
+// Optimization in clang or gcc can recognize this as a single `bswap` instruction.
+void flip32(uint32_t& n) {
+  n = ((n & 0xFF000000u) >> 24u) | ((n & 0x00FF0000u) >> 8u) | ((n & 0x0000FF00u) << 8u) |
+      ((n & 0x000000FFu) << 24u);
+}
+uint32_t ntohl32(uint32_t n) {
+#if defined(ABSL_IS_LITTLE_ENDIAN)
```
jmarantz, line 36: ditto -- actually this sounds similar -- is this just an API you can call or would that introduce more portability issues?

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,78 @@
...
+  flip32(n);
+#endif
+  return n;
+}
+uint64_t htonl32(uint32_t n) { return ntohl32(n); }
+} // namespace
+
+// Beginning of file should be "CACH".
+const uint32_t CacheFileFixedBlock::expectedFileId =
```
jmarantz, line 45: contexpr should work here

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,78 @@
...
+    (static_cast<uint32_t>('C') << 24) + (static_cast<uint32_t>('A') << 16) +
+    (static_cast<uint32_t>('C') << 8) + static_cast<uint32_t>('H');
+
+// Next 4 bytes of file should be "0000".
+// Increment this to invalidate old cache files if the format changes.
+// Formatted string-style rather than as an actual int to make it easily human-readable.
+const uint32_t CacheFileFixedBlock::expectedCacheVersionId =
```
jmarantz, line 52: here too

jmarantz (commented):
I'm going to flush a few comments, but I would definitely think you will get more traction with the top-down doc (which you have modulo comments) and a bottom-up impl.

Otherwise `time(CodeReview) = K*2^(# Lines in PR)`

jmarantz:
I thought I'd add Todd to the review for caching semantics. Todd -- I shared the doc with you also. You might want to look at that first.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: Done, and file-scoped it as well.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: Yeah, the import for ntohl is not well standardized unfortunately. Since it compiles down to one inlined assembler instruction it's a zero cost execution-wise, but I fully agree that doing this is annoying for maintenance.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,78 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+namespace {
+// in-place byte order swap for a uint64.
+// Optimization in clang or gcc can recognize this as a single `bswap` instruction.
+void flip64(uint64_t& n) {
```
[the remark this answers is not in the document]
ravenblackx: Done.

source/extensions/extensions_build_config.bzl
ravenblackx: > @ravenblackx are you going to move the simple cache in a separate PR?

Yes, I'm waiting on that until the other precursor PR that touches it (#23717) is merged so I don't get into horrible merge conflicts with myself.

ravenblackx:
> a bottom-up impl broken into more PRs

In order to do that I'd have to write bundles of essentially pointless test cases, because otherwise the small PRs wouldn't be able to pass CI due to coverage constraints (which constraints are overcome by the *meaningful* test cases further up the stack).

If you really think such bundles of pointless test cases would have value then I can grudgingly do it that way. It also makes things awkward because git is not good at "this depends on two branches that don't depend on each other" modeling, which would mean either, e.g. cache_file_fixed_block would have to depend on #23717 even though it doesn't have any connection, or a lot of painful manual wrangling of inverted branch dependency where the dependency *actually* occurs.

`time(CodeReview) = K*2^(# Lines in PR)` seems like a very strange approximation - adding one line doubles the review time? And there would also have to be like 500 additional lines of pointless tests to make the smaller PRs pass CI. Surely just reviewing one file at a time wouldn't be *that* different from doing the same thing in separate PRs?

jmarantz:
RE pointless tests: I think in general we want unit tests for classes in addition to the mid-level tests that show them in context. It's much easier to hit the corner-cases of a lower-level API directly.

RE 2^N formula: it's just an approximation: probably it doesn't double from 1 line to 2 line. But the exponential observation is just that I rarely get a solid block of time where I can focus on reviewing a huge PR. What happens is I get 10 minutes here, 15 minutes there, in between other things. So if I can get through the whole thing in 20m I can do it reasonably efficiently. But if I have to come back and pick up where I left off then typically I have to re-review earlier parts to get context.

However if I have a top-down spec and review the bottom layer first, I can understand the context in which it will be used, but then just rip through the dead simple implementation and its simple tests :) Then the layers on top of that will also be simple.

RE git limitations: yes I agree. Actually I'm not good at advanced source control and I wind up doing a lot of patching and diffing to cobble together pieces from various things. Usually what I like to do is start with a complete implementation which is impossible to review. Then I peel off pieces bottom up and get them reviewed in a different git client(s). Once those merge I do the git pull and merge it in and the weight of that PR decreases over time, at the cost of me having to resolve conflicts if there were changes required by the reviewers. But the benefit is that the pieces are each reviewable (hopefully).

toddmgreer (changes requested):

jmarantz:
/wait

marking as such as when the constituent smaller PRs land  this will either be obsoleted by them or become simple enough to easily review.

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.h
```diff
@@ -0,0 +1,38 @@
...
+CacheFileHeader protoFromHeadersAndMetadata(const Key& key,
+                                            const Http::ResponseHeaderMap& response_headers,
+                                            const ResponseMetadata& metadata);
+CacheFileTrailer protoFromTrailers(const Http::ResponseTrailerMap& response_trailers);
+
+size_t headerProtoSize(const CacheFileHeader& proto);
+
+Buffer::OwnedImpl bufferFromProto(const CacheFileHeader& proto);
+Buffer::OwnedImpl bufferFromProto(const CacheFileTrailer& proto);
+std::string serializedStringFromProto(const CacheFileHeader& proto);
```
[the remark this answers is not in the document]
ravenblackx: I agree about proto.SerializeAsString being preferable - it's here because Envoy CI enforces not using proto.SerializeAsString. I had to explicitly add cache_file_header_proto_util.cc to an exclusions file for this, which I think is preferable to adding several files as exclusions and juggling around if those usages get moved.

bufferFromProto was also added specifically for that reason, more than for clarity - it was added first, then I ended up still needing one case of serializing to not-a-buffer!

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
toddmgreer: Are these any different from absl::ghtonl, ghtonll, gntohl, and gntohll?

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,82 @@
...
+                                    (static_cast<uint32_t>('A') << 16) +
+                                    (static_cast<uint32_t>('C') << 8) + static_cast<uint32_t>('H');
+
+// The expected next four bytes of the header - if cacheVersionId() doesn't match
+// expectedCacheVersionId then the file is from an incompatible cache version and should
+// be removed from the cache.
+// Next 4 bytes of file should be "0000".
+// Increment this to invalidate old cache files if the format changes.
+// Formatted string-style rather than as an actual int to make it easily human-readable.
+constexpr uint32_t expectedCacheVersionId =
```
toddmgreer, line 50: Seems like it would be simpler to just say ` = 0` (or more in future), and use the byteswap functions where needed.

jmarantz: Hey Todd -- I'm going to flip this PR into draft. Per my request, Raven has broken out this particular code into #23847 and the code you are commenting on here has diverged. PTAL at that one. Thank you!

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,74 @@
...
+  size_t bodySize() const { return getUint64(contents_.body_size_); }
+  void setFileId(uint32_t id) { setUint32(contents_.file_id_, id); }
+  void setCacheVersionId(uint32_t id) { setUint32(contents_.cache_version_id_, id); }
+  void setHeadersSize(size_t sz) { setUint32(contents_.header_size_, sz); }
+  void setTrailersSize(size_t sz) { setUint32(contents_.trailer_size_, sz); }
+  void setBodySize(size_t sz) { setUint64(contents_.body_size_, sz); }
+  static size_t offsetToHeaders() { return size(); }
+  size_t offsetToBody() const { return offsetToHeaders() + headerSize(); }
+  size_t offsetToTrailers() const { return offsetToBody() + bodySize(); }
+  absl::string_view stringView() const { return {contents_.as_str_, size()}; }
```
toddmgreer, line 43: This access to contents_.as_str_ is undefined behavior because as_str_ is not the active member of the union. If you ditch the union and have contents_ directly hold the uints, this can be replaced with `return { reinterpret_cast<char*>(&contents_), size()}; }` (one of the rare uses of reinterpret_cast that doesn't lead to UB).

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,74 @@
...
+#include "absl/strings/string_view.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+// This represents a minimal header block on the cache entry file; it uses a struct that
+// is required to be packed, and explicit network byte order, to ensure compatibility.
```
toddmgreer, line 17: I assume that the use of network byte order is to enable writing an entry with one computer, then reading it with a different computer with a different endianness. If so, please explicitly mention this in a comment.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,82 @@
...
+  setTrailersSize(0);
+  setBodySize(0);
+}
+
+void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
+  ASSERT(s.size() == size());
+  safeMemcpyUnsafeSrc(&contents_, s.begin());
+}
+
+uint32_t CacheFileFixedBlock::getUint32(const uint32_t& t) const { return ntohl32(t); }
```
toddmgreer, line 69: I don't see the value of these get/set functions--seems like it would be easier to understand their callers if they just call the ntoh/hton functions directly.

toddmgreer: Please ignore my comments. jmarantz just pointed out that I'm commenting on the stale PR.

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -0,0 +1,74 @@
...
+  //
+  // [#not-implemented-hide:]
+  google.protobuf.UInt64Value max_cache_size_bytes = 3;
+
+  // The maximum size of a cache entry in bytes - larger responses will not be cached.
+  //
+  // If unset there is no limit.
+  //
+  // [#not-implemented-hide:]
+  google.protobuf.UInt64Value max_cache_entry_size_bytes = 4;
```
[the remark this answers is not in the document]
ravenblackx: Clarified in description (includes headers, trailers, metadata but not filesystem overhead).

jmarantz:
Looks like there are still a few CI issues

/wait

source/extensions/extensions_build_config.bzl
ravenblackx: This is done. (#24112)

toddmgreer (approved):

source/extensions/http/cache/file_system_http_cache/lookup_context.h
```diff
@@ -0,0 +1,91 @@
...
+public:
+  FileLookupContext(FileSystemHttpCache& cache, LookupRequest&& lookup)
+      : cache_(cache), key_(lookup.key()), lookup_(std::move(lookup)) {}
+
+  // From LookupContext
+  void getHeaders(LookupHeadersCallback&& cb) final;
+  void getBody(const AdjustedByteRange& range, LookupBodyCallback&& cb) final;
+  void getTrailers(LookupTrailersCallback&& cb) final;
+  void onDestroy() final;
+  // This shouldn't be necessary since onDestroy is supposed to always be called, but in some
```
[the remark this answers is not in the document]
ravenblackx: Is it undefined behavior if the destructor blocks until the other thread has completed? I would expect it to be undefined behavior only when the vtable of the object is downgraded [and since there's no subclass of FileLookupContext, no changes to vtable occur until the *end* of the destructor].

PR #23176, opened 2022-09-19
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/check-deps.yml
+++ b/.github/workflows/check-deps.yml
@@ -6,6 +6,9 @@ on:
 
   workflow_dispatch:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/pr_notifier.yml
+++ b/.github/workflows/pr_notifier.yml
@@ -3,8 +3,15 @@ on:
   schedule:
   - cron: '0 5 * * 1,2,3,4,5'
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   pr_notifier:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+      statuses: read # for pr_notifier.py
+      pull-requests: read # for pr_notifier.py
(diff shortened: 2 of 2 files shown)


adisuissa:
/assign @phlax 
Can you PTAL?

phlax:
this came up before - my understanding is that this is not necessary - iirc it only applies to private repos - im struggling to find the previous conversation - i can follow up to check this

sashashura:
I respectfully disagree, that it applies to private repositories. It is not a vulnerability, but about running workflows with minimal privileges needed. Like not running with sudo unnecessarily. For example the last run of pr_notifier, see below, it does pip install. If any dependency is ever compromised it gets full access to the repository.
image

phlax:
> I respectfully disagree that it applies to private repositories.

ok - that was my vague memory - probably incorrect

checking further - it seems to depend on the default "Workflow permissions" (in `Actions` - `General`)

i think it does make sense to add this for scheduled actions - probably wont hurt for pull requests also

im not sure about `pull_request` because surely you could just change that in the pull request (perhaps it reads from current main in determining permissions)

phlax (approved):
lgtm, thanks @sashashura

PR #23180, opened 2022-09-20
remove unsed member of connection pool container

wbpcode:
Commit Message: remove unsed member of connection pool container
Additional Description:

n/a.

Risk Level: Low.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.
Platform Specific Features: n/a.

PR #23182, opened 2022-09-20
unified drain connection pools in the cluster manager to avoid duplicate code and logic

wbpcode:
Commit Message: unified drain connection pools in the cluster manager to avoid duplicate code and logic
Additional Description:

Commit Message: unified drain connection pools in the cluster manager to avoid duplicate code and logic.

Risk Level: mid. core code update.
Testing: may be some more unit tests are necessary?
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
Waiting for #23180 to be merged.

wbpcode:
Hi, @mattklein123 could you take a look to this PR first when you have free time? Thanks. :) Does this make sense?

This PR mainly does follow changes:
- split close and drain logic more clearly by explicit parameter.
- unified all core logic of draining and closing to single `drainOrCloseConnPools`.
- avoid closing tcp connections when `CLOSE_CONNECTIONS_ON_HOST_HEALTH_FAILURE` is set and the `ClusterManagerImpl::drainConnections` is called.

source/common/upstream/cluster_manager_impl.h
```diff
@@ -519,12 +519,15 @@ class ClusterManagerImpl : public ClusterManager,
                        absl::optional<uint32_t> overprovisioning_factor,
                        HostMapConstSharedPtr cross_priority_host_map);
 
-      // Drains any connection pools associated with the removed hosts.
+      // Drains any connection pools associated with the removed hosts. All connections will be
+      // closed gracefully and new connection is invalid for these hosts.
```
mattklein123, line 523:
```suggestion
      // closed gracefully and no new connections will be created.
```

source/common/upstream/cluster_manager_impl.h
```diff
@@ -519,12 +519,15 @@ class ClusterManagerImpl : public ClusterManager,
...
       void drainConnPools(const HostVector& hosts_removed);
-      // Drains idle clients in connection pools for all hosts.
+      // Drains any connection pools associated with the all hosts. All connections will be
+      // closed gracefully and new connection is invalid for these hosts.
```
mattklein123, line 526:
```suggestion
      // closed gracefully and no new connections will be created.
```

source/common/upstream/cluster_manager_impl.h
```diff
@@ -519,12 +519,15 @@ class ClusterManagerImpl : public ClusterManager,
...
       void drainConnPools();
-      // Drain all clients in connection pools for all hosts.
-      void drainAllConnPools(DrainConnectionsHostPredicate predicate);
+      // Drain any connection pools associated with the hosts that filtered by the predicate.
```
mattklein123, line 528:
```suggestion
      // Drain any connection pools associated with the hosts filtered by the predicate.
```

source/common/upstream/cluster_manager_impl.h
```diff
@@ -560,15 +563,12 @@ class ClusterManagerImpl : public ClusterManager,
...
-    // ConnPoolImplBase::drainConnections() get cleaned up. The code in onHostHealthFailure and the
-    // code in ThreadLocalClusterManagerImpl::drainConnPools(const HostVector& hosts) is very
-    // similar and can be merged in a similar fashion to the ConnPoolImplBase case.
-    void drainConnPools(const HostVector& hosts);
-    void drainConnPools(HostSharedPtr old_host, ConnPoolsContainer& container);
-    void drainTcpConnPools(TcpConnPoolsContainer& container);
-    void drainAllConnPoolsWorker(const HostSharedPtr& host);
+
+    // Drain or close connections of host. If no any drain behavior is provided then closing will
+    // be applied.
```
mattklein123, line 568:
```suggestion
    // Drain or close connections of host. If no drain behavior is provided then closing will
    // be immediate.
```

source/common/upstream/cluster_manager_impl.cc
```diff
@@ -1511,6 +1476,9 @@ void ClusterManagerImpl::ThreadLocalClusterManagerImpl::onHostHealthFailure(
       TcpConnectionsMap& container = it->second;
       container.connections_.begin()->first->close(Network::ConnectionCloseType::NoFlush);
     }
+
```
mattklein123, line 1479: nit: remove newline

source/common/upstream/cluster_manager_impl.h
```diff
@@ -560,15 +563,12 @@ class ClusterManagerImpl : public ClusterManager,
...
+    void drainOrCloseConnPools(const HostSharedPtr& host,
+                               absl::optional<ConnectionPool::DrainBehavior> drain_or_close);
```
mattklein123, line 570: nit: I would recommend s/drain_or_close/drain_behavior. I think it reads a bit better and it becomes more clear that if the optional is not set it means close? Same rename elsewhere.

mattklein123 (changes requested):
Thanks for the cleanup. In general this LGTM. TBH it's really hard to review the change and understand if it's correct since the logic is unfortunately intricate and fragile. Do you feel we have good coverage via existing tests or is there more coverage that we could have? I would imagine the existing tests cover this code pretty well but I'm not 100% sure.

cc @ggreenway who is familiar with this code also.

Just some small nits.

/wait

wbpcode:
This code is covered pretty good by current tests. 

Although I have a worry. This PR change/fix a simple behavior, but no one test was broken. Seem at least I need to add a test for this scene. 🤔 
see
```
avoid closing tcp connections when CLOSE_CONNECTIONS_ON_HOST_HEALTH_FAILURE is set and the ClusterManagerImpl::drainConnections is called.
```

source/common/upstream/cluster_manager_impl.cc
```diff
@@ -1588,49 +1556,32 @@ ClusterManagerImpl::ThreadLocalClusterManagerImpl::ClusterEntry::ClusterEntry(
...
-          Envoy::ConnectionPool::DrainBehavior::DrainExistingConnections);
+      if (drain_or_close.has_value()) {
+        container->pools_->drainConnections(drain_or_close.value());
+      } else {
+        // TODO(wbpcode): in the original implementation, if the
+        // 'CLOSE_CONNECTIONS_ON_HOST_HEALTH_FAILURE' is set, the http connection pool will be
+        // drained rather then be closed. I am not sure why but just keep it.
+        container->pools_->drainConnections(
+            ConnectionPool::DrainBehavior::DrainExistingConnections);
+      }
```
wbpcode, line 1574: I found why we implemented this in this way.

The `CLOSE_CONNECTIONS_ON_HOST_HEALTH_FAILURE` is only supported for TCP connections. And there is no `closeConnections` interface for HTTP pools for now. I will add a comment to record it and then add a test to guard this behavior.

mattklein123: OK cool thanks for the follow up!

mattklein123 (approved):
Thanks!

PR #23183, opened 2022-09-20
Call onDestroy for InsertContext in cache tests

ravenblackx:
Commit Message: Call onDestroy for InsertContext in cache tests
Additional Description: Minor test cleanup; the onDestroy function for InsertContext was not called in tests, which is a potential leak for future cache implementations. Also, calling it correctly improves coverage.
Risk Level: None, test-only.
Testing: Test-only change.
Docs Changes: n/a
Release Notes: n/a


adisuissa (approved):
LGTM, thanks!

Issue #23167, opened 2022-09-19
OAuth2 filter fails a request with Non-ASCII characters


Alexcei88:
There is a listener with Oauth2 filter. 

User is not yet authorized.
User sends a request with non-ASCII characters to Envoy like this

`https%3A%2F%2Freg-portal.ca.dev.kontur.ru%3A4433%2FRC%2FSearch%2FLocal?stateGroup%3D0%26typeGroup%3D0%26isFromForm%3DFalse%26Query%3D%D0%B2%D0%B2%D1%81`

Once user authorize (login and password has been inputed), IDP server calls a callback url according our configuration that looks like

`https://reg-portal.ca.dev.kontur.ru:4433/oauth2/callback?state=https%3A%2F%2Freg-portal.ca.dev.kontur.ru%3A4433%2FRC%2FSearch%2FLocal?stateGroup%3D0%26typeGroup%3D0%26isFromForm%3DFalse%26Query%3D%D0%B2%D0%B2%D1%81&session_state=2f0c1f44-6726-47ca-ae04-9b27ab470403&code=d4245176-0970-4250-b536-62e5eb8cc6c9.2f0c1f44-6726-47ca-ae04-9b27ab470403.bac706d3-bede-49cf-b52d-a38a44c98d3c`

If we'll decode this url we get

`https://reg-portal.ca.dev.kontur.ru:4433/oauth2/callback?state=https://reg-portal.ca.dev.kontur.ru:4433/RC/Search/Local?stateGroup=0&typeGroup=0&isFromForm=False&Query=ввс&session_state=2f0c1f44-6726-47ca-ae04-9b27ab470403&code=d4245176-0970-4250-b536-62e5eb8cc6c9.2f0c1f44-6726-47ca-ae04-9b27ab470403.bac706d3-bede-49cf-b52d-a38a44c98d3c`

Most important here is that a state query parameter is a equals our source url and contains Non-ASCII characters (**Query=ввс**)

The problem is that Envoy cannot initialize `Http::Utility::Url` for provided url.

https://github.com/envoyproxy/envoy/blob/7461266aae623c7a0c9858fe9055aaf37aed95b6/source/extensions/filters/http/oauth2/filter.cc#L347-L353

As result I have the error 

> OAuth flow failed

instead of successfull flow.

adisuissa:
cc @derekargueta @snowp as codeowners

Alexcei88:
[...]

PR #23188, opened 2022-09-20
OAuth2 filter fix handling callback url with Non-ASCII characters

Alexcei88:
Commit Message: Fix handling urls with Non-ASCII symbols on OAuth2 filter
Additional Description: http_parser library was replaced by GURL library in the OAuth2 filter for valudation url
Risk: Medium
Testing: I have tested manually. As for me for testing changes it's enough current oauth2_filter tests 
Release notes: Fixed processing callback url with Non-ASCII characters in the OAuth2 filter
Fix issue #23167

### DIFF
--- a/source/extensions/filters/http/oauth2/BUILD
+++ b/source/extensions/filters/http/oauth2/BUILD
@@ -36,6 +36,7 @@ envoy_cc_library(
         "//source/common/http:utility_lib",
         "//source/common/protobuf:message_validator_lib",
         "//source/extensions/filters/http/oauth2:oauth_callback_interface",
+        "@com_googlesource_googleurl//url",
         "@envoy_api//envoy/config/core/v3:pkg_cc_proto",
     ],
 )
--- a/source/extensions/filters/http/oauth2/filter.cc
+++ b/source/extensions/filters/http/oauth2/filter.cc
@@ -27,6 +27,11 @@
 #include "absl/strings/str_replace.h"
 #include "absl/strings/str_split.h"
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#include "url/gurl.h"
+#pragma GCC diagnostic pop
+
 namespace Envoy {
 namespace Extensions {
 namespace HttpFilters {
@@ -259,13 +264,14 @@ Http::FilterHeadersStatus OAuth2Filter::decodeHeaders(Http::RequestHeaderMap& he
 
       const auto state =
           Http::Utility::PercentEncoding::decode(query_parameters.at(queryParamsState()));
-      Http::Utility::Url state_url;
-      if (!state_url.initialize(state, false)) {
+
+      GURL state_url(state);
+      if (!state_url.is_valid()) {
         sendUnauthorizedResponse();
         return Http::FilterHeadersStatus::StopIteration;
       }
       // Avoid infinite redirect storm
-      if (config_->redirectPathMatcher().match(state_url.pathAndQueryParams())) {
+      if (config_->redirectPathMatcher().match(state_url.PathForRequest())) {
         sendUnauthorizedResponse();
         return Http::FilterHeadersStatus::StopIteration;
       }
@@ -346,8 +352,8 @@ Http::FilterHeadersStatus OAuth2Filter::decodeHeaders(Http::RequestHeaderMap& he
   auth_code_ = query_parameters.at(queryParamsCode());
   state_ = Http::Utility::PercentEncoding::decode(query_parameters.at(queryParamsState()));
 
-  Http::Utility::Url state_url;
-  if (!state_url.initialize(state_, false)) {
+  GURL state_url(state_);
+  if (!state_url.is_valid()) {
     sendUnauthorizedResponse();
     return Http::FilterHeadersStatus::StopIteration;
   }


mattklein123:
Can you provide more context around this change? This is huge and it sounds quit scary from a security perspective. Thank you.

/wait

Alexcei88:
It's very strange why it's showed so many commits and changed files. I have changed only 2 files. May be I'll recreate PR

Alexcei88:
It's better now. You can look one commit with a small changes.

mattklein123 (changes requested):
Unfortunately I don't understand the full implications of this change so I'm going to tag @yanavlasov who is more familiar with our plans around URL parsing for further guidance.

This change definitely needs test and a release note also, thanks.

/wait

yanavlasov:
But non-ASCII characters are not allowed in the URI path accprding to https://datatracker.ietf.org/doc/html/rfc3986#section-3.3 ?

Are you trying to fix this for an application that sends malformed URI in the request?

/wait-any

Alexcei88:
@yanavlasov , No, no. Did you see a bug description?

I agree that non-ASCII characters are not allowed in the URI path. But all Non-ASCII symbols encode by URL encode according the same spec(https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) . Any application can send Unicode symbols in URI using by HTML URL Encoding. My application sends all right. 

The OAuth2 filter receives URI form only ASCII symbols. But after decoding URL to string, the string already contains Non-ASCII symbols, they were decoded. 
And in this moment, http_parser library already cannot parser such string properly. GURL library is OK. It works good with URL with Non-ASCII symbols.

yanavlasov:
I see. This is a bug in the filter. the `Http::Utility::PercentEncoding::decode` decodes every single character, but in fact only characters from the reserved and unreserved character sets should be decoded, if I understand it correctly.

I do not think this is the right change, since it will produce an incorrect redirect URL with unescaped non-ASCII characters.

The right fix is to add a new `Http::Utility::PercentEncoding::decodeURL` function that only decodes characters from the [reserved](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) and [unreserved](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3) sets and leaves all other encoding sequences as is.

Alexcei88:
I don't mind your suggestion. But I am not clear how to implement such function. Do you have a chance to create such function? I can just revert my changes and replace to your function and test after that

Alexcei88:
Also note that in the filter implementation we compare URL object with URL from setting of the filter. If URL will contain non-decoded characters in this case user need to know about it and have to set HTML encoding URL in the filter settings.
I think it's no user friendly.

yanavlasov:
I unfortunately do not have time to work on fixing this, but I can help you implement it.
You can see how the `Http::Utility::PercentEncoding::decode` function is implemented here: https://github.com/envoyproxy/envoy/blob/8285fca7c3da4b0ab9aa35e3febab659164ba987/source/common/http/utility.cc#L1024

The Http::Utility::PercentEncoding::decodeURI is the same but with additional check here https://github.com/envoyproxy/envoy/blob/8285fca7c3da4b0ab9aa35e3febab659164ba987/source/common/http/utility.cc#L1045

to see if `ch` is in either reserved or unreserved sets according to RFC: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2

if it is - it should be decoded, if it is not in either of these sets `ch` should be set to `%` and `i += 2;` is skipped.

Then you can add unit tests similar to how it is done here: https://github.com/envoyproxy/envoy/blob/8285fca7c3da4b0ab9aa35e3febab659164ba987/test/common/http/utility_test.cc#L1614 in you case you can just decode a string with every ASCII character %encoded and check that only expected characters were decoded.

We can then go through making changes in the o-auth filter.

I have also added current owners @derekargueta @snowp of this filter, in case they would want to take this on.

Alexcei88:
@yanavlasov , could you provide test input and output example? I am not sure that I clear undestand you what I need to do

yanavlasov:
The [standard](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) allows query to contain: `unreserved | pct-encoded | sub-delims | : | @ | / | ?` This allows 4 characters out of the gen-delims set to be unencoded and only requires the `#[]` characters to be encoded.

The bigger question is if the redirect URI also contains query string. I do not know if OAuth allows query and fragment in the redirect URL. If it does, it means that the key/value separators in the redirect URL also have to be %-encoded to prevent incorrect interpretation. 

If query and fragment are allowed in the redirect URL and assume that query key/value separators are = and &, then the test strings are:

input: `%00%01%02 ...<replace ... with %-encoded sequences for all intervening characters> %FE%FF%`
output: `%00%01%02...%22#%24%25&%27...%3C=%3E...%5A[%5C]%5E...%FE%FF%`

This is my best approximation for how URL should be encoded as part of a query parameter.

Alexcei88:
I dont' have time to reedit changes. I hope I will find a time for this

Alexcei88:
I had have been more time to think about your suggestion. I think my changes are correct. I am trying to explain.

You suggest to create a special decode function. I think there are already two function to encode and decode URL and they works excellent. Why you need to create third function for URL? You suggest to make a partial decoding in new function. I think it's strange.

Beside that, I don't produce non valid URL. The filter decodes the URL. The Http::Utility::PercentEncoding::decode function works right according specification. After decoding the redirect url is true by structure. GURL library confirms that this is a valid URL unlike the Http::Utility::Url. 
We works with decoded URL in the filter. The redirect url will be encoded back by the filter at the end.

The GURL library is recieved just a plain URL. The url is not needed to be encoded. The Http::Utility::Url also need to work with any decoded URL true by structure, but it's not. And it's main problem

Alexcei88:
@yanavlasov, @mattklein123, I think my changes are perfect and fix the problem. I told above why I think so

Alexcei88:
I am busy to actualize code

yanavlasov:
This PR may be closed in favor of the #25278

PR #23189, opened 2022-09-21
listener: support update socket_options

### LINKED ISSUE #22634: The update of v4_compat flag for the listener will be ignored
When updating a listener with a different v4_compat flag value, the new value will be ignored. When updating an existing listener, we will clone the socket from the old listener:
https://github.com/envoyproxy/envoy/blob/01fb798af4276da0e4b93ac1a60287fc477598ad/source/server/listener_manager_impl.cc#L505

And the old v4_compat flag value will be clone also. Then the new value is lost. 

The possible fix are:

* changing the `ListenerImpl::hasCompatibleAddresses`

https://github.com/envoyproxy/envoy/blob/01fb798af4276da0e4b93ac1a60287fc477598ad/source/server/listener_impl.cc#L1047

Then compare the v4_compat flag, if the value is different, then consider that as non-compatible address, then will create a new socket for the listener update. This would be an easy fix but this shouldn't work for `reuse_port` with `false`.

* updating the v6only flag and `IPV6_V6ONLY` option on the old socket, that will change the sematic of clone
https://github.com/envoyproxy/envoy/blob/01fb798af4276da0e4b93ac1a60287fc477598ad/source/server/listener_impl.h#L72-L74

Pass the new v4_compat value to the clone, and generate new address instance and set the new value for `IPV6_V6ONLY` socket option. But not sure this option can be changed for already bind socket. Even it can be changed, the old listener loose the v4 compact capabilities before new listener warmed.

Also not sure there is really someone who want to update the `v4_compat` value. So the first question should be whether we want to support this? if not we can reject this directly also.

soulxu:

Commit Message: listener: support update socket_options
Additional Description:
Support to update socket_options through create a new socket

Risk Level: high
Testing: unittest added
Docs Changes: API doc
Release Notes: a behavior change item added
Fixes #22634


soulxu:
@mattklein123 @ggreenway would you like to give a high-level review on this? I didn't add any tests yet but I did a manual test. And there are some checks that should be added for rejecting the update non-STATE_LISTENING socket options when using clone.

@kfaseela if you are interested, you can have a test, and ensure this is what are you looking for.

api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -342,6 +342,15 @@ message Bootstrap {
...
   core.v3.TypedExtensionConfig xds_delegate_extension = 35;
+
+  // There are two behaviors can be chosen when update the `socket_options` on the existing listener,
+  // the default behavior is
+  // creating a new socket when the socket_options updated. Another behavior is clone the existing socket
+  // when the socket_options updated, and only the socket option which is set at `STATE_LISTENING` state
+  // can be updated. Whether a socket option can be updated, it depends on the platform. Envoy doesn't
+  // have knowledge about that. The user should ensure the socket update behavior on the platform first,
+  // ensure it works as expected.
+  bool clone_socket_when_options_updated = 36;
```
soulxu, line 353: stepping back a little bit, and asking again, should we support updating option through clone socket? is it just creating a new socket when the option update enough?

I guess this option was purposed for providing another option for the user if creating a new socket with new options doesn't work
https://github.com/envoyproxy/envoy/issues/22634#issuecomment-1247349596, right?

soulxu: note that creating a new socket with `enable_reuse_port` as false won't work, but I guess most people using `enable_reuse_port` as true today. I should update this in the comment

mattklein123, line 353: Just from an API perspective, do we need a configuration option for this? Couldn't we do something like:
1) Find the difference in socket options during an update.
2) Based on the difference, ask the platform whether they can be updated or not. The default answer can be no, in which case the config update is NACKd to make it more clear what is happening. Then known safe options or sets of options can be added later?

What happens today? Are the updates just always dropped because we can't do the updates? Or does it depend on reuse_port vs. not and what the exact options are? I think some summary of the current state would be helpful. Thank you!

mattklein123 (commented):
Thanks for working on this. A high level comment to get started.

api/envoy/config/bootstrap/v3/bootstrap.proto
soulxu: > Just from an API perspective, do we need a configuration option for this? Couldn't we do something like:

If we are going to only support Linux platform, then I prefer to not add this config option, and we just always create new socket and apply the new options. And that avoids adding a lot of complexity to the code.

But as the doc said https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#config-listener-v3-listener, the Linux platform is the only platform that work well with reuse port. If not set reuse port, we aren't able to create a new socket since the port bind conflict.

If we are going to support all the platform, then clone is the only choice can work on the all platforms. Probably we needn't add this config option also, since the create new socket has a downside as you comment in the code:
https://github.com/envoyproxy/envoy/blob/ac5792e1da4c966656f4afc7a184fb39644f24fb/source/server/listener_impl.cc#L128, then clone will become the better option here.

> 
> 1. Find the difference in socket options during an update.

For creating new socket, we needn't do that, we can just apply all the socket_options in the new listener config directly. That is pretty simple.

For clone the existing socket, I think there are two options: First option, we could find out the difference in the socket options. but it is more complex, like the user could remove a socket option, then we should set this option back to the default value. Then another problem shows up, that is collecting all the default value for all the option on the all the platform.  Second option is more simple, we just apply all the new socket options on the cloned socket, asking the user explicitly set an option back to the default value when the user wants to remove an option. of course, this is based on the all socket options update is idempotent, although I didn't see any doc clear about it, but I think why it isn't

> 2. Based on the difference, ask the platform whether they can be updated or not. The default answer can be no, in which case the config update is NACKd to make it more clear what is happening. Then known safe options or sets of options can be added later?

For creating new socket, we needn't care about this. Since it just applies all socket options in the new listener config.

For clone existing socket, we probably needed a rollback mechanism, if the platform reject the update, then we rollback. Or we also can leave the responsibility to the user, asking the user test the socket option update behavior on the platform first.

> 
> What happens today? Are the updates just always dropped because we can't do the updates? Or does it depend on reuse_port vs. not and what the exact options are? I think some summary of the current state would be helpful. Thank you!

Today, we will clone(duplicate) the existing socket when update listener, the socket options are cloned from the old listener also. Then the change in the new listener config is just ignored.
https://github.com/envoyproxy/envoy/blob/ac5792e1da4c966656f4afc7a184fb39644f24fb/source/server/listener_impl.cc#L122

To summarize the solution here:
* only support creating a new socket option. The implementation is simple. The downside is only supported Linux platform, does not support reuse_port is false. And https://github.com/envoyproxy/envoy/blob/ac5792e1da4c966656f4afc7a184fb39644f24fb/source/server/listener_impl.cc#L128
* only support clone existing socket option. support all the platforms, and support both the resuse_port is true and false. The downside is adding a lot of complexity to the code. But we could reduce the complexity by dropping some responsibilities if that are acceptable.

mattklein123: Thanks for the detailed explanation. IMO this is super complicated and probably not worth the effort to do something even more complicated. I would probably:
1) Only support changing socket options when reuse_port is true, so effectively only on Linux currently.
2) NACK the config update in all other situations of the socket options change so it's more clear what is going on.

Can we start there?

soulxu: thanks! yes, we can start here, that is what I prefer too.

soulxu:
@mattklein123 we have three listener fields for three specific socket options, they are `transparent`, `freebind` and `tcp_fast_open_queue_length`, I think they should work with an update. Do we want to support updating those fields if they work? If we reject that, but people still can do a workaround by setting those options in the `socket_options` manually, then they got the ability to update them.

soulxu:
> @mattklein123 we have three listener fields for three specific socket options, they are `transparent`, `freebind` and `tcp_fast_open_queue_length`, I think they should work with an update. Do we want to support updating those fields if they work? If we reject that, but people still can do a workaround by setting those options in the `socket_options` manually, then they got the ability to update them.
> 
feel `freebind`'s update is meanless, if people change its value, then the address should be changed also.

soulxu:
/wait

the duplicate address check isn't right, working on a fix

mattklein123:
> @mattklein123 we have three listener fields for three specific socket options, they are transparent, freebind and tcp_fast_open_queue_length, I think they should work with an update. Do we want to support updating those fields if they work? If we reject that, but people still can do a workaround by setting those options in the socket_options manually, then they got the ability to update them.

I would support update if possible so it's not surprising. cc @ggreenway

api/envoy/config/listener/v3/listener.proto
```diff
@@ -233,7 +233,10 @@ message Listener {
   google.protobuf.BoolValue freebind = 11;
 
   // Additional socket options that may not be present in Envoy source code or
-  // precompiled binaries.
+  // precompiled binaries. The socket options can be updated for a listener and
+  // it only supports when the
+  // :ref:`enable_reuse_port <envoy_v3_api_field_config.listener.v3.Listener.enable_reuse_port>`
+  // is `true`.
```
mattklein123, line 239:
```suggestion
  // precompiled binaries. The socket options can be updated for a listener
  // only when
  // :ref:`enable_reuse_port <envoy_v3_api_field_config.listener.v3.Listener.enable_reuse_port>`
  // is `true`. Otherwise, if socket options change during a listener update the updated will be NACKd 
  // to make it clear that the options were not updated.
```

We should also put the new behavior behind a runtime flag as this might cause people issues if they didn't know they were ignoring changes to configs.

changelogs/current.yaml
```diff
@@ -5,6 +5,10 @@ behavior_changes:
 - area: build
   change: |
     official released binary is now built on Ubuntu 20.04, requires glibc >= 2.30.
+- area: listener
+  change: |
+    Previously the listener update with different :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` will be ignored. Now, when the :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` updated, a new socket will created for the listener and
+    apply the new :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` to it.
```
mattklein123, line 11:
```suggestion
    Previously a listener update with different :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` was ignored. Now, when the :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` are updated, a new socket will be created for the listener and the new :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` applied to it. This only happens when reuse_port is true. Otherwise if socket options change the update is NACKd.
```

Similar note about the runtime flag.

source/server/listener_impl.cc
```diff
@@ -1123,6 +1135,30 @@ void ListenerImpl::closeAllSockets() {
   }
 }
 
+bool ListenerMessageUtil::compareSocketOptions(const envoy::config::listener::v3::Listener& lhs,
+                                               const envoy::config::listener::v3::Listener& rhs) {
```
mattklein123, line 1139: Per my other comments in the docs, I think we should NACK the update if reuse_port is false and the options change? Also I think this behavior should be behind a runtime flag.

Also, can we try to have an integration test for both reuse_port true and false in the LDS integration tests for changing socket options?

mattklein123 (changes requested):
Awesome to see this fixed. Some high level comments to get started, thank you.

/wait

source/server/listener_impl.cc
soulxu: I plan the platform to reject that when binding the listen socket, then we needn't add more code to the envoy, when you update with reuse_port as false, you will get something as below:
```
[2022-09-28 00:55:38.401][65125][error][config] [source/server/listener_manager_impl.cc:1083] listener 'my_listener' failed to bind or apply socket options: cannot bind '0.0.0.0:13333': Address already in use
[2022-09-28 00:55:38.402][65125][warning][config] [source/common/config/filesystem_subscription_impl.cc:60] Filesystem config update rejected: Error adding/updating listener(s) my_listener: cannot bind '0.0.0.0:13333': Address already in use
``

Let me know if we prefer to reject in Envoy instead of platform.

I will try to add integration tests.

thanks!

Comment (soulxu) [reply source/server/listener_impl.cc:1139]: it turns out, we can't test reuse_port as false, since our test is listening on zero port address, so it won't fail due to conflict address when binding port.

Comment (soulxu) [reply source/server/listener_impl.cc:1143]: that makes sense, the message isn't very clear for the user. reject that case should be easy, let me add it. thanks

File: source/server/listener_manager_impl.cc [hunk 5]
```diff
@@ -423,6 +423,33 @@ bool ListenerManagerImpl::addOrUpdateListener(const envoy::config::listener::v3:
   return false;
 }
 
+void ListenerManagerImpl::setupSocketFactoryForListener(ListenerImpl& new_listener,
+                                                        const ListenerImpl& existing_listener) {
+  bool same_socket_options = true;
+  if (Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.enable_update_listener_socket_options")) {
+    if (new_listener.reusePort() != existing_listener.reusePort()) {
+      incListenerCreateFailureStat();
```
soulxu, line 432: I added this for the integration test can know the listener update failed https://github.com/envoyproxy/envoy/pull/23189/files#diff-7fe1ad3a088b5d0ea44d9bace0dfa0675d5009bcb538afb778ebad50c802ddeaR1489, But I feel this stat definition isn't very clear for now, it doesn't increase in many listener update failed case.

soulxu:
/wait

still have test need to fix

test/integration/listener_lds_integration_test.cc
```diff
@@ -1356,6 +1356,146 @@ TEST_P(ListenerFilterIntegrationTest, UpdateListenerFilterOrder) {
   tcp_client3->close();
 }
 
+#if !defined(__APPLE__) && !defined(WIN32)
```
soulxu, line 1359: I should add a test for the mac and windows

soulxu: emm...no, `UpdateListenerWithDifferentSocketOptionsWhenReusePortDisabled` is enough, there should be no different to test on explicit set enable_reuse_port as false or test on the default value of enable_reuse_port as false.

test/integration/listener_lds_integration_test.cc
```diff
@@ -1356,6 +1356,146 @@ TEST_P(ListenerFilterIntegrationTest, UpdateListenerFilterOrder) {
...
+  tcp_client->close();
+
+  auto* socket_option = listener_config_.add_socket_options();
+  socket_option->set_level(IPPROTO_IP);
+  socket_option->set_name(IP_TOS);
+  socket_option->set_int_value(8); // IPTOS_THROUGHPUT
+  socket_option->set_state(envoy::config::core::v3::SocketOption::STATE_LISTENING);
+  ENVOY_LOG_MISC(debug, "listener config: {}", listener_config_.DebugString());
+  sendLdsResponse({MessageUtil::getYamlStringFromMessage(listener_config_)}, "2");
+  test_server_->waitForCounterGe("listener_manager.listener_create_failure", 1);
```
soulxu, line 1489: Is there a way I can't ensure the listener update reject for a specific reason?

api/envoy/config/listener/v3/listener.proto
```diff
@@ -233,7 +233,10 @@ message Listener {
   google.protobuf.BoolValue freebind = 11;
 
   // Additional socket options that may not be present in Envoy source code or
-  // precompiled binaries.
+  // precompiled binaries. The socket options can be updated for a listener when the
```
mattklein123, line 236:
```suggestion
  // precompiled binaries. The socket options can be updated for a listener when
```

api/envoy/config/listener/v3/listener.proto
```diff
@@ -233,7 +233,10 @@ message Listener {
...
+  // :ref:`enable_reuse_port <envoy_v3_api_field_config.listener.v3.Listener.enable_reuse_port>`
+  // is `true`. Otherwise, if socket options change during a listener update the updated will be rejected.
```
mattklein123, line 238:
```suggestion
  // is `true`. Otherwise, if socket options change during a listener update the update will be rejected
```

changelogs/current.yaml
```diff
@@ -5,6 +5,18 @@ behavior_changes:
 - area: build
   change: |
     official released binary is now built on Ubuntu 20.04, requires glibc >= 2.30.
+- area: listener
+  change: |
+    Previously a listener update with different :ref:`transparent <envoy_v3_api_field_config.listener.v3.Listener.transparent>`, :ref:`freebind <envoy_v3_api_field_config.listener.v3.Listener.freebind>`,
+    :ref:`tcp_fast_open_queue_length <envoy_v3_api_field_config.listener.v3.Listener.tcp_fast_open_queue_length>` or :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` was ignored.
+    Now, when those fields are updated, a new socket will be created for
+    the listener and the new value of those fields applied to it. This only happens when reuse_port is true.
```
mattklein123, line 13: nit ref link to reuse_port

changelogs/current.yaml
```diff
@@ -5,6 +5,18 @@ behavior_changes:
...
+    Otherwise if those fields change the update is rejected.
+    New runtime flag ``envoy.reloadable_features.enable_update_listener_socket_options`` is added for revert this behavior.
+- area: listener
+  change: |
+    Previously a listener update with different :ref:`enable_reuse_port <envoy_v3_api_field_config.listener.v3.Listener.enable_reuse_port>` value will be ignored. Now,
+    this kind of update will be rejected. Also the runtime flag ``envoy.reloadable_features.enable_update_listener_socket_options`` can be used for revert this behavior.
```
mattklein123, line 19:
```suggestion
    this kind of update will be rejected. The runtime flag ``envoy.reloadable_features.enable_update_listener_socket_options`` can be used for revert this behavior.
```

source/server/listener_impl.cc
```diff
@@ -1123,6 +1145,39 @@ void ListenerImpl::closeAllSockets() {
...
+        found = true;
+        break;
+      }
+    }
+    if (!found) {
+      return false;
+    }
+  }
+  return true;
+}
```
mattklein123, line 1179: I don't feel very strongly about it, but you could probably make this simpler by just doing a direct comparison vs. an unordered comparison. Since it's an array the order shouldn't change and I don't personally think it's worth checking for unordered equal but up to you.

source/server/listener_impl.cc
```diff
@@ -1090,6 +1103,15 @@ bool ListenerImpl::hasCompatibleAddress(const ListenerImpl& other) const {
 }
 
 bool ListenerImpl::hasDuplicatedAddress(const ListenerImpl& other) const {
+  // Skip the duplicate address check if this is the case of a listener update with new socket
+  // options.
+  if (Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.enable_update_listener_socket_options") &&
```
mattklein123, line 1109: Since this string is used in multiple places, can you define it once in some constexpr? There are a few others that are defined like that that you can look at.

source/server/listener_manager_impl.cc
```diff
@@ -423,6 +423,33 @@ bool ListenerManagerImpl::addOrUpdateListener(const envoy::config::listener::v3:
...
+      throw EnvoyException(
+          fmt::format("Listener {}: reuse port doesn't support change", new_listener.name()));
```
mattklein123, line 434:
```suggestion
          fmt::format("Listener {}: reuse port cannot be changed during an update", new_listener.name()));
```

source/server/listener_manager_impl.cc
```diff
@@ -423,6 +423,33 @@ bool ListenerManagerImpl::addOrUpdateListener(const envoy::config::listener::v3:
...
+    }
+
+    same_socket_options = existing_listener.compareSocketOptions(new_listener);
+    if (!same_socket_options && new_listener.reusePort() == false) {
+      incListenerCreateFailureStat();
+      throw EnvoyException(fmt::format("Listener {}: doesn't support update any socket options "
```
mattklein123, line 440:
```suggestion
      throw EnvoyException(fmt::format("Listener {}: socket options cannot be changed during an update "
```

test/integration/listener_lds_integration_test.cc
mattklein123: Perhaps a better way would be to use https://github.com/envoyproxy/envoy/blob/main/source/server/listener_hooks.h and add a new hook for a failure reason that you can trap in tests? cc @alyssawilk to see if she has any other thoughts about the test part of this change.

mattklein123 (changes requested):
Thanks this is great. Flushing out some more comments. For all the new tests, please try to run them many times to look for flakes. https://github.com/envoyproxy/envoy/tree/main/test/integration#reproducing-test-flakes

/wait

test/integration/listener_lds_integration_test.cc
alyssawilk: This results in the config being rejected yes?  If so that should always come with a warning / error log explaining why the config is invalid, and you can use EXPECT_LOG_CONTAINS to make sure you got the cause you expected.

source/server/listener_impl.cc
```diff
@@ -1123,6 +1145,39 @@ void ListenerImpl::closeAllSockets() {
   }
 }
 
+bool ListenerMessageUtil::compareSocketOptions(const envoy::config::listener::v3::Listener& lhs,
```
ggreenway, line 1148: Rename to `socketOptionsEqual` for clarity. The current name doesn't make it clear what true or false means.

source/server/listener_impl.cc
ggreenway: +1. I'd rather keep it simpler.

ggreenway (changes requested):
This approach LGTM. 

Will this also work for hot-restart? I think socket options changes aren't applied for hot-restart either.

soulxu:
> This approach LGTM.
> 
> Will this also work for hot-restart? I think socket options changes aren't applied for hot-restart either.

Yes, we doesn't change socket options when hot-restart, the existing socket option already applied on the socket which passed from the parent, the OS suppose ensure there is no change.

Also checked, although we have the same address between new listener and drainning listener, the hot restart always get the socket from the active listener
https://github.com/envoyproxy/envoy/blob/ad186c448cf8c5e78454d2adf5f228e9c90a997c/source/server/hot_restarting_parent.cc#L110

`ListenerManager::listeners()` return active listener default. So hotrestart won't get old socket with old options from the parent.

source/server/listener_impl.cc
soulxu: will do

test/integration/listener_lds_integration_test.cc
soulxu: I can use `EXPECT_LOG_CONTAINS` to get expected message, like
```
EXPECT_LOG_CONTAINS(
      "warning",
      "gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error "
      "adding/updating listener(s) listener_foo: Listener listener_foo: doesn't support update any "
      "socket options when the reuse port isn't enabled",
      test_server_->waitForCounterGe("listener_manager.lds.update_rejected", 1))
```
but there isn't a point I can ensure the log already emit

I wait on the stat `listener_manager.lds.update_rejected`, but I can't ensure the log emitted. Even I change the code to emit the log first, I still see the test flaky.
https://github.com/envoyproxy/envoy/blob/ad186c448cf8c5e78454d2adf5f228e9c90a997c/source/common/config/grpc_subscription_impl.cc#L127-L128

Maybe I still need to use listener hook.

soulxu: Emm.. I can put something more into the EXPECT_LOG_CONTAINS

```
EXPECT_LOG_CONTAINS(
      "warning",
      "gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error "
      "adding/updating listener(s) listener_foo: Listener listener_foo: doesn't support update any "
      "socket options when the reuse port isn't enabled",
      {
        sendLdsResponse({MessageUtil::getYamlStringFromMessage(listener_config_)}, "2");
        test_server_->waitForCounterGe("listener_manager.lds.update_rejected", 1);
        IntegrationTcpClientPtr tcp_client2 = makeTcpConnection(lookupPort("test_listener"));
        ASSERT_TRUE(tcp_client2->write(data));
        FakeRawConnectionPtr fake_upstream_connection2;
        ASSERT_TRUE(fake_upstreams_[0]->waitForRawConnection(fake_upstream_connection2));
        ASSERT_TRUE(fake_upstream_connection2->waitForData(data.size(), &data));
        tcp_client2->close();
      }
  );
```

Then I got enough time to wait the log emitted, I think that is enough

soulxu:
> Thanks this is great. Flushing out some more comments. For all the new tests, please try to run them many times to look for flakes. https://github.com/envoyproxy/envoy/tree/main/test/integration#reproducing-test-flakes
> 
> /wait

good call, thanks! I saw a segmentation fault in the end of `UpdateListenerWithDifferentSocketOptionsWhenReusePortDisabled`, let me dig into it.

ggreenway:
> Yes, we doesn't change socket options when hot-restart, the existing socket option already applied on the socket which passed from the parent, the OS suppose ensure there is no change.

I meant will it work if the socket options have changed in the listener configuration between the parent and the child process in hot-restart? It should probably have the same semantics as an LDS update; logically it's the same change, even though the mechanism is a bit different.

soulxu:
> > Yes, we doesn't change socket options when hot-restart, the existing socket option already applied on the socket which passed from the parent, the OS suppose ensure there is no change.
> 
> I meant will it work if the socket options have changed in the listener configuration between the parent and the child process in hot-restart? It should probably have the same semantics as an LDS update; logically it's the same change, even though the mechanism is a bit different.

I see now, thanks for explaining to me again! I assume people don't change the config when hot-restart at the beginning since some experience told me it will be easy to have separate steps to do the upgrade, then things become easy. But after second thought, seems like everything should be fine except the config related to socket for envoy hot restart

For this case, I don't think it works perfectly, I can imagine two cases don't work well.
* if the user removed a socket option after hot-restart, that setup won't be clear after duplicated from the parent. And there is no way to clear the setup since we don't know the platform default value for each socket option.
* prebind socket option can't ensure works as expected, since the code will apply the prebind socket option again, but the socket is already bound since it is duplicated from the parent.
   https://github.com/envoyproxy/envoy/blob/a3b59dd87abfe6cc65b3f0af31dda1c42b725796/source/common/network/listen_socket_impl.h#L83-L87
   `setListenSocketOptions` will set the prebind socket options

The simple solution here is to suggest the user don't change the socket option when hot-restart, the user can do another update for socket option after hot-restart. Also the user only can append new socket option for `STATE_LISTENING` when hot-restart. We can document those somewhere. Does make sense?

ggreenway:
> The simple solution here is to suggest the user don't change the socket option when hot-restart, the user can do another update for socket option after hot-restart. Also the user only can append new socket option for `STATE_LISTENING` when hot-restart. We can document those somewhere. Does make sense?

Another option would be, in the hot-restart protocol, for the parent to send the current socket options, and if they're different from the child socket options, invoke the same code path you're writing in this PR to update them. But right now, the child doesn't know the current socket options, so this can't be done without adding that additional information to hot-restart.

soulxu:
> > The simple solution here is to suggest the user don't change the socket option when hot-restart, the user can do another update for socket option after hot-restart. Also the user only can append new socket option for `STATE_LISTENING` when hot-restart. We can document those somewhere. Does make sense?
> 
> Another option would be, in the hot-restart protocol, for the parent to send the current socket options, and if they're different from the child socket options, invoke the same code path you're writing in this PR to update them. But right now, the child doesn't know the current socket options, so this can't be done without adding that additional information to hot-restart.

Yea, we probably need a separate hot-restart API to get the parent's socket options config. @mattklein123 WDYT? do you have preference on which solution?

mattklein123:
Personally I would probably try to keep it simple until someone actively wants this to work for hot restart. If we do want to implement it, passing the old socket options seems like a fine approach but it would be a bunch of work.

soulxu:
> Personally I would probably try to keep it simple until someone actively wants this to work for hot restart. If we do want to implement it, passing the old socket options seems like a fine approach but it would be a bunch of work.

I did a check the hot restart api, we probably need to add a new parameter to this hot restart API
https://github.com/envoyproxy/envoy/blob/f3c6c6d250b2387121e113f3ae5c47b81312e05a/source/server/hot_restarting_child.h#L17

The new parameter is used to pass the new socket_options config to the parent. Then the parent will find a socket with the same address and same socket_options, otherwise consider no socket can pass to the child. Then the child found there is invalid fd returned from the parent, then it will create new socket
https://github.com/envoyproxy/envoy/blob/f3c6c6d250b2387121e113f3ae5c47b81312e05a/source/server/listener_manager_impl.cc#L248-L249

new parameter should be a compatible API change for the hot restart API, the old envoy just ignore that parameter.

mattklein123:
Yeah I agree ^ will work fine. I just wouldn't implement it until someone actually asks for the support.

soulxu:
> Yeah I agree ^ will work fine. I just wouldn't implement it until someone actually asks for the support.

got it, thanks for the feedback! let me add some doc for the hotrestart.

soulxu:
> Thanks LGTM with small comment. I would suggest we hold this until after the release. @ggreenway is this OK with you regarding hot restart? Any other comments?
> 
> /wait

agree with that, it isn't worth rush before release

docs/root/intro/arch_overview/operations/hot_restart.rst
```diff
@@ -35,6 +35,11 @@ will exit as described above. The :option:`--base-id` or :option:`--use-dynamic-
 may be used to allow multiple, distinctly configured Envoys to run on the same host and hot restart
 independently.
 
+.. note::
+
+ The update of the listener's :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` with a hot restart doesn't support well.
+ Suggest doing the update listener's :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` separate with the hot restart.
```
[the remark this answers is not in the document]
soulxu: always thanks for the help on the writing!

soulxu:
/wait

waiting for 1.24 release, then this PR is ready for merge

ggreenway:
Yeah, hot-restart should be a separate PR if/when someone wants it. I agree with merging this post-release.

changelogs/current.yaml
```diff
@@ -2,6 +2,18 @@ date: Pending
...
 behavior_changes:
 # *Changes that are expected to cause an incompatibility if applicable; deployment changes are likely required*
+- area: listener
+  change: |
+    Previously a listener update with different :ref:`transparent <envoy_v3_api_field_config.listener.v3.Listener.transparent>`, :ref:`freebind <envoy_v3_api_field_config.listener.v3.Listener.freebind>`,
+    :ref:`tcp_fast_open_queue_length <envoy_v3_api_field_config.listener.v3.Listener.tcp_fast_open_queue_length>` or :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` was ignored.
+    Now, when those fields are updated, a new socket will be created for
+    the listener and the new value of those fields applied to it. This only happens when :ref:`enable_reuse_port <envoy_v3_api_field_config.listener.v3.Listener.enable_reuse_port>` is true.
+    Otherwise if those fields change the update is rejected.
+    New runtime flag ``envoy.reloadable_features.enable_update_listener_socket_options`` is added for revert this behavior.
```
mattklein123, line 12:
```suggestion
    New runtime flag ``envoy.reloadable_features.enable_update_listener_socket_options`` can be used to revert this behavior.
```

docs/root/intro/arch_overview/operations/hot_restart.rst
```diff
@@ -35,6 +35,11 @@ will exit as described above. The :option:`--base-id` or :option:`--use-dynamic-
 may be used to allow multiple, distinctly configured Envoys to run on the same host and hot restart
 independently.
 
+.. note::
+
+ Currently, updating a listener's :ref:`socket_options <envoy_v3_api_field_config.listener.v3.Listener.socket_options>` during a hot restart isn't supported.
+ The old process socket options will be used. If it is required to update socket options either do a full restart or perform an LSD based listener updated.
```
mattklein123, line 41:
```suggestion
 The old process socket options will be used. If it is required to update socket options either do a full restart or perform an LDS based listener updated.
```

mattklein123 (changes requested):
LGTM with small remaining doc comments, thanks.

/wait

soulxu:
@mattklein123 thanks!

mattklein123 (approved):

mattklein123:
@ggreenway if no further comments can you approve/merge?

KBaichoo:
@soulxu seems like this needs merge to resolve conflict

@ggreenway hope you are feeling better, can you take a look when you have a chance? Thank you!

/wait

ggreenway:
I'm hoping to review next week (hopefully early in the week)

soulxu:
> I'm hoping to review next week (hopefully early in the week)

thanks! no hurry.

ggreenway (dismissed):
LGTM. Needs merge conflicts resolved.

soulxu:
> LGTM. Needs merge conflicts resolved.

done, thanks!

kfaseela:
Thanks a lot @soulxu @ggreenway @mattklein123  :)

ggreenway (approved):

PR #23063, opened 2022-09-11
WIP: Local cert provider

liverbirdkte:
Local certificate provider instance

A certificate provider instance to generate certificates locally.  It's based on the certificate provider framework and bumping filter could use this instance for testing.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #22036, opened 2022-07-06
SNI-based cert selection during TLS handshake

LuyaoZhong:
Envoy supports selecting certs by selecting filter chain based on SNI.
But it is possible that we access different services via one filter
chain, which requires SNI-based cert selection in one single filter
chain during handshake.

Risk Level: Medium
Testing: unit tests
Docs Changes: ssl
Release Notes: yes
Fixes #21739

Signed-off-by: Luyao Zhong <luyao.zhong@intel.com>

PR #19582, opened 2022-01-18
Implement Certificate Provider Framework (#19308)

LuyaoZhong:
Implement Certificate Provider Framework

Constructing and managing the certificate provider instances in Bootstrap
Implement the ca_certificate_provider_instance in CertificateValidationContext
Implement mplement tls_certificate_provider_instance in CommonTlsContext

Risk Level: Low
Testing: Unit tests
Docs Changes: N/A
Release Notes: N/A
Fixes #19308

Signed-off-by: Luyao Zhong <luyao.zhong@intel.com>

PR #23192, opened 2022-09-21
[PoC][Draft] TLS Bumping integration 

LuyaoZhong:
**NOT READY FOR REVIEW**

This PR is to merge following parts, and do integration test. 

Certificate Provider framework https://github.com/envoyproxy/envoy/pull/19582
SNI-based cert selection in tls transport socket https://github.com/envoyproxy/envoy/pull/22036
A new network filter - BumpingFilter https://github.com/envoyproxy/envoy/pull/22582
Certificate Provider instance - LocalMimicCertProvider https://github.com/envoyproxy/envoy/pull/23063

Signed-off-by: LeiZhang <lei.a.zhang@intel.com>


LuyaoZhong:
@mattklein123 Could you help reopen this draft PR?

ohadvano:
@LuyaoZhong in your design and implementation, is it expected that the TCP proxy/HCM would re-use the connection that was opened for fetching the certificate? Or would Envoy end up with 2 upstream connections per 1 downstream request that includes mimicking?

LuyaoZhong:
> n your design and implementation, is it expected that the TCP proxy/HCM would re-use the connection that was opened for fetching the certificate? Or would Envoy end up with 2 upstream connections per 1 downstream request that includes mimicking?

We  will not re-use the connection. So envoy would have 2 upstream connections per 1 downstream request if mimicking is required. And when corresponding mimic cert is ready, we don't need addition upstream connection, then it will be 1 upstream connection per 1 downstream request.

Would you mind share your use case?

ohadvano:
My use case is somewhat similar, but instead of having Envoy do the mimicking, I would like to query an external service that would do the mimicking and respond with a certificate. The problem I had in mind was that if an external service does this, then there would be no re-use of connections, and we'll end up with 2 TCP streams per 1 downstream request. However, seems like there's no better option at this time and supporting this on Envoy would be very expensive. 

In my case I'm investigating the option to integrate this as part of the SDS. Meaning that if a secret is unknown at TLS handshake time, Envoy will pause the handshake and call SDS to fetch the certificate on-demand. Then, since SDS is external service, it can do whatever it wants at the time the request for fetch was invoked. It can either respond with yet-provided secret (that is known to SDS), or it can have logic to do mimicking on the fly.
Since you were doing some work on this, I was hoping to get your ideas on this method and has this been looked into.

Current problem with SDS is that it does not support "Give me all you have" or "Fetch on-demand" features. You have to subscribe to a named resource statically and SDS will only provide secrets for these subscriptions.
Further context: https://github.com/envoyproxy/envoy/issues/25958

LuyaoZhong:
wip

LuyaoZhong:
wip

MarkRx:
What ever happened to this?

PR #23208, opened 2022-09-22
refactor: factory cleanup

alyssawilk:
As follow-on clean up for upstream filters, cleaning up more factory code.

Risk Level: low
Testing: existing tests
Docs Changes: n/a
Release Notes: n/a


adisuissa (approved):
Thanks for this cleanup!
One small question, could the context change mid-run, or is it set once? (trying to understand whether there's implication on when these context members are latched).
LGTM otherwise.

snowp (approved):
Thanks!

PR #23225, opened 2022-09-23
fuzzy: avoid reconnected in health check fuzzy test

botengyao:

We should not reconnect the session for TCP onEvent. Only TCP checker processes the `CONNECTED` onEvent, there is no impact on HTTP and GRPC

Commit Message:
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


yanavlasov:
Let's discuss this internally. There are few more crashes in this fuzzer for the same reason - invalid event sequence. I think the event generation in the fuzzer is rather broken and this fixes only one small part. This whole approach needs to be rethought.

botengyao:
> Let's discuss this internally. There are few more crashes in this fuzzer for the same reason - invalid event sequence. I think the event generation in the fuzzer is rather broken and this fixes only one small part. This whole approach needs to be rethought.

Got it, thanks!

PR #23233, opened 2022-09-23
tls: reduce flakiness of ssl_integration_test

benjaminp:
A few async certificate validation tests want to make assertions while async validation is pending. If the vagaries of scheduling caused async validation to finish before the test code finished running, the test would fail.

To reduce the chance of this flakiness, increase the time the fake "timed" certificate validator waits before validating the certificate. This mitigation is inspired by [similar code in the QUIC integration tests](https://github.com/envoyproxy/envoy/blob/deded6dcf86b2762ab885af88e0e3ae3f27eee7b/test/integration/quic_http_integration_test.cc#L1084-L1091).

Also, modify some loops in tests so that they do not loop forever if the handshake state machine advances past the state they are waiting for.

Signed-off-by: Benjamin Peterson <benjamin@engflow.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
+++ b/test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
@@ -399,7 +399,7 @@ typed_config:
   const auto* socket = dynamic_cast<const Extensions::TransportSockets::Tls::SslHandshakerImpl*>(
       connection->ssl().get());
   ASSERT(socket);
-  while (socket->state() != Ssl::SocketState::HandshakeInProgress) {
+  while (socket->state() == Ssl::SocketState::PreHandshake) {
     dispatcher_->run(Event::Dispatcher::RunType::NonBlock);
   }
   ASSERT_EQ(connection->state(), Network::Connection::State::Open);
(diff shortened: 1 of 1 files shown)


test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
```diff
@@ -399,7 +399,7 @@ name: "envoy.tls.cert_validator.timed_cert_validator"
   const auto* socket = dynamic_cast<const Extensions::TransportSockets::Tls::SslHandshakerImpl*>(
       connection->ssl().get());
   ASSERT(socket);
-  while (socket->state() != Ssl::SocketState::HandshakeInProgress) {
+  while (socket->state() < Ssl::SocketState::HandshakeInProgress) {
```
giantcroc, line 402: Hi, I think it's confusing to use `<` for enum type, maybe we can replace it with `== Ssl::SocketState::PreHandshake`.

giantcroc:
```suggestion
  while (socket->state() == Ssl::SocketState::PreHandshake) {
```

yanavlasov (approved):

Issue #19527, opened 2022-01-13
postgres: Support StartTLS upstream and plain-text downstream


terorie:
*Title*: *Allow Postgres proxy to upgrade unencrypted downstream connection into encrypted upstream connection*

*Description*:

The PostgreSQL filter supports TLS termination: https://github.com/envoyproxy/envoy/pull/14634 In this mode, it accepts an encrypted client connection (downstream), strips TLS, and creates an outgoing unencrypted connection to the SQL DB (upstream).

Conversely, it should support the opposite where Envoy accepts unencrypted (`sslmode=disable`) Postgres connections, adds StartTLS on top, and dials securely to the upstream.

This is particularly useful for Google Cloud SQL which enforces the use of client certificates.
Unfortunately a few Postgres clients (such as Rust's `tokio-postgres`) don't support client certs.
So a service proxy is required to establish a secure connection.
(Google Cloud SQL specifically provides the `cloud_sql_proxy` tool for this reason but obviously Envoy is a preferred choice given its maturity)

Defining a `starttls` transport socket on the `cluster` config in Envoy v1.21 results in surprising behavior.
When making an insecure connection to Envoy, the Postgres filter will just pass through the connection attempt to the upstream without attempting to upgrade the socket into TLS mode.

*Proposed fix*:

* Introduce a new `upstream_sslmode` string that mirrors the available PostgreSQL `sslmode` settings.
* When `upstream_sslmode` is defined, send that to the upstream instead of the `sslmode` specified by the client.
* When `upstream_sslmode` is `prefer`, `require`, `verify-ca`, or `verify-full` and the downstream connection is unencrypted, initiate an upstream StartTLS upgrade.

*Useful links*:

- PostgreSQL SSL: https://www.postgresql.org/docs/9.1/libpq-ssl.html
- Google Cloud SQL SSL settings: https://cloud.google.com/sql/docs/postgres/configure-ssl-instance#client-certs
- Google `cloud_sql_proxy`: https://github.com/GoogleCloudPlatform/cloudsql-proxy

alyssawilk:
cc @fabriziomello @cpakulski @dio

cpakulski:
Seems that there is demand for this feature. I will try to implement this.

fabriziomello:
@cpakulski doing that we'll loose the ability to decode the protocol and expose metrics. We introduced starttls into Postgres filter to terminate the SSL and able us to decode the protocol even for encrypted client connections.

[...]

PR #23249, opened 2022-09-26
filter manager: ability to convert upstream Starttls to secure mode

cpakulski:
Commit Message:
filter manager: ability to convert upstream Starttls to secure mode

Additional Description:
This PR provides a signalling path from a downstream read filter to a terminal filter in order to convert StartTLS upstream transport socket from non-secure mode to secure mode. StartTLS transport socket is basically a wrapper around raw (cleartext socket) and TLS socket. StartTLS transport socket always starts in cleartext and in order to convert it to secure mode, `startSecureTransport` method must be called. The main use of StartTLS transport socket is implementation of Opportunistic TLS protocol, where the first few messages are exchanged in cleartext and then TLS handshake is done on the same TCP session. The first few cleartext messages are protocol specific. Postgres messages are different than for example SMTP messages. Therefore a Read downstream filter is needed to parse those protocol specific messages and recognize the need to start TLS. The problem is that the Read filter (postgres, smtp, mysql) does not have an access to upstream connection, where the StartTLS transport socket is attached. The only filter which has a knowledge about both downstream and upstream connections it the terminal filter. Therefore the signal must be somehow conveyed from one of the downstream Read filters to the terminal filter. The result of the operation (whether upstream startTLS has been successfully converted to secure mode) should be returned to the Read filter.
That problem was discussed in one of the community meetings. @mattklein123 suggested to extend network FilterManager's callbacks and this PR uses that idea. The network filter manager iterates over all downstream filters and calls `startUpstreamSecureTransport` until one of them returns `true`. The only terminal filter which implements `startUpstreamSecureTransport` method is tcp_proxy. All other filters use `startUpstreamSecureTransport` defined in base class which is empty/no-op. @ggreenway raised a concern that tcp_proxy may be used for tunnelling. To address that issue only non-http upstream has `startUpstreamSecureTransport` implemented. The idea is described below.

FM

I used a modified postgres filter to exercise the new signalling path and it works properly.

Risk Level: Low. 
Testing: Added unit and integration testing.
Docs Changes: No.
Release Notes: No
Platform Specific Features: No.
Part of #19527

### DIFF
--- a/source/common/network/filter_manager_impl.cc
+++ b/source/common/network/filter_manager_impl.cc
@@ -88,6 +88,19 @@ void FilterManagerImpl::onRead() {
   onContinueReading(nullptr, connection_);
 }
 
+bool FilterManagerImpl::startUpstreamSecureTransport() {
+  for (auto& filter : upstream_filters_) {
+    if (!(filter)->filter_) {
+      continue;
+    }
+    if (filter->filter_->startUpstreamSecureTransport()) {
+      // Success. The filter converted upstream's transport socket to secure mode.
+      return true;
+    }
+  }
+  return false;
+}
+
 FilterStatus FilterManagerImpl::onWrite() { return onWrite(nullptr, connection_); }
 
 FilterStatus FilterManagerImpl::onWrite(ActiveWriteFilter* filter,
--- a/source/common/network/filter_manager_impl.cc
+++ b/source/common/network/filter_manager_impl.cc
@@ -90,10 +90,7 @@ void FilterManagerImpl::onRead() {
 
 bool FilterManagerImpl::startUpstreamSecureTransport() {
   for (auto& filter : upstream_filters_) {
-    if (!(filter)->filter_) {
-      continue;
-    }
-    if (filter->filter_->startUpstreamSecureTransport()) {
+    if (filter->filter_ != nullptr && filter->filter_->startUpstreamSecureTransport()) {
       // Success. The filter converted upstream's transport socket to secure mode.
       return true;
     }
--- a/source/common/tcp_proxy/tcp_proxy.cc
+++ b/source/common/tcp_proxy/tcp_proxy.cc
@@ -599,6 +599,8 @@ Network::FilterStatus Filter::onNewConnection() {
   return establishUpstreamConnection();
 }
 
+bool Filter::startUpstreamSecureTransport() { return upstream_->startUpstreamSecureTransport(); }
+
 void Filter::onDownstreamEvent(Network::ConnectionEvent event) {
   if (event == Network::ConnectionEvent::LocalClose ||
       event == Network::ConnectionEvent::RemoteClose) {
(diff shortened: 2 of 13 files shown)


ggreenway:
@lizan please review, or unassign yourself and I'll find another assignee.

source/common/network/filter_manager_impl.cc
```diff
@@ -88,6 +88,19 @@ void FilterManagerImpl::onRead() {
   onContinueReading(nullptr, connection_);
 }
 
+bool FilterManagerImpl::startUpstreamSecureTransport() {
+  for (auto& filter : upstream_filters_) {
+    if (!(filter)->filter_) {
```
lizan, line 93: nit: compare with nullptr and this if can be combined with the one below:

`if (filter->filter_ != nullptr && filter->filter_->startUpstreamSecureTransport())`

source/server/api_listener_impl.h
```diff
@@ -67,6 +67,10 @@ class ApiListenerImplBase : public ApiListener,
     void injectReadDataToFilterChain(Buffer::Instance&, bool) override {
       IS_ENVOY_BUG("Unexpected call to injectReadDataToFilterChain");
     }
+    bool startUpstreamSecureTransport() override {
+      IS_ENVOY_BUG("Unexpected call to injectUpstreamConnectionEvent");
```
lizan, line 71:
```suggestion
      IS_ENVOY_BUG("Unexpected call to startUpstreamSecureTransport");
```

lizan (commented):
this is in the right shape, a couple nits

lizan:
/assign-from @envoyproxy/envoy-maintainers

for non-Tetrands review

phlax:
im probably not the best maintainer to be assigned 8/ - @ggreenway is it ok to assign you ?

source/common/network/filter_manager_impl.cc
cpakulski: Done. Thanks.

RyanTheOptimist:
@ggreenway would you mind giving this a look now that @lizan is happy with it?

test/common/tcp_proxy/tcp_proxy_test.cc
```diff
@@ -1328,6 +1328,17 @@ TEST_F(TcpProxyTest, UpstreamConnectFailureStreamInfoAccessLog) {
   EXPECT_EQ(access_log_data_, "test_transport_failure");
 }
 
+// Test that call to tcp_proxy filter's startUpstreamSecureTransport results
+// in upstream's startUpstreamSecureTransport call.
+TEST_F(TcpProxyTest, UpstreamStartSecuteTransport) {
```
ggreenway, line 1333:
```suggestion
TEST_F(TcpProxyTest, UpstreamStartSecureTransport) {
```

ggreenway (changes requested):
LGTM modulo one typo

/wait

ggreenway (approved):

source/common/tcp_proxy/tcp_proxy.cc
```diff
@@ -599,6 +599,8 @@ Network::FilterStatus Filter::onNewConnection() {
   return establishUpstreamConnection();
 }
 
+bool Filter::startUpstreamSecureTransport() { return upstream_->startUpstreamSecureTransport(); }
```
soulxu, line 602: Emm... seems like there should be an assertion to ensure upstream_ isn't nullptr.

PR #23253, opened 2022-09-26
Minor cleanup for TLS inspector filter.

KBaichoo:

Commit Message:Cleanup for TLS inspector. 
Additional Description:
1) Avoid making share ptr copy
2) Use std algo vs loop
3) Make member const since it's only read after init

Risk Level:low
Testing: ran test
Docs Changes: n/a
Release Notes: n/a
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/extensions/filters/listener/tls_inspector/tls_inspector_integration_test.cc
+++ b/test/extensions/filters/listener/tls_inspector/tls_inspector_integration_test.cc
@@ -166,7 +166,7 @@ TEST_P(TlsInspectorIntegrationTest, ContinueOnListenerTimeout) {
   // will continue wait. Then the listener filter timeout timer will be triggered.
   Buffer::OwnedImpl buffer("fake data");
   client_->write(buffer, false);
-  // the timeout is set as one seconds, sleep 5 to trigger the timeout.
+  // the timeout is set as one seconds, advance 2 seconds to trigger the timeout.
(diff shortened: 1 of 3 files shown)


test/extensions/filters/listener/tls_inspector/tls_inspector_integration_test.cc
```diff
@@ -166,7 +166,7 @@ TEST_P(TlsInspectorIntegrationTest, ContinueOnListenerTimeout) {
   // will continue wait. Then the listener filter timeout timer will be triggered.
   Buffer::OwnedImpl buffer("fake data");
   client_->write(buffer, false);
-  // the timeout is set as one seconds, sleep 5 to trigger the timeout.
+  // the timeout is set as one seconds, advance 2 seconds to trigger the timeout.
```
ggreenway, line 169: Optional: capitalize first letter of the comment (even though it was already like that)

ggreenway (approved):
LGTM

test/extensions/filters/listener/tls_inspector/tls_inspector_integration_test.cc
KBaichoo: done, thanks for speedy review

ggreenway (approved):

PR #23255, opened 2022-09-26
upstream: Add support for sharing alt-svc entries between hosts.

RyanTheOptimist:
Add a "canonical suffix" list to the Alt-Svc cache so that Alt-Svc entries can be shared across origins which share the same hostname suffix.

Risk Level: Low
Testing: New unit tests
Docs Changes: Update proto docs
Release Notes: Updated

### DIFF
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -171,6 +171,9 @@ message AlternateProtocolsCacheOptions {
 
   // Allows pre-populating the cache with entries, as described above.
   repeated AlternateProtocolsCacheEntry prepopulated_entries = 4;
+
+  // Optional list of hostnames suffixes for which Alt-Svc entries can be shared.
+  repeated string canonical_suffixes = 5;
 }
 
 // [#next-free-field: 7]
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -127,6 +127,7 @@ message UpstreamHttpProtocolOptions {
 // make an HTTP connection to an origin server. See https://tools.ietf.org/html/rfc7838 for
 // HTTP Alternative Services and https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-04
 // for the "HTTPS" DNS resource record.
+// [#next-free-field: 6]
 message AlternateProtocolsCacheOptions {
   // Allows pre-populating the cache with HTTP/3 alternate protocols entries with a 7 day lifetime.
   // This will cause Envoy to attempt HTTP/3 to those upstreams, even if the upstreams have not
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -174,10 +174,14 @@ message AlternateProtocolsCacheOptions {
   repeated AlternateProtocolsCacheEntry prepopulated_entries = 4;
 
   // Optional list of hostnames suffixes for which Alt-Svc entries can be shared. For example, if
-  // this list contained the value `.c.example.com`, then an Alt-Svc entry for `foo.c.example.com`
-  // could be shared with 'bar.c.example.com' but would not be shared with 'baz.example.com'. On
-  // the other hand, if the list contained the value '.youtube.com' then all three hosts could share
-  // Alt-Svc entries.
+  // this list contained the value ``.c.example.com``, then an Alt-Svc entry for ``foo.c.example.com``
+  // could be shared with ``bar.c.example.com`` but would not be shared with ``baz.example.com``. On
+  // the other hand, if the list contained the value ``.example.com`` then all three hosts could share
+  // Alt-Svc entries. Each entry must start with ``.``.  If a hostname matches multiple suffixes, the
+  // first listed suffix will be used.
+  //
+  // Since lookup in this list is O(n), it is recommended that the number of suffixes be limited.
+  // [#not-implemented-hide:]
   repeated string canonical_suffixes = 5;
 }
 
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -173,7 +173,11 @@ message AlternateProtocolsCacheOptions {
   // Allows pre-populating the cache with entries, as described above.
   repeated AlternateProtocolsCacheEntry prepopulated_entries = 4;
 
-  // Optional list of hostnames suffixes for which Alt-Svc entries can be shared.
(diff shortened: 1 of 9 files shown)


api/envoy/config/core/v3/protocol.proto
```diff
@@ -171,6 +172,9 @@ message AlternateProtocolsCacheOptions {
 
   // Allows pre-populating the cache with entries, as described above.
   repeated AlternateProtocolsCacheEntry prepopulated_entries = 4;
+
+  // Optional list of hostnames suffixes for which Alt-Svc entries can be shared.
```
alyssawilk, line 176: WDYT of adding an example.com based example, along with matches?
also I'm not clear if foo.c.youtube.com would match both .c.youtube.com and .youtube.com or only the former so let's make that clear here.

source/common/http/http_server_properties_cache_manager_impl.cc
```diff
@@ -51,9 +51,12 @@ HttpServerPropertiesCacheSharedPtr HttpServerPropertiesCacheManagerImpl::getCach
         factory.createStore(kv_config, data_.validation_visitor_, dispatcher, data_.file_system_);
   }
 
-  HttpServerPropertiesCacheSharedPtr new_cache = std::make_shared<HttpServerPropertiesCacheImpl>(
-      dispatcher, std::move(store), options.max_entries().value());
+  auto new_cache = std::make_shared<HttpServerPropertiesCacheImpl>(dispatcher, std::move(store),
+                                                                   options.max_entries().value());
 
+  for (const std::string& suffix : options.canonical_suffixes()) {
+    new_cache->addCanonicalSuffix(suffix);
```
alyssawilk, line 58: should we have input validation this starts with . and is otherwise a valid hostname?

alyssawilk (commented):
oh yeah!  I'm stoked to have this one crossed off our list :-)
/wait

source/common/http/http_server_properties_cache_manager_impl.cc
RyanTheOptimist: Added the "starts with '.'" check though I'm not sure if we really want to do a valid hostname check. I could imagine some internal usage where folks use "intranet hostnames". But if you feel strongly, happy to reconsider.

api/envoy/config/core/v3/protocol.proto
RyanTheOptimist: Good idea. PTAL!

RyanTheOptimist (commented):
Thanks!

api/envoy/config/core/v3/protocol.proto
```diff
@@ -171,6 +172,13 @@ message AlternateProtocolsCacheOptions {
 
   // Allows pre-populating the cache with entries, as described above.
   repeated AlternateProtocolsCacheEntry prepopulated_entries = 4;
+
+  // Optional list of hostnames suffixes for which Alt-Svc entries can be shared. For example, if
+  // this list contained the value `.c.example.com`, then an Alt-Svc entry for `foo.c.example.com`
```
htuch, line 177: @phlax this is a place we could do with better guidance on canonical formatting as style guide.

api/envoy/config/core/v3/protocol.proto
```diff
@@ -171,6 +172,13 @@ message AlternateProtocolsCacheOptions {
...
+  // could be shared with 'bar.c.example.com' but would not be shared with 'baz.example.com'. On
+  // the other hand, if the list contained the value '.youtube.com' then all three hosts could share
```
htuch, line 179: API looks good, but just a nit on whether we expect to have a leading dot or not for the suffix. Is this a true string suffix or is it a domain zone essentially?

api/envoy/config/core/v3/protocol.proto
phlax, line 177:
```suggestion
  // this list contained the value ``.c.example.com``, then an Alt-Svc entry for ``foo.c.example.com``
```

phlax: yep - not sure - i would say these are literals - defo not italics tho - below uses single quotes - i would say they are all literal

test/common/http/http_server_properties_cache_impl_test.cc
```diff
@@ -372,6 +372,23 @@ TEST_F(HttpServerPropertiesCacheImplTest, GetOrCreateHttp3StatusTracker) {
   EXPECT_FALSE(protocols_->getOrCreateHttp3StatusTracker(origin1_).isHttp3Broken());
 }
 
+TEST_F(HttpServerPropertiesCacheImplTest, CanonicalSuffix) {
+  std::string suffix = ".example.com";
+  std::string host1 = "first.example.com";
+  std::string host2 = "second.example.com";
```
alyssawilk, line 378: extra.sections.example.com ?

source/common/http/http_server_properties_cache_impl.cc
```diff
@@ -286,5 +293,41 @@ HttpServerPropertiesCacheImpl::getOrCreateHttp3StatusTracker(const Origin& origi
...
+  return it->second;
+}
+
+void HttpServerPropertiesCacheImpl::maybeSetCanonicalOrigin(const Origin& origin) {
+  absl::string_view suffix = getCanonicalSuffix(origin.hostname_);
+  if (suffix.empty()) {
+    return;
+  }
+
+  canonical_alt_svc_map_[std::string(suffix)] = origin;
```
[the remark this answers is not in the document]
RyanTheOptimist: Oh good point. I wasn't really think about the KV store other than to note that we don't need to persist the suffixes. But yeah, on load it really should just do the right thing.

Of course, because I wasn't thinking about it, I didn't write a test for it. And because there was no test for the right behavior, it had the wrong behavior :) This required passing in the suffixes to the constructor which I've done, and I've added a test. Thanks!

RyanTheOptimist (commented):
Thanks for the great comments, all!

changelogs/current.yaml
```diff
@@ -196,6 +196,9 @@ new_features:
 - area: access_log
   change: |
     added support for number values in substitution format string in json_format.
+- area: upstream
+  change: |
+    added support for sharing alt-svc entries between hosts which share a suffix.
```
alyssawilk, line 201: consider removing as we're hiding the API

alyssawilk (commented):
Cool, LGTM modulo hiding the API.
@danzh2010 or @RenjieTang would one of you take a look as well?

alyssawilk (approved):

RenjieTang (approved):

htuch (commented):
/lgtm api

PR #23261, opened 2022-09-27
generic proxy: provides per filter config API to L7 filters to get route level config by the filter config name

wbpcode:
Commit Message: generic proxy: make the l7 filter could be configured repeatedly
Additional Description:

Add a new `perFilterConfig` API to the `StreamFilterCallbacks` of generic proxy. This API will get route level filter config from the route entry by the filter config name.

It's similar to #21525.

**I think there is no third-party L7 filter for generic proxy for now. So, this change won't break anything.**

Risk Level: Low.
Testing: Unit.
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
Hi, @mattklein123 because this is similar to #21525, could you take a look when you have free time? Thanks.

mattklein123 (approved):

PR #23267, opened 2022-09-28
Disable RFC9113 header value whitespace validation

### LINKED ISSUE #23266: Envoy now rejecting header values with trailing whitespace
*After upgrading nghttp2 to 1.49.0 (#23014), Envoy is now rejecting header values with trailing whitespace*

*Description*:

nghttp2 v1.49.0 rejects trailing whitespaces in header values to follow validation introduced in [RFC9113](https://datatracker.ietf.org/doc/html/rfc9113#section-8.2.1). This is newly introduced behavior which can break existing deployments of Envoy using HTTP2.

At Lyft, the impact of this change has been significantly increased HTTP2 connection churn (`upstream_cx_destroy_remote`) leading to widespread UC and 503's between our Envoy front proxy and Envoy service mesh sidecars.

*Repro steps*:

Deploy any version of Envoy that uses nghttp2 v1.49.0 and send requests with whitespace-padded header values. These requests should fail but will succeed if Envoy uses v1.48.0.

xjtian:
Commit Message: Disable RFC9113 header value whitespace validation (fixes #23266)

Additional Description: nghttp2 v1.49.0 rejects header values with leading or trailing whitespace, a new behavior which follows [RFC9113](https://datatracker.ietf.org/doc/html/rfc9113#section-8.2.1). Envoy deployments in environments which may see these kinds of header values will see increased HTTP2 connection churn and 5xx. This change configures updates nghttp2 to v1.50.0 and configures it to not perform the RFC9113 whitespace validation.

See additional discussion in https://github.com/nghttp2/nghttp2/issues/1786 and https://github.com/nghttp2/nghttp2/issues/1796 for more context.

Risk Level: low (?) - this maintains Envoy's existing behavior when encountering these kinds of header values, but it does ignore a security-related part of the RFC.

Testing: new unit test

Docs Changes:

Release Notes:

Platform Specific Features:


alyssawilk (approved):
Thank you so much for the quick fix!

yanavlasov (approved):

mattklein123 (approved):
Thank you!

PR #23270, opened 2022-09-28
upstream: deprecated additional_source_addresses with new repeated message

soulxu:

Commit Message: upstream: deprecated additional_source_addresses with new repeated message
Additional Description:
For dual-stack case, people may set different config for different source address (like different socket options, and it discussed at slack https://envoyproxy.slack.com/archives/C78HA81DH/p1664228598624269), the original API isn't future-proof. This PR deprecates the additional_source_addresses with a new repeated of ExtraSourceAddress message, then it enables to add more config bind to the additional source address in the future. Fix this before this feature release in the last release.

https://github.com/envoyproxy/envoy/blob/6bba2d69a17e4cf374eb09f7f51fbf0f0ce4b7da/changelogs/current.yaml#L189-L191

Risk Level: low
Testing: unittest
Docs Changes: API doc
Release Notes: there is an existing release note


soulxu:
cc @ravenblackx also

mattklein123:
I will check with @envoyproxy/api-shepherds to see if we can land this since this was not yet released. I'm inclined to let it in as I can't imagine anyone is using this yet and this will avoid a deprecation.

api/envoy/config/core/v3/address.proto
```diff
@@ -142,7 +147,7 @@ message BindConfig {
   // than one additional address specified, only the first address matched IP version will be
   // returned. If there is no same IP version address found, the address in the `source_address`
   // will be returned.
-  repeated SocketAddress additional_source_addresses = 4;
+  repeated AdditionalSourceAddress additional_source_addresses = 4;
```
wbpcode, line 150: Maybe we can just use `repeated BindConfig` here?

soulxu: emm.. we still need to deprecate `additional_source_addresses`. If we put the additional addresses inside the BindConfig, we still have a place to add config which applies to all addresses.

soulxu:
> Unfortunately according to to the policy (https://github.com/envoyproxy/envoy/blob/main/api/API_VERSIONING.md) we can only make changes up to 14 days after the initial API went in, so I think we have to deprecate.
> 
> /wait

got it, sorry about that, let me deprecate.

api/envoy/config/core/v3/address.proto
soulxu: Two options in my mind:
1. Deprecated `additional_source_addresses`, then add new field called `extra_source_addresses`
2. Introduce a new field `additional_source_address_configs`, then match the additional source address and corresponding config by order.

@mattklein123 WDYT?

wbpcode: I'd like the first one. 🤔

soulxu: > IMO, I'd like the first one. 🤔 Although the second one can avoid the deprecating, but it's a little weird that to match additional source address and corresponding config by order

Yea, the second one is pretty like a 'workaround', maybe not worth it.

api/envoy/config/core/v3/address.proto
```diff
@@ -134,15 +141,20 @@ message BindConfig {
...
   // version with the target host's address IP version will be used as bind address. If more
-  // than one additional address specified, only the first address matched IP version will be
+  // than one extra address specified, only the first address matched IP version will be
   // returned. If there is no same IP version address found, the address in the `source_address`
   // will be returned.
-  repeated SocketAddress additional_source_addresses = 4;
+  repeated ExtraSourceAddress extra_source_addresses = 5;
+
+  // Deprecated by
+  // :ref:`extra_source_addresses <envoy_v3_api_field_config.cluster.v3.UpstreamBindConfig.extra_source_addresses>`
```
soulxu, line 155: The doc test seems failed on the label https://dev.azure.com/cncf/envoy/_build/results?buildId=117819&view=logs&j=91e633a5-4907-5da1-5862-92a79fe2387a&t=f29e24e7-a093-5724-80bd-d79975d268cb&l=102

but I couldn't reproduce this error in my local.

is the label case sensitive? I saw the warning message using the low case of UpstreamBindConfig, ` envoy_v3_api_field_config.cluster.v3.upstreambindconfig.extra_source_addresses`. Or I missed something. 

@phlax do you know what happened here? thanks in advance!

wbpcode: May be the correct label should be `envoy_v3_api_field_config.core.v3.BindConfig.extra_source_addresses`?

soulxu: oh, right, thanks!

this is seems strange https://github.com/envoyproxy/envoy/pull/23270/files#diff-7d01a4eae2ec35f22ebc30d24d2e08800e159a3f913efc9fb50f6de3c59e24a7R133, we have another message called UpstreamBindConfig

soulxu: see if that can be removed https://github.com/envoyproxy/envoy/pull/23323

api/envoy/config/core/v3/address.proto
```diff
@@ -115,6 +116,12 @@ message TcpKeepalive {
   google.protobuf.UInt32Value keepalive_interval = 3;
 }
 
+message ExtraSourceAddress {
+  // The additional address to bind.
+  SocketAddress source_address = 1 [(validate.rules).message = {required: true}];
```
mattklein123, line 121: nit: s/source_address/address, since it's not redundant inside the message.

source/common/upstream/upstream_impl.cc
```diff
@@ -85,6 +113,8 @@ getSourceAddressFnFromBindConfig(const std::string& cluster_name,
           "and additional_source_addresses fields",
           cluster_name.empty() ? "Bootstrap" : fmt::format("Cluster {}", cluster_name)));
     }
+    ENVOY_LOG_MISC(warn, "The `additional_source_addresses` field is already deprecated by "
```
mattklein123, line 116: This can be deleted. This is already handled by the deprecated field tooling.

mattklein123 (changes requested):
Thanks LGTM with small comments.

/wait

api/envoy/config/core/v3/address.proto
soulxu: got it

source/common/upstream/upstream_impl.cc
soulxu: got it

changelogs/current.yaml
```diff
@@ -196,7 +196,7 @@ new_features:
     added a filter state object to control the destination address in :ref:`ORIGINAL_DST clusters <arch_overview_load_balancing_types_original_destination_request_header_filter_state>`.
 - area: upstream
   change: |
-    added a new field :ref:`additional_source_addresses <envoy_v3_api_field_config.core.v3.BindConfig.additional_source_addresses>` to the BindConfig, it enables to specify multiple source addresses, and the source address selection is based on target host's address' version.
+    added a new field :ref:`extra_source_addresses <envoy_v3_api_field_config.core.v3.BindConfig.extra_source_addresses>` to the BindConfig, it enables to specify multiple source addresses, and the source address selection is based on target host's address' version.
```
soulxu, line 199: I change the existing release note directly, it seems we needn't bother the user here, since the `additional_source_address` is deprecated immediately.

PR #23281, opened 2022-09-28
xds: Add xDS delegate feature support to the unified mux

abeyad:
https://github.com/envoyproxy/envoy/pull/22473 and https://github.com/envoyproxy/envoy/pull/22921 added support for the xDS resources delegate in the gRPC Sotw mux. This PR adds support for the xDS resources delegate feature to the unified gRPC Sotw implementation, which is intended to be the default implementation in the future.

### DIFF
--- a/contrib/config/test/kv_store_xds_delegate_integration_test.cc
+++ b/contrib/config/test/kv_store_xds_delegate_integration_test.cc
@@ -56,16 +56,19 @@ std::string invalidProtoKvStoreDelegateConfig() {
     )EOF";
 }
 
-class KeyValueStoreXdsDelegateIntegrationTest : public HttpIntegrationTest,
-                                                public Grpc::GrpcClientIntegrationParamTest {
+class KeyValueStoreXdsDelegateIntegrationTest
+    : public HttpIntegrationTest,
+      public Grpc::UnifiedOrLegacyMuxIntegrationParamTest {
 public:
   KeyValueStoreXdsDelegateIntegrationTest()
       : HttpIntegrationTest(Http::CodecType::HTTP2, ipVersion(),
                             ConfigHelper::baseConfigNoListeners()) {
     use_lds_ = false;
-    // TODO(abeyad): add UnifiedSotw tests too when implementation is ready.
-    sotw_or_delta_ = Grpc::SotwOrDelta::Sotw;
     skip_tag_extraction_rule_check_ = true;
+
+    if (isUnified()) {
+      config_helper_.addRuntimeOverride("envoy.reloadable_features.unified_mux", "true");
+    }
 
     // Make the default cluster HTTP2.
     config_helper_.addConfigModifier([](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
@@ -276,7 +279,7 @@ protected:
 };
 
 INSTANTIATE_TEST_SUITE_P(IpVersions, KeyValueStoreXdsDelegateIntegrationTest,
-                         GRPC_CLIENT_INTEGRATION_PARAMS);
+                         UNIFIED_LEGACY_GRPC_CLIENT_INTEGRATION_PARAMS);
 
 TEST_P(KeyValueStoreXdsDelegateIntegrationTest, BasicSuccess) {
   on_server_init_function_ = [this]() {
@@ -354,10 +357,10 @@ TEST_P(KeyValueStoreXdsDelegateIntegrationTest, BasicSuccess) {
   test_server_->waitForCounterGe("runtime.load_success", 2);
 
   // Verify that the latest resource values are used by Envoy.
-  EXPECT_EQ(2, test_server_->counter("xds.kv_store.load_success")->value());
-  EXPECT_EQ(0, test_server_->counter("xds.kv_store.resources_not_found")->value());
-  EXPECT_EQ(0, test_server_->counter("xds.kv_store.resource_missing")->value());
-  EXPECT_EQ(0, test_server_->counter("xds.kv_store.parse_failed")->value());
+  EXPECT_GE(test_server_->counter("xds.kv_store.load_success")->value(), 2);
+  EXPECT_EQ(test_server_->counter("xds.kv_store.resources_not_found")->value(), 0);
+  EXPECT_EQ(test_server_->counter("xds.kv_store.resource_missing")->value(), 0);
+  EXPECT_EQ(test_server_->counter("xds.kv_store.parse_failed")->value(), 0);
   checkSecretExists(std::string(CLIENT_CERT_NAME), /*version_info=*/"1");
   EXPECT_EQ("whatevs", getRuntimeKey("foo"));
   EXPECT_EQ("yar", getRuntimeKey("bar"));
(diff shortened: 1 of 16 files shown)


abeyad:
@adisuissa friendly ping

source/common/config/xds_mux/sotw_subscription_state.cc
```diff
@@ -71,14 +73,69 @@ void SotwSubscriptionState::handleGoodResponse(
...
+      }
+
+      TRY_ASSERT_MAIN_THREAD {
+        auto decoded_resource = DecodedResourceImpl::fromResource(*resource_decoder_, resource);
+        setResourceTtl(*decoded_resource);
+        decoded_resources.emplace_back(std::move(decoded_resource));
+      }
+      END_TRY
+      catch (const EnvoyException& e) {
+        xds_resources_delegate_->onResourceLoadFailed(source_id, resource.name(), e);
```
adisuissa, line 125: Please add a test that covers this scenario.

Just to make sure, the idea is that all other valid resources will be used, correct?

source/common/config/xds_mux/sotw_subscription_state.cc
```diff
@@ -71,14 +73,69 @@ void SotwSubscriptionState::handleGoodResponse(
...
+      }
+    }
+
+    callbacks().onConfigUpdate(decoded_resources, version_info);
+  }
+  END_TRY
+  catch (const EnvoyException& e) {
+    // TODO(abeyad): do something more than just logging the error?
+    ENVOY_LOG(warn, "xDS delegate failed onEstablishmentFailure() for {}: {}", source_id.toKey(),
```
adisuissa, line 134: Please add a test that covers this scenario.

source/common/config/xds_mux/sotw_subscription_state.cc
```diff
@@ -71,14 +73,69 @@ void SotwSubscriptionState::handleGoodResponse(
...
+        decoded_resources.emplace_back(std::move(decoded_resource));
+      }
+      END_TRY
+      catch (const EnvoyException& e) {
+        xds_resources_delegate_->onResourceLoadFailed(source_id, resource.name(), e);
+      }
+    }
+
+    callbacks().onConfigUpdate(decoded_resources, version_info);
+  }
```
adisuissa, line 130: What should be the values of `last_good_version_info_` and `last_good_nonce_` if this ingestion is successful?

adisuissa (commented):

contrib/config/test/kv_store_xds_delegate_integration_test.cc
```diff
@@ -354,10 +357,10 @@ TEST_P(KeyValueStoreXdsDelegateIntegrationTest, BasicSuccess) {
   test_server_->waitForCounterGe("runtime.load_success", 2);
 
   // Verify that the latest resource values are used by Envoy.
-  EXPECT_EQ(2, test_server_->counter("xds.kv_store.load_success")->value());
-  EXPECT_EQ(0, test_server_->counter("xds.kv_store.resources_not_found")->value());
-  EXPECT_EQ(0, test_server_->counter("xds.kv_store.resource_missing")->value());
-  EXPECT_EQ(0, test_server_->counter("xds.kv_store.parse_failed")->value());
+  EXPECT_GE(test_server_->counter("xds.kv_store.load_success")->value(), 2);
```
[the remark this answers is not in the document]
abeyad: I changed it back and amended `previously_fetched_data_` to ensure we fetch from the store the minimal amount.

source/common/config/xds_mux/sotw_subscription_state.cc
abeyad: Added test `HandleEstablishmentFailureWithInvalidResource`. Yes, the idea is to load all other valid resources.

source/common/config/xds_mux/sotw_subscription_state.cc
abeyad: Added test `HandleEstablishmentFailureWithDelegateThrowingException`

source/common/config/xds_mux/sotw_subscription_state.cc
abeyad: I'm interested in your thoughts on that.  I didn't think I should touch `last_good_version_info_` and `last_good_nonce_` as part of loading cached resources, since those variables are used in request/response interactions with the xDS server.  Do you believe that the cached config should set those values (specifically the `last_good_version_info_` value)?  If so, why?

adisuissa: I think it impacts what will be sent to the server on the [next request](https://github.com/envoyproxy/envoy/blob/ad186c448cf8c5e78454d2adf5f228e9c90a997c/source/common/config/xds_mux/sotw_subscription_state.cc#L88-L90), so it can avoid the response, if the version hasn't changed.

abeyad: My thought was, even if the version is the same, once the management server can be reached again, we should send a request, even if the version is the latest one.  Would resending the same nonce make sense, even if we sent over the cached version number in the next request?

abeyad: Reading https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#basic-protocol-overview, it seems like a nonce does not survive stream restarts/reconnects, so I assume it's already possible to send a non-empty version info with an empty nonce in a DiscoveryRequest?  If that's the case, I can do the same with using the persisted config.  Is my understanding correct?

adisuissa: The nonce is reset, but not the version.
The idea is that by sending the version, the server is able to know if the client already holds the latest and greatest, and if it does, the server won't need to send an update.
Not sure if your use-case fits should also follow this.

abeyad: I think it's tricky.  I'm imagining a couple scenarios: 

1. What if the DiscoveryRequest requests resources {"a", "b", "c"}, but the xDS delegate's getResources() function only had {"a", "b"} cached, but not "c".  Do we still want to set the last known good version?  Or is it guaranteed that, if there's a new resource added to the list (in this case, "c"), the resources would have a new version too?

2. What if the xDS delegate returned {"a", "b", "c"}, but "c" was malformed (perhaps an invalid enum value) and so doesn't get used.  Should we still set the last known good version to the version found in "a" and "b"?

abeyad: It seems like the version always gets updated if any resource gets changed/added/removed for that resource type: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ack-nack-and-resource-type-instance-version.  I pushed https://github.com/envoyproxy/envoy/pull/23281/commits/08722407f5e6adb6283087e167d1b930c8d04b7b which addresses this scenario. PTAL.

abeyad:
@adisuissa any other concerns/feedback on the PR?

source/common/config/xds_mux/sotw_subscription_state.cc
```diff
@@ -71,14 +73,83 @@ void SotwSubscriptionState::handleGoodResponse(
...
+  const XdsConfigSourceId source_id{target_xds_authority_, type_url_};
+  TRY_ASSERT_MAIN_THREAD {
+    const std::vector<std::string> resource_names{names_tracked_.begin(), names_tracked_.end()};
+    std::vector<envoy::service::discovery::v3::Resource> resources =
+        xds_resources_delegate_->getResources(source_id, resource_names);
+
+    std::vector<DecodedResourcePtr> decoded_resources;
+    const auto scoped_update = ttl_.scopedTtlUpdate();
+    std::string version_info;
+    absl::flat_hash_set<std::string> unaccounted{names_tracked_.begin(), names_tracked_.end()};
```
adisuissa, line 111: Looking at this, it made me think why `resource_names` and `resources` are defined as vectors (from a first glance, at least `resource_names` might not be needed as one can pass `names_tracked_` IIUC).

Also, is tracking the actual set is needed, or can it be a number of parsed resources (again assuming they are all different).

adisuissa (approved):
LGTM, thanks!

/retest

abeyad:
/assign @lizan 

Alyssa is OOO for the next few days, assigning to Lizan instead

RyanTheOptimist:
@snowp can you do a cross-org /sr maintainer review? @lizan seems to be offline

PR #23292, opened 2022-09-28
ci: Add branch and remote to Bazel get_workspace_status

WillEngFlow:
Signed-off-by: Will Martin <will@engflow.com>

Commit Message:
This change adds the BUILD_SCM_REMOTE and BUILD_SCM_BRANCH variables to those reported by Bazel. This will allow downstream consumers of the BEP within the CI system to include information regarding the repository and branch that a build was triggered from.

Additional Description: 
Specifically, this adds repository and branch to the [EngFlow UI][1] for envoy-mobile but may be used by any BEP consumer.

Risk Level: Low
Testing: Manual
Docs Changes: None
Release Notes: None

[1]: https://envoy.cluster.engflow.com/

### DIFF
--- a/bazel/get_workspace_status
+++ b/bazel/get_workspace_status
@@ -45,8 +45,8 @@ git_branch=$(git rev-parse --abbrev-ref HEAD)
 echo "BUILD_SCM_BRANCH ${git_branch}"
 
 git_remote=$(git remote get-url origin)
-git_remote_code=$?
-if [[ git_remote_code == 0 ]]; then
+git_remote_exit_code=$?
+if [[ git_remote_exit_code == 0 ]]; then
     echo "BUILD_SCM_REMOTE ${git_remote}"
 fi
 
--- a/bazel/get_workspace_status
+++ b/bazel/get_workspace_status
@@ -40,3 +40,13 @@ git diff-index --quiet HEAD -- || {
 }
 echo "BUILD_SCM_STATUS ${tree_status}"
 echo "STABLE_BUILD_SCM_STATUS ${tree_status}"
+
+git_branch=$(git rev-parse --abbrev-ref HEAD)
+echo "BUILD_SCM_BRANCH ${git_branch}"
+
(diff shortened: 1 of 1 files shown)


phlax (dismissed):
lgtm, thanks @WillEngFlow

phlax:
seems like its failing shellcheck

```console
In bazel/get_workspace_status line 48:
if [[ $? == 0 ]]; then
      ^-- SC2181 (style): Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
```

WillEngFlow:
Ah, thanks guys! This is a little tricky because I want to store the value in a variable and read the exit code separately. Let me try a different way and see if I can please shellcheck!

bazel/get_workspace_status
```diff
@@ -45,8 +45,8 @@ git_branch=$(git rev-parse --abbrev-ref HEAD)
 echo "BUILD_SCM_BRANCH ${git_branch}"
 
 git_remote=$(git remote get-url origin)
-git_remote_code=$?
-if [[ git_remote_code == 0 ]]; then
+git_remote_exit_code=$?
```
phlax, line 48: how about just checking if `git_remote` was set - shellcheck doesnt like checking the return code and i cant off hand think of a way of iffing and setting at the same time - however, i think if previous command exitted non 0 then `git_remote` would not be set - so i think you can `if [[ -n "$git_remote" ]]; then ...`

phlax: for more assurance you could change to...

```bash
git_remote=$(git remote get-url origin || echo "")
```

WillEngFlow: I like that idea, let me give it a go!

phlax (approved):
lgtm, thanks @WillEngFlow

WillEngFlow:
Thanks for walking me through that @phlax! Is there any more action required on my part before this can be merged?

phlax:
can be merged - i was waiting on ci

PR #23293, opened 2022-09-28
Clarify that filters with no filter_chain_match are always matched

bradjones1:
Signed-off-by: Brad Jones <brad@kinksters.dating>

Commit Message: Clarifies documentation on omitted `filter_chain_match`
Additional Description:
Risk Level:
Testing:
Docs Changes: Clarifies documentation on omitted `filter_chain_match`
Release Notes:
Platform Specific Features:

### DIFF
--- a/api/envoy/config/listener/v3/listener_components.proto
+++ b/api/envoy/config/listener/v3/listener_components.proto
@@ -221,6 +221,7 @@ message FilterChain {
   reserved "tls_context";
 
   // The criteria to use when matching a connection to this filter chain.
+  // If omitted, the filter is matched always/unconditionally.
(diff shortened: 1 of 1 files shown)


bradjones1:
At least, I think this is the case? Various example recipes in the docs omit this value, and the docs are unclear?

htuch:
I think it is the most specific match wins. If you have an empty matcher and no competing matchers, sure, I think this is successful. https://github.com/envoyproxy/envoy/blob/02489bbf8990faddb6be1f2e0f22851b64c3fc85/api/envoy/config/listener/v3/listener.proto#L132

Please share any examples you'd like to use to motivate hate.

PR #23294, opened 2022-09-28
Skipping DNS lookup for the DFP when proxying.

alyssawilk:
When doing upstream connect proxying using the new transport socket, upstream connections go to the proxy address rather than the host address. This PR allows skipping DNS lookup for the host while creating the host in the DFP cluster.

Risk Level: medium 
Testing: new integration test
Docs Changes: n/a
Release Notes: inline
Runtime guard: envoy.reloadable_features.skip_dns_lookup_for_proxied_requests


source/common/upstream/upstream_impl.cc
```diff
@@ -62,6 +62,13 @@ namespace Envoy {
 namespace Upstream {
 namespace {
 
+std::string addressToString(Network::Address::InstanceConstSharedPtr address) {
+  if (!address) {
+    return "";
+  }
+  return address->asString();
+}
```
mattklein123, line 70: I'm concerned that there is a lot of code that probably assumes that hosts always have addresses. Did you consider making some kind of synthetic address as a sentinel for the proxied case? That might also make a lot of the special cases in this PR simpler?

mattklein123 (commented):
Just for my understanding, this is a performance improvement only, right? Is there a functionality/bug fix here also? One high level question first. Thanks for working on this.

source/common/upstream/upstream_impl.cc
alyssawilk: We already have to handle nullptr addresses for UDS, which is why the fixes were in the DynamicCluster code and not generic upstream code.  I'm inclined to think we're better off handling null addresses than having something synthetic, but I added a runtime guard in case the current testing wasn't sufficient (rather than relying on Lyft's already off-by-default rollout of the proxy code)

alyssawilk: Oh also this is largely a latency/perf fix.  It's generally a latency win doing a DNS lookup for an endpoint we're not connecting to (proxy code connects to proxy address) but it's also possible that folks setting up proxying on a corp network would block DNS so this would make requests work where the might not otherwise.

source/extensions/common/dynamic_forward_proxy/dns_cache_impl.cc
```diff
@@ -177,7 +180,13 @@ void DnsCacheImpl::startCacheLoad(const std::string& host, uint16_t default_port
   }
 
   primary_host = createHost(host, default_port);
-  startResolve(host, *primary_host);
+  // If the DNS request was simply to create a host endpoint in a Dynamic Forward Proxy cluster,
+  // fast fail the look-up as the address is not needed.
+  if (is_proxy_lookup) {
+    finishResolve(host, Network::DnsResolver::ResolutionStatus::Failure, {}, {}, true);
```
mattklein123, line 186: This is going to increment the failure stat, right? I think that might be confusing? Should this either not increment that stat or be some kind of success type?

envoy/upstream/upstream.h
```diff
@@ -1163,8 +1163,9 @@ namespace fmt {
 template <> struct formatter<Envoy::Upstream::Host> : formatter<absl::string_view> {
   template <typename FormatContext>
   auto format(const Envoy::Upstream::Host& host, FormatContext& ctx) -> decltype(ctx.out()) {
-    absl::string_view out =
-        host.hostname().empty() ? host.hostname() : host.address()->asStringView();
+    absl::string_view out = host.hostname().empty() ? host.hostname()
+                            : host.address()        ? host.address()->asStringView()
```
mattklein123, line 1167: Not your code but shouldn't this be: ?

```suggestion
    absl::string_view out = !host.hostname().empty() ? host.hostname()
                            : host.address()        ? host.address()->asStringView()
```

mattklein123 (changes requested):
Thanks LGTM with 2 small comments, thank you.

/wait

mattklein123 (approved):
Thanks!

PR #23308, opened 2022-09-29
extensions_schema: add category envoy.tls_handshakers.

ipuustin:
Commit Message: extensions: add category envoy.tls_handshakers.
Additional Description:

`HandshakerFactory` at [envoy/ssl/handshaker.h](https://github.com/envoyproxy/envoy/blob/main/envoy/ssl/handshaker.h#L130) specifies extension category `envoy.tls_handshakers`, but it's missing from the extensions schema. Add it so that we can specify extensions in that category.

Risk Level: low
Testing: manual
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/tools/extensions/extensions_schema.yaml
+++ b/tools/extensions/extensions_schema.yaml
@@ -113,6 +113,7 @@ categories:
 - envoy.filters.http.upstream
 - envoy.path.match
 - envoy.path.rewrite
+- envoy.tls_handshakers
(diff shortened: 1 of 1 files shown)


kyessenov (approved):
The reason why it hasn't been a problem is because we don't have any typed extension for handshakers. They are built-in and can only be instantiated internally.

PR #23316, opened 2022-09-29
Allow disabling the guard dog

jpsim:
Commit Message: Allow disabling the guard dog
Additional Description: Which Envoy Mobile should never enable, because terminating the Envoy process would mean terminating the host iOS or Android app. See https://github.com/envoyproxy/envoy-mobile/issues/2581 for more details.
Risk Level: Low, compile-time flag, no-op by default.
Testing: Building Envoy Mobile locally.
Docs Changes: None.
Release Notes: None.
Platform Specific Features: Will be disabled when building Envoy Mobile.
Fixes: https://github.com/envoyproxy/envoy-mobile/issues/2581

### DIFF
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -302,6 +302,27 @@ config_setting(
 )
 
 bool_flag(
+    name = "guarddog",
+    build_setting_default = True,
+    visibility = ["//visibility:private"],
+)
+
+config_setting(
+    name = "disable_guarddog_setting",
+    flag_values = {
+        ":guarddog": "False",
+    },
+    visibility = ["//visibility:private"],
+)
+
+selects.config_setting_group(
+    name = "disable_guarddog",
+    match_any = [
+        ":disable_guarddog_setting",
+    ],
+)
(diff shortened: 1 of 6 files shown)


jpsim:
As discussed in today's Envoy Mobile community call. /cc @alyssawilk @RyanTheOptimist

Thanks @Reflejo for flagging this.

jpsim:
Testing in Envoy Mobile:

```console
# Before
$ ./bazelw build --config=ios //examples/swift/async_await:app
$ nm Payload/app.app/app | grep GuardDogImpl | wc -l
641
# After
$ ./bazelw build --config=ios //examples/swift/async_await:app
$ nm Payload/app.app/app | grep GuardDogImpl | wc -l
0
```

jpsim:
Hmmm, I think the watch dog is currently responsible for actually _running_ the main dispatch loop, so to disable it I'll actually need to run the main dispatch loop some other way.

I've marked this PR as being a draft until I can do that.

PR #23318, opened 2022-09-29
Persist key-value-store TTLs to disk in flush()

caschoener:
Turns out the TTL functionality I added needs the ability to persist to disk.  Do this by piggybacking on the existing persistence code; writing an optional absolute TTL to disk and then converting it back to relative form for the `ttl_manager` to use.

### DIFF
--- a/source/common/common/key_value_store_base.cc
+++ b/source/common/common/key_value_store_base.cc
@@ -1,5 +1,4 @@
 #include "source/common/common/key_value_store_base.h"
-#include "source/common/common/key_value_store_base.h"
 
 #include <algorithm>
 #include <chrono>
--- a/source/common/common/key_value_store_base.cc
+++ b/source/common/common/key_value_store_base.cc
@@ -1,7 +1,5 @@
 #include "source/common/common/key_value_store_base.h"
 
-#include <bits/chrono.h>
-
 #include <algorithm>
 #include <chrono>
 
--- a/source/common/common/key_value_store_base.cc
+++ b/source/common/common/key_value_store_base.cc
@@ -79,14 +79,15 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
         ttl.emplace(std::chrono::duration_cast<std::chrono::seconds>(
             std::chrono::time_point<std::chrono::system_clock>(std::chrono::seconds(ttl_int)) -
             time_source_.systemTime()));
+        if (ttl <= std::chrono::seconds(0)) {
+          continue;
+        }
       } else {
         ENVOY_LOG(warn, "TTL was read from disk but failed to convert to integer");
         return false;
       }
     }
-    if (!ttl.has_value() || ttl > std::chrono::seconds(0)) {
-      addOrUpdate(key.value(), value.value(), ttl);
-    }
+    addOrUpdate(key.value(), value.value(), ttl);
   }
   return true;
 }
--- a/source/common/common/key_value_store_base.cc
+++ b/source/common/common/key_value_store_base.cc
@@ -50,7 +50,8 @@ KeyValueStoreBase::KeyValueStoreBase(Event::Dispatcher& dispatcher,
         flush_timer_->enableTimer(flush_interval);
       })),
       ttl_manager_([this](const std::vector<std::string>& expired) { onExpiredKeys(expired); },
-                   dispatcher, dispatcher.timeSource()) {
+                   dispatcher, dispatcher.timeSource()),
+      time_source_(dispatcher.timeSource()) {
   if (flush_interval.count() > 0) {
     flush_timer_->enableTimer(flush_interval);
   }
@@ -75,7 +76,9 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
       if (token.has_value()) {
         if (absl::SimpleAtoi(token.value(), &ttlInt)) {
           ttl.emplace(std::chrono::duration_cast<std::chrono::seconds>(
-              std::chrono::seconds(ttlInt) - std::chrono::system_clock::now().time_since_epoch()));
+              std::chrono::time_point<std::chrono::steady_clock>(std::chrono::seconds(ttlInt)) -
+              time_source_.monotonicTime()));
+          std::cout << ttl.value().count();
         }
       }
     }
@@ -108,8 +111,8 @@ void KeyValueStoreBase::addOrUpdate(absl::string_view key_view, absl::string_vie
   }
   if (ttl) {
     ttl_manager_.add(std::chrono::milliseconds(ttl.value()), key);
-    std::chrono::seconds absolute_ttl = std::chrono::duration_cast<std::chrono::seconds>(
-        ttl.value() + std::chrono::system_clock::now().time_since_epoch());
+    auto absolute_ttl = ttl.value() + std::chrono::duration_cast<std::chrono::seconds>(
+                                          time_source_.monotonicTime().time_since_epoch());
     if (!ttl_store_.emplace(key, absolute_ttl)
              .second) { // ttl_store_ should probably be merged with store_ as a tuple
       ttl_store_.erase(key);
--- a/source/common/common/key_value_store_base.cc
+++ b/source/common/common/key_value_store_base.cc
@@ -1,6 +1,7 @@
 #include "source/common/common/key_value_store_base.h"
 
 #include <algorithm>
+#include <bits/chrono.h>
 #include <chrono>
 
 #include "absl/cleanup/cleanup.h"
@@ -31,6 +32,14 @@ absl::optional<absl::string_view> getToken(absl::string_view& contents, std::str
   return token;
 }
 
+bool checkForTtl(absl::string_view& contents) {
+  if (contents.size() > 3 && contents.substr(0, 3) == "TTL") {
+    contents.remove_prefix(3);
+    return true;
+  }
+  return false;
+}
+
 } // namespace
 
 KeyValueStoreBase::KeyValueStoreBase(Event::Dispatcher& dispatcher,
@@ -51,6 +60,7 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
   while (!contents.empty()) {
     absl::optional<absl::string_view> key = getToken(contents, error);
     absl::optional<absl::string_view> value;
+    absl::optional<std::chrono::seconds> ttl = absl::nullopt;
     if (key.has_value()) {
       value = getToken(contents, error);
     }
@@ -58,7 +68,17 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
       ENVOY_LOG(warn, error);
       return false;
     }
-    addOrUpdate(key.value(), value.value(), absl::nullopt);
+    if (checkForTtl(contents)) {
+      uint64_t ttlInt;
+      auto token = getToken(contents, error);
+      if (token.has_value()) {
+        if (absl::SimpleAtoi(token.value(), &ttlInt)) {
+          ttl.emplace(std::chrono::duration_cast<std::chrono::seconds>(std::chrono::seconds(ttlInt) - std::chrono::system_clock::now().time_since_epoch()));
+        }
+      }
+    }
+
+    addOrUpdate(key.value(), value.value(), ttl);
   }
   return true;
 }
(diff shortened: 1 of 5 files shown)


source/extensions/key_value/file_based/config.cc
```diff
@@ -36,6 +36,13 @@ void FileBasedKeyValueStore::flush() {
     file->write(it.first);
     file->write(absl::StrCat(it.second.length(), "\n"));
     file->write(it.second);
+    auto ttl_it = ttl_store_.find(it.first);
+    if (ttl_it != ttl_store_.end()) {
+      std::string ttl = std::to_string(ttl_it->second.count());
+      file->write("TTL");
```
abeyad, line 42: can we factor this out into a variable?  i also suggest something more obscure like `__KV_TTL__` just to minimize the chances of conflicting with the ttl key in the serialized file being part of the key in another.

source/common/common/key_value_store_base.cc
```diff
@@ -31,6 +31,14 @@ absl::optional<absl::string_view> getToken(absl::string_view& contents, std::str
   return token;
 }
 
+bool checkForTtl(absl::string_view& contents) {
+  if (contents.size() > 3 && contents.substr(0, 3) == "TTL") {
```
abeyad, line 35: can we factor out into a variable and then use TTL_KEY.size() instead of hard coding 3?

source/common/common/key_value_store_base.cc
```diff
@@ -51,14 +60,28 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+      uint64_t ttlInt;
+      auto token = getToken(contents, error);
+      if (token.has_value()) {
+        if (absl::SimpleAtoi(token.value(), &ttlInt)) {
+          ttl.emplace(std::chrono::duration_cast<std::chrono::seconds>(
+              std::chrono::time_point<std::chrono::steady_clock>(std::chrono::seconds(ttlInt)) -
+              time_source_.monotonicTime()));
+          std::cout << ttl.value().count();
+        }
+      }
```
abeyad, line 81: may want to at least log that we got a malformed TTL value (could indicate a problem in parsing)

abeyad (commented):
Thanks Chris! Left a few comments

source/extensions/key_value/file_based/config.cc
caschoener: There isn't actually a risk of a conflict.  This file looks like `\n{len(key)}{key}\n{len(value)}{value}\nTTL{len(ttl)}{ttl}`

"TTL" goes before the length, so the only options are to see "TTL" or an integer representing length.

source/common/common/key_value_store_base.cc
```diff
@@ -70,7 +93,6 @@ void KeyValueStoreBase::addOrUpdate(absl::string_view key_view, absl::string_vie
   std::string value(value_view);
   // Do not add if ttl is <= 0
   if (ttl && ttl <= std::chrono::seconds(0)) {
-    ASSERT(false);
```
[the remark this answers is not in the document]
caschoener: Previously, TTL<0 represented a malformed input since there's no good reason e.g. traffic director would send a negative TTL (I think).  But this update actually uses negative TTLs here, if a TTL expires before the store gets reloaded from disk.

source/extensions/key_value/file_based/config.cc
abeyad: ah you're right, makes sense, thanks!

source/common/common/key_value_store_base.cc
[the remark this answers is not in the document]
abeyad: I think if the TTL has expired, we shouldn't reload from disk, i.e. we can not call addOrUpdate if the ttl in https://github.com/envoyproxy/envoy/pull/23318/files/537f0b1df57b9a7fc03898dca95d653db241d739#diff-b9ac1e7fb952c169f6ee95b58cb15c65ac31c8e8b421f521c13546e6b511fd8eR76 is a negative value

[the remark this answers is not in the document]
caschoener: Yeah.  But the way it works out as written is that we call `addOrUpdate` with a negative TTL, and then hit this `if` statement and return without doing anything.

The alternative would be to put the logic enforcing ttl>0 in two different functions which seems worse to me.

source/common/common/key_value_store_base.cc
caschoener: This is a great point, I'm probably supposed to put "TTL" somewhere as a constant and just reference it from there.

Tricky though... I need to access "TTL" in both checkForTtl and FileBasedKeyValueStore::flush.  I think I'd have to refactor these files to share a constant. lol

abeyad: True, I think the best option is to put it in key_value_store_base.h, since FileBasedKeyValueStore needs to depend on that file already.

source/common/common/key_value_store_base.cc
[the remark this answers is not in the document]
abeyad: In `addOrUpdate`, you could just keep the assert (or log a warning, or something).  That way, we know if we passed in something with a negative TTL, which should be an error.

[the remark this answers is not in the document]
caschoener: Ok I will return it to the previous format.

source/common/common/key_value_store_base.cc
```diff
@@ -71,22 +70,23 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
           ENVOY_LOG(warn, error);
         } else {
-          ENVOY_LOG(warn, "TTL was read from disk, but aoti() failed");
+          ENVOY_LOG(warn, "TTL was read from disk but failed to convert to integer");
         }
       }
     }
-
-    addOrUpdate(key.value(), value.value(), ttl);
+    if (!(ttl.has_value() && ttl <= std::chrono::seconds(0))) {
```
abeyad, line 87: nit: personal preference, but I find this harder to read... I think it would be easier to read if it was:

`if (!ttl.has_value() || ttl > std::chrono::seconds(0))`

abeyad (commented):
Thanks Chris, looks good! Just left a couple minor comments.

abeyad (approved):
Thanks Chris!

abeyad (dismissed):

source/common/common/key_value_store_base.cc
```diff
@@ -51,14 +61,32 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+            time_source_.monotonicTime()));
+      } else {
+        if (!token.has_value()) {
+          ENVOY_LOG(warn, error);
+        } else {
+          ENVOY_LOG(warn, "TTL was read from disk but failed to convert to integer");
+        }
+      }
+    }
+    if (!ttl.has_value() || ttl > std::chrono::seconds(0)) {
```
snowp, line 87: Would we ever encode a 0 TTL?

snowp (changes requested):
Thanks, a few questions

source/common/common/key_value_store_base.cc
```diff
@@ -73,21 +101,27 @@ void KeyValueStoreBase::addOrUpdate(absl::string_view key_view, absl::string_vie
     ASSERT(false);
     return;
   }
+  absl::optional<std::chrono::seconds> absolute_ttl = absl::nullopt;
+  if (ttl) {
+    absolute_ttl.emplace(ttl.value() + std::chrono::duration_cast<std::chrono::seconds>(
+                                           time_source_.monotonicTime().time_since_epoch()));
```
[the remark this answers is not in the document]
abeyad: I think Snow is right, since we're loading from disk after restarts, I don't think `std::chrono::steady_clock` would allow us to capture the time elapsed since the value was persisted.

source/common/common/key_value_store_base.cc
caschoener: These TTLs are being calculated from disk.  So in theory if we happen to load a key from disk the exact second it's TTL expires, we should drop it.

source/common/common/key_value_store_base.cc
```diff
@@ -51,14 +61,32 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+    if (checkForTtl(contents)) {
+      uint64_t ttl_int;
+      auto token = getToken(contents, error);
+      if (token.has_value() && absl::SimpleAtoi(token.value(), &ttl_int)) {
+        ttl.emplace(std::chrono::duration_cast<std::chrono::seconds>(
+            std::chrono::time_point<std::chrono::steady_clock>(std::chrono::seconds(ttl_int)) -
+            time_source_.monotonicTime()));
+      } else {
+        if (!token.has_value()) {
+          ENVOY_LOG(warn, error);
```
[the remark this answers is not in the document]
caschoener: Did this.  On second look I also changed it to return `false` for the whole `parseContents()` call if we get an invalid TTL after `checkForTtl()`  says there should be one.  Seems fair to match the behavior of the rest of the function.

source/common/common/key_value_store_base.cc
```diff
@@ -51,14 +60,33 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
+      if (absl::SimpleAtoi(token.value(), &ttl_int)) {
+        ttl.emplace(std::chrono::duration_cast<std::chrono::seconds>(
+            std::chrono::time_point<std::chrono::system_clock>(std::chrono::seconds(ttl_int)) -
+            time_source_.systemTime()));
+      } else {
+        ENVOY_LOG(warn, "TTL was read from disk but failed to convert to integer");
+        return false;
+      }
+    }
+    if (!ttl.has_value() || ttl > std::chrono::seconds(0)) {
```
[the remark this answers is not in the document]
caschoener: Make sense!

abeyad (approved):

source/common/common/key_value_store_base.cc
```diff
@@ -51,14 +60,34 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
     if (!key.has_value() || !value.has_value()) {
       ENVOY_LOG(warn, error);
       return false;
     }
-    addOrUpdate(key.value(), value.value(), absl::nullopt);
+    if (checkForTtl(contents)) {
+      uint64_t ttl_int;
+      auto token = getToken(contents, error);
+      if (!token.has_value()) {
+        ENVOY_LOG(warn, error);
```
snowp, line 75: Maybe prefix this error message with something? "Failed to read token" or something like that

source/common/common/key_value_store_base.cc
```diff
@@ -72,21 +71,23 @@ bool KeyValueStoreBase::parseContents(absl::string_view contents) {
...
-        } else {
-          ENVOY_LOG(warn, "TTL was read from disk but failed to convert to integer");
+            std::chrono::time_point<std::chrono::system_clock>(std::chrono::seconds(ttl_int)) -
+            time_source_.systemTime()));
+        if (ttl <= std::chrono::seconds(0)) {
+          continue;
         }
+      } else {
+        ENVOY_LOG(warn, "TTL was read from disk but failed to convert to integer");
+        return false;
```
snowp, line 87: I think you could do an early return to reduce some of this nesting, e.g. `if (!absl::SimpleAtoi(...)) { return false }`

snowp (changes requested):
Thanks, a few more nits and then I think this is good to go

test/config/utility.cc
```diff
@@ -994,9 +996,12 @@ void ConfigHelper::setPorts(const std::vector<uint32_t>& ports, bool override_po
         auto locality_lb = cluster->mutable_load_assignment()->mutable_endpoints(j);
         for (int k = 0; k < locality_lb->lb_endpoints_size(); ++k) {
           auto lb_endpoint = locality_lb->mutable_lb_endpoints(k);
+          std::cout << "WE ARE IN THERE";
           if (lb_endpoint->endpoint().address().has_socket_address()) {
+            std::cout << "1";
             if (lb_endpoint->endpoint().address().socket_address().port_value() == 0 ||
                 override_port_zero) {
+              std::cout << "2";
               RELEASE_ASSERT(ports.size() > port_idx, "");
```
snowp, line 1005: Probably don't want to leave these in :)

snowp (changes requested):

snowp (approved):
Thanks!

PR #6018, opened 2019-02-21
upstream: add transport socket failure reason to stream info and log

lizan:
*Description*:
Fixes #5603 

*Risk Level*: Low (not changing flow, add more info)
*Testing*: unit test,
*Docs Changes*: Added
*Release Notes*: Added

PR #23319, opened 2022-09-29
tls: fix error-reporting in doSynchronousVerifyCertChain

davidben:
This PR is a preparatory fix for PR #23320.

The intent was to upload a pair of dependent PRs. I've made them into separate PRs because it looked like you all prefer one commit per PR (since PR descriptions ask for a commit message), and it seemed worth keeping the changes ultimately separate. Let me know if this isn't the preferred workflow.

---

Commit Message:

This fixes a bug in the pre-tls_async_cert_validation codepath. The tls_async_cert_validation codepath is unaffected.

Per both OpenSSL and BoringSSL documentation, if you use SSL_CTX_set_cert_verify_callback, the library expects to find the error code in X509_STORE_CTX. This error determines the TLS alert. X509_verify_cert does this for you, but Envoy's callback was not doing this when doSynchronousVerifyCertChain runs the extra checks in verifyCertAndUpdateStatus. This change fixes this and sets the generic X509_V_ERR_APPLICATION_VERIFICATION.

Before this commit, the error would usually be X509_V_OK, but sometimes it would be something else. Envoy implements the
allow_expired_certificate using the X509_STORE_CTX verify callback, via ignoreCertificateExpirationCallback. This callback, when it suppresses an error, will sometimes leave stray errors in X509_STORE_CTX. These are usually benign, *but* if Envoy then does extra checks, which fail, and then report some *other* failure, it will inadvertantly return the stray error to BoringSSL!

That is why FailedClientCertAllowExpiredBadHashVerification expected a SSLV3_ALERT_CERTIFICATE_EXPIRED alert. The alert makes no sense because the connection sets allow_expired_certificate! Rather, the test actually caught an Envoy bug, but https://github.com/envoyproxy/envoy/pull/6018 mistakenly codified it as expected behavior, rather than correctly diagnosing it as a bug. This commit fixes that bug.

Unfortunately, OpenSSL/BoringSSL map X509_V_ERR_APPLICATION_VERIFICATION to handshake_failure rather than certificate_unknown, so this does change the alert returned by the pre-tls_async_cert_validation codepath in other cases. (certificate_unknown came from X509_V_OK hitting the default in a switch-case.) Arguably certificate_unknown is a better alert than handshake_failure, so I'm slightly inclined to fix BoringSSL to map that differently. But since Envoy needs to support an old BoringSSL, I've just updated the expectations for now. Either way, tls_async_cert_validation gives full control over the alert and will make this moot.

Signed-off-by: David Benjamin <davidben@google.com>

Additional Description:
Risk Level: Low
Testing: `bazelist test //test/extensions/transport_sockets/tls`
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -109,6 +109,9 @@ bug_fixes:
 - area: upstream
   change: |
     fixed a bug where custom transport socket hashes might not be respected by wrapper passthrough sockets. This change can be temporarily reverted by setting runtime guard ``envoy.reloadable_features.fix_hash_key`` to false.
+- area: tls
+  change: |
+    fixed a bug where, when runtime guard ``envoy.reloadable_features.tls_async_cert_validation`` is set to false, the wrong TLS alerts would sometimes be sent in response to certificate validation failures.
 
 removed_config_or_runtime:
 # *Normally occurs at the end of the* :ref:`deprecation period <deprecated>`
--- a/source/extensions/transport_sockets/tls/cert_validator/default_validator.cc
+++ b/source/extensions/transport_sockets/tls/cert_validator/default_validator.cc
@@ -209,10 +209,12 @@ int DefaultCertValidator::doSynchronousVerifyCertChain(
       return allow_untrusted_certificate_ ? 1 : ret;
     }
   }
-  return verifyCertAndUpdateStatus(ssl_extended_info, &leaf_cert, transport_socket_options, nullptr,
-                                   nullptr)
-             ? 1
-             : 0;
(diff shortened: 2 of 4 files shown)


ggreenway (changes requested):
Thanks for the fix! Can you please add a release note [here](https://github.com/envoyproxy/envoy/blob/3fe2e475b14d7408d35ea6c574296a3649aa1962/changelogs/current.yaml#L92)?

davidben:
> Can you please add a release note [here](https://github.com/envoyproxy/envoy/blob/3fe2e475b14d7408d35ea6c574296a3649aa1962/changelogs/current.yaml#L92)?

Done. Let me know if did that wrong.

ggreenway (approved):

PR #23320, opened 2022-09-29
tls: use X509_V_FLAG_NO_CHECK_TIME to implement allow_expired_certificate

davidben:
~~This PR includes PR #23319 as its first commit. For review, I'd suggest just looking at the second commit, and leaving the first commit to be reviewed in the base PR.~~

~~The intent was to upload a pair of dependent PRs. I've made them into separate PRs because it looked like you all prefer one commit per PR (since PR descriptions ask for a commit message), and it seemed worth keeping the changes ultimately separate. Let me know if this isn't the preferred workflow.~~

Update: #23319 is now merged, so this has been rebased on top of it.

---

Commit Message:
Newer revisions of BoringSSL provide OpenSSL's X509_V_FLAG_NO_CHECK_TIME API. Envoy currently implements allow_expired_certificate by using the X509_STORE_CTX verify callback. This callback is difficult for BoringSSL to support as it exposes internal details of the verification process. (What order errors come in, that there isn't a real path-builder, etc.) It also leaves X509_STORE_CTX in a funny state (per the error-reporting bug fixed before this commit).

This change replaces this callback with a more direct API. Without removing the callback, it will be very difficult to migrate Envoy to the future replacement certificate verifier in BoringSSL.

As Envoy doesn't track BoringSSL HEAD, and needs to support a very old boringssl_fips config, I've retained the callback-based path for when X509_V_FLAG_NO_CHECK_TIME is unavailable. Once Envoy's minimum BoringSSL revision is new enough, this can all be removed in favor of just calling X509_STORE_CTX_set_flags in default_validator.cc and spiffe_validator.cc.

utility_test.cc is removed because it wasn't actually testing anything, just that callback recognized a pair of return values, without even driving a certificate verification. The actual test coverage came from spiffe_validator_test and ssl_socket_test.

Signed-off-by: David Benjamin <davidben@google.com>

Additional Description:
Risk Level: Low
Testing: `bazelist test //test/extensions/transport_sockets/tls/...`, also with BoringSSL updated to HEAD to make sure both codepaths worked
Docs Changes: N/A
Release Notes: N/A


source/extensions/transport_sockets/tls/cert_validator/utility.cc
```diff
@@ -5,7 +5,12 @@ namespace Extensions {
 namespace TransportSockets {
 namespace Tls {
 
-int CertValidatorUtil::ignoreCertificateExpirationCallback(int ok, X509_STORE_CTX* store_ctx) {
+// When the minimum supported BoringSSL includes
+// https://boringssl-review.googlesource.com/c/boringssl/+/53965, remove this and have
+// callers just set `X509_V_FLAG_NO_CHECK_TIME` directly.
```
ggreenway, line 10: Can you add a compile-time failure for when that happens, so we don't forget to do this? Something like 

```
#if defined(BORINGSSL_FIPS) && defined(X509_V_FLAG_NO_CHECK_TIME)
#error "Remove legacy codepath"
#endif
```

ggreenway (changes requested):
/wait

source/extensions/transport_sockets/tls/cert_validator/utility.cc
ggreenway: Sure, a comment in repository_locations sounds good.

davidben: Done

ggreenway:
> Forced-pushed to rebase atop the merge of #23319.

In the future, please merge instead of rebasing; it makes PRs to easier to review on github (due to github's poor support for rebasing).

PR #7236, opened 2019-06-11
 upstream: simplify warming cluster CDS pausing

fredlas:
Replaced callback with a simpler, more centralized approach: the time to potentially pause/unpause is whenever the warming cluster count changes. That centralization removes some previous fragility around removing still-warming clusters.

Risk Level: medium
Testing: moved testing from cds_api_impl_test to ads_integration_test.

Issue #5933, opened 2019-02-13
LDS cannot find new cluster because it's still warming up


winguse:
We are using an ADS service to deliver LDS/CDS to envoy. We found in some cases, it will fail to add new cluster because of LDS complaining about the newly added cluster is unknown.

Here is the sequence diagram of added a new cluster:

![image](https://user-images.githubusercontent.com/1443504/52702879-138ce700-2fb8-11e9-87ee-56c0dcdc7bb8.png)

So:

1. Is that possible to make Envoy ack CDS after warming is done?
2. Or how do we decide what is the correct time for delivering LDS update to Envoy?

Thanks!

Envoy version: 1.9.0

Envoy Log:

```
[2019-02-13 09:04:40.313][000007][info][upstream] [source/common/upstream/cluster_manager_impl.cc:495] add/update cluster zipkin during init
[2019-02-13 09:04:40.315][000007][info][upstream] [source/common/upstream/cluster_manager_impl.cc:495] add/update cluster local_cluster during init
[2019-02-13 09:04:40.315][000007][info][upstream] [source/common/upstream/cluster_manager_impl.cc:495] add/update cluster kinesis_cluster during init
[2019-02-13 09:04:40.341][000007][info][upstream] [source/common/upstream/cluster_manager_impl.cc:136] cm init: all clusters initialized
[2019-02-13 09:04:40.341][000007][info][main] [source/server/server.cc:435] all clusters initialized. initializing init manager
[2019-02-13 09:04:40.355][000007][info][upstream] [source/server/lds_api.cc:80] lds: add/update listener 'ranking_ingress'
[2019-02-13 09:04:40.355][000007][info][upstream] [source/server/lds_api.cc:80] lds: add/update listener 'ranking_egress'
[2019-02-13 09:04:40.356][000007][info][upstream] [source/server/lds_api.cc:80] lds: add/update listener 'kinesis_egress'
[2019-02-13 09:04:40.356][000007][info][config] [source/server/listener_manager_impl.cc:961] all dependencies initialized. starting workers
[2019-02-13 09:04:46.297][000007][info][upstream] [source/common/upstream/cluster_manager_impl.cc:501] add/update cluster datascience starting warming
[...]

PR #5887, opened 2019-02-08
Pause CDS while there are clusters in the warming up state.

ivitjuk:
*Description*: Pause CDS API while there are clusters in the warming state.
*Risk Level*: High, as it changes the flow control of the CDS.
*Testing*:
 - Extended ClusterManagerImplTest.DynamicAddRemove to verify warming callback is correctly invoked.
 - Added CdsApiImplTest.CdsPauseOnWarming to verify CDS API pause/resume is correctly performed.

*Docs Changes*: N/A
*Release Notes*: N/A
*Fixes*: #5174

PR #23321, opened 2022-09-30
Fix the regression in pausing cds ack until cluster is in warming state

vikaschoudhary16:
Signed-off-by: Vikas Choudhary <choudharyvikas16@gmail.com>

Commit Message:
Functionality of pausing the cds ack was introduced in https://github.com/envoyproxy/envoy/pull/5887. During the refactoring done in https://github.com/envoyproxy/envoy/pull/7236, it got broken. 
There is also discussion about this regression on [this issue](https://github.com/envoyproxy/envoy/issues/5933#issuecomment-604011264) which got closed on stale state.

Additional Description:
>what cds-ack pausing scenario this PR is solving

Use case is like this. Let say [foo.com](http://foo.com/) hostname is being routed by route r1 to svc1 . Continuously new http requests are coming. Now it is desired to serve route r1  to svc-new .
t0:  Control plane starts updating the change by first pushing the new cluster for svc-new 
t1: while the envoy is still processing CDS for svc-new, control plane sends updated route. Updated route gets loaded which points to still warming cluster svc-new
t2: new http requests for [foo.com](http://foo.com/) gets "unknown cluster svc-new" 503
t3: svc-new warming gets completed, all good at datapath
After the fix, control-plane could wait for the CDS ack before pushing updated route. That would eliminate intermittent 503s.

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/upstream/cluster_manager_impl.cc
+++ b/source/common/upstream/cluster_manager_impl.cc
@@ -924,7 +924,7 @@ void ClusterManagerImpl::updateClusterCounts() {
   if (all_clusters_initialized && ads_mux_) {
     const auto type_url = Config::getTypeUrl<envoy::config::cluster::v3::Cluster>();
     const uint64_t previous_warming = cm_stats_.warming_clusters_.value();
-    if (previous_warming == 0 && !warming_clusters_.empty()) {
+    if (!warming_clusters_.empty()) {
       resume_cds_ = ads_mux_->pause(type_url);
     } else if (previous_warming > 0 && warming_clusters_.empty()) {
       ASSERT(resume_cds_ != nullptr);


source/common/upstream/cluster_manager_impl.cc
```diff
@@ -924,7 +924,7 @@ void ClusterManagerImpl::updateClusterCounts() {
   if (all_clusters_initialized && ads_mux_) {
     const auto type_url = Config::getTypeUrl<envoy::config::cluster::v3::Cluster>();
     const uint64_t previous_warming = cm_stats_.warming_clusters_.value();
-    if (previous_warming == 0 && !warming_clusters_.empty()) {
+    if (!warming_clusters_.empty()) {
```
vikaschoudhary16, line 927: `previous_warming` cant be 0 when `updateClusterCounts()` in invoked from `onClusterInit()` i.e after warming is done. This is because `loadCluster()` is also invoking `updateClusterCounts()` which increases it before the warming starts.
```
 const auto previous_cluster =
      loadCluster(cluster, new_hash, version_info, true, warming_clusters_);
  auto& cluster_entry = warming_clusters_.at(cluster_name);
  if (!all_clusters_initialized) {
    ENVOY_LOG(debug, "add/update cluster {} during init", cluster_name);
    init_helper_.addCluster(*cluster_entry);
  } else {
    ENVOY_LOG(debug, "add/update cluster {} starting warming", cluster_name);
    cluster_entry->cluster_->initialize([this, cluster_name] {
      ENVOY_LOG(debug, "warming cluster {} complete", cluster_name);
      auto state_changed_cluster_entry = warming_clusters_.find(cluster_name);
      onClusterInit(*state_changed_cluster_entry->second);
```

vikaschoudhary16:
cc @htuch

adisuissa:
Thanks for contributing!
I'm trying to understand what cds-ack pausing scenario this PR is solving. Can you add a test that this PR fixes (this is also good as a regression test)?
We may also want to runtime guard this.

source/common/upstream/cluster_manager_impl.cc
vikaschoudhary16: My bad. Pausing cds ack is working already. 
>previous_warming cant be 0 when updateClusterCounts() in invoked from onClusterInit() i.e after warming is done. This is because loadCluster() is also invoking updateClusterCounts() which increases it before the warming starts.

This is true but it does not stop/break pausing of cds ack. cds ack gets paused from the `loadCluster()` itself.

vikaschoudhary16:
>I'm trying to understand what cds-ack pausing scenario this PR is solving. Can you add a test that this PR fixes (this is also good as a regression test)?

Hey, while trying to add a test, I realized there is already a test for this `CdsPausedDuringWarming` and functionality is already working. When I looked closer, found out that my change is not needed and ack is paused from the loadCluster() itself. Hence closing the PR. Thanks for taking a look.

PR #3036, opened 2018-04-09
[WIP] Reconcile envoyproxy/data-plane-api and envoyproxy/envoy

htuch:
This is WiP for the process described in https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/envoy-dev/KcVHFH-zQwQ.

Still TODO:
- Update contributing docs and other user facing process information.
- Write script to push to data-plane-api on each commit affecting `//api`.
- More testing, CI fixing, debugging, etc. 

Risk Level: Medium (there might be unintentional breakage of dependent builds).
Testing: TBD

Signed-off-by: Harvey Tuch <htuch@google.com>

PR #23323, opened 2022-09-30
upstream: remove unused message UpstreamBindConfig

soulxu:

Commit Message: upstream: remove unused message UpstreamBindConfig
Additional Description:
Risk Level: low
Testing: n/a
Docs Changes: API doc
Release Notes: n/a

### DIFF
--- a/api/envoy/config/cluster/v3/cluster.proto
+++ b/api/envoy/config/cluster/v3/cluster.proto
@@ -1224,15 +1224,6 @@ message LoadBalancingPolicy {
   repeated Policy policies = 1;
 }
 
-// An extensible structure containing the address Envoy should bind to when
-// establishing upstream connections.
-message UpstreamBindConfig {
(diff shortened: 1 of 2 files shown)


adisuissa (commented):
This a breaking change, and generally is not allowed.
The message was first added into Envoy in #3036, and at the time there was no use of this field.
I think it is ok to remove, but would like to hear from others @envoyproxy/api-shepherds

htuch:
If there are literally zero other references, I think it's fine to remove.

soulxu:
@adisuissa @htuch thanks for the review!

I found this very old PR https://github.com/envoyproxy/data-plane-api/pull/150, it seems we add BindConfig message to instead of UpstreamBindConfig, since we want to support BindConfig both in bootstrap config and cluster config.

htuch (approved):

PR #23334, opened 2022-09-30
CacheFilter: handle encodeHeaders while lookups are outstanding

mkbehr:
Commit Message: Handle CacheFilter::encodeHeaders while lookups are outstanding.
Additional Description: CacheFilter stops filter chain iteration while a cache lookup is outstanding, but the filter chain manager may send a local reply anyway, e.g. in case of a stream idle timeout. Handle that case instead of assuming that the lookup has finished.
Risk Level: Low
Testing: Unit test
Docs Changes: N/A
Release Notes: N/A


test/extensions/filters/http/cache/cache_filter_test.cc
```diff
@@ -762,6 +762,104 @@ TEST_F(CacheFilterTest, FilterDeletedBeforePostedCallbackExecuted) {
...
+    dispatcher_->run(Event::Dispatcher::RunType::Block);
+
+    filter->onStreamComplete();
+    EXPECT_THAT(lookupStatus(), IsOkAndHolds(LookupStatus::RequestIncomplete));
+    filter->onDestroy();
+    filter.reset();
+  }
+}
+
+using CacheFilterDeathTest = CacheFilterTest;
```
jmarantz, line 810: what does this nickname do?

source/extensions/filters/http/cache/cache_filter.cc
```diff
@@ -102,6 +102,18 @@ Http::FilterHeadersStatus CacheFilter::encodeHeaders(Http::ResponseHeaderMap& he
     return Http::FilterHeadersStatus::Continue;
   }
 
+  if (lookup_result_ == nullptr) {
+    // Filter chain iteration is paused while a lookup is outstanding, but the filter chain manager
+    // can still generate a local reply. One case where this can happen is when a downstream idle
+    // timeout fires, which may mean that the HttpCache isn't correctly setting deadlines on its
+    // asynchronous operations or is otherwise getting stuck.
```
jmarantz, line 109: are there other cases? It looks like in the timeout case ENVOY_BUG fires, logs, increments a stat or crashes.

In what other case does this block fire?

test/extensions/filters/http/cache/cache_filter_test.cc
```diff
@@ -762,6 +762,104 @@ TEST_F(CacheFilterTest, FilterDeletedBeforePostedCallbackExecuted) {
...
+    // Make sure that the filter doesn't try to encode the cached response after processing the
+    // local reply.
+    EXPECT_CALL(decoder_callbacks_, continueDecoding).Times(0);
+    EXPECT_CALL(decoder_callbacks_, encodeHeaders_).Times(0);
+
+    // Run events on the dispatcher so that the lookup callback is invoked after the local reply.
+    dispatcher_->run(Event::Dispatcher::RunType::Block);
+
+    filter->onStreamComplete();
+    EXPECT_THAT(lookupStatus(), IsOkAndHolds(LookupStatus::RequestIncomplete));
```
jmarantz, line 804: in this test should we be looking for NotServingFromCache somehow?

jmarantz (commented):
lgtm basically; just questions about comments and the test.

test/extensions/filters/http/cache/cache_filter_test.cc
mkbehr: NotServingFromCache is an implementation detail and not visible to the test, but we exercise its behavior by confirming that the filter doesn't make any calls to the decoder callbacks when the lookup callback comes off the dispatcher.

test/extensions/filters/http/cache/cache_filter_test.cc
mkbehr: Tells googletest it's a death test for threading purposes. See https://google.github.io/googletest/advanced.html#death-test-naming

source/extensions/filters/http/cache/cache_filter.cc
jmarantz: This makes sense, if by async you mean same-thread, but after the cache filter has yielded control.

jmarantz (commented):
just one comment nit and i'm good to go.

test/extensions/filters/http/cache/cache_filter_test.cc
mkbehr: Done.

jmarantz (approved):
thanks!

PR #23338, opened 2022-10-01
Thrift: Defer route cache after apply applyDecoderFilters

JuniorHsu:

Commit Message:

Filters could change cluster_header's value, which affect the routing decision.
Therefore, we need to apply filter chain before the routing decision is made and cached.

Risk Level: low
Testing: unit

### DIFF
--- a/source/extensions/filters/network/thrift_proxy/conn_manager.cc
+++ b/source/extensions/filters/network/thrift_proxy/conn_manager.cc
@@ -841,6 +841,7 @@ FilterStatus ConnectionManager::ActiveRpc::messageBegin(MessageMetadataSharedPtr
 
   // Filters could change cluster_header's value, which affect the routing decision.
   // Therefore, we need to apply filter chain before the routing decision is made and cached.
+  // TODO(kuochunghsu): implement FilterCallbacks::clearRouteCache
   auto result = applyDecoderFilters(DecoderEvent::MessageBegin, metadata);
 
   const auto& route_ptr = route();
--- a/source/extensions/filters/network/thrift_proxy/conn_manager.cc
+++ b/source/extensions/filters/network/thrift_proxy/conn_manager.cc
@@ -839,6 +839,10 @@ FilterStatus ConnectionManager::ActiveRpc::messageBegin(MessageMetadataSharedPtr
     ASSERT(upgrade_handler_ != nullptr);
   }
 
+  // Filters could change cluster_header's value, which affect the routing decision.
+  // Therefore, we need to apply filter chain before the routing decision is made ad cached.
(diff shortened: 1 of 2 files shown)


source/extensions/filters/network/thrift_proxy/conn_manager.cc
```diff
@@ -839,6 +839,10 @@ FilterStatus ConnectionManager::ActiveRpc::messageBegin(MessageMetadataSharedPtr
     ASSERT(upgrade_handler_ != nullptr);
   }
 
+  // Filters could change cluster_header's value, which affect the routing decision.
+  // Therefore, we need to apply filter chain before the routing decision is made ad cached.
```
rgs1, line 843: nit: made and cached.

Hmm, but this would be different than what happens in HTTP land (not that we need to copy everything that we do there). In HTTP we lookup the route before filters run and if a filter does something that might alter the picked route it explicitly calls clearRouteCache() (e.g.: #3216). The reason for this is that route lookup might be expensive (think a lot of routes and a linear lookup). 

Should we try to stay on a somewhat similar path as what happens in the HTTP conn manager?

cc: @mattklein123 optionally for some thoughts on keeping similarities across different connection managers.

rgs1 (commented):
Thanks! Just one comment and also I think we might want to drop in a changelog entry for this change.

source/extensions/filters/network/thrift_proxy/conn_manager.cc
rgs1: cc: @fishcakez

JuniorHsu: deferring route decision won't make performance worse, actually `clearRouteCache` make the route decision probably called multiple time. 

One of the reason HTTP needs a routes before applying filter chain is support of `RouteSpecificFilterConfig` We need a pre-calculated routing decision for the filter configs.

Moreover, no matter we have `clearRouteCache` or not, the access log should log the *real* routing decision instead of the per-calculated one.  We want to get `route()` again (it may be a cached one) after filter chain and right before setting the metadata. 

I agree that we should have `clearRouteCache` in the next PR since we allow filter to access and cache the route. 
https://github.com/envoyproxy/envoy/blob/5f57678636d52b6e4fb3aca0ec256fa71a99f3f3/source/extensions/filters/network/thrift_proxy/filters/filter.h#L54
Hence, this diff is just a gate stopper, which doesn't prevent corner cases unfortunately.

rgs1 (approved):
Ok, if the computed route is not even available to filters as of now this is probably fine and shouldn't break any existing filter that's out there.

Over to @zuercher 

P.S.: a changelog entry would still be nice, or a TODO comment that says what you are getting at next.

zuercher (approved):

PR #23335, opened 2022-09-30
Revert "delta-xds: perform resource state tracking after resource ing…

mattklein123:
…estion (#23214)"

This reverts commit 6e6a3a17e4c1890b20b8832280a81385f3eb8573.

Coverage is failing on main.

PR #23343, opened 2022-10-03
delta-xds: resource state tracking after resource ingestion

### LINKED ISSUE #20790: [delta-XDS] delta-XDS state tracking is broken when there is an exception thrown when watchmap loading the config.

adisuissa:
Commit Message: delta-xds: resource state tracking after resource ingestion
Additional Description:
Fixes delta-xDS protocol resource state tracking.
Prior to this PR, the implementation updated the state of a resource (tracked version, TTL), even if the fetched config was NACKed. This PR changes it to only update the tracked state after a resource was successfully ingested, and an ACK will be sent.
Note that this PR impacts the TTL of a resource, as now TTL timer starts after the resource was ingested rather than before.
This PR is essentially #23214 (which was reverted in #23335) with more coverage.

Risk Level: Medium for delta-xDS usage.
Testing: Added unit tests.
Docs Changes: N/A.
Release Notes: Added release note.
Runtime guard: Can be reverted by setting envoy.reloadable_features.delta_xds_subscription_state_tracking_fix to false.
Fixes #20790


htuch (approved):
Interesting, it seems like the earlier code you copy+pasted over the ingest line was missing coverage as well. LGTM.

adisuissa:
> Interesting, it seems like the earlier code you copy+pasted over the ingest line was missing coverage as well. LGTM.

Indeed. It passed pre-submit, but by the time the PR was merged post-submit failed.

PR #23344, opened 2022-10-03
upstream: logging shadow requests

alyssawilk:
Changing shadow requests to use the router's filter config.
This results in the shadow requests behaving more like the original request in a number of ways, upstream access logging among them.

Risk Level: medium
Docs Changes: n/a
Release Notes: yes
[Optional Runtime guard:] yes


alyssawilk:
cc @paul-r-gall

source/extensions/filters/http/router/config.cc
```diff
@@ -19,8 +19,11 @@ Http::FilterFactoryCb RouterFilterConfig::createFilterFactoryFromProtoTyped(
       prefix.statName(), context,
       std::make_unique<Router::ShadowWriterImpl>(context.clusterManager()), proto_config));
 
-  return [filter_config](Http::FilterChainFactoryCallbacks& callbacks) -> void {
-    callbacks.addStreamDecoderFilter(std::make_shared<Router::ProdFilter>(*filter_config));
+  return [filter_config, &stat_prefix,
+          &context](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+    Stats::StatNameManagedStorage prefix(stat_prefix, context.scope().symbolTable());
```
paul-r-gall, line 24: prefix here is unused?

changelogs/current.yaml
```diff
@@ -73,6 +73,9 @@ minor_behavior_changes:
 - area: health_check
   change: |
     support custom health check address via :ref:`health_check_config <envoy_v3_api_msg_config.endpoint.v3.endpoint.healthcheckconfig>`.
+- area: http
+  change: |
+     changed shadow requests to more closely behave like the requests they are shadowing. This includes matching the upstream logging for the original request, dynamic stats, suppressing Envoy headers, respecting expected request timeout, suppressing grpc request failure code stats and strict header checks. This behaviorial change can be temporarily reverted by setting runtime guard ``envoy.reloadable_features.closer_shadow_behavior`` to false.
```
snowp, line 78: Wrap this line maybe? Not sure if we care about formatting much in this file, but it is kinda long

snowp (approved):
Nice, this was much easier than I thought it would have been

changelogs/current.yaml
alyssawilk: I think the style of this file is generally single wrapped lines

PR #23366, opened 2022-10-04
gcp_authn: Modify the token time constraint check

tyxia:
`VerifyTimeConstraint()` interface is correct for the token consumer. However, as the token producer here, we should instead include the clock skew as the part of the `now` time up front to account for the clock skew on the consumer side where the token will be consumed. Also, in this way, we can avoid changing the third-party library interface `VerifyTimeConstraint()` (which might requires longer turnaround time)

### DIFF
--- a/source/extensions/filters/http/gcp_authn/token_cache.h
+++ b/source/extensions/filters/http/gcp_authn/token_cache.h
@@ -32,6 +32,8 @@ public:
   void insert(const std::string& key, std::unique_ptr<TokenType>&& token);
   TokenType* validateTokenAndReturn(const std::string& key, TokenType* const found_token);
 
+  uint64_t capacity() { return lru_cache_.maxSize(); }
+
   ~TokenCacheImpl() {
     // Remove all entries from the cache.
     lru_cache_.clear();
--- a/source/extensions/filters/http/gcp_authn/token_cache.h
+++ b/source/extensions/filters/http/gcp_authn/token_cache.h
@@ -87,8 +87,12 @@ TokenType* TokenCacheImpl<TokenType>::validateTokenAndReturn(const std::string& 
   if constexpr (std::is_same<TokenType, JwtToken>::value) {
     ASSERT(found_token != nullptr);
     // Verify the validness of the token by checking its expiration time field.
-    if (found_token->verifyTimeConstraint(DateUtil::nowToSeconds(time_source_)) ==
-        ::google::jwt_verify::Status::JwtExpired) {
+    // Note: verifyTimeConstraint() interface is correct to be used by token consumer. However, for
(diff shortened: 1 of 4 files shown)


fiadliel:
One way of looking at this is that people/services expect a provided token to not expire _immediately_, so some positive time addition is better than none. And as these tokens generally have a lifetime of an hour, there's a big buffer to work with; no need to use a token right up to the last second.

What that extra amount should be is a judgment call; `kClockSkewInSecond` is probably as good as any. I think this would solve the issue in the Google environment (where, also, most systems are running NTP, so there's very little actual skew).

tyxia:
> One way of looking at this is that people/services expect a provided token to not expire _immediately_, so some positive time addition is better than none. And as these tokens generally have a lifetime of an hour, there's a big buffer to work with; no need to use a token right up to the last second.
> 
> What that extra amount should be is a judgment call; `kClockSkewInSecond` is probably as good as any. I think this would solve the issue in the Google environment (where, also, most systems are running NTP, so there's very little actual skew).
> 
> (so, TLDR, LGTM)

Thanks for the review! Yes, as I commented in the code the clock skew here actually can be configurable by client in the future if it is needed.

tyxia:
/assign @yanavlasov 

Could you take a look at this PR when you have free time? Thanks!

yanavlasov (approved):

yanavlasov (changes requested):
Just realized that we do not have unit tests for this code. The return statement on line 96 (token is valid) is not covered by tests.

Can you add unit tests for the TokenCacheImpl class please? I think it would be good to have tests that cover the skew.

/wait

yanavlasov:
> > Just realized that we do not have unit tests for this code. The return statement on line 96 (token is valid) is not covered by tests.
> > Can you add unit tests for the TokenCacheImpl class please? I think it would be good to have tests that cover the skew.
> > /wait
> 
> Good point! I have verified the token cache locally before but because the token expiration is timing related we can not have deterministic behavior in test over time (i.e., token will be expired eventually). I will figure out the way of testing it (e.g., generating the token on each test run by leveraging time source etc). Per offline discussion, currently I left this as TODO so that we can check in this change unblock the customer side. The token related functionalities are verified by the test in jwt library itself.

You have control over system time with the TimeSource class. There are tests that do what you need to do in the test/extensions/filters/http/jwt_authn/jwt_cache_test.cc. You can use one of the token strings from the test and manipulate the tests system time object to set the time to what you need.

/wait-any

KBaichoo:
ping @yanavlasov  PTAL

phlax:
cc @yanavlasov i think this is waiting on your review

yanavlasov (approved):

PR #23367, opened 2022-10-04
Add API field in Router for Upstream Filters.

paul-r-gall:

Commit Message: Add API field in Router Config for upstream filters.
Additional Description: Upstream filters specified in the router will apply if there are no filters specified in the cluster.
Risk Level: Low.
Testing: Unit and Integration testing.


paul-r-gall:
cc @mattklein123 @htuch @alyssawilk -- for API discussion.

api/envoy/extensions/filters/http/router/v3/router.proto
```diff
@@ -89,4 +90,20 @@ message Router {
...
+  //
+  // These filters will be applied for all requests that pass through the router.
+  // They will also be applied to shadowed requests.
+  // Upstream filters cannot change route or cluster.
+  //
+  // If using upstream filters, please be aware that local errors sent by
+  // upstream filters will not trigger retries, and local errors sent by
+  // upstream filters will count as a final response if hedging is configured.
+  // [#extension-category: envoy.filters.http.upstream]
+  repeated network.http_connection_manager.v3.HttpFilter upstream_http_filters = 8;
```
htuch, line 108: How do these interact with per-cluster upstream HTTP filters? Upstream HTTP filters will vary depending on cluster, even if we have a "default" for a given HCM filter chain.

yanavlasov: Normally more specific configuration overwrites less specific configuration.

Also if we are going this way, should we make this part of the usual configuration hierarchy, HCM -> VH -> Route and then Cluster ?

htuch: Hard override is fine if it matches @paul-r-gall use case, since we have a forward compatible path for other composition. Can you document this @paul-r-gall?

api/envoy/extensions/filters/http/router/v3/router.proto
```diff
@@ -89,4 +90,20 @@ message Router {
...
+  //   This warning will be removed as support moves beyond alpha.
+  //
+  // Optional HTTP filters for the upstream filter chain.
+  //
+  // These filters will be applied for all requests that pass through the router.
+  // They will also be applied to shadowed requests.
+  // Upstream filters cannot change route or cluster.
+  //
+  // If using upstream filters, please be aware that local errors sent by
+  // upstream filters will not trigger retries, and local errors sent by
```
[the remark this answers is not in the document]
paul-r-gall: That's a question for Alyssa -- this is a copy of the documentation comment around upstream filters in upstream http protocol options.

api/envoy/extensions/filters/http/router/v3/router.proto
paul-r-gall: Hard override to start off is fine with me.

htuch (commented):
/lgtm api

htuch:
LGTM, will let others ack before merge.

wbpcode:
friendly ping @alyssawilk

paul-r-gall:
@alyssawilk added implementation.

alyssawilk:
sweet.  mind taking a look at CI?

api/envoy/extensions/filters/http/router/v3/router.proto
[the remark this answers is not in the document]
alyssawilk: because we retry when the peer sends an error.  the upstream filter chain sendLocalReply kicks off the downstream sendLocalReply chain which bypasses the "should we retry" logic.

source/common/router/upstream_request.cc
```diff
@@ -125,7 +125,13 @@ UpstreamRequest::UpstreamRequest(RouterFilterInterface& parent,
       parent_.callbacks()->connection(), parent_.callbacks()->streamId(),
       parent_.callbacks()->account(), true, parent_.callbacks()->decoderBufferLimit(),
       *parent_.cluster(), *this);
-  parent_.cluster()->createFilterChain(*filter_manager_);
+  if (parent_.cluster()->filterChainLength() <= 1 && parent_.config().filterChainLength() > 0) {
```
alyssawilk, line 128: can you check out the comments I made on the internal version?  I don't see them applied here
/wait

paul-r-gall: You're talking about having createFilterChain return bool, right? The thing is that clusters will always have the codec filter attached to them, and so `parent_.cluster()->createFilterChain()` would always return `true`. I guess something like 
```
if (parent_.cluster()->filterChainLength() > 1 // there is a nontrivial cluster filter chain
    || !parent_.config().createFilterChain(*filter_manager_) /* there is no router filter chain */ ) {
  parent_.cluster()->createFilterChain(*filter_manager_);
}
```
would do the same thing as what I've got above, but I don't know that it's clearer.

Another possibility is to just totally split the codec filter off of the filter chain; so the above would become:
```
parent_.cluster()->createFilterChain(...) || parent_.config().createFilterChain(...);
parent_.cluster()->appendCodecFilter(...);
```
The first line will only run the router filter chain creation if cluster filter chain creation returns false.

alyssawilk: oh yeah sorry I had it inverted and thought router trumped cluster so we could do something like we do for the upgrade filter chain but we can't because as you say the "nothing configured" case is empty.

OK new idea, which I'm writing in a hurry so poke me this afternoon if it doesn't make sense.

ow about adding a boolean to createFilterChain so we only create it if there's more than one entry, i.e if that filter chain is actually configured?  then in the long run we could

bool created = cluster->createUpgradeFilterChain(filter_manager)
if (!created) {
  created = cluster->createFilterChain(filter_manager_, true /*only_create_if_configured*/)
}
if (!created) { 
  created = router->createUpgradeFilterChain()
}
createFilterChain(filter_manager_, false /* only_create_if_configured */)

which I think is cleaner than a bunch of size checks?  alternately change your size check to isFilterChainConfigured but then we have to check existence of regular filter chains or best effort try-create upgrade filter chains, which I don't love.

paul-r-gall: My other concern with this is that in the future we may add Filter Config to the UpstreamCodecFilter, in which case clusters could actually configure UpstreamCodecFilter. There wouldn't be a good way to distinguish between an UpstreamCodecFilter that has the default values, and one that has values assigned by the cluster.

alyssawilk: I actually see that as a feature and not a bug.  now that I think of it I really don't like length.
If you have a route config filterA filterB codec, and you want an override of "don't run any filters" how would you do it?   the override filter length is 1 so we'd skip?  I think we need to go by the user configuration proto - if one was present at all or if we used the default config, rather than the length of the filter chain, so again ideally the 'only if configured' bool but alternately a filterChainConfigured() method returning bool

paul-r-gall: I think I prefer the `isCustomFilterChainConfigured()` method returning `bool`, personally.

snowp (changes requested):
Seems reasonable to me, a few questions/comments

source/common/router/router.h
```diff
@@ -235,7 +239,39 @@ class FilterConfig {
...
+      Http::FilterChainHelper<Server::Configuration::UpstreamHttpFactoryContext,
+                              Server::Configuration::UpstreamHttpFilterConfigFactory>
+          helper(*filter_config_provider_manager, server_factory_ctx, *upstream_ctx_, prefix);
+      helper.processFilters(upstream_http_filters, "router upstream http", "router upstream http",
+                            upstream_http_filter_factories_);
+    }
+  }
+
+  bool createFilterChain(Http::FilterChainManager& manager, bool) const override {
+    if (upstream_http_filter_factories_.empty()) {
```
alyssawilk, line 261: why are we ignoring only_create_if_configured?

source/common/router/upstream_request.cc
```diff
@@ -125,10 +125,17 @@ UpstreamRequest::UpstreamRequest(RouterFilterInterface& parent,
...
       *parent_.cluster(), *this);
-  parent_.cluster()->createFilterChain(*filter_manager_);
+  bool created = parent_.cluster()->createFilterChain(*filter_manager_,
+                                                      /*only_create_if_configured=*/true);
+  if (!created) {
+    created = parent_.config().createFilterChain(*filter_manager_, false);
+  }
+  if (!created) {
+    created = parent_.cluster()->createFilterChain(*filter_manager_, false);
+  }
```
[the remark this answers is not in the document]
alyssawilk: yeah I'm confused at this. 
I thought this would create cluster-if-configured else create router no matter what.  
let's do e2e tests for the 4 config combinations

source/common/router/router.h
```diff
@@ -235,7 +239,39 @@ class FilterConfig {
     for (const auto& upstream_log : config.upstream_log()) {
       upstream_logs_.push_back(AccessLog::AccessLogFactory::fromProto(upstream_log, context));
     }
+    if (config.upstream_http_filters_size() > 0) {
+      auto& server_factory_ctx = context.getServerFactoryContext();
+      const Http::FilterChainUtility::FiltersList& upstream_http_filters =
```
alyssawilk, line 244: the answer is probably no but anything else we can do re code sharing between this and preexisting code?

source/common/upstream/upstream_impl.h
```diff
@@ -939,6 +926,7 @@ class ClusterInfoImpl : public ClusterInfo,
   const std::unique_ptr<Server::Configuration::CommonFactoryContext> factory_context_;
   std::vector<Network::FilterFactoryCb> filter_factories_;
   Http::FilterChainUtility::FilterFactoriesList http_filter_factories_;
+  bool has_custom_http_filters_{false};
```
alyssawilk, line 929: I don't think it needs to be custom, just configued yeah?
I think it's also worth a comment why we have to latch the bool, e.g. the difference between empty (i.e. implicit codec filter) and present (but maybe not custom configured) codec filter)

alyssawilk (commented):
sorry for lag - here's a few more thoughts
/wait

source/common/upstream/upstream_impl.h
paul-r-gall: Maybe this is semantics, but all clusters will have the codec filter configured in http_filter_factories_, but not all clusters will have custom filters specified in the proto config.

alyssawilk: important semantics!
I still think if we have an HCM filter chain [A B codec] and cluster A with no cluster config and cluster B with cluster config codec filter, requests to cluster A should have A B codec, and requests to cluster B should have codec.
So it's not about if there's more than one filter or a custom filter, it's if there's any actual configuration, vs no configuratoin (which is implicitly codec filter)

paul-r-gall: SGTM. will change.

source/common/router/upstream_request.cc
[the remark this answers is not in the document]
paul-r-gall: Router doesn't have codec filter automatically applied; the cluster does. We could move that logic away from the cluster onto the router if you think that makes more sense?

test/common/router/router_upstream_filter_test.cc
```diff
@@ -0,0 +1,158 @@
...
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks_;
+  std::shared_ptr<FilterConfig> config_;
+  std::shared_ptr<TestFilter> router_;
+  std::shared_ptr<NiceMock<Upstream::MockClusterInfo>> cluster_info_;
+  NiceMock<StreamInfo::MockStreamInfo> stream_info_;
+};
+
+TEST_F(RouterUpstreamFilterTest, UpstreamFilter) {
+  TestScopedRuntime tsr;
+  tsr.mergeValues({{"envoy.reloadable_features.allow_upstream-filters", "true"},
```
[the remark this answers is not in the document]
paul-r-gall: ah; allow_upstream_filters is on by default now.

source/common/router/router.h
paul-r-gall: The router default is an empty filter list -- there's nothing to do if the chain is not configured.

source/common/router/router.h
paul-r-gall: Not without some moderate refactoring. I think there's enough that's sufficiently different that we don't gain much by sharing code.

test/common/router/router_upstream_filter_test.cc
```diff
@@ -0,0 +1,158 @@
...
+                   {"envoy.reloadable_features.no_extension_lookup_by_name", "false"}});
+  HttpFilter add_header_filter;
+  add_header_filter.set_name("add-header-filter");
+  HttpFilter codec_filter;
+  codec_filter.set_name("envoy.filters.http.upstream_codec");
+  init({add_header_filter, codec_filter});
+  auto headers = run({});
```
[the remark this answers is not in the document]
paul-r-gall: Fixed.

source/common/router/router.h
alyssawilk: nothing to do today.  how about a comment and an assert in case that changes?

source/common/upstream/upstream_impl.h
```diff
@@ -856,8 +838,13 @@ class ClusterInfoImpl : public ClusterInfo,
   upstreamHttpProtocol(absl::optional<Http::Protocol> downstream_protocol) const override;
 
   // Http::FilterChainFactory
-  void createFilterChain(Http::FilterChainManager& manager) const override {
-    Http::FilterChainUtility::createFilterChainForFactories(manager, http_filter_factories_);
+  bool createFilterChain(Http::FilterChainManager& manager,
+                         bool only_create_if_configured) const override {
+    if (!only_create_if_configured || has_configured_http_filters_) {
```
alyssawilk, line 843: optional: might be easier to read if we invert and early return false.

alyssawilk (commented):
cool almost there!
/wait

alyssawilk (approved):
over to @snowp

snowp (approved):
LGTM, thanks!

htuch (approved):
/lgtm api

PR #23374, opened 2022-10-05
contrib: make qat and cryptomb unit tests only run on specific platform

VillePihlava:
Commit Message: Make bazel only run QAT and cryptomb private key provider tests on `linux_x86_64` platform.
Additional Description:

The QAT and cryptomb private key provider features are hardware dependent, but the tests use mock libraries which can run on other platforms. This pull request changes the tests to only run on the `linux_x86_64` platform.

Risk Level: Low
Testing: Manual testing
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: Change unit tests to only run on `linux_x86_64`

### DIFF
--- a/contrib/cryptomb/private_key_providers/test/BUILD
+++ b/contrib/cryptomb/private_key_providers/test/BUILD
@@ -4,6 +4,10 @@ load(
     "envoy_cc_test_library",
     "envoy_contrib_package",
 )
+load(
+    "//contrib:all_contrib_extensions.bzl",
+    "envoy_contrib_linux_x86_64_constraints",
+)
 
 licenses(["notice"])  # Apache 2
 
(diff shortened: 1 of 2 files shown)


VillePihlava:
@ipuustin @phlax I believe this should remove the tests from being run on arm as was talked about in https://github.com/envoyproxy/envoy/pull/22717. I was able to locally test that the tests can be excluded on a platform by disabling them on a `linux_x86_64` machine.

I was able to verify that the unit tests do run on `linux_x86_64` by following the link in the ci logs, for example on this line:
https://dev.azure.com/cncf/envoy/_build/results?buildId=118027&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec&l=607 

I couldn't find a similar link in the arm build logs, but it does say that some tests are skipped which I couldn't find in the other logs.
Is there a way to view the arm build/test logs in a similar way?

phlax:
>  Is there a way to view the arm build/test logs in a similar way?

unfortunately we dont have resultstore set up for arm - i guess because it doesnt use RBE - but im wondering if it could still use resultstore - cc @lizan

ipuustin (approved):
Looks good to me. I think it makes sense to run the tests with the same constraints that affect the "real" library dependencies.

lizan:
> > Is there a way to view the arm build/test logs in a similar way?
> 
> unfortunately we dont have resultstore set up for arm - i guess because it doesnt use RBE - but im wondering if it could still use resultstore - cc @lizan

It could use result store but not available right now, I just needed to make local cache transport secured

keith (approved):

PR #23388, opened 2022-10-06
refactor: code sytle consistency

sekirocc:
`ManagedResourceImpl` reuse parent api when possible, just like inc/decBy

Signed-off-by: Sekiro CC <sekirocc@gmail.com>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/upstream/resource_manager_impl.h
+++ b/source/common/upstream/resource_manager_impl.h
@@ -25,7 +25,6 @@ struct ManagedResourceImpl : public BasicResourceLimitImpl {
   }
 
   // BasicResourceLimitImpl
-  bool canCreate() override { return BasicResourceLimitImpl::canCreate(); }
   void inc() override {
     BasicResourceLimitImpl::inc();
     updateRemaining();
--- a/source/common/upstream/resource_manager_impl.h
+++ b/source/common/upstream/resource_manager_impl.h
@@ -25,7 +25,7 @@ struct ManagedResourceImpl : public BasicResourceLimitImpl {
   }
 
   // BasicResourceLimitImpl
-  bool canCreate() override { return current_ < max(); }
+  bool canCreate() override { return BasicResourceLimitImpl::canCreate(); }
   void inc() override {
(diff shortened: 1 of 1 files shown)


source/common/upstream/resource_manager_impl.h
```diff
@@ -25,7 +25,7 @@ struct ManagedResourceImpl : public BasicResourceLimitImpl {
   }
 
   // BasicResourceLimitImpl
-  bool canCreate() override { return current_ < max(); }
+  bool canCreate() override { return BasicResourceLimitImpl::canCreate(); }
```
giantcroc, line 28: Hi, if this could work, why not use `BasicResourceLimitImpl::canCreate` directly?

sekirocc: HI, frankly i'm not sure why not use the parent api directly, maybe because of some design decisions? I just found the method (kind of) repeating its parent and made a simple suggestion. :)

KBaichoo: I think we should just remove this since it's the same implementation as the parent.

KBaichoo (commented):
Thanks for cleaning this up

KBaichoo (approved):
thanks

KBaichoo:
@sekirocc CI seems to be stuck, can you push an empty commit to retry it?

KBaichoo:
can you try merging main? perhaps that might resolve the issue.

If not I'll try to get someone who can handle when the CI hangs like this. Sorry for the inconvenience.

sekirocc:
hi @KBaichoo it seems the CI won't succeed anymore :( . Feel free to close the merge if you think it is appropriate

KBaichoo:
Thanks for merging @alyssawilk ! and thanks for the patience @sekirocc , sorry about CI.

PR #23391, opened 2022-10-06
router: evaluate all headers before modifying

### LINKED ISSUE #23378: Headers referencing other headers in route actions could create high CPU and memory utilization
*Title*: *Headers referencing other headers in route actions could create high CPU and memory utilization*

_Approved by the security group to be a public issue, because it requires an appropriate config._

*Description*:
PR #21932 fixed an issue where a header that references another header (possibly same header) can be added in a route action.
With a wrong configuration this could create high CPU and memory utilization.

The [evaluateHeaders](https://github.com/envoyproxy/envoy/blob/28e8c5c16d1c7a6d783f6425deb59b65bc545eef/source/common/router/header_parser.cc#L335) method iterates over the "headers-to-add" list, and calls the formatter on each of them. While doing so, it reads the current header value, and updates it to be the same value multiplied. This is done a few times, which could causes large memory consumption, and high CPU utilization.

Example:
1) adding a header with some value:
```
headers_to_add {
  header {
    key: "a"
    value: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx="
  }
  append_action: OVERWRITE_IF_EXISTS_OR_ADD
}
```

2) referencing this value (this entry can repeat itself a few times):
```
headers_to_add {
  header {
    key: "a"
    value: "%START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)%"
  }
  append_action: OVERWRITE_IF_EXISTS_OR_ADD
}

headers_to_add {
  header {
    key: "a"
    value: "%START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)% %START_TIME( )%%REQ(a)%"
  }
  append_action: OVERWRITE_IF_EXISTS_OR_ADD
}

...
```

Fuzzer reference: [51803](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51803)

cpakulski:
Commit Message:
evaluate all headers before modifying

Additional Description:
As reported in #23378, if there are multiple actions for the same header, the output of the previous formatter was used as input for the next formatter. For example, it a request header has a value "a" : "xxxxxx" and there are 2 formatters which modify the same header "a" and at the same time the formatter refers to the header "a", the changes were stack on top of each other. For example if the first formatter was "%REQ(a)% %REQ(a)%" and the second formatter was "%REQ(a)% %REQ(a)% %REQ(a)%", the proper output should be "xxxxxx xxxxxx" and "xxxxxx xxxxxx xxxxxx". However, the code evaluated the first formatter and overwrote the value of header "a" to be "xxxxxx xxxxxx". When the second formatter was executed, it used not the original value "xxxxxx", but the output of the previous formatter "xxxxxx xxxxxx". Eventually the second formatter produced the "xxxxxx xxxxxxxxxxxx xxxxxxxxxxxx xxxxxx".
The change is basically to run all formatters using original header values and only overwrite/modify the header values when all formatters has been executed.

Risk Level: Low.
Testing: Added unit tests.
Docs Changes: No.
Release Notes: No
Fixes #23378

### DIFF
--- a/source/common/router/header_parser.cc
+++ b/source/common/router/header_parser.cc
@@ -342,6 +342,13 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
     headers.remove(header);
   }
 
+  // Create local copy of headers to add and replace. This is required
+  // to execute all formatters using the original received headers.
+  // Only after all the formatters produced the new values of the headers, the headers are set.
+  std::vector<std::pair<const Http::LowerCaseString&, std::string>> headers_to_add;
+  absl::flat_hash_map<absl::string_view, std::pair<const Http::LowerCaseString&, std::string>>
+      headers_to_overwrite;
+
   for (const auto& [key, entry] : headers_to_add_) {
     const std::string value =
         stream_info != nullptr
@@ -351,18 +358,33 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
       switch (entry.append_action_) {
         PANIC_ON_PROTO_ENUM_SENTINEL_VALUES;
       case HeaderValueOption::APPEND_IF_EXISTS_OR_ADD:
-        headers.addReferenceKey(key, value);
+        headers_to_add.emplace_back(key, value);
         break;
       case HeaderValueOption::ADD_IF_ABSENT:
         if (auto header_entry = headers.get(key); header_entry.empty()) {
-          headers.addReferenceKey(key, value);
+          headers_to_add.emplace_back(key, value);
         }
         break;
       case HeaderValueOption::OVERWRITE_IF_EXISTS_OR_ADD:
-        headers.setReferenceKey(key, value);
+        // TODO(cpakulski): Hash map is used to track multiple OVERWRITE actions
+        // and store only the last one. It would be more efficient to do it
+        // when config is processed. All OVERWRITE actions for the same header could be dropped
+        // except the last one.
+        headers_to_overwrite.erase(key);
+        headers_to_overwrite.try_emplace(key, key, value);
         break;
       }
     }
+  }
+
+  // First overwrite all headers which need to be overwritten.
+  for (const auto& [key, value] : headers_to_overwrite) {
+    headers.setReferenceKey(value.first, value.second);
+  }
+
+  // Now add headers which should be added.
+  for (const auto& header : headers_to_add) {
+    headers.addReferenceKey(header.first, header.second);
   }
 }
 
--- a/test/common/router/header_formatter_test.cc
+++ b/test/common/router/header_formatter_test.cc
@@ -38,6 +38,7 @@ using ::testing::Pair;
 using ::testing::Return;
 using ::testing::ReturnPointee;
 using ::testing::ReturnRef;
+using ::testing::UnorderedElementsAre;
 
 static envoy::config::route::v3::Route parseRouteFromV3Yaml(const std::string& yaml) {
   envoy::config::route::v3::Route route;
@@ -2013,8 +2014,6 @@ response_headers_to_remove: ["x-foo-header"]
   resp_header_parser->evaluateHeaders(header_map, stream_info);
   EXPECT_EQ("bar", header_map.get_("x-foo-header"));
 }
-
-using ::testing::UnorderedElementsAre;
 
 TEST(HeaderParserTest, GetHeaderTransformsWithFormatting) {
   const std::string yaml = R"EOF(
--- a/test/common/router/header_formatter_test.cc
+++ b/test/common/router/header_formatter_test.cc
@@ -1664,6 +1664,91 @@ request_headers_to_add:
   EXPECT_EQ(1, counts["x-request-start"]);
 }
 
+TEST(HeaderParserTest, EvaluateMultipleOverwriteHeadersWithTheSameKey) {
+  // Each next OVERWRITE_IF_EXISTS_OR_ADD should eclipse the previous one
+  // and the last one will determine the actual header value.
+  const std::string yaml = R"EOF(
+match: { prefix: "/new_endpoint" }
+route:
+  cluster: "www2"
+  prefix_rewrite: "/api/new_endpoint"
+request_headers_to_add:
+  - header:
+      key: "a"
+      value: "abcdefghijklmnopqrstwxyz"
+    append_action: OVERWRITE_IF_EXISTS_OR_ADD
+  - header:
+      key: "a"
+      value: "second-%REQ(a)%%REQ(a)%"
+    append_action: OVERWRITE_IF_EXISTS_OR_ADD
+  - header:
+      key: "a"
+      value: "third-%REQ(a)%%REQ(a)%%REQ(a)%"
+    append_action: OVERWRITE_IF_EXISTS_OR_ADD
+)EOF";
+
+  HeaderParserPtr req_header_parser =
+      HeaderParser::configure(parseRouteFromV3Yaml(yaml).request_headers_to_add());
+  Http::TestRequestHeaderMapImpl header_map{{":method", "POST"}, {"a", "value-a"}};
+  NiceMock<Envoy::StreamInfo::MockStreamInfo> stream_info;
+  EXPECT_CALL(stream_info, getRequestHeaders).WillRepeatedly(Return(&header_map));
+  req_header_parser->evaluateHeaders(header_map, stream_info);
+  std::vector<absl::string_view> header_values;
+  header_map.iterate([&header_values](const Http::HeaderEntry& header) -> Http::HeaderMap::Iterate {
+    if (header.key() == "a") {
+      header_values.push_back(header.value().getStringView());
+    }
+    return Http::HeaderMap::Iterate::Continue;
+  });
+
+  EXPECT_EQ(1, header_values.size());
+  EXPECT_EQ("third-value-avalue-avalue-a", header_values[0]);
+}
+
+TEST(HeaderParserTest, EvaluateMultipleAppendAndOverwriteHeadersWithTheSameKey) {
+  // Append header should be added and two headers with OVERWRITE action
+  // should overwrite the original value.
+  // The first header with OVERWRITE action is basically no-op because it will
+  // be overwritten by the second OVERWRITE action.
+  const std::string yaml = R"EOF(
+match: { prefix: "/new_endpoint" }
+route:
+  cluster: "www2"
+  prefix_rewrite: "/api/new_endpoint"
+request_headers_to_add:
+  - header:
+      key: "a"
+      value: "second-%REQ(a)%%REQ(a)%"
+    append_action: OVERWRITE_IF_EXISTS_OR_ADD
+  - header:
+      key: "a"
+      value: "abcdefghijklmnopqrstwxyz"
+    append_action: APPEND_IF_EXISTS_OR_ADD
+  - header:
+      key: "a"
+      value: "third-%REQ(a)%%REQ(a)%%REQ(a)%"
+    append_action: OVERWRITE_IF_EXISTS_OR_ADD
+)EOF";
+
+  HeaderParserPtr req_header_parser =
+      HeaderParser::configure(parseRouteFromV3Yaml(yaml).request_headers_to_add());
+  Http::TestRequestHeaderMapImpl header_map{{":method", "POST"}, {"a", "value-a"}};
+  NiceMock<Envoy::StreamInfo::MockStreamInfo> stream_info;
+  EXPECT_CALL(stream_info, getRequestHeaders).WillRepeatedly(Return(&header_map));
+  req_header_parser->evaluateHeaders(header_map, stream_info);
+  std::vector<absl::string_view> header_values;
+  header_map.iterate([&header_values](const Http::HeaderEntry& header) -> Http::HeaderMap::Iterate {
+    if (header.key() == "a") {
+      header_values.push_back(header.value().getStringView());
+    }
+    return Http::HeaderMap::Iterate::Continue;
+  });
+
+  EXPECT_EQ(2, header_values.size());
+  EXPECT_THAT(header_values, testing::UnorderedElementsAre("abcdefghijklmnopqrstwxyz",
+                                                           "third-value-avalue-avalue-a"));
+}
+
 TEST(HeaderParserTest, EvaluateResponseHeaders) {
   const std::string yaml = R"EOF(
 match: { prefix: "/new_endpoint" }
@@ -1929,6 +2014,8 @@ response_headers_to_remove: ["x-foo-header"]
   EXPECT_EQ("bar", header_map.get_("x-foo-header"));
 }
 
+using ::testing::UnorderedElementsAre;
+
 TEST(HeaderParserTest, GetHeaderTransformsWithFormatting) {
   const std::string yaml = R"EOF(
 match: { prefix: "/new_endpoint" }
@@ -1967,9 +2054,10 @@ response_headers_to_remove: ["x-baz-header"]
   auto transforms = resp_header_parser->getHeaderTransforms(stream_info);
   EXPECT_THAT(transforms.headers_to_append_or_add,
               ElementsAre(Pair(Http::LowerCaseString("x-foo-header"), "foo")));
-  EXPECT_THAT(transforms.headers_to_overwrite_or_add,
-              ElementsAre(Pair(Http::LowerCaseString("x-bar-header"), "bar"),
-                          Pair(Http::LowerCaseString("x-per-request-header"), "test_value")));
+  EXPECT_THAT(
+      transforms.headers_to_overwrite_or_add,
+      UnorderedElementsAre(Pair(Http::LowerCaseString("x-bar-header"), "bar"),
+                           Pair(Http::LowerCaseString("x-per-request-header"), "test_value")));
   EXPECT_THAT(transforms.headers_to_remove, ElementsAre(Http::LowerCaseString("x-baz-header")));
 }
 
@@ -2013,9 +2101,9 @@ response_headers_to_remove: ["x-baz-header"]
   EXPECT_THAT(transforms.headers_to_append_or_add,
               ElementsAre(Pair(Http::LowerCaseString("x-foo-header"), "foo")));
   EXPECT_THAT(transforms.headers_to_overwrite_or_add,
-              ElementsAre(Pair(Http::LowerCaseString("x-bar-header"), "bar"),
-                          Pair(Http::LowerCaseString("x-per-request-header"),
-                               "%PER_REQUEST_STATE(testing)%")));
+              UnorderedElementsAre(Pair(Http::LowerCaseString("x-bar-header"), "bar"),
+                                   Pair(Http::LowerCaseString("x-per-request-header"),
+                                        "%PER_REQUEST_STATE(testing)%")));
   EXPECT_THAT(transforms.headers_to_remove, ElementsAre(Http::LowerCaseString("x-baz-header")));
 }
 


source/common/router/header_parser.cc
```diff
@@ -351,19 +358,34 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
       switch (entry.append_action_) {
         PANIC_ON_PROTO_ENUM_SENTINEL_VALUES;
       case HeaderValueOption::APPEND_IF_EXISTS_OR_ADD:
-        headers.addReferenceKey(key, value);
+        headers_to_add.emplace_back(key, value);
```
jmarantz, line 361: perf nit: if stream_info is nullptr then a pointless string-copy happens for every header.

Consider having the stream_info!=null check outside the loop and do something entirely different in that case.

jmarantz (commented):
This needs a microcbenchmark.

Also see the note I left after the merge of https://github.com/envoyproxy/envoy/pull/21932 -- I have a lot of concerns abou the use of std::regex there generally, although the lowest hanging fruit is not to recompile it on every call.

cpakulski:
@jmarantz Thanks for reviewing the code. I will add microbenchmark code plus some results.

source/common/router/header_parser.cc
```diff
@@ -351,19 +358,34 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
...
         break;
       case HeaderValueOption::ADD_IF_ABSENT:
         if (auto header_entry = headers.get(key); header_entry.empty()) {
-          headers.addReferenceKey(key, value);
+          headers_to_add.emplace_back(key, value);
```
adisuissa, line 365: I think this changes the semantics, and I think there's a need for more tests and docs clarification of the desired behavior.

Previous semantics:
APPEND_IF_EXISTS_OR_ADD["a", "val1"]
ADD_IF_ABSENT["a", "val2"]
will result with {"a"->"val1"}

New semantics:
APPEND_IF_EXISTS_OR_ADD["a", "val1"]
ADD_IF_ABSENT["a", "val2"]
will result with {"a"->"val1,val2"}

Similarly multiple ADD_IF_ABSENT["a", "val1"], ADD_IF_ABSENT["a", "val2"]

adisuissa: Taking a step back, I wonder if the right way to look at this is to reject a config that has multiple headers_to_add_ keys with the same value (i.e., it should be a map in the first place).

source/common/router/header_parser.cc
```diff
@@ -351,19 +358,34 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
...
         }
         break;
       case HeaderValueOption::OVERWRITE_IF_EXISTS_OR_ADD:
-        headers.setReferenceKey(key, value);
+        // TODO(cpakulski): Hash map is used to track multiple OVERWRITE actions
+        // and store only the last one. It would be more efficient to do it
+        // when config is processed. All OVERWRITE actions for the same header could be dropped
+        // except the last one.
+        headers_to_overwrite.erase(key);
+        headers_to_overwrite.try_emplace(key, key, value);
```
adisuissa, line 374: Can `insert_or_assign` be used here?

adisuissa (commented):
Thanks for working on this!

source/common/router/header_parser.cc
cpakulski: I think that the same problem can be recreated with multiple headers too, not with only one:
start with: `a = "aaaaaa", b="bbbbbb"`

set b to `%REQ(a)%%REQ(a)%` -> b is aaaaaaaaaaaa
add new header c as `%REQ(b)%%REQ(b)%`. It should be bbbbbbbbbbbb but it will be aaaaaaaaaaaaaaaaaaaaaaaa

cpakulski: > I think this changes the semantics, and I think there's a need for more tests and docs clarification of the desired behavior.
> 
> Previous semantics:
> APPEND_IF_EXISTS_OR_ADD["a", "val1"]
> ADD_IF_ABSENT["a", "val2"]
> will result with {"a"->"val1"}
> 
> New semantics:
> APPEND_IF_EXISTS_OR_ADD["a", "val1"]
> ADD_IF_ABSENT["a", "val2"]
> will result with {"a"->"val1,val2"}
> 
> Similarly multiple ADD_IF_ABSENT["a", "val1"], ADD_IF_ABSENT["a", "val2"]

To be honest, I am not sure if we should worry about that, because it is kind of "artificial" config, which specifies multiple values for the same header, which in practice should never happen. IMHO the only thing we should be concerned here is that a wrong config does not lead to CPU/memory overload or exhaustion (what is done in this PR). In the worst case a user may disable runtime guard to go back to the original behaviour. I think that your suggestion to catch things at the config time (allow only one modifier per header name) is much better idea. WDYT @adisuissa ?

source/common/router/header_parser.cc
cpakulski: I tried it and it is not possible at the moment. It complains that `Http::LowerCaseString` has deleted copy constructor (or something around that). Let me investigate it more. I agree that `insert_or_assign` is cleaner than doing `erase` and `emplace`.

source/common/router/header_parser.cc
cpakulski: @jmarantz Would you mind elaborating more about:

`perf nit: if stream_info is nullptr then a pointless string-copy happens for every header.` ? Thanks!

source/common/router/header_parser.cc
adisuissa: I think rejecting is a better approach if it doesn't break compatibility with the previous version of this code.

jmarantz:
Once #23447 lands can you add in a re-run of the benchmarks w/o removing the current ones from the description? I'm curious what impact that has.

cpakulski:
@jmarantz fixes in #23447 improved configuration path but the newly added benchmark measures speed of evaluating headers on data path. The results are the same.

jmarantz:
So after the PR, if there are 20 headers in the config there's a per-request overhead of 5.8 ms? That seems like a lot still. Why does it take that long? Do we have flame-graphs? Sorry I didn't dive into the actual code yet. :)

source/common/router/header_parser.cc
jmarantz: it looks like it's valid to call evaluateHeaders with stream_info==nullptr. If you follow the code, you'll. see we check for that for every header in headers_to_add_, and we'll be executing
```
  const std::string value = false ? .... : entry.original_value;
```
that will do a string copy and allocation which is not needed. Instead you could hoist the `stream_info!=nullptr` check outside the loop (saving a bunch of branches even if stream_info!=nullptr) and early-exit (saving allocations and string copies if stream_info==nullptr).

cpakulski:
If I am not mistaken, benchmark measures execution time of code only within `for (auto_ : state)`

 ```
for (auto _ : state) { // NOLINT: Silences warning about dead store
    // Code to benchmark
    header_parser->evaluateHeaders(request_header, request_header,
                                   *Http::StaticEmptyHeaders::get().response_headers, *stream_info);
  }
```

The body on benchmark function `bmEvaluateHeaders` prepares config, but I think that part is ignored during benchmarking.

**Comment by jmarantz** [comment]
Yes benchmarking will only measure inside that loop.

File: source/common/router/header_parser.cc (hunk 1 above)
Comment (cpakulski) [reply source/common/router/header_parser.cc:361]: Got it. Thanks for explaining. Let me try to refactor it and benchmark again.

File: source/common/router/header_parser.cc (hunk 3 above)
Comment (jmarantz) [reply source/common/router/header_parser.cc:374]: i've never used try_emplace and do not understand why `key` is specified twice. But how about just writing `headers_to_overwrite[key] = value;` ?

**Comment by cpakulski** [comment]
> So after the PR, if there are 20 headers in the config there's a per-request overhead of 5.8 ms? That seems like a lot still. Why does it take that long?

The numbers are practically the same using a "old" code (before PR). Let me dig into it more.

**Comment by cpakulski** [comment]
> So after the PR, if there are 20 headers in the config there's a per-request overhead of 5.8 ms? That seems like a lot still. Why does it take that long? Do we have flame-graphs? Sorry I didn't dive into the actual code yet. :)

I produced flame-graphs and realized that the speed test was using `Http::TestRequestHeaderMapImpl` which has "extra" routines for checking consistency and was adding significant delay. After switching to `Http::RequestHeaderMapImpl`, speed test returns much better results, but the changes in this PR add some delay and I am not sure if this is acceptable. Updated speed test results are in PR description.

File: source/common/router/header_parser.cc (hunk 1 above)
Comment (cpakulski) [reply source/common/router/header_parser.cc:361]: Going back to string copy: I thought that `std::string`'s copy constructor and assignment do not really copy the underlying string, but only where the string is. The copy happens only when one of the instances decides to modify the string and produces its own copy of underlying string before modifying it. Not sure if this is still the case, but I remember reading about that some time ago.

Comment (jmarantz) [reply source/common/router/header_parser.cc:361]: That was true 10+ years ago (copy-on-write) but is no longer true at least for the STL impls I know. The locking overhead needed for that feature turned out to make it a net loss.

**Comment by jmarantz** [comment]
Can you add before/after metrics for the case I called out in my earlier comment, where stream_info is nullptr?

I would like to have this PR be close to zero penalty if this feature is not being used.

**Comment by cpakulski** [comment]
> Can you add before/after metrics for the case I called out in my earlier comment, where stream_info is nullptr?
> 
> I would like to have this PR be close to zero penalty if this feature is not being used.
> 

Yes, I will do that. I am still trying to find the most optimum way (with minimum penalty) of handling cases where `stream_info` is not `nullptr`.

File: source/common/router/header_parser.cc (hunk 3 above)
Comment (cpakulski) [reply source/common/router/header_parser.cc]: Refactored code not to use hash map, so this is not an issue anymore.

File: source/common/router/header_parser.cc (hunk 2 above)
Comment (cpakulski) [reply source/common/router/header_parser.cc:371]: Modified config processing code to reject multiple OVERWRITE_IF_EXISTS_OR_ADD for the same header. This simplifies code when doing actual header evaluation.

**Comment by cpakulski** [comment]
@jmarantz I run performance tests for PR code and "before PR" code and results are not what I would expect for `stream_info` equal to `nullptr`. (results are in PR description).

For case when `stream_info` is valid, there is small degradation comparing to "before PR" code, but this is understandable, because headers must evaluated based on original values and new values must be temporarily stored in a vector and later on moved to the actual header map, so there are some extra steps.

I modified the code to handle `stream_info == nullptr` on a separate path, but it did not improve the timing. Actually it is even a bit worse comparing to "before PR" code.

File: source/common/router/header_parser.cc [hunk 4]
```diff
@@ -342,25 +355,63 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
...
+
+    // Now add headers which should be added.
+    for (const auto& header : headers_to_add) {
+      headers.addReferenceKey(header.first, header.second);
+    }
+  } else {
+    // The logic below is very similar to code executed when stream_info != nullptr,
+    // It is possible to combine two branches of "if" statement into more elegant code,
+    // but based on performance testing, using separate branches depending on value
+    // of stream_info resulted in the most efficient code.
```
jmarantz, line 398: oh I didn't realize we still had a complex loop to run if stream_info was null; I was thinking this meant  we could short-circuit entirely.

source/common/router/header_parser.cc
```diff
@@ -251,6 +251,7 @@ HeaderFormatterPtr parseInternal(const envoy::config::core::v3::HeaderValue& hea
 
 HeaderParserPtr
 HeaderParser::configure(const Protobuf::RepeatedPtrField<HeaderValueOption>& headers_to_add) {
+  std::set<absl::string_view> headers_to_overwrite;
```
jmarantz, line 254: Can we sink this variable into a scope so it is freed before line 358, and also rename one of the  two different headers_to_overwrite variables? It's unambiguous to the compiler but confusing to humans.

Also I think this can be a flat_hash_set.

source/common/router/header_parser.cc
jmarantz: WDYT of factoring out lines 367-382, and lines 400-416, which are basically the same, into a parameterized helper method or lambda that takes the value as a formal.

source/common/router/header_parser.cc
cpakulski: Changed the type to `absl::flat_hash_set`. Thanks!

Hmm, it is true that I used `headers_to_overwrite` in two places, but they are in separate functions. The scope of the first `headers_to_overwrite` is limited to `HeaderParser::configure` method and the second one is in `HeaderParser::evaluateHeaders` method. I can still change the name, but have problems with finding a good name :-)

source/common/router/header_parser.cc
cpakulski: I have actually already done it (lambdas). The code looks more elegant, but the performance is worse. I am not sure if this is related to capturing params or by introducing extra function calls:

  ```
std::vector<std::pair<const Http::LowerCaseString&, const std::string>> headers_to_add;
  std::vector<std::pair<const Http::LowerCaseString&, const std::string>> headers_to_overwrite;

  std::function<void(const Http::LowerCaseString&, const HeadersToAddEntry&)> header_append_func,
      header_overwrite_func;
  if (stream_info == nullptr) {
    header_append_func = [&headers](const Http::LowerCaseString& key,
                                    const HeadersToAddEntry& entry) {
      if (!entry.original_value_.empty() || entry.add_if_empty_) {
        headers.addReferenceKey(key, entry.original_value_);
      }
    };
    header_overwrite_func = [&headers](const Http::LowerCaseString& key,
                                       const HeadersToAddEntry& entry) {
      if (!entry.original_value_.empty() || entry.add_if_empty_) {
        headers.setReferenceKey(key, entry.original_value_);
      }
    };
  } else {
    header_append_func = [&request_headers, &response_headers, &stream_info, &headers_to_add](
                             const Http::LowerCaseString& key, const HeadersToAddEntry& entry) {
      const std::string value =
          entry.formatter_->format(request_headers, response_headers, *stream_info);
      if (!value.empty() || entry.add_if_empty_) {
        headers_to_add.emplace_back(key, value);
      }
    };
    header_overwrite_func = [&request_headers, &response_headers, &stream_info,
                             &headers_to_overwrite](const Http::LowerCaseString& key,
                                                    const HeadersToAddEntry& entry) {
      const std::string value =
          entry.formatter_->format(request_headers, response_headers, *stream_info);
      if (!value.empty() || entry.add_if_empty_) {
        headers_to_overwrite.emplace_back(key, value);
      }
    };
  }

  for (const auto& [key, entry] : headers_to_add_) {
    switch (entry.append_action_) {
      PANIC_ON_PROTO_ENUM_SENTINEL_VALUES;
    case HeaderValueOption::APPEND_IF_EXISTS_OR_ADD:
      header_append_func(key, entry);
      break;
    case HeaderValueOption::ADD_IF_ABSENT:
      if (auto header_entry = headers.get(key); header_entry.empty()) {
        header_append_func(key, entry);
      }
      break;
    case HeaderValueOption::OVERWRITE_IF_EXISTS_OR_ADD:
      header_overwrite_func(key, entry);
      break;
    }
  }

  // First overwrite all headers which need to be overwritten.
  for (const auto& header : headers_to_overwrite) {
    headers.setReferenceKey(header.first, header.second);
  }

  // Now add headers which should be added.
  for (const auto& header : headers_to_add) {
    headers.addReferenceKey(header.first, header.second);
  }
```

So, for valid stream_info the results are:
when there are 2 separate paths with quite similar code `if (stream_info != null) {switch: ...} else {switch: ...}`
```
bmEvaluateHeaders/2        8767 ns         8759 ns        78696
bmEvaluateHeaders/4       21864 ns        21835 ns        31680
bmEvaluateHeaders/6       34422 ns        34383 ns        20185
bmEvaluateHeaders/8       47098 ns        47038 ns        14951
bmEvaluateHeaders/10      59249 ns        59183 ns        11892
bmEvaluateHeaders/12      72207 ns        72126 ns         9793
bmEvaluateHeaders/14      83693 ns        83585 ns         8450
bmEvaluateHeaders/16      95415 ns        95343 ns         7336
bmEvaluateHeaders/18     107372 ns       107341 ns         6471
bmEvaluateHeaders/20     120339 ns       120320 ns         5736
```
and when using lambdas:
```
bmEvaluateHeaders/2        9905 ns         9904 ns        70521
bmEvaluateHeaders/4       23442 ns        23440 ns        29858
bmEvaluateHeaders/6       35770 ns        35756 ns        19602
bmEvaluateHeaders/8       48154 ns        48152 ns        14576
bmEvaluateHeaders/10      60328 ns        60320 ns        11551
bmEvaluateHeaders/12      72932 ns        72918 ns         9601
bmEvaluateHeaders/14      84671 ns        84649 ns         8287
bmEvaluateHeaders/16      96296 ns        96254 ns         7235
bmEvaluateHeaders/18     109546 ns       109532 ns         6387
bmEvaluateHeaders/20     122509 ns       122487 ns         5653
```

And when stream_info is nullptr:
when there are 2 separate paths with quite similar code `if (stream_info != null) {switch: ...} else {switch: ...}`
```
bmEvaluateHeaders/2        6699 ns         6694 ns       104734
bmEvaluateHeaders/4       17388 ns        17387 ns        40493
bmEvaluateHeaders/6       27539 ns        27535 ns        25916
bmEvaluateHeaders/8       38423 ns        38417 ns        18217
bmEvaluateHeaders/10      48765 ns        48759 ns        14305
bmEvaluateHeaders/12      60458 ns        60448 ns        11471
bmEvaluateHeaders/14      69731 ns        69722 ns        10092
bmEvaluateHeaders/16      79537 ns        79528 ns         8744
bmEvaluateHeaders/18      91209 ns        91205 ns         7470
bmEvaluateHeaders/20     102027 ns       102022 ns         6784
```
and with lambdas:
```
bmEvaluateHeaders/2        7157 ns         7154 ns        97919
bmEvaluateHeaders/4       18311 ns        18306 ns        38525
bmEvaluateHeaders/6       28535 ns        28529 ns        24947
bmEvaluateHeaders/8       39269 ns        39262 ns        17732
bmEvaluateHeaders/10      50440 ns        50416 ns        13879
bmEvaluateHeaders/12      60673 ns        60662 ns        11518
bmEvaluateHeaders/14      70968 ns        70952 ns         9895
bmEvaluateHeaders/16      81705 ns        81686 ns         8635
bmEvaluateHeaders/18      92027 ns        92003 ns         7501
bmEvaluateHeaders/20     103374 ns       103325 ns         6796
```
So, the current code with `if (stream_info != nullptr) {switch: ....} else {switch: ....}` gives the most performance. Switching to lambdas adds around (5-10)% for small number of headers and about 2% for 20 headers.

I like lambdas as the code is more elegant. I also measured timing using code compiled with `dbg` flag, so maybe optimized code results in less of performance loss.

Comment (jmarantz) [reply source/common/router/header_parser.cc:398]: Thanks -- Can you run the benchmarks with just "-c opt"? Sorry for the trouble :)

I think the use of the lambdas will cause more stack-frames to be generated, and you'd want to see those while debugging, but it's possible the optimizer would remove them if it doesn't have to keep around all the stack-frames for debugging. E.g. typically in flame graphs on optimized code you won't see the entire call-stack.

Comment (cpakulski) [reply source/common/router/header_parser.cc:398]: I re-run all tests using optimized code and timings are completely different. Thanks for suggesting it. The PR introduces a small delay (because it needs to temporarily store results in a vector while new headers values are calculated), but the delay is not significant. I also replaced `std::vector` in favour of `absl::InlinedVector` and it improved the timings. lambdas are still a bit slower than using if (stream_info) {...} else {...}, but the difference is small now. The exact numbers are below. Thanks for analyzing my changes with such detail. This code is executed for each HTTP packet, so I am totally OK with trying different approached and profiling it.

```
OPTIMIZED BUILD
===============

stream_info is not nullptr

original:
bmEvaluateHeaders/2         239 ns          239 ns      2933249
bmEvaluateHeaders/4        1058 ns         1058 ns       660594
bmEvaluateHeaders/6        1758 ns         1758 ns       498900
bmEvaluateHeaders/8        2093 ns         2093 ns       422586
bmEvaluateHeaders/10       2275 ns         2274 ns       326849
bmEvaluateHeaders/12       2621 ns         2621 ns       271865
bmEvaluateHeaders/14       3042 ns         3041 ns       227186
bmEvaluateHeaders/16       3538 ns         3538 ns       196645
bmEvaluateHeaders/18       3968 ns         3968 ns       174172
bmEvaluateHeaders/20       4558 ns         4558 ns       157163

this PR (if (strem_info == nullptr) {...} else {...}
bmEvaluateHeaders/2         255 ns          255 ns      2754158
bmEvaluateHeaders/4         992 ns          992 ns       703809
bmEvaluateHeaders/6        1688 ns         1688 ns       442525
bmEvaluateHeaders/8        2053 ns         2053 ns       374797
bmEvaluateHeaders/10       2544 ns         2543 ns       288706
bmEvaluateHeaders/12       2985 ns         2985 ns       235778
bmEvaluateHeaders/14       3489 ns         3489 ns       202675
bmEvaluateHeaders/16       3984 ns         3983 ns       178978
bmEvaluateHeaders/18       4514 ns         4514 ns       156524
bmEvaluateHeaders/20       5096 ns         5094 ns       134769

this PR lambdas
bmEvaluateHeaders/2         324 ns          324 ns      2152446
bmEvaluateHeaders/4        1031 ns         1031 ns       664590
bmEvaluateHeaders/6        1729 ns         1729 ns       430220
bmEvaluateHeaders/8        2131 ns         2131 ns       361016
bmEvaluateHeaders/10       2623 ns         2623 ns       281483
bmEvaluateHeaders/12       3046 ns         3045 ns       226996
bmEvaluateHeaders/14       3598 ns         3597 ns       197573
bmEvaluateHeaders/16       4044 ns         4044 ns       173756
bmEvaluateHeaders/18       4697 ns         4696 ns       153256
bmEvaluateHeaders/20       5319 ns         5318 ns       129045

stream_info is nullptr

original (before PR):
bmEvaluateHeaders/2         164 ns          164 ns      4256833
bmEvaluateHeaders/4         710 ns          710 ns       979517
bmEvaluateHeaders/6        1183 ns         1183 ns       695658
bmEvaluateHeaders/8        1539 ns         1539 ns       629900
bmEvaluateHeaders/10       1663 ns         1663 ns       483293
bmEvaluateHeaders/12       1973 ns         1973 ns       403045
bmEvaluateHeaders/14       2147 ns         2147 ns       338610
bmEvaluateHeaders/16       2617 ns         2617 ns       295247
bmEvaluateHeaders/18       2702 ns         2702 ns       252027
bmEvaluateHeaders/20       3040 ns         3040 ns       233345

this PR (if (strem_info == nullptr) {...} else {...}
bmEvaluateHeaders/2         159 ns          159 ns      4410206
bmEvaluateHeaders/4         690 ns          689 ns      1003884
bmEvaluateHeaders/6        1145 ns         1145 ns       698180
bmEvaluateHeaders/8        1354 ns         1354 ns       622681
bmEvaluateHeaders/10       1561 ns         1561 ns       486003
bmEvaluateHeaders/12       1762 ns         1762 ns       413769
bmEvaluateHeaders/14       2076 ns         2076 ns       338057
bmEvaluateHeaders/16       2405 ns         2405 ns       297029
bmEvaluateHeaders/18       2692 ns         2691 ns       262014
bmEvaluateHeaders/20       2990 ns         2990 ns       233128

this PR lambdas:
bmEvaluateHeaders/2         171 ns          171 ns      4083766
bmEvaluateHeaders/4         690 ns          690 ns      1002158
bmEvaluateHeaders/6        1154 ns         1154 ns       694188
bmEvaluateHeaders/8        1357 ns         1356 ns       606530
bmEvaluateHeaders/10       1571 ns         1571 ns       480932
bmEvaluateHeaders/12       1790 ns         1790 ns       403867
bmEvaluateHeaders/14       2085 ns         2085 ns       344280
bmEvaluateHeaders/16       2383 ns         2382 ns       297992
bmEvaluateHeaders/18       2679 ns         2679 ns       262552
bmEvaluateHeaders/20       3037 ns         3037 ns       236217
```

Comment (jmarantz) [reply source/common/router/header_parser.cc:398]: It's fun optimizing stuff that matters :)

I will take a look later but I had a slightly different model in mind for how to use lambdas but let me look in detail at what you have when I get a bit of time.

Comment (jmarantz) [reply source/common/router/header_parser.cc:398]: OK I just scanned over this again and I think this is what I would try, and why. I think the big-if-statement probably is what you want the compiler to generate for speed, but it's not readable/maintainable.

I think probably the lambdas are hard for the compiler to optimize away, and they are also not that easy to read -- they make this function very long.

I've thought of some other ways to do this but I am not confident the compiler will optimize them well. One I had been thinking was that the whole loop body in a lambda or function, which took the value as a formal. Then you'd have two copies of the for-loop but they would be one-liners calling the body one way or the other:

```
if (stream_info != nullptr) {
  for ([key, entry]. ,...) {
    handleKeyValue(request_headers, response_headers, stream_info, headers_to_overwrite, 
                   entry.formatter_->format(request_headers, response_headers, *stream_info));
  }
} else {
    for ([key, entry]. ,...) {
    handleKeyValue(request_headers, response_headers, &stream_info, &headers_to_overwrite, 
                   entry.original_value_);
  }
}
```
But I have another idea also. My main objection to your original loop was that you did a lot of pointless string copying in the case where stream_info==null. You could also avoid by using a string_view to hold the value, and a single string buffer declared outside the loop that you use to store the value when stream_info!=null.

Something like:
```
std::string value_buffer;
for (...) {
  absl::string_view value;
  if (stream_info != nullptr) {
    value_buffer = entry.formatter_->format(request_headers, response_headers, *stream_info);
    value = value_buffer;
  } else {
    value = entry.original_value_;
  }
  switch ...
}
```

I'll let you mull these options over; pick something that balances readability (ie functions not too long) and having decent performance.

One thing I'm not certain of: strings are actually pretty efficient these days if they are 22 bytes or less (search for "c++ small string optimization" -- I couldn't find any one authoritative post). So if the actual headers are mostly small your code is not that bad -- the main problem is the extra branch not the copies. But the moment you hit 23 bytes or 24 bytes  (I think) it needs a heap allocation. So depending on how you write your perf test you will get very different results.

Please pick something and document the tradeoffs in a comment and we can just go with it.

Comment (cpakulski) [reply source/common/router/header_parser.cc]: Thanks for the great feedback. I think that your idea with using `absl::string_view`:
```
std::string value_buffer;
for (...) {
  absl::string_view value;
  if (stream_info != nullptr) {
    value_buffer = entry.formatter_->format(request_headers, response_headers, *stream_info);
    value = value_buffer;
  } else {
    value = entry.original_value_;
  }
  switch ...
}
```
strikes the best balance between readability and performance. 
I compared speed test results to `if {big block} else {big block}`  and new code results are almost identical for valid `*stream_info` and a bit worse for `stream_info` equal to `nullptr`.

As for small string optimizations, you are right that for clang, the cut off size is 22 bytes. The test results are different for long strings, but IMHO the performance hit  is unavoidable for any approach: _if {big block} else {big block}_, lambdas or the current code.

File: source/common/router/header_parser.cc [hunk 6]
```diff
@@ -342,28 +355,48 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
     headers.remove(header);
   }
 
+  // Temporary storage to hold evaluated values of headers to add and replace. This is required
+  // to execute all formatters using the original received headers.
+  // Only after all the formatters produced the new values of the headers, the headers are set.
+  // absl::InlinedVector is optimized for 4 headers. After that it behaves as normal std::vector.
```
jmarantz, line 361: nit: absl;;InlineVector is I think capable of any optimal size; it's you that's picking the '4', right?

Can you explain why you picked 4? benchmarks?

test/common/router/header_formatter_speed_test.cc
```diff
@@ -0,0 +1,51 @@
...
+namespace Envoy {
+namespace Router {
+
+static void bmEvaluateHeaders(benchmark::State& state) {
+  auto request_header = Http::RequestHeaderMapImpl::create();
+  request_header->addCopy(Http::LowerCaseString("bar"), "a");
+  request_header->addCopy(Http::LowerCaseString("foo"), 1);
+  request_header->addCopy(Http::LowerCaseString("test1"), "to_overwrite");
+
+  MockTimeSystem time_system;
```
jmarantz, line 18: I would avoid mocks for tests generally. I'm always suspicious they are slow.

Can you just use `Event::SimulatedTimeSystem time_system` instead?

jmarantz (approved):
lgtm modulo nits from a readability/perf/maintainability perspective.

But I'd like @adisuissa to give the final approval on semantic change.

source/common/router/header_parser.cc
cpakulski: The reasoning behind picking 4 was that I believe that in the majority of use cases the number of added/replaced headers is rather small. I do not have any concrete data though. I was leaning towards 2,3 or 4. Above that I believe would be over-optimization. The benchmark actually tests adding and replacing 1,2,3,...,10 headers respectively, so it goes beyond 4 and I do not see a significant jump in timing after exceeding 4.

test/common/router/header_formatter_speed_test.cc
cpakulski: OK. Let me do that. Thanks.

source/common/router/header_parser.cc
jmarantz: sounds good. Throw that in the comment and I'm good to go :)

source/common/router/header_parser.cc
```diff
@@ -269,6 +270,18 @@ HeaderParser::configure(const Protobuf::RepeatedPtrField<HeaderValueOption>& hea
       append_action = header_value_option.append_action();
     }
 
+    // There should be only one OVERWRITE_IF_EXISTS_OR_ADD per each header.
```
adisuissa, line 273: This change breaks backward compatibility, and rejecting it should be done with care (runtime guarding it, and allowing users and management server implementers update their code in a reasonable time).
To accept the config and to keep the semantics, can the same be achieved by linearizing the header_to_add list (i.e., ignoring the previous addition of the header key)?

Note that my suggestion assumes that a header cannot have circular reference to itself (IIRC before #21932 this was not allowed).

source/common/router/header_parser.cc
```diff
@@ -269,6 +270,18 @@ HeaderParser::configure(const Protobuf::RepeatedPtrField<HeaderValueOption>& hea
...
+    if (append_action == HeaderValueOption::OVERWRITE_IF_EXISTS_OR_ADD) {
+      if (headers_to_overwrite.find(header_value_option.header().key()) !=
+          headers_to_overwrite.end()) {
+        throw EnvoyException(
+            fmt::format("Multiple OVERWRITE_IF_EXISTS_OR_ADD actions for header {} are not allowed",
+                        header_value_option.header().key()));
+      } else {
+        headers_to_overwrite.emplace(header_value_option.header().key());
```
adisuissa, line 281: Should this also take into account headers inserted with `APPEND_IF_EXISTS_OR_ADD`?

source/common/router/header_parser.cc
cpakulski: Accepting the config with multiple `OVERWRITE_IF_EXISTS_OR_ADD` is easy (above and below lines from here should be deleted). It works fine, but CPU cycles will be wasted because whatever is calculated before the last occurrence of  `OVERWRITE_IF_EXISTS_OR_ADD` will be discarded and overwritten. So, to keep backwards compatibility for config I think that we should allow multiple `OVERWRITE_IF_EXISTS_OR_ADD` for the same headers.

As or semantics, I believe that what we are suggesting now is the cleanest (and probably compatible with what was before #21932, but I have to check). I think that when documentation uses words `IF_EXISTS` it means the headers in received packet, not "intermediate" headers created by for example `APPEND_IF_EXISTS_OR_ADD` when a new header is built.

source/common/router/header_parser.cc
cpakulski: As per above comment, I suggest that we drop that check and allow not-optimal config. It does not really affect the main issue this PR addresses: circular input which grows each time it is executed.

source/common/router/header_parser.cc
adisuissa: RE wasted CPU cycles, I agree. This can be addressed by linearizing the list of operations, and removing any operation that is overridden (iterating the list backwards, and maintaining the first encounter of each key).

I believe that the point you make about the semantics of `IF_EXISTS` is correct. I'm worried that rejecting a config that was valid before may cause some harm on current deployments that for some reason send a config that has multiple `OVERWRITE_IF_EXISTS_OR_ADD` actions for the same header.

cpakulski: @adisuissa 

> RE wasted CPU cycles, I agree. This can be addressed by linearizing the list of operations, and removing any operation that is overridden (iterating the list backwards, and maintaining the first encounter of each key).

Let me do it as a separate PR. In case we need to revert it, it will be cleaner. Wasted CPU is a pre-existing issue, not related what we are addressing in this PR.

So, at the moment the code does not introduce any backward compatibility issues and if you do not see any other issues I think it it complete.

source/common/router/header_parser.cc
```diff
@@ -342,28 +342,53 @@ void HeaderParser::evaluateHeaders(Http::HeaderMap& headers,
...
+  // Only after all the formatters produced the new values of the headers, the headers are set.
+  // absl::InlinedVector is optimized for 4 headers. After that it behaves as normal std::vector.
+  // It is assumed that most of the use cases will add or modify fairly small number of headers
+  // (<=4). If this assumption changes, the number of inlined capacity should be increased.
+  // header_formatter_speed_test.cc provides micro-benchmark for evaluating speed of adding and
+  // replacing headers and should be used when modifying the code below to access the performance
+  // impact of code changes.
+  absl::InlinedVector<std::pair<const Http::LowerCaseString&, const std::string>, 4> headers_to_add,
+      headers_to_overwrite;
+  std::string value_buffer;
```
adisuissa, line 356: nit: not needed.

adisuissa (commented):
LGTM, modulo small nit.
Thanks for fixing this!

source/common/router/header_parser.cc
jmarantz: It is needed to avoid superfluous string-copies when stream_info==nullptr

actually I suspect we could also make entry.formatter_->format avoid the string-copy when it's not doing anything, which would make it matter more to have this buffer.

@adisuissa see comment-thread at line 369.

adisuissa: Not sure I fully grok, if `stream_info == nullptr` then `value_buffer` won't be used, or am I missing something here? (FWIW, I see the benefit of `value`, but not of `value_buffer`)
In any case, if this is some intentional code pattern, I suggest adding some comments that describe this.

jmarantz: This is to have a place to store the computed value from entry.formatter_->format() when stream_info!=null, but avoiding the string alloc+copy when stream_info==null.

I agree that could be added as a comment.

jmarantz:
@cpakulski I think we are just waiting for you to add a comment to the value_buffer; I did explain it offline to Adi but probably a comment makes sense.

Thanks!

/wait

jmarantz:
Actually I'll go ahead and queue up senior maintainer for this also, as this is deep in the data plane.

cpakulski:
Added the comment to explain the code structure. @jmarantz Thanks again for guiding me through low level details!

adisuissa (approved):
LGTM, thanks!

alyssawilk (commented):
Thanks so much for tackling this change.  Quick question before I review - does this actually result in (correct) changes to the resultant headers?  If so I think we need to runtime guard and release note
/wait-any

cpakulski:
> Quick question before I review - does this actually result in (correct) changes to the resultant headers? If so I think we need to runtime guard and release note

@alyssawilk For "normal" use cases this PR does not bring any changes. Notice that all unit and integration tests are unchanged. It affected only specific, not practical scenarios, when the same header is modified multiple times and additionally it refers to itself, like the following pseudo configuration:
step1: overwrite header "x-header" to be %REQ(x-header)% %REQ(x-header)%
step2: overwrite header "x-header" to be %REQ(x-header)% %REQ(x-header)%
step3: overwrite header "x-header" to be %REQ(x-header)% %REQ(x-header)%
step4: overwrite header "x-header" to be %REQ(x-header)% %REQ(x-header)%

Documentation does not describe such cases, but I think it is assumed that REQ(x-header) always refers to the original received header, not x-header modified by previous steps. 

Adding runtime guard is simple, because the code is concentrated and I can easily add it if you think it should be added.

alyssawilk (approved):
hm, with your explanation I'm fine landing as-is but I'm going to cc in Yan for a second google pass as we've been doing lots of discussing user-visible changes this week

adisuissa:
I think this technically a user-visible change, but somewhat different than the one we typically address.

Prior to #21932, a headers_to_add update could not reference a header (maybe even self reference the same header).
#21932 enabled it, but introduced an issue where the loop that updated the headers values, used the "live" (during the loop) value of the header. This is the first user-visible change, and unfortunately we also have this in the released v1.24.0.

This PR attempts to revert the behavior back, so the header value used will be the non-"live" (before the loop) value of the header, which is compatible with the state prior to #21932.

IMHO the reasons for not introducing a runtime guard is because a reasonable user would not have used the "headers_to_add" this way, and that this PR "reverts" the behavior to the way it was before.
That said, if adding a runtime guard can easily be done, then it may be better to do so.

cpakulski:
After some thinking I believe that runtime guard is not required here. #21932 itself has runtime guard to go back to the original version. So, another runtime guard (in this PR) would be only for users who use v1.24.0, want new unified formatters and additionally used a weird, not practical config.

PR #23401, opened 2022-10-07
http conn man: fix memory leak with watermark buffer

Hexta:
Fix memory leak with stream recreated when watermark buffer tracking enabled.

Do not create a new memory account for the recreated stream in ConnectionManagerImpl::ActiveStream::recreateStream.
Re-use an account from the old stream.

Risk Level: low
Testing: integration

### DIFF
--- a/source/common/http/conn_manager_impl.cc
+++ b/source/common/http/conn_manager_impl.cc
@@ -305,12 +305,17 @@ RequestDecoder& ConnectionManagerImpl::newStream(ResponseEncoder& response_encod
 
   ENVOY_CONN_LOG(debug, "new stream", read_callbacks_->connection());
 
-  // Create account, wiring the stream to use it for tracking bytes.
-  // If tracking is disabled, the wiring becomes a NOP.
-  auto& buffer_factory = read_callbacks_->connection().dispatcher().getWatermarkFactory();
   Buffer::BufferMemoryAccountSharedPtr downstream_stream_account =
-      buffer_factory.createAccount(response_encoder.getStream());
-  response_encoder.getStream().setAccount(downstream_stream_account);
+      response_encoder.getStream().account();
+
+  if (downstream_stream_account == nullptr) {
+    // Create account, wiring the stream to use it for tracking bytes.
+    // If tracking is disabled, the wiring becomes a NOP.
+    auto& buffer_factory = read_callbacks_->connection().dispatcher().getWatermarkFactory();
+    downstream_stream_account = buffer_factory.createAccount(response_encoder.getStream());
+    response_encoder.getStream().setAccount(downstream_stream_account);
+  }
+
   ActiveStreamPtr new_stream(new ActiveStream(*this, response_encoder.getStream().bufferLimit(),
                                               std::move(downstream_stream_account)));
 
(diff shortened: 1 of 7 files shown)


source/common/http/conn_manager_impl.cc
```diff
@@ -1766,6 +1766,11 @@ void ConnectionManagerImpl::ActiveStream::recreateStream(
   // Make sure to not check for deferred close as we'll be immediately creating a new stream.
   connection_manager_.doEndStream(*this, /*check_for_deferred_close*/ false);
 
+  auto oldAccount = response_encoder->getStream().account();
```
KBaichoo, line 1769: can we follow style guide: https://google.github.io/styleguide/cppguide.html#Variable_Names for naming local variable

source/common/http/conn_manager_impl.cc
```diff
@@ -1766,6 +1766,11 @@ void ConnectionManagerImpl::ActiveStream::recreateStream(
...
+  if (oldAccount != nullptr) {
+    oldAccount->clearDownstream();
+  }
+
```
KBaichoo, line 1773: The slices in the `request_data` still are referring to the old account, we should have them transfered to refer to the account of the recreated stream (crediting the old account which will be destroyed at the end of this function)

test/integration/buffer_accounting_integration_test.cc
```diff
@@ -237,6 +237,51 @@ TEST_P(Http2BufferWatermarksTest, ShouldCreateFourBuffersPerAccount) {
...
+  ASSERT_TRUE(response->complete());
+
+  if (streamBufferAccounting()) {
+    EXPECT_TRUE(buffer_factory_->waitForExpectedAccountUnregistered(2)) << printAccounts();
+  } else {
+    EXPECT_TRUE(buffer_factory_->waitForExpectedAccountUnregistered(0));
+  }
+
+  if (streamBufferAccounting()) {
+    EXPECT_EQ(buffer_factory_->numAccountsCreated(), 2);
```
KBaichoo, line 279: can we validate the transfer of bytes in request body so that we include them in our accounting? (possibly in another test, this one has no request body).

KBaichoo (commented):
Thanks for catching this! Few improvements I think we can make.

source/common/http/conn_manager_impl.cc
Hexta: Probably, we can re-use the old account because the downstream connection is still the same.
WDYT?

KBaichoo: yeah I think that'd work and be simpler. We would need to modify `ConnectionManagerImpl::newStream` for the case there's an existing account associated which there will be if we're recreating the stream.

test/integration/buffer_accounting_integration_test.cc
```diff
@@ -237,6 +237,91 @@ TEST_P(Http2BufferWatermarksTest, ShouldCreateFourBuffersPerAccount) {
...
+    EXPECT_EQ(buffer_factory_->numAccountsCreated(), 1) << printAccounts();
+    EXPECT_TRUE(buffer_factory_->waitForExpectedAccountUnregistered(1)) << printAccounts();
+  } else {
+    EXPECT_EQ(buffer_factory_->numAccountsCreated(), 0);
+    EXPECT_TRUE(buffer_factory_->waitForExpectedAccountUnregistered(0));
+  }
+}
+
+TEST_P(Http2BufferWatermarksTest, AccountsAndInternalRedirectWithRequestBody) {
+  Http::TestResponseHeaderMapImpl redirect_response_{
```
KBaichoo, line 282: nit: naming, should be `redirect_response` the `_` suffix is reserved for object fields. Consider making const since it's not modified and shifting it down right where it's used.

Same with above test.

KBaichoo (commented):
Just minor changes, otherwise pending CI this looks good. Thanks for catching this.

KBaichoo (approved):
lgtm, thanks!

Issue #22926, opened 2022-08-31
Broken types when JSON transcoding with URL path fields


benjle:
**If you are reporting *any* crash or *any* potential security issue, *do not*
open an issue in this repo. Please report the issue via emailing
envoy-security@googlegroups.com where the issue will be triaged appropriately.**

*Title*: URL-path field types are inconsistently enforced

*Description*:
When a URL path parameter is provided with JSON transcoding, enforcement of type correctness is inconsistent depending on whether the body is empty or has any JSON content, even if the protobuf message does not have any non-URL-formatted parameters.

*Repro steps*:
1. Use this minimal proto/RPC to repro:
``` 
service TestService {
  rpc TestInt (TestIntRequest) returns (TestIntResponse) {
    option (google.api.http) = {
      post: "/v1/test/{url_int}/{url_string}"
      body: "*"
    };
  }
}

message TestIntRequest {
  int32 url_int = 1;
  string url_string = 2;
}

message TestIntResponse {
  int32 url_int = 1;
  string url_string = 2;
}
```

2. Implement this RPC in a service behind Envoy that just echos back the url_int and url_string parameters in the response.
3. Run a curl with int and string path parameters to verify happy path is working:
```
$ curl -X POST http://localhost:8080/v1/test/1234/qwer
{
 "urlInt": 1234,
 "urlString": "qwer"
}
```
4. Run a curl with string and string path parameters to show the first broken path:
```
$ curl -X POST http://localhost:8080/v1/test/rewq/qwer
{
 "urlInt": 0,
 "urlString": "qwer"
}
```
4. Run a curl with any JSON request body and string path parameters to show the second inconsistently-broken path:
```
$ curl -d '{}' -H 'Content-Type: application/json' -X POST http://localhost:8080/v1/test/rewq/qwer
url_int: invalid value "rewq" for type TYPE_INT32
```

Expected behavior is for response on steps 3 and 4 to be consistent in either rejecting the request or accepting the request with zero defaults.

>**Note**: The [Envoy_collect tool](https://github.com/envoyproxy/envoy/blob/main/tools/envoy_collect/README.md)
gathers a tarball with debug logs, config and the following admin
endpoints: /stats, /clusters and /server_info. Please note if there are
privacy concerns, sanitize the data prior to sharing the tarball/pasting.

*Admin and Stats Output*:
Stuff might look wonky because I deployed the test RPC in the same service as the one I was working on:
```
cluster.backend_service.assignment_stale: 0
[...]

PR #23403, opened 2022-10-07
grpc_transcoder: fix a bug with wrong binding type

qiwzhang:
To fix https://github.com/envoyproxy/envoy/issues/22926

As reported, If a binding field type is integer, but path segment is string,  the request should be rejected.  But current behaviors is not consistent.   When the request body is not empty,  it works as expected, the request is rejected.  But if request body is empty,  the request is not rejected.

Found two bugs:
* In [grpc_httpjson_transcoding](https://github.com/grpc-ecosystem/grpc-httpjson-transcoding) lib,  the error status is not propagated out properly in empty body case.
* In grpc_transcoder filter,  check transcode status is in the wrong place, it should be after readToBuffer() call.

Changes:
* Get the [fix](https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/pull/85)
* Fix the code in grpc_transcoder_filter
* Add an integration test to verify the fix

Risk Level:  None
Testing:  Unit tested and integration tested
Docs Changes: None
Release Notes:  Yes

### DIFF
--- a/test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
+++ b/test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
@@ -831,14 +831,14 @@ TEST_P(GrpcJsonTranscoderIntegrationTest, UnaryDelete) {
       "{}");
 }
 
-TEST_P(GrpcJsonTranscoderIntegrationTest, UnaryPatchWrongBindingType) {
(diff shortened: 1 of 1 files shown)


phlax:
cc @lizan

qiwzhang:
https://github.com/orgs/envoyproxy/teams/dependency-shepherds:  Please review and approve,  thanks

lizan (approved):

phlax (approved):
/lgtm deps

thanks @qiwzhang

PR #23405, opened 2022-10-07
[NO REVIEWS REQUIRED] updated envoy jwt api

aryan16:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
+++ b/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
@@ -277,6 +277,9 @@ message JwtProvider {
   // Enables JWT cache, its size is specified by *jwt_cache_size*.
   // Only valid JWT tokens are cached.
   JwtCacheConfig jwt_cache_config = 12;
+
+  // Add JWT Claim to Header
+  repeated JWTClaimToHeader claim_to_header = 15;
 }
 
 // This message specifies JWT Cache configuration.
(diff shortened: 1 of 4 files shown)


api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -277,6 +277,9 @@ message JwtProvider {
   // Enables JWT cache, its size is specified by *jwt_cache_size*.
   // Only valid JWT tokens are cached.
   JwtCacheConfig jwt_cache_config = 12;
+
+  // Add JWT Claim to Header
```
htuch, line 281: This feature needs some documentation on how to use etc., preferably in https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/jwt_authn_filter.html or related Envoy docs. 

It's a little unclear to me why we can't just write to dynamic metadata and then use Envoy's existing header addition capabilities to do this. CC @lizan @qiwzhang

aryan16: sure, I will update the documentation. And this has been a popular ask, so it's good to add a standalone field for this. 
Closed this PR due to some issue with this branch, currently working on this one - https://github.com/envoyproxy/envoy/pull/23444

aryan16: And Even if we go ahead with dynamic metadata solution, we would require the claim key name and header name from user which would require a new field in the existing API.

Issue #23322, opened 2022-09-30
Thrift Payload-To-Metadata Filter


JuniorHsu:
*Title*: a filter to set metadata from a thrift request field

*Description*:

This filter is configured with `request_rules` that will be matched against requests. A `field_selector` of a rule represents the head of a linked list, each node of the linked list has a `name` for logging and an `id` for matching. The `field_selector` is tied to a payload field when the linked list corresponds to a downward path which rooted in the top-level of the request message structure. `on_present` is triggered when corresponding the payload is present. Otherwise, `on_missing` is triggered.

This filter is designed to support payload passthrough. By performing payload to metadata filter can do deserialization once, and pass the metadata to other filters. This means that load balancing decisions, consumed from log and routing could all use payload information with a single parse. Also notably performing the parsing in payload passthrough buffer will mean deserialization once and not re-serializing, which is the most performant outcome.

If any of the filter chain doesn’t support payload passthrough, a customized non passthrough filter to setup metadata is encouraged from point of performance view.

A typical use case for this filter is to dynamically match requests with load balancer subsets. For this, a given payload field’s value would be extracted and attached to the request as dynamic metadata which would then be used to match a subset of endpoints.

*Example*:

A sample filter configuration to route traffic to endpoints based on the presence or absence of a version payload could be:
``` yaml
thrift_filters:
- name: envoy.filters.thrift.payload_to_metadata
  typed_config:
    "@type": type.googleapis.com/envoy.extensions.filters.network.thrift_proxy.filters.payload_to_metadata.v3.PayloadToMetadata
    request_rules:
    - method_name: foo
      field_selector:
        name: info
        id: 2
        child:
          name: version
          id: 1
      on_present:
        metadata_namespace: envoy.lb
        key: version
      on_missing:
        metadata_namespace: envoy.lb
        key: default
        value: 'unknown'
```
The request thrift structure could be:
``` proto
message Request {
  message Info {
    string version = 1;
  }
  string data = 1;
  Info info = 2;
}
```
This would then allow requests of method name foo with the version payload field which is under info field set to be matched against endpoints with the corresponding version. Whereas requests with that payload missing would be matched with the default endpoints.

The regex matching and substitution is similiar with [header to metadata filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/other_protocols/thrift_filters/header_to_metadata_filter).

*Proposed Proto*

``` proto
message PayloadToMetadata {
  enum ValueType {
    STRING = 0;
[...]

PR #23409, opened 2022-10-08
Thrift: Payload to metadata filter

JuniorHsu:
Commit Message:
Add a thrift filter to support adding dynamic metadata by payload

Additional Description:
This filter is configured with request_rules that will be matched against requests. A field_selector of a rule represents the head of a linked list, each node of the linked list has a name for logging and an id for matching. The field_selector is tied to a payload field when the linked list corresponds to a downward path which rooted in the top-level of the request message structure. on_present is triggered when corresponding the payload is present. Otherwise, on_missing is triggered.

This filter is designed to support payload passthrough. By performing payload to metadata filter can do deserialization once, and pass the metadata to other filters. This means that load balancing decisions, consumed from log and routing could all use payload information with a single parse. Also notably performing the parsing in payload passthrough buffer will mean deserialization once and not re-serializing, which is the most performant outcome.

Risk Level: low
Testing: unit
Docs Changes: multiple rst
[Optional Fixes #Issue] #23322


JuniorHsu:
I address everything on top of my head except allowing multiple rules in a payload node.
https://github.com/envoyproxy/envoy/pull/23409/files#diff-7c9ea70a9bcf34c08c4213b100d808ad3c5462ba16624cbc37dace202cd87621R63

Will address this and do more testing on next biz days

JuniorHsu:
I think this is ready for the first round review.

cc @zuercher @rgs1 @fishcakez @PeterL328

api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
```diff
@@ -0,0 +1,105 @@
...
+      (validate.rules).int32 = {lte: 32767 gte: -32768},
+      (validate.rules).message = {required: true}
+    ];
+
+    // next node of the field selector
+    FieldSelector child = 3;
+  }
+
+  // The list of rules to apply to requests.
+  repeated Rule request_rules = 1 [(validate.rules).repeated = {min_items: 1}];
```
rgs1, line 104: is there any value in reusing some of these definitions from the existing header-to-metadata filter?

docs/root/configuration/other_protocols/thrift_filters/payload_to_metadata_filter.rst
```diff
@@ -0,0 +1,95 @@
...
+decisions, consumed from log and routing could all use payload information with a single parse.
+Also notably performing the parsing in payload passthrough buffer will mean deserialization once
+and not re-serializing, which is the most performant outcome.
+
+If any of the filter chain doesn't support payload passthrough, a customized non-passthrough
+filter to setup metadata is encouraged from point of performance view.
+
+A typical use case for this filter is to dynamically match requests with load balancer
+subsets. For this, a given payload field's value would be extracted and attached to the request
+as dynamic metadata which would then be used to match a subset of endpoints.
```
rgs1, line 56: Can you move this paragraph up to the start of the document? I think it's the first thing people would want to read. Also, could you add a sentence saying why would you need payload-to-metadata instead of header-to-metadata (I think the answer is if you are not using THeader)?

source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
```diff
@@ -0,0 +1,322 @@
...
+    finalizeDynamicMetadata();
+  }
+
+  return FilterStatus::Continue;
+}
+
+} // namespace PayloadToMetadataFilter
+} // namespace ThriftFilters
+} // namespace Extensions
+} // namespace Envoy
```
rgs1, line 322: is there any significant chunk of duplicate code in this file that could be shared with header-to-metadata? If so, let's move it to a common/ path (it can happen later too, and let's only do it if it's a non-trivial amount of lines).

test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter_test.cc
```diff
@@ -0,0 +1,1226 @@
...
+    proto->writeFieldBegin(msg, "third_field", FieldType::Struct, 3);
+
+    proto->writeStructBegin(msg, "payload");
+    proto->writeFieldBegin(msg, "f1", FieldType::Bool, 1);
+    proto->writeBool(msg, true);
+    proto->writeFieldEnd(msg);
+
+    proto->writeFieldBegin(msg, "f2", FieldType::Byte, 2);
+    proto->writeByte(msg, 2);
+    proto->writeFieldEnd(msg);
```
rgs1, line 106: nit: i think we have helpers for making this stuff shorter, iirc there's something like:

```
addField(msg, "f2", FieldType::Byte, 2);
```

that will reduce the boilerplate... if not, would be nice to have!

rgs1 (changes requested):
This looks good, just a few questions and asks -- thanks!

test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter_test.cc
JuniorHsu: Unfortunately we don't have a helper to the best of my knowledge.
The definition is here
https://github.com/envoyproxy/envoy/blob/9e4ddc314b78ecbad21645e0128e7e0a05964926/source/extensions/filters/network/thrift_proxy/protocol.h#L278

We used this pattern here and there.
Here's an example
https://github.com/envoyproxy/envoy/blob/9e4ddc314b78ecbad21645e0128e7e0a05964926/test/extensions/filters/network/thrift_proxy/conn_manager_test.cc#L263

I do agree with you that we want some helper but that needs some careful deliberation so we won't use the helper wrong. 
We want same experience among struct, container(map, list, set) and single fields so we can do something like
``` cpp
ProtocolHelper helper(proto, msg, "wrapper_struct_name"); // start the wrapper
helper.addInt64Field("first_field", 1, 1);
helper.addStringField("second_field", 2, "two");
helper.addStructField("third_field", 3, [&]() {
   helper.addBoolField("f1", 1, true);
   // ...
});
helper.addMapField("fourth_field", 4, FieldType::I32, {{1, 1}, {2,2});
helper.done(); //finish the wrapper

transport->encodeFrame(buffer, metadata, msg);
```

There we can fully hide `writeFieldBegin` otherwise it's so easy to mis-use the helper.
I would foreseen it takes some time on the lambda callback, testing and template programming on the container field.
If you agree, I can do this as a follow up and refactor the existing tests here and there.

docs/root/configuration/other_protocols/thrift_filters/payload_to_metadata_filter.rst
JuniorHsu: Sure, will do!
Missing THeader in framed transport is one reason. Another reason is that we can not trust the downstream service always copy the field to the THeader correctly, and we want single truth of source.

api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
JuniorHsu: Great question, I'm happy this is raised.
The most reusable part is `message KeyValuePair` and I did think through if we want to DIY or not.

I'm not sure how we handled api dependency between filters. Tag @mattklein123 since he's assigned as the api shepherd.

It's kinda weird to let p2m depends on h2m, since we need to build both filters if we only want to use p2m.
However extract `message KeyValuePair` to something like `common/key_value.proto` would break the existing usage. 
IMO the `KeyValuePair` per se is not that big, so it might be okay-ish to dup that in two filters.
If we really want to DIY in proto, we might want to keep both `KeyValuePair` in `h2m.proto` and  ``common/key_value.proto`` and mark `KeyValuePair` in h2m as deprecated, maintain two code paths to make both work and remove one path in some day. It apparently needs a great amount of care. 

That's the reason I dup `KeyValuePair` in p2m and h2m. Does it make sense?

source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
JuniorHsu: I put a TODO item here
https://github.com/envoyproxy/envoy/pull/23409/files#diff-7c9ea70a9bcf34c08c4213b100d808ad3c5462ba16624cbc37dace202cd87621R172

The most apparent part to extract is `applyKeyValue` and `addMetadata` 

The reasons I did not reuse are:
(a) the implementation replies on the proto so it would need either template or some additional proto->enum mapping
https://github.com/envoyproxy/envoy/blob/6bba2d69a17e4cf374eb09f7f51fbf0f0ce4b7da/source/extensions/filters/network/thrift_proxy/filters/header_to_metadata/header_to_metadata_filter.cc#L100

and 
https://github.com/envoyproxy/envoy/blob/6bba2d69a17e4cf374eb09f7f51fbf0f0ce4b7da/source/extensions/filters/network/thrift_proxy/filters/header_to_metadata/header_to_metadata_filter.cc#L140

where `KeyValuePair` is tied to proto. 

(b) h2m `setDynamicMetadata` once but p2m does it while we parse the tree.

It's definitely doable but the refactor is non-trivial.
Hence, I'd like to do the refactor once the filter is stable in proto and production.

rgs1 (dismissed):
LGTM, thanks!

Over to @zuercher for merging (after you've merged main).

changelogs/current.yaml
```diff
@@ -17,6 +17,48 @@ removed_config_or_runtime:
...
+    added stats for downstream connection close to detect SR drop.
+- area: thrift
+  change: |
+    added payload to metadata filter which matches a given payload field's value would be extracted and attached to the request as dynamic metadata.
+- area: compression
+  change: |
+    added support for :ref:`choose_first<envoy_v3_api_msg_extensions.filters.http.compressor.v3.Compressor>`.
+- area: cors
+  change: |
+    added support for cors PNA. This behavioral change can be temporarily reverted by setting runtime guard ``envoy_reloadable_features_cors_private_network_access`` to false. More details refer to https://developer.chrome.com/blog/private-network-access-preflight.
```
zuercher, line 61: Merge issue?

api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
```diff
@@ -0,0 +1,105 @@
...
+    // of the missing field value.
+    KeyValuePair on_missing = 5;
+  }
+
+  message FieldSelector {
+    // field name to log
+    string name = 1 [(validate.rules).string = {min_len: 1}];
+
+    // field id to match. Limit to the range of int16 to fit thrift format.
+    google.protobuf.Int32Value id = 2 [
```
zuercher, line 94: Can this just be an int32? It's a little easier to deal with given that it's required anyway.

source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
```diff
@@ -0,0 +1,322 @@
...
+      method_or_service_name_ = rule_.service_name() + ":";
+    } else {
+      method_or_service_name_ = rule_.service_name();
+    }
+    break;
+  case ProtoRule::MatchSpecifierCase::MATCH_SPECIFIER_NOT_SET:
+    PANIC_DUE_TO_CORRUPT_ENUM;
+  }
+
+  FieldSelector field_selector = rule_.field_selector();
```
zuercher, line 65: minor: I think we can make this a `FieldSelector*` or `FieldSelector&` and avoid Swap and the need to explain it later.

source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
```diff
@@ -0,0 +1,322 @@
...
+  static const std::string& payloadToMetadata = "envoy.filters.thrift.payload_to_metadata";
+  return nspace.empty() ? payloadToMetadata : nspace;
+}
+
+bool PayloadToMetadataFilter::addMetadata(const std::string& meta_namespace, const std::string& key,
+                                          std::string value, ValueType type) {
+  ProtobufWkt::Value val;
+  ASSERT(!value.empty());
+
+  if (value.size() >= MAX_PAYLOAD_VALUE_LEN) {
```
zuercher, line 221: I couldn't find this documented anywhere, and I think it should be. Also consider making the the log message error-level instead of debug.

changelogs/current.yaml
JuniorHsu: yeah. thanks

api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
JuniorHsu: we want `int16` but if we don't set that is required for `int16`. IIRC it would be default to 0 without setting which is error-prone.

source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
JuniorHsu: pointer works. thanks for great suggestion

source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
JuniorHsu: thanks! i doc it and make the log error

api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
zuercher: But the validations make it required so it’s an error either way.

zuercher (dismissed):
I’ll leave the discussion about the google.protobuf.Int32 vs int32 for the api shepherds.

api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
lizan: Yeah this doesn't have to be Int32Value but could just be int16 since it is required anyway.

JuniorHsu: I did a quick test and it doesn't raise `ProtoValidationException` or any other exception. 
Am i missing anything, @lizan ? But yeah i can switch to int32 if that's preferred.

```
[ RUN      ] PayloadToMetadataFilterConfigTest.SimpleConfigWithoutId
[       OK ] PayloadToMetadataFilterConfigTest.SimpleConfigWithoutId (105 ms)
``` 


```
diff --git a/api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto b/api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
index 0148a22626..894df76e1d 100644
--- a/api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
+++ b/api/envoy/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/v3/payload_to_metadata.proto
@@ -4,8 +4,6 @@ package envoy.extensions.filters.network.thrift_proxy.filters.payload_to_metadat
 
 import "envoy/type/matcher/v3/regex.proto";
 
-import "google/protobuf/wrappers.proto";
-
 import "udpa/annotations/status.proto";
 import "validate/validate.proto";
 
@@ -90,11 +88,8 @@ message PayloadToMetadata {
     // field name to log
     string name = 1 [(validate.rules).string = {min_len: 1}];
 
-    // field id to match. Limit to the range of int16 to fit thrift format.
-    google.protobuf.Int32Value id = 2 [
-      (validate.rules).int32 = {lte: 32767 gte: -32768},
-      (validate.rules).message = {required: true}
-    ];
+    // field id to match
+    int32 id = 2 [(validate.rules).int32 = {lte: 32767 gte: -32768}];
 
     // next node of the field selector
     FieldSelector child = 3;
diff --git a/source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc b/source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
index c90911a572..5476eb5720 100644
--- a/source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
+++ b/source/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter.cc
@@ -65,7 +65,7 @@ Rule::Rule(const ProtoRule& rule, uint16_t rule_id, TrieSharedPtr root)
   const FieldSelector* field_selector = &rule_.field_selector();
   TrieSharedPtr node = root;
   while (true) {
-    int16_t id = static_cast<int16_t>(field_selector->id().value());
+    int16_t id = static_cast<int16_t>(field_selector->id());
     if (node->children_.find(id) == node->children_.end()) {
       node->children_[id] = std::make_shared<Trie>(node);
     }
diff --git a/test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/config_test.cc b/test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/config_test.cc
index 8dacdf50e2..f112c21d92 100644
--- a/test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/config_test.cc
+++ b/test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/config_test.cc
@@ -183,6 +183,25 @@ void testForbiddenConfig(const std::string& yaml, const std::string& message) {
                             EnvoyException, message);
 }
 
+// does not allow rule without id in field selector
+TEST(PayloadToMetadataFilterConfigTest, SimpleConfigWithoutId) {
+  const std::string yaml = R"EOF(
+request_rules:
+  - method_name: foo
+    field_selector:
+      name: info
+    on_present:
+      metadata_namespace: envoy.lb
+      key: version
+    on_missing:
+      metadata_namespace: envoy.lb
+      key: default
+      value: unknown
+  )EOF";
+
+  testValidConfig(yaml);
+}
+
 // does not allow rule without either on_present or on_missing
 TEST(PayloadToMetadataFilterConfigTest, EmptyOnPresentOnMissing) {
   std::string yaml =

```

zuercher (approved):

JuniorHsu:
@lizan do you mind to take a look at api part again? Thanks.

lizan (approved):

PR #23415, opened 2022-10-09
api: add socket_options for multiple addresses

soulxu:

Commit Message: api: add socket_options for multiple addresses
Additional Description:
Since the ipv4 and ipv6 have different socket option flags, when using multiple addresses, the user has to specify different socket options for the ipv4 address and the ipv6 address.

For the listener, the additional address can be the Ipv6 address, then it should be able to set an Ipv6 flag corresponding to the ipv4 one. Add socket_option field for each additional address.

For the upstream, the endpoint can be ipv4 or ipv6, currently, the user can specify the ipv4 and ipv6 local bind address in the bind config, but there is only a global socket_options that apply to both the ipv4 and ipv6 addresses. Add socket_options for each extra source address. https://envoyproxy.slack.com/archives/C78HA81DH/p1664228598624269

Risk Level: low
Testing: n/a
Docs Changes: API doc
Release Notes: n/a

### DIFF
--- a/api/envoy/config/core/v3/address.proto
+++ b/api/envoy/config/core/v3/address.proto
@@ -119,6 +119,16 @@ message TcpKeepalive {
 message ExtraSourceAddress {
   // The additional address to bind.
   SocketAddress address = 1 [(validate.rules).message = {required: true}];
+
+  // [#not-implemented-hide:]
+  oneof socket_options_specifier {
+    // [#not-implemented-hide:]
+    // Additional socket options that may not be present in Envoy source code or
+    // precompiled binaries. If no socket options specified, the BindConfig's
+    // :ref:`socket_options <envoy_v3_api_field_config.core.v3.BindConfig.socket_options>`
+    // will apply to this address.
+    repeated core.SocketOption socket_options = 2;
+  }
 }
 
 // [#next-free-field: 6]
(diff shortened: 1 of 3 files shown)


soulxu:
cc @mattklein123 @ggreenway @ravenblackx

api/envoy/config/core/v3/address.proto
```diff
@@ -119,6 +119,19 @@ message TcpKeepalive {
 message ExtraSourceAddress {
   // The additional address to bind.
   SocketAddress address = 1 [(validate.rules).message = {required: true}];
+
+  // [#not-implemented-hide:]
+  // Additional socket options that may not be present in Envoy source code or
+  // precompiled binaries. If no socket options specified, the BindConfig's
+  // :ref:`socket_options <envoy_v3_api_field_config.core.v3.BindConfig.socket_options>`
+  // will apply to this address.
+  repeated SocketOption socket_options = 2;
```
lizan, line 128: Would prefer extract this to a message as `socket_options_override`, then you don't need `no_socket_options`. If the field is `null` means no override, non-null with empty means no socket options will be applied.

soulxu: got it, thanks! that is good idea

lizan (approved):

PR #23419, opened 2022-10-10
compressor benchmark: add brotli

giantcroc:
Signed-off-by: giantcroc [changran.wang@intel.com](mailto:changran.wang@intel.com)

Commit Message: add brotli to fulfill compressor benchmark 

```
Running ./bazel-bin/test/extensions/filters/http/compressor/compressor_filter_speed_test
Run on (8 X 4061.8 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 8192 KiB (x1)
Load Average: 0.51, 0.54, 0.55
---------------------------------------------------------------------------------------
Benchmark                                             Time             CPU   Iterations
---------------------------------------------------------------------------------------
compressFullWithGzip/0/manual_time              5.96 ms         6.01 ms          117
compressFullWithGzip/1/manual_time              2.54 ms         2.59 ms          276
compressFullWithGzip/2/manual_time              1.75 ms         1.80 ms          400
compressFullWithGzip/3/manual_time              6.40 ms         6.45 ms          105
compressFullWithGzip/4/manual_time              4.25 ms         4.30 ms          164
compressFullWithGzip/5/manual_time              8.34 ms         8.39 ms           82
compressFullWithGzip/6/manual_time              6.45 ms         6.50 ms          105
compressFullWithGzip/7/manual_time              4.26 ms         4.31 ms          163
compressFullWithGzip/8/manual_time              8.32 ms         8.37 ms           82
compressChunks16384WithGzip/0/manual_time       5.93 ms         5.99 ms          119
compressChunks16384WithGzip/1/manual_time       2.40 ms         2.45 ms          293
compressChunks16384WithGzip/2/manual_time       1.69 ms         1.74 ms          412
compressChunks16384WithGzip/3/manual_time       6.00 ms         6.05 ms          112
compressChunks16384WithGzip/4/manual_time       4.04 ms         4.08 ms          175
compressChunks16384WithGzip/5/manual_time       8.20 ms         8.25 ms           84
compressChunks16384WithGzip/6/manual_time       6.14 ms         6.19 ms          110
compressChunks16384WithGzip/7/manual_time       4.21 ms         4.27 ms          174
compressChunks16384WithGzip/8/manual_time       7.89 ms         7.94 ms           85
compressChunks8192WithGzip/0/manual_time        6.03 ms         6.08 ms          109
compressChunks8192WithGzip/1/manual_time        2.57 ms         2.62 ms          271
compressChunks8192WithGzip/2/manual_time        1.84 ms         1.90 ms          376
compressChunks8192WithGzip/3/manual_time        6.53 ms         6.59 ms          104
compressChunks8192WithGzip/4/manual_time        4.31 ms         4.36 ms          162
compressChunks8192WithGzip/5/manual_time        8.43 ms         8.49 ms           81
compressChunks8192WithGzip/6/manual_time        6.45 ms         6.50 ms          103
compressChunks8192WithGzip/7/manual_time        4.31 ms         4.36 ms          162
compressChunks8192WithGzip/8/manual_time        8.50 ms         8.56 ms           80
...
compressFullWithBrotli/0/manual_time              0.328 ms        0.372 ms         2148
compressFullWithBrotli/1/manual_time               1.27 ms         1.32 ms          539
compressFullWithBrotli/2/manual_time               1.64 ms         1.69 ms          424
compressFullWithBrotli/3/manual_time               2.60 ms         2.65 ms          266
compressFullWithBrotli/4/manual_time               5.48 ms         5.53 ms          121
compressFullWithBrotli/5/manual_time               5.64 ms         5.69 ms          119
compressFullWithBrotli/6/manual_time               6.68 ms         6.74 ms           99
compressFullWithBrotli/7/manual_time               7.23 ms         7.29 ms           92
compressFullWithBrotli/8/manual_time               9.61 ms         9.68 ms           65
compressFullWithBrotli/9/manual_time               58.6 ms         58.7 ms           11
compressFullWithBrotli/10/manual_time               106 ms          106 ms            7
compressChunks16384WithBrotli/0/manual_time       0.761 ms        0.810 ms          907
compressChunks16384WithBrotli/1/manual_time        1.20 ms         1.24 ms          580
compressChunks16384WithBrotli/2/manual_time        1.56 ms         1.61 ms          443
compressChunks16384WithBrotli/3/manual_time        2.59 ms         2.64 ms          272
compressChunks16384WithBrotli/4/manual_time        5.30 ms         5.35 ms          127
compressChunks16384WithBrotli/5/manual_time        5.41 ms         5.46 ms          125
compressChunks16384WithBrotli/6/manual_time        6.49 ms         6.56 ms           94
compressChunks16384WithBrotli/7/manual_time        6.92 ms         6.98 ms           87
compressChunks16384WithBrotli/8/manual_time        9.06 ms         9.12 ms           77
compressChunks16384WithBrotli/9/manual_time        43.8 ms         43.9 ms           17
compressChunks16384WithBrotli/10/manual_time       70.1 ms         70.2 ms           10
compressChunks8192WithBrotli/0/manual_time        0.732 ms        0.782 ms          955
compressChunks8192WithBrotli/1/manual_time         1.35 ms         1.40 ms          511
compressChunks8192WithBrotli/2/manual_time         1.75 ms         1.80 ms          397
compressChunks8192WithBrotli/3/manual_time         2.92 ms         2.98 ms          236
compressChunks8192WithBrotli/4/manual_time         5.90 ms         5.96 ms          106
compressChunks8192WithBrotli/5/manual_time         6.02 ms         6.08 ms          112
compressChunks8192WithBrotli/6/manual_time         7.04 ms         7.10 ms           95
compressChunks8192WithBrotli/7/manual_time         7.66 ms         7.73 ms           79
compressChunks8192WithBrotli/8/manual_time         9.91 ms         9.98 ms           58
compressChunks8192WithBrotli/9/manual_time         42.3 ms         42.4 ms           17
compressChunks8192WithBrotli/10/manual_time        73.1 ms         73.2 ms            9
...
```


test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -111,11 +140,50 @@ CompressorFilterConfigSharedPtr makeZstdConfig(Stats::IsolatedStoreImpl& stats,
...
+  const auto quality = std::get<0>(params);
+  Envoy::Compression::Compressor::CompressorFactoryPtr compressor_factory =
+      std::make_unique<MockBrotliCompressorFactory>(quality);
+  CompressorFilterConfigSharedPtr config = std::make_shared<CompressorFilterConfig>(
+      compressor, "test.", stats, runtime, std::move(compressor_factory));
+
+  return config;
+}
+
+static uint64_t TestDataSize = 122880;
```
daixiang0, line 158: still need `constexpr`

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -111,11 +140,50 @@ CompressorFilterConfigSharedPtr makeZstdConfig(Stats::IsolatedStoreImpl& stats,
...
 
 Buffer::OwnedImpl generateTestData() {
   Buffer::OwnedImpl data;
-  TestUtility::feedBufferWithRandomCharacters(data, TestDataSize);
+  std::string filename = std::to_string(TestDataSize) + ".data";
```
daixiang0, line 162: When do we delete this file?

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -520,6 +593,120 @@ BENCHMARK(compressChunks1024WithZstd)
     ->UseManualTime()
     ->Unit(benchmark::kMillisecond);
 
+static std::vector<CompressionParams> brotli_compression_params = {
```
daixiang0, line 596: Is there a better way to generate it?

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: This file is used to make sure that we can use same data when run multiple times. So the deletion of this file is decided by users.

daixiang0: Benchmark is still a test, all files generated during tests should be deleted.

Besides, why can not get the same data with a fixed data size?

giantcroc: Sorry, I didn't notice that `feedBufferWithRandomCharacters` is implemented with fixed random seed, and we don't need this file.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: I think it's more intuitive to list all params used and also convient to modify each param.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -71,6 +71,33 @@ class MockZstdCompressorFactory : public Envoy::Compression::Compressor::Compres
...
+
+  const std::string& statsPrefix() const override {
+    CONSTRUCT_ON_FIRST_USE(std::string, "brotli.");
+  }
+  const std::string& contentEncoding() const override {
+    return Http::CustomHeaders::get().ContentEncodingValues.Brotli;
+  }
+
+private:
+  const uint64_t chunk_size_{4096};
```
jmarantz, line 92: would it make sense to parameterize your performance tests over some of these values? chunk size in particular?

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
jmarantz: no non-pod statics. You can use a std::array or C array here and specify it as constexpr. I'm not sure if your tuple implementation would work for CompressionParams in that case; you might need to make it a struct (which IMO would be more readable anyway :)

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -520,6 +567,120 @@ BENCHMARK(compressChunks1024WithZstd)
...
+
+    // level10 + default
+    {10, 0, 0, 0},
+
+    // level11 + default
+    {11, 0, 0, 0}};
+
+// NOLINTNEXTLINE(readability-identifier-naming)
+static void compressFullWithBrotli(benchmark::State& state) {
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> decoder_callbacks;
```
jmarantz, line 606: this is pre-existing but I've seen in the past that gmock structures are not built for speed, and may be a bottleneck in a benchmark.

jmarantz (commented):
lgtm basically with minor nits

Can you put the benchmark results in the PR description? Thank you!

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: OK, I have changed tuple to struct, but find that `constexpr` can't work here, so I change it to `const`. Thanks!

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: Yeah, it may be a issue, but I don't have any other idea to address it for now. And we more care about relative values of different compression algorithms in our work.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -520,6 +575,120 @@ BENCHMARK(compressChunks1024WithZstd)
     ->UseManualTime()
     ->Unit(benchmark::kMillisecond);
 
+static const std::array<CompressionParams, 11> brotli_compression_params = {
```
jmarantz, line 578: const is not sufficient to avoid this being initialized like a non-pod. You need to use constexpr. I might have been wrong about std::array being an appropriate choice; can you just use a C array like
`static constexpr CompressionParams BrotliCompressionParams[] = ....` 

otherwise you can use CONSTRUCT_ON_FIRST_USE to build complex static const objects.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
jmarantz: you can simply use the real objects. See other *_speed_test.cc around the hierarchy for examples. That's the preferred solution.

But at a minimum you could get some performance data (eg flame graphs) to see whether you are really just measuring the speed of the slow mocks. When you do this, please use both optimization *and* debug so you get accurate symbols in your flame graphs.

I don't think you can learn enough about relative speed values if the time is dominated by mocks. I'm not sure it is, but I can't really trust benchmark data unless we get some better view.

jmarantz (commented):
you need to run your benchmarks with a "-c opt" build.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: 2022-10-13 15-35-56 的屏幕截图
I have got the result of flamegraph, it seems `encodeData` takes up more than 80% of the time, and `Mock` is about 7% totally.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: Brotli has its default values for params in `config.cc`, but they are only able to used in this file.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
jmarantz: Did you take that flame-graph from an optimized build? Actually I'd recommend a build with debug *and* optimization: `--compilation_mode=opt --cxxopt=-g --cxxopt=-ggdb3`.

But let's take that 7% overhead and do some math. Some say (and I think that used to be true) that Brotli is 10x slower than gzip.

So if Brotli takes 93 seconds to do an operation, with 7 seconds for mocks, and Gzip takes 9.3 seconds, with 7 seconds per mocks, your perf data will look like the delta is 100/(9.3+7) = 6x slower when it's really 10x slower.

That's why I think it's worth getting rid of the mock overhead by using real objects.

I also think it's OK for benchmarks to skip the filtering infrastructure and just call the common compressor API, which might reduce the number of objects you need to instantiate.

giantcroc: I rebuild with `--compilation_mode=opt --cxxopt=-g --cxxopt=-ggdb3`, it looks that the percentage of `encodeData` in case of gzip and brotli is increased up to 96%, and `Mock` is down to about 1.5%.
2022-10-14 10-35-24 的屏幕截图
2022-10-14 10-38-21的屏幕截图

And the percentage of `encodeData` in case of zstd is increased up to 92%, and `Mock` is down to about 3%.
2022-10-14 10-36-15 的屏幕截图

I think that the time of `Mock` is related to the total time of benchamrk, and its percentage is roughly fixed in one benchmark.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: I have tried to use `static constexpr CompressionParams BrotliCompressionParams[] = ....`,
but there is an error says 
```
error: the type 'const Envoy::Extensions::HttpFilters::Compressor::CompressionParams []' of 'constexpr' variable 'Envoy::Extensions::HttpFilters::Compressor::gzip_compression_params' is not literal
```
I guess that it is caused by struct.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: OK, thanks!

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
jmarantz: I think that's cause you have a constructor, which you don't need. This compiles fine:
```
struct Foo {
  int a;
  int b;
  int c;
  int d;
};

static constexpr Foo foo_array[] = {
  {0, 1, 2, 3},
  {4, 5, 6, 7}
};
```

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
jmarantz: OK I think this is acceptable but please comment that the gmock overhead has been measured with flame graphs to be pretty low.

I think the reason it was 7% earlier was that optimization affects different code-paths differently, and in particular heavy use of STL really needs the optimizer to work fast, whereas more C-like code (e.g. found in zlib or brotli) may is less reliant on optimization. When you compiled with some optimization you reduced the relative overhead of STL-heavy mocks. That's my theory anyway.

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: OK, it works, thanks!

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: OK, I have added comments. 
I learned a lot from your comments. 
Thank you very much! :)

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
jmarantz: So this still needs to be resolved; the 4096 chunk size constant should only be in one place. Same with the other constants below (24, and 18).

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -71,7 +71,39 @@ class MockZstdCompressorFactory : public Envoy::Compression::Compressor::Compres
   const uint64_t chunk_size_{4096};
 };
 
-using CompressionParams = std::tuple<int64_t, uint64_t, int64_t, uint64_t>;
+class MockBrotliCompressorFactory : public Envoy::Compression::Compressor::CompressorFactory {
+public:
+  MockBrotliCompressorFactory(uint32_t quality) : quality_(quality) {}
```
jmarantz, line 76: explicit

jmarantz (commented):
lgtm modulo the config constants. thank you!

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -111,6 +142,21 @@ CompressorFilterConfigSharedPtr makeZstdConfig(Stats::IsolatedStoreImpl& stats,
   return config;
 }
 
+CompressorFilterConfigSharedPtr makeBrotliConfig(Stats::IsolatedStoreImpl& stats,
+                                                 testing::NiceMock<Runtime::MockLoader>& runtime,
+                                                 CompressionParams params) {
```
jmarantz, line 147: pass by const ref

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: OK, I have changed to use default values in config for gzip and brotli, but there is no default value for zstd. Do you think if I can add some default values in zstd's config?

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
giantcroc: OK, fixed, thanks!

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -45,12 +47,11 @@ class MockGzipCompressorFactory : public Envoy::Compression::Compressor::Compres
   const Compression::Gzip::Compressor::ZlibCompressorImpl::CompressionStrategy strategy_;
   const int64_t window_bits_;
   const uint64_t memory_level_;
-  const uint64_t chunk_size_{4096};
 };
 
 class MockZstdCompressorFactory : public Envoy::Compression::Compressor::CompressorFactory {
 public:
-  MockZstdCompressorFactory(uint32_t level, uint32_t strategy)
+  explicit MockZstdCompressorFactory(uint32_t level, uint32_t strategy)
```
jmarantz, line 54: rm `explicit`

test/extensions/filters/http/compressor/compressor_filter_speed_test.cc
```diff
@@ -18,19 +21,18 @@ namespace Envoy {
 namespace Extensions {
 namespace HttpFilters {
 namespace Compressor {
-
 class MockGzipCompressorFactory : public Envoy::Compression::Compressor::CompressorFactory {
 public:
-  MockGzipCompressorFactory(
+  explicit MockGzipCompressorFactory(
```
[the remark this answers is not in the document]
giantcroc: OK, fixed, thanks!

lizan (approved):

PR #23341, opened 2022-10-03
move http HeaderString to common code tree for common usage

wbpcode:
Commit Message: move http HeaderString to common code tree for common usage
Additional Description:

See #23327 for background.

Risk Level: Low.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.

PR #23427, opened 2022-10-10
http: optimize the UnionString append validation

botengyao:

#23341 refactored http HeaderString to common usage before. This PR is going to add back the optimization to only valid the input (n) rather than the whole appended string buffer each time (n^2). Also added one timeout fuzzy corpus.

Commit Message:  optimize the UnionString validation in append method
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


envoy/common/union_string.h
```diff
@@ -256,6 +257,8 @@ template <class Validator> class UnionStringBase {
 
   bool valid() const { return Validator()(getStringView()); }
 
+  bool validInput(absl::string_view data) const { return Validator()(data); }
+
```
wbpcode, line 261: or just 
```
bool valid(absl::optional<absl::string_view> view) { 
  return view.has_value() ? Validator()(view.value()) : Validator()(getStringView());
}
```

botengyao: Good point, done.

wbpcode (approved):
Thanks. LGTM.

PR #23430, opened 2022-10-11
ci: accelerate local build

daixiang0:

Create a TAR waste lots of time, skip it if not release.

Before this:
```
Copying binary for image build...

real    0m32.585s
user    0m30.830s
sys     0m2.557s
```

After this:
```
Copying binary for image build...

real    0m1.912s
user    0m0.806s
sys     0m1.105s
```

Save 30s in total.

Commit Message:
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -91,8 +91,8 @@ function cp_binary_for_image_build() {
   mkdir -p "${BASE_TARGET_DIR}"/"${TARGET_DIR}"_stripped
   strip "${FINAL_DELIVERY_DIR}"/envoy -o "${BASE_TARGET_DIR}"/"${TARGET_DIR}"_stripped/envoy
 
-  # Copy for azp which doesn't preserve permissions, creating a tar archive
-  tar czf "${ENVOY_BUILD_DIR}"/"${EXE_NAME}"_binary.tar.gz -C "${BASE_TARGET_DIR}" "${TARGET_DIR}" "${TARGET_DIR}"_stripped
+  # Copy for azp which doesn't preserve permissions, creating a tar archive only when release
+  [[ "${BINARY_TYPE}" == "release" ]] && tar czf "${ENVOY_BUILD_DIR}"/"${EXE_NAME}"_binary.tar.gz -C "${BASE_TARGET_DIR}" "${TARGET_DIR}" "${TARGET_DIR}"_stripped
 
   # Remove binaries to save space, only if BUILD_REASON exists (running in AZP)
   [[ -z "${BUILD_REASON}" ]] || \
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -91,13 +91,15 @@ function cp_binary_for_image_build() {
   mkdir -p "${BASE_TARGET_DIR}"/"${TARGET_DIR}"_stripped
   strip "${FINAL_DELIVERY_DIR}"/envoy -o "${BASE_TARGET_DIR}"/"${TARGET_DIR}"_stripped/envoy
 
-  # Copy for azp which doesn't preserve permissions, creating a tar archive only when release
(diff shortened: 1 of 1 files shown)


phlax:
@daixiang0 i think the problem with this is that if you are using RBE it slows things down - im wondering if this can be done only for local builds (this also affects eg docs build which does lots of un/tarring for similar reasons)

daixiang0:
I do not disable RBE, only skip tar create step. Since the tar is useless local, it is helpful.

phlax:
i think this is a good idea but would be good to confirm with @lizan 

not sure exactly when this is/not called as a "release" binary and wondering if we just want to prevent the tarballing for non-azp runs - also not sure if anyone is expecting this output

im thinking lets wait till after the 1.24 release before landing

daixiang0:
@phlax  Actually, I find that we only use this TAR when "release" during CI, so if there is a ENV that can mark this case, we can use it directly.

BTW, what is the last day of 1.24 code freeze?

phlax:
> BTW, what is the last day of 1.24 code freeze?

release is expected early next week - so we are in a partial code freeze now

daixiang0:
@phlax could we consider it?

phlax:
> @phlax could we consider it?

yep, apologies for delay - i will look over more thoroughly tomorrow

KBaichoo:
hey @phlax , any other thoughts on this?

phlax:
apologies - i just didnt get around to checking further - ill check shortly ...

phlax:
azp seems to be having issues atm so im unable to check further - ill follow up as soon as im able to get back in and see jobs/artefacts etc

ci/do_ci.sh
```diff
@@ -91,8 +91,8 @@ function cp_binary_for_image_build() {
   mkdir -p "${BASE_TARGET_DIR}"/"${TARGET_DIR}"_stripped
   strip "${FINAL_DELIVERY_DIR}"/envoy -o "${BASE_TARGET_DIR}"/"${TARGET_DIR}"_stripped/envoy
 
-  # Copy for azp which doesn't preserve permissions, creating a tar archive
-  tar czf "${ENVOY_BUILD_DIR}"/"${EXE_NAME}"_binary.tar.gz -C "${BASE_TARGET_DIR}" "${TARGET_DIR}" "${TARGET_DIR}"_stripped
+  # Copy for azp which doesn't preserve permissions, creating a tar archive only when release
+  [[ "${BINARY_TYPE}" == "release" ]] && tar czf "${ENVOY_BUILD_DIR}"/"${EXE_NAME}"_binary.tar.gz -C "${BASE_TARGET_DIR}" "${TARGET_DIR}" "${TARGET_DIR}"_stripped
```
phlax, line 95: looking through the code more thoroughly - if we only want to create the tarball in azp then it seems like the next condition below - ie `[[ -z "${BUILD_REASON}" ]]` can be used

how about just moving the tar line into that condition ? (perhaps adding `if .. fi` to make it more readable)

phlax (commented):
apologies if i was cautious landing this - the code there is not easy to follow and its not entirely clear whether this will break anyones expectations

on balance i think the change is good and if it does break anything it will be an easy fix or we can revert - so im happy to add this altho not sure how is best

ci/do_ci.sh
daixiang0: Sure.

phlax: i dont see the change

/wait

daixiang0: sorry for push issue. now you can see it.

phlax (approved):
lgtm, thanks @daixiang0 

apologies again for slow review

daixiang0:
@phlax thank you.

PR #23432, opened 2022-10-11
upstream: new healthStatus method to provide more detailed host health status

### LINKED ISSUE #23213: override_host_set conflates different EDS health statuses
This is something we discovered while in the process of implementing Envoy's stateful session affinity feature (see #16698) in gRPC.

From reading the code, it appears that there's a bug in Envoy's handling of the [`override_host_status` field](https://github.com/envoyproxy/envoy/blob/8e5c602e3cbafa3787269e94fe7fd0a81e62ca81/api/envoy/config/cluster/v3/cluster.proto#L672).  The field allows setting individual health statuses for which the host-override behavior should be honored.  However, internally, Envoy does not actually track all of the health statuses individually; in particular, it treats all of `DRAINING`, `UNHEALTHY`, and `TIMEOUT` as being exactly the same thing, and it treats `UNKNOWN` and `HEALTHY` as being the same thing (see [code](https://github.com/envoyproxy/envoy/blob/ea0eb2d5ea3f82250f3577c7ac5a898087d0799f/source/common/upstream/load_balancer_impl.cc#L529)).  So for example, if someone sets the `override_host_status` set to include `DRAINING` but not `UNHEALTHY` or `TIMEOUT`, Envoy will actually honor the host-override for all three of those states.  I think this is a bug; it seems very clear from the API what is actually intended here.

I think the underlying problem here is that Envoy does not internally record the actual EDS health status in the `Host` object; instead, it does a lossy conversion to `Host::Health`.  I suspect that the right fix is to have the `Host` object record the original EDS health status with full fidelity and then look at that when evaluating the `override_host_status` set.

CC @wbpcode @htuch @yanavlasov @sanjaypujare

wbpcode:

Commit Message: upstream: new healthStatus method to provide more detailed host health status
Additional Description:

Part of work to fix #23213.

Because the canonical `health()` method is used widely and is enough for most of cases. We can treat it as abbreviated status. So, we won't update the `health()` directly.

This PR added a new `healthStatus()` method to the host to provide more specific health status (with type of EDS `envoy::config::core::v3::HealthStatus`). 

**This status is hybrid of EDS status and runtime active status (from active health checker or outlier detection). Runtime active status will be taken as a priority.**

This is simplest implementation for this feature. 

Risk Level: none. not change any exist code but add a new method.
Testing: unit.
Docs Changes: n/a.
Release Notes: n/a.

### DIFF
--- a/source/common/conn_pool/conn_pool_base.cc
+++ b/source/common/conn_pool/conn_pool_base.cc
@@ -77,7 +77,7 @@ bool ConnPoolImplBase::shouldCreateNewConnection(float global_preconnect_ratio) 
   // upstream selection logic may result in bypassing this upstream entirely.
   // If an Envoy user wants preconnecting for degraded upstreams this could be
   // added later via extending the preconnect config.
-  if (host_->health() != Upstream::Host::Health::Healthy) {
+  if (host_->coarseHealth() != Upstream::Host::Health::Healthy) {
     return pending_streams_.size() > connecting_stream_capacity_;
   }
 
--- a/source/common/upstream/upstream_impl.h
+++ b/source/common/upstream/upstream_impl.h
@@ -233,8 +233,8 @@ public:
       : HostDescriptionImpl(cluster, hostname, address, metadata, locality, health_check_config,
                             priority, time_source),
         health_status_(health_status) {
-    // This EDS flags setting is still necessary for stats, configuration dump, canonical health()
-    // etc.
+    // This EDS flags setting is still necessary for stats, configuration dump, canonical
+    // coarseHealth() etc.
     setEdsHealthFlag(health_status);
     HostImpl::weight(initial_weight);
   }
@@ -284,7 +284,7 @@ public:
     return health_status_;
   }
 
-  Host::Health health() const override {
+  Host::Health coarseHealth() const override {
     // If any of the unhealthy flags are set, host is unhealthy.
     if (healthFlagGet(HealthFlag::FAILED_ACTIVE_HC) ||
         healthFlagGet(HealthFlag::FAILED_OUTLIER_CHECK) ||
--- a/source/common/upstream/upstream_impl.h
+++ b/source/common/upstream/upstream_impl.h
@@ -231,7 +231,10 @@ public:
            uint32_t priority, const envoy::config::core::v3::HealthStatus health_status,
            TimeSource& time_source)
       : HostDescriptionImpl(cluster, hostname, address, metadata, locality, health_check_config,
-                            priority, time_source) {
+                            priority, time_source),
+        health_status_(health_status) {
+    // This EDS flags setting is still necessary for stats, configuration dump, canonical health()
+    // etc.
     setEdsHealthFlag(health_status);
     HostImpl::weight(initial_weight);
   }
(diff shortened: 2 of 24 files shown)


wbpcode:
I think this new method interface should be OK. But the inner implementation could be re-evaluated again.
I want to refactor the `HealthFlag` to store all EDS status and dynamic runtime flags. Then we can get all detailed status by the `healthFlagGet()`. But it would be some more code changes. So i'd defer it to future PR, after I do more validation. 
Some TODOs are added in the code tree.

source/common/upstream/upstream_impl.h
```diff
@@ -305,6 +325,10 @@ class HostImpl : public HostDescriptionImpl,
 
   std::atomic<uint32_t> health_flags_{};
   std::atomic<uint32_t> weight_;
+  // TODO(wbpcode): should we store the EDS health status to health_flags_ to get unified status or
```
markdroth, line 328: I think it might ultimately be cleaner to just eliminate `Host::Health` and instead expose only the `HealthStatus` value, which would represent the final state after all inputs are applied (EDS health status, active health checking, etc).

htuch: I think this is fair, but maybe use better names, since `health` and `healthStatus` are both seemingly the same thing. Maybe rename the existing one to `coarseHealth` or something with similar intent (naming is hard).

wbpcode: naming is hard. 🤣 I think `coarseHealth` is good enough. I will rename it.

wbpcode:
friendly ping @htuch 

PS: waiting for 1.24.0 to be released first.

source/common/conn_pool/conn_pool_base.cc
```diff
@@ -77,7 +77,7 @@ bool ConnPoolImplBase::shouldCreateNewConnection(float global_preconnect_ratio)
   // upstream selection logic may result in bypassing this upstream entirely.
   // If an Envoy user wants preconnecting for degraded upstreams this could be
   // added later via extending the preconnect config.
-  if (host_->health() != Upstream::Host::Health::Healthy) {
+  if (host_->coarseHealth() != Upstream::Host::Health::Healthy) {
```
htuch, line 80: Not in this PR, but in a followup, should we adjust as many of these `coarseHealth` to `healthStatus` as we can? I'm hoping that we can generally get rid of most of the old-style, since it's a bit confusing to read code and wonder why we need to treat the two cases differently. The alternative is to move everything over at once, not sure what would break though.

source/common/upstream/upstream_impl.h
htuch: As per my other comment, I wonder how many places we could just switch to the new health status. If it's too risky to do it all at once, I think it's fine to do what you have in this PR, and maybe under runtime flag protection migrate the existing callsites?

source/common/conn_pool/conn_pool_base.cc
wbpcode: If we just want to remove this `coarseHealth` method from `Host`, it's easy for us to do this. We can use a simple util function + `healthStatus` to replace this method in the future.

But it's a little hard to complete remove this simple status abstraction from Envoy.

htuch: Sure. I think that would be a good place to get to, where we just do the abstraction where we need it.

wbpcode: I will add a todo to my task list. This wouldn't be too hard.

htuch (approved):
LGTM, thanks!

PR #23435, opened 2022-10-11
Sip proxy: Support for outbound transactions

arejasco:
Commit Message: Add support for outbound transactions to SIP proxy extension
Co-authored-by: Jonah Murphy [jonamurp@cisco.com](mailto:jonamurp@cisco.com)
Additional Description: See below
Risk Level: Medium
Testing: Unit testing & Manual testing
Docs Changes: API for SIP proxy plugin settings updated
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

This PR provides to SIP proxy extension for Envoy with support for handling outbound transactions, that is, transactions being initiated in the SIP server towards the client.

### Background

SIP proxy contrib extension was implemented in September 2021 to enable Service Mesh for SIP protocol, taking into account SIP specific router mechanisms and features.

Current behavior of SIP proxy extension supports the forwarding of inbound messages (request coming from the client to be redirected to an endpoint), but only responses to existing transactions initiated by an inbound request are supported to be forwarded from the endpoint to the client. The forwarding of outbound request coming from an endpoint is not currently supported by SIP proxy extension, as Envoy does not know to which downstream connection has to forward the message when received.

This PR provides through the usage of a new header the routing mechanisms for forwarding an outbound request coming from an endpoint to the target client, providing in that way support for outbound initiated transactions.

### Design

The basic idea is to provide support for outbound transactions by adding a new header, X-Envoy-Origin-Ingress, which will provide information about the thread and downstream connection where to forward any outbound transaction initiated. This X-Envoy-Origin-Ingress header will be composed of two parameters:

- Thread ID: Identify the thread instance to which the client is connected, with a format thread-number@local-envoy-address.
- Downstream connection ID: Identify the downstream connection instance being used by the thread instance for connecting with the client, with a format remote-client-address:remote-client-port@random-uuid.

These parameters will be initiated in the beginning, when a client connects to Envoy and a ConnectionManager is initiated. For each thread, a local map will relate the downstream connection ID with its ConnectionManager instance through a wrapper object called DownstreamConnection (new class).

Each message coming from the client (either request or response) will be modified by adding the X-Envoy-Origin-Ingress header before being forwarded to an endpoint.

The endpoint, when receiving a request coming from Envoy, will process the request as usual, but storing the value from X-Envoy-Origin-Ingress header for being used in future outbound transactions.

Each outbound request coming from a server will need to have the X-Envoy-Origin-Ingress header. It will be parsed and the DownstreamConnection associated will be recovered, providing the ConnectionManager instance to be used for forwarding the message to the client. An UpstreamTransaction instance (new class) will be created and mapped in a thread local map, saving the address of the endpoint sending the request.

Each response coming from the client to an outbound request will be processed by getting the UpstreamTransaction instance from the transaction ID and routing the message but setting the address of the endpoint which started the transaction as preferred destination. In that way the response to an outbound request will be routed by setting an affinity with the origin endpoint of the request, having the response forwarded to the same endpoint.

### System architecture

This is the relationship between classes in the forwarding of an outbound initiated request to a client:

image

And this is the relationship between classes in the forwarding of the response to an outbound initiated request to the original endpoint:

image

### Call flows

Here is the call flow for the connection and register of the client to the SIP server through Envoy:

image

Here is the call flow for a successful outbound transaction (an INVITE in this case):

image

### Configuration

For configuring the SIP proxy plugin to allow the management of upstream transactions as an optional feature, API files for SIP proxy extension has been modified to include a new node for SIP proxy clusters, called _upstream_transactions_. In that way, it will be possible to enable/disable upstream transactions at a cluster level.

```
static_resources:
  clusters:
  - name: cluster_0
    type:
      ...
    typed_extension_protocol_options:
      envoy.filters.network.sip_proxy:
        "@type": type.googleapis.com/envoy.extensions.filters.network.sip_proxy.v3alpha.SipProtocolOptions
         ...
         upstream_transactions:
            enabled: true
         ...
```

This node will define for now just one parameter, _enabled_, which enables or disables the support for outbound transactions in that cluster. Disabling outbound transactions will provoke the following actions in the cluster:

- X-Envoy-Origin-Ingress header won't be added in messages being forwarded upstream to the cluster.
- Request messages received upstream will be discarded.
- Indirectly, response messages received downstream will be always discarded, as the thread-local map for upstream transactions won't contain and upstream transaction instance with the transaction ID.

### Main changes done

Here is the summary of the main changes done in this PR:

- Parsing of X-Envoy-Origin-Ingress header during decoding, addition in all messages forwarded upstream and removal in all requests forwarded downstream.
- Creation of DownstreamConnection class for mapping ConnectionManager instances to their downstream connection IDs. Creation of a TLS-base thread-local map for this mapping.
- Modify ActiveTrans class to make it abstract, and create two implementations of this class:
    - DownstreamActiveTrans, representing all downstream initiated transactions. Replace the old use of ActiveTrans for DownstreamActiveTrans.
    - UpstreamActiveTrans, representing all upstream initiated transactions. Creation of a TLS-base thread-local map for mapping these upstream transactions with their transaction ID.
- Rename of some existing classes for better sense:
   - Router::UpstreamRequest -> Router::UpstreamConnection.
   - Router::ResponseDecoder -> Router::UpstreamMessageDecoder.
- Behavior of SIP proxy extension modified for working according to what's documented above.
- Set of statistics created for upstream transactions, mirroring downstream ones (still pending to review stats behavior).
   - Current stats remains unchanged but could be worth to add downstream_ prefix to them.
- Possibility to send local replies for outbound initiated transactions, although not used for now (for now all errors detected in outbound transactions will result in the discarding of the message).
- Metadata function for validating message headers in both inbound and outbound paths, and logic for discarding in both paths the message if any of the mandatory SIP headers which may provoke a crash is not present.


arejasco:
@durd07 This is the draft PR. Thanks for your time

durd07:
@arejasco thanks for your teams great contribution. we will go through them later. I will let you know if there are any comments.

phlax:
@durd07 would you mind giving this another pass

durd07:
As this is so big a PR, I need some time to review, I will began to review this soon.

phlax:
thanks

/wait-any

arejasco:
Thanks @durd07 @phlax

contrib/sip_proxy/filters/network/source/sip.h
```diff
@@ -103,7 +118,26 @@ class HeaderTypeMap {
       {"WWW-Authenticate", HeaderType::WAuth},
       {"Authorization", HeaderType::Auth},
       {"TopLine", HeaderType::TopLine},
-      {"P-Nokia-Cookie-IP-Mapping", HeaderType::PCookieIPMap}};
+      {"P-Nokia-Cookie-IP-Mapping", HeaderType::PCookieIPMap},
+      {"X-Envoy-Origin-Ingress", HeaderType::XEnvoyOriginIngress}};
+
+  const std::map<HeaderType, absl::string_view> sip_header_type_reverse_map_{
```
durd07, line 124: please help to replace `std::map` into `absl_flat_map`

arejasco:
Thanks for your review @durd07. I'll try to address all your comments in the following days

durd07:
@arejasco  I merged one PR https://github.com/envoyproxy/envoy/pull/24165#event-7941986437 with the latest stats change, please merge this and adapt the new stats changes.

Sorry for any inconvenience.

alyssawilk:
assigning myself to merge after this has @durd07's LGTM
/wait

arejasco:
Sorry for the lack of activity on this PR, I've been on a long abscence. I'll try to address your comments between this week and the next one

arejasco:
Still working on it

Issue #15062, opened 2021-02-16
Enable jwt_claims to header for jwt_based routing


mandarjog:
*Add the ability to copy jwt_claims to headers to enable jwt_based routing*

*Description*:
[jwt_authn_filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter)
validates jwt tokens and rejects a token if it is invalid. May times we want to route a request at ingress based on the jwt-token presented. For example a claim me be related to `canary` in which case the user should be routed to the canary backend.

At present this can be done using Lua or a Wasm filter, however this is an often requested feature and makes sense for it to be directly supported.

Options

**A: jwt_authn_filter configuration**
1. Add `action` configuration to the `jwt_authn_filter`, which will copy claims to a header.
2. Ensure that the target header is sanitized.
3. Relies on RBAC to happen afterwards
4. flush route to generate new route.

**B: support jwt-match in http routes**
1. Add nested jwt claims as route matches.
2. The match will be done on `unvalidated` claims, which is fine since it will be rejected by jwt later anyway.
3. This saves us from flushing routes. It makes it equivalent to a client that *also* populates the said header along with sending the jwt token.

@yangminzhu @kyessenov

qiwzhang:
For A:  we just add following into JwtProvider:
```
message PayloadToHeader {
   // it is the path in payload json, such as "a.c"
   string payload_path = 1;

   // the header key
   string header_key = 2;
};

repeated PayloadToHeader payload_to_headers = 13;

bool clear_route_cache = 14;

```

yangminzhu:
for A, after the jwt filter clear the route cache, will the request go through the same filter chain again? If so how does the jwt filter know if should not clear the cache again?

for B, the routing part probably still need to validate the token to some extent, sometimes a path is excluded from the JWT so the jwt filter might be configured to bypass validation of it, in that case, the routing should not use any information from the token.

or maybe `C: add a metadata_to_header filter`, the metadata will only be populated after validation and the filter will be a bit more general, but this has the same drawback for flushing routing.

mandarjog:
@yangminzhu if you believe “B” is equivalent to a client also populating the header then there is no validation needed during routing. If a route is not protected then it does not need a validated token. Can you help me understand the failure case?

qiwzhang:
@yangminzhu clear_route_cache will not cause request to re-pass filter chain.

qiwzhang:
[...]

PR #23444, opened 2022-10-11
Copy JWT Claim to HTTP Header

aryan16:
Commit Message: This patch adds the `claim_to_headers` field to the `JwtProvider` config to copy jwt claims to http headers.
Additional Description:
Risk Level: Low, a new feature
Testing: Added
Docs Changes: Added
Release Notes: Added
[Optional Runtime guard:]
[Optional Fixes #Issue] https://github.com/envoyproxy/envoy/issues/15062
[Optional Deprecated:]


api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +721,14 @@ message PerRouteConfig {
     string requirement_name = 2 [(validate.rules).string = {min_len: 1}];
   }
 }
+
+// This message specifies a combination of header name and claim name.
+message JwtClaimToHeader {
+  
+  // The HTTP header name.
```
htuch, line 728: Same comments as in the previous PR, LMK when this is ready for review.

aryan16: PTAL

aryan16: To answer you previous comment - This has been a popular ask, so it's good to add a standalone field for this. And even if we use dynamic metadata, we need the claim name and header name from user in some form. IMO this could be the easiest solution from user point of view.

aryan16:
Previously opened issue - https://github.com/envoyproxy/envoy/issues/15062

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +728,13 @@ message PerRouteConfig {
     string requirement_name = 2 [(validate.rules).string = {min_len: 1}];
   }
 }
+
+// This message specifies a combination of header name and claim name.
+message JwtClaimToHeader {
+  // The HTTP header name.
+  string name = 1
```
qiwzhang, line 735: maybe `header_name`  and `claim_name`?

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +728,13 @@ message PerRouteConfig {
     string requirement_name = 2 [(validate.rules).string = {min_len: 1}];
   }
 }
+
+// This message specifies a combination of header name and claim name.
+message JwtClaimToHeader {
+  // The HTTP header name.
```
qiwzhang, line 734: The HTTP header name to copy the claim to

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +728,13 @@ message PerRouteConfig {
...
+  string name = 1
+      [(validate.rules).string = {min_len: 1 well_known_regex: HTTP_HEADER_NAME strict: false}];
+
+  // JWT Claim
```
qiwzhang, line 738: Please specify how to use nested json names

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -286,6 +286,56 @@ void AuthenticatorImpl::handleGoodJwt(bool cache_hit) {
     }
   }
 
+  // Copy JWT Claim to Header
+  if (provider.claim_to_header_size() != 0) {
+    Json::ObjectSharedPtr payload_json;
+    try {
+      payload_json = Json::Factory::loadFromString(jwt_->payload_str_);
```
qiwzhang, line 293: Prefer not to do another JSON parsing again here.


jwt_verify_lib repo already parsed it into proto.Struct [here](https://github.com/google/jwt_verify_lib/blob/master/jwt_verify_lib/jwt.h#L49)

Prefer you add a couple get methods in "Jwt" class to retrieve claim fields

aryan16: Did you mean to add new getters (for object and boolean) in https://github.com/google/jwt_verify_lib/blob/master/src/jwt.cc ??

qiwzhang: Yes

aryan16: https://github.com/google/jwt_verify_lib/pull/74

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -286,6 +288,27 @@ void AuthenticatorImpl::handleGoodJwt(bool cache_hit) {
...
+    for (const auto& headerAndClaim : provider.claim_to_header()) {
+      if (!headerAndClaim.claim_name().empty() && !headerAndClaim.header_name().empty()) {
+        StructUtils payload_getter(jwt_->payload_pb_);
+        std::string stringClaimValue;
+        uint64_t intClaimValue = 0;
+        bool boolClaimValue;
+        if (payload_getter.GetString(headerAndClaim.claim_name(), &stringClaimValue) == StructUtils::OK) {
+          headers_->addCopy(Http::LowerCaseString(headerAndClaim.header_name()), stringClaimValue);
+        } else if(payload_getter.GetUInt64(headerAndClaim.claim_name(), &intClaimValue) == StructUtils::OK) {
+          headers_->addCopy(Http::LowerCaseString(headerAndClaim.header_name()), std::to_string(intClaimValue));
```
qiwzhang, line 302: this is not efficient with 3 lookup.    maybe you can add 
```
// This will fetch the field with primitive type, double, bool and string into a string
GetPrimitiveFieldToString(const string& name, string* out_string);
```
into jwt_verify_lib  StructUtil class

test/extensions/filters/http/jwt_authn/authenticator_test.cc
```diff
@@ -136,6 +136,43 @@ TEST_F(AuthenticatorTest, TestCompletePaddingInJwtPayload) {
...
+
+// This test verifies when wrong claim is passed in claim_to_header
+TEST_F(AuthenticatorTest, TestClaimToHeaderWithHeaderReplace) {
+  createAuthenticator();
+  EXPECT_CALL(*raw_fetcher_, fetch(_, _))
+      .WillOnce(Invoke([this](Tracing::Span&, JwksFetcher::JwksReceiver& receiver) {
+        receiver.onJwksSuccess(std::move(jwks_));
+      }));
+
+  Http::TestRequestHeaderMapImpl headers{
```
qiwzhang, line 166: I did not see a claim header passed-in here.

We need a test to verify a pass-in claim header is replaced with a good token,  and is removed with a bad token.

test/extensions/filters/http/jwt_authn/provider_verifier_test.cc
```diff
@@ -68,10 +68,15 @@ TEST_F(ProviderVerifierTest, TestOkJWT) {
   auto headers = Http::TestRequestHeaderMapImpl{
       {"Authorization", "Bearer " + std::string(GoodToken)},
       {"sec-istio-auth-userinfo", ""},
+      {"x-jwt-claim-sub", ""},
+      {"x-jwt-claim-nested", ""},
+
```
qiwzhang, line 73: remove this blank line

test/extensions/filters/http/jwt_authn/test_common.h
```diff
@@ -203,6 +221,19 @@ const char OtherGoodToken[] =
     "drRvLcvlT5gB4adOIOlmhm8xtXgYpvqrXfmMJCHbP9no7JATFaTEAkmA3OOxDsaOju4BFgMtRZtDM8p12QQG0rFl_FE-"
     "2FqYX9qA4q41HJ4vxTSxgObeLGA";
 
+const char NestedGoodToken[] =
```
qiwzhang, line 224: please add comment on what the claim looks like

test/extensions/filters/http/jwt_authn/authenticator_test.cc
aryan16: That part is in the verification, this test assumes that jwt is already verified and header sanitization wouldn't be affected. I am testing that logic here - https://github.com/envoyproxy/envoy/blob/96addf2784fe8da718945325688db77711512701/test/extensions/filters/http/jwt_authn/provider_verifier_test.cc#L203

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +728,14 @@ message PerRouteConfig {
...
 }
+
+// This message specifies a combination of header name and claim name.
+message JwtClaimToHeader {
+  // The HTTP header name to copy the claim to.
+  string header_name = 1
+      [(validate.rules).string = {min_len: 1 well_known_regex: HTTP_HEADER_NAME strict: false}];
+
+  // The field name for the JWT Claim : it can be a nested claim of type (eg. "claim.nested.key", "sub")
+  // String separated with "." in case of nested claims or a single value.
```
qiwzhang, line 739: remove "or a single value"
maybe:
The nested claim name must use dot "." to separate the JSON name path.

test/extensions/filters/http/jwt_authn/test_common.h
qiwzhang: can the JSON in the nice print format with new line?

test/extensions/filters/http/jwt_authn/authenticator_test.cc
qiwzhang: I see. thanks

qiwzhang:
presubmit test faled with format error
```
-            payload_getter.GetValue(header_and_claim.claim_name(),found);
+            payload_getter.GetValue(header_and_claim.claim_name(), found);
```

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -286,6 +288,36 @@ void AuthenticatorImpl::handleGoodJwt(bool cache_hit) {
     }
   }
 
+  // Copy JWT Claim to Header
+  if (provider.claim_to_headers_size() != 0) {
+    for (const auto& header_and_claim : provider.claim_to_headers()) {
+      if (!header_and_claim.claim_name().empty() && !header_and_claim.header_name().empty()) {
```
qiwzhang, line 294: not need this "emtpy" check,  validation will fail if any of them is empty.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -286,6 +288,36 @@ void AuthenticatorImpl::handleGoodJwt(bool cache_hit) {
...
+        StructUtils payload_getter(jwt_->payload_pb_);
+        const ::google::protobuf::Value* found;
+        google::jwt_verify::StructUtils::FindResult claim_value =
```
qiwzhang, line 297: move these codes for each header/claim into a separate function?

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -286,6 +288,36 @@ void AuthenticatorImpl::handleGoodJwt(bool cache_hit) {
...
+            payload_getter.GetValue(header_and_claim.claim_name(),found);
+        if (claim_value == StructUtils::OK) {
+          if (found->kind_case() == google::protobuf::Value::kStringValue) {
+            headers_->addCopy(Http::LowerCaseString(header_and_claim.header_name()),
+                              found->string_value());
+          } else if (found->kind_case() == google::protobuf::Value::kNumberValue) {
+            headers_->addCopy(Http::LowerCaseString(header_and_claim.header_name()),
+                              std::to_string(static_cast<uint64_t>(found->number_value())));
```
qiwzhang, line 305: Why cast  `double` to uint64_t?   This jwt_authn filter can be used by others,  can we just do 
std::to_string<double>

source/extensions/filters/http/jwt_authn/extractor.cc
```diff
@@ -186,6 +188,8 @@ class ExtractorImpl : public Logger::Loggable<Logger::Id::jwt>, public Extractor
   absl::btree_map<std::string, CookieLocationSpec> cookie_locations_;
 
   std::vector<LowerCaseString> forward_payload_headers_;
+
+  std::vector<LowerCaseString> claim_to_headers_;
```
qiwzhang, line 192: combined to use single vector, call it `headers_to_sanitize`,  add both `forward_payload_headers` and `claim_to_headers` to it.

also rename the function santizePayloadHeaders to sanitizeHeaders

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: If we go ahead without this check and any of the value is empty, it could create issues while fetching claim or copying the value to the header. It's safe to add this check IMO

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: sure

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: The reason to convert this uint64_t becuase our method `GetValue()` can't return type `uint64`. It just returns this https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.pb.h#L350 which is of type double. And Generally JWT Claims are not double or float, so it's safe to convert them to Int

source/extensions/filters/http/jwt_authn/extractor.cc
aryan16: Sure

source/extensions/filters/http/jwt_authn/authenticator.cc
qiwzhang: ok, then.  But please check its range before static_cast<uint64>, same as the jwt_verify_lib.

source/extensions/filters/http/jwt_authn/authenticator.cc
qiwzhang: envoy filter_config should have been validated before passing it to the jwt_authn filter to use.   per your validation rules, an empty header or claim will fail the config,  the config will not be used.  That is why you don't need to check here.

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: sure

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: understood, thanks for the clarification

source/extensions/filters/http/jwt_authn/extractor.cc
```diff
@@ -136,9 +136,7 @@ class ExtractorImpl : public Logger::Loggable<Logger::Id::jwt>, public Extractor
 
   std::vector<JwtLocationConstPtr> extract(const Http::RequestHeaderMap& headers) const override;
 
-  void sanitizePayloadHeaders(Http::HeaderMap& headers) const override;
-
-  void sanitizeClaimHeaders(Http::HeaderMap& headers) const override;
+  void sanitizePayloadAndClaimHeaders(Http::HeaderMap& headers) const override;
```
qiwzhang, line 139: just call it `sanitizeHeaders`

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -272,6 +275,37 @@ void AuthenticatorImpl::verifyKey() {
   handleGoodJwt(/*cache_hit=*/false);
 }
 
+bool AuthenticatorImpl::addJWTClaimToHeader(std::string claim_name, std::string header_name) {
```
qiwzhang, line 278: use `const std::string&` for both parameters

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -272,6 +275,37 @@ void AuthenticatorImpl::verifyKey() {
...
+  google::jwt_verify::StructUtils::FindResult status =
+      payload_getter.GetValue(claim_name, claim_value);
+  if (status == StructUtils::OK) {
+    if (claim_value->kind_case() == google::protobuf::Value::kStringValue) {
+      headers_->addCopy(Http::LowerCaseString(header_name), claim_value->string_value());
+    } else if (claim_value->kind_case() == google::protobuf::Value::kNumberValue) {
+      if (claim_value->number_value() < 0 ||
+          claim_value->number_value() >=
+              static_cast<double>(std::numeric_limits<uint64_t>::max())) {
+        return false;
```
qiwzhang, line 290: return value is not needed, just return void

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -272,6 +275,37 @@ void AuthenticatorImpl::verifyKey() {
   handleGoodJwt(/*cache_hit=*/false);
 }
 
+bool AuthenticatorImpl::addJWTClaimToHeader(std::string claim_name, std::string header_name) {
+  StructUtils payload_getter(jwt_->payload_pb_);
+  const ::Envoy::ProtobufWkt::Value* claim_value;
+  google::jwt_verify::StructUtils::FindResult status =
+      payload_getter.GetValue(claim_name, claim_value);
```
qiwzhang, line 282: use `auto`, easier to read
```
const auto status = 
```

source/extensions/filters/http/jwt_authn/authenticator.cc
qiwzhang: here, log an error. also print out the claim name and value for easy debug.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -272,6 +275,42 @@ void AuthenticatorImpl::verifyKey() {
...
+  if (status == StructUtils::OK) {
+    if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kStringValue) {
+      headers_->addCopy(Http::LowerCaseString(header_name), claim_value->string_value());
+      ENVOY_LOG(debug, "--------claim : {} with value : {} is added to the header : {} -----------",
+                claim_name, claim_value->string_value(), header_name);
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kNumberValue) {
+      if (claim_value->number_value() < 0 ||
+          claim_value->number_value() >=
+              static_cast<double>(std::numeric_limits<uint64_t>::max())) {
+        ENVOY_LOG(debug, "--------claim : {} is not a valid integer -----------", claim_name);
```
qiwzhang, line 292: please log the "double" value too. also log this as error as I see it should be rare.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -272,6 +275,42 @@ void AuthenticatorImpl::verifyKey() {
...
+      headers_->addCopy(Http::LowerCaseString(header_name), std::to_string(int_claim_value));
+      ENVOY_LOG(debug, "--------claim : {} with value : {} is added to the header : {} -----------",
+                claim_name, std::to_string(int_claim_value), header_name);
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kBoolValue) {
+      const std::string str_bool_claim = claim_value->bool_value() ? "true" : "false";
+      headers_->addCopy(Http::LowerCaseString(header_name), str_bool_claim);
+      ENVOY_LOG(debug, "--------claim : {} with value : {} is added to the header : {} -----------",
+                claim_name, str_bool_claim, header_name);
+    } else {
+      ENVOY_LOG(debug, "--------claim : {} is not of type string, int or bool -----------",
```
qiwzhang, line 304: "is not a primitive types of int, string or bool.'

qiwzhang: It seems that following two lines are duplicated 3 time
```
headers->addCopy
ENVOY_LOG(debug...
```

Try to remove the duplication codes.

qiwzhang (approved):

qiwzhang:
@envoyproxy/api-shepherds 
envoyproxy/dependency-shepherds
...

api/envoy/extensions/filters/http/jwt_authn/v3/README.md
```diff
@@ -30,6 +30,18 @@ If a custom location is desired, `from_headers` or `from_params` can be used to
 
 If a JWT is valid, its payload will be passed to the backend in a new HTTP header specified in `forward_payload_header` field. Its value is base64url-encoded JWT payload in JSON.
 
+## Copy Validated JWT Claims to HTTP Request Headers
```
htuch, line 33: @qiwzhang why is this parallel README.md doc being maintain to the Envoy RST docs? Seems it would be good to converge them for single source-of-truth.

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -277,6 +277,17 @@ message JwtProvider {
   // Enables JWT cache, its size is specified by ``jwt_cache_size``.
   // Only valid JWT tokens are cached.
   JwtCacheConfig jwt_cache_config = 12;
+
+  // Add JWT Claim to HTTP Header
+  // Specify the Claim name you want to copy in which HTTP header. For examples, following config:
```
htuch, line 282: What are the semantics if the header already has a value? append? replace?

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +728,15 @@ message PerRouteConfig {
...
+
+// This message specifies a combination of header name and claim name.
+message JwtClaimToHeader {
+  // The HTTP header name to copy the claim to.
+  string header_name = 1
+      [(validate.rules).string = {min_len: 1 well_known_regex: HTTP_HEADER_NAME strict: false}];
+
+  // The field name for the JWT Claim : it can be a nested claim of type (eg. "claim.nested.key", "sub")
+  // String separated with "." in case of nested claims. The nested claim name must use dot "." to separate
+  // the JSON name path.
```
htuch, line 740: How does this work with arrays?

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
```diff
@@ -130,6 +136,10 @@ JWT payload will be added to the request header as following format::
 
     x-jwt-payload: base64url_encoded(jwt_payload_in_JSON)
 
+JWT Claim ("sub" and "nested.claim.key") will be added to HTTP headers as following format::
+    x-jwt-claim-sub: <JWT Claim>
+    x-jwt-claim-nested-key: <JWT Claim>
+
```
htuch, line 142: Needs an entry in https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml

api/envoy/extensions/filters/http/jwt_authn/v3/README.md
qiwzhang: OK,  it is a good idea.  I will do that in a separate pr.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -270,6 +275,43 @@ void AuthenticatorImpl::verifyKey() {
...
+                "--------claim : {} is not a primitive type of int, string, or bool -----------",
+                claim_name);
+    }
+    if (!str_claim_value.empty()) {
+      headers_->addCopy(Http::LowerCaseString(header_name), str_claim_value);
+      ENVOY_LOG(debug, "--------claim : {} with value : {} is added to the header : {} -----------",
+                claim_name, str_claim_value, header_name);
+    }
+  } else {
+    ENVOY_LOG(debug,
```
htuch, line 309: Probably should bump some stat - debug logging is not useful in production and we are basically silently failing here.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -270,6 +275,43 @@ void AuthenticatorImpl::verifyKey() {
...
+      str_claim_value = claim_value->string_value();
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kNumberValue) {
+      if (claim_value->number_value() < 0 ||
+          claim_value->number_value() >=
+              static_cast<double>(std::numeric_limits<uint64_t>::max())) {
+        ENVOY_LOG(debug, "--------claim : {} with value : {} not a valid integer -----------",
+                  claim_name, claim_value->number_value());
+      }
+      const uint64_t int_claim_value = static_cast<uint64_t>(claim_value->number_value());
+      str_claim_value = std::to_string(int_claim_value);
```
htuch, line 295: We're doing a double -> uint64 -> string here? Seems potentially dangerous, i.e. the signed claim was on some double value, we are now forwarding it with some truncation of the floating portion. @qiwzhang WDYT?

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
aryan16: Replace, if header has any value. No one can set any value to this header, it's just reserved for JWT-Claims

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
aryan16: We do not support arrays as of now.

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
aryan16: sure, will add this.

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
qiwzhang: The header is sanitized first in the filter.

source/extensions/filters/http/jwt_authn/authenticator.cc
qiwzhang: It is not dangerous, but we may have some precision loss from two fronts

1)  Using "double" in proto.Struct to represent JSON number,  but its [RFC](https://www.rfc-editor.org/rfc/rfc7159#section-6) say that it is ok.
2)  The code in this pr drop negative and fraction. 

For 2),  I will suggest to use std::to_string<double> directly.

aryan16: The only problem with this is we will never get integer output from jwt_verify library(https://github.com/google/jwt_verify_lib/blob/master/src/struct_utils.cc#L105). Because we never get int from google protobuf, it's of double type (https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.pb.h#L350). So when user has passed a integer (eg. 500), the `jwt-verify` returns something like (`500.0000`) which could be an issue. Generally claims are non negative integers, so it's better to stick to this. I can do one thing, if it is not an valid integer then just add it as a double value in the claim ? @qiwzhang @htuch wdyt ?

aryan16: added one solution, lmk what you guys think about it ?

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
qiwzhang: Please use lower case for Claim.  

Please also mention that the header name will be sanitized and replaced.

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -717,3 +728,15 @@ message PerRouteConfig {
     string requirement_name = 2 [(validate.rules).string = {min_len: 1}];
   }
 }
+
+// This message specifies a combination of header name and claim name.
+message JwtClaimToHeader {
+  // The HTTP header name to copy the claim to.
```
qiwzhang, line 734: Please also mention that the header name will be sanitized and replaced.

api/envoy/extensions/filters/http/jwt_authn/v3/README.md
qiwzhang, line 33: Please move this new section to `docs/root/configuration/http/http_filters/jwt_authn_filter.rst`.  This doc will be removed.

source/extensions/filters/http/jwt_authn/authenticator.cc
qiwzhang: SG

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -270,6 +275,41 @@ void AuthenticatorImpl::verifyKey() {
...
+  if (status == StructUtils::OK) {
+    if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kStringValue) {
+      str_claim_value = claim_value->string_value();
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kNumberValue) {
+      if (claim_value->number_value() < 0 ||
+          claim_value->number_value() >=
+              static_cast<double>(std::numeric_limits<uint64_t>::max())) {
+        // if claim is not an integer, adding a double type value in the claim.
+        headers_->addCopy(Http::LowerCaseString(header_name),
+                          std::to_string(claim_value->number_value()));
```
qiwzhang, line 293: write function to convert double to string,  in the function,  we can detect if it is an integer, without fraction with in correct range, then cast to int, otherwise print as double.

api/envoy/extensions/filters/http/jwt_authn/v3/README.md
aryan16: should I delete the content from `README.md` as well ?

qiwzhang: Yes, I am going to remove this doc in https://github.com/envoyproxy/envoy/pull/23923

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -123,6 +126,17 @@ std::string AuthenticatorImpl::name() const {
   return "_UNKNOWN_";
 }
 
+std::string AuthenticatorImpl::convertDoubleOrIntToString(const double& double_value) {
```
qiwzhang, line 129: primitive type, e.g. double, not need to use "const &"  just copy it,  just "double"

qiwzhang, line 129: this should be just a static global function, not a class member function.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -123,6 +126,17 @@ std::string AuthenticatorImpl::name() const {
...
+  double fraction;
+  if ((double_value < 0 ||
+       double_value->number_value() >= static_cast<double>(std::numeric_limits<uint64_t>::max())) &&
+      modf(double_value, &fraction) != 0) {
```
qiwzhang, line 133: this is integ_part, not fraction part

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -285,14 +300,7 @@ void AuthenticatorImpl::addJWTClaimToHeader(const std::string& claim_name,
...
     } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kNumberValue) {
-      if (claim_value->number_value() < 0 ||
-          claim_value->number_value() >=
-              static_cast<double>(std::numeric_limits<uint64_t>::max())) {
-        ENVOY_LOG(debug, "--------claim : {} with value : {} not a valid integer -----------",
-                  claim_name, claim_value->number_value());
-      }
-      const uint64_t int_claim_value = static_cast<uint64_t>(claim_value->number_value());
-      str_claim_value = std::to_string(int_claim_value);
+      str_claim_value = convertDoubleOrIntToString(claim_value->number_value())
```
qiwzhang, line 303: just `convertDoubleToString`

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: `modf(double_value, &fraction)` this will check the fraction part.

qiwzhang: From the document of `modf(double,  &integer_part)`,  the 2nd argument is returning the integer part of the double.    The variable name of `fraction` is misleading, it should be `integer_part`

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -123,6 +127,17 @@ std::string AuthenticatorImpl::name() const {
   return "_UNKNOWN_";
 }
 
+std::string AuthenticatorImpl::convertDoubleToString(double double_value) {
+  double int_part;
```
qiwzhang, line 131: Please add comment for this function as
// If the number is unsigned 64 bit integer,  convert to string as integer,
// otherwise, convert to string as double.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -54,6 +54,9 @@ class AuthenticatorImpl : public Logger::Loggable<Logger::Id::jwt>,
 
   TimeSource& timeSource() { return time_source_; }
 
+  // convert double to string
+  static std::string convertDoubleToString(double double_value);
```
qiwzhang, line 58: Make this as global static function,  not a static class function.  so no need to declare in the class

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -123,6 +127,17 @@ std::string AuthenticatorImpl::name() const {
   return "_UNKNOWN_";
 }
 
+std::string AuthenticatorImpl::convertDoubleToString(double double_value) {
```
qiwzhang, line 130: remove `AuthenticatorImpl::`   not a class function.

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
```diff
@@ -216,3 +207,24 @@ Above config uses more complex *group* requirements:
 
 * The first *rule* specifies *requires_any*; if any of **provider1** or **provider2** requirement is satisfied, the request is OK to proceed.
 * The second *rule* specifies *requires_all*; only if both **provider1** and **provider2** requirements are satisfied, the request is OK to proceed.
+
+Copy validated JWT claims to HTTP request headers example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If a JWT is valid, you can add some of its claims of type (string, integer, boolean) to a new HTTP header to pass to the upstream.You can specify claims and headers in
+`claim_to_headers` field. Nested claims are also supported.
```
qiwzhang, line 215: use  :ref:`claim_to_headers <envoy_v3_api_field_extensions.filters.http.jwt_authn.v3.JwtProvider.claim_to_headers>`

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
```diff
@@ -216,3 +207,24 @@ Above config uses more complex *group* requirements:
...
+  providers:
+    provider_name2:
+      issuer: https://example2.com
+      claim_to_headers:
+      - header_name: x-jwt-claim-sub
+        claim_name: sub
+      - header_name: x-jwt-claim-nested-key
+        claim_name: nested.claim.key
+
+JWT Claim ("sub" and "nested.claim.key") will be added to HTTP headers as following format::
```
qiwzhang, line 228: lower case claim

changelogs/current.yaml
```diff
@@ -117,5 +117,8 @@ new_features:
 - area: tracing
   change: |
     added support for setting the hostname used when sending spans to a Datadog collector using the :ref:`collector_hostname <envoy_v3_api_field_config.trace.v3.DatadogConfig.collector_hostname>` field.
+- area: jwt_authn
+  change: |
+    added support for copying nested jwt claims to http headers.
```
qiwzhang, line 122: remove "nested"

qiwzhang:
`precheck format` test failed with following

```
 }
 
 void AuthenticatorImpl::addJWTClaimToHeader(const std::string& claim_name,
-                                       const std::string& header_name) {
+                                            const std::string& header_name) {

```

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
```diff
@@ -216,3 +207,29 @@ Above config uses more complex *group* requirements:
...
+
+Copy validated JWT claims to HTTP request headers example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`claim_to_headers <envoy_v3_api_field_extensions.filters.http.jwt_authn.v3.JwtProvider.claim_to_headers>` has two fields:
+
+* Field *header_name* specifies the name of new http header reserved for jwt claim.
+* Field *claim_name* specifies the claim from verified jwt token.
+
+If a JWT is valid, you can add some of its claims of type (string, integer, boolean) to a new HTTP header to pass to the upstream.You can specify claims and headers in
```
qiwzhang, line 219: Please put a space between "to the update.You" here.

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
```diff
@@ -216,3 +207,29 @@ Above config uses more complex *group* requirements:
 
 * The first *rule* specifies *requires_any*; if any of **provider1** or **provider2** requirement is satisfied, the request is OK to proceed.
 * The second *rule* specifies *requires_all*; only if both **provider1** and **provider2** requirements are satisfied, the request is OK to proceed.
+
+Copy validated JWT claims to HTTP request headers example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`claim_to_headers <envoy_v3_api_field_extensions.filters.http.jwt_authn.v3.JwtProvider.claim_to_headers>` has two fields:
```
qiwzhang, line 214: line 214-218 seems odd at the top of section. Is this indented?  I would suggest move them below line 220

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
```diff
@@ -216,3 +207,29 @@ Above config uses more complex *group* requirements:
...
+
+* Field *header_name* specifies the name of new http header reserved for jwt claim.
+* Field *claim_name* specifies the claim from verified jwt token.
+
+If a JWT is valid, you can add some of its claims of type (string, integer, boolean) to a new HTTP header to pass to the upstream.You can specify claims and headers in
+`claim_to_headers` field. Nested claims are also supported.
```
qiwzhang, line 220: can you put a link in ":ref:" format for any field name?

docs/root/configuration/http/http_filters/jwt_authn_filter.rst
qiwzhang: Actually this paragraph can be re-worded as

The field claim_to_headers is a repeat of message JwtClaimToHeader(use :ref: for msg link,  it is a different format as field link).  The message JwtClaimToHeader has two fields:

...

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -286,6 +330,13 @@ void AuthenticatorImpl::handleGoodJwt(bool cache_hit) {
     }
   }
 
+  // Copy JWT Claim to Header
+  if (provider.claim_to_headers_size() != 0) {
```
qiwzhang, line 334: Actually, you don't need to check the size().  Just use "for" loop.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -8,21 +8,36 @@
...
 
 namespace Envoy {
 namespace Extensions {
 namespace HttpFilters {
 namespace JwtAuthn {
 namespace {
 
+// If the number is unsigned 64 bit integer, convert to string as integer,
+// otherwise, convert to string as double.
+static std::string convertDoubleToString(double double_value) {
```
qiwzhang, line 31: convertClaimDoubleToString  since it is specifically used for claim.

qiwzhang (approved):

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
htuch: Please explain in the docs that this is replace semantics.

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
htuch: Please explain in docs.

source/extensions/filters/http/jwt_authn/authenticator.cc
htuch: Any followup here?

source/extensions/filters/http/jwt_authn/authenticator.cc
htuch: Can you folks explain for my benefit (and posterity), why it's safe to ignore the float portion? My naïve thinking here is that a JWT claim is about a specific value, e.g. "Joe Doe is allowed access to this document when the weather is warmer than 34.5 degrees celsius". If we ignore the float, Joe Doe gets access when the weather is 34 degrees celsius. Maybe this example is contrived, but unless there is some convention about discarding float portion or precision here, this seems a plausible attack vector when floating point claims are used.

qiwzhang: Sorry, we did not make it clear. We never ignore any fraction.   Here is what we do
```
   if   detect double is unsigned integer:  
          return std::to_string<uint64_t> ()
   else 
       return std::to_string<double>() 
```
If a number has fraction, we still print it as "double".   Just for nice print for integer, we do such optimization.
If we always use std::to_string<double>,  a "100" integer, will be "100.0000", this is not nice.

aryan16:
@htuch PTAL

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: cc : @qiwzhang

qiwzhang: I feel that adding stat for these cases will not be so useful.  The symptoms of such errors will show up in other ways. In this case, the backend will expect certain claim headers, but due some mistakes, such as claim name typo or wrong type, we will not copy them, the backend will not receive them, it will complain.  They can be easily found in the development test as JWT payload format usually don't change.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -270,6 +288,32 @@ void AuthenticatorImpl::verifyKey() {
...
+  if (status == StructUtils::OK) {
+    if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kStringValue) {
+      str_claim_value = claim_value->string_value();
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kNumberValue) {
+      str_claim_value = convertClaimDoubleToString(claim_value->number_value());
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kBoolValue) {
+      str_claim_value = claim_value->bool_value() ? "true" : "false";
+    } else {
+      ENVOY_LOG(debug,
+                "--------claim : {} is not a primitive type of int, string, or bool -----------",
```
htuch, line 306: s/int/number/?

test/extensions/filters/http/jwt_authn/authenticator_test.cc
```diff
@@ -136,6 +136,43 @@ TEST_F(AuthenticatorTest, TestCompletePaddingInJwtPayload) {
...
+      {"Authorization", "Bearer " + std::string(NestedGoodToken)}};
+
+  expectVerifyStatus(Status::Ok, headers);
+
+  EXPECT_EQ(headers.get_("x-jwt-claim-sub"), "test@example.com");
+  EXPECT_EQ(headers.get_("x-jwt-claim-nested"), "value1");
+  EXPECT_EQ(headers.get_("x-jwt-bool-claim"), "true");
+  EXPECT_EQ(headers.get_("x-jwt-int-claim"), "9999");
+}
+
```
htuch, line 157: Please add a test to validate replace vs. append semantics.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -270,6 +288,32 @@ void AuthenticatorImpl::verifyKey() {
...
 }
 
+void AuthenticatorImpl::addJWTClaimToHeader(const std::string& claim_name,
+                                            const std::string& header_name) {
+  StructUtils payload_getter(jwt_->payload_pb_);
+  const ::Envoy::ProtobufWkt::Value* claim_value;
+  const auto status = payload_getter.GetValue(claim_name, claim_value);
+  std::string str_claim_value;
+  if (status == StructUtils::OK) {
+    if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kStringValue) {
```
htuch, line 298: This whole block should be a separate utility function with a complete set of tests for all behaviors. It's got basic coverage below, but I'm not convinced all the various branching cases and state space is covered. This is security critical, so I think this is worth doing.

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -270,6 +288,32 @@ void AuthenticatorImpl::verifyKey() {
...
+      str_claim_value = claim_value->string_value();
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kNumberValue) {
+      str_claim_value = convertClaimDoubleToString(claim_value->number_value());
+    } else if (claim_value->kind_case() == Envoy::ProtobufWkt::Value::kBoolValue) {
+      str_claim_value = claim_value->bool_value() ? "true" : "false";
+    } else {
+      ENVOY_LOG(debug,
+                "--------claim : {} is not a primitive type of int, string, or bool -----------",
+                claim_name);
```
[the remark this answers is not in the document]
aryan16: There are many comments so that comment for lost somewhere. @qiwzhang responded to your comment with 
`I feel that adding stat for these cases will not be so useful. The symptoms of such errors will show up in other ways. In this case, the backend will expect certain claim headers, but due some mistakes, such as claim name typo or wrong type, we will not copy them, the backend will not receive them, it will complain. They can be easily found in the development test as JWT payload format usually don't change.`

test/extensions/filters/http/jwt_authn/authenticator_test.cc
aryan16: These kind of tests are already there - https://github.com/envoyproxy/envoy/pull/23444/files#diff-302a5e5ea85804ebdb3e270f87987fab8dfe876ae1cb85866dedac4833468d52R77

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: What do you think is not covered in the tests ? imo we have added all the possible tests

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -277,6 +277,17 @@ message JwtProvider {
...
+
+  // Add JWT claim to HTTP Header
+  // Specify the claim name you want to copy in which HTTP header. For examples, following config:
+  // The claim must be of type; string, int, bool. Array type claims are not supported
+  // .. code-block:: yaml
+  //
+  //   claim_to_headers:
+  //   - name: x-jwt-claim-nested-claim
+  //     claim: claim.nested.key
+
```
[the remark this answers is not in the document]
aryan16: thanks, resolved

source/extensions/filters/http/jwt_authn/authenticator.cc
aryan16: Updated the testing token with an array type value (https://github.com/envoyproxy/envoy/pull/23444/files#diff-ccb8c0f9dfe4011484c9d7ca83d56abbe8b60c7b1b8dcd8efbc3debe0c6c5a1cR237) and added one more test on this.

htuch (approved):

htuch:
@aryan16 can you update the top-level commit message with description, issue, risk level, testing etc?

aryan16:
> @aryan16 can you update the top-level commit message with description, issue, risk level, testing etc?

Updated, thanks !!

PR #23449, opened 2022-10-11
quic: extends QUIC connection ID generator

danzh2010:
Commit Message: make QUIC connection ID generation an extension point with currently in-use `EnvoyDeterministicConnectionIdGenerator` as the default implementation.
Additional Description: fix some previously unused QUICHE build targets.
Risk Level: low, control plane change
Testing: added new unit tests
Docs Changes: docs/root/api-v3/config/quic/quic_extensions.rst
Release Notes: N/A
[API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md): interface naming and documentation


danzh2010:
@martinduke please review the CID generator factory interface to make sure I didn't miss anything important: https://github.com/envoyproxy/envoy/pull/23449/files#diff-3298485d385ae40e90a429315ce68fc1163d95cb435bed897ea5dc19f22692ec

source/extensions/quic/connection_id_generator/envoy_deterministic_connection_id_generator.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+absl::optional<quic::QuicConnectionId>
+EnvoyDeterministicConnectionIdGenerator::MaybeReplaceConnectionId(
+    const quic::QuicConnectionId& original, const quic::ParsedQuicVersion& version) {
+  auto new_cid = DeterministicConnectionIdGenerator::MaybeReplaceConnectionId(original, version);
+  if (new_cid.has_value()) {
+    adjustNewConnectionIdForRoutine(new_cid.value(), original);
+  }
+  return new_cid;
```
martinduke, line 28: This should check if new_cid == original after all the changes and return nullopt if so.

source/extensions/quic/connection_id_generator/envoy_deterministic_connection_id_generator.cc
```diff
@@ -0,0 +1,94 @@
...
+namespace Quic {
+
+absl::optional<quic::QuicConnectionId>
+EnvoyDeterministicConnectionIdGenerator::GenerateNextConnectionId(
+    const quic::QuicConnectionId& original) {
+  auto new_cid = DeterministicConnectionIdGenerator::GenerateNextConnectionId(original);
+  if (new_cid.has_value()) {
+    adjustNewConnectionIdForRoutine(new_cid.value(), original);
+  }
+  return new_cid;
```
martinduke, line 18: This should check if new_cid == original after all the changes and return nullopt if so.

source/extensions/quic/connection_id_generator/envoy_deterministic_connection_id_generator.h
```diff
@@ -1,7 +1,13 @@
...
+#include "source/common/quic/envoy_quic_connection_id_generator_factory.h"
+
 #include "quiche/quic/core/deterministic_connection_id_generator.h"
 
+#if defined(__linux__)
+#include <linux/filter.h>
+#endif
+
 namespace Envoy {
 namespace Quic {
```
martinduke, line 13: Please note that there is a new virtual function in ConnectionIdGeneratorInterface called ConnectionIdLength(uint8_t first_byte). I'm not sure when it will make its way into Envoy Quiche.

That said, there's no need to override the implementation in DeterministicConnectionIdGenerator, so I don't think there's any code change necessary in this PR.

martinduke (commented):
The new instance of ConnectionIdGeneratorInterface looks good to me, minus some nits.

I do not see exactly how Google-specific stuff is going to interact with the classes you have set up here, but as you've seen the relevant design I presume there's a roadmap to integrate them.

danzh2010:
> I do not see exactly how Google-specific stuff is going to interact with the classes you have set up here, but as you've seen the relevant design I presume there's a roadmap to integrate them.

The internal CID generator implementation will be plugged in via the newly added `EnvoyQuicConnectionIdGeneratorFactory` interface. So I'd like to make sure the factory interface is compatible to our implementation of `ConnectionIdGeneratorInterface`.

martinduke (dismissed):
ConnectionIdGeneratorInterface parts LGTM

source/common/quic/envoy_quic_connection_id_generator_factory.h
```diff
@@ -0,0 +1,60 @@
...
+  createQuicConnectionIdGenerator(quic::LoadBalancerEncoder& lb_encoder,
+                                  uint32_t worker_index) PURE;
+
+  /**
+   * the length of connection IDs to be generated when there is no active config
+   */
+  virtual uint8_t getConnectionIdLengthWithoutRouteConfig() PURE;
+
+  /**
+   * Returns true if the connection ID length should be encoded in first byte of the generated
```
RyanTheOptimist, line 32: nit: "should be" => "will be", or perhaps just "is"

source/extensions/quic/connection_id_generator/envoy_deterministic_connection_id_generator.cc
```diff
@@ -0,0 +1,95 @@
...
+      {0x05, 0, 0, 0x00000001}, //                   ja return
+      {0x20, 0, 0,              // packet_too_short: ld rxhash
+       static_cast<uint32_t>(SKF_AD_OFF + SKF_AD_RXHASH)},
+      {0x94, 0, 0, concurrency}, // return:         mod #socket_count
+      {0x16, 0, 0, 0000000000},  //                 ret a
+  };
+  // SPELLCHECKER(on)
+
+  // Note that this option refers to the BPF program data above, which must live until the
+  // option is used. The program is kept as a member variable for this purpose.
```
RyanTheOptimist, line 82: fwiw, I think we could also make this a static const variable instead of a member variable, though I'm not sure if that would be better or not.

danzh2010: Yeah, it won't make things much cleaner than today, especially both `filter_` and `prog_` need to become static const variables in that case.

danzh2010:
@penguingao FYI, this is the Envoy extension point for QUIC CID stuff.

RyanTheOptimist (approved):

PR #23451, opened 2022-10-12
dubbo codec: remove Http HeaderMap in the dubbo message

wbpcode:
Commit Message: dubbo codec: remove Http HeaderMap in the dubbo message
Additional Description:

In the original implementation of dubbo proxy, `Http::HeaderMap` is used to store a copy of dubbo attachments. By this way, we can reuse `HeaderMatcher`-related functions/classes. It's not necessary for now because the unified matcher API will be used.

And the key of `Http::HeaderMap` is case-insensitive and only string value could be stored in. So, we must use a `Hessian2::Object::UntypedMap` to store complete attachments. It would bring some additional overhead.

This PR removed `Http::HeaderMap` from our new dubbo codec.

Risk Level: Low
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.

### DIFF
--- a/source/extensions/common/dubbo/BUILD
+++ b/source/extensions/common/dubbo/BUILD
@@ -42,7 +42,6 @@ envoy_cc_library(
     deps = [
         ":message_lib",
         "//source/common/buffer:buffer_lib",
-        "//source/common/http:header_map_lib",
     ],
 )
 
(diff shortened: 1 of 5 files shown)


lizan (dismissed):

wbpcode:
cc @lizan could your take a look again? Thanks. 😺

lizan (approved):

Issue #23390, opened 2022-10-06
Visibility failure on Bazel 6.0


gregestren:
*Title*: *Visibility failure on Bazel 6.0*

*Description*:
[Bazel 6.0 pre-release tests](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1290#0183aad6-fffa-400c-a8e6-989970b479f1) show envoy failing with `--incompatible_enforce_config_setting_visibility`.

This isn't a surprise. That flag [fixes](https://github.com/bazelbuild/bazel/issues/12932) a long-standing omission in Bazel's visibility checking: `config_setting` has never been visibility-checked. That means that if a `config_setting` is defined in a package with private visibility, the `config_setting` is still public. The flag updates `config_setting` to properly honor visibility.

So flipping the flag makes it possible for existing `config_setting` references to fail. The fix is to update their visibility to reflect actual usage.

*Repro steps*:
```
$ git clone -v https://github.com/envoyproxy/envoy.git
$ USE_BAZEL_VERSION=latest bazel build --nobuild //test/common/common/... //test/integration/... //test/exe/... --incompatible_enforce_config_setting_visibility

ERROR: /usr/local/google/home/gregce/.cache/bazel/_bazel_gregce/160b5d944412957a853c062174d7f0d8/external/proxy_wasm_cpp_host/BUILD:49:11:
in cc_library rule @proxy_wasm_cpp_host//:base_lib: alias '@proxy_wasm_cpp_host//bazel:crypto_system' referring to target 
'@proxy_wasm_cpp_host//bazel:linux_s390x' is not visible from target '@proxy_wasm_cpp_host//:base_lib'. Check the visibility 
declaration of the former target if you think the dependency is legitimate
ERROR: /usr/local/google/home/gregce/.cache/bazel/_bazel_gregce/160b5d944412957a853c062174d7f0d8/external/proxy_wasm_cpp_host/BUILD:49:11:
 Analysis of target '@proxy_wasm_cpp_host//:base_lib' failed
ERROR: Analysis of target '@proxy_wasm_cpp_host//:base_lib' failed; build aborted: 
```

This leads to [these config settings](https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/bazel/BUILD) which need broader visibility. One option is adding this to their package:

```
package(default_visibility = ["//visibility:public"])
```

Other options [here](https://bazel.build/concepts/visibility).

[...]

PR #23434, opened 2022-10-11
deps: Bump `v8` to 10.7.193.13 and `proxy_wasm_cpp_host` to b0a0594

dio:
Commit Message: Update `v8` to 10.7.193.13 and `proxy_wasm_cpp_host` to https://github.com/proxy-wasm/proxy-wasm-cpp-host/commit/b0a0594e8ab4023857393fab349170fb487ec5a9.
Risk Level: Medium (but mostly wasm related)
Additional Description: In the process, we remove `com_googlesource_chromium_zlib`, since now v8 can be compiled without it and can be ignored: https://github.com/envoyproxy/envoy/issues/23258#issuecomment-1264980271.
 
Fixes #23258 #23390

Signed-off-by: Dhi Aurrahman <dio@rockybars.com>

PR #23232, opened 2022-09-23
deps: bump `com_github_wasmtime` -> 1.0.0

RiverPhillips:
Commit Message: deps: bump `com_github_wasmtime` -> 1.0.0
Additional Description: Resolves #23193
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #23460, opened 2022-10-12
deps: update proxy-wasm-cpp-host

keith:
Fixes https://github.com/envoyproxy/envoy/issues/23390


keith:
not sure what I'm in for here

phlax:
cc @kfaseela

dio:
@phlax do we want to land this before https://github.com/envoyproxy/envoy/pull/23434?

keith:
I removed the v8 version bump from this one to only take that bazel fix, if we can land yours instead w/o other blockers we should do that and close this one

phlax:
> @phlax @keith did we finish updating proxy-wasm-cpp-host first?

im not sure - previously i believe there was a need to update these deps together (and do some cargo-fu)

it seems like they are not in sync now - altho maybe the important deps are

Envoy's version of wasmtime:

https://github.com/envoyproxy/envoy/blob/7a4bbb3c5b52b84625ad0626f916be1c99399ba8/bazel/repository_locations.bzl#L982-L996

which is using a recent hash version (maybe we should switch to a releaese version there)

current wasmtime upstream

https://github.com/bytecodealliance/wasmtime/releases/tag/v1.0.1

but it doesnt seem like the wasmtime crates have been updated:

https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/master/bazel/cargo/wasmtime/BUILD.bazel#L44

keith:
There are only 2 commits being pulled in with this change https://github.com/proxy-wasm/proxy-wasm-cpp-host/compare/4fcf895fa2433a1cdf20704926b8b7e4039a6a04...25d6a99de9378e65a6bb6a56a2ca5073870ddc3c so if the other change is blocked on anything I would think updating this would be pretty safe

PiotrSikora (approved):
> it seems like they are not in sync now - altho maybe the important deps are

They are not. Wasmtime is out-of-sync and it only works by accident because there were no breaking changes in Wasmtime's C API (updated in #23232), but Envoy is now using C API from a version different than Wasmtime core.

@RiverPhillips is working on fixing that in https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/309, and hopefully we can land it and merge into Envoy before the next release.

Regarding this PR, it can be either merged as-is or it could be dropped in favor of #23434, which also contains those changes and was opened first.

htuch:
Maybe just do https://github.com/envoyproxy/envoy/pull/23434?

keith:
https://github.com/envoyproxy/envoy/pull/23434

PR #23465, opened 2022-10-13
dlb: optimize auto detect logic

daixiang0:

The default of type uint32 in protobuf is 0, so the `else` field can not run into.

Fix the above issue and add tests to cover some cases.

DLB mock functions are in development, so only test the config logic for now.

Commit Message: 
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
+++ b/contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
@@ -36,10 +36,13 @@ DlbConnectionBalanceFactory::createConnectionBalancerFromProto(
   }
 
   const uint& config_id = dlb_config.id();
-  const uint& device_id = detectDlbDevice(config_id, "/dev");
-  if (device_id < 0) {
+  const auto& result = detectDlbDevice(config_id, "/dev");
+  if (!result.has_value()) {
     ExceptionUtil::throwEnvoyException("no available dlb hardware");
-  } else if (device_id != config_id) {
+  }
+
+  const uint& device_id = result.value();
+  if (device_id != config_id) {
     ENVOY_LOG(warn, "dlb device {} is not found, use dlb device {} instead", config_id, device_id);
   }
 
--- a/contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
+++ b/contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
@@ -40,18 +40,12 @@ DlbConnectionBalanceFactory::createConnectionBalancerFromProto(
 #ifdef DLB_DISABLED
   throw EnvoyException("X86_64 architecture is required for Dlb.");
 #else
-  int device_id = 0;
+  int device_id = dlb_config.id();
   Api::OsSysCalls& os_sys_calls = Api::OsSysCallsSingleton::get();
   struct stat buffer;
 
-  if (dlb_config.id()) {
-    device_id = dlb_config.id();
-    const std::string& device_name = fmt::format("/dev/dlb{}", device_id);
-    if (os_sys_calls.stat(device_name.c_str(), &buffer).return_value_ != 0) {
-      ExceptionUtil::throwEnvoyException(fmt::format("dlb hardware {} not found", device_name));
-    }
-  } else {
-    std::string device_name;
+  std::string device_name = fmt::format("/dev/dlb{}", device_id);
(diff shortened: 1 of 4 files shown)


daixiang0:
@adisuissa @wrowe could you help review?

contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
```diff
@@ -40,25 +40,21 @@ DlbConnectionBalanceFactory::createConnectionBalancerFromProto(
 #ifdef DLB_DISABLED
   throw EnvoyException("X86_64 architecture is required for Dlb.");
 #else
-  int device_id = 0;
+  int device_id = dlb_config.id();
```
adisuissa, line 43: Could this be a negative number? Should the config be rejected if it is?

adisuissa (commented):
LGTM, just left a question.
RE how to test this, I don't think the test should create "/dev/..." files.
Can you refactor the code so it reads it from a different location? Then the test will use the functions from https://github.com/envoyproxy/envoy/blob/main/test/test_common/environment.h to create the temp files.

contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
daixiang0: Here the id is `uint32`, which cannot be negative, refer to https://stackoverflow.com/questions/765916/is-there-ever-a-good-time-to-use-int32-instead-of-sint32-in-google-protocol-buff

daixiang0:
@adisuissa I have refactored it and added new tests, please review and merge if possible.

contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
```diff
@@ -37,35 +35,17 @@ DlbConnectionBalanceFactory::createConnectionBalancerFromProto(
         "please decrease the number of threads by `--concurrency`");
   }
 
+  const int& config_id = dlb_config.id();
```
adisuissa, line 38: Sorry for not being clear earlier.
The [id](https://github.com/envoyproxy/envoy/blob/1ad89a900b9239539820d9ff5c322414efa4fd07/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto#L32) is supposed to be unsigned int. Why is it being converted to a signed int?

contrib/network/connection_balance/dlb/source/connection_balancer_impl.h
```diff
@@ -49,6 +51,33 @@ class DlbBalancedConnectionHandlerImpl : public Envoy::Network::BalancedConnecti
   Envoy::Event::FileEventPtr dlb_event_;
 };
 
+// The dir should always be "/dev" in production.
+// For test it is a temporary directory.
+// Return Dlb device id, -1 means error.
+static int detectDlbDevice(const int& config_id, const std::string& dir) {
```
adisuissa, line 57: nit: no need to pass int by ref

contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc
daixiang0: Sure, I will update it.

contrib/network/connection_balance/dlb/source/connection_balancer_impl.h
```diff
@@ -49,6 +51,33 @@ class DlbBalancedConnectionHandlerImpl : public Envoy::Network::BalancedConnecti
   Envoy::Event::FileEventPtr dlb_event_;
 };
 
+// The dir should always be "/dev" in production.
+// For test it is a temporary directory.
+// Return Dlb device id, absl::nullopt means error.
+static absl::optional<uint> detectDlbDevice(const uint config_id, const std::string& dir) {
```
adisuissa, line 57: nit: you can use `absl::StatusOr` as it allows you to specify the error.

adisuissa (approved):
LGTM, left a comment for future reference.

contrib/network/connection_balance/dlb/source/connection_balancer_impl.h
daixiang0: Sorry for slow response due to sickness, I will optimize it in the next PR.

PR #23470, opened 2022-10-13
Revert "listener filter: hide envoy internals from ListenerFilterCallbacks interface (#22732)

soulxu:

Commit Message: Revert "listener filter: hide envoy internals from ListenerFilterCallbacks interface (#22732)
Additional Description:
https://envoyproxy.slack.com/archives/C78HA81DH/p1665661946626769 there is a case showing up, people still need a way to trigger the listener filter iteration in the listener filter.

This reverts commit 350c7f947149fcff41c4bdeac633aab80ae207fe.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/envoy/network/filter.h
+++ b/envoy/network/filter.h
@@ -287,6 +287,20 @@ public:
   virtual ConnectionSocket& socket() PURE;
 
   /**
+   * @return the Dispatcher for issuing events.
+   */
+  virtual Event::Dispatcher& dispatcher() PURE;
+
+  /**
+   * If a filter stopped filter iteration by returning FilterStatus::StopIteration,
+   * the filter should call continueFilterChain(true) when complete to continue the filter chain,
+   * or continueFilterChain(false) if the filter execution failed and the connection must be
+   * closed.
+   * @param success boolean telling whether the filter execution was successful or not.
+   */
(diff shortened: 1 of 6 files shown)


soulxu:
cc @mattklein123 @ggreenway

soulxu:
/wait

need to ensure that case is valid first.

soulxu:
Or I should just expose the continueFilterChain again instead of revert this PR if the `dispatch()` consider as not expose the listener filter.

soulxu:
@mattklein123 looking for feedback on this. This is removed in 1.24 release, if we think it is a valid case, we can add it back before the release. I can also submit a PR and only add that method back since the original PR did some cleanup and refactoring. But if we rush the time, it is also ok to revert this PR.

soulxu:
cc @phlax also, in case we want to bring this in before the release

KBaichoo:
@soulxu I think given this wasn't release blocking, and has some other improvements we can just enable the use case vs reverting all of this. Thank you for handling this!

soulxu:
> @soulxu I think given this wasn't release blocking, and has some other improvements we can just enable the use case vs reverting all of this. Thank you for handling this!

got it, thanks! I will add another pr enable this.

PR #23474, opened 2022-10-13
load balancer: update related methods for load balancer extensions

### LINKED ISSUE #20634: xds: Envoy should support its built-in LB policies as extensions

wbpcode:
Commit Message: load balancer: update related methods for load balancer extensions
Additional Description:

Part of work to land #20634. This is split out from #23472 and is some necessary pre-work for adding load balancer extension.

I made this as an independent PR because this updates some core code and I wish this could be reviewed separately.

Risk Level: mid.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


envoy/upstream/load_balancer.h
```diff
@@ -158,9 +168,15 @@ class LoadBalancerFactory {
...
-   * @return LoadBalancerPtr a new load balancer.
+   * @return LoadBalancerPtr a new worker local load balancer.
+   * TODO(wbpcode): remove this method in the future and used the new method below.
    */
   virtual LoadBalancerPtr create() PURE;
+
+  /**
+   * @return LoadBalancerPtr a new worker local load balancer.
+   */
+  virtual LoadBalancerPtr create(LoadBalancerParams params) PURE;
```
wbpcode, line 179: Why this is necessary? 
If the `load_balancing_policy` is enabled, all Lb type will be treated as `ThreadAwareLoadBalancer`. But for Lb type like `least request`, thread local `priority_set` and `local_priority_set` are necessary. So, we must add this `params` to help the factory to create these types of Lb.

envoy/upstream/upstream.h
```diff
@@ -836,11 +836,10 @@ class ClusterInfo : public Http::FilterChainFactory {
...
 
   /**
-   * @return const envoy::config::cluster::v3::LoadBalancingPolicy_Policy& the load balancing policy
-   * to use for this cluster.
+   * @return const ProtobufWkt::Message& the validated load balancing policy configuration to use
+   * for this cluster.
    */
-  virtual const envoy::config::cluster::v3::LoadBalancingPolicy_Policy&
-  loadBalancingPolicy() const PURE;
+  virtual const ProtobufTypes::MessagePtr& loadBalancingPolicy() const PURE;
```
wbpcode, line 843: Why we need this change?
Load balancer will be created at runtime in every worker. We shouldn't use non-validated raw `Policy` to create load balancers. Uncaught exceptions may be throwed.  
So, we should unpack and validate this typed config in the configure phase.

wbpcode:
Hmmm, the ci take too much time run the coverage tests. Seems that the miss-rate of remote cache is very high because I update the `upstream.h` and `local_balancer.h`.

Anyway, this PR shouldn't be merged before 1.24.0 release.

/wait

mattklein123 (approved):
LGTM, thanks.

PR #23485, opened 2022-10-13
xDS: add xDS config tracker extension point

botengyao:

This extension added hooks to allow tracking xDS responses or resources in external components,
e.g., external tracer or monitor. It provides the process point when receive, ingest, or fail to process xDS resources and messages.

Commit Message:
Additional Description:
Risk Level: L
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -342,6 +342,10 @@ message Bootstrap {
   // TODO(abeyad): Add public-facing documentation.
   // [#not-implemented-hide:]
   core.v3.TypedExtensionConfig xds_delegate_extension = 35;
+
+  // Optional XdsConfigTracer configuration, which allows tracking xDS resources and responses.
```
htuch, line 346: Can you add some documentation explaining how this is intended to be used? Thanks.

botengyao: Thanks for taking a look! Sure, will do and it is still WIP and will make this PR more polished.

envoy/config/xds_config_tracer.h
```diff
@@ -0,0 +1,96 @@
...
+public:
+  virtual ~XdsConfigTracer() = default;
+
+  /**
+   * Log the decoded SotW xDS resources that are about to be ingested.
+   * @param type_url The type url of xDS message.
+   * @param resources List of decoded resources that reflect the latest state.
+   * @param details The log point state and details.
+   */
+  virtual void log(const absl::string_view type_url,
```
htuch, line 54: Why is this "log" and not just a generic hook (e.g. "onConfigIngestion")  in the xDS ingestion lifecycle? Is there some documentation on when this is called relative to the xDS delegate, etc?

botengyao: Thanks for taking a look! And updated the PR. There is a high level doc internally, but doesn't include the details related to this PR. I will add more comments in the PR and here. The initial purpose is to add some log point and persist the log to adapt the external package, therefore `log` is used.

For both SotW and Delta response message, we will have three log points: RCEIVE, INGESTED, FAILED. 
For SotW, the first log method is used twice, one is at the time when the resource is successfully parsed and before the ingestion. The other one is when the resource is successfully applied to Envoy and are about to be ACK'ed. The TraceDetails implies the status of ingestion process, e.g., TraceState::INGESTED. The log point TraceState::INGESTED is the same place as xDS delegate, like we can use `onConfigIngestion`. 

Similar to Delta, the log point is distinguished by the TraceState: RCEIVE, INGESTED, FAILED. The extension is more like a logger, but it could be used by other purpose. This is my initial thinking, happy to learn any thoughts. Thank you!

adisuissa: I think what @htuch implies is that this can be generalized to some other use-cases that would want to know the state of resources during the config ingestion process. In other words, not only for tracing/logging, but more generic.

Specifically, the callback methods can be called `resourcesProcessingUpdate`/`configProcessingUpdate` and `TraceState` could be renamed to `ProcessingState` (as it is already under the `Envoy::Config` namespace).

envoy/config/xds_config_tracer.h
```diff
@@ -0,0 +1,111 @@
...
+                   const std::vector<DecodedResourcePtr>& resources,
+                   const TraceDetails& details) PURE;
+
+  /**
+   * Log point for SotW xDS discovery response.
+   *
+   * This is mainly callded when there is any exception during the parse and ingestion process
```
adisuissa, line 69: Can you give a more detailed description instead of saying `mainly`?

envoy/config/xds_config_tracer.h
```diff
@@ -0,0 +1,111 @@
...
+   * for SotW. The TraceDetails includes the error details and TraceState.
+   *
+   * @param message The SotW discovery response message body.
+   * @param details The log point state and details.
+   */
+  virtual void log(const envoy::service::discovery::v3::DiscoveryResponse& message,
+                   const TraceDetails& details) PURE;
+
+  /**
+   * Log point for Delta xDS discovery response message when it is received, successfully applied,
```
adisuissa, line 79: The asymmetry between SotW (2 callbacks) vs Delta (1 callback) seems a bit strange.
Could there be a more symmetric approach here?
In other words, why the first `log` function is only called for SotW?

envoy/config/xds_config_tracer.h
```diff
@@ -0,0 +1,111 @@
...
+                   const TraceDetails& details) PURE;
+
+  /**
+   * Log point for SotW xDS discovery response.
+   *
+   * This is mainly callded when there is any exception during the parse and ingestion process
+   * for SotW. The TraceDetails includes the error details and TraceState.
+   *
+   * @param message The SotW discovery response message body.
+   * @param details The log point state and details.
```
adisuissa, line 73: Can this be called for something other than "FAILED" state?

envoy/config/xds_config_tracer.h
```diff
@@ -0,0 +1,111 @@
...
+ * An interface for hooking into xDS update events to provide the ablility to use some
+ * external logger or processor in xDS update. This tracer provides the log point when
+ * the resources are received , when they are successfully processed and applied, and when
+ * there is any failure.
+ *
+ * Instance of this interface get invoked on the main Envoy thread. Thus, it is important
+ * for implementations of this interface to not execute any blocking operations on the same
+ * thread.
+ */
+class XdsConfigTracer {
```
adisuissa, line 46: As discussed in an external comment, this can probably be more generic any config processing operation (not only tracing/logging).

envoy/config/subscription.h
```diff
@@ -59,6 +59,11 @@ class DecodedResource {
    * @return bool does the xDS discovery response have a set resource payload?
    */
   virtual bool hasResource() const PURE;
+
+  /**
+   * @return absl::optional<envoy::config::core::v3::Metadata> of a resource.
+   */
+  virtual absl::optional<envoy::config::core::v3::Metadata> metadata() const PURE;
```
adisuissa, line 66: Should this return const OptRef?

adisuissa (commented):
Thanks for working on this, very useful!
Left a few high-level design comments.

envoy/config/xds_config_tracer.h
htuch: Yep - I don't think we need any deep changes, but just to model this hook as part of the xDS configuration ingestion lifecycle, rather than being super opinionated about this being a log or trace thingy (maybe we just want to bump a stat, or post an event to some Dispatcher as part of on-demand xDS etc.)

botengyao: Gotcha, generalization makes sense. Will change & thank you all!

envoy/config/xds_config_tracer.h
botengyao: Done.

envoy/config/xds_config_tracer.h
botengyao: Re-modeled this hook. Can you take another pass? Thank you!

botengyao:
Hi @adisuissa, I remodeled the hook, and now it should look better, and PTAL. Thanks!
The CodeQL build failure is not related to my changes.

envoy/config/xds_config_tracker.h
```diff
@@ -0,0 +1,115 @@
...
+   * @param added_resources A list of decoded resources to add to the current state.
+   * @param removed_resources A list of resources to remove from the current state.
+   */
+  virtual void
+  onConfigIngested(const absl::string_view type_url,
+                   const std::vector<DecodedResourcePtr>& added_resources,
+                   const Protobuf::RepeatedPtrField<std::string>& removed_resources) PURE;
+
+  /**
+   * Invoked when received or failed to process the DiscoveryResponse.
```
htuch, line 72: This is a bit of a confusing name / semantics. First, what is a "failed" `DiscoveryResponse`? Second, how do we think about this lifecycle-wise vs. `onConfigIngested`? Always first right?

An xDS config tracker is tracking the applied configuration, not the transport protocol or proxy NACKs due to rejections I think. Or does it do this as well? I think being clear on the mission here, and whether we are tracking the applied configuration, vs. the wire configuration, is really important to a clean interface.

htuch: @adisuissa as well

botengyao: We tracked the ingested configurations, and also provide the hooks for the message level when receiving or there is any exception during processing the response.

botengyao: Thanks for reviewing!

This hook is like we will throw an exception when we fail to decode and apply the configs from `DiscoveryResponse`, the hook is applied there and when receiving the message as well. I am thinking we can change the name to `onConfigUpdateReceivedorFailed`. WDYT? Thanks!

htuch: I think it might be better to split this into two separate callbacks for the two cases. That way we have less need for each implementation to switch themselves on the two fairly distinct cases. Also, I still think the different cases are very confusing name wise. Here would be my suggestion, three methods:
* `onConfigReceived` which is called when config is first received in the xDS ingestion lifecycle (I think?! please clarify when this is and why we even need to think about this).
* `onConfigRejected` which is called if config is rejected during xDS ingestion.
* `onConfigAccepted` which is called if config is accepted during xDS ingestion (i.e. what `onConfigIngested` is today).

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,291 @@
...
+      ++ReceiveCount;
+      break;
+    case Config::ProcessingState::FAILED:
+      ++FailureCount;
+      break;
+    };
+  }
+};
+
+std::atomic<int> TestXdsConfigTracker::ReceiveCount;
```
adisuissa, line 80: I think this can be done by introducing these as test counters (see https://github.com/envoyproxy/envoy/blob/f4144e8b02181d927fe4bea5eba0e95316b5ef7a/test/integration/filters/test_network_filter.cc#L14 as an example).

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  void acceptXdsConnection() {
+    createXdsConnection();
+    AssertionResult result = xds_connection_->waitForNewStream(*dispatcher_, xds_stream_);
+    RELEASE_ASSERT(result, result.message());
+    xds_stream_->startGrpcStream();
+  }
+
+  std::string getRuntimeKey(const std::string& key) {
```
adisuissa, line 179: Why is the runtime and the admin needed for this test?
Can you use something that doesn't require it, maybe sending back a static cluster?

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,291 @@
...
+
+  std::string name() const override { return "envoy.config.xds.test_xds_tracer"; };
+
+  Config::XdsConfigTrackerPtr createXdsConfigTracker(const ProtobufWkt::Any&,
+                                                     ProtobufMessage::ValidationVisitor&) override {
+    return std::make_unique<TestXdsConfigTracker>();
+  }
+};
+
+class XdsConfigTrackerIntegrationTest : public Grpc::DeltaSotwIntegrationParamTest,
```
adisuissa, line 101: Please also test this against the unified mux

envoy/config/xds_config_tracker.h
adisuissa: +1 on using explicit callback methods rather than sending the state.

Feel free to correct me if I'm wrong, but compared to the current state of the protocol, does this tracker essentially adds the ability to know when a resource was received, or is it something else?
In other words, at the moment, given a response, Envoy processes it and either accepts (all resources "ingested") or rejects the response. This tracker allows to know when a resource arrived (received) at the Envoy.
Can you provide more details around the usefulness of this new info?

source/common/config/decoded_resource_impl.h
```diff
@@ -91,6 +98,7 @@ class DecodedResourceImpl : public DecodedResource {
   const std::string version_;
   // Per resource TTL.
   const absl::optional<std::chrono::milliseconds> ttl_;
+  const OptRef<const envoy::config::core::v3::Metadata> metadata_;
```
adisuissa, line 101: Can you add a comment about the lifetime of the original metadata, and why it is ok to keep a ref here.

envoy/config/xds_config_tracker.h
adisuissa: Another thought of where things may be different is if the Envoy received a resource that it hasn't subscribed for, and so you'll see it on the "received" side, but not on the processed?

source/common/config/watch_map.cc
```diff
@@ -242,6 +242,11 @@ void WatchMap::onConfigUpdate(
       cur_watch->callbacks_.onConfigUpdate({}, {}, system_version_info);
     }
   }
+
+  // Processing point when resources are successfully ingested.
+  if (xds_config_tracker_.has_value()) {
+    xds_config_tracker_->onConfigIngested(type_url_, decoded_resources, removed_resources);
+  }
```
adisuissa, line 249: Why not do this inside the new_grpc_mux_impl?

test/integration/xds_config_tracker_integration_test.cc
botengyao: Thanks! Added the stats counter for the test.

test/integration/xds_config_tracker_integration_test.cc
botengyao: Done. the test is using RTDS to call the extension, and removed the admin.

test/integration/xds_config_tracker_integration_test.cc
botengyao: Done for the unified mux.

source/common/config/watch_map.cc
botengyao: Changed. For `legacy` and `unified` mux, I think the subscription state is the best place to put the `onConfigAccepted`. It is called right after the watch map callbacks, which represents the actual accepted resources.
Before I am thinking to pass the decoded_resources, but for delta we can put under `handleGoodResponse`.

envoy/config/xds_config_tracker.h
botengyao: Thank you all for the input. I am aligned with the `onConfigReceived` may not add too much value here. We care more about the accepted configs and any exceptions, and if there is any error or unsubscribed type_url, we either directly NACK or throw an exception.

source/common/config/watch_map.cc
botengyao: PTAL. Thank you!

htuch:
API now LGTM, will defer to @adisuissa on the rest. Thanks.

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,248 @@
...
+  COUNTER(on_config_rejected)
+
+/**
+ * Struct definition for stats. @see stats_macros.h
+ */
+struct TestXdsTrackerStats {
+  ALL_TEST_XDS_TRACKER_STATS(GENERATE_COUNTER_STRUCT)
+};
+
+class TestXdsConfigTracker : public Config::XdsConfigTracker {
```
adisuissa, line 39: Please add comments describing what this class is and what is it used for.

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,248 @@
...
+  XdsConfigTrackerIntegrationTest()
+      : HttpIntegrationTest(Http::CodecType::HTTP2, ipVersion(),
+                            ConfigHelper::baseConfigNoListeners()) {
+
+    use_lds_ = false;
+    create_xds_upstream_ = true;
+    sotw_or_delta_ = sotwOrDelta();
+
+    config_helper_.addRuntimeOverride("envoy.reloadable_features.unified_mux",
+                                      (this->sotwOrDelta() == Grpc::SotwOrDelta::Sotw ||
```
adisuissa, line 99: Should this be `UnifiedDelta`?

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,248 @@
...
+TEST_P(XdsConfigTrackerIntegrationTest, XdsConfigTrackerFailureCount) {
+  TestXdsConfigTrackerFactory factory;
+  Registry::InjectFactory<Config::XdsConfigTrackerFactory> registered(factory);
+
+  initialize();
+  acceptXdsConnection();
+
+  EXPECT_TRUE(compareDiscoveryRequest(Config::TypeUrl::get().Runtime, "", {"some_rtds_layer"},
+                                      {"some_rtds_layer"}, {}, true));
+  auto some_rtds_layer = TestUtility::parseYaml<envoy::service::runtime::v3::Runtime>(R"EOF(
```
adisuissa, line 211: Is this used? if not, let's remove.

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,248 @@
...
+    name: different_rtds_layer
+    layer:
+      foo: bar
+      baz: meh
+  )EOF");
+
+  auto route_config = TestUtility::parseYaml<envoy::config::route::v3::RouteConfiguration>(R"EOF(
```
adisuissa, line 218: nit: const

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,248 @@
...
+        api_config_source:
+          api_type: GRPC
+          transport_api_version: V3
+          grpc_services:
+            envoy_grpc:
+              cluster_name: xds_cluster
+    )EOF");
+
+  FakeStream* stream = xds_stream_.get();
+  envoy::service::discovery::v3::DiscoveryResponse discovery_response;
```
adisuissa, line 232: This seems odd, for non-SotW responses. Should this cover both Delta and SotW?

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,248 @@
...
+
+  void initialize() override {
+    // The tests infra expects the xDS server to be the second fake upstream, so
+    // we need a dummy data plane cluster.
+    setUpstreamCount(1);
+    setUpstreamProtocol(Http::CodecType::HTTP2);
+    HttpIntegrationTest::initialize();
+    registerTestServerPorts({});
+    initial_xds_update_ = test_server_->counter("test_xds_tracker.on_config_accepted")->value();
+    initial_xds_reject_ = test_server_->counter("test_xds_tracker.on_config_rejected")->value();
```
adisuissa, line 163: Why is this needed?

test/common/config/delta_subscription_test_harness.h
```diff
@@ -51,12 +52,14 @@ class DeltaSubscriptionTestHarness : public SubscriptionTestHarness {
       xds_context_ = std::make_shared<Config::XdsMux::GrpcMuxDelta>(
           std::unique_ptr<Grpc::MockAsyncClient>(async_client_), dispatcher_, *method_descriptor_,
           random_, stats_store_, rate_limit_settings_, local_info_, false,
-          std::make_unique<NiceMock<MockCustomConfigValidators>>());
+          std::make_unique<NiceMock<MockCustomConfigValidators>>(),
+          /*xds_config_tracker=*/XdsConfigTrackerOptRef());
```
adisuissa, line 56: Why is this not providing a mock object?

test/common/config/decoded_resource_impl_test.cc
```diff
@@ -49,6 +49,28 @@ TEST(DecodedResourceImplTest, All) {
     EXPECT_EQ("foo", decoded_resource.version());
     EXPECT_THAT(decoded_resource.resource(), ProtoEq(ProtobufWkt::Empty()));
     EXPECT_TRUE(decoded_resource.hasResource());
+    EXPECT_FALSE(decoded_resource.metadata().has_value());
+  }
+
+  {
```
adisuissa, line 55: Please add a comment describing what this code block is validating

test/integration/xds_config_tracker_integration_test.cc
botengyao: Thanks! Done.

test/integration/xds_config_tracker_integration_test.cc
botengyao: My bad. Yes, we need to simulate both, there is an existing template method but is type strict, and this simulate an exception scenario.

test/integration/xds_config_tracker_integration_test.cc
botengyao: We can delete it for my test.

test/common/config/delta_subscription_test_harness.h
botengyao: Emmm, an `OptRef` of an abstract class is okay for the test, right? We are testing it at least from the integration test to verify the extension method is called.

botengyao:
Hi @adisuissa, can you take another look? Thanks!

jmarantz:
Adi -- any more comments?

envoy/config/xds_config_tracker.h
```diff
@@ -0,0 +1,105 @@
...
+#include "source/common/protobuf/protobuf.h"
+
+#include "google/rpc/status.pb.h"
+
+namespace Envoy {
+namespace Config {
+
+/**
+ * An interface for hooking into xDS update events to provide the ablility to use some external
+ * processor in xDS update. This traker provides the process point when the discovery response
```
adisuissa, line 21:
```suggestion
 * processor in xDS update. This tracker provides the process point when the discovery response
```

envoy/config/xds_config_tracker.h
```diff
@@ -0,0 +1,105 @@
...
+ * is received, when the resoures are successfully processed and applied, and when there is any
```
adisuissa, line 22:
```suggestion
 * is received, when the resources are successfully processed and applied, and when there is any
```

envoy/config/xds_config_tracker.h
```diff
@@ -0,0 +1,105 @@
...
+ */
+class XdsConfigTracker {
+public:
+  virtual ~XdsConfigTracker() = default;
+
+  /**
+   * Invoked when SotW xDS configuration updates have been successfully parsed, applied on
+   * the Envoy instance, and are about to be ACK'ed.
+   *
+   * For SotW, resources are all the resouces except for the heart-beat ones in the original
```
adisuissa, line 37:
```suggestion
   * For SotW, the passed resources contain all the received resources except for the heart-beat ones in the original
```

envoy/config/xds_config_tracker.h
```diff
@@ -0,0 +1,105 @@
...
+   * @param resources A list of decoded resources to add to the current state.
+   */
+  virtual void onConfigAccepted(const absl::string_view type_url,
+                                const std::vector<DecodedResourcePtr>& resources) PURE;
+
+  /**
+   * Invoked when Delta xDS configuration updates have been successfully accepted, applied on
+   * the Envoy instance, and are about to be ACK'ed.
+   *
+   * For Delta, added_resources are all the received added resouces except for the heart-beat ones
```
adisuissa, line 51:
```suggestion
   * For Delta, added_resources contains all the received added resources except for the heart-beat ones
```

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,256 @@
...
+    resource.mutable_resource()->PackFrom(route_config);
+    resource.set_name("my_route");
+    resource.set_version("1");
+    resources.emplace_back(resource);
+    const auto delta_discovery_response = createExplicitResourcesDeltaDiscoveryResponse(
+        Config::TypeUrl::get().Runtime, resources, {});
+
+    // Message's TypeUrl != Resource's
+    stream->sendGrpcMessage(delta_discovery_response);
+  }
```
adisuissa, line 249: I'm probably missing something here, but why can't this just be a call to `sendDiscoveryResponse`. I'm thinking of something similar to the following:
```
sendDiscoveryResponse<envoy::service::runtime::v3::Runtime>(
      Config::TypeUrl::get().Runtime, {route_config}, {route_config}, {}, "1");
```
Will the above work instead of this code?

adisuissa (commented):
Thanks, overall LGTM.
Some nits about comments, and a question about the test.
/wait

test/integration/xds_config_tracker_integration_test.cc
botengyao: This is a template method, and the type of resources must be same as the message's type_url as well as the template type. My code simulates a simple exception scenario which the types are not same.

botengyao:
Hi @adisuissa, can you take another pass? Thanks!

adisuissa (dismissed):
Thanks!
/lgtm api

test/integration/xds_config_tracker_integration_test.cc
botengyao: Thanks! `sendDiscoveryResponse<envoy::config::route::v3::RouteConfiguration>( Config::TypeUrl::get().Runtime, {route_config}, {route_config}, {}, "1");` works. I should know this.

botengyao:
Maybe /assign @mattklein123

adisuissa (changes requested):
/wait

envoy/config/xds_config_tracker.h
```diff
@@ -0,0 +1,109 @@
...
+                                const std::vector<DecodedResourcePtr>& resources) PURE;
+
+  /**
+   * Invoked when Delta xDS configuration updates have been successfully accepted, applied on
+   * the Envoy instance, and are about to be ACK'ed.
+   *
+   * For Delta, added_resources contains all the received added resources except for the heart-beat
+   * ones in the original message, and the removed resouces are the same in the xDS message.
+   *
+   * Note: this method is called when *all* the resouces in a response are accepted.
```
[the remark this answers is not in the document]
botengyao: Thanks! Done

test/integration/xds_config_tracker_integration_test.cc
```diff
@@ -0,0 +1,236 @@
...
+  EXPECT_EQ(1, test_server_->counter("test_xds_tracker.on_config_rejected")->value());
+}
+
+TEST_P(XdsConfigTrackerIntegrationTest, XdsConfigTrackerPartialUpdate) {
+  TestXdsConfigTrackerFactory factory;
+  Registry::InjectFactory<Config::XdsConfigTrackerFactory> registered(factory);
+
+  initialize();
+  // The first of duplicates has already been successfully applied, and a duplicate exception should
+  // be threw. Only the first cluster is added.
```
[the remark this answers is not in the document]
botengyao: Thanks! modified, for SotW 2 will be added, while for Delta, the message is denied at the beginning if there is any duplication.

adisuissa (approved):
LGTM, thanks!

botengyao:
Hi @mattklein123, @lizan, PTAL and can we get this merged? Thanks!

api/envoy/config/bootstrap/v3/bootstrap.proto
```diff
@@ -342,6 +342,13 @@ message Bootstrap {
   // TODO(abeyad): Add public-facing documentation.
   // [#not-implemented-hide:]
   core.v3.TypedExtensionConfig xds_delegate_extension = 35;
+
+  // Optional XdsConfigTracker configuration, which allows tracking xDS responses in external components,
+  // e.g., external tracer or monitor. It provides the process point when receive, ingest, or fail to
+  // process xDS resources and messages.
+  // If a value is not specified, no XdsConfigTracker will be used.
+  // [#not-implemented-hide:]
```
mattklein123, line 350: Why is this not-implemented-hide? Isn't it implemented? Also, can you please add a release note? Thank you.

/wait

botengyao: Thanks! This PR only added an api extension interface right now, and there is no actual extension implementation. This enables the ability to extend this interface for different use case. Should I add a release note only for this interface?

mattklein123: The fact that it can be implemented means it's implemented, so I would remove the not-implemented-hide and still add a release note for it. It would be also good to `:repo:` link to the test extension so people can look at an example of the API, and I would also mention there are no in-repo extensions currently, etc.

botengyao: Thanks! Done, and waiting CI.

mattklein123 (approved):
Thanks!

PR #23489, opened 2022-10-13
outh2: passthrough of authorization header when filter is matched

### LINKED ISSUE #23080: [OAuth2] Cannot passthrough on authorization header
*Title*: OAuth2 - Cannot passthrough on authorization header

*Description*:
We're using the envoy.filters.http.oauth2 filter in some of our sidecar proxies for protecting web applications. Since a recent version, the passthrough configuration we are using does not work anymore:
```
              pass_through_matcher:
                - name: authorization
                  prefix_match: Bearer
```
This was needed to let requests issued with existing bearer token (from mobile applications for example) to skip the oauth2 filter. Now when a client issues a request with a bearer token, instead of bypassing the oauth2 filter and routing the request to the upstream service, the client now receives a redirection response.

The bug (or feature?) was introduced in May 2022, this line here seems to be the culprit: 
```
headers.removeInline(authorization_handle.handle());
```
(https://github.com/envoyproxy/envoy/blame/main/source/extensions/filters/http/oauth2/filter.cc#L233)

A potential fix could be to sanitize the authorization header after (instead of before) evaluating the passthrough conditions.

timo-haas-by:
Commit Message: 
Authorization header sanitation altered headers before the passthrough evaluation(removes the authorization header if present). Therefore, move of passthrough header matching to the beginning if the header evaluation, so the unaltered request is passed through.
Introduces a new metric: oauth_passthrough counter.

Risk Level: low
Testing: unit test
Docs Changes: added
Release Notes: added
fixes #23080
fixes commit e0c52894b0adaee575e3e981814bece8ece58b40

Signed-off-by: Timo Haas <haastimo@gmx.de>


changelogs/current.yaml
```diff
@@ -96,6 +96,9 @@ minor_behavior_changes:
 - area: dynamic_forward_proxy
   change: |
     No longer waiting on DNS responses in the dynamic forward proxy filter if upstream proxying is turned on. This behaviorial change can be reverted by setting runtime guard ``envoy.reloadable_features.skip_dns_lookup_for_proxied_requests`` to false.
+- area: oauth
+  change: |
+    Passthrough request do now have their own metric ``oauth_passthrough`` and aren't included in ``oauth_success`` anymore.
```
wbpcode, line 101: `aera: oauth2`

changelogs/current.yaml
```diff
@@ -126,6 +129,9 @@ bug_fixes:
 - area: tls
   change: |
     fixed a bug where, when runtime guard ``envoy.reloadable_features.tls_async_cert_validation`` is set to false, the wrong TLS alerts would sometimes be sent in response to certificate validation failures.
+- area: oauth
```
wbpcode, line 132: ditto.

changelogs/current.yaml
```diff
@@ -126,6 +129,9 @@ bug_fixes:
...
+  change: |
+    fixed a bug when passthrough header was matched, envoy would always remove the authorization header.
```
wbpcode, line 134: Is this a bug fix or behavior change? And maybe this will break some exist users? 🤔

timo-haas-by: I think it is a bug fix:
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/oauth2_filter.html#notes

> pass_through_matcher provides an interface for users to provide specific header matching criteria such that, when applicable, the OAuth flow is **entirely** skipped. When this occurs, the oauth_success metric is still incremented.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -221,12 +221,23 @@ const std::string& OAuth2Filter::bearerPrefix() const {
...
- * 3) user is authorized
- * 4) user is unauthorized
+ * 1) pass through header is present
+ * 2) user is signing out
+ * 3) /_oauth redirect
+ * 4) user is authorized
+ * 5) user is unauthorized
  */
 Http::FilterHeadersStatus OAuth2Filter::decodeHeaders(Http::RequestHeaderMap& headers, bool) {
+  // Stop Filter if Passthrough header is present
```
snowp, line 231: I wouldn't use the term Stop here since it makes it sound like we're going into Stop iteration

Maybe something like `// Do nothing if this matches one of the pass through matchers`?

changelogs/current.yaml
```diff
@@ -96,6 +96,9 @@ minor_behavior_changes:
 - area: dynamic_forward_proxy
   change: |
     No longer waiting on DNS responses in the dynamic forward proxy filter if upstream proxying is turned on. This behaviorial change can be reverted by setting runtime guard ``envoy.reloadable_features.skip_dns_lookup_for_proxied_requests`` to false.
+- area: oauth2
+  change: |
+    Passthrough request do now have their own metric ``oauth_passthrough`` and aren't included in ``oauth_success`` anymore.
```
snowp, line 101:
```suggestion
    Passthrough request now have their own metric ``oauth_passthrough`` and aren't included in ``oauth_success`` anymore.
```

test/extensions/filters/http/oauth2/filter_test.cc
```diff
@@ -574,10 +576,30 @@ TEST_F(OAuth2Test, OAuthOptionsRequestAndContinue) {
       {Http::Headers::get().Path.get(), "/anypath"},
       {Http::Headers::get().Method.get(), Http::Headers::get().MethodValues.Options},
   };
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue, filter_->decodeHeaders(request_headers, false));
+}
+
+/**
+ * Scenario: Authorization header is one of the headers in pass_through_matcher
+ *
+ * Expected behavior: Passthrough unaltered request (especially Authorization header)
```
snowp, line 585: This test makes it sound like the behavior changes when the authorization header is used as one of the matchers, but I don't think that's the case? Can we just include a check to the existing pass through test that shows that the authorization header is left untouched?

changelogs/current.yaml
snowp: Sounds like the docs there will need to be updated with this PR since we're no longer incrementing success

While the behavior in this PR seems more correct than before, it does change the behavior slightly and might surprise people. Can we add a runtime check to this change to ensure that people have a way out in case they were relying on this behavior (be it the stats or the sanization)? https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#runtime-guarding

timo-haas-by:
Will look into runtime guarding.

timo-haas-by:
@snowp Is there something else I should do?

snowp (approved):
Thanks, this LGTM!

timo-haas-by:
@snowp @rgs1 Is there a timeline for merging this PR?

PR #23490, opened 2022-10-13
tracing: fixing a regression in router initiated spans

alyssawilk:
Fixes an issue where route spans were not performed correctly when HCM config was absent. Now if HCM is not configured, the egress config defaults will again be used.

Risk Level: low
Testing: integration test
Docs Changes: n/a
Release Notes: n/a
Fixes commit #22987


RyanTheOptimist (approved):

mattklein123 (approved):
Thank you as always Alyssa!

wbpcode:
Thanks for this fix. But seems that this issue is not introduced by the #22987. So, we may need do more backports. I could discuss this problem with you offline.

PR #23498, opened 2022-10-14
oauth: fix the Max-Age of OAuth cookies

### LINKED ISSUE #23492: Max-Age in oauth2 cookies is set to an absolute timestamp instead of a duration relative to the current time
https://github.com/envoyproxy/envoy/blob/817cfee4b1188d4e6ce57b2ab0964c317b36130f/source/extensions/filters/http/oauth2/filter.cc#L419-L421
is used to set the Max-Age field in oauth2 cookies:
https://github.com/envoyproxy/envoy/blob/817cfee4b1188d4e6ce57b2ab0964c317b36130f/source/extensions/filters/http/oauth2/filter.cc#L53
https://github.com/envoyproxy/envoy/blob/817cfee4b1188d4e6ce57b2ab0964c317b36130f/source/extensions/filters/http/oauth2/filter.cc#L444
but it contains an absolute timestamp instead of a duration relative to the current time. This causes Oauth2 cookies to expire after the intended duration, causing Envoy to respond with a 403 when a request is received with expired cookies instead of starting a new Oauth flow.

qfiard:
The Max-Age in OAuth cookies was incorrectly set to an absolute timestamp instead of a duration relative to the current time. This pull request fixes Max-Age while continuing to use an absolute timestamp for the OAuthExpires cookie and HMAC payload.

Fixes #23492


changelogs/current.yaml
```diff
@@ -132,6 +132,9 @@ bug_fixes:
 - area: tls
   change: |
     fixed a bug where, when runtime guard ``envoy.reloadable_features.tls_async_cert_validation`` is set to false, the wrong TLS alerts would sometimes be sent in response to certificate validation failures.
+- area: oauth
+  change: |
+    fixed the Max-Age of OAuth cookies.
```
wbpcode, line 137: could make the change log more detailed?

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -416,19 +416,19 @@ void OAuth2Filter::onGetAccessTokenSuccess(const std::string& access_code,
   id_token_ = id_token;
   refresh_token_ = refresh_token;
 
-  const auto new_epoch = time_source_.systemTime() + expires_in;
-  new_expires_ = std::to_string(
-      std::chrono::duration_cast<std::chrono::seconds>(new_epoch.time_since_epoch()).count());
+  new_expires_ = time_source_.systemTime() + expires_in;
```
wbpcode, line 419: It would be better just to pass the `expires_in` to the `finishFlow()` to avoid re-calculate the MaxAge

wbpcode (changes requested):
Thanks for this contribution. Some comments are added.

/wait

source/extensions/filters/http/oauth2/filter.cc
qfiard: Done.

qfiard:
@wbpcode Should be ready for another pass, thanks for the review.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -430,19 +430,18 @@ void OAuth2Filter::onGetAccessTokenSuccess(const std::string& access_code,
...
-
-  finishFlow();
+  finishFlow(expires_in);
 }
 
-void OAuth2Filter::finishFlow() {
+void OAuth2Filter::finishFlow(std::chrono::seconds expires_in) {
   std::string token_payload;
+  auto new_expires = time_source_.systemTime() + expires_in;
+  std::string new_expires_str = std::to_string(
```
snowp, line 439: Both of these can be `const` I think

snowp (changes requested):
Makes sense to me, just the one comment

source/extensions/filters/http/oauth2/filter.cc
qfiard: Done, also moved these before the definition of token_payload.

snowp (dismissed):
Thanks!

qfiard:
Anything I can do to help get this merged?

snowp (approved):
Thanks!

snowp:
Seems like CI is stuck here, mind pushing an empty commit to trigger a fresh run?

snowp:
Ping here - this needs to pick up latest main to pass CI

KBaichoo:
@qfiard can you merge main to fix CI? Thank you

Issue #22845, opened 2022-08-25
error default value of header append for ext_proc/ext_authz


wbpcode:
The default value of `append` of `HeaderValueOption` should be true according to the API.

```
 // Should the value be appended? If true (default), the value is appended to
  // existing values. Otherwise it replaces any existing values.
  // This field is deprecated and please use
  // :ref:`append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>` as replacement.
  google.protobuf.BoolValue append = 2;
```

However, both the ext_authz & ext_proc use false as the default value of `append`.

Snippet of ext_proc:
```
    const bool append = PROTOBUF_GET_WRAPPED_OR_DEFAULT(sh, append, false);
```

Snippet of ext_authz:
```
          if (header.append().value()) {
            authz_response->response_headers_to_add.emplace_back(
                Http::LowerCaseString(header.header().key()), header.header().value());
          } else {
            authz_response->response_headers_to_set.emplace_back(
                Http::LowerCaseString(header.header().key()), header.header().value());
          }
```

wbpcode:
May be a bug?

mattklein123:
I vaguely recall encountering this at some point and wondering the same thing, but I can't remember the history. Unfortunately I don't think we can change the defaults in the code so we probably just need to document it?

wbpcode:
> I vaguely recall encountering this at some point and wondering the same thing, but I can't remember the history. Unfortunately I don't think we can change the defaults in the code so we probably just need to document it?

OK, may be we need some minor doc or proto comment updates.

wbpcode:
I almost forget this. I will create a minor pr for this quickly.

wbpcode:
Hmmm, I found a new problem.

We deprecated `append` of `HeaderValueOption` in the #22723 and use `append_action` (an enum) as alternative.
The default value of `append_action` enum is `APPEND_IF_EXISTS_OR_ADD`. It behaves same as not setting `append`.

However, ext_proc/ext_authz have error default value for `append` field. If we remove `append` finally, there is no way to distinguish between `not setting append_action` and `setting append_action with APPEND_IF_EXISTS_OR_ADD`. 

This may be break behaviors of ext_proc/ext_authz.

cc @mattklein123

wbpcode:
Or we need to deprecate current `HeaderValueOption` and create a new option that include a meaningless default enum value.

Then different extensions could apply different default behaviors.
😞

[...]

PR #23518, opened 2022-10-17
api: add attention to HeaderValueOption.append to avoid removing this field

wbpcode:

Commit Message: api: add attention to HeaderValueOption.append to avoid removing this field
Additional Description:

See #22845.

Risk Level: n/a.
Testing: n/a.
Docs Changes: added.
Release Notes: n/a.

### DIFF
--- a/api/envoy/config/core/v3/base.proto
+++ b/api/envoy/config/core/v3/base.proto
@@ -362,7 +362,7 @@ message HeaderValueOption {
   // .. attention::
   //   The :ref:`ext_authz filter <envoy_v3_api_msg_extensions.filters.http.ext_authz.v3.ExtAuthz>` and
   //   :ref:`ext_proc filter <envoy_v3_api_msg_extensions.filters.http.ext_proc.v3.ExternalProcessor>` have
-  //   error default value (false) for this field.
+  //   error default value (``false``) for this field.
   //   And the default value of
   //   :ref:`new append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>`
   //   is always APPEND_IF_EXISTS_OR_ADD.
--- a/api/envoy/config/core/v3/base.proto
+++ b/api/envoy/config/core/v3/base.proto
@@ -362,7 +362,7 @@ message HeaderValueOption {
   // .. attention::
   //   The :ref:`external authorization service <envoy_v3_api_msg_service.auth.v3.CheckResponse>` and
   //   :ref:`external processor service <envoy_v3_api_msg_service.ext_proc.v3.ProcessingResponse>` have
-  //   error default value (``false``) for this field.
+  //   default value (``false``) for this field.
   //   And the default value of field
   //   :ref:`append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>`
   //   is always ``APPEND_IF_EXISTS_OR_ADD``.
--- a/api/envoy/config/core/v3/base.proto
+++ b/api/envoy/config/core/v3/base.proto
@@ -360,15 +360,14 @@ message HeaderValueOption {
   // :ref:`append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>` as replacement.
   //
   // .. attention::
-  //   The :ref:`ext_authz filter <envoy_v3_api_msg_extensions.filters.http.ext_authz.v3.ExtAuthz>` and
-  //   :ref:`ext_proc filter <envoy_v3_api_msg_extensions.filters.http.ext_proc.v3.ExternalProcessor>` have
+  //   The :ref:`external authorization service <envoy_v3_api_msg_service.auth.v3.Authorization>` and
+  //   :ref:`external processor service <envoy_v3_api_msg_service.ext_proc.v3.ExternalProcessor>` have
   //   error default value (``false``) for this field.
-  //   And the default value of
-  //   :ref:`new append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>`
+  //   And the default value of field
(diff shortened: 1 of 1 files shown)


api/envoy/config/core/v3/base.proto
```diff
@@ -358,6 +358,17 @@ message HeaderValueOption {
...
+  // .. attention::
+  //   The :ref:`ext_authz filter <envoy_v3_api_msg_extensions.filters.http.ext_authz.v3.ExtAuthz>` and
+  //   :ref:`ext_proc filter <envoy_v3_api_msg_extensions.filters.http.ext_proc.v3.ExternalProcessor>` have
+  //   error default value (false) for this field.
+  //   And the default value of
+  //   :ref:`new append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>`
+  //   is always APPEND_IF_EXISTS_OR_ADD.
+  //   It means if we remove this field, the default behavior in these filter will be changed and may break our
+  //   users.
+  //   So we should keep this field and couldn't remove it until the above issue is resolved one day.
```
soulxu, line 371: Emm...The last two sentences seems like a note for the developer, it shouldn't be part of the API document.

api/envoy/config/core/v3/base.proto
```diff
@@ -358,6 +358,17 @@ message HeaderValueOption {
   // existing values. Otherwise it replaces any existing values.
   // This field is deprecated and please use
   // :ref:`append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>` as replacement.
+  //
+  // .. attention::
+  //   The :ref:`ext_authz filter <envoy_v3_api_msg_extensions.filters.http.ext_authz.v3.ExtAuthz>` and
+  //   :ref:`ext_proc filter <envoy_v3_api_msg_extensions.filters.http.ext_proc.v3.ExternalProcessor>` have
+  //   error default value (false) for this field.
```
phlax, line 365:
```suggestion
  //   error default value (``false``) for this field.
```

api/envoy/config/core/v3/base.proto
```diff
@@ -358,6 +358,17 @@ message HeaderValueOption {
...
+  //   And the default value of
+  //   :ref:`new append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>`
+  //   is always APPEND_IF_EXISTS_OR_ADD.
```
phlax, line 368:
```suggestion
  //   is always ``APPEND_IF_EXISTS_OR_ADD``.
```

phlax (commented):
couple of format nits inline...

api/envoy/config/core/v3/base.proto
mattklein123: As I mentioned in Slack, at this point the APIs are forever. We are never removing anything, so I think we can remove this. Can we rewrite this as end user documentation? Thank you.

/wait

wbpcode: It's easy to do some comments update. But I think there still is some info gap. 🤣 I re-read the API version policy. There are some conclusions by my understanding. Feel free to point int out if I am wrong.

1. When we say `the APIs are forever`, it means that this field is forever in the `core.v3.HeaderValueOption`.
2. One day the v4 APIs is enabled, then this field will be removed from the `core.v4.HeaderValueOption`. But the break changes are acceptable for major API versions update.

api/envoy/config/core/v3/base.proto
```diff
@@ -358,6 +358,16 @@ message HeaderValueOption {
...
+  //
+  // .. attention::
+  //   The :ref:`external authorization service <envoy_v3_api_msg_service.auth.v3.CheckResponse>` and
+  //   :ref:`external processor service <envoy_v3_api_msg_service.ext_proc.v3.ProcessingResponse>` have
+  //   error default value (``false``) for this field.
+  //   And the default value of field
+  //   :ref:`append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>`
+  //   is always ``APPEND_IF_EXISTS_OR_ADD``.
+  //   It means that the default behavior of ``HeaderValueOption`` in these two services will be changed after
+  //   this field is removed in the future version of API.
```
mattklein123, line 370: Per other comments, it's not going to get removed, so let's remove this. It's still not clear to me what we are telling the end user. Are we just telling them that the default is false when used in these 2 services? If so can we just say that clearly?

api/envoy/config/core/v3/base.proto
```diff
@@ -358,6 +358,16 @@ message HeaderValueOption {
   // existing values. Otherwise it replaces any existing values.
   // This field is deprecated and please use
   // :ref:`append_action <envoy_v3_api_field_config.core.v3.HeaderValueOption.append_action>` as replacement.
+  //
+  // .. attention::
+  //   The :ref:`external authorization service <envoy_v3_api_msg_service.auth.v3.CheckResponse>` and
+  //   :ref:`external processor service <envoy_v3_api_msg_service.ext_proc.v3.ProcessingResponse>` have
+  //   error default value (``false``) for this field.
```
mattklein123, line 365:
```suggestion
  //   default value (``false``) for this field.
```

mattklein123 (changes requested):
/wait

api/envoy/config/core/v3/base.proto
mattklein123: I haven't gotten around to doing an official policy update, but at this point realistically a) we are never doing v3 and b) we are never actually deleting any fields/code from Envoy. It's too disruptive. cc @envoyproxy/api-shepherds

wbpcode: got it.

mattklein123 (approved):
Thanks!

PR #23521, opened 2022-10-17
dfp: adding support for coalesced connections

alyssawilk:
Unhiding a config option.  I added tests and we queue for DFP so it's working as intended.

Testing: preexisting
Docs Changes: inline
Release Notes: inline


changelogs/current.yaml
```diff
@@ -14,4 +14,8 @@ removed_config_or_runtime:
 
 new_features:
 
+- area: http
+  change: |
+    made :ref:`allow_coalesced_connections <envoy_v3_api_field_extensions.clusters.dynamic_forward_proxy.v3.ClusterConfig.allow_coalesced_connections>` visible, enabling connection coalescing for HTTP/2 and HTTP/3 dynamic forward proxy connections.
```
mattklein123, line 19: nit: This is the first time the user has seen this, so can you write it from the perspective of it being a new feature?

mattklein123 (changes requested):
LGTM with small release not ask, thank you.

/wait

mattklein123 (approved):
Cool, thanks.

PR #23529, opened 2022-10-17
Stricter boundary checks for Duration when it is converted to std::chrono::milliseconds

yanavlasov:
Commit Message:
Internally protobuf first converts seconds to nanoseconds before converting to milliseconds. This can overflow if the default bound checks are applied to the `seconds` value.

This can potentially be a breaking change for configuration that erroneously supplied values higher than 9223372036, since Envoy will now be rejecting such config. Before it was accepting it and using the "wrapped around" value.

This case is also verified by the test/server/server_corpus/clusterfuzz-testcase-server_fuzz_test-5988544525893632 corpus, however the asan sanitizer needs to be of a newer version to catch the error.

Risk Level: Low. Small possibility of a breaking change.
Testing: Unit Test 
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/protobuf/utility.cc
+++ b/source/common/protobuf/utility.cc
@@ -919,22 +919,34 @@ ProtobufWkt::Value ValueUtil::listValue(const std::vector<ProtobufWkt::Value>& v
 
 namespace {
 
-void validateDuration(const ProtobufWkt::Duration& duration) {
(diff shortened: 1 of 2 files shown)


wbpcode:
LGTM. Still defer this to @envoyproxy/api-shepherds for a quick check.

lizan (approved):

PR #23541, opened 2022-10-18
header-parser: fix config regex infinite loop

adisuissa:
Commit Message: header-parser: fix config regex infinite loop
Additional Description:
#23447 changed the regex engine to RE2, but introduced a way to set a config that cannot be ingested (results in an infinite loop), because the use `RE2::Replace` would replaces occurrences of `\0` in the original string with the original matched string. This PR changes to a `RE2::Match` -> `absl::StrReplaceAll` pattern, which doesn't replace the `\digit` occurrences.

Risk Level: Low.
Testing: Added a unit test.
Docs Changes: N/A.
Release Notes: N/A.
Fixes fuzz issues [52448](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52448) and [52449](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52449).


source/common/router/header_parser_utils.cc
```diff
@@ -48,11 +48,18 @@ std::string HeaderParser::translateMetadataFormat(const std::string& header_valu
           "Header formatter: JSON format of {} parameters has been obsoleted. Use colon format: {}",
           metadata_type.as_string() + "_METADATA", new_format.c_str());
 
-      re2::RE2::Replace(&new_header_value, re, new_format);
+      if (!re2::RE2::Replace(&new_header_value, re, new_format)) {
```
cpakulski, line 51: I looked at the corpus files and still do not fully grasp how it is that regex finds the pattern but fails in replacing it.

source/common/router/header_parser.h
```diff
@@ -94,7 +96,7 @@ class HeaderParser : public Http::HeaderEvaluator {
                                              bool do_formatting = true) const;
 
   static std::string translateMetadataFormat(const std::string& header_value);
```
cpakulski, line 98: Why translateMetadataFormat does not return `absl::StatusOr<T>`?

source/common/router/header_parser_utils.cc
adisuissa: I've dug into this a bit deeper. In `RE2::Replace` backslash-escaped digits are used to refer to specific matches.
If the replaced string had that, it could keep the original (wrong) pattern after the replacement, causing an infinite loop.
I've updated the code to use `RE2::Match` and `absl::StrReplaceAll` instead.

source/common/router/header_parser.h
adisuissa: I actually changed it to return StatusOr, but then saw that your original implementation intentionally returns the original header, and so I tried to keep it that way.
Alas, the use of `Match` and `StrReplaceAll` doesn't need `StatusOr` for both functions.

wrowe:
CI appears borked, hopefully a resolution shortly.

wrowe:
And excuse my confusion, apparently first-pass-reviewer is initially sufficient, it was the waiting on CI success, not assignment.

wrowe:
Appears to be known test failures, you may have to merge main as well...

/retest

source/common/router/header_parser_utils.cc
```diff
@@ -25,34 +26,34 @@ std::string HeaderParser::translateMetadataFormat(const std::string& header_valu
   const re2::RE2& re = getMetadataTranslatorPattern();
   ASSERT(re.ok());
   std::string new_header_value = header_value;
-  re2::StringPiece json_array, metadata_type;
-  while (re.PartialMatch(new_header_value, re, &metadata_type, &json_array)) {
-    std::string new_format;
+  re2::StringPiece matches[4];
```
cpakulski, line 29: I think 3 is enough. There are 2 capturing groups plus one for entire string.

cpakulski (approved):
Just a small nit related to size of matching groups. Thanks for catching this issue and fixing it.

source/common/router/header_parser_utils.cc
cpakulski: Thanks for explaining it.

source/common/router/header_parser.h
cpakulski: Great!. Thanks.

source/common/router/header_parser_utils.cc
adisuissa: You are correct, thanks!

KBaichoo:
friendly ping @snowp, I think this is waiting on you

snowp (approved):
Makes sense to me - can you update the PR description now that since this is no longer using Replace?

adisuissa:
> Makes sense to me - can you update the PR description now that since this is no longer using Replace?

Good idea, done.

Issue #2181, opened 2017-12-09
build does not work from tarball, but only from git checkout


cs224:
*Title*: build does not work from tarball, but only from git checkout

*Description*:

When downloading a release tarball, unpacking and trying to build you get the following error:

    ERROR: Process exited with status 1: Process exited with status 1
    fatal: Not a git repository (or any parent up to mount point /source)
    Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
    Target //source/exe:envoy-static.stamped failed to build
    INFO: Elapsed time: 57.978s, Critical Path: 0.03s
    FAILED: Build did NOT complete successfully

It would be nice if the build worked also on the downloadable tarballs.

Thanks,
Christian

lavagetto:
The build process depends on the checkout being a git repository indeed - stumbled upon the same problem myself.

mattklein123:
The main need is here: https://github.com/envoyproxy/envoy/blob/master/bazel/get_workspace_status

It should be possible to alter that script to work without git and feed in the versions, etc. from a file (or something). I'm not sure what other git needs bazel has, if any.

merge03:
Is there a work around for this issue?

dio:
@merge03 there are two workarounds that I can think of:

1. Create a `SOURCE_VERSION` file and fill it with the commit hash of the release package. e.g. for v1.6.0 release, the commit hash value will be `4dd49d8809f7aaa580538b3c228dd99a2fae92a4`.
2. Replace the `bazel/get_workspace_status` with the one in here: https://github.com/dio/envoy/blob/workaround-2181/bazel/get_workspace_status (it is "not clean" indeed, but at least it gets the SHA from Github API).

merge03:
@dio Option #1 worked just fine.  Just noting that the file needed to be in the ./ci directory.  Didn't try #2.

Thanks for the help!

exell-christopher:
Is there any plan to include generation of the SOURCE_VERSION file as part of the tarball generation? If someone can point me to the tooling for that I'd be happy to add it.

yurivict:
Could you just remove git command invocation in the build?
git is a tool to manage sources and for collaboration, it is not to be called from inside of the build.

All version information should be provided from outside through command arguments.

mattklein123:
Fixed by https://github.com/envoyproxy/envoy/pull/23577

PR #23577, opened 2022-10-20
tools, github: Add current SOURCE_VERSION writer

dio:
Commit Message: This adds a script to produce the `SOURCE_VERSION` file which is useful to build envoy in a non-git (from an extracted release tarball) directory. See: `bazel/get_workspace_status` for more information.

Relevant: https://github.com/envoyproxy/envoy/issues/2181.

Additional Description: This is useful to have a way to produce `SOURCE_VERSION`, for example, to solve: https://github.com/Homebrew/homebrew-core/blob/923bb7a622b86876a662d1ed4a7ac51ccb2befcf/Formula/envoy.rb#L4. So downstream builder can consume release tarballs to build envoy without adding custom scripts to create `SOURCE_VERSION`.
Risk Level: Low, since this is an optional tool.
Testing: Run the script in git and non-git environments.
Docs Changes: Added.
Release Notes: N/A

### DIFF
--- a/tools/github/write_current_source_version.py
+++ b/tools/github/write_current_source_version.py
@@ -8,42 +8,42 @@
 # Note: This script can only be executed from project root directory of an extracted "release"
 # tarball.
 
-import sys
 import json
+import pathlib
+import sys
+import urllib.request
 
-from urllib import request
-from pathlib import Path
+if __name__ == '__main__':
+    # Simple check if a .git directory exists. When we are in a Git repo, we should rely on git.
+    if pathlib.Path(".git").exists():
+        print(
+            "Failed to create SOURCE_VERSION. "
+            "Run this script from an extracted release tarball directory.")
+        sys.exit(1)
 
-# Simple check if a .git directory exists. When we are in a Git repo, we should rely on git.
-if Path(".git").exists():
-    print(
-        "Failed to create SOURCE_VERSION. "
-        "Run this script from an extracted release tarball directory.")
-    sys.exit(1)
+    # Check if we have VERSION.txt available
+    current_version_file = pathlib.Path("VERSION.txt")
+    if not current_version_file.exists():
+        print(
+            "Failed to read VERSION.txt. "
+            "Run this script from project root of an extracted release tarball directory.")
+        sys.exit(1)
 
-# Check if we have VERSION.txt available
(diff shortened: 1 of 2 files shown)


phlax (commented):
looks good @dio 

im wondering if we should wrap the code in a `main` function and use a `if __name__ == "__main__":` block to trigger it - it just means it wont run just from being imported

im also wondering if we need to add some info somewhere to say to run this before running any bazel tasks (for a tarball build)

tools/github/write_current_source_version.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# The SOURCE_VERSION file is produced by reading current version tag from VERSION.txt file then
+# fetch the corresponding commit hash from GitHub.
+#
+# Note: This script can only be executed from project root directory of an extracted "release"
+# tarball.
+
+import sys
+import json
+
```
phlax, line 13: nit: all of these imports are stdlib so i think no need for a space

also can we use alpha ordering for imports

dio:
Thanks for the review @phlax. I added some lines in `bazel/README.md`. Please let me know if that makes sense (presented as a subsection, under the **Production environments** section, since probably that is what people want to do with release tarballs).

tools/github/write_current_source_version.py
dio: Resolved.

phlax (commented):
one python suggestion and then i think this should be good to land

tools/github/write_current_source_version.py
```diff
@@ -0,0 +1,49 @@
...
+        sys.exit(1)
+
+    # Fetch the current version commit information from GitHub.
+    with urllib.request.urlopen("https://api.github.com/repos/envoyproxy/envoy/commits/v"
+                                + current_version) as response:
+        commit_info = json.loads(response.read())
+        source_version_file = pathlib.Path("SOURCE_VERSION")
+        with source_version_file.open("w", encoding="utf-8") as source_version:
+            # Write the extracted current version commit hash "sha" to SOURCE_VERSION.
+            source_version.write(commit_info["sha"])
```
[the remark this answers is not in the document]
dio: Ah thanks, done.

phlax (approved):
lgtm, thanks @dio

Issue #23557, opened 2022-10-19
Using `regex_rewrite` in RedirectAction appends original path to the resulting path


Ackuq:
*Title*: *Using `regex_rewrite` in RedirectAction append original path to the end*

*Description*:

When using a `regex_rewrite` configuration for a redirect route, the original path always gets appended to the end of the resulting path. This happens even though the regex pattern does not find a match in the requested path. In my included configuration, I use one of the examples for this field, that is presented in the [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-redirectaction). According to the docs this should regex should rewrite the path `/service/foo/v1/api` to `/v1/api/instance/foo`. However, the resulting path becomes `/v1/api/instance/fooservice/foo/v1/api`. As shown here:

```bash
curl -H "Host: example.com" localhost/service/foo/v1/api -v
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /service/foo/v1/api HTTP/1.1
> Host: example.com
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< location: http://example.com/v1/api/instance/fooservice/foo/v1/api
< date: Wed, 19 Oct 2022 08:43:27 GMT
< server: envoy
< content-length: 0
< 
* Connection #0 to host localhost left intact
```

This issue also occurs when the regex is not matched, e.g.:
```bash
curl -H "Host: example.com" localhost/some-path -v
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /some-path HTTP/1.1
> Host: example.com
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< location: http://example.com/some-pathsome-path
< date: Wed, 19 Oct 2022 08:46:52 GMT
< server: envoy
< content-length: 0
< 
* Connection #0 to host localhost left intact
```

*Repro steps*:
Envoy version: v1.23.1
1. Define a redirect route for a virtual host in your HTTP listener
2. Add a `regex_rewrite` configuration for your redirect action
3. Make a request that gets routed to that route

*Config*:
```
admin:
  address:
    socket_address: { address: 0.0.0.0, port_value: 9901 }

static_resources:
[...]

PR #23587, opened 2022-10-20
router: fix `regex_rewrite` for redirects matched on prefix

Ackuq:
Commit Message: router: fix `regex_rewrite` for redirects matched on prefix
Additional Description:
These changes fixes an issue when using `regex_rewrite` as a part of a redirect when the path that is matched does not equal the full path. The issue was that only the matched section of the original path (e.g. prefix) was replaced with the substitution of the regex rewrite, while the expected behaviour is that the full path should be replaced.

This solves this by adding a case for these scenarios, i.e. if we have a redirect and use `regex_rewrite`, replace the full path with the rewritten path.

Risk Level: medium (changes the behaviour of how `regex_rewrite` and redirects works together)
Testing: Implemented unit tests to ensure the expected behaviour.
Docs Changes: N/A
Release Notes: Added
Fixes #Issue: #23557

Signed-off-by: Axel Pettersson <axel@pettersson.cc>

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -11,6 +11,9 @@ bug_fixes:
 - area: grpc_json_transcoder
   change: |
     fix a bug when using http2, request body has google.api.HttpBody and the size is < 16KB, it will cause EOF from the backend grpc server.
+- area: router
+  change: |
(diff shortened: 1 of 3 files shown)


Ackuq:
@wbpcode Fixed!

mattklein123 (approved):
Thanks!

Issue #23593, opened 2022-10-20
envoy_reloadable_features_deprecate_global_ints deprecation


alyssawilk:
Your change #20168 (runtime: deprecating global ints) introduced a runtime guarded feature. It has been 6 months since the new code has been exercised by default, so it's time to remove the old code path. This issue tracks source code cleanup so we don't forget.

PR #12656, opened 2020-08-14
[http] Headermap Lazy map addition

adisuissa:
Commit Message: Adding a lazy map to the current header-map implementation.

Additional Description: This is a continuation of #9261. Adding a map to headermap's HeaderList to access headers by key in O(1) instead of O(n).
In addition some benchmarks that emulate encoding and decoding of headers, and route matching by headers are added.

Risk Level: High - changes to headermap.
Testing: Added `RemoveIf` test, and some perf tests.
Docs Changes: Once the design is settled, I'll update the [headermap doc](https://github.com/envoyproxy/envoy/blob/master/source/docs/header_map.md)
Release Notes: None.

Signed-off-by: Adi Suissa-Peleg <adip@google.com>

PR #23613, opened 2022-10-20
deprecating envoy.reloadable_features.deprecate_global_int

alyssawilk:
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: inline
Fixes https://github.com/envoyproxy/envoy/issues/23593


source/common/http/header_map_impl.cc
```diff
@@ -22,6 +22,7 @@ namespace {
 
 constexpr absl::string_view DelimiterForInlineHeaders{","};
 constexpr absl::string_view DelimiterForInlineCookies{"; "};
+const static int kMinHeadersForLazyMap = 3; // Arbitrary hard-coded value.
```
jmarantz, line 25: I think it's better than arbitrary :) @adisuissa did ran microbenchmarks to at least sanity-check that choice.

source/common/http/header_map_impl.h
```diff
@@ -181,10 +181,9 @@ class HeaderMapImpl : NonCopyable {
   /**
    * List of HeaderEntryImpl that keeps the pseudo headers (key starting with ':') in the front
    * of the list (as required by nghttp2) and otherwise maintains insertion order.
-   * When the list size is greater or equal to the envoy.http.headermap.lazy_map_min_size runtime
-   * feature value (defaults to 3, if not set), all headers are added to a map, to allow
-   * fast access given a header key. Once the map is initialized, it will be used even if the number
-   * of headers decreases below the threshold.
+   * When the list size is greater or equal to the default min size (3), all headers are added to a
```
jmarantz, line 184: s/default// ?

jmarantz (commented):
basically lgtm; makes sense to me to drop that settings as no one ever really changed it to my knowledge :)

Is there some general reason not to support int settings though? It seems this PR is part of a bigger mission I was not aware of.

source/common/http/header_map_impl.cc
adisuissa: 3 was chosen after doing some benchmarks (#12656).

jmarantz (approved):
thanks!

PR #23630, opened 2022-10-22
access log: added support to flush tcp access log periodically

### LINKED ISSUE #23579: Enhance ALS TCP to support periodic reports when the TCP connection is not closed
*Title*: *Enhance ALS TCP to support periodic reports when the TCP connection is not closed.*

*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Currently, the HTTP ALS reports for every HTTP request, and TCP ALS only reports after the connection is closed. 
In the observability case of SkyWalking for Istio mesh, we rely on the ALS to generate the topology map with metrics. But when TCP is set through service mesh, envoy TCP ALS would not report TCP ALS until the connection is closed.
Meanwhile, a lot of TCP connections are running in keep-alive mode, so, it lasts days to not close. As a result, we can't observe the relationship and metrics.
We hope this could be enhanced to add a new option to ALS to support TCP ALS reports periodically. And we should consider a new field in the ALS indicating the reporting mode, a) `end-of-the-session`(same as now), b) `intermediate`/`periodical`(report every several seconds even if the connection is still kept)

cc @cpakulski @dio 
FYI @wbpcode As you are well known SkyWalking using ALS :) 

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

- SkyWalking ALS analysis doc, https://skywalking.apache.org/docs/main/next/en/setup/envoy/als_setting/

wbpcode:

Commit Message: access log: added support to flush tcp access log periodically
Additional Description:

To close #23579

Risk Level: mid.
Testing: waiting.
Docs Changes: n/a.
Release Notes: n/a


wbpcode:
I think this is what you want. And it should work and is compatible with exist implementation. cc @wu-sheng

wbpcode:
cc @dio A draft is completed. 😸 I test it locally and it works. But may more detailed unit tests are necessary.

wu-sheng:
Yes, I think this would fix the gap. SkyWalking could run topology analysis with metrics using rate, as metrics are counter from the connection established.

api/envoy/data/accesslog/v3/accesslog.proto
```diff
@@ -40,6 +40,10 @@ message TCPAccessLogEntry {
 
   // Properties of the TCP connection.
   ConnectionProperties connection_properties = 2;
+
+  // If this log entry is final log entry that flushed after the connection closed or
+  // intermediate log entry that flushed periodically during the connection.
+  bool intermediate_log_entry = 3;
```
lizan, line 46: When this is true, how do an ALS determine whether two TCPAccessLogEntry is from the same connection or not?

wbpcode: then, we need a unique id for this. A uuid would be better than the connection id.

wbpcode:
```
  // All metadata encountered during request processing, including endpoint
  // selection.
  //
  // This can be used to associate IDs attached to the various configurations
  // used to process this request with the access log entry. For example, a
  // route created from a higher level forwarding rule with some ID can place
  // that ID in this field and cross reference later. It can also be used to
  // determine if a canary endpoint was used or not.
  config.core.v3.Metadata metadata = 17;
```

How about inject a uuid to the metadata of TCPAccessLogEntry. Seems like this field is designed for this scene.
If we need determine whether multiple intermediate log entry if from same connection or not, the the related id should be provided in the metadata.

wu-sheng: Good suggestion, I forgot a connection ID(any ID to represent the TCP connection) is crucial .

source/common/tcp_proxy/tcp_proxy.cc
```diff
@@ -754,6 +770,27 @@ void Filter::onMaxDownstreamConnectionDuration() {
   read_callbacks_->connection().close(Network::ConnectionCloseType::NoFlush);
 }
 
+void Filter::onAccessLogFlushInterval() {
+  for (const auto& access_log : config_->accessLogs()) {
+    access_log->log(nullptr, nullptr, nullptr, getStreamInfo());
```
cpakulski, line 775: Is there any way to distinguish intermediate log from the final one when the session is closed? I see in the comments above to use uuid for correlation, so multiple intermediate updates are linked and do not look as separate sessions, but there should be a final log which must inform that the session is completed. Otherwise it will look as another intermediate update and mess the stats.

wbpcode: Yeah, of course. The intermediate log entries have specified flag `intermediate_log_entry = true`.

cpakulski: Ah, yes - you are right. I see it for grpc sink. But I am guessing that entries flushed to a file/stdout will be undistinguished, correct?

wbpcode: > Ah, yes - you are right. I see it for grpc sink. But I am guessing that entries flushed to a file/stdout will be undistinguished, correct?

Yes, for now. It would be easy to add a new command to support distinguishing in the file/stdout in the future.

toffentoffen: I think we should not base or infer a concrete state or behaviour based on the presence or not of another field(Duration in this case) as it could potentially change in the feature and the connection/dependency created to infer the state(intermediate or final) can break.

Also, I think that having %DURATION% on intermediate `access_log` is quite valuable for the consumer of the entry.
For instance, if the consumer would like to know how many bytes per minute have been send, knowing the duration is crucial and make processing way easier, otherwise it would need to know when the last intermediate entry was send to calculate the duration(and it will not be exact).

toffentoffen: A follow up question based on my previous message:

On each intermediate message send, what will the values in the different properties be? Concretely metric style properties like for instance byes received/send found at `connection_properties { received_bytes: 6 sent_bytes: 33 }` or perhaps even duration?
Would they be an ever increasing value, or will they be reset and only send the observed values during the `access_log_flush_interval`?

source/common/tcp_proxy/tcp_proxy.cc
```diff
@@ -623,6 +641,17 @@ Network::FilterStatus Filter::onNewConnection() {
     connection_duration_timer_->enableTimer(config_->maxDownstreamConnectionDuration().value());
   }
 
+  if (config_->accessLogFlushInterval().has_value()) {
+    access_log_flush_timer_ = read_callbacks_->connection().dispatcher().createTimer(
```
cpakulski, line 645: Just a quick question. Will the timer and `tcp_proxy` filter run on the same thread? I am thinking about concurrency when accessing `stream_info`. Timer's callback will access `stream_info` to create access log and at the same time `tcp_proxy` may access `stream_info` to update it. Maybe dispatcher assures proper access.

wbpcode: yeah.

source/common/tcp_proxy/tcp_proxy.cc
wbpcode: The `received/send` would be ever increasing values from the connection was established.

wbpcode:
Hi, @lizan, could you check the API again when you have free time? Thanks.

api/envoy/data/accesslog/v3/accesslog.proto
lizan: Sounds, good, worth a comment in the documentation.

api/envoy/data/accesslog/v3/accesslog.proto
```diff
@@ -40,6 +40,15 @@ message TCPAccessLogEntry {
...
   ConnectionProperties connection_properties = 2;
+
+  // If this log entry is final log entry that flushed after the connection closed or
+  // intermediate log entry that flushed periodically during the connection.
+  // There may be multiple intermediate log entries and only one final log entry for each
+  // TCP connection.
+  // And if it is necessary, some IDs or identifiers can be added to the log entry
+  // :ref:`metadata <envoy_v3_api_field_data.accesslog.v3.AccessLogCommon.metadata>` to
+  // correlate all these intermediate log entries and final log entry.
+  bool intermediate_log_entry = 3;
```
lizan, line 51: Another thought, this might be in `common_properties`, IIRC there are similar request for long-live H2 stream (gRPC stream).

wbpcode: ha, yeah, I will move it to the common_properties.

phlax:
@lizan could you give this another look

/retest

lizan:
/wait

This generally in the right track, will address #23838 first and then use that for stream id.

wbpcode:
friendly ping @lizan

wbpcode:
Added the substitution formatter for the stream id in this PR to avoid creating a separated PR for it.

api/envoy/data/accesslog/v3/accesslog.proto
```diff
@@ -202,6 +202,19 @@ message AccessLogCommon {
 
   // Connection termination details may provide additional information about why the connection was terminated by Envoy for L4 reasons.
   string connection_termination_details = 25;
+
+  // Optional unique id of stream (TCP connection, long-live HTTP2 stream, HTTP request) for logging and tracing.
+  // This could be any format string that could be used to identify one stream.
+  string stream_unique_id = 26;
```
lizan, line 208: prefer simply `stream_id`. same as substitute formatter.

wbpcode: Done. Thanks.

wbpcode:
friendly ping cc @lizan

Issue #23632, opened 2022-10-22
test flaky: tls/ocsp:ocsp_test and tls:ssl_socket_test failed out of the expectation


wbpcode:
see the ci pipelines. and here is an example https://dev.azure.com/cncf/envoy/_build/results?buildId=118940&view=logs&jobId=8c169225-0ae8-53bd-947f-07cb81846cb5&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec

wbpcode:
Hi, any idea about it? cc @ggreenway

PR #23635, opened 2022-10-23
ocsp: rotate certs of test data to fix flaky tests

wbpcode:
Commit Message: ocsp: rotate certs of test data to fix flaky tests
Additional Description:

See #23632.

Risk Level: none. test only update.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
cc @ggreenway cc @lizan 

Almost all PRs was blocked by the flaky test and this PR should fixed it.

wbpcode:
cc @alyssawilk Here fixed the flaky ci.

alyssawilk (approved):
Thanks for the fix!

phlax:
i wonder if we can give these certs a longer ttl so we dont need to update them so frequently

wbpcode:
> i wonder if we can give these certs a longer ttl so we dont need to update them so frequently

I agree. Now it's two years, I think. For test, I think a longer ttl is completely ok.

wbpcode:
We can update these certs refresh script with a pr to achieve this target. Or even generate certs dynamically at runtime?

phlax:
either way seems ok - not sure how long they take to generate - but probably they can just be generated dynamically

ggreenway:
I'd prefer to have the certs generated by bazel and not checked into the repo; this would fix the issue that old versions of envoy will always fail tests.

PR #23637, opened 2022-10-23
tools, github: Silently exit when writing sha in git

dio:
Commit Message: This patch allows to silently exit when writing the current version SHA when in a git repository. User can supply `--skip_error_in_git`.
Additional Description: This is to support running this script when in a git repository for automation, for example in homebrew's `brew --HEAD` scenario. This script can be still executed (either in a git repo or an extracted tarball) without sophisticated checks and carry-on using git to do the workspace status check. 
Risk Level: Low
Testing: Manual
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/tools/github/write_current_source_version.py
+++ b/tools/github/write_current_source_version.py
@@ -8,41 +8,55 @@
 # Note: This script can only be executed from project root directory of an extracted "release"
 # tarball.
 
+import argparse
 import json
 import pathlib
 import sys
 import urllib.request
 
 if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Write current source version')
+    parser.add_argument(
+        '--skip_error_in_git',
+        dest='skip_error_in_git',
+        help='Skip returning error on exit when the current directory is a git repository.',
+        action='store_true')
+    args = parser.parse_args()
+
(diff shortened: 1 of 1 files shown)


tools/github/write_current_source_version.py
```diff
@@ -8,41 +8,55 @@
...
+    parser.add_argument(
+        '--skip_error_in_git',
+        dest='skip_error_in_git',
+        help='Skip returning error on exit when the current directory is a git repository.',
+        action='store_true')
+    args = parser.parse_args()
+
     # Simple check if a .git directory exists. When we are in a Git repo, we should rely on git.
-    if pathlib.Path(".git").exists():
+    if pathlib.Path('.git').exists():
```
phlax, line 27: im wondering why change the quotation marks - we dont have a hard rule about this - but i think how you had it is more consistent

phlax (commented):
hi @dio - change seems fine - added a small nit about formatting

tools/github/write_current_source_version.py
dio: Goctha. Sorry. Seems like my formatter did it :(.

dio: @phlax I have updated the quotation marks.

phlax (approved):
lgtm, thanks @dio

Issue #22583, opened 2022-08-05
debug assertion failure: `assert failure: isIdleImpl()`


howardjohn:
*Title*: debug assertion failure: `assert failure: isIdleImpl()`

*Description*:
In our CI with debug builds, we are regularly seeing this assertion failure.

```
2022-08-04T22:11:20.554576Z	critical	envoy assert	assert failure: isIdleImpl().
2022-08-04T22:11:20.554714Z	critical	envoy backtrace	Caught Aborted, suspect faulting address 0x53900000018
2022-08-04T22:11:20.554729Z	critical	envoy backtrace	Backtrace (use tools/stack_decode.py to get line numbers):
2022-08-04T22:11:20.554739Z	critical	envoy backtrace	Envoy version: 01baf88a8942c273181c830b4307bfe70bd839c0/1.24.0-dev/Clean/DEBUG/BoringSSL
2022-08-04T22:11:20.646444Z	critical	envoy backtrace	#0: Envoy::SignalAction::sigHandler() [0x55787082aedc]
2022-08-04T22:11:20.646627Z	critical	envoy backtrace	#1: __restore_rt [0x7fe4307ab140]
2022-08-04T22:11:20.729793Z	critical	envoy backtrace	#2: Envoy::Tcp::ConnPoolImpl::deleteIsPending() [0x55786ee5e6e5]
2022-08-04T22:11:20.816228Z	critical	envoy backtrace	#3: Envoy::Event::DispatcherImpl::deferredDelete() [0x557870440645]
2022-08-04T22:11:20.901629Z	critical	envoy backtrace	#4: Envoy::Upstream::ClusterManagerImpl::ThreadLocalClusterManagerImpl::tcpConnPoolIsIdle() [0x55786edd54b1]
2022-08-04T22:11:20.994375Z	critical	envoy backtrace	#5: Envoy::Upstream::ClusterManagerImpl::ThreadLocalClusterManagerImpl::ClusterEntry::tcpConnPoolImpl()::$_24::operator()() [0x55786ee05891]
2022-08-04T22:11:21.111388Z	critical	envoy backtrace	#6: std::__1::__invoke<>() [0x55786ee0582d]
2022-08-04T22:11:21.222222Z	critical	envoy backtrace	#7: std::__1::__invoke_void_return_wrapper<>::__call<>() [0x55786ee057dd]
[...]

PR #23645, opened 2022-10-24
test: adding original dst tests.

alyssawilk:
Part of https://github.com/envoyproxy/envoy/issues/22583

### DIFF
--- a/test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
+++ b/test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
@@ -29,6 +29,7 @@ ConfigHelper::ConfigModifierFunction setOriginalDstCluster(int port) {
 
     auto* listener = bootstrap.mutable_static_resources()->mutable_listeners(0);
     listener->mutable_address()->mutable_socket_address()->set_address("0.0.0.0");
+    listener->set_traffic_direction(envoy::config::core::v3::INBOUND);
 
     auto* listener_filter = listener->add_listener_filters();
     listener_filter->set_name("envoy.filters.listener.original_dst");
--- a/test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
+++ b/test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
@@ -1,0 +1,132 @@
+#include "envoy/config/bootstrap/v3/bootstrap.pb.h"
+#include "envoy/event/dispatcher.h"
+#include "envoy/extensions/filters/listener/original_dst/v3/original_dst.pb.h"
+#include "envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.pb.h"
+#include "envoy/http/header_map.h"
+#include "source/common/http/headers.h"
+#include "source/common/network/utility.h"
+#include "test/common/upstream/utility.h"
+#include "test/integration/http_integration.h"
+#include "test/integration/http_protocol_integration.h"
+#include "test/integration/utility.h"
+
+#include "gtest/gtest.h"
+
+namespace Envoy {
+
+using OriginalDstIntegrationTest = HttpProtocolIntegrationTest;
+
+ConfigHelper::ConfigModifierFunction setOriginalDstCluster(int port) {
+  return [port](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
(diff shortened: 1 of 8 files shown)


test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
```diff
@@ -0,0 +1,134 @@
...
+
+TEST_P(OriginalDstTcpProxyIntegrationTest, TestManyConnections) {
+  if (version_ != Network::Address::IpVersion::v4) {
+    return;
+  }
+
+  // Create an upstream on 0.0.0.0
+  auto address = Network::Test::getAnyAddress(Network::Address::IpVersion::v4);
+  Network::DownstreamTransportSocketFactoryPtr factory =
+      upstream_tls_ ? createUpstreamTlsContext(upstreamConfig())
```
soulxu, line 97: ditto

test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
```diff
@@ -0,0 +1,134 @@
...
+    auto response =
+        sendRequestAndWaitForResponse(default_request_headers_, 0, default_response_headers_, 0);
+    ASSERT_TRUE(fake_upstream_connection_->close());
+    fake_upstream_connection_.reset();
+    codec_client_->close();
+  }
+}
+
+class OriginalDstTcpProxyIntegrationTest
+    : public testing::TestWithParam<Network::Address::IpVersion>,
```
soulxu, line 80: There is check in the test, it will return directly when the test isn't ipv4, then why we still add param for the test?

test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
```diff
@@ -0,0 +1,134 @@
...
+  config_helper_.addConfigModifier(
+      setOriginalDstCluster(fake_upstreams_[0]->localAddress()->ip()->port()));
+  initialize();
+
+  const int32_t kMaxConnections = 100;
+  for (int i = 0; i < kMaxConnections; ++i) {
+    // Set up the connection
+    IntegrationTcpClientPtr tcp_client = makeTcpConnection(
+        lookupPort("listener_0"), nullptr, nullptr, fmt::format("127.0.0.{}", i + 1));
+    FakeRawConnectionPtr fake_upstream_connection;
```
soulxu, line 111: Should we validate the fake upstream connection's local address, also the overrides port?

test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
```diff
@@ -0,0 +1,134 @@
...
+      upstream_tls_ ? createUpstreamTlsContext(upstreamConfig())
+                    : Network::Test::createRawBufferDownstreamSocketFactory();
+  fake_upstreams_.emplace_back(std::make_unique<FakeUpstream>(std::move(factory), address,
+                                                              configWithType(upstreamProtocol())));
+
+  config_helper_.addConfigModifier(
+      setOriginalDstCluster(fake_upstreams_[0]->localAddress()->ip()->port()));
+  initialize();
+
+  const int32_t kMaxConnections = 100;
```
soulxu, line 106: just a question ,why we need to test many connections?

test/extensions/filters/listener/original_dst/original_dst_integration_test.cc
```diff
@@ -0,0 +1,144 @@
...
+#include "test/integration/http_protocol_integration.h"
+#include "test/integration/utility.h"
+
+#include "gtest/gtest.h"
+
+namespace Envoy {
+
+using OriginalDstIntegrationTest = HttpProtocolIntegrationTest;
+
+ConfigHelper::ConfigModifierFunction setOriginalDstCluster(int port) {
```
kyessenov, line 21: Just a high-level feedback: I think the tricky part about OriginalDST is the host management. That means creation of host on-demand, deletion of the host after idle/cleanup, and re-creation of the host. With the port override, it's just one host, so we need to make sure pools remove them in-between requests/connections. For example, for TCP, I think we have to wait for "cleanup" interval .

kyessenov (approved):
Thanks for adding the test! I have just a couple of minor comments. We can build off this to test "cleanup" and other things.

PR #23648, opened 2022-10-25
[WIP] Deferred logging of response time (QUIC only)

pksohn:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


envoy/http/codec.h
```diff
@@ -227,6 +194,56 @@ class RequestDecoder : public virtual StreamDecoder {
...
+   * @return List of shared pointers to access loggers for this stream.
+   */
+  virtual std::list<AccessLog::InstanceSharedPtr> accessLogHandlers() PURE;
+};
+
+/**
+ * Stream encoder used for sending a response (server to client). Virtual inheritance is required
+ * due to a parallel implementation split between the shared base class and the derived class.
+ */
+class ResponseEncoder : public virtual StreamEncoder {
```
alyssawilk, line 213: I'm probably missing something but why move this?

source/common/http/conn_manager_impl.cc
```diff
@@ -278,7 +278,14 @@ void ConnectionManagerImpl::doDeferredStreamDestroy(ActiveStream& stream) {
 
   stream.completeRequest();
   stream.filter_manager_.onStreamComplete();
-  stream.filter_manager_.log();
+  if (codec_ && codec_->protocol() == Protocol::Http3 &&
+      !stream.filter_manager_.sawDownstreamReset() &&
+      Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.quic_defer_logging_to_ack_listener")) {
```
alyssawilk, line 284: this should be called out in release notes.

source/common/http/conn_manager_impl.cc
```diff
@@ -278,7 +278,14 @@ void ConnectionManagerImpl::doDeferredStreamDestroy(ActiveStream& stream) {
 
   stream.completeRequest();
   stream.filter_manager_.onStreamComplete();
-  stream.filter_manager_.log();
+  if (codec_ && codec_->protocol() == Protocol::Http3 &&
```
alyssawilk, line 281: let's comment what's going on here.

source/common/http/conn_manager_impl.cc
```diff
@@ -1783,6 +1790,7 @@ void ConnectionManagerImpl::ActiveStream::recreateStream(
   connection_manager_.doEndStream(*this, /*check_for_deferred_close*/ false);
 
   RequestDecoder& new_stream = connection_manager_.newStream(*response_encoder, true);
+  response_encoder->setRequestDecoder(new_stream);
```
alyssawilk, line 1793: where was this before?

envoy/http/codec.h
```diff
@@ -227,6 +194,56 @@ class RequestDecoder : public virtual StreamDecoder {
    * @return StreamInfo::StreamInfo& the stream_info for this stream.
    */
   virtual StreamInfo::StreamInfo& streamInfo() PURE;
+
+  /**
+   * @return Shared pointer to the stream_info for this stream.
```
alyssawilk, line 199: Shared -> shared

envoy/http/codec.h
```diff
@@ -227,6 +194,56 @@ class RequestDecoder : public virtual StreamDecoder {
...
+   */
+  virtual std::shared_ptr<StreamInfo::StreamInfo> streamInfoSharedPtr() PURE;
```
alyssawilk, line 201: I'm concerned about folks adding things to stream info that aren't also shared pointers, and will result in issues if the info outlives the stream.
Can you think of a way to test this?

source/common/stream_info/stream_info_impl.h
```diff
@@ -356,6 +364,7 @@ struct StreamInfoImpl : public StreamInfo {
   // TODO(agrawroh): Check if the owner of this storage outlives the StreamInfo. We should only copy
   // the string if it could outlive the StreamInfo.
   absl::optional<std::string> virtual_cluster_name_;
+  absl::optional<DeferredLoggingInfo> deferred_logging_info_;
```
alyssawilk, line 367: comment

alyssawilk (commented):
very cool addition!  First pass thoughts follow
/wait

envoy/http/codec.h
pksohn: It's moved lower in the file so that is can refer to RequestDecoder (in the new setRequestDecoder function).

alyssawilk: WDYT of just pre-declaring so we don't lose the diffs?  If that doesn't work you can just mark this resolved.

source/common/http/http2/codec_impl.h
```diff
@@ -520,12 +520,13 @@ class ConnectionImpl : public virtual Connection,
     void encodeTrailers(const ResponseTrailerMap& trailers) override {
       encodeTrailersBase(trailers);
     }
+    void setRequestDecoder(Http::RequestDecoder& decoder) override { request_decoder_ = &decoder; }
```
danzh2010, line 523: Is request_decoder_ assigned somewhere already? Please make it private and use this setter in those call sites.

test/integration/quic_http_integration_test.cc
```diff
@@ -1167,6 +1167,86 @@ TEST_P(QuicHttpIntegrationTest, MultipleNetworkFilters) {
...
+  initialize();
+  codec_client_ = makeHttpConnection(makeClientConnection(lookupPort("http")));
+  sendRequestAndWaitForResponse(default_request_headers_, /*request_size=*/0,
+                                default_response_headers_,
+                                /*response_size=*/0,
+                                /*upstream_index=*/0, TestUtility::DefaultTimeout);
+  codec_client_->close();
+
+  std::string log = waitForAccessLog(access_log_name_);
+  std::string delimiter = ",";
```
danzh2010, line 1181: const

test/integration/quic_http_integration_test.cc
```diff
@@ -1167,6 +1167,86 @@ TEST_P(QuicHttpIntegrationTest, MultipleNetworkFilters) {
...
+  std::string log = waitForAccessLog(access_log_name_, 0, false, nullptr);
+  std::string delimiter = ",";
+  std::string protocol = log.substr(0, log.find(delimiter));
+  std::string roundtrip_duration = log.substr(log.find(delimiter) + 1, log.length());
+  EXPECT_EQ(protocol, "HTTP/3");
+  EXPECT_EQ(roundtrip_duration, "-");
+}
+
+TEST_P(QuicHttpIntegrationTest, DeferredLoggingWithReset) {
+  useAccessLog("%PROTOCOL%,%ROUNDTRIP_DURATION%");
```
danzh2010, line 1210: please also check that other metrics are logged.

test/integration/quic_http_integration_test.cc
```diff
@@ -1167,6 +1167,86 @@ TEST_P(QuicHttpIntegrationTest, MultipleNetworkFilters) {
...
+  ASSERT_TRUE(response->complete());
+
+  std::string log = waitForAccessLog(access_log_name_);
+  std::string delimiter = ",";
+  std::string protocol = log.substr(0, log.find(delimiter));
+  std::string roundtrip_duration = log.substr(log.find(delimiter) + 1, log.length());
+  EXPECT_EQ(protocol, "HTTP/3");
+  EXPECT_EQ(roundtrip_duration, "-");
+}
+
```
danzh2010, line 1249: Please also add a test case where the client reset the request after fake_upstream_ has finish sending response to verify that log_handlers log() is called all the time.

source/common/http/conn_manager_impl.h
```diff
@@ -192,7 +192,7 @@ class ConnectionManagerImpl : Logger::Loggable<Logger::Id::http>,
     void maybeEndDecode(bool end_stream);
 
     // Http::RequestDecoder
-    void decodeHeaders(RequestHeaderMapPtr&& headers, bool end_stream) override;
+    void decodeHeaders(RequestHeaderMapSharedPtr&& headers, bool end_stream) override;
```
danzh2010, line 195: Is this interface change needed? It can take a unique_ptr and store the pointee in a shared_ptr.

envoy/http/codec.h
pksohn: What do you think about a test similar to StreamInfoImplTest::SetFrom where we:

1) assert that the size of StreamInfo hasn't changed (if it has, print warning that new fields need to be shared)
2) set all the fields of a StreamInfo, have the original variables all go out of scope, assert that each of those fields still point to something.

source/common/quic/quic_stats_gatherer.h
```diff
@@ -0,0 +1,44 @@
...
+  }
+  void setAccessLogHandlers(std::list<AccessLog::InstanceSharedPtr> handlers) {
+    access_log_handlers_ = handlers;
+  }
+  void setTimeSource(Envoy::TimeSource* time_source) { time_source_ = time_source; }
+
+private:
+  uint64_t bytes_outstanding_ = 0;
+  bool fin_sent_ = false;
+  std::list<std::shared_ptr<StreamInfo::StreamInfo>> stream_info_{};
```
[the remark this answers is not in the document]
pksohn: The case this is for is an internal redirect / recreated stream where we log multiple times despite a single response to the downstream client.

source/common/quic/quic_stats_gatherer.cc
```diff
@@ -0,0 +1,35 @@
...
+  if (bytes_outstanding_ == 0 && fin_sent_ && !logging_done_) {
+    doDeferredLog();
+  }
+}
+
+void QuicStatsGatherer::doDeferredLog() {
+  logging_done_ = true;
+  for (const std::shared_ptr<StreamInfo::StreamInfo>& stream_info : stream_info_) {
+    if (!stream_info->deferredLoggingInfo().has_value()) {
+      continue;
```
danzh2010, line 20: Please add a test to verify that in redirect case, the response time is correctly calculated.

source/common/http/conn_manager_impl.h
pksohn: Agreed that keeping the interface unchanged is preferable.

I tried changing the interface back to accept a unique ptr, and leaving the member variable as a shared ptr, but ran into the problem that in ActiveStream::RecreateStream we call decodeHeaders on the request_headers_ member variable: https://github.com/envoyproxy/envoy/blob/3752119b94ae4a5ff337364f585f54305406dc7f/source/common/http/conn_manager_impl.cc#L1788. I'm not sure if there is a simple solution to this that I'm not seeing, do you have any suggestion?

envoy/http/codec.h
pksohn: done, thanks for the suggestion.

source/common/http/conn_manager_impl.cc
pksohn: done.

source/common/http/conn_manager_impl.cc
pksohn: done.

envoy/http/codec.h
pksohn: done.

source/common/stream_info/stream_info_impl.h
pksohn: Done.

source/common/quic/quic_stats_gatherer.h
```diff
@@ -0,0 +1,44 @@
...
+#include "envoy/http/codec.h"
+#include "envoy/http/header_map.h"
+#include "envoy/stream_info/stream_info.h"
+
+#include "quiche/quic/core/quic_ack_listener_interface.h"
+
+namespace Envoy {
+namespace Quic {
+
+class QuicStatsGatherer : public quic::QuicAckListenerInterface {
```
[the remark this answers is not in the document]
pksohn: Done.

source/common/http/http2/codec_impl.h
pksohn: Made this private. The one assignment was already changed to using this setter.

source/common/quic/quic_stats_gatherer.cc
```diff
@@ -0,0 +1,35 @@
...
+
+void QuicStatsGatherer::OnPacketAcked(int acked_bytes,
+                                      quic::QuicTime::Delta /* delta_largest_observed */) {
+  bytes_outstanding_ -= acked_bytes;
+  if (bytes_outstanding_ == 0 && fin_sent_ && !logging_done_) {
+    doDeferredLog();
+  }
+}
+
+void QuicStatsGatherer::doDeferredLog() {
```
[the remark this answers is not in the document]
pksohn: Done.

source/common/quic/envoy_quic_server_stream.h
```diff
@@ -18,7 +20,12 @@ class EnvoyQuicServerStream : public quic::QuicSpdyServerStreamBase,
                         envoy::config::core::v3::HttpProtocolOptions::HeadersWithUnderscoresAction
                             headers_with_underscores_action);
 
-  void setRequestDecoder(Http::RequestDecoder& decoder) { request_decoder_ = &decoder; }
+  void setRequestDecoder(Http::RequestDecoder& decoder) override {
+    request_decoder_ = &decoder;
+    stats_gatherer_->addStreamInfo(request_decoder_->streamInfoSharedPtr());
+    stats_gatherer_->setAccessLogHandlers(request_decoder_->accessLogHandlers());
+    stats_gatherer_->setTimeSource(&connection()->dispatcher().timeSource());
```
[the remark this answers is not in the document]
pksohn: Done.

source/common/quic/envoy_quic_server_stream.cc
```diff
@@ -381,6 +388,7 @@ void EnvoyQuicServerStream::OnClose() {
     return;
   }
   clearWatermarkBuffer();
+  stats_gatherer_ = nullptr;
```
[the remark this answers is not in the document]
pksohn: My understanding was that it's fine and expected that the last ack may happen after this point, and that as long as the QUIC stream has a pointer to it (https://github.com/google/quiche/blob/ce1352545bf94e572e76afed3e72d64a2198436f/quiche/quic/core/http/quic_spdy_stream.h#L350), it should live on long enough to do its logging. Let me know if I'm missing something.

source/common/http/conn_manager_impl.cc
pksohn: This is new, we didn't previously set the new request decoder. Perhaps related to the comment in 1773-1775 (and perhaps I should change that comment).

test/integration/quic_http_integration_test.cc
pksohn: I thought that the existing `DeferredLoggingWithReset` covered this case:

```
  waitForNextUpstreamRequest(0);
  codec_client_->close();
```

Can you clarify if I'm missing something?

source/common/quic/quic_stats_gatherer.h
```diff
@@ -16,6 +16,8 @@ namespace Quic {
 // logging after the final ack.
 class QuicStatsGatherer : public quic::QuicAckListenerInterface {
 public:
+  QuicStatsGatherer(Envoy::TimeSource* time_source) : time_source_(time_source) {}
```
danzh2010, line 19: missing explicit modifier.

envoy/stream_info/stream_info.h
```diff
@@ -758,6 +773,16 @@ class StreamInfo {
    * This function is called on Envoy fatal errors so should avoid memory allocation.
    */
   virtual void dumpState(std::ostream& os, int indent_level = 0) const PURE;
+
+  /**
+   * @param deferredLoggingInfo request/response info for deferred logging.
```
danzh2010, line 778: nits: s/info/headers and trailers

How about changing this interface to be `storeDeferredLoggignInfo(request_header_map_shard_ptr, response_header_map_shard_ptr, trailer_map_shard_ptr)` so that ActiveStream::deferredLoggingInfo() is no longer needed.

test/integration/quic_http_integration_test.cc
danzh2010: `waitForNextUpstreamRequest(0)` doesn't guarantee that response has been sent back. The test case that I asked for would be able to verify that the deferred logging should be triggered by stream reset while waiting for the ACK.

source/common/quic/quic_stats_gatherer.h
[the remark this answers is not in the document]
danzh2010: Ah, I see. In the internal redirect case, should we not defer logging at all? So that this field only points to the latest stream_info created.

source/common/quic/envoy_quic_server_stream.cc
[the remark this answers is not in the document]
danzh2010: I'm concerned that the stream get destroyed without calling deferred logging at all, i.e. getting reset.

danzh2010:
> Have you thought about make adding setRequestDecoder() to ResponseEncoder interface a stand alone PR?

WDYT about this?

pksohn:
```
Have you thought about make adding setRequestDecoder() to ResponseEncoder interface a stand alone PR?

WDYT about this?
```

Ack, I'll open a separate PR for this piece, as a prerequisite for this PR.

envoy/stream_info/stream_info.h
pksohn: I added ActiveStream::deferredLoggingInfo() to avoid having to create new getters for each of the shared pointers. There are getters for the raw pointers here: https://github.com/envoyproxy/envoy/blob/main/source/common/http/conn_manager_impl.h#L237-L251, but I'd need to add versions for each of the shared ones. LMK if you have any better ideas.

pksohn: I will change all of the "info"s to "headers and trailers".

source/common/quic/quic_stats_gatherer.h
pksohn: done.

test/integration/quic_http_integration_test.cc
pksohn: Added several more metrics.

test/integration/quic_http_integration_test.cc
pksohn: (now obsolete.)

source/common/quic/quic_stats_gatherer.h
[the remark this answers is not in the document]
pksohn: I think we should log as usual (directly from the stream destruction process) in the internal redirect case, as to not change behavior.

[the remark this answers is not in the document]
pksohn: Sorry, my previous comment contradicted itself. Let me think about this further.

source/common/http/conn_manager_impl.cc
```diff
@@ -278,7 +278,20 @@ void ConnectionManagerImpl::doDeferredStreamDestroy(ActiveStream& stream) {
...
+
+  // For HTTP/3, skip access logging here and add deferred logging info
+  // to stream info for QuicStatsGatherer to use later. If there was a
+  // downstream reset, log here as usual.
+  if (codec_ && codec_->protocol() == Protocol::Http3 &&
+      !stream.filter_manager_.sawDownstreamReset() &&
+      Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.quic_defer_logging_to_ack_listener")) {
+    stream.filter_manager_.streamInfo().setDeferredLoggingHeadersAndTrailers(
+        stream.DeferredLoggingHeadersAndTrailers());
```
danzh2010, line 290: Can you wrap this logic into into ActiveStream::deferLoggingHeadersAndTrailers()?

source/common/quic/envoy_quic_server_stream.h
```diff
@@ -18,7 +20,11 @@ class EnvoyQuicServerStream : public quic::QuicSpdyServerStreamBase,
                         envoy::config::core::v3::HttpProtocolOptions::HeadersWithUnderscoresAction
                             headers_with_underscores_action);
 
-  void setRequestDecoder(Http::RequestDecoder& decoder) { request_decoder_ = &decoder; }
+  void setRequestDecoder(Http::RequestDecoder& decoder) override {
+    request_decoder_ = &decoder;
+    stats_gatherer_->setStreamInfo(request_decoder_->streamInfoSharedPtr());
```
danzh2010, line 25: As we update the request_decoder_ reference in the stream now, can the QuicStatsGatherer hold a handle to the stream and get the stream info and access log handlers from the stream in maybeDoDeferredLog() instead?

danzh2010: To answer myself, request_decoder_ short lives the QuicStream so this is not a good way.

source/common/stream_info/stream_info_impl.h
```diff
@@ -231,6 +231,16 @@ struct StreamInfoImpl : public StreamInfo {
     return virtual_cluster_name_;
   }
 
+  void setDeferredLoggingHeadersAndTrailers(
```
danzh2010, line 234: Can DeferredLoggingHeadersAndTrailers object lives in response decoder implementation which should be accessible by HCM::ActiveStream. The response decoder implementation also handles deferred logging, so there can be less indirection.

source/common/quic/quic_stats_gatherer.cc
pksohn: Done.

source/common/http/conn_manager_impl.h
```diff
@@ -201,6 +201,25 @@ class ConnectionManagerImpl : Logger::Loggable<Logger::Id::http>,
                         absl::string_view details) override {
       return filter_manager_.sendLocalReply(code, body, modify_headers, grpc_status, details);
     }
+    std::shared_ptr<StreamInfo::StreamInfo> streamInfoSharedPtr() override {
+      return filter_manager_.streamInfoSharedPtr();
+    }
+    std::list<AccessLog::InstanceSharedPtr> accessLogHandlers() override {
+      return filter_manager_.accessLogHandlers();
+    }
+    StreamInfo::DeferredLoggingInfo deferredLoggingInfo() {
```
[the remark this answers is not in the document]
pksohn: (this comment is now obsolete -- this has changed to a function to call setDeferredLoggingHeadersAndTrailers on the response encoder).

source/common/http/conn_manager_impl.cc
pksohn: Done.

source/common/stream_info/stream_info_impl.h
pksohn: Done.

envoy/http/codec.h
pksohn: After a lot of tinkering with this, @danzh2010 and I decided not to make this a shared ptr in the end, and instead copy the stream info into the deferred-logging object at the time of stream destroy. This involves a new `setFrom` method similar to `setFromForRecreateStream`, and an expanded test in case StreamInfo gets new fields.

source/common/quic/envoy_quic_server_stream.cc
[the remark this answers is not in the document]
pksohn: I've added a fix and test for this in https://github.com/envoyproxy/envoy/pull/23648/commits/9b31dcd30db1e24729eed8d7d1777cb030f21533.

test/integration/quic_http_integration_test.cc
pksohn: I've added a fix and test for this in https://github.com/envoyproxy/envoy/commit/9b31dcd30db1e24729eed8d7d1777cb030f21533.

test/common/stream_info/stream_info_impl_test.cc
```diff
@@ -227,29 +232,88 @@ TEST_F(StreamInfoImplTest, MiscSettersAndGetters) {
...
+         "need to add them to setFromForRecreateStream or setFrom! Current size "
       << sizeof(s1);
 #endif
 #endif
 #endif
 
   StreamInfoImpl s2(Http::Protocol::Http11, test_time_.timeSystem(), nullptr);
-  s2.setFromForRecreateStream(s1);
+  Http::TestRequestHeaderMapImpl headers2;
+  s2.setFrom(s1, &headers2);
```
danzh2010, line 277: Instead of changing the existing test to test a new interface, can you add another unit test that test the extra thing setFrom() does?

test/integration/quic_http_integration_test.cc
```diff
@@ -1167,6 +1167,207 @@ TEST_P(QuicHttpIntegrationTest, MultipleNetworkFilters) {
...
+  EXPECT_EQ(/* PROTOCOL */ metrics.at(0), "HTTP/3");
+  EXPECT_EQ(/* ROUNDTRIP_DURATION */ metrics.at(1), "-");
+  EXPECT_EQ(/* REQUEST_DURATION */ metrics.at(2), "-");
+  EXPECT_EQ(/* RESPONSE_DURATION */ metrics.at(3), "-");
+  EXPECT_EQ(/* RESPONSE_CODE */ metrics.at(4), "400");
+  EXPECT_EQ(/* BYTES_RECEIVED */ metrics.at(5), "0");
+  EXPECT_EQ(/* request headers */ metrics.at(19), metrics.at(20));
+}
+
+TEST_P(QuicHttpIntegrationTest, DeferredLoggingWithInternalRedirect) {
```
danzh2010, line 1293: Nice test!

danzh2010 (commented):
This is much cleaner! Thanks for cleaning up the interfaces!

source/common/quic/quic_stats_gatherer.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+  // Log this stream using available stream info and access loggers.
+  void maybeDoDeferredLog(bool record_ack_timing = true);
+  // Set list of pointers to access loggers.
+  void setAccessLogHandlers(std::list<AccessLog::InstanceSharedPtr> handlers) {
+    access_log_handlers_ = handlers;
+  }
+  // Set headers and trailers used for deferred logging.
+  void setDeferredLoggingHeadersAndTrailers(
+      Http::DeferredLoggingHeadersAndTrailers& headers_and_trailers) {
```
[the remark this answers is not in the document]
pksohn: I've removed DeferredLoggingHeadersAndTrailers entirely, thanks for the suggestion.

test/common/quic/envoy_quic_server_stream_test.cc
```diff
@@ -99,10 +102,20 @@ class EnvoyQuicServerStreamTest : public testing::Test {
     }
   }
 
+  void setStatsGathererDetails(std::list<AccessLog::InstanceSharedPtr> access_loggers,
```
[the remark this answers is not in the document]
pksohn: I've removed the friend declaration and added a getter for QuicStatsGatherer in the quic stream.

test/common/stream_info/stream_info_impl_test.cc
pksohn: Done.

source/common/http/conn_manager_impl.h
```diff
@@ -204,26 +204,16 @@ class ConnectionManagerImpl : Logger::Loggable<Logger::Id::http>,
...
-      }
       std::unique_ptr<StreamInfo::StreamInfoImpl> stream_info =
           std::make_unique<StreamInfo::StreamInfoImpl>(
               connection_manager_.codec_->protocol(), connection_manager_.time_source_,
               connection_manager_.read_callbacks_->connection().connectionInfoProviderSharedPtr());
       stream_info->setFrom(streamInfo(), request_headers_.get());
-      headers_and_trailers.stream_info = std::move(stream_info);
-      response_encoder_->setDeferredLoggingHeadersAndTrailers(headers_and_trailers);
+      response_encoder_->setDeferredLoggingHeadersAndTrailers(
+          request_headers_, response_headers_, response_trailers_, std::move(stream_info));
```
danzh2010, line 216: nice! One more nit: can you pass in a const reference of the current stream info and create the new one in setDeferredLoggingHeadersAndTrailers()?

danzh2010 (approved):
LGTM

RyanTheOptimist (approved):
LGTM! Thanks for doing this. Just two small comments. That being said, it's a bummer that this on stat requires switching StreamInfo over to shared ownership from explicit ownership. But I guess there's really no alternative.

@alyssawilk would you like to take another look or shall we send it off to cross-company review.

test/integration/base_integration_test.cc
```diff
@@ -524,6 +526,10 @@ std::string BaseIntegrationTest::waitForAccessLog(const std::string& filename, u
           << contents;
       return entries[entry];
     }
+    if (i % 25 == 0 && client_connection != nullptr) {
```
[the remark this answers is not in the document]
pksohn: It's the default QUIC delayed ack timer ([quiche code link](https://github.com/google/quiche/blob/c06013fca03cc95f662cb3b09ad582b0336258aa/quiche/quic/core/quic_constants.h#L137)). I'll clarify the comment, thanks.

danzh2010:
> This requires the stream info and everything in it to have shared ownership so that it can be used for logging even after its originating stream is destroyed.

StreamInfo doesn't need to have shared ownership now. Can you update this description accordingly?

pksohn:
> StreamInfo doesn't need to have shared ownership now. Can you update this description accordingly?

Done @danzh2010.

> That being said, it's a bummer that this on stat requires switching StreamInfo over to shared ownership from explicit ownership. But I guess there's really no alternative.

@RyanTheOptimist FYI the shared pointer to streaminfo was removed in 613f1c38e3fea659da394225d29335d4cef8b455. I'll update the now-out-of-date description.

RyanTheOptimist:
> > StreamInfo doesn't need to have shared ownership now. Can you update this description accordingly?
> 
> Done @danzh2010.
> 
> > That being said, it's a bummer that this on stat requires switching StreamInfo over to shared ownership from explicit ownership. But I guess there's really no alternative.
> 
> @RyanTheOptimist FYI the shared pointer to streaminfo was removed in [613f1c3](https://github.com/envoyproxy/envoy/commit/613f1c38e3fea659da394225d29335d4cef8b455). I'll update the now-out-of-date description.

Oh yes, true. Thought it does change ownerhship of the headers in the codec from unique to shared, right? We should probably mention that I think. @mattklein123 I'd love to get your eyes on that part of this PR. I think it's fine but would appreciate your input.

envoy/http/codec.h
```diff
@@ -174,6 +175,19 @@ class ResponseEncoder : public virtual StreamEncoder {
...
+   * @param request_header_map Request headers for this stream.
+   * @param response_header_map Response headers for this stream.
+   * @param response_trailer_map Response trailers for this stream.
+   * @param stream_info Stream info for this stream.
+   */
+  virtual void
+  setDeferredLoggingHeadersAndTrailers(Http::RequestHeaderMapSharedPtr request_header_map,
+                                       Http::ResponseHeaderMapSharedPtr response_header_map,
+                                       Http::ResponseTrailerMapSharedPtr response_trailer_map,
+                                       StreamInfo::StreamInfo& stream_info) PURE;
```
mattklein123, line 190: You probably already discussed this, but FWIW, I find this dance a bit odd. It's a bit confusing how the headers are passed back and forth now. Perhaps more comments? Also, if this is for logging, can we use a shared_ptr to const object here to make it more clear that these things are not modified after this point?

I do wonder if it would have been easier to understand to keep the deferred logging system inside the HCM (some kind of deferred logging manager) somehow triggered by a codec callback later, but that would be a bigger change so probably not worth it.

source/common/quic/envoy_quic_server_stream.h
```diff
@@ -32,6 +38,20 @@ class EnvoyQuicServerStream : public quic::QuicSpdyServerStreamBase,
...
+                                            Http::ResponseHeaderMapSharedPtr response_header_map,
+                                            Http::ResponseTrailerMapSharedPtr response_trailer_map,
+                                            StreamInfo::StreamInfo& stream_info) override {
+    std::unique_ptr<StreamInfo::StreamInfoImpl> new_stream_info =
+        std::make_unique<StreamInfo::StreamInfoImpl>(
+            filterManagerConnection()->dispatcher().timeSource(),
+            filterManagerConnection()->connectionInfoProviderSharedPtr());
+    new_stream_info->setFrom(stream_info, request_header_map.get());
+    stats_gatherer_->setDeferredLoggingHeadersAndTrailers(
+        request_header_map, response_header_map, response_trailer_map, std::move(new_stream_info));
```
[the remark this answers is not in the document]
pksohn: > If we need stream info to live longer, why don't we just make that a shared_ptr and allow it to live longer?

That was indeed the original plan, and this PR changed course in https://github.com/envoyproxy/envoy/commit/613f1c38e3fea659da394225d29335d4cef8b455. We ran into a dead end when ensuring that all of the members of the StreamInfo implementations were also shared, specifically, get/set requestHeaders in StreamInfoImpl. I believe it would have required updating a very widely used Filter interface with hundreds of call sites.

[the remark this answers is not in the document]
mattklein123: OK. I wonder if my alternate implementation idea would somehow make this cleaner? (Have deferred logging built into the HCM so that relevant things could be moved instead of copied?)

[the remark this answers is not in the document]
pksohn: The reason for this codec-centric implementation is to take advantage of the QUIC AckListener interface so we don't have to deal with any of the transport layer details.

We'd like to add deferred logging for H/1 and H/2 as well, and it seems like then it might make sense to take a look at building deferred logging into HCM. In any case I'd love to hear if you have any thoughts/ideas/sketches about how it would work in HCM.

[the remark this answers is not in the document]
mattklein123: > We'd like to add deferred logging for H/1 and H/2 as well, and it seems like then it might make sense to take a look at building deferred logging into HCM. In any case I'd love to hear if you have any thoughts/ideas/sketches about how it would work in HCM.

I would probably just have a "deferred logging manager" that is a Singleton created as part of the HCM config. For deferred logging, the relevant structures could be moved in to live longer than the HCM level connection at teardown. If we do this (move) you might actually not need most/all of these changes. Then, you could have the codec potentially register to provide callbacks to the deferred logging manager for a particular stream (and build time outs, etc. in to the  manager). In this way I think you could still mostly rely on the codec level ACK manager and to provide the callbacks.

envoy/http/codec.h
pksohn: > Also, if this is for logging, can we use a shared_ptr to const object here to make it more clear that these things are not modified after this point?

Done.

> Perhaps more comments?

I'll take a pass at more thorough comments/docstrings. Is there anything specifically/particularly confusing that I should clear up?

source/common/quic/envoy_quic_server_stream.h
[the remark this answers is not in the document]
danzh2010: We are not sure if **moving** deferred logging relevant objects when HCM finishes sending response is a safe approach, in case where the moved objects gets accessed by HCM ActiveStream afterwards. So we decided to shared the ownership. I would think even if we do all the deferred logging stuff in HCM, the relevant objects currently living ActiveStream still need to be extended by shared ownership instead of moving because of the same issue. Or did I miss understand your suggestion?

Another approach is to extend the ActiveStream lifetime, but it seems non-ideal to keep all the filters alive while waiting for ACKs.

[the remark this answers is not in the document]
mattklein123: TBH I haven't though about it / analyzed it in detail so you may be correct. This was just my gut instinct. Naively, I would think that if we defer logging until after ActiveStream goes away, we can just move things out of it? Why does it need to be shared? Basically we can move at the point we destroy ActiveStream.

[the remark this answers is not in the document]
mattklein123: Right but it could be moved even at the point of deferred destruction. IMO moving would be simpler to reason about in this case.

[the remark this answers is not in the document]
danzh2010: > Right but it could be moved even at the point of deferred destruction.

Do you mean moving the objects when the ActiveStream object is actually destroyed, i.e. in the destructor? That's an interesting idea! But it would need non-trival refactoring I guess. And given that ACK callbacks may be triggered before the actual destruction of ActiveStream, my gut feeling is that something still need to be shared with the codec.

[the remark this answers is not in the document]
mattklein123: > And given that ACK callbacks may be triggered before the actual destruction of ActiveStream, my gut feeling is that something still need to be shared with the codec.

I don't think anything would need to be shared if the codec triggers the cleanup callback. If it happens before destruction, you log and then destroy. If it happens after destruction it would be handled by the "cleanup manager."

envoy/http/codec.h
pksohn: I've added a few more comments on the methods that pass headers through to the codec.

source/common/quic/envoy_quic_server_stream.h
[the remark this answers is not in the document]
pksohn: @mattklein123 Thanks for the discussion. Are you ok with moving forward with the current implementation or do you think a change/refactor is required?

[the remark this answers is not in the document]
danzh2010: Hmm, I would think unique ownership requires more sophisticated lifetime handling than shared ownership. Do you think we can go with this implementation for now and figure out how to shared code with other codecs and how to manage the object lifetime better when we add implementations for HTTP/2 and HTTP/1?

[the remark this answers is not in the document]
mattklein123: I'm not thrilled about the current implementation as I think it's somewhat confusing and will have to be redone anyway once we do HTTP/1 and HTTP/2, but I'm not going to block it if you want to redo it later. I would at least add a TODO somewhere about where we want to go with this?

/wait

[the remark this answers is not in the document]
pksohn: Thanks Matt for the discussion here. I've added a TODO and we'll revisit this for later implementations as you suggest.

mattklein123 (approved):
Thank you! Will defer to @RyanTheOptimist for further review/merge.

RyanTheOptimist (approved):

RyanTheOptimist (approved):
Woohoo!

envoy/http/codec.h
```diff
@@ -236,6 +253,11 @@ class RequestDecoder : public virtual StreamDecoder {
    * @return StreamInfo::StreamInfo& the stream_info for this stream.
    */
   virtual StreamInfo::StreamInfo& streamInfo() PURE;
+
+  /**
+   * @return List of shared pointers to access loggers for this stream.
+   */
+  virtual std::list<AccessLog::InstanceSharedPtr> accessLogHandlers() PURE;
```
adisuissa, line 260: A drive-by comment:
Should this return a reference to the list?
Additionally, should this be const?
Looking at the uses of this method below I think the answer to the above 2 questions may be yes, and may avoid unnecessary copies.

PR #23666, opened 2022-10-25
[cache] Add completion callback to updateHeaders

ravenblackx:
Commit Message: Add completion callback to updateHeaders in HttpCache
Additional Description: This is necessary to facilitate testing of asynchronous cache implementations.
Risk Level: Low - cache filter is still tagged WIP.
Testing: Covered by existing tests (and future tests that rely on this change.)
Docs Changes: n/a
Release Notes: May require any existing HttpCache implementations to update to match the new interface.

### DIFF
--- a/source/extensions/filters/http/cache/http_cache.h
+++ b/source/extensions/filters/http/cache/http_cache.h
@@ -253,7 +253,7 @@ public:
   // This is called when an expired cache entry is successfully validated, to
   // update the cache entry.
   //
-  // The on_complete callback should be called with true if the update is successful,
+  // The on_complete callback is called with true if the update is successful,
   // false if the update was not performed.
   virtual void updateHeaders(const LookupContext& lookup_context,
                              const Http::ResponseHeaderMap& response_headers,
--- a/test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
+++ b/test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
@@ -61,7 +61,7 @@ HttpCacheImplementationTest::~HttpCacheImplementationTest() {
   delegate_->tearDown();
 }
 
-void HttpCacheImplementationTest::updateHeaders(
+bool HttpCacheImplementationTest::updateHeaders(
     absl::string_view request_path, const Http::TestResponseHeaderMapImpl& response_headers,
     const ResponseMetadata& metadata) {
   LookupContextPtr lookup_context = lookup(request_path);
@@ -71,7 +71,9 @@ void HttpCacheImplementationTest::updateHeaders(
   auto update_future = update_promise->get_future();
   if (std::future_status::ready != update_future.wait_for(std::chrono::seconds(5))) {
     EXPECT_TRUE(false) << "timed out in updateHeaders " << request_path;
+    return false;
   }
+  return update_future.get();
 }
 
 LookupContextPtr HttpCacheImplementationTest::lookup(absl::string_view request_path) {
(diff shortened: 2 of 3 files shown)


source/extensions/filters/http/cache/http_cache.h
```diff
@@ -252,9 +252,13 @@ class HttpCache {
   //
   // This is called when an expired cache entry is successfully validated, to
   // update the cache entry.
+  //
+  // The on_complete callback should be called with true if the update is successful,
```
jmarantz, line 256: s/should be/is/

test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
```diff
@@ -65,7 +65,13 @@ void HttpCacheImplementationTest::updateHeaders(
     absl::string_view request_path, const Http::TestResponseHeaderMapImpl& response_headers,
     const ResponseMetadata& metadata) {
   LookupContextPtr lookup_context = lookup(request_path);
-  cache()->updateHeaders(*lookup_context, response_headers, metadata);
+  auto update_promise = std::make_shared<std::promise<bool>>();
+  cache()->updateHeaders(*lookup_context, response_headers, metadata,
+                         [update_promise](bool result) { update_promise->set_value(result); });
```
jmarantz, line 70: can we test the value of result after the promise is called?

test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
```diff
@@ -161,12 +167,15 @@ absl::Status HttpCacheImplementationTest::insert(absl::string_view request_path,
 
 Http::ResponseHeaderMapPtr HttpCacheImplementationTest::getHeaders(LookupContext& context) {
   Http::ResponseHeaderMapPtr response_headers_ptr;
-  context.getHeaders([&response_headers_ptr](LookupResult&& lookup_result) {
+  auto headers_promise = std::make_shared<std::promise<Http::ResponseHeaderMapPtr>>();
+  context.getHeaders([headers_promise](LookupResult&& lookup_result) {
     EXPECT_NE(lookup_result.cache_entry_status_, CacheEntryStatus::Unusable);
     EXPECT_NE(lookup_result.headers_, nullptr);
-    response_headers_ptr = move(lookup_result.headers_);
+    headers_promise->set_value(move(lookup_result.headers_));
```
jmarantz, line 174: pre-existing, but where does `move` come from? should this be std::move?

jmarantz (commented):
lgtm modulo minor nits

source/extensions/filters/http/cache/http_cache.h
ravenblackx: I've made this change, but I'm not entirely comfortable with it - this is an interface against which the person reading the documentation is likely implementing a cache instance (since there's no other reason to be looking at this particular code). "Is" sounds like something that already happens. I would favor "must be", as a description of what the implementer needs to make happen. But if you want "is", you can have "is". :)

test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
ravenblackx: Made updateHeaders return the result, so it can be validated in either direction at the callsite, and added the corresponding validations (and also made the name of one of the tests more descriptive accordingly).

test/extensions/filters/http/cache/http_cache_implementation_test_common.cc
ravenblackx: Done.

jmarantz (approved):
thanks!

jmarantz (commented):
Actually I'll go ahead and merge -- you can write a PR to change "is" to "must be" if you like, though there was a pre-existing parallel use of "is" in the paragraph above the description of updateHeaders() which is why I think I suggested that in the first place.

Issue #23614, opened 2022-10-20
matchers: support format string as input and action


kyessenov:
Feature request for https://github.com/envoyproxy/envoy/issues/23503.

Add an extension that registers [SubstitutionFormatString](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/substitution_format_string.proto#envoy-v3-api-msg-config-core-v3-substitutionformatstring) as an HTTP and network input. Additionally, allow its usage as an `action` in the selection of the filter chain name in a listener.

This general approach makes it possible to use the filter state and the dynamic metadata in the unified matching framework.

PR #23669, opened 2022-10-25
matching: expose stream info to filter chain matching

kyessenov:

Commit Message: Supply `StreamInfo` to `findFilterChain` to allow matching on dynamic metadata and the filter state populated by the listener filters. Context: https://github.com/envoyproxy/envoy/issues/23614
Risk Level: low
Testing: regression (refactor)
Docs Changes: none
Release Notes: none

### DIFF
--- a/source/common/quic/active_quic_listener.cc
+++ b/source/common/quic/active_quic_listener.cc
@@ -54,8 +54,8 @@ ActiveQuicListener::ActiveQuicListener(
   crypto_config_ = std::make_unique<quic::QuicCryptoServerConfig>(
       absl::string_view(reinterpret_cast<char*>(random_seed_), sizeof(random_seed_)),
       quic::QuicRandom::GetInstance(),
-      proof_source_factory.createQuicProofSource(listen_socket_,
-                                                 listener_config.filterChainManager(), stats_),
+      proof_source_factory.createQuicProofSource(
+          listen_socket_, listener_config.filterChainManager(), stats_, dispatcher.timeSource()),
       quic::KeyExchangeSource::Default());
   auto connection_helper = std::make_unique<EnvoyQuicConnectionHelper>(dispatcher_);
   crypto_config_->AddDefaultConfig(random, connection_helper->GetClock(),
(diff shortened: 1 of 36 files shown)


soulxu (approved):
lgtm

kyessenov:
CC @ggreenway do you want to take a look?

source/common/quic/active_quic_listener.cc
```diff
@@ -54,8 +54,8 @@ ActiveQuicListener::ActiveQuicListener(
   crypto_config_ = std::make_unique<quic::QuicCryptoServerConfig>(
       absl::string_view(reinterpret_cast<char*>(random_seed_), sizeof(random_seed_)),
       quic::QuicRandom::GetInstance(),
-      proof_source_factory.createQuicProofSource(listen_socket_,
-                                                 listener_config.filterChainManager(), stats_),
+      proof_source_factory.createQuicProofSource(
+          listen_socket_, listener_config.filterChainManager(), stats_, dispatcher.timeSource()),
```
ssripadham, line 58: What is the last argument for?

kyessenov: StreamInfo needs stream start time.

KBaichoo:
friendly ping @jmarantz , can you take a look? Thank you

jmarantz (commented):
this lgtm; thanks for plumbing this through.

My only thought is that it might have been better to plumb through an Api::Api& which can then give you the TimeSource, but would also make available other global-ish context such as stats root scope, thread factory etc.

That way if we need any of those in the future we'll won't need to re-plumb it.

WDYT?

/wait

kyessenov:
@jmarantz Looks like `QuicActiveListener`, which owns the proof verifier, doesn't actually have `Api::Api` to expose. The best we could do is `Dispatcher`?

jmarantz (approved):

ssripadham (approved):
Thanks for this change @kyessenov

kyessenov:
@yanavlasov I can follow up if I missed some comments. This is just a refactor so seems low risk.

PR #23670, opened 2022-10-25
thrift_proxy: Add reloadable feature flag to allow negative field ids

fishcakez:
Commit Message: Add a runtime reloadable feature flag to allow thrift negative field ids
Additional Description:
Add a runtime reloadable feature flag to allow negative field ids: envoy.reloadable_features.thrift_allow_negative_field_ids. Since apache thrift 0.8 the apache thrift compiler has by default disallowed negative field ids, however --allow-neg-keys can be used to support negative field ids introduced in a thrift schemas before 0.8. Unfortunately such old thrift schemas exist, and therefore the thrift_proxy is changed to be generous with what it accepts. Ideally there would be a deprecation window in this but apache thrift hasn't removed the deprecated feature in 11 years.

See https://github.com/apache/thrift/commit/c7206a40117da7d1f4a8a98ea52099825bea3b45
Risk Level: low - a closed source filter could depend on positive only field ids or upstream could depend on clients migrating to positive field ids and previously be blocking negative field ids
Testing: unit test
Docs Changes: N/A
Release Notes: Yes
Runtime guard: envoy.reloadable_features.thrift_allow_negative_field_ids


fishcakez:
cc @JuniorHsu.

Conflicted about adding a feature flag to allow deprecated behavior in apache thrift. However apache thrift haven't removed this behavior and we have one service using negative field ids from long ago.

wrowe:
/assign @zuercher 

for review once @fishcakez has a chance to resolve these regressions;
//test/extensions/filters/network/thrift_proxy:conn_manager_test
https://dev.azure.com/cncf/envoy/_build/results?buildId=119066&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec&l=1062

/wait

fishcakez:
@rgs1 @zuercher please take a look when you have time :)

rgs1 (changes requested):
Generally LGTM, with a nit/ask:

any chance we could consolidate all those checks in a `validateFieldId()` method that deals with the runtime check, differences between compact/binary and anything else? Would be nice to have one place with all the logic related to field id handling.

Thanks!

source/extensions/filters/network/thrift_proxy/compact_protocol_impl.cc
```diff
@@ -111,14 +112,20 @@ bool CompactProtocolImpl::peekReplyPayload(Buffer::Instance& buffer, ReplyType&
...
   // successful response struct in field id 0, error (IDL exception) in field id greater than 0
   reply_type = id == 0 ? ReplyType::Success : ReplyType::Error;
   return true;
 }
 
+void CompactProtocolImpl::validateFieldId(int32_t id) {
+  if ((id < 0 && !Runtime::runtimeFeatureEnabled(
+                     "envoy.reloadable_features.thrift_allow_negative_field_ids")) ||
+      id < std::numeric_limits<int16_t>::min() || id > std::numeric_limits<int16_t>::max()) {
+    throw EnvoyException(absl::StrCat("invalid compact protocol field id ", id));
```
zuercher, line 125: nit: this is pretty hard to read -- do an initial check for ids in the always-valid range and return early and then do the extra check for the flag to decide whether to throw or not?

zuercher (commented):
Generally looks ok.

Do you intend to make negative field ids generally valid after some time? E.g., after a release or so this flag would be removed and Envoy would just accept negative field ids?

If not, this should probably a flag in the API.

source/extensions/filters/network/thrift_proxy/compact_protocol_impl.cc
fishcakez: I refactored both protocol's check to the suggested style.

fishcakez:
> Do you intend to make negative field ids generally valid after some time? E.g., after a release or so this flag would be removed and Envoy would just accept negative field ids?

Given it looks like apache thrift is not removing negative field id support, likely we want to be generous with what we allow in the thrift proxy if the thrift compiler supports it. Therefore I'd expected to remove it in the release after next. However happy to move to an api config if maintainers think supporting it shouldn't be on by default in the long run if the thrift compiler doesn't allow it by default (but does allow opt in).

rgs1:
> > Do you intend to make negative field ids generally valid after some time? E.g., after a release or so this flag would be removed and Envoy would just accept negative field ids?
> 
> Given it looks like apache thrift is not removing negative field id support, likely we want to be generous with what we allow in the thrift proxy if the thrift compiler supports it. Therefore I'd expected to remove it in the release after next. However happy to move to an api config if maintainers think supporting it shouldn't be on by default in the long run if the thrift compiler doesn't allow it by default (but does allow opt in).

I don't think we should expose this via the API, it's good as is 👍.

rgs1 (approved):
Thanks!

zuercher (approved):
+1 to removing the flag in a future release.

Issue #23580, opened 2022-10-20
com_github_google_tcmalloc: 'asm' clobber conflict with output operand


sergey-safarov:
Fedora 37
```
ERROR: /root/.cache/bazel/_bazel_root/221703495c2e97a5482194eda3ea2f8b/external/com_github_google_tcmalloc/tcmalloc/BUILD:92:11: Compiling tcmalloc/tcmalloc.cc failed: (Exit 1): process-wrapper failed: error executing command 
  (cd /root/.cache/bazel/_bazel_root/221703495c2e97a5482194eda3ea2f8b/sandbox/processwrapper-sandbox/34/execroot/envoy && \
  exec env - \
    BAZEL_LINKLIBS=-l%:libstdc++.a \
    BAZEL_LINKOPTS=-lm \
    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
    PWD=/proc/self/cwd \
    TMPDIR=/tmp \
[...]

PR #23671, opened 2022-10-25
bazel: update tcmalloc

keith:
Fixes https://github.com/envoyproxy/envoy/issues/23580

Full diff: https://github.com/google/tcmalloc/compare/59400332b9cff9920b6a1da203ac1575272a9f44...e33c7bc60415127c104006d3301c96902f98d42a


wrowe:
/lgtm deps

However, CI pipelines need surgery

wrowe:
CI failures here appear to be regressions of //test/integration:stats_*, ptal

/wait

test/common/stats/thread_local_store_test.cc
```diff
@@ -1381,7 +1381,7 @@ TEST_F(StatsThreadLocalStoreTestNoFixture, MemoryWithoutTlsRealSymbolTable) {
   TestUtil::forEachSampleStat(
       100, true, [this](absl::string_view name) { store_.counterFromString(std::string(name)); });
   EXPECT_MEMORY_EQ(memory_test.consumedBytes(), 688080); // July 2, 2020
-  EXPECT_MEMORY_LE(memory_test.consumedBytes(), 0.75 * million_);
+  EXPECT_MEMORY_LE(memory_test.consumedBytes(), 0.85 * million_);
```
keith, line 1384: I do not know the implications of bumping these

wrowe: I believe this just indicates the predictable pattern isn't so predictable. But there's a second question of whether this particular sort of tests belongs in envoyproxy or should be part of the testing of the underlying cpplib/absl implementations.

wrowe: It's just as possible that the change occurred in our own TestUtil::MemoryTest logic?

keith: it pretty reliably fails with this change, so I feel like there must be something in tcmalloc affecting it, but I agree that it could be better as an upstream test if that's all we're really testing. I didn't check how on the cusp of exceeding the previous limits we were before

Issue #23569, opened 2022-10-19
source/common/common/logger.cc: comparison is always true due to limited range of data type


sergey-safarov:
Fedora 37
```
ERROR: /tmp/envoy/source/common/common/BUILD:193:17: Compiling source/common/common/logger.cc failed: (Exit 1): process-wrapper failed: error executing command 
  (cd /root/.cache/bazel/_bazel_root/221703495c2e97a5482194eda3ea2f8b/sandbox/processwrapper-sandbox/2072/execroot/envoy && \
  exec env - \
    BAZEL_LINKLIBS=-l%:libstdc++.a \
    BAZEL_LINKOPTS=-lm \
    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
    PWD=/proc/self/cwd \
    TMPDIR=/tmp \
[...]

PR #23675, opened 2022-10-25
build: fix char signed vs unsigned warning

keith:
char defaults to signed on most x86 platforms, but unsigned on arm64. When targeting linux arm64 this produced a warning because character was always >= 0. This change circumvents this warning without changing behavior even though it seems pretty unreasonable to have a character less than 0 in this use case. Theoretically we could disable a warning flag instead but that differs between clang and gcc.

Fixes https://github.com/envoyproxy/envoy/issues/23569

### DIFF
--- a/source/common/common/json_escape_string.h
+++ b/source/common/common/json_escape_string.h
@@ -64,7 +64,7 @@ public:
         position += 2;
         break;
       default:
-        if (character >= 0x00 && character <= 0x1f) {
+        if (character == 0x00 || (character > 0x00 && character <= 0x1f)) {
           // Print character as unicode hex.
           sprintf(&result[position + 1], "u%04x", static_cast<int>(character));
           position += 6;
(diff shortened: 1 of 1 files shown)


keith:
The riskier approach would just be to drop the 0x00 comparison all together, I don't feel educated enough in the practical implications of that to make that without someone else who thinks that's safe 🙃

source/common/common/json_escape_string.h
```diff
@@ -64,7 +64,7 @@ class JsonEscaper {
         position += 2;
         break;
       default:
-        if (character >= 0x00 && character <= 0x1f) {
+        if (character == 0x00 || (character > 0x00 && character <= 0x1f)) {
```
zhxie, line 67: So in the x86 platform, will the extra comparison come with a performance cost?

wrowe: Easy enough to verify using https://godbolt.org/ to examine the resulting cpu assembly code, would you mind checking that at -O2?

zhxie: Oh, the site works like magic. There is no overhead after compiler optimization.

wrowe (approved):
lgtm

source/common/common/json_escape_string.h
keith: thanks for testing!

keith: for reference: https://godbolt.org/z/xrEfsd5Mq

PR #23676, opened 2022-10-25
build: making alwayslink optional in Enovy CC libraries

alyssawilk:
This adds a variable to change Envoy library defaults to not always link, while fixing up Envoy libraries using singletons and factory registrations to still always link. 

Risk Level: high
Testing: using existing tests
Docs Changes: n/a
Release Notes: yes
Part of https://github.com/envoyproxy/envoy-mobile/issues/175


alyssawilk:
cc @keith I really want to run one of the Envoy builds with this true and I'm not sure how to do that.  Ideas?

keith:
I don't follow, do you mean you want to run the binary, or CI, and do you mean envoy or envoy mobile?

alyssawilk:
I want to run one of the Envoy CIs with LIBRARY_DEFAULT_ALWAYSLINK = 0
Given this is currently defined in a file I don't think I can - I'd have to fix it upstream in E-M.  Is there a way to get it from an environment variable or other so I can set it to 0 in one of the Envoy builds and prevent regressions?

ci/osx-build-config/extensions_build_config.bzl
```diff
@@ -18,9 +18,6 @@ EXTENSION_CONFIG_VISIBILITY = ["//:extension_config"]
...
 
-# Set this variable to true to disable alwayslink for envoy_cc_library.
-LIBRARY_DEFAULT_ALWAYSLINK = 1
-
-# For any users who want to further disable alwayslink, this can be set to 0.
-# TODO(alyssawilk) audit uses of this in source/ and migrate all libraries to extensions.
-EXTENSION_DEFAULT_ALWAYSLINK = 1
+# As part of (https://github.com/envoyproxy/envoy-mobile/issues/175) we turned down alwayslink for envoy libraries
+# This tracks libraries that should be registered as extensions.
+LEGACY_ALWAYSLINK = 1
```
keith, line 23: looks like the call sites are trying to use `AUTOLINK`, not `ALWAYSLINK`. I think this one is more correct because autolinking does exist and isn't related to this

bazel/envoy_library.bzl
```diff
@@ -101,10 +104,19 @@ def envoy_cc_library(
...
         include_prefix = None,
         textual_hdrs = None,
+        alwayslink = None,
         defines = []):
     if tcmalloc_dep:
         deps += tcmalloc_external_deps(repository)
 
+    # If alwayslink is not specified, allow turning it off via --define=library_autolink=disabled
+    # alwayslink is defaulted on for envoy_cc_extensions to ensure the REGISTRY macros work.
+    if not (alwayslink):
```
keith, line 114: if this is just the normal `not` keyword I think we need to check against None explicitly, otherwise I think `0` will be false-y

```suggestion
    if alwayslink == None:
```

keith (approved):

PR #23679, opened 2022-10-26
test: clean up unused variable

JuniorHsu:

Risk Level: low
Testing: unit


wrowe:
/assign @ggreenway

PTAL - I think you were one of the authors? Just making sure this wasn't an indication of an intended test that wasn't fleshed out.

PR #23654, opened 2022-10-25
repo: Yaml cleanups

phlax:
This is an initial cleanup PR ahead of addressing https://github.com/envoyproxy/pytooling/issues/267 to resolve https://github.com/envoyproxy/envoy/issues/23348 and https://github.com/envoyproxy/envoy/issues/22222

Signed-off-by: Ryan Northey <ryan@synca.io>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #23708, opened 2022-10-27
repo: Add yaml linting

### LINKED ISSUE #22222: Improve changelog linting
A couple of issues are cropping up in the changelog files

Currently it doesnt check for the text type (ie missing `|`) on the change strings

there is also a regression where it no longer checks for capitalization on the first word

there may be others that we want to add

---

- check structure of changelogs - ie have area/change for every entry
- prevent editing of historical changelogs

### LINKED ISSUE #23348: improved runtime guard checks
right now runtime guard checks appear to only work if the guard is mentioned in the first change.  If one adds envoy.reloadable_features.foo

```yaml
area: http
  change: |
    changed implemented something guarded by  ``envoy.reloadable_features.foo`` 
```

will pass but

```yaml
area: http
  change: |
    preexisting feature
  change: |
    changed implemented something guarded by  ``envoy.reloadable_features.foo`` 
```

appears to fail

phlax:
Fix #22222 
Fix #23348 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/tools/base/requirements.in
+++ b/tools/base/requirements.in
@@ -4,8 +4,8 @@ aiohttp>=3.8.1
 cffi>=1.15.0
 colorama
 coloredlogs
-envoy.base.utils>=0.3.7
-envoy.code.check>=0.3.0
+envoy.base.utils>=0.3.8
+envoy.code.check>=0.3.3
 envoy.dependency.check>=0.1.5
 envoy.dependency.pip_check>=0.1.2
 envoy.distribution.release>=0.0.8
--- a/tools/base/requirements.txt
+++ b/tools/base/requirements.txt
@@ -324,9 +324,9 @@ docutils==0.16 \
     #   sphinx
     #   sphinx-rtd-theme
     #   sphinx-tabs
-envoy-base-utils==0.3.7 \
-    --hash=sha256:5c79aa70de6b7014a5b955822eb83324fd222b7cf7bb656e12a018c747b4f2df \
-    --hash=sha256:ac7850ec1c90253964a5d7f7847a3c3317f9d77607e6f90d1fb5aeb975ac2842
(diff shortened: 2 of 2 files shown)


phlax:
before landing this we need to clean up the yaml - #23654 

we also need to resolve the changelog formatter which is currently adding an extraneous new line https://github.com/envoyproxy/pytooling/issues/616

phlax:
cc @mattklein123

phlax:
here you can see the checker failing the current yaml - ie before #23654 

https://dev.azure.com/cncf/envoy/_build/results?buildId=119208&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39&l=557

and here you can see current with the cleanups from #23654 added:

https://dev.azure.com/cncf/envoy/_build/results?buildId=119210&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39&l=492

note that the config is set to warn for lines > 200 - which triggers a lot of warnings - i will add a cleanup PR subsequently to address the long lines - its mostly changelogs

i think warning and the 200 limit is the right balance for this - but open to suggestions otherwise

some files are excluded - mostly templates - currently this is hardcoded upstream - i will make it configurable in a future iteration

there is no autofix currently added - it would be possible to add with a little complexity - but in order to do so we would need to be much stricter about the format - i think just linting with yamllint is good for now

RyanTheOptimist (approved):

PR #23717, opened 2022-10-27
Fix simple_http_cache updateHeaders for vary

ravenblackx:
Commit Message: Fix simple_http_cache updateHeaders for vary
Additional Description: There was a TODO for this, and a test that validates the behavior was incorrect in a way that matched the implementation, with comments falsely suggesting it was validating the correct behavior. This change makes the behavior correct and the validation correct, thereby enabling other correct cache implementations to pass the test suite.
Risk Level: Very low. Changes only to WIP filter, only fixing incorrect behavior.
Testing: Fixed existing test.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
+++ b/source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
@@ -12,6 +12,25 @@ namespace HttpFilters {
 namespace Cache {
 namespace {
 
+// Returns a Key with the vary header added to custom_fields.
+// It is an error to call this with headers that don't include vary.
+// Returns nullopt if the vary headers in the response are not
+// compatible with the VaryAllowList in the LookupRequest.
+absl::optional<Key> variedRequestKey(const LookupRequest& request,
+                                     const Http::ResponseHeaderMap& response_headers) {
+  absl::btree_set<absl::string_view> vary_header_values =
+      VaryHeaderUtils::getVaryValues(response_headers);
+  ASSERT(!vary_header_values.empty());
+  const absl::optional<std::string> vary_identifier = VaryHeaderUtils::createVaryIdentifier(
+      request.varyAllowList(), vary_header_values, request.requestHeaders());
+  if (!vary_identifier.has_value()) {
+    return absl::nullopt;
+  }
+  Key varied_request_key = request.key();
+  varied_request_key.add_custom_fields(vary_identifier.value());
+  return varied_request_key;
+}
+
 class SimpleLookupContext : public LookupContext {
 public:
   SimpleLookupContext(SimpleHttpCache& cache, LookupRequest&& request)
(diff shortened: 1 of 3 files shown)


source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
```diff
@@ -148,19 +148,25 @@ void SimpleHttpCache::updateHeaders(const LookupContext& lookup_context,
...
     return;
   }
-  auto& entry = iter->second;
-
-  // TODO(tangsaidi) handle Vary header updates properly
-  if (VaryHeaderUtils::hasVary(*(entry.response_headers_))) {
-    on_complete(false);
-    return;
+  if (VaryHeaderUtils::hasVary(*iter->second.response_headers_)) {
+    auto varied_key =
```
jmarantz, line 157: please use an explicit type here as the type returned by variedRequestKey is not obvious at the call-site.

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
```diff
@@ -224,25 +230,35 @@ bool SimpleHttpCache::insert(const Key& key, Http::ResponseHeaderMapPtr&& respon
...
+  return varied_request_key;
+}
+
+SimpleHttpCache::Entry
+SimpleHttpCache::varyLookup(const LookupRequest& request,
+                            const Http::ResponseHeaderMapPtr& response_headers) {
+  // This method should be called from lookup, which holds the mutex for reading.
+  mutex_.AssertReaderHeld();
+
+  auto maybe_key = variedRequestKey(request, *response_headers);
```
jmarantz, line 257: explicit type

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
```diff
@@ -148,19 +148,25 @@ void SimpleHttpCache::updateHeaders(const LookupContext& lookup_context,
...
+      on_complete(false);
+      return;
+    }
+    iter = map_.find(varied_key.value());
+    if (iter == map_.end() || !iter->second.response_headers_) {
+      on_complete(false);
+      return;
+    }
   }
+  auto& entry = iter->second;
```
jmarantz, line 169: I think this needs to be an explicit type also

jmarantz (commented):
lgtm generally. Hopefully when you fix this, merge main, and upload CI will fix itself :)

jmarantz (commented):
this lgtm from a maintainer perspective. I'd just like to get @toddmgreer to weigh in on vary semantics.

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.h
```diff
@@ -25,6 +25,12 @@ class SimpleHttpCache : public HttpCache, public Singleton::Instance {
     Http::ResponseTrailerMapPtr trailers_;
   };
 
+  // Returns a Key with the vary header added to custom_fields.
+  // It is an error to call this with headers that don't include vary.
+  // Returns nullopt if the vary headers are no longer valid.
```
toddmgreer, line 30: Can you say what "no longer valid" means?

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
```diff
@@ -224,25 +230,35 @@ bool SimpleHttpCache::insert(const Key& key, Http::ResponseHeaderMapPtr&& respon
...
 }
 
-SimpleHttpCache::Entry
-SimpleHttpCache::varyLookup(const LookupRequest& request,
-                            const Http::ResponseHeaderMapPtr& response_headers) {
-  // This method should be called from lookup, which holds the mutex for reading.
-  mutex_.AssertReaderHeld();
-
+absl::optional<Key>
+SimpleHttpCache::variedRequestKey(const LookupRequest& request,
```
toddmgreer, line 234: variedRequestKey doesn't use any members of SimpleHttpCache, so there's no reason for it to be a member function. It would be better as either a file-scoped static or unnamed namespace member, or in VaryHeaderUtils.

toddmgreer (approved):
Thank you for improving SimpleHttpCache!

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.h
ravenblackx: Updated to
```
// Returns nullopt if the vary headers in the response are not
// compatible with the VaryAllowList in the LookupRequest.
```

source/extensions/filters/http/cache/simple_http_cache/simple_http_cache.cc
ravenblackx: Tried to move it to VaryHeaderUtils but that creates a circular dependency for LookupRequest, or requires splitting that parameter into an ugly three parameters which would make it sensible to add a helper wrapper in this file, at which point it's not really worth the extra indirection, so file-scoped static it is.

jmarantz:
Todd are you good with this or does it need further comments?

toddmgreer (approved):
Thanks! Looks good.

jmarantz (approved):

Issue #22661, opened 2022-08-11
[Question] Is it possible to provide a time_since_last_update metric in xDS subscription statistics?


cancecen:
I am aware of the `update_time` metric. My team wants to use this metric to observe if instances are receiving xDS updates in a timely fashion in our fleet and this is especially helpful in gray failure scenarios. What we want to alarm is really `time_since_last_update`. 

However `update_time` being an epoch timestamp is complicating monitoring/alerting for us. I would like to think this may be a complication in other alerting system as well. For instance:
1. We are able to monitor computed metrics in our system, however the system does not support a System.now() metric, so we cannot do a `update_time` - System.now() computation.
2. After a node goes down, our metrics system cache the gauges it last reported and they continue to appear on the timeline for a while. So, it is difficult to distinguish nodes going down from nodes going stale for xDS updates due to other reasons. If we had a `time_since_last_update` metric instead, this wouldn't be a problem.

My main question is - does this metric look reasonable to you, or is there a reason this was not provided in the first place?

cancecen:
@kyessenov since you labeled this issue - does this mean it is something that's doable and there's not a strong reason against it(in terms of current design/flow) ? Knowing that would help, and perhaps we can contribute.

kyessenov:
Envoy metrics have evolved over time and don't necessarily follow best modern practices. I think it's reasonable to propose usability improvements in the area of xDS status monitoring.

cancecen:
I am working on this.

PR #23720, opened 2022-10-27
subscription stats: Add a new metric to track the time since the last successful xds update.

cancecen:
This change addresses the following issue:
https://github.com/envoyproxy/envoy/issues/22661

Signed-off-by: Can Cecen <ccecen@netflix.com>

Commit Message: Add a new metric to track the time since the last successful xds update.
Additional Description: More context here https://github.com/envoyproxy/envoy/issues/22661
Adding a time_since_last_update metric is makes it easier to alarm on, and can make the metric easier to reason about.
Risk Level: Low
Testing: Unit tests and manual testing.
Docs Changes: Updated metrics list.
Release Notes: n/a

### DIFF
--- a/envoy/config/subscription.h
+++ b/envoy/config/subscription.h
@@ -239,6 +239,7 @@ using SubscriptionPtr = std::unique_ptr<Subscription>;
   COUNTER(update_success)                                                                          \
   GAUGE(update_time, NeverImport)                                                                  \
   GAUGE(version, NeverImport)                                                                      \
+  GAUGE(time_since_last_update, NeverImport)                                                       \
   HISTOGRAM(update_duration, Milliseconds)                                                         \
   TEXT_READOUT(version_text)
 
@@ -249,6 +250,8 @@ struct SubscriptionStats {
   ALL_SUBSCRIPTION_STATS(GENERATE_COUNTER_STRUCT, GENERATE_GAUGE_STRUCT,
                          GENERATE_TEXT_READOUT_STRUCT, GENERATE_HISTOGRAM_STRUCT)
 };
+
+constexpr uint32_t PERIODIC_STATS_TIMER_REFRESH_MS = 10 * 1000;
 
 } // namespace Config
 } // namespace Envoy
--- a/docs/root/configuration/overview/mgmt_server.rst
+++ b/docs/root/configuration/overview/mgmt_server.rst
@@ -60,6 +60,7 @@ The following statistics are generated for all subscriptions.
  update_failure, Counter, Total API fetches that failed because of network errors
  update_rejected, Counter, Total API fetches that failed because of schema/validation errors
  update_time, Gauge, Timestamp of the last successful API fetch attempt as milliseconds since the epoch. Refreshed even after a trivial configuration reload that contained no configuration changes.
+ time_since_last_update, Gauge, Approximate milliseconds since the last successful API fetch attempt. Even a trivial configuration reload that contains no changes resets this stat to zero.
  version, Gauge, Hash of the contents from the last successful API fetch
  version_text, TextReadout, The version text from the last successful API fetch
  control_plane.connected_state, Gauge, A boolean (1 for connected and 0 for disconnected) that indicates the current connection state with management server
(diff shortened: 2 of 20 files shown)


envoy/config/subscription.h
```diff
@@ -250,5 +251,7 @@ struct SubscriptionStats {
                          GENERATE_TEXT_READOUT_STRUCT, GENERATE_HISTOGRAM_STRUCT)
 };
 
+constexpr uint32_t PERIODIC_STATS_TIMER_REFRESH_MS = 10 * 1000;
```
cancecen, line 254: resolution of this metric is not critical, so I picked an timer refresh rate here that I imagined to be not intrusive, but thinking about it making this something like every second should also not really have any impact. I'll wait for the reviewer to comment before changing anything.

adisuissa:
Thanks for your contribution.
I've got two main reservations about this approach:
1. The new stat is less accurate than `update_time`, as it is dependent on the timer kicking in in the right time.
2. Adding per-subscription stat that essentially contains similar info as the `update_time` is quite expensive (O(N) timers and stats, where N is the number of subscriptions).

Taking a step back, will adding an Envoy instance `timestamp` stat help in easing the alarm creation (assuming one can take 2 metrics, subtract one from the other, and create an alarm on the result)?

jmarantz:
+1 -- why can't the consumer read the last-update-time and subtract from the current time?

cancecen:
> Thanks for your contribution. I've got two main reservations about this approach:

Thanks for the review! @adisuissa 

> 
> 1. The new stat is less accurate than `update_time`, as it is dependent on the timer kicking in in the right time.

Correct. However, for "time since last success alerting" pattern, this is acceptable. We would like to know instances that are stale for whatever reason in a matter of minutes, so timer kicking off a bit later should be acceptable.

> 2. Adding per-subscription stat that essentially contains similar info as the `update_time` is quite expensive (O(N) timers and stats, where N is the number of subscriptions).

You are right. I think I have not thought about this in the right context (just thought about our current operating conditions) - defining the problem as O(N) makes this approach undesirable.
 
> Taking a step back, will adding an Envoy instance `timestamp` stat help in easing the alarm creation (assuming one can take 2 metrics, subtract one from the other, and create an alarm on the result)?
>+1 -- why can't the consumer read the last-update-time and subtract from the current time?

I am also going to answer @jmarantz's question here. Our system does not support a "macro" like `Time.now()`, I imagine some others would. So, adding a timestamp stat would help with the alarm creation. But I think I found a way better way of implementing this. Today we have a stats_sink that perform some transformations on the stats before sending it to the observability platform sidecar - and the sidecar itself supports implementing [this pattern](https://github.com/Netflix-Skunkworks/spectatord#metric-types). This way, when the stat makes it to the metrics backend, it'll already be computed (should give better accuracy for alarming.)

That being said, for those who might be in a similar situation, I think we could still solve this in Envoy - and I am willing to make that change if you think it would add value. Can you imagine doing this at a higher layer, i.e. one metric per process instead of subscription - essentially a new stat in `ControlPlaneStats`. I am very new to the codebase, so not sure how involved this would be. But again, my first question is whether you think this would add value or not.

adisuissa:
> I am also going to answer @jmarantz's question here. Our system does not support a "macro" like `Time.now()`, I imagine some others would. So, adding a timestamp stat would help with the alarm creation. But I think I found a way better way of implementing this. Today we have a stats_sink that perform some transformations on the stats before sending it to the observability platform sidecar - and the sidecar itself supports implementing [this pattern](https://github.com/Netflix-Skunkworks/spectatord#metric-types). This way, when the stat makes it to the metrics backend, it'll already be computed (should give better accuracy for alarming.)
> 
I see how this could work, and how it could transform any type/content as you need.

> That being said, for those who might be in a similar situation, I think we could still solve this in Envoy - and I am willing to make that change if you think it would add value. Can you imagine doing this at a higher layer, i.e. one metric per process instead of subscription - essentially a new stat in `ControlPlaneStats`. I am very new to the codebase, so not sure how involved this would be. But again, my first question is whether you think this would add value or not.

`ControlPlaneStats` is only used for gRPC-based servers, and so this will not be supported for filesystem/http subscriptions.
I think that either having a "global" proxy current time stat that will be used for subtraction later on the metrics ingestion part if possible, or a generic way to add the mutation during the sink step as you suggested are better approaches to cover the requirement.

Issue #23340, opened 2022-10-03
cannot build com_google_absl on Fedora:37


sergey-safarov:
I building envoy on Fedora 37 using command
```
bazel build //:envoy --config=clang --verbose_failures --sandbox_debug --subcommands --jobs 1 --cxxopt="-Wno-deprecated-builtins" --copt="-Wno-deprecated-builtins" --copt="-Wno-error" --cxxopt="-Wno-error"
```
But getting error
```
SUBCOMMAND: # @envoy_api//envoy/extensions/retry/host/omit_canary_hosts/v3:pkg_cc_proto_validate [action 'ProtoGenValidateCcGenerate external/envoy_api/envoy/extensions/retry/host/omit_canary_hosts/v3/omit_canary_hosts.pb.validate.h', configuration: 153a50b53db87b7ca58ddc5183810b4747ab3127cb18aaa8cfeea3386c020e25, execution platform: @local_config_platform//:host]
(cd /root/.cache/bazel/_bazel_root/4e47b77d977b209382cf04b3ae32963e/execroot/envoy && \
  exec env - \
    BAZEL_COMPILER=clang \
    BAZEL_LINKLIBS=-l%:libstdc++.a \
    BAZEL_LINKOPTS=-lm \
    CC=clang \
    CXX=clang++ \
    PATH=/root/.local/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
[...]

PR #23722, opened 2022-10-27
deps: update googleurl

keith:
This slightly improves our patch because upstream started validating `__has_attribute` exists itself now.

There is a little bit of new patching due to https://quiche-review.googlesource.com/c/googleurl/+/11180

Fixes https://github.com/envoyproxy/envoy/issues/23340


keith:
patch benefits ended up being a wash, but the good part is that most of the current patch should eventually make it upstream, which i never expected of the other pieces we're able to remove!

RyanTheOptimist (approved):

Issue #23696, opened 2022-10-26
Return 429s instead of 503s in adaptive concurrency filter


hochuenw-dd:
Hi team, 
When the adaptive concurrency filter of a server is triggered, clients may feel confused that the server is unavailable instead of that the server is proactively rejecting requests. I'm wondering if it makes more sense to change the status to 429s instead of [503s](https://github.com/envoyproxy/envoy/blob/v1.24.0/source/extensions/filters/http/adaptive_concurrency/adaptive_concurrency_filter.cc#L39)? It seems netflix's concurrency limits library uses [429s](https://github.com/Netflix/concurrency-limits/blob/ff90760ebda2af6c09d7e26a6eb58edb67b4ab40/concurrency-limits-servlet/src/main/java/com/netflix/concurrency/limits/servlet/ConcurrencyLimitServletFilter.java#L39).

RiverPhillips:
It would make sense to me if this was [configurable like the global rate limit.](https://github.com/envoyproxy/envoy/blob/ba81ae537e722ad99906599f5d3ad367ca854433/api/envoy/extensions/filters/http/ratelimit/v3/rate_limit.proto#L123) We could default to 503 but allow users to override it.

PR #23741, opened 2022-10-29
Enable returning a custom status code from the Adaptive Concurrency Filter

RiverPhillips:
Commit Message: Enable returning a custom status code from the Adaptive Concurrency Filter
Additional Description: Like global rate limit will not allow blocking with a successful status code, if an invalid status code is encountered it will fall back to 503.
Risk Level: Low
Testing: Unit tests added
Docs Changes: proto docs added
Release Notes: n/a
Fixes: #23696


api/envoy/extensions/filters/http/adaptive_concurrency/v3/adaptive_concurrency.proto
```diff
@@ -105,4 +106,12 @@ message AdaptiveConcurrency {
   // If set to false, the adaptive concurrency filter will operate as a pass-through filter. If the
   // message is unspecified, the filter will be enabled.
   config.core.v3.RuntimeFeatureFlag enabled = 2;
+
+  // This field allows for a custom HTTP response status code to the downstream client when
+  // the concurrency limit has been exceeded
```
adisuissa, line 111: nit:
```suggestion
  // the concurrency limit has been exceeded.
```

api/envoy/extensions/filters/http/adaptive_concurrency/v3/adaptive_concurrency.proto
```diff
@@ -105,4 +106,12 @@ message AdaptiveConcurrency {
...
+  // Defaults to 503 (Service Unavailable)
```
adisuissa, line 112: nit:
```suggestion
  // Defaults to 503 (Service Unavailable).
```

source/extensions/filters/http/adaptive_concurrency/adaptive_concurrency_filter.h
```diff
@@ -34,11 +34,21 @@ class AdaptiveConcurrencyFilterConfig {
 
   bool filterEnabled() const { return adaptive_concurrency_feature_.enabled(); }
   TimeSource& timeSource() const { return time_source_; }
+  Http::Code concurrencyLimitExceededStatus() const { return concurrency_limit_exceeded_status_; }
 
 private:
+  static Http::Code toErrorCode(uint64_t status) {
```
adisuissa, line 40: nit: Please move to an unnamed namespace inner function in the cc file.

test/extensions/filters/http/adaptive_concurrency/adaptive_concurrency_filter_test.cc
```diff
@@ -370,6 +370,79 @@ TEST_F(AdaptiveConcurrencyFilterTest, DisregardHealthChecks) {
...
+
+  Http::TestRequestHeaderMapImpl request_headers;
+
+  EXPECT_CALL(*controller_, forwardingDecision()).WillOnce(Return(RequestForwardingAction::Block));
+  EXPECT_CALL(decoder_callbacks_, sendLocalReply(Http::Code::TooManyRequests, _, _, _, _));
+  EXPECT_EQ(Http::FilterHeadersStatus::StopIteration,
+            filter_->decodeHeaders(request_headers, true));
+}
+
+TEST_F(AdaptiveConcurrencyFilterTest,
```
adisuissa, line 409: Please add a short comment before the tests describing them.

adisuissa (changes requested):
Thanks for adding this feature!
Overall looks good. Left a few minor comments.
/wait

source/extensions/filters/http/adaptive_concurrency/adaptive_concurrency_filter.h
RiverPhillips: I'm not super familiar with modern C++, in fact this is the first bit of C++ I've written since undergrad. Can you clarify what you mean by an "unnamed namespace inner function" or link to an example?

RiverPhillips:
Thanks for the feedback, I'm on holiday and will try to address this when I return early next week

source/extensions/filters/http/adaptive_concurrency/adaptive_concurrency_filter.h
adisuissa: Sure. Currently `toErrorCode` is a function that is only used in the cc file, and is also defined as private (shouldn't be used by other classes).
In this case we can move the function into the cc file inside an unnamed namespace, e.g.,
```
namespace {
  Http::Code toErrorCode(...) {
    ...
  }
}
```

Example can be seen [here](https://github.com/envoyproxy/envoy/blob/a2b595da4fd48c7c6fa399637f5001d0c0f0f6a5/source/extensions/filters/http/oauth2/config.cc#L25-L38) where `secretsProvider()` is defined only internally in this file.

RiverPhillips:
Should be resolved now

api/envoy/extensions/filters/http/adaptive_concurrency/v3/adaptive_concurrency.proto
```diff
@@ -105,4 +106,12 @@ message AdaptiveConcurrency {
...
   // message is unspecified, the filter will be enabled.
   config.core.v3.RuntimeFeatureFlag enabled = 2;
+
+  // This field allows for a custom HTTP response status code to the downstream client when
+  // the concurrency limit has been exceeded.
+  // Defaults to 503 (Service Unavailable).
+  //
+  // .. note::
+  //   If this is set to < 400, 503 will be used instead.
+  type.v3.HttpStatus concurrency_limit_exceeded_status = 3;
```
adisuissa, line 116: Sorry, one more question that I missed before, why is this in `AdaptiveConcurrency` and not in `GradientControllerConfig`?

RiverPhillips: I tried to keep adaptive concurrency consistent with `RateLimit`. The 503 was returned by `AdaptiveConcurrency` so I thought it made sense to configure the status code at this level. I got the impression that the `GradientControllerConfig` was related more to calculating the value than handling the response.

RiverPhillips: I'm happy either way, made sense for me to put it here as that's where the existing 503 was returned but a second opinion can't hurt

RiverPhillips: I looked at refactoring, at the moment the `GradientController` doesn't have any HTTP level concepts in it. It seems a bit strange to me to  introduce it for this use case

adisuissa: The way I read it, it seems that GradientController is currently a type of HTTP adaptive controller (could be others in the future), and that it imposes the limits.
As it also defines the concurrency limits, it may also define the return val.
That said, we can probably have this field as a filter level setting, like you are suggesting, because it points out to a configuration of the HTTP filter side of things. I guess we can have an override for this in the future, if additional controllers are added.

adisuissa (commented):
/lgtm api

Please add a release note for the new feature.

lizan (approved):
LGTM, defer merging to @adisuissa when release note is ready.

RiverPhillips:
Hi @adisuissa is there anything I can do to help move this forward?

alyssawilk:
@adisuissa ping!

adisuissa (approved):
Apologies for the late reply.
/lgtm api

adisuissa:
Sorry, just saw that @lizan already approved. Merging.

Issue #23719, opened 2022-10-27
IPv6 link local downstream endpoints not working


fcrisciani:
*Title*: *IPv6 link local downstream endpoints not working*

*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

Given a cluster with a backend endpoint such:
```
[...]
endpoint {
    address {
       socket_address {
            address: "fe80::1:1%<interface_name>"
             port_value: 1234
        }
     }
}
[...]
```
2 things should happen:
1) Envoy should be able to parse the IPv6 address with scope
2) Envoy should be able to create a socket that correctly sets the `sin6_scope_id` sockaddr_in6 (more info on link local (https://labs.ripe.net/author/philip_homburg/whats-the-deal-with-ipv6-link-local-addresses)

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

- IP parsing issue:
1) Create a cluster with link local IPv6 address and [scope](https://en.wikipedia.org/wiki/IPv6_address#Scoped_literal_IPv6_addresses_(with_zone_index))

today Envoy will reject this format.

- Socket for link local IP
1)  Create a cluster with link local IPv6 address

today Envoy will return:
```
D connection [third_party/envoy/src/source/common/network/connection_impl.cc:924] [C6380] connecting to [fe80::1:1]:1234
D connection [third_party/envoy/src/source/common/network/connection_impl.cc:948] [C6380] immediate connect error: Invalid argument
```
The reason is that the kernel wants sin6_scope_id to be set.

fcrisciani:
I validated that the issue is the one mentioned above just adding sin6_scope_id in [utility.cc](https://github.com/envoyproxy/envoy/blob/main/source/common/network/utility.cc#L106) and I was able to reach the downstream endpoint.

fcrisciani:
My proposal for the fix is the following:

1) change the IP parsing to allow to specify the scope
2) when available add the sin6_scope_id to the sockaddr_in6 returned.

DavidSchinazi:
IPv6 link-local identifiers aren't meant to be shared outside of a host. I'd suggest using routable IPv6 addresses such as ULA for your use-case.

yanavlasov:
My sense is that it is reasonable to expect Envoy to work with IPv6 link local addresses, since it is used to simplify network configuration. 

[...]

PR #23760, opened 2022-10-31
Add IPv6 link local address parsing

fcrisciani:
This commit adds support for IPv6 link local addresses. Link local addresses require the scope to be specified. The change basically identifies if the address string contains the `%` symbol and if so uses the `getaddrinfo()` method to parse the address.

Signed-off-by: Flavio Crisciani <fcrisciani@google.com>

Commit Message: add support for IPv6 link local addresses. Link local addresses require the scope to be specified. The change basically identifies if the address string contains the `%` symbol and if so uses the `getaddrinfo()` method to parse the address.
Additional Description:
Risk Level: low
Testing: manual testing. Configured a cluster passing a downstream socket address with IPv6 link local address with scope.
Docs Changes:
Release Notes: add support for IPv6 link local addresses.
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


fcrisciani:
This PR should address https://github.com/envoyproxy/envoy/issues/23719

yanavlasov (changes requested):
/wait

test/common/network/utility_test.cc
```diff
@@ -28,6 +33,41 @@ namespace Envoy {
...
+StatusOr<Interface> getLocalNetworkInterface() {
+  struct ifaddrs* ifaddr;
+
+  if (getifaddrs(&ifaddr) == -1) {
+    return absl::FailedPreconditionError("getifaddrs failed");
+  }
+
+  // Loop on all the interfaces and try to find one with IPv6 address.
+  for (struct ifaddrs* ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
+    if (ifa->ifa_addr == nullptr) {
```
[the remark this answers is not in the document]
fcrisciani: I don't think it is necessary just because the getaddrinfo and the if_nametoindex just care about the interface existence so any interface should be good.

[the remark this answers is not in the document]
fcrisciani: Updated the misleading comment that was referring to looking for an IPv6 address.

yanavlasov (approved):

test/common/network/utility_test.cc
```diff
@@ -140,6 +218,16 @@ TEST(NetworkUtility, ParseInternetAddress) {
   EXPECT_EQ("[1::2:3]:0", Utility::parseInternetAddressNoThrow("1::2:3")->asString());
   EXPECT_EQ("[a::1]:0", Utility::parseInternetAddressNoThrow("a::1")->asString());
   EXPECT_EQ("[a:b:c:d::]:0", Utility::parseInternetAddressNoThrow("a:b:c:d::")->asString());
+
+  StatusOr<Interface> ifc = getLocalNetworkInterface();
+  if (ifc.ok()) {
+    EXPECT_EQ(
+        absl::StrCat("[fe80::1]:0"),
```
soulxu, line 225: This means the address string won't output the scope. Two addresses with different scope will get the same string output. And `Utility::parseInternetAddressNoThrow` is used for the listener to parse the listening address also.

So a question about this, is there any case people want to use this kind of address in the listener? if yes, we have some code in the listener using the address string as the hash map key, for example:
https://github.com/envoyproxy/envoy/blob/60e0a125d84488e5aaf1e60d1770a672076c5e11/source/server/connection_handler_impl.h#L148, that won't work if different address generates same string.

Another question is, does the compare two addresses work for this kind of address?
https://github.com/envoyproxy/envoy/blob/60e0a125d84488e5aaf1e60d1770a672076c5e11/source/common/network/address_impl.cc#L294, maybe some code depends on the address comparing.

soulxu (changes requested):
two questions about whether Ipv6Instance support this kind of address well. Thanks!

yanavlasov:
Tagging @envoyproxy/api-shepherds This change has user visible behavior changes for config parsing. it will allow the scope to be added to IPv6 addresses. Before this change such config was rejected.

htuch:
Probably should be runtime guarded if it was possible to be parsing with `%` at all before, otherwise not. Out of curiosity, why not always use `getaddrinfo`?

fcrisciani:
@adisuissa @htuch I see this more as a bugfix just because the failure was not from the IP parsing but on the connect syscall returning Invalid Argument.

@htuch I was trying to keep the change as less invasive as possible, leaving everything as is and just add new logic if there was a scope. I'm happy to move to `getaddrinfo` for all parsing if you want. I noticed that between the utility file and the [address_impl.cc](https://github.com/envoyproxy/envoy/blob/60e0a125d84488e5aaf1e60d1770a672076c5e11/source/common/network/address_impl.cc) the parsing logic is duplicated making it more difficult to fix the link local parsing for all cases. Probably there is space for some cleanup that unify the way we go from string to sock_addr.

test/common/network/utility_test.cc
fcrisciani: for q1 I would say in theory yes, if you want to listen on the link local you will need the scope to be properly parse and set on the sockaddr_in6. Regarding the change to the way the string is formed I think it's more tricky. I empirically experienced that there is not a string format that is widely accepted. Example: the kernel likes the scope id as a uint32. The grpc server builder was throwing me an error if I was passing <ipv6>%<scope id> and wanted <ipv6>%<interface name>. 
Because of the above I'm unsure on what is the best path forward, maybe can be to leave the status quo and address it when there is a clear use case?

for q2 I introduced the scope to the Ipv6 base class, updated the `operator==` and added a unit test.

source/common/network/utility.cc
```diff
@@ -88,24 +88,63 @@ Api::IoCallUint64Result receiveMessage(uint64_t max_rx_datagram_size, Buffer::In
...
+    memset(&hints, 0, sizeof(hints));
+    struct addrinfo* res = nullptr;
+    // Suppresses any potentially lengthy network host address lookups and inhibit the invocation of
+    // a name resolution service.
+    hints.ai_flags = AI_NUMERICHOST | AI_NUMERICSERV;
+    hints.ai_family = AF_INET6;
+    // Hint that getaddrinfo() need not return a linked list of answers.
+    hints.ai_socktype = SOCK_DGRAM;
+    hints.ai_protocol = IPPROTO_UDP;
+    if (getaddrinfo(ip_address.c_str(), /*service=*/nullptr, &hints, &res) != 0) {
```
htuch, line 124: You probably want https://github.com/envoyproxy/envoy/blob/d93afd72e5b861ea8edcd97ebc26417d61f758c3/envoy/api/os_sys_calls.h#L285

htuch: I can't remember all the details, but I thought some platforms didn't have great support for `getaddrinfo`. But I might be confusing with `getaddrinfo_a`.

source/common/network/utility.cc
```diff
@@ -88,24 +88,63 @@ Api::IoCallUint64Result receiveMessage(uint64_t max_rx_datagram_size, Buffer::In
...
-  if (inet_pton(AF_INET6, ip_address.c_str(), &sa6.sin6_addr) == 1) {
+  const auto scope_pos = ip_address.rfind('%');
+  if (scope_pos == std::string::npos) {
+    // Parse IPv6 with no scope.
+    if (inet_pton(AF_INET6, ip_address.c_str(), &sa6.sin6_addr) != 1) {
+      return absl::FailedPreconditionError("failed parsing ipv6 with no scope");
+    }
     sa6.sin6_family = AF_INET6;
-    sa6.sin6_port = htons(port);
+  } else {
```
htuch, line 112: I think having a single path for parsing IP address is a win for simplicity. If you are concerned about possible disruption of this PR, I'd recommend adding a runtime flag to guard this.

test/common/network/utility_test.cc
soulxu: > for q1 I would say in theory yes, if you want to listen on the link local you will need the scope to be properly parse and set on the sockaddr_in6. Regarding the change to the way the string is formed I think it's more tricky. I empirically experienced that there is not a string format that is widely accepted. Example: the kernel likes the scope id as a uint32. The grpc server builder was throwing me an error if I was passing % and wanted %. Because of the above I'm unsure on what is the best path forward, maybe can be to leave the status quo and address it when there is a clear use case?
> 

I'm not sure whether we can leave it as, since it will be undefined behavior when people use it. I leave this to the maintainer to make the decision. @htuch @adisuissa 

There are other options here:
* Not using address string as the key. This is todo I left when I added multi-addresses listener, I could help this. But yea, need the maintainer to decide whether we can move this PR forward first.
* Just use '%' in the string output. Another benefit is we have correct log output. But this need to ensure we didn't pass the address string to other lib which doesn't support scope address as you said.

> for q2 I introduced the scope to the Ipv6 base class, updated the `operator==` and added a unit test.

Thanks for the update!

soulxu:
@fcrisciani just a reminder, it is better to not use force-push for updating the PR https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#submitting-a-pr
```
Once your PR is under review, please do not rebase it. If you rebase, you will need to force push to github, and github's user 
interface will force your reviewer to review the PR from scratch rather than simply look at your latest changes. It's much easier 
to review new commits and/or merges. We squash rebase the final merged commit so the number of commits you have in the 
PR don't matter. Again once your PR is assigned a reviewer, unless you need to fix DCO please do not force push. If you need to 
pull recent changes you can run
```

Thanks!

source/common/network/utility.cc
fcrisciani: Looks like the parsing between linux and windows is different and some basic parsing tests that are passing on linux are actually failing on windows. As suggested I'm following up with @wrowe for some suggestion on how to move forward here.  One option would be to go back to the first implementation where we use the inet_pton for everything and we fallback to the `getaddrinfo` only for IP parsing that has the zone specified. I will follow up once there is new information.

source/common/network/utility.cc
fcrisciani: resolving this one and we can continue the discussion on the comment above.

fcrisciani:
@soulxu noted! Sorry also @yanavlasov mentioned that to me but I thought was an issue only if I was modifying a commit that got reviewed, I will just pile them up then. I'm having some struggles with windows so during the weekend I had amended the topmost commit a bunch of times.

source/common/network/utility.cc
fcrisciani: Given the different behavior of windows, at this point I would rather decouple this fix from adopting the `getaddrinfo` for the parsing while there is a resolution to the discussion that I started on slack.

test/common/network/utility_test.cc
soulxu: Yea, I will take a look at hashing IP address, that is my todo list. And since you already update the code to output the scope in the address string, then we can move forward without waiting on hashing IP address.

source/common/network/address_impl.cc
```diff
@@ -247,6 +257,13 @@ std::string Ipv6Instance::Ipv6Helper::makeFriendlyAddress() const {
   char str[INET6_ADDRSTRLEN];
   const char* ptr = inet_ntop(AF_INET6, &address_.sin6_addr, str, INET6_ADDRSTRLEN);
   ASSERT(str == ptr);
+  if (address_.sin6_scope_id != 0) {
+    char if_name[IF_NAMESIZE]{'\0'};
+    if (if_indextoname(address_.sin6_scope_id, if_name) != nullptr) {
```
yanavlasov, line 262: This is will create a performance bottleneck if config has a lot of link local IPv6 addresses. We had to revert a PR that was resolving interface name for logging some time ago, because `if_indextoname` can take time to complete. This is not a problem right now, since such configs do not exist.

While I think it is nice to have the interface name instead of index in the friendly name, I do not think it is required. I think index would do just fine. My suggestion is to just use index for now and avoid any potential timing issues with calling `if_indextoname`.

yanavlasov (changes requested):
/wait

source/common/network/address_impl.cc
fcrisciani: done and added a comment to capture the above.

source/common/network/address_impl.cc
```diff
@@ -247,6 +257,12 @@ std::string Ipv6Instance::Ipv6Helper::makeFriendlyAddress() const {
   char str[INET6_ADDRSTRLEN];
   const char* ptr = inet_ntop(AF_INET6, &address_.sin6_addr, str, INET6_ADDRSTRLEN);
   ASSERT(str == ptr);
+  if (address_.sin6_scope_id != 0) {
+    // Note that here we don't use the `if_indextoname` that will give a more user friendly
+    // output just because in the past created a performance bottleneck if the machine had a
+    // lot of IPv6 Link local addresses.
```
htuch, line 263: @yanavlasov can we us  `if_indextoname` is this is strictly new behavior and opt-in?

source/common/network/utility.cc
```diff
@@ -88,24 +88,68 @@ Api::IoCallUint64Result receiveMessage(uint64_t max_rx_datagram_size, Buffer::In
...
+}
+
+StatusOr<sockaddr_in6> parseV6Address(const std::string& ip_address, uint16_t port) {
   sockaddr_in6 sa6;
   memset(&sa6, 0, sizeof(sa6));
-  if (inet_pton(AF_INET6, ip_address.c_str(), &sa6.sin6_addr) == 1) {
+  const auto scope_pos = ip_address.rfind('%');
+  // Even though it would be nice to do any IPv6 parsing only with the getaddrinfo at the moment
+  // Windows parsing is slightly different in behavior than other platforms. For this reason we use
+  // the inet_pton for any parsing that does not contain the zone id.
```
htuch, line 108: Please tag with TODO the places we will cleanup with https://github.com/envoyproxy/envoy/issues/23952.

fcrisciani: done

source/common/network/address_impl.cc
yanavlasov: Yes, it is safe for existing deployments. My concern was that it may come back to bite us when someone will use IPv6 link local addresses on machines with large number of IP interfaces.

As an aside the `makeFriendlyName` needs to be optimized (it is worth a few percentage points of CPU for configs with large number of EDS endpoints) and not even call inet_ntop for endpoint addresses that are already IP address strings and can just simply be copied into the friendly name. This will take care of resolving the interface name as well for the most part.

source/common/network/address_impl.cc
```diff
@@ -1,5 +1,13 @@
 #include "source/common/network/address_impl.h"
 
+#ifndef WIN32
```
yanavlasov, line 3: These headers should not be needed now.

yanavlasov (changes requested):
LGTM modulo cleanup of includes.

/wait

source/common/network/address_impl.cc
fcrisciani: removed

source/common/network/address_impl.cc
fcrisciani: Maybe I can try to take a look at that as part of [#23952](https://github.com/envoyproxy/envoy/issues/23952)

htuch (commented):
LGTM

htuch:
@ggreenway if you have a few cycles, could you take a look at this for a non-Google pass? Author and both reviewers are Googlers and we're making some decisions here that impact IPv6. From the Google side this LGTM.

yanavlasov (approved):

source/common/network/utility.cc
```diff
@@ -88,24 +88,68 @@ Api::IoCallUint64Result receiveMessage(uint64_t max_rx_datagram_size, Buffer::In
...
+  } else {
+    // Parse IPv6 with optional scope using getaddrinfo().
+    struct addrinfo hints;
+    memset(&hints, 0, sizeof(hints));
+    struct addrinfo* res = nullptr;
+    // Suppresses any potentially lengthy network host address lookups and inhibit the invocation of
+    // a name resolution service.
+    hints.ai_flags = AI_NUMERICHOST | AI_NUMERICSERV;
+    hints.ai_family = AF_INET6;
+    // Hint that getaddrinfo() need not return a linked list of answers.
```
ggreenway, line 124: Can you elaborate on this? How does setting DGRAM/UDP hint this? Is this linux-specific?

ggreenway (changes requested):
Looks good overall.

/wait

source/common/network/utility.cc
fcrisciani: It was mainly an empirical observation. I specify only the node and not the service, so specifying the socket_type I noticed that I was getting only one `addrinfo` struct while if leave it as 0 at least linux returns 2 elements SOCK_DGRAM & SOCK_STREAM.

fcrisciani: I reworked a bit the comment to give more context, let me know if it is less esoteric this way.

ggreenway (approved):
LGTM. Thanks!

Issue #23692, opened 2022-10-26
Build error: argument 1 null where non-null expected 


zonque:
This is seen on Fedora 36 with gcc 12.2.1:

```
In file included from /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/hashtable_policy.h:35,
                 from /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/hashtable.h:35,
                 from /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/unordered_map:46,
                 from /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/functional:61,
                 from external/v8/src/execution/isolate.h:10,
                 from external/v8/src/objects/js-temporal-objects.h:8,
                 from external/v8/src/objects/js-temporal-objects.cc:5:
In static member function 'static _Tp* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(const _Tp*, const _Tp*, _Tp*) [with _Tp = v8::internal::{anonymous}::Unit; bool _IsMove = false]',
    inlined from '_OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = const v8::internal::{anonymous}::Unit*; _OI = v8::internal::{anonymous}::Unit*]' at /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/stl_algobase.h:495:30,
    inlined from '_OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = const v8::internal::{anonymous}::Unit*; _OI = v8::internal::{anonymous}::Unit*]' at /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/stl_algobase.h:522:42,
    inlined from '_OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = const v8::internal::{anonymous}::Unit*; _OI = v8::internal::{anonymous}::Unit*]' at /usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/stl_algobase.h:529:31,
[...]

PR #23765, opened 2022-10-31
bazel: workaround invalid nonnull warning in v8

keith:
Since gcc 12 it seems there are some false positives around this warning.

Upstream gcc reports:

- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107252
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100366
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106199

We can't use `--per_file_copt` in this case because we build this in the host configuration https://github.com/bazelbuild/bazel/issues/12406

Upstream v8 report: https://bugs.chromium.org/p/v8/issues/detail?id=13441

Fixes https://github.com/envoyproxy/envoy/issues/23692

### DIFF
--- a/bazel/v8.patch
+++ b/bazel/v8.patch
@@ -31,7 +31,7 @@ index 4e89f90e7e..3fcb38b3f3 100644
  
  # Default setting for v8_enable_pointer_compression.
 diff --git a/bazel/defs.bzl b/bazel/defs.bzl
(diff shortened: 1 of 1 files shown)


phlax (approved):
lgtm, thanks @keith

PR #23766, opened 2022-10-31
tools, github: Pick authorization prefix based on provided token

dio:
Commit Message: Pick the authorization prefix based on the provided token. 
Additional Description: Reference: https://github.com/octokit/auth-token.js/blob/902a172693d08de998250bf4d8acb1fdb22377a4/src/with-authorization-prefix.ts#L6-L12
Risk Level: Low
Testing: Manual
Docs Changes: N/A
Release Notes: N/A

Signed-off-by: Dhi Aurrahman <dio@rockybars.com>

### DIFF
--- a/tools/github/write_current_source_version.py
+++ b/tools/github/write_current_source_version.py
@@ -70,7 +70,7 @@ if __name__ == "__main__":
         "https://api.github.com/repos/envoyproxy/envoy/commits/v" + current_version)
     if github_token := os.environ.get(args.github_api_token_env_name):
         # Reference: https://github.com/octokit/auth-token.js/blob/902a172693d08de998250bf4d8acb1fdb22377a4/src/with-authorization-prefix.ts#L6-L12.
-        authorization_header_prefix = "bearer" if len(github_token.split(".")) is 3 else "token"
+        authorization_header_prefix = "bearer" if len(github_token.split(".")) == 3 else "token"
         # To avoid rate-limited API calls.
         commit_info_request.add_header(
             "Authorization", f"{authorization_header_prefix} {github_token}")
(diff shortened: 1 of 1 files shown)


dio:
Hopefully, this works now.

tools/github/write_current_source_version.py
```diff
@@ -68,10 +68,12 @@
     # Fetch the current version commit information from GitHub.
     commit_info_request = urllib.request.Request(
         "https://api.github.com/repos/envoyproxy/envoy/commits/v" + current_version)
-    github_token = os.environ.get(args.github_api_token_env_name)
-    if github_token != None:
+    if github_token := os.environ.get(args.github_api_token_env_name):
+        # Reference: https://github.com/octokit/auth-token.js/blob/902a172693d08de998250bf4d8acb1fdb22377a4/src/with-authorization-prefix.ts#L6-L12.
+        authorization_header_prefix = "bearer" if len(github_token.split(".")) is 3 else "token"
```
phlax, line 73:
```suggestion
        authorization_header_prefix = "bearer" if len(github_token.split(".")) == 3 else "token"
```

```console
F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
```

i think our flake8 is a bit limited and needs to have more rules added

phlax (approved):
lgtm, thanks @dio 

im wondering if we need to add some info to the README about using a token to generate the `SOURCE_VERSION` file

also wondering about error handling - its very common for tokens to expire for example

dio:
Sure. Will add. When you do it locally seems like it is rare to experience rate-limited API calls for commit info (e.g. https://api.github.com/repos/envoyproxy/envoy/commits/v1.24.0). But yeah when you do from CI, e.g. GitHub Actions it has a bigger probability. Also, +1 for clean error handling.

phlax:
cool - i can review any follow ups - ill wait on the backport pr/s until you are happy with this on `main`

PR #23768, opened 2022-11-01
bazel: remove genrule_repository

keith:
This rule seems to have been used a lot more in the past, likely pre rules_foreign_cc. At this point it was only used for boringssl_fips, and we can easily replace that use with a normal genrule that just calls the same script. The script had to be updated slightly to remove the few dependencies on bazel specifics.


keith:
my actual motivation here was bazel keeps messing with Labels which broke the genrule_cmd accessor yet again. I think I can come up with a new solution for that but it doesn't seem like we need to maintain this anymore either

htuch (approved):
Nice cleanup!

Issue #23653, opened 2022-10-25
Utility::PercentEncoding::encode doesn't encode non-ascii character correctly


boringuy:
*Title*: *Utility::PercentEncoding::encode doesn't encode non-ascii character correctly*

*Description*:
Utility::PercentEncoding::encode() does not handle binary correctly because it ended up getting interpret as "negative" number due to the most significant bit being 1. 

*Repro steps*:
> Utility::PercentEncoding::encode("São Paulo") produces "S%-3D%-5Do Paulo" but it should be "S%C3%A3o Paulo"

>**Note**: N/A

*Admin and Stats Output*: N/A

>**Note**: N/A

*Config*: N/A

*Logs*: N/A

>**Note**: N/A

*Call Stack*: N/A

boringuy:
This is a simple fix and I already update the unit test to reproduce this. I will be submitting PR soon.

KBaichoo:
Thank you for looking into this @boringuy!

boringuy:
The fix is easy but I have not submitted the PR yet because I need to re-read the PR guideline as I have not done it for a while. It should happen this week.

PR #23781, opened 2022-11-01
http: fixed Utility::PercentEncoding::encode() to treat all values as unsigned

boringuy:
Fixes [#23653]

Signed-off-by: Andy Fong <boringuy@gmail.com>

Commit Message:  http: fixed Utility::PercentEncoding::encode() to treat all values as unsigned
Additional Description:
Risk Level: low
Testing: Added new input to existing tests to make sure the bug is fixed
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]
[Optional API Considerations:]


source/common/http/utility.cc
```diff
@@ -1006,7 +1006,7 @@ std::string Utility::PercentEncoding::encode(absl::string_view value, const size
     if (ch < ' ' || ch >= '~' || reserved_char_set.find(ch) != reserved_char_set.end()) {
       // For consistency, URI producers should use uppercase hexadecimal digits for all
       // percent-encodings. https://tools.ietf.org/html/rfc3986#section-2.1.
-      absl::StrAppend(&encoded, fmt::format("%{:02X}", ch));
+      absl::StrAppend(&encoded, fmt::format("%{:02X}", static_cast<const unsigned char&>(ch)));
```
wbpcode, line 1009: nit: would it better to convert `ch` to an unsigned char directly rather than convert the ref type by `static_cast<unsigned char>(ch)`

wbpcode:
And please fix the conflict first. thanks.

source/common/http/utility.cc
boringuy: I have not actually tried but if I converted ch to unsigned, will need to cast the const char in the if statement and when calling find(), so it's not as clean.

boringuy:
> 
conflict fixed. Can you review again before another conflict comes?

KBaichoo (approved):
Thanks for fixing this @boringuy

PR #23694, opened 2022-10-26
moving original_dst cluster to extensions

alyssawilk:
Any Envoy users who customize their pre-built extensions will need to evaluate if they need this cluster.

Risk Level: medium
Testing: n/a
Docs Changes: n/a
Release Notes: inline

PR #23799, opened 2022-11-02
Logical dns: moving to an extension

alyssawilk:
Any Envoy users who customize their pre-built extensions will need to evaluate if they need this cluster.
Akin to https://github.com/envoyproxy/envoy/pull/23694

Risk Level: medium
Testing: n/a
Docs Changes: n/a
Release Notes: inline


source/extensions/extensions_metadata.yaml
```diff
@@ -95,6 +95,11 @@ envoy.clusters.redis:
   - envoy.clusters
   security_posture: requires_trusted_downstream_and_upstream
   status: stable
+envoy.clusters.logical_dns:
+  categories:
+  - envoy.clusters
+  security_posture: requires_trusted_downstream_and_upstream
```
adisuissa, line 101: Is this the right posture? I was under the impression that as part of core clusters it was "robust_to_untrusted_downstream_and_upstream", but maybe I'm wrong.

test/exe/BUILD
```diff
@@ -69,6 +69,7 @@ envoy_cc_test(
         "//source/common/api:api_lib",
         "//source/exe:envoy_main_common_with_core_extensions_lib",
         "//source/exe:platform_impl_lib",
+        "//source/extensions/clusters/logical_dns:logical_dns_cluster_lib",
```
adisuissa, line 72: Hmm... I wonder if this target can be built w/o reliance on source/extensions.

test/integration/BUILD
```diff
@@ -589,6 +589,7 @@ envoy_cc_test_library(
         ":http_protocol_integration_lib",
         ":socket_interface_swap_lib",
         "//source/common/http:header_map_lib",
+        "//source/extensions/clusters/logical_dns:logical_dns_cluster_lib",
```
adisuissa, line 592: Similar question to the `//test/exe:main_common_test` target.
I guess it is ok because these are tests, but wonder if the extension is inherent to the tests functionality or is it testing non-core functionality (and then the test itself may be moved to the be part of the extension's tests)?

adisuissa (commented):
Thanks, left a few questions/comments.

test/exe/BUILD
alyssawilk: no, there's already a number of registered extensions such as source/extensions/transport_sockets/raw_buffer/config.h that envoy won't build without

source/extensions/extensions_metadata.yaml
alyssawilk: good catch thanks, fixed.

test/exe/BUILD
alyssawilk: that said this is why we have the visibility rules and an extra presubmit around extensions not generally being visible to core tests.  so it's a pretty well cross-checked system :-)

adisuissa (approved):
LGTM, thanks!

KBaichoo (approved):

Issue #23764, opened 2022-10-31
redis: hostnames in MOVED (or ASK) responses are not supported


rgs1:
The Redis Proxy assumes that `MOVED` and `ASK` responses will contain an IPv4/6 address as the given endpoint:

https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/network/redis_proxy/conn_pool_impl.cc#L312

And here is the Redis integration test failing when non IP endpoint is passed:

https://github.com/envoyproxy/envoy/blob/main/test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc#L822

However, the Redis cluster spec allows for hostnames to be used as well:

"The error includes the hash slot of the key (3999) and the endpoint:port of the instance that can serve the query. The client needs to reissue the query to the specified node's endpoint address and port. **The endpoint can be either an IP address, a hostname, or it can be empty (e.g. -MOVED 3999 :6380)**. An empty endpoint indicates that the server node has an unknown endpoint, and the client should send the next request to the same endpoint as the current request but with the provided port."

https://redis.io/docs/reference/cluster-spec/

Example response from memorydb in AWS that includes hostname in the response:

```
(error) MOVED 16287 something.some-region.amazonaws.com:6739
```

sc0ttbeardsley:
Welcome back @rgs1

rgs1:
So I think that it should be fairly straight forward to add support for DNS lookups on top of #23802. The only not great thing would be that the dns cache currently resides in `source/extensions/common/dynamic_forward_proxy/` and it might be nice to have it in a more common area path, but not a big deal.

rgs1:
Ok so I missed a path: `MOVED` responses that were triggered by a request within a transaction won't get a DNS lookup even if they contain hostnames.

PR #23802, opened 2022-11-02
[redis] refactor and simplify onRedirection()

rgs1:
As part of #23764, we need to simplify what the Redis Client does after calling onRedirection(), so that eventually the ConnPool can cleanly handle async DNS lookups.

Signed-off-by: Raul Gutierrez Segales <rgs@itevenworks.net>


mattklein123 (approved):
LGTM, thanks.

PR #23807, opened 2022-11-02
bazel: update googleurl patch with missing header

keith:
Upstream fix: https://quiche-review.googlesource.com/c/googleurl/+/11260


alyssawilk (approved):
Thanks for the fix!

htuch (approved):

PR #23808, opened 2022-11-02
build: Moving the strict DNS cluster into an extension directory

alyssawilk:
Any Envoy users who customize their pre-built extensions will need to evaluate if they need this cluster.
Akin to https://github.com/envoyproxy/envoy/pull/23694

Risk Level: medium
Testing: n/a
Docs Changes: n/a
Release Notes: inline


jmarantz (commented):
This looks simple enough but I'd like @yanjunxiang-google to look at it.

test/common/upstream/upstream_impl_test.cc
```diff
@@ -23,7 +23,7 @@
 #include "source/common/network/utility.h"
 #include "source/common/singleton/manager_impl.h"
 #include "source/common/upstream/static_cluster.h"
-#include "source/common/upstream/strict_dns_cluster.h"
+#include "source/extensions/clusters/strict_dns/strict_dns_cluster.h"
```
yanjunxiang-google, line 26: Does this break the rule that the core tests should not depend on extensions?

yanjunxiang-google: And will the envoy.clusters.strict_dns: be in the required extensions as in here:
https://github.com/envoyproxy/envoy/blob/e4ae2259add5a327c4d67de18a82df5d78a801b4/source/extensions/all_extensions.bzl#L6

alyssawilk: it does not - we frequently have core code in extensions directory (e.g. raw buffer socket) 
Also I don't think it hsould be in required extensions as the whole point of moving it out was to not link it for Envoy Mobile and we can't override that.

jmarantz (approved):

PR #23818, opened 2022-11-03
local ratelimit: make sure the timer of local limit could be destoryed safely

wbpcode:
Commit Message: local ratelimit: make sure the timer of local limit could be destoryed safely
Additional Description:

Timer is used by the `LocalRatelimitImpl`, so we should ensure the `LocalRatelimitImpl` always be destroyed in the thread where its inner timer is created and running.
However, the `LocalRatelimitImpl` in the `FilterConfig` of local rate limit filter may be created in the main thread and destroyed in the worker theads when the route configuration is updated by the RDS.

Risk Level: Low.
Testing: integration test.
Docs Changes: n/a.
Release Notes: n/a.

### DIFF
--- a/source/extensions/filters/http/local_ratelimit/local_ratelimit.h
+++ b/source/extensions/filters/http/local_ratelimit/local_ratelimit.h
@@ -74,10 +74,10 @@ public:
                Stats::Scope& scope, Runtime::Loader& runtime, bool per_route = false);
   ~FilterConfig() override {
     if (!main_dispatcher_.isThreadSafe()) {
-      main_dispatcher_.post(
-          [limiter_wrapper = std::make_shared<
-               std::unique_ptr<Filters::Common::LocalRateLimit::LocalRateLimiterImpl>>(
-               std::move(rate_limiter_))]() { limiter_wrapper->reset(); });
+      auto shared_ptr_wrapper =
+          std::make_shared<std::unique_ptr<Filters::Common::LocalRateLimit::LocalRateLimiterImpl>>(
+              std::move(rate_limiter_));
+      main_dispatcher_.post([shared_ptr_wrapper]() { shared_ptr_wrapper->reset(); });
     }
   }
   const LocalInfo::LocalInfo& localInfo() const { return local_info_; }
--- a/source/extensions/filters/http/local_ratelimit/local_ratelimit.h
+++ b/source/extensions/filters/http/local_ratelimit/local_ratelimit.h
@@ -70,9 +70,16 @@ private:
 class FilterConfig : public Router::RouteSpecificFilterConfig {
 public:
   FilterConfig(const envoy::extensions::filters::http::local_ratelimit::v3::LocalRateLimit& config,
-               const LocalInfo::LocalInfo& local_info, Event::Dispatcher& dispatcher,
+               const LocalInfo::LocalInfo& local_info, Event::Dispatcher& main_dispatcher,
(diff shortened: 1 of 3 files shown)


source/extensions/filters/http/local_ratelimit/local_ratelimit.h
```diff
@@ -70,9 +70,16 @@ class PerConnectionRateLimiter : public StreamInfo::FilterState::Object {
 class FilterConfig : public Router::RouteSpecificFilterConfig {
 public:
   FilterConfig(const envoy::extensions::filters::http::local_ratelimit::v3::LocalRateLimit& config,
-               const LocalInfo::LocalInfo& local_info, Event::Dispatcher& dispatcher,
+               const LocalInfo::LocalInfo& local_info, Event::Dispatcher& main_dispatcher,
                Stats::Scope& scope, Runtime::Loader& runtime, bool per_route = false);
-  ~FilterConfig() override = default;
+  ~FilterConfig() override {
+    if (!main_dispatcher_.isThreadSafe()) {
```
mattklein123, line 76: I would probably just skip this check and always post for consistency. I don't think it needs to be optimized. Can you also add more comments on why this is needed?

mattklein123 (changes requested):
Thanks. This makes sense to me, though it's unfortunate that this is the solution since it's pretty fragile. I've been out of this code for a while so I'm going to also assign to @kyessenov who has dealt with more of these types of lifetime issues. It sure seems like we could potentially have this type of problem in a lot of other places but perhaps timers are not that common in shared configuration for filters?

Either way thank you for fixing!

/wait

kyessenov:
I think at some point we were using the connection worker dispatcher? Was there some change that made the timers be from the main thread?

The solution looks fine, we did a similar thing for ECDS since Wasm config must also be deleted on main with something like `tls_.set(empty_value)` .

wbpcode:
> I think at some point we were using the connection worker dispatcher? Was there some change that made the timers be from the main thread?

Sorry, I made some unclear description. In fact, the timer should be destroyed in the thread where it is created and running.

So, if timer is created with worker dispatcher in the worker, then timer should be destroyed there. I think may the connection lifetime management could achieve this point (?)

So, only the `LocalRatelimitImpl` in the route-level `FilterConfig` has this issue.

wbpcode:
> Thanks. I'm still trying to understand if this is some structural issue with RDS and/or local rate limit or just a bug inside the filter. Many filters don't work well when destroyed on a worker, Wasm comes to mind. Let me take a closer look at where this is instantiated.

Every request (ActiveStream of HCM) will cache a shared pointer to the route config to ensure all the route configs will have longer lifetime than the request.
It means that when the route config is updated by the `rds`, the route config that cached in the request may destruct in the worker thread at last.

mattklein123:
I think in this case the config (and timer) are always made on the main thread, but due to shared_ptr latching it can be destroyed on workers. So I think this fix is correct. With that said, this comes up enough that I wish we had a more structural way of fixing this but that would be a ton of work so it seems OK to me to land this with comments. @kyessenov wdyt?

kyessenov (approved):
OK, thanks for explaining the RDS problem. Yeah, this is definitely a good enough fix for now. This is only affecting filters that use timers and/or TLS slots.

wbpcode:
cc @mattklein123 I think a new review from you is necessary to unblock the merge. thanks. 🌷

mattklein123 (approved):

PR #23825, opened 2022-11-03
[redis] do a dns lookup for endpoints on MOVED/ASK responses

### LINKED ISSUE #23764: redis: hostnames in MOVED (or ASK) responses are not supported

rgs1:
Fixes #23764.

Signed-off-by: Raul Gutierrez Segales <rsegales@nvidia.com>


api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
```diff
@@ -321,6 +322,11 @@ message RedisProxy {
   // client. If an AUTH command is received when the password is not set, then an "ERR Client sent
   // AUTH, but no ACL is set" error will be returned.
   config.core.v3.DataSource downstream_auth_username = 7 [(udpa.annotations.sensitive) = true];
+
+  // The DNS cache configuration that the filter will attach to and that will be used to resolve
+  // hostnames that are returned with MOVED and ASK responses. If not provided, no DNS lookups
+  // will be performed (and thus the MOVED/ASK errors will be propagated to the user).
+  common.dynamic_forward_proxy.v3.DnsCacheConfig dns_cache_config = 10;
```
rgs1, line 329: I just realized this should actually be another `ConnPoolSettings`, where `enable_redirection` is since a) DNS lookups will be handled by the ConnPool and b) it's a no-op to enable it if `enable_redirection` is `off`.

I'll fix it.

cc: @mattklein123

rgs1:
@mattklein123 ok now it's ready to go I think, PTAL -- thanks!

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
```diff
@@ -86,6 +87,12 @@ message RedisProxy {
     // original error is passed downstream unchanged. By default, this support is not enabled.
     bool enable_redirection = 3;
 
+    // If ``enable_redirection`` is set to true this option configures the DNS cache that the
+    // connection pool will use to resolve hostnames that are returned with MOVED and ASK responses.
+    // If no configuration is provided, DNS lookups will not be performed (and thus the MOVED/ASK errors
+    // will be propagated verbatim to the user).
+    common.dynamic_forward_proxy.v3.DnsCacheConfig dns_cache_config = 9;
```
mattklein123, line 94: Can you update the Redis Arch docs to talk about this and maybe have a config example of how to set this up? I would go as far as to say that we should error a config if this is not set since it's broken, but that would be a breaking change for existing users. Can we at least output a warning of what will happen if enable_redirection is set, this is not set, and the server returns MOVED/ASK with host names? I'm mostly foreseeing other people not realizing they have to set this.

changelogs/current.yaml
```diff
@@ -95,5 +95,8 @@ new_features:
 - area: mobile
   change: |
     started merging the Envoy mobile library into the main Envoy repo.
+- area: redis
+  change: |
+    added support for DNS resolving endponts returned by MOVED/ASK responses.
```
mattklein123, line 100: Can you ref link to the new field and also the arch docs that I mention adding above?

source/extensions/filters/network/redis_proxy/conn_pool_impl.cc
```diff
@@ -430,6 +434,66 @@ void InstanceImpl::PendingRequest::onFailure() {
...
+      host->cluster().stats().upstream_internal_redirect_failed_total_.inc();
+      return;
+    }
+    PANIC_DUE_TO_CORRUPT_ENUM;
+  }
+
+  doRedirection(std::move(value), host_address, ask_redirection);
+}
+
+std::string InstanceImpl::PendingRequest::formatAddress(const Envoy::Network::Address::Ip* ip) {
```
mattklein123, line 473: In practice I think this has to be non-null? Can you pass by reference?

source/extensions/filters/network/redis_proxy/conn_pool_impl.cc
```diff
@@ -430,6 +434,66 @@ void InstanceImpl::PendingRequest::onFailure() {
...
+        parent_.dns_cache_->loadDnsCacheEntry(host_address, parent_.default_port_, false, *this);
+
+    cache_load_handle_ = std::move(result.handle_);
+    switch (result.status_) {
+    case Extensions::Common::DynamicForwardProxy::DnsCache::LoadDnsCacheEntryStatus::InCache: {
+      ASSERT(cache_load_handle_ == nullptr);
+      if (!result.host_info_.has_value() || !result.host_info_.value()->address()) {
+        auto host = host_;
+        onResponse(std::move(resp_value_));
+        host->cluster().stats().upstream_internal_redirect_failed_total_.inc();
```
mattklein123, line 450: nit: you might want to add some debug logs for these error cases here and below.

source/extensions/filters/network/redis_proxy/proxy_filter.cc
```diff
@@ -19,12 +19,15 @@ namespace RedisProxy {
...
-    Api::Api& api)
+    Api::Api& api,
+    Extensions::Common::DynamicForwardProxy::DnsCacheManagerFactory& cache_manager_factory)
     : drain_decision_(drain_decision), runtime_(runtime),
       stat_prefix_(fmt::format("redis.{}.", config.stat_prefix())),
       stats_(generateStats(stat_prefix_, scope)),
       downstream_auth_username_(
-          Config::DataSource::read(config.downstream_auth_username(), true, api)) {
+          Config::DataSource::read(config.downstream_auth_username(), true, api)),
+      port_(static_cast<uint16_t>(6379)), dns_cache_manager_(cache_manager_factory.get()),
```
mattklein123, line 29: nit: can you put 6379 in a constant somewhere to make it more clear?

source/extensions/filters/network/redis_proxy/conn_pool_impl.cc
```diff
@@ -83,10 +84,11 @@ InstanceImpl::makeRequestToHost(const std::string& host_address,
...
 }
 
-InstanceImpl::ThreadLocalPool::ThreadLocalPool(std::shared_ptr<InstanceImpl> parent,
-                                               Event::Dispatcher& dispatcher,
-                                               std::string cluster_name)
+InstanceImpl::ThreadLocalPool::ThreadLocalPool(
+    std::shared_ptr<InstanceImpl> parent, Event::Dispatcher& dispatcher, std::string cluster_name,
+    const Extensions::Common::DynamicForwardProxy::DnsCacheSharedPtr& dns_cache)
     : parent_(parent), dispatcher_(dispatcher), cluster_name_(std::move(cluster_name)),
+      dns_cache_(dns_cache), default_port_(static_cast<uint16_t>(6379)),
```
mattklein123, line 91: Same comment about named constant for 6379, also the mem var can be constant I think. And if it's always constant and not configurable do we even need any of these member variables for now?

test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc
```diff
@@ -773,6 +805,20 @@ TEST_P(RedisProxyWithRedirectionIntegrationTest, RedirectToKnownServer) {
   simpleRedirection(fake_upstreams_[1], request, redirection_error.str(), "$3\r\nbar\r\n");
 }
 
+// This test sends a simple Redis command to a sequence of fake upstream
+// Redis servers. The first server replies with a MOVED redirection
+// error that specifies the hostname as its target.
+// The target server responds to a possibly transformed request, and its response
+// is received unchanged by the fake Redis client.
+
```
mattklein123, line 813: nit: remove new line

mattklein123 (changes requested):
Nice, thanks for fixing. Generally LGTM with some small comments.

/wait

api/envoy/extensions/filters/network/redis_proxy/v3/redis_proxy.proto
rgs1: Yeah I'll update the docs. 

I do think we should emit the warning just once (e.g.: when the filter is instantiated), as opposed to an every MOVED/ASK request since that'd be very spammy.

mattklein123: Yeah I meant at config load time once.

rgs1: Ah yes, you did say what _will_ happen. Sorry lack of ☕️.

rgs1: I'll also open an issue to explore enabling the DNS cache by default, once this has been used for a while.

source/extensions/filters/network/redis_proxy/conn_pool_impl.cc
rgs1: Yeah as far as I know Redis can't be used on Unix domain sockets and such.

rgs1: Or at least, typically isn't used... less so for cluster mode.

source/extensions/filters/network/redis_proxy/conn_pool_impl.cc
rgs1: yeah let's drop it as a member var for now and make it a constant somewhere accessible.

mattklein123:
LGTM modulo adding the warning, thanks.

/wait

rgs1:
> LGTM modulo adding the warning, thanks.

Ah yes, just pushed (missed it somehow).

rgs1:
FWIW, tested against memorydb and it's looking good 👍

source/extensions/filters/network/redis_proxy/proxy_filter.cc
```diff
@@ -28,6 +28,12 @@ ProxyFilterConfig::ProxyFilterConfig(
           Config::DataSource::read(config.downstream_auth_username(), true, api)),
       dns_cache_manager_(cache_manager_factory.get()), dns_cache_(getCache(config)) {
 
+  if (config.settings().enable_redirection() && !config.settings().has_dns_cache_config()) {
+    ENVOY_LOG(
+        warn,
+        "redirections enabled without DNS lookups so some responses might generate client errors");
```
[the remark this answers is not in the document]
rgs1: Sure, is something like:

```
redirections without DNS lookups enabled might cause client errors, set the dns_cache_config field within the connection pool settings to avoid them
```

good enough?

mattklein123 (approved):
LGTM, thanks!

PR #23826, opened 2022-11-03
health_check: add the support to disable active hc for single endpoint

botengyao:

This PR is going to add an optional flag in the Endpoint to disable or enable active health check for it. 
**Note**, it will impact all type of clusters if health checker is configured, e.g. EDS, strict_dns. 
However, we skip the endpoint at the message level for HDS.

Commit Message:
Additional Description:
Risk Level: M
Testing: unit and integration tests
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/upstream/upstream_impl.cc
+++ b/source/common/upstream/upstream_impl.cc
@@ -1418,8 +1418,17 @@ void ClusterImplBase::onPreInitComplete() {
 void ClusterImplBase::onInitDone() {
   if (health_checker_ && pending_initialize_health_checks_ == 0) {
     for (auto& host_set : prioritySet().hostSetsPerPriority()) {
-      pending_initialize_health_checks_ += host_set->hosts().size();
+      uint64_t health_check_count = 0;
+      for (auto& host : host_set->hosts()) {
+        if (host->disableActiveHealthCheck()) {
+          continue;
+        }
+        ++health_check_count;
+      }
+      pending_initialize_health_checks_ += health_check_count;
     }
+    ENVOY_LOG(debug, "Clsuter onInitDone pending initialize health checks count {}",
+              pending_initialize_health_checks_);
 
     // TODO(mattklein123): Remove this callback when done.
     health_checker_->addHostCheckCompleteCb([this](HostSharedPtr, HealthTransition) -> void {
@@ -1505,6 +1514,7 @@ void ClusterImplBase::reloadHealthyHostsHelper(const HostSharedPtr&) {
     prioritySet().updateHosts(priority,
                               HostSetImpl::partitionHosts(hosts_copy, hosts_per_locality_copy),
                               host_set->localityWeights(), {}, {}, absl::nullopt);
+    ENVOY_LOG(debug, "Boteng prioritySet.updateHosts done.");
   }
 }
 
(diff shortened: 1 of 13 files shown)


botengyao:
cc @yanavlasov as well. Thanks!

htuch:
What's the use case for this?

botengyao:
Thanks for taking a look! This is mainly to support hybrid NEGs health check. E.g., only health check the non-GCE NEGs. Envoy would consume EDS provided health data for GCE NEGs (coming from UHC) while hybrid NEGs would be health checked by Envoy. This is the interim solution for Mixed NEGs based on [go/statera-active-health-checks](https://goto.google.com/statera-active-health-checks)

htuch (commented):
/lgtm api

botengyao:
Hi @tonya11en, gental ping for the review ;).

source/common/upstream/upstream_impl.cc
```diff
@@ -1418,8 +1418,17 @@ void ClusterImplBase::onPreInitComplete() {
...
     for (auto& host_set : prioritySet().hostSetsPerPriority()) {
-      pending_initialize_health_checks_ += host_set->hosts().size();
+      uint64_t health_check_count = 0;
+      for (auto& host : host_set->hosts()) {
+        if (host->disableActiveHealthCheck()) {
+          continue;
+        }
+        ++health_check_count;
+      }
+      pending_initialize_health_checks_ += health_check_count;
```
tonya11en, line 1428: You could forego creating `health_check_count` and just increment `pending_initialize_health_checks_` directly.

source/common/upstream/upstream_impl.cc
```diff
@@ -1418,8 +1418,17 @@ void ClusterImplBase::onPreInitComplete() {
...
     }
+    ENVOY_LOG(debug, "Clsuter onInitDone pending initialize health checks count {}",
```
tonya11en, line 1430: Misspelled `Cluster`

tonya11en (approved):
Took the first pass and LG overall. Only some minor nits.

Maintainers should take a look at this point.

source/common/upstream/upstream_impl.cc
botengyao: Good point, done.

source/common/upstream/upstream_impl.cc
botengyao: Thanks! Done

test/integration/eds_integration_test.cc
```diff
@@ -61,8 +61,9 @@ class EdsIntegrationTest : public testing::TestWithParam<Network::Address::IpVer
   void setEndpoints(uint32_t total_endpoints, uint32_t healthy_endpoints,
                     uint32_t degraded_endpoints, bool remaining_unhealthy = true,
                     absl::optional<uint32_t> overprovisioning_factor = absl::nullopt,
-                    bool await_update = true) {
+                    bool await_update = true, uint32_t disable_active_hc_endpoints = 0) {
```
htuch, line 64: I think at this point, the number of integer positional parameters has gotten out-of-hand. I would make this an options struct so we can set defaults and clearly see what is being configured.

htuch (commented):
LGTM modulo nit.

test/integration/eds_integration_test.cc
botengyao: Good point & changed to an option struct. Seems designated initializer is a C++20 feature, but we are not allowing to use that.

htuch (approved):
LGTM, thanks!

PR #23828, opened 2022-11-03
rbac: add filter state object option

kyessenov:

Commit Message: Add another option to read principal from the filter state instead of metadata. The use case is passing the value from a network filter to an HTTP filter (RBAC), and the dynamic metadata does not support inheritance. For tunneled requests, the principal needs to be set at the tunnel connection, not the internal connection used for HTTP processing.
Risk Level: low
Testing: unit
Docs Changes: none
Release Notes: none


phlax:
@yanavlasov would you mind giving this a pass, thanks

kyessenov:
@markdroth Can you please take a look at the API change?

source/common/common/matchers.cc
```diff
@@ -96,6 +96,28 @@ MetadataMatcher::MetadataMatcher(const envoy::type::matcher::v3::MetadataMatcher
   value_matcher_ = ValueMatcher::create(v);
 }
 
+FilterStateMatcher::FilterStateMatcher(const envoy::type::matcher::v3::FilterStateMatcher& matcher)
+    : key_(matcher.key()) {
+  switch (matcher.matcher_case()) {
+  case envoy::type::matcher::v3::FilterStateMatcher::MatcherCase::kStringMatch:
```
yanavlasov, line 102: Small suggestion that can improve const correctness. You can move this switch statement into a method in anonymous namespace, that returns StringMatcherPtr and then you can use it to initialize value_matcher_ in the constructor initializer list and make it const. I find it easier to reason that configuration objects are immutable after they were created.

yanavlasov (approved):

source/common/common/matchers.cc
kyessenov: Good point, immutability also helps compilers.

yanavlasov (approved):

kyessenov:
Gentle ping @markdroth or any other @envoyproxy/api-shepherds .

PR #23831, opened 2022-11-04
listener filter: add continueFilterChain for ListenerFilterCallback

soulxu:

Commit Message: listener filter: add continueFilterChain for ListenerFilterCallback
Additional Description:
https://envoyproxy.slack.com/archives/C78HA81DH/p1665661946626769 there is a case showing up, people still need a way to trigger the listener filter iteration in the listener filter.

Risk Level: low
Testing: continueFilterChain is used by the existing flow, it should be tested to work well.
Docs Changes: n/a
Release Notes:  release note should be added


soulxu:
cc @KBaichoo @mattklein123

soulxu:
my first question is whether we should bring the `dispatcher()`, I guess people need the dispatcher to create their own event, then they are available to get a chance to calling back to filter iteration.

soulxu:
> my first question is whether we should bring the `dispatcher()` back, I guess people may have the case of using the dispatcher to create their own event? then they are available to get a chance to call back to filter iteration.

if the people need the dispatcher to register their own event, it seems like we are back to the old day. We probably missed something else in the listener interface.

KBaichoo (commented):
Yea, I think this would need the dispatcher as well in the ListenerFilterCallbacks. We can expand this if the need arises. Thank you

changelogs/current.yaml
```diff
@@ -86,6 +86,9 @@ new_features:
 - area: generic_proxy
   change: |
     added :ref:`generic rds support <envoy_v3_api_field_extensions.filters.network.generic_proxy.v3.GenericProxy.generic_rds>`.
+- area: listener
+  change: |
+    added ``continueFilterChain()`` and ``dispatcher()`` methods to the ``ListenerFilterCallback``, enable the listener filter back to the lister filter iteration after some async process done.
```
KBaichoo, line 91: maybe instead of :`, enable the listener filter back to the lister filter iteration after some async process done.`

something like :`. This allows listener filters to continue listener filter iteration after stopping iteration e.g. if the listener filter depends on an async process.`

KBaichoo (commented):
looks great otherwise, thank you!

changelogs/current.yaml
soulxu: thanks! done

KBaichoo (approved):

PR #23834, opened 2022-11-04
deps: bump liburing -> 2.3

ipuustin:
Commit Message: deps: bump liburing to 2.3.
Additional Description: Update libiouring. It seems that dependabot isn't picking liburing updates, since we missed the 2.2 update.
Risk Level: low
Testing: unit tests
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: only supported on Linux


phlax:
these bazel deps dont get managed by dependabot - we have an internal system that tracks them that can be seen here:

https://github.com/envoyproxy/envoy/issues?q=is%3Aissue+is%3Aopen+newer+release+available

that said it doesnt seem to have picked up the liburing update there either

moderation:
Wonder why this isn't being picked up by the dependabot?

/lgtm deps

phlax:
> Not being picked up by the dependabot because there are no releases, only tags?

yep, most likely - there is some tag handling in the code - but i think there are some situations where it doesnt work - not sure why but i will look next time im in that code

phlax (approved):
lgtm, thanks @ipuustin

PR #23845, opened 2022-11-04
matchers: add formatter action

kyessenov:
Commit Message: Add a formatter action for network filter chain matching.
Additional Description: This feature handles a long tail of matching capabilities by allowing arbitrary format strings to be used as the "action", which is the filter chain name in the connection handler. The use case is using a filter state object or dynamic metadata to pre-compute the filter chain name in a listener filter.

Example:
```yaml
    filter_chain_matcher:
      matcher_tree:
        input:
          name: ip
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.matching.common_inputs.network.v3.DestinationIPInput
        custom_match:
          name: ip-matcher
          typed_config:
            "@type": type.googleapis.com/xds.type.matcher.v3.IPMatcher
            range_matchers:
            - ranges:
              - address_prefix: 127.0.0.1
                prefix_len: 32
              on_match:
                action:
                  name: foo
                  typed_config:
                    "@type": type.googleapis.com/envoy.config.core.v3.SubstitutionFormatString
                    text_format_source:
                      inline_string: "%FILTER_STATE(my_filter.com)%"


```
Risk Level: low
Testing: yes
Docs Changes: yes
Release Notes: yes
Issue: https://github.com/envoyproxy/envoy/issues/23614


kyessenov:
CC @yanjunxiang-google

source/extensions/matching/actions/format_string/config.h
```diff
@@ -0,0 +1,45 @@
...
+#include "source/common/matcher/matcher.h"
+#include "source/server/filter_chain_manager_impl.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace Matching {
+namespace Actions {
+namespace FormatString {
+
+class Action : public Server::FilterChainBaseAction,
```
yanavlasov, line 18: Would it be possible to simplify the inheritance here and avoid the diamond pattern? I think this is what causes you to change all static_casts to dynamic, right? Is it possible to make FilterChainBaseAction a template and derive it from Matcher::ActionBase?

yanavlasov (commented):
/wait-any

source/extensions/matching/actions/format_string/config.h
kyessenov: I've tried that. The problem is this line:
`return result->getTyped<FilterChainBaseAction>().get(filter_chains_by_name_, info);`

We'd need to say something like "\exists proto T . result->getTyped<FilterChainBaseAction<T>>` which I can't figure out how to express. Is there some construct to do that? It's almost like that we do want traits here, not a strict hierarchy.

yanavlasov: The only way to flatten inheritance chain here is to use the "curiously recurring" pattern. The ActionBase class needs to change to support it:

```
template <typename ProtoType, typename Base=Envoy::Matcher::Action> class ActionBase : public Base {
public:
  template <typename... Args>
  ActionBase(Args... args) : Base(args...) {}

  absl::string_view typeUrl() const override { return staticTypeUrl(); }

  static absl::string_view staticTypeUrl() {
    const static std::string typeUrl = ProtoType().GetTypeName();

    return typeUrl;
  }
};
```

The `...::FormatString::Action` just needs to derive from Server::FilterChainBaseAction.

And then you can add the helper alias `...::FormatString::ActionImpl = Matcher::ActionBase<envoy::config::core::v3::SubstitutionFormatString, ...::FormatString::Action>;`

And then in the factory instead of creating object of Action class, create ActionImpl with the same arguments.

yanavlasov (commented):
/wait-any

source/extensions/matching/actions/format_string/config.h
kyessenov: Done.

kyessenov:
/retest
(flake)

PR #23847, opened 2022-11-04
Add cache_file_fixed_block as first piece of filesystem_cache

ravenblackx:
Commit Message: cache_file_fixed_block: part one of filesystem_cache
Additional Description: #23175 is the target end state of this.
Risk Level: None, it is dead-code until the API part is added.
Testing: Unit tested.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
+++ b/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
@@ -25,7 +25,7 @@ constexpr std::array<char, 4> ExpectedCacheVersionId = {'0', '0', '0', '0'};
 
 CacheFileFixedBlock::CacheFileFixedBlock()
     : file_id_(ExpectedFileId), cache_version_id_(ExpectedCacheVersionId), header_size_(0),
-      body_size_(0), trailer_size_(0) {}
+      trailer_size_(0), body_size_(0) {}
 
 void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
   // The string view should be the size of the buffer, and
@@ -38,8 +38,8 @@ void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
   std::copy(s.begin(), s.begin() + 4, file_id_.begin());
   std::copy(s.begin() + 4, s.begin() + 8, cache_version_id_.begin());
   header_size_ = absl::big_endian::Load32(&s[8]);
-  body_size_ = absl::big_endian::Load64(&s[12]);
-  trailer_size_ = absl::big_endian::Load32(&s[20]);
+  trailer_size_ = absl::big_endian::Load32(&s[12]);
+  body_size_ = absl::big_endian::Load64(&s[16]);
 }
 
 void CacheFileFixedBlock::serializeToBuffer(Buffer::Instance& buffer) {
@@ -53,8 +53,8 @@ void CacheFileFixedBlock::serializeToBuffer(Buffer::Instance& buffer) {
   std::copy(file_id_.begin(), file_id_.end(), &b[0]);
   std::copy(cache_version_id_.begin(), cache_version_id_.end(), &b[4]);
   absl::big_endian::Store32(&b[8], header_size_);
-  absl::big_endian::Store64(&b[12], body_size_);
-  absl::big_endian::Store32(&b[20], trailer_size_);
+  absl::big_endian::Store32(&b[12], trailer_size_);
+  absl::big_endian::Store64(&b[16], body_size_);
   // Append that buffer into the target buffer object.
   buffer.add(absl::string_view{b, size()});
 }
--- a/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
+++ b/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
@@ -13,13 +13,13 @@ namespace {
 // The expected first four bytes of the header - if fileId() doesn't match ExpectedFileId
 // then the file is not a cache file and should be removed from the cache.
 // Beginning of file should be "CACH".
-std::array<char, 4> ExpectedFileId = {'C', 'A', 'C', 'H'};
+constexpr std::array<char, 4> ExpectedFileId = {'C', 'A', 'C', 'H'};
 
 // The expected next four bytes of the header - if cacheVersionId() doesn't match
 // ExpectedCacheVersionId then the file is from an incompatible cache version and should
 // be removed from the cache.
 // Next 4 bytes of file should be "0000".
-std::array<char, 4> ExpectedCacheVersionId = {'0', '0', '0', '0'};
+constexpr std::array<char, 4> ExpectedCacheVersionId = {'0', '0', '0', '0'};
 
 // Deserialize 4 bytes into a uint32_t.
 uint32_t deserializeUint32(const char* p) {
@@ -63,8 +63,8 @@ CacheFileFixedBlock::CacheFileFixedBlock() {
 void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
   ASSERT(s.size() == size() && size() == 24);
   // Serialize the values from the string_view s into the member values.
-  std::copy_n(s.begin(), 4, file_id_.begin());
-  std::copy_n(s.begin() + 4, 4, cache_version_id_.begin());
+  std::copy(s.begin(), s.begin() + 4, file_id_.begin());
+  std::copy(s.begin() + 4, s.begin() + 8, cache_version_id_.begin());
   header_size_ = deserializeUint32(&s[8]);
   body_size_ = deserializeUint64(&s[12]);
   trailer_size_ = deserializeUint32(&s[20]);
@@ -74,8 +74,8 @@ void CacheFileFixedBlock::serializeToBuffer(Buffer::Instance& buffer) {
   char b[size()];
   ASSERT(size() == 24);
   // Serialize the values from the member values into the stack buffer b.
-  std::copy_n(file_id_.begin(), 4, &b[0]);
-  std::copy_n(cache_version_id_.begin(), 4, &b[4]);
+  std::copy(file_id_.begin(), file_id_.end(), &b[0]);
+  std::copy(cache_version_id_.begin(), cache_version_id_.end(), &b[4]);
   serializeUint32(header_size_, &b[8]);
   serializeUint64(body_size_, &b[12]);
   serializeUint32(trailer_size_, &b[20]);
--- a/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
+++ b/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
@@ -13,22 +13,18 @@ namespace {
 // The expected first four bytes of the header - if fileId() doesn't match ExpectedFileId
 // then the file is not a cache file and should be removed from the cache.
 // Beginning of file should be "CACH".
-constexpr uint32_t ExpectedFileId = (static_cast<uint32_t>('C') << 24) |
-                                    (static_cast<uint32_t>('A') << 16) |
-                                    (static_cast<uint32_t>('C') << 8) | static_cast<uint32_t>('H');
+std::array<char, 4> ExpectedFileId = {'C', 'A', 'C', 'H'};
 
 // The expected next four bytes of the header - if cacheVersionId() doesn't match
 // ExpectedCacheVersionId then the file is from an incompatible cache version and should
 // be removed from the cache.
 // Next 4 bytes of file should be "0000".
-// Increment this to invalidate old cache files if the format changes.
-// Formatted string-style rather than as an actual int to make it easily human-readable.
-constexpr uint32_t ExpectedCacheVersionId =
-    (static_cast<uint32_t>('0') << 24) | (static_cast<uint32_t>('0') << 16) |
-    (static_cast<uint32_t>('0') << 8) | static_cast<uint32_t>('0');
+std::array<char, 4> ExpectedCacheVersionId = {'0', '0', '0', '0'};
 
 // Deserialize 4 bytes into a uint32_t.
 uint32_t deserializeUint32(const char* p) {
+  // chars must be cast to uint8_t first so that e.g. -1 becomes 0xff, before casting to uint32_t
+  // - otherwise -1 becomes 0xffffffff.
   return static_cast<uint32_t>(static_cast<uint8_t>(p[0])) << 24 |
          static_cast<uint32_t>(static_cast<uint8_t>(p[1])) << 16 |
          static_cast<uint32_t>(static_cast<uint8_t>(p[2])) << 8 |
@@ -66,13 +62,9 @@ CacheFileFixedBlock::CacheFileFixedBlock() {
 
 void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
   ASSERT(s.size() == size() && size() == 24);
-  // Bytewise copy the values from the string_view into the fields.
-  // This ensures consistent endianness, and is the only safe way to copy values
-  // from a byte array into value types.
-  // chars must be cast to uint8_t first so that e.g. -1 becomes 0xff, before casting to uint64_t
-  // - otherwise -1 becomes 0xffffffffffffffff.
-  file_id_ = deserializeUint32(&s[0]);
-  cache_version_id_ = deserializeUint32(&s[4]);
+  // Bytewise serialize the values from the string_view s into the member values.
+  std::copy_n(s.begin(), 4, file_id_.begin());
+  std::copy_n(s.begin() + 4, 4, cache_version_id_.begin());
   header_size_ = deserializeUint32(&s[8]);
   body_size_ = deserializeUint64(&s[12]);
   trailer_size_ = deserializeUint32(&s[20]);
@@ -81,11 +73,13 @@ void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
 void CacheFileFixedBlock::serializeToBuffer(Buffer::Instance& buffer) {
   char b[size()];
   ASSERT(size() == 24);
-  serializeUint32(file_id_, &b[0]);
-  serializeUint32(cache_version_id_, &b[4]);
+  // Bytewise serialize the values from the member values into the stack buffer b.
+  std::copy_n(file_id_.begin(), 4, &b[0]);
+  std::copy_n(cache_version_id_.begin(), 4, &b[4]);
   serializeUint32(header_size_, &b[8]);
   serializeUint64(body_size_, &b[12]);
   serializeUint32(trailer_size_, &b[20]);
+  // Append that buffer into the target buffer object.
   buffer.add(absl::string_view{b, size()});
 }
 
--- a/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
+++ b/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
@@ -21,35 +21,6 @@ constexpr std::array<char, 4> ExpectedFileId = {'C', 'A', 'C', 'H'};
 // Next 4 bytes of file should be "0000".
 constexpr std::array<char, 4> ExpectedCacheVersionId = {'0', '0', '0', '0'};
 
-// Deserialize 4 bytes into a uint32_t.
-uint32_t deserializeUint32(const char* p) {
-  // chars must be cast to uint8_t first so that e.g. -1 becomes 0xff, before casting to uint32_t
-  // - otherwise -1 becomes 0xffffffff.
-  return static_cast<uint32_t>(static_cast<uint8_t>(p[0])) << 24 |
-         static_cast<uint32_t>(static_cast<uint8_t>(p[1])) << 16 |
-         static_cast<uint32_t>(static_cast<uint8_t>(p[2])) << 8 |
-         static_cast<uint32_t>(static_cast<uint8_t>(p[3]));
-}
-
-// Deserialize 8 bytes into a uint64_t.
-uint64_t deserializeUint64(const char* p) {
-  return static_cast<uint64_t>(deserializeUint32(p)) | deserializeUint32(p + 4);
-}
-
-// Serialize a uint32 into 4 bytes at pointer p.
-void serializeUint32(uint32_t value, char* p) {
-  p[0] = value >> 24;
-  p[1] = value >> 16 & 0xff;
-  p[2] = value >> 8 & 0xff;
-  p[3] = value & 0xff;
-}
-
-// Serialize a uint64 into 8 bytes at pointer p.
-void serializeUint64(uint64_t value, char* p) {
-  serializeUint32(value >> 32, p);
-  serializeUint32(value & 0xffffffff, p + 4);
-}
-
 } // namespace
 
 CacheFileFixedBlock::CacheFileFixedBlock() {
@@ -61,24 +32,33 @@ CacheFileFixedBlock::CacheFileFixedBlock() {
 }
 
 void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
+  // The string view should be the size of the buffer, and
+  // Since we're explicitly reading byte offsets here, the size should match
+  // what we read.
+  // (This will remind us to change this function if we change the size
+  // and vice-versa!)
   ASSERT(s.size() == size() && size() == 24);
   // Serialize the values from the string_view s into the member values.
   std::copy(s.begin(), s.begin() + 4, file_id_.begin());
   std::copy(s.begin() + 4, s.begin() + 8, cache_version_id_.begin());
-  header_size_ = deserializeUint32(&s[8]);
-  body_size_ = deserializeUint64(&s[12]);
-  trailer_size_ = deserializeUint32(&s[20]);
+  header_size_ = absl::big_endian::Load32(&s[8]);
+  body_size_ = absl::big_endian::Load64(&s[12]);
+  trailer_size_ = absl::big_endian::Load32(&s[20]);
 }
 
 void CacheFileFixedBlock::serializeToBuffer(Buffer::Instance& buffer) {
   char b[size()];
+  // Since we're explicitly writing byte offsets here, the size should match
+  // what we write.
+  // (This will remind us to change this function if we change the size
+  // and vice-versa!)
   ASSERT(size() == 24);
   // Serialize the values from the member values into the stack buffer b.
   std::copy(file_id_.begin(), file_id_.end(), &b[0]);
   std::copy(cache_version_id_.begin(), cache_version_id_.end(), &b[4]);
-  serializeUint32(header_size_, &b[8]);
-  serializeUint64(body_size_, &b[12]);
-  serializeUint32(trailer_size_, &b[20]);
+  absl::big_endian::Store32(&b[8], header_size_);
+  absl::big_endian::Store64(&b[12], body_size_);
+  absl::big_endian::Store32(&b[20], trailer_size_);
   // Append that buffer into the target buffer object.
   buffer.add(absl::string_view{b, size()});
 }
--- a/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
+++ b/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
@@ -27,6 +27,33 @@ constexpr uint32_t ExpectedCacheVersionId =
     (static_cast<uint32_t>('0') << 24) | (static_cast<uint32_t>('0') << 16) |
     (static_cast<uint32_t>('0') << 8) | static_cast<uint32_t>('0');
 
+// Deserialize 4 bytes into a uint32_t.
+uint32_t deserializeUint32(const char* p) {
+  return static_cast<uint32_t>(static_cast<uint8_t>(p[0])) << 24 |
+         static_cast<uint32_t>(static_cast<uint8_t>(p[1])) << 16 |
+         static_cast<uint32_t>(static_cast<uint8_t>(p[2])) << 8 |
+         static_cast<uint32_t>(static_cast<uint8_t>(p[3]));
+}
+
+// Deserialize 8 bytes into a uint64_t.
+uint64_t deserializeUint64(const char* p) {
+  return static_cast<uint64_t>(deserializeUint32(p)) | deserializeUint32(p + 4);
+}
+
+// Serialize a uint32 into 4 bytes at pointer p.
+void serializeUint32(uint32_t value, char* p) {
+  p[0] = value >> 24;
+  p[1] = value >> 16 & 0xff;
+  p[2] = value >> 8 & 0xff;
+  p[3] = value & 0xff;
+}
+
+// Serialize a uint64 into 8 bytes at pointer p.
+void serializeUint64(uint64_t value, char* p) {
+  serializeUint32(value >> 32, p);
+  serializeUint32(value & 0xffffffff, p + 4);
+}
+
 } // namespace
 
 CacheFileFixedBlock::CacheFileFixedBlock() {
@@ -44,59 +71,21 @@ void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
   // from a byte array into value types.
   // chars must be cast to uint8_t first so that e.g. -1 becomes 0xff, before casting to uint64_t
   // - otherwise -1 becomes 0xffffffffffffffff.
-  file_id_ = static_cast<uint32_t>(static_cast<uint8_t>(s[0])) << 24 |
-             static_cast<uint32_t>(static_cast<uint8_t>(s[1])) << 16 |
-             static_cast<uint32_t>(static_cast<uint8_t>(s[2])) << 8 |
-             static_cast<uint32_t>(static_cast<uint8_t>(s[3]));
-  cache_version_id_ = static_cast<uint32_t>(static_cast<uint8_t>(s[4])) << 24 |
-                      static_cast<uint32_t>(static_cast<uint8_t>(s[5])) << 16 |
-                      static_cast<uint32_t>(static_cast<uint8_t>(s[6])) << 8 |
-                      static_cast<uint32_t>(static_cast<uint8_t>(s[7]));
-  header_size_ = static_cast<uint32_t>(static_cast<uint8_t>(s[8])) << 24 |
-                 static_cast<uint32_t>(static_cast<uint8_t>(s[9])) << 16 |
-                 static_cast<uint32_t>(static_cast<uint8_t>(s[10])) << 8 |
-                 static_cast<uint32_t>(static_cast<uint8_t>(s[11]));
-  body_size_ = static_cast<uint64_t>(static_cast<uint8_t>(s[12])) << 56 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[13])) << 48 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[14])) << 40 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[15])) << 32 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[16])) << 24 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[17])) << 16 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[18])) << 8 |
-               static_cast<uint64_t>(static_cast<uint8_t>(s[19]));
-  trailer_size_ = static_cast<uint32_t>(static_cast<uint8_t>(s[20])) << 24 |
-                  static_cast<uint32_t>(static_cast<uint8_t>(s[21])) << 16 |
-                  static_cast<uint32_t>(static_cast<uint8_t>(s[22])) << 8 |
-                  static_cast<uint32_t>(static_cast<uint8_t>(s[23]));
+  file_id_ = deserializeUint32(&s[0]);
+  cache_version_id_ = deserializeUint32(&s[4]);
+  header_size_ = deserializeUint32(&s[8]);
+  body_size_ = deserializeUint64(&s[12]);
+  trailer_size_ = deserializeUint32(&s[20]);
 }
 
 void CacheFileFixedBlock::serializeToBuffer(Buffer::Instance& buffer) {
   char b[size()];
   ASSERT(size() == 24);
-  b[0] = file_id_ >> 24;
-  b[1] = file_id_ >> 16 & 0xff;
-  b[2] = file_id_ >> 8 & 0xff;
-  b[3] = file_id_ & 0xff;
-  b[4] = cache_version_id_ >> 24;
-  b[5] = cache_version_id_ >> 16 & 0xff;
-  b[6] = cache_version_id_ >> 8 & 0xff;
-  b[7] = cache_version_id_ & 0xff;
-  b[8] = header_size_ >> 24;
-  b[9] = header_size_ >> 16 & 0xff;
-  b[10] = header_size_ >> 8 & 0xff;
-  b[11] = header_size_ & 0xff;
-  b[12] = body_size_ >> 56;
-  b[13] = body_size_ >> 48 & 0xff;
-  b[14] = body_size_ >> 40 & 0xff;
-  b[15] = body_size_ >> 32 & 0xff;
-  b[16] = body_size_ >> 24 & 0xff;
-  b[17] = body_size_ >> 16 & 0xff;
-  b[18] = body_size_ >> 8 & 0xff;
-  b[19] = body_size_ & 0xff;
-  b[20] = trailer_size_ >> 24;
-  b[21] = trailer_size_ >> 16 & 0xff;
-  b[22] = trailer_size_ >> 8 & 0xff;
-  b[23] = trailer_size_ & 0xff;
+  serializeUint32(file_id_, &b[0]);
+  serializeUint32(cache_version_id_, &b[4]);
+  serializeUint32(header_size_, &b[8]);
+  serializeUint64(body_size_, &b[12]);
+  serializeUint32(trailer_size_, &b[20]);
   buffer.add(absl::string_view{b, size()});
 }
 
--- a/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
+++ b/source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
@@ -1,0 +1,82 @@
+#include "source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h"
+
+#include "source/common/common/assert.h"
+#include "source/common/common/safe_memcpy.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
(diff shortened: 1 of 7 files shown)


alyssawilk:
@toddmgreer would you be willing to do first pass?

ravenblackx:
fyi @jmarantz has the most context from looking at the larger PR already.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,74 @@
...
+#include "absl/strings/string_view.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+// This represents a minimal header block on the cache entry file; it uses a struct that
+// is required to be packed, and explicit network byte order, to ensure compatibility.
```
jmarantz, line 17: it's not clear to me that's necessary. I wouldn't expect to share a cache data file between machines with different endian-ness. Instead you could write a byte at the top to make sure a cache file was not read by a machine with the wrong architecture.

In general I expect these files to be accessed locallly; they would not perform well IMO if you access them via a network mount.

jmarantz (commented):
Thanks for breaking this out.

I have a number of questions that would've been daunting to add to a much larger review.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,82 @@
...
+  setTrailersSize(0);
+  setBodySize(0);
+}
+
+void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
+  ASSERT(s.size() == size());
+  safeMemcpyUnsafeSrc(&contents_, s.begin());
+}
+
+uint32_t CacheFileFixedBlock::getUint32(const uint32_t& t) const { return ntohl32(t); }
```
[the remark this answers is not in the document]
ravenblackx: Removed in favor of inlining.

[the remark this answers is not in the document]
ravenblackx: Removed in favor of inlining.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,74 @@
...
+      uint64_t body_size_;
+    };
+    char as_str_[1];
+  } contents_;
+  static_assert(sizeof(contents_) ==
+                    sizeof(contents_.file_id_) + sizeof(contents_.cache_version_id_) +
+                        sizeof(contents_.header_size_) + sizeof(contents_.trailer_size_) +
+                        sizeof(contents_.body_size_),
+                "contents_ must be fully packed for consistency");
+  uint32_t getUint32(const uint32_t& t) const;
```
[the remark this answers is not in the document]
ravenblackx: Removed.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Changed compatibility to consistency. You're right that it's not likely to be necessary in any common case, but it's generally a good practice to have file formats be consistent (in case someone sends us one saying "why isn't this working", for example.)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,82 @@
...
+#include "source/common/common/safe_memcpy.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+namespace {
+uint64_t ntohl64(uint64_t n) {
```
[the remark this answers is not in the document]
ravenblackx: Found there was one already there, so this was a good idea. 😁

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,74 @@
...
+namespace FileSystemHttpCache {
+
+// This represents a minimal header block on the cache entry file; it uses a struct that
+// is required to be packed, and explicit network byte order, to ensure compatibility.
+class CacheFileFixedBlock {
+public:
+  CacheFileFixedBlock();
+  void populateFromStringView(absl::string_view s);
+  static size_t size() { return sizeof(contents_); }
+  // fileId is a fixed value used to identify that this is a cache file.
```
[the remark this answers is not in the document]
ravenblackx: I'm not really familiar with what that means, but a cursory search for what it means, compared against the comments in e.g. various files in source/common, suggests that it is not an envoy standard to use doxygen format. (I would hope that if it was something we should be doing then a linter should be asking me to do it?)

In the absence of a mandate, I personally prefer *not* doxygen format (assuming the thing you're asking for is e.g. @param annotations, which, again, I'm not really sure even is what you're asking for!)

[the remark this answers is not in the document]
jmarantz: Mandate is here: https://github.com/envoyproxy/envoy/blob/main/STYLE.md
> API-level comments should follow normal Doxygen conventions. Use @param to describe parameters and @return <return-type> for return values. Internal comments for methods and member variables may be regular C++ // comments or Doxygen at developer discretion. Where possible, methods should have meaningful documentation on expected input and state preconditions.

It's unclear to me what API-level means; maybe this is below that level of visibility. I was just going by that syntax generally. But we definitely use doxygen format in envoy code a lot.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,121 @@
...
+  static size_t size() { return sizeof(contents_); }
+
+  // fileId is a fixed value used to identify that this is a cache file.
+  uint32_t fileId() const { return fromEndianness<ByteOrder::BigEndian>(contents_.file_id_); }
+
+  // cacheVersionId is a value that should be consistent between versions of the file
+  // cache implementation. Changing version in code will invalidate all cache
+  // entries where the version ID does not match.
+  uint32_t cacheVersionId() const {
+    return fromEndianness<ByteOrder::BigEndian>(contents_.cache_version_id_);
```
jmarantz, line 37: this feels heavy -- why don't we transform the data when reading/writing it from files so it's always in native format in-memory?

If nothing else it makes debugging (with gdb) annoying to have the data mangled, especially if you store big-endian.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,121 @@
...
+
+  union {
+    struct {
+      uint32_t file_id_;
+      uint32_t cache_version_id_;
+      uint32_t header_size_;
+      uint32_t trailer_size_;
+      uint64_t body_size_;
+    };
+    char as_str_[1];
```
jmarantz, line 106: suggest renaming to 'data_'. I see what you are trying to do here but think calling it 'data 'is more helpful than "as string".

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
```diff
@@ -0,0 +1,96 @@
...
+  EXPECT_TRUE(default_block.isValid());
+}
+
+TEST_F(CacheFileFixedBlockTest, GettersRecoverValuesThatWereSet) {
+  CacheFileFixedBlock block;
+  setFileId(block, 98765);
+  setCacheVersionId(block, 56789);
+  block.setBodySize(999999);
+  block.setHeadersSize(1234);
+  block.setTrailersSize(4321);
```
jmarantz, line 31: I guess this makes sense to test this way as your getters do real work.

My suggestion would be to instead do the work on read/write from files and make the getter trivial (expressed in impl).

But in either case I think you could probably make a second CacheFileFixedBlock and populate it by reading from it 'block' and writing to the new one, and *then* checking the getters.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
```diff
@@ -0,0 +1,96 @@
...
+  EXPECT_EQ(block.cacheVersionId(), 56789);
+  EXPECT_EQ(block.bodySize(), 999999);
+  EXPECT_EQ(block.headerSize(), 1234);
+  EXPECT_EQ(block.trailerSize(), 4321);
+}
+
+TEST_F(CacheFileFixedBlockTest, IsValidReturnsFalseOnBadFileId) {
+  CacheFileFixedBlock block;
+  setFileId(block, 98765);
+  EXPECT_FALSE(block.isValid());
```
jmarantz, line 42: what is that makes this block invalid? Just not having all the fields set?

Maybe call this test UnsetFieldXXXIsInvalid, for XXXX with each individual field that needs to be set?

ravenblackx: ... The test is named "IsValidReturnsFalseOnBadFileId"? [Most] unset fields are valid.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
```diff
@@ -0,0 +1,96 @@
...
+  EXPECT_EQ(block2.offsetToHeaders(), CacheFileFixedBlock::size());
+  EXPECT_EQ(block2.offsetToBody(), CacheFileFixedBlock::size() + 100);
+  EXPECT_EQ(block2.offsetToTrailers(), CacheFileFixedBlock::size() + 1100);
+  EXPECT_TRUE(block2.isValid());
+}
+
+TEST_F(CacheFileFixedBlockTest, NumbersAreInNetworkByteOrder) {
+  CacheFileFixedBlock block;
+  block.setHeadersSize(0xfaaf);
+  EXPECT_EQ(static_cast<unsigned char>(block.stringView()[8]), 0);
```
jmarantz, line 85: nit: I'd capture the string_view, check its size, to make this less reptitious, like:

```
absl::string_view str = block.stringView();
ASSERT_GE(str.size(), 12);
EXPECT_EQ(str[9], 0);
....
```
Up to you on capturing the temp -- I think it reads better and would definitely be easier to debug. But it's best practice to ASSERT sizes of arrays before using them in tests.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
jmarantz: My misunderstanding. To clarify for the reader: a CacheFileFixedBlock with unset fields is valid, but the FileId of 98765 is invalid for some reason? Maybe add a comment?

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,121 @@
...
+
+  // sets the cacheVersionId, a value that should be consistent between versions
+  // of the file cache implementation.
+  // This is private so only the unit test can call it, as the version ID should be
+  // a compile-time constant in ordinary use.
+  void setCacheVersionId(uint32_t id) {
+    contents_.cache_version_id_ = toEndianness<ByteOrder::BigEndian>(id);
+  }
+
+  union {
```
jmarantz, line 98: You probably need to defend the position that you need to manually serialize/de-serialize all these fields including network byte-ordering hacks, rather than simply using protobuf serialization. Please put the rationale in a comment.

Actually I'm not clear myself on why you went this way. I was caught up mainly in the details of *how* you did this but I'm finally catching up on *why* :) 

To be clear: I'm not necessarily objecting to the design decision but it does need to be documented.

jmarantz (commented):
/wait

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: It was documented in the proto file where the whole structure of the file is explained rather than the header alone, but okay.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
ravenblackx: Done (and for cacheVersionId).

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
ravenblackx: Doing the work on read/write would mean allocating a new buffer into which to write the byte-reordered values. The whole point of this object is to be that serialized buffer - creating a separate buffer for serialize/deserialize wouldn't really make sense.

There is a test case that validates values (via offsets) after a serialize/deserialize, but having that test and not this test would not reveal a misbehavior in which, say, 'set' flips the byte order, one of serialize or deserialize flips the byte order, and 'get' does not. Which is the sort of misbehavior you might get if someone decided doing the work on read/write was better and then made that change poorly!

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Disagree: data_ doesn't really mean anything; all the other values in the union are also data. The purpose of the union with as_str_ is to provide a string-like interface. Calling it data wouldn't even hint at what its purpose is.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: This object is essentially the serialized buffer. Transforming on read/write would mean allocating another buffer.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,121 @@
...
+  CacheFileFixedBlock();
+
+  // Deserializes the string representation of a CacheFileFixedBlock into this instance.
+  void populateFromStringView(absl::string_view s);
+
+  // Returns the size in bytes of a CacheFileFixedBlock.
+  static size_t size() { return sizeof(contents_); }
+
+  // fileId is a fixed value used to identify that this is a cache file.
+  uint32_t fileId() const { return fromEndianness<ByteOrder::BigEndian>(contents_.file_id_); }
```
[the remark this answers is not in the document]
ravenblackx: Sure, done. Slightly weird because typically network byte order is the "consistency" one, but if we care about a single instruction of performance in the context of large disk operations then the optimization is worthwhile.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
jmarantz: I was trying to stay analogous to std::string::data() and absl::string_view::data(), both of which can only be used in the context of a size(), and do not null-terminate, which is why I might expect for something that claims to be a "string", but provides only a char*.

The pattern you give looks a lot more like a structure with variable-length array at the end, which at first is what I thought you were doing until I read more closely.

This whole block of code is correct -- not disputing that -- but it took me a while to understand why you had done it this way and why it's safe. This is why my higher level thought is to drop all this in favor of a s protobuf with 4 integer members, and use protobuf utils to serialize/deserialize. When people trying to ensure that Envoy is secure see code that hacks bytes like this it raises a lot of questions about security and correctness that are easier to reason about if we work at a higher level. I will not be the last one to read this code and need to think deeply about it.

As I said there are times when you need to go low-level but I think our default position is that higher level constructs are used unless we document why we are not doing that, in the place where we have the low-level byte-hacking code.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
jmarantz: understood -- I think it's better to answer the question fo why you need to be hacking network byte-order in your getter as opposed to in your serialzier/deserializer.

Let's resolve that first.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Explanation for why not protobuf is now in the class comment.
The difference between this and those string classes is that in the string classes "data" is the sole underlying representation of all the data the class wraps. In this case, the struct of variables is *also* the underlying representation of all the data, so the word data could mean that, or it could mean the raw data (as it would), which makes its lack of meaning a problem. Renamed to raw_, to clarify that it's not a str, and gave it an explicit size to clarify that it's not variable-length.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
ravenblackx: Because changing the byte order in the serialize/deserialize operation would mean you need *another* buffer that you're serializing or deserializing into. The whole point of this class is to represent a serialized buffer.
(For 'read' it's not important because it's already reading from an existing buffer, but for 'write' you'd need the caller to provide a memcpy-recipient pointer pointing at a sufficiently sized block of data, which doesn't play well with Envoy::Buffer or safety, OR provide an Envoy::Buffer& which doesn't play well with being able to do anything else.)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -171,7 +171,7 @@ class CacheFileFixedBlock {
       uint32_t trailer_size_;
       uint64_t body_size_;
     };
-    char as_str_[1];
+    char raw_[sizeof(uint32_t) * 4 + sizeof(uint64_t)];
```
jmarantz, line 174: `raw_` member var name SGTM.

name the struct and use sizeof on the structure.

The point is moot if we decide to drop the manual serialization and just use protobufs for this.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,189 @@
...
+      uint32_t file_id_;
+      uint32_t cache_version_id_;
+      uint32_t header_size_;
+      uint32_t trailer_size_;
+      uint64_t body_size_;
+    };
+    char raw_[sizeof(uint32_t) * 4 + sizeof(uint64_t)];
+  } contents_;
+  static_assert(sizeof(contents_) ==
+                    sizeof(contents_.file_id_) + sizeof(contents_.cache_version_id_) +
```
jmarantz, line 177: this would be much simpler and future-proof if you named the struct.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
jmarantz: This makes sense, but comments in the code would help. I'd like another opinion about it, as I think it's a bit unusual to do it this way. I'll pull someone in once there are comments explaining this decision in the code.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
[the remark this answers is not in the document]
jmarantz: I have never loved every aspect of a style guide that someone else wrote :)

But I think it's a lot less cognitive load on the reader to be consistent with the existing style, so I just do it :)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Removed as part of serializing differently.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * CacheFileFixedBlock represents a minimal header block on the cache entry file; it
+ * uses a struct that is required to be packed, and explicit byte order, to ensure
+ * consistency.
+ *
+ * This data is serialized as a flat object rather than protobuf serialization because it
+ * needs to be readable from the start of the file, but writeable after the rest of the
+ * file has been completely written (as the body size and trailer size aren't necessarily
+ * known until the entire content has been streamed). Serialized proto messages can
+ * change size when values change, which makes them unsuited for this purpose.
```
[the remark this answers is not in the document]
ravenblackx: Yes, proto integer types do something similar to utf8, such that small integers with large types can still serialize to one byte. (It makes higher level testing with mocks unnecessarily difficult due to the microseconds field of a proto.Timestamp potentially changing the header_size_ value!)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Apparently an anonymous struct can't have a named type. Deanonymizing the struct makes all the code horrible. So I just moved the serialization entirely since you were pushing for that anyway.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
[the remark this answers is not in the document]
jmarantz: makes sense. thanks.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -38,8 +38,66 @@ CacheFileFixedBlock::CacheFileFixedBlock() {
...
 void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
-  ASSERT(s.size() == size());
-  safeMemcpyUnsafeSrc(&contents_, s.begin());
+  ASSERT(s.size() == size() && size() == 24);
+  // Bytewise copy the values from the string_view into the fields.
+  // This ensures consistent endianness, and is the only safe way to copy values
+  // from a byte array into value types.
+  // chars must be cast to uint8_t first so that e.g. -1 becomes 0xff, before casting to uint64_t
+  // - otherwise -1 becomes 0xffffffffffffffff.
+  file_id_ = static_cast<uint32_t>(static_cast<uint8_t>(s[0])) << 24 |
```
jmarantz, line 47: this was my punishment for being too picky :)

I think the previous state was obviously better from a readability perspective.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -159,25 +144,13 @@ class CacheFileFixedBlock {
    * a compile-time constant in ordinary use.
    * @param id The cache version ID to set.
    */
-  void setCacheVersionId(uint32_t id) {
-    contents_.cache_version_id_ = toEndianness<ByteOrder::LittleEndian>(id);
-  }
-
-  union {
-    struct {
```
jmarantz: what horrible thing is it that happens if the struct is named? Just having to inject "file_block_." or whatever in references to the fields?

BTW I'm not against using `char raw_[1];` rather than specifying the size as long a you document what magic you are up to with a comment.

I retract my comment about the static-assert now that I see what it's for. That seems fine - sorry for the trouble.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
jmarantz: Also I was thinking you could still read into the struct but just correct the endian-ness (if needed) when you do that.

For writing I guess you could either copy the FileBlock out into a temp and twiddle the bits (if you need to) before appending to the Envoy buffer. Compared with the cost of fileI/O that extra buffer copy seems pretty miniscule.

One more thing: do you want to make size be 64 bits?

ravenblackx: Well, the struct was relatively problematic anyway (it's *borderline* undefined behavior to union bytes over a bunch of ints, especially with the dangers of packing). Having made this change I'm reluctant to change it back. I have made the serialize-on-demand more readable with some inline functions for it. I think the case for making these functions 'common' is weaker than it was for endian-swap functions, but I'm not sure it's something nobody else would want (I wasn't able to find any existing impl this time though!)

I'm more comfortable with body size as 64 bits because, yeah, maybe someone *really does* want to put a full length HD movie into a cache! I don't want to be accidentally creating the next FAT32 situation. (I'm even a little nervous about 32-bit headers and trailers sizes - sure *today* they're generally smaller than 1K, but who knows what madness the future holds!)

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
ravenblackx: Hideousness now reduced with helper functions.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: (There is [this](https://sourcegraph.com/github.com/envoyproxy/envoy/-/blob/source/common/grpc/codec.cc?L19:43) suggesting maybe there are other uses for serializeUint32, but that's the only existing one, in only one direction, and it's not going onto chars, so I think it's more likely that each use-case is sufficiently unique that it's not worth putting in common.)

jmarantz: I am pretty slammed today (prepping for an exec review), so I may not have time for another round, but please explain where the UB is?

I think the main thing in my mind is that to avoid  a spurious uninitialized read you should memset the structure before populating it, depending on how you populate it.

But apart from that I don't see the UB.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,94 @@
...
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+namespace {
+// The expected first four bytes of the header - if fileId() doesn't match ExpectedFileId
+// then the file is not a cache file and should be removed from the cache.
+// Beginning of file should be "CACH".
+std::array<char, 4> ExpectedFileId = {'C', 'A', 'C', 'H'};
```
jmarantz, line 16: these need to be constexpr to avoid an init-order fiasco per style guide. But they don't need to be std::array; can't they be
```
constexpr absl::string_view ExpectedFileId("CACH"}; 
constexpr absl::string_view ExpectedCacheVersionId{"\0\0\0\0", 4|;

File: source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc [hunk 19]
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// Deserialize 8 bytes into a uint64_t.
+uint64_t deserializeUint64(const char* p) {
+  return static_cast<uint64_t>(deserializeUint32(p)) | deserializeUint32(p + 4);
+}
+
+// Serialize a uint32 into 4 bytes at pointer p.
+void serializeUint32(uint32_t value, char* p) {
+  p[0] = value >> 24;
```
jmarantz, line 41: I guess if you're going to go all C on me you could write `*p++ = value >> 24` :)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,94 @@
...
+  setCacheVersionId(ExpectedCacheVersionId);
+  setHeadersSize(0);
+  setTrailersSize(0);
+  setBodySize(0);
+}
+
+void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
+  ASSERT(s.size() == size() && size() == 24);
+  // Serialize the values from the string_view s into the member values.
+  std::copy_n(s.begin(), 4, file_id_.begin());
```
jmarantz, line 66: so much byte hacking. I really didn't mind the old code with the union and the calling of existing byte-order hacking methods; I just wanted it to be commented better.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
jmarantz: I'm trying to understand the practical implication of what you are saying. The downside of using the struct is that there might be some platform in the future where two uint32_t are each padded to 8-byte boundaries, and you want to be able to share cache files between that platform and an arm or x86. Is that right?

I question the premise of ever sharing these files cross-platform in the first place, but if it's not too expensive to support I can't complain if it's possible. E.g. storing the data as little-endian and calling an existing conversion to the native type (which would be a no-op on x86 and arm) seems pretty reasonable :)

But this approach you have here is a whole lot of byte-hacking code that needs to be reviewed, not just by me, but by future people concerned about potential security issues. So I am not sure it's worth it to allow file-sharing with future architectures.

Like I said you could easily write a platform signature to the top of the structure and reject reading files that don't match the current architecture, saving a lot of cognitive load.

jmarantz (commented):
/wait

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: std::array makes them explicitly four bytes long, known by the compiler. string_view makes them runtime four bytes long.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: If I wanted it to be less clear that it was writing precisely the four bytes starting at p, yes. I expect when optimized both ways would turn into identical code.

jmarantz: yeah I was just kind of kidding; I think optimizers are very good at data-path rewriting now.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: Counterpoint: really very little byte hacking at this point, and comparable to [other serialization code in the envoy codebase](https://sourcegraph.com/github.com/envoyproxy/envoy/-/blob/source/common/grpc/codec.cc?L19:43).

And significantly *less* dangerous than the union method that you originally didn't like, and no longer requires the half-safe memcpy function or any endian-ness conditions.

jmarantz: We think there is too much byte hacking in Envoy and there have been various waves of work to eliminate them. Some of them need to stay when there is a strong need to do it for performance or whatever.

I apologize if it came across that I didn't like using unions -- I just wanted to clarify the rationale why you were doing it and suggested minor tweaks to make it more readable. I think I like it more than byte hacking like this :)

ravenblackx: It still ends up with unsafe byte-copying operations with the union option - there's at the very least one safeMemcpyUnsafeSrc involved to copy the data from a string_view to the union-object, even if we were to go with an inconsistent serialization option like you seem to prefer now.

There's really no getting away from the fact that serialization of integers into a fixed size data stream is going to involve byte-twiddling. It's just a matter of if you really think an explicitly unsafe copy operation and the awkwardness of a union is sufficiently better than moving 16 bytes with shifting. I don't think it is, I started changing it back to a union either as the class's internal representation *or* just as a temporary in the serialization functions, and IMO either way is less readable and looks less safe than this current implementation.

Arguably it could just be `ASSERT(incoming_string.size() == sizeof(*this)); safeMemcpyUnsafeSrc(incoming_string, *this)` with no inner union or struct, if we're not messing with bytes and not caring about inconsistent serialization, because within the same binary the serialization will be consistent and the class is acting as a struct itself, but to me that does not look safer, and it can definitely produce an inconsistent cache file layout that will at some point bite for debugging problems. And will go wrong in surprising ways if someone tries to add some other member to the class that isn't supposed to be part of the serialization, or make it virtual, etc.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,94 @@
...
+  // - otherwise -1 becomes 0xffffffff.
+  return static_cast<uint32_t>(static_cast<uint8_t>(p[0])) << 24 |
+         static_cast<uint32_t>(static_cast<uint8_t>(p[1])) << 16 |
+         static_cast<uint32_t>(static_cast<uint8_t>(p[2])) << 8 |
+         static_cast<uint32_t>(static_cast<uint8_t>(p[3]));
+}
+
+// Deserialize 8 bytes into a uint64_t.
+uint64_t deserializeUint64(const char* p) {
+  return static_cast<uint64_t>(deserializeUint32(p)) | deserializeUint32(p + 4);
```
jmarantz, line 36: Hm I thought I commented on this already but I think this is wrong (need to left-shift one of those). Can this be (a) moved to common/common (where maybe we'll find similar functionality we can either de-dup or put near) and (b) unit-tested?

I don't actually see anything specific but maybe in common/network/utility.h or common/common/utility.h in class StringUtil ?

There's no particular synergy with the existing functions there but it would just be the kind of place I'd look for such functions.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,162 @@
...
+   * of the file cache implementation.
+   * This is private so only the unit test can call it, as the version ID should be
+   * a compile-time constant in ordinary use.
+   * @param id The cache version ID to set.
+   */
+  void setCacheVersionId(std::array<char, 4> id) { cache_version_id_ = id; }
+
+  std::array<char, 4> file_id_;
+  std::array<char, 4> cache_version_id_;
+  size_t header_size_;
```
jmarantz, line 151: you'll see that in most places Envoy style is to use uint64_t rather than size_t. 

Also I thought you were going to keep headers & trailers at 32 bits though I' not really that concerned about it.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,162 @@
...
+   * the offset from the start of the file to the start of the serialized trailers proto.
+   * @return the offset in bytes.
+   */
+  off_t offsetToTrailers() const { return offsetToBody() + bodySize(); }
+
+  /**
+   * is this a valid cache file header block for the current code version?
+   * @return True if the block's cache version id and file id match the current version.
+   */
+  bool isValid() const;
```
jmarantz, line 129: this is excellent doc.

jmarantz (commented):
OK I'm fine with this approach, modulo concerns noted blow.

Thanks and the doc/comments really help.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
toddmgreer: I think these could be replaced with absl::big_endian::Load[32|64] and Store[32|64]

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
```diff
@@ -0,0 +1,94 @@
...
+CacheFileFixedBlock::CacheFileFixedBlock() {
+  setFileId(ExpectedFileId);
+  setCacheVersionId(ExpectedCacheVersionId);
+  setHeadersSize(0);
+  setTrailersSize(0);
+  setBodySize(0);
+}
+
+void CacheFileFixedBlock::populateFromStringView(absl::string_view s) {
+  ASSERT(s.size() == size() && size() == 24);
```
toddmgreer, line 64: If we're serializing our the members separately, do we need this ASSERT on size? I don't think padding would matter now.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,162 @@
...
+
+private:
+  /**
+   * sets the fileId, a fixed value used to identify that this is a cache file.
+   * This is private so only the unit test can call it, as the file ID should be
+   * a compile-time constant in ordinary use.
+   * @param id The file ID to set.
+   */
+  void setFileId(std::array<char, 4> id) { file_id_ = id; }
```
toddmgreer, line 138: Rather than modifying these logically constant values for tests, I think it would be cleaner and more representative of reality to have a test just modify the appropriate bytes of a file, and verify that the file gets rejected.

toddmgreer (changes requested):

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: size_t is consistent with the AsyncFileHandle interface, which is what this connects to eventually.
They're still being serialized as 32 bit, but size_t for descriptive purpose and AsyncFileHandle-relation.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
```diff
@@ -0,0 +1,162 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+/**
+ * CacheFileFixedBlock represents a minimal header block on the cache entry file; it
+ * uses a struct that is required to be packed, and explicit byte order, to ensure
+ * consistency.
```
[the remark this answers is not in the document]
ravenblackx: Elaborated in the comment.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: Added a comment explaining why the assert. (It's more a "keep this function and the size function synchronized" thing than anything, but also the string_view should be the right size since that's the caller saying "this is a serialized CacheFileFixedBlock".)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Good call. Done.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.cc
ravenblackx: Thanks so much for knowing about those. 100% not matching anything I would have searched for when looking for these functions.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
jmarantz: For the record I personally am fine with size_t for these kinds of usages but that has not been the practice in Envoy. That's also in conflict with common patterns in Envoy then :)

But I do not see that called out specifically in the style guide. But you can see how waves of inconsistency across the codebase can be a concern. Let me check with other maintainers about whether we should try to clean that up and clarify in the style guide, or let it go.

jmarantz:
Windows CI failure seems real:
```//test/extensions/http/cache/file_system_http_cache:cache_file_fixed_block_test FAILED in 2 out of 2 in 0.3s```

File: source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h (hunk 22 above)
Comment (ravenblackx) [reply source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h]: Well, trying changing it here because I think that's what would cause the Windows CI failure!

File: source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h [hunk 27]
```diff
@@ -133,9 +133,9 @@ class CacheFileFixedBlock {
 private:
   std::array<char, 4> file_id_;
   std::array<char, 4> cache_version_id_;
-  size_t header_size_;
-  size_t trailer_size_;
-  size_t body_size_;
+  uint32_t header_size_;
+  uint64_t body_size_;
+  uint32_t trailer_size_;
```
jmarantz, line 138: nit: put the uint32_t together for better packing :)

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
```diff
@@ -76,7 +65,7 @@ TEST_F(CacheFileFixedBlockTest, ReturnsCorrectOffsets) {
 TEST_F(CacheFileFixedBlockTest, SerializesAndDeserializesCorrectly) {
   CacheFileFixedBlock block;
   block.setHeadersSize(100);
-  block.setBodySize(1000);
+  block.setBodySize(10000000000);
```
jmarantz, line 68: nit: put a const above and express all large numbers in terms of that, like
```
constexpr uint64_t one_billion = 1000*1000*1000;
```
I feel like on some platform/compiler, 10000000000 would not compile because it's not got a size spec like ULL (non-portable) and won't fit in an int, but the above hack will work around that without excessive ugliness cause one billion will fit in an int.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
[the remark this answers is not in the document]
toddmgreer: Future us thanks you for documenting these reasons!

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
toddmgreer: Please also reorder their serialized representations, for two reasons
1. It's easier to understand if the memory and serialized formats match, and
2. Copying 64 bits from an aligned location is probably more efficient.

(OK, any efficiency difference will be completely swamped by the filesystem operations, but reason 1 is real.)

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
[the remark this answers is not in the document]
toddmgreer: BTW, it is possible to make proto messages that have deterministic serialization (in this case by using fixed32/64 fields and making everything `optional`), but the documentation is clear that we should not rely on that. Such determinism is an undocumented property of the implementation. The current approach is better.

test/extensions/http/cache/file_system_http_cache/cache_file_fixed_block_test.cc
ravenblackx: Good call. Absolutely bonkers that `uint64_t x = 10000000000` will compile and work [on clang] but `uint64_t x = 1000*1000*1000*10` does not.

source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h
ravenblackx: Done.

jmarantz (approved):
thanks!

jmarantz:
@toddmgreer lmk if you are good with this and I'll merge.

toddmgreer (approved):

PR #23851, opened 2022-11-06
stats: Scope should not be a parent-class of Store

jmarantz:
Commit Message: Refactors so that Stats::Store is no longer a subclass of Stats::Scope. Previously it was, so that the root-scope was effectively the same thing as the Store. This causes issues and confusion because Stores are typically created directly as test-class member variables, but Scopes were *mostly* created as shared_ptr for various reasons. This refactor attempts to cleanly, but minimally, enable the memory management of Scopes to be *strictly* shared_ptr. It also separate two different concepts.

Additional Description:

This leaves behind a bit of technical debt to keep this refactor reviewable. This is tracked in #24007 and can be addressed in a follow-up PR that will be much larger than this one, but will be strictly a series of simple replacements (`makeStats(store_) --> makeStats(*store_->rootScope())`)

To reviewers: I recommend looking at scope.h and store.h carefully to see the real change that was done here. Then also look at `isolated_store_impl.*`, `stat_test_utility.*`, and the changes to the mocks. Finally look at `test/integration/server.h` for the change in the thread-safe wrapper around isolated store there. Everything else should be pretty quick to review once you have that context.

Risk Level: medium -- this is a pretty large refactor
Testing: //test/...
Docs Changes: n/a
Release Notes: n/a
Fixes: #23862


jmarantz:
Xin -- can you take a pass through this?

envoy/stats/store.h
```diff
@@ -27,10 +27,53 @@ class Sink;
...
+ * individual stats. Each stat is defined in a scope. There is a single root
+ * scope created in the Store, and more sub-scopes can be created.
+ *
+ * Stores enable iteration over all stats in its owned Scopes.
+ *
+ * There is typically one Store instance in a test or binary, though Isolated
+ * Stores can be created in some scenarios. Stores are typically allocated
+ * as part of other objects or via std::unique_ptr.
+ *
+ * In contast, Scopes are managed via shared_ptr, unique nickname
```
pradeepcrao, line 40: s/contast/contrast

envoy/stats/store.h
```diff
@@ -27,10 +27,53 @@ class Sink;
 class SinkPredicates;
 
 /**
- * A store for all known counters, gauges, and timers.
+ * Store keeps track of all Scopes created in it, and the Scopes manage
+ * individual stats. Each stat is defined in a scope. There is a single root
+ * scope created in the Store, and more sub-scopes can be created.
+ *
```
stevenzzzz, line 33: could you clarify a little bit here: 
could sub-scopes create a tree-like scope hierarchy? and anything special about their life-cycle management?

The store can do iteration over all-stats in its-owned scoped, including sub-scopes(children,grand-children scopes...)?

envoy/stats/store.h
```diff
@@ -27,10 +27,53 @@ class Sink;
...
+
+  /**
+   * @return the root scope (creating it if necessary)
+   */
+  virtual ScopeSharedPtr rootScope() PURE;
+
+  /**
+   * @return the root scope (creating it if necessary)
+   */
+  virtual ConstScopeSharedPtr constRootScope() const PURE;
```
stevenzzzz, line 60: not sure if this "constXXXX" is a pattern in Envoy. 
I think it's usually just 
const xxx()const

envoy/stats/store.h
```diff
@@ -81,6 +158,50 @@ class Store : public Scope {
   virtual void forEachSinkedCounter(SizeFn f_size, StatFn<Counter> f_stat) const PURE;
```
stevenzzzz, line 158: it took me a while to understand the semantics of the two parameters, I know this is not your fault, but it would be nice if we can add some comments about the parameters.

envoy/stats/store.h
jmarantz: I have taken to using it because sometimes the overload is ambiguous for my purposes.

envoy/stats/store.h
jmarantz: the doxygen comments are above. Do you have a clarification for those?

envoy/stats/store.h
jmarantz: Yes good point. I added some more text. PTAL.

envoy/stats/store.h
```diff
@@ -70,6 +115,40 @@ class Store : public Scope {
...
+  /**
+   * @param The name of the stat, obtained from the SymbolTable.
+   * @return a reference to a text readout within the entire store, if it exists.
+   */
+  virtual TextReadoutOptConstRef findTextReadout(StatName name) const PURE;
+
+  /**
+   * @return a null counter that will ignore increemts and always return 0.
+   */
+  virtual Counter& nullCounter() PURE;
```
stevenzzzz, line 145: do we need a const version?

envoy/stats/store.h
stevenzzzz: sorry, it's folded so I didn't see the parameters. All set.

source/common/stats/isolated_store_impl.cc
```diff
@@ -34,21 +33,37 @@ IsolatedStoreImpl::IsolatedStoreImpl(SymbolTable& symbol_table)
         return alloc_.makeTextReadout(name, name, StatNameTagVector{});
       }),
       null_counter_(new NullCounterImpl(symbol_table)),
-      null_gauge_(new NullGaugeImpl(symbol_table)),
-      default_scope_(std::make_shared<ScopePrefixer>("", *this)) {}
+      null_gauge_(new NullGaugeImpl(symbol_table)) {}
+
+ScopeSharedPtr IsolatedStoreImpl::rootScope() {
+  if (lazy_default_scope_ == nullptr) {
+    StatNameManagedStorage name_storage("", symbolTable());
```
stevenzzzz, line 40: could you help me to understand the lifecycle management of name_storage vs lazy_default_scope_ ?

source/common/stats/isolated_store_impl.cc
```diff
@@ -34,21 +33,37 @@ IsolatedStoreImpl::IsolatedStoreImpl(SymbolTable& symbol_table)
...
+    lazy_default_scope_ = makeScope(StatName(name_storage.statName()));
```
stevenzzzz, line 41: why we need another StatName here when statName() is already returning a StatName?

source/common/stats/isolated_store_impl.cc
stevenzzzz: plus, is there a thread-safe issue on lazy_default_scope_?

jmarantz: Yes: the scope that's created below copies and owns the StatName.

Remember, StatName is like string_view: APIs make take a string-view arg, and will usually need to copy it into owned storage.

source/common/stats/isolated_store_impl.cc
jmarantz: um we don't. removing.

jmarantz (commented):
Thanks for the comments. I see something has changed in the coverage metrics and this has fallen slightly out of limits and will follow-up to get CI to pass.

source/common/stats/isolated_store_impl.cc
stevenzzzz: what about the thread-safety question?

jmarantz: Oh, sorry. IsolatedStoreImpl is not a thread-safe class. There is a thread-safe wrapper in test/server.

jmarantz: I will add a comment (but not push it right now, no need to waste the CI cycles as I need to do another push to fix coverage anyway).

source/common/stats/isolated_store_impl.h
```diff
@@ -268,9 +243,126 @@ class IsolatedStoreImpl : public StoreImpl {
   IsolatedStatsCache<TextReadout> text_readouts_;
   RefcountPtr<NullCounterImpl> null_counter_;
   RefcountPtr<NullGaugeImpl> null_gauge_;
-  ScopeSharedPtr default_scope_;
+  mutable ScopeSharedPtr lazy_default_scope_;
```
pradeepcrao, line 246: I could not find a reason that requires this to be mutable. Maybe it needed to be before but not anymore? (Or I missed it)

jmarantz: Quite right! It is conceptually mutable, because there's a constRootScope() method which can write it. That is implemented with a const_cast to avoid duplicating code, but conceptually I think lazy_default_scope_ should be considered mutable.

If we didn't use a const_cast in the impl it would have to be mutable (as well as the symbol table).

test/common/stats/thread_local_store_test.cc
```diff
@@ -1955,7 +1990,62 @@ TEST_F(HistogramThreadTest, ScopeOverlap) {
   EXPECT_EQ(0, numTlsHistograms());
 
   shutdownThreading();
-  store_->histogramFromString("histogram_after_shutdown", Histogram::Unit::Unspecified);
+  scope_.histogramFromString("histogram_after_shutdown", Histogram::Unit::Unspecified);
+}
+
+class TestSinkPredicates : public Stats::SinkPredicates {
```
jmarantz, line 1996: @pradeepcrao I added this because coverage complained about ThreadLocalStore::setSinkPredicates not being covered. I realize this might conflict with your histogram predicates PR, but I don't think it will be hard to resolve. I feel like in that PR you might have broken out TestSinkPredicates into a header so we should clean this when that lands.

pradeepcrao: SGTM.

pradeepcrao (commented):
LGTM

envoy/stats/store.h
```diff
@@ -70,6 +115,40 @@ class Store : public Scope {
   virtual void forEachHistogram(SizeFn f_size, StatFn<ParentHistogram> f_stat) const PURE;
   virtual void forEachScope(SizeFn f_size, StatFn<const Scope> f_stat) const PURE;
 
+  /**
+   * @param The name of the stat, obtained from the SymbolTable.
+   * @return a reference to a counter within the entire store, if it exists.
+   */
+  virtual CounterOptConstRef findCounter(StatName name) const PURE;
```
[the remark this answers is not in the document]
jmarantz: Hm. Not all of them. I could not find a use for findCounter, but Scope::findGauge is used in stat_merger.cc which is used to support preserving gauge values across hot restarts, and requires lookups by name.

However those find* methods *are* used in tests when the implementation is thread_local_store.cc; not sure the test-specific classes. So I think we need to leave them in. We *could* rename findCounter to be findCounterForTest, but we'd need to leave findGauge named as is.

Also, these methods are not new; they were in Scope; but I added them to Store now that it does not subclass Scope. What I could do is leave a TODO to rename the Store ones to findCounterForTest, but I don't want to blow up this PR as there are a lot of call-sites in tests.

jmarantz:
Greg - I think this is ready for a pass; no super great rush but IMO this is the right direction, and will also make it easier for @stevenzzzz  to build and test a mechanism that uses scopes to hold onto live cluster stats across and xDS updaets.

envoy/stats/store.h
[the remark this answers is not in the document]
jmarantz: reflecting on this -- I think all the 'find' methods in Store were not needed. And it turns out we only needed one for Gauges, and it works fine to have that be a free test-util function.

envoy/stats/store.h
```diff
@@ -70,6 +115,16 @@ class Store : public Scope {
   virtual void forEachHistogram(SizeFn f_size, StatFn<ParentHistogram> f_stat) const PURE;
   virtual void forEachScope(SizeFn f_size, StatFn<const Scope> f_stat) const PURE;
 
+  /**
+   * @return a null counter that will ignore increemts and always return 0.
```
ggreenway, line 119:
```suggestion
   * @return a null counter that will ignore increments and always return 0.
```

envoy/stats/store.h
```diff
@@ -81,6 +136,50 @@ class Store : public Scope {
...
+  // of https://github.com/envoyproxy/envoy/pull/23851, which detaches the
+  // inheritance of Scope as a parent of Store. There is semantic complexity to
+  // that PR, so it's going to be easier review if it's as small as possible.
+  //
+  // A follow-up PR is required to remove the functions below, which will
+  // require a large number of files to be trivially changed, by explicitly
+  // accessing the rootScope() to call these methods.
+  operator Scope&() { return *rootScope(); }
+
+  // Delegate somea methods to the root scope; these are exposed to make it more
```
ggreenway, line 163:
```suggestion
  // Delegate some methods to the root scope; these are exposed to make it more
```

envoy/stats/store.h
```diff
@@ -81,6 +136,50 @@ class Store : public Scope {
...
+  // convenient to use stats_macros.h. We mayconsider dropping them if desired,
```
ggreenway, line 164:
```suggestion
  // convenient to use stats_macros.h. We may consider dropping them if desired,
```

source/common/stats/isolated_store_impl.h
```diff
@@ -268,9 +236,126 @@ class IsolatedStoreImpl : public StoreImpl {
...
```
ggreenway, line 239: Why is this lazy-created? Why is there ever not a root scope, and so why can't it be created upon construction? It seems like some other parts could be simpler if `rootScope()` never has to create it.

ggreenway (changes requested):
This looks good overall. I like the direction it's going.

/wait

source/common/stats/isolated_store_impl.h
jmarantz: It makes it easier to subclass IsoaltedStoreImpl, allowing the infrastructure to create and maintain scopes to be shared, using a virtual method to construct a Scope of the appropriate type.

I will add a comment.

envoy/stats/store.h
jmarantz: done

envoy/stats/store.h
jmarantz: done

envoy/stats/store.h
jmarantz: done

source/common/stats/isolated_store_impl.h
jmarantz: Perhaps, but it's a little messy since there are two levels of subclassing of this in our test stores, and so the intermediate one (which is often used directly) would I guess need to make one and then have it erased by its subclass.

I felt it was cleaner to do this lazily after all construction is done

ggreenway (approved):

PR #23874, opened 2022-11-08
dlb: add retry logic

daixiang0:
When big traffic come, dlb_send() may fail due to temporarily unavailable resource.

Add retry logic to ensure every connection is sent successfully by dlb.

Risk Level: Low
Testing: Add
Release Notes: N/A

### DIFF
--- a/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
+++ b/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
@@ -33,5 +33,5 @@ message Dlb {
 
   // Control how many times Envoy makes an attempt to send event to DLB hardware.
   // No retry as default.
-  uint32 retry_times = 2;
+  uint32 most_send_retries = 2;
 }
--- a/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
+++ b/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
@@ -31,7 +31,7 @@ message Dlb {
   // If not specified, use the first available device as default.
   uint32 id = 1;
 
-  // Control how many times Envoy makes an attempt to send event to DLB hardware.
+  // Maximum number of retries when sending to DLB device fails.
   // No retry as default.
-  uint32 most_send_retries = 2;
+  uint32 max_retries = 2;
 }
--- a/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
+++ b/api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
@@ -30,4 +30,8 @@ message Dlb {
   // The ID of the Dlb hardware, start from 0.
   // If not specified, use the first available device as default.
   uint32 id = 1;
+
+  // Control how many times Envoy makes an attempt to send event to DLB hardware.
+  // No retry as default.
+  uint32 retry_times = 2;
 }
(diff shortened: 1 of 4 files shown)


phlax:
ci errors look genuine:

```console
contrib/network/connection_balance/dlb/source/connection_balancer_impl.cc:49:3: error: use of undeclared identifier 'retry_times'
  retry_times = dlb_config.retry_times();
```

daixiang0:
@phlax thanks, have fixed it.

api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
```diff
@@ -30,4 +30,8 @@ message Dlb {
   // The ID of the Dlb hardware, start from 0.
   // If not specified, use the first available device as default.
   uint32 id = 1;
+
+  // Control how many times Envoy makes an attempt to send event to DLB hardware.
+  // No retry as default.
+  uint32 retry_times = 2;
```
adisuissa, line 36: nit: consider renaming to `max_post_retries`, or something that hints that this is for posting/sending an event.
Also consider adding a maximal value here, if there is something that makes sense.

contrib/network/connection_balance/dlb/test/config_test.cc
```diff
@@ -41,17 +41,20 @@ TEST_F(DlbConnectionBalanceFactoryTest, MakeDefaultConfig) {
...
   envoy::config::core::v3::TypedExtensionConfig typed_config;
 
   envoy::extensions::network::connection_balance::dlb::v3alpha::Dlb dlb;
   dlb.set_id(10);
+  dlb.set_retry_times(12);
 
   makeDlbConnectionBalanceConfig(typed_config, dlb);
   verifyDlbConnectionBalanceConfigAndUnpack(typed_config, dlb);
   EXPECT_EQ(10, dlb.id());
+  EXPECT_EQ(12, dlb.retry_times());
```
adisuissa, line 57: Just a suggestion: these tests do not actually test that the feature performs as it should, they just test that the proto works. You may want to consider testing the actual code.

daixiang0: Thanks for your suggestion, I will do it when mock functions are ready.

api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
daixiang0: Since the value depends on specific cases, so hard to limit max here.

api/contrib/envoy/extensions/network/connection_balance/dlb/v3alpha/dlb.proto
```diff
@@ -30,4 +30,8 @@ message Dlb {
   // The ID of the Dlb hardware, start from 0.
   // If not specified, use the first available device as default.
   uint32 id = 1;
+
+  // Control how many times Envoy makes an attempt to send event to DLB hardware.
+  // No retry as default.
+  uint32 most_send_retries = 2;
```
adisuissa, line 36: Sorry, re-read this doc and I think it doesn't reflect the semantics of the field, because the field counts retries, and the comment says attempts.
The number of attempts equals the number of retries + 1.

The field name should be either `max_attempts` or `max_retries`, and the comment and implementation should reflect its semantics.

daixiang0: Updated.

adisuissa (approved):
/lgtm api
LGTM, thanks!

daixiang0:
@adisuissa could we merge it?

Issue #23560, opened 2022-10-19
GCP auth filter supports configuring header name


rulex123:
*Title*: *GCP Authentication Filter supports configuring request header that holds GCP authentication token.*

*Description*:
Currently, the filter fetches the GCP authentication token from the GCE metadata server and adds it
to the request in the HTTP Authorization header. This might be fine for most use cases, however there
might be exceptions where services expect the GCP token to arrive in a different HTTP header. 

Our backend services are one such exception, so it would help us if the filter supported configuration of the 
header name. 

If this seems alright, I could work on contributing a patch.

PR #23877, opened 2022-11-08
[gcp_auth] Support configuring header that holds token

rulex123:
Commit Message: add support for configuring header that holds token fetched from the GCE metadata server  (see https://github.com/envoyproxy/envoy/issues/23560)
Additional Description: n/a
Risk Level: low
Testing: updated integration tests
Docs Changes: updated API protos
Release Notes: updated release notes


api/envoy/extensions/filters/http/gcp_authn/v3/gcp_authn.proto
```diff
@@ -32,6 +32,11 @@ message GcpAuthnFilterConfig {
 
   // Token cache configuration. This field is optional.
   TokenCacheConfig cache_config = 3;
+
+  // Name of the request header that holds the token. If this field is not specified, the token
```
mattklein123, line 36: nit:

```suggestion
  // Name of the request header that holds the token. If this field is empty, the token
```

source/extensions/filters/http/gcp_authn/gcp_authn_filter.cc
```diff
@@ -13,9 +13,14 @@ namespace Extensions {
...
 namespace {
-void addTokenToRequest(Http::RequestHeaderMap& hdrs, absl::string_view token_str) {
-  std::string id_token = absl::StrCat("Bearer ", token_str);
-  hdrs.setCopy(authorizationHeaderKey(), id_token);
+void addTokenToRequest(Http::RequestHeaderMap& hdrs, absl::string_view token_str,
+                       absl::string_view header_name) {
+  if (!header_name.empty()) {
+    hdrs.setCopy(Http::LowerCaseString(header_name), token_str);
+  } else {
+    std::string id_token = absl::StrCat("Bearer ", token_str);
```
mattklein123, line 21: It seems odd to me that we prefix with "Bearer" only in this case, but not in the other case. If there is a reason for that can you document that also?

source/extensions/filters/http/gcp_authn/gcp_authn_filter.cc
```diff
@@ -91,11 +96,11 @@ void GcpAuthnFilter::onComplete(const Http::ResponseMessage* response) {
   state_ = State::Complete;
   if (!initiating_call_) {
     if (response != nullptr) {
-      // Modify the request header to include the ID token in an `Authorization: Bearer ID_TOKEN`
-      // header.
+      // Modify the request header to include the ID token in a header (bu default, the
```
mattklein123, line 99:
```suggestion
      // Modify the request header to include the ID token in a header (by default, the
```

mattklein123 (changes requested):
LGTM with small comments, thanks.

/wait

source/extensions/filters/http/gcp_authn/gcp_authn_filter.cc
erikjoh: For our use case we would like to have it like this.
```
<custom header name>: <token>
```
But it could be that others may want the `Bearer` prefix when using a custom header name.
Wdyt about taking some inspiration from the [JWT filter](https://www.envoyproxy.io/docs/envoy/v1.24.0/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtheader) and allow configuring the prefix for this filter as well?
So the default would something like:
```yaml
token_header:
  name: "Authorization"
  value_prefix: "Bearer "
```
and then for our use case we can just do this:
```yaml
token_header:
  name: "our-custom-header"
  value_prefix: ""
```

rulex123: Seems like a good compromise to me, I will go ahead and update the code.

changelogs/current.yaml
```diff
@@ -98,5 +98,8 @@ new_features:
 - area: redis
   change: |
     extended :ref:`cluster support <arch_overview_redis_cluster_support>` by adding a :ref:`dns_cache_config <envoy_v3_api_field_extensions.filters.network.redis_proxy.v3.RedisProxy.ConnPoolSettings.dns_cache_config>` option that can be used to resolve hostnames returned by MOVED/ASK responses.
+- area: gcp_authn
+  change: |
+    added support for configuring header that holds token fetched from GCE metadata server.
```
mattklein123, line 103: Can you ref link to the new field?

mattklein123 (changes requested):
LGTM with small comment and also needs main merge. Thank you!

/wait

test/extensions/filters/http/gcp_authn/gcp_authn_filter_integration_test.cc
```diff
@@ -222,6 +235,26 @@ TEST_P(GcpAuthnFilterIntegrationTest, Basicflow) {
...
+  initializeConfig(/*add_audience=*/true, true);
+  HttpIntegrationTest::initialize();
+  int num = 2;
+  // Send multiple requests.
+  for (int i = 0; i < num; ++i) {
+    initiateClientConnection();
+    // Send the request to cluster `gcp_authn`.
+    waitForGcpAuthnServerResponse();
+    // Send the request to cluster `cluster_0` and validate the response.
+    sendRequestToDestinationAndValidateResponse(/*with_audience=*/true, true);
```
tyxia, line 248: nit : please add annotation for code readability, so that people don't need to check the function definition to figure out which field is set to true.
`/*configure_token_header=*/true`

test/extensions/filters/http/gcp_authn/gcp_authn_filter_test.cc
```diff
@@ -38,6 +38,7 @@ constexpr char DefaultConfig[] = R"EOF(
         base_interval: 1s
         max_interval: 10s
       num_retries: 5
+    header_name: customheader
```
tyxia, line 41: What is this `header_name` pointing to in the config? You are adding the `TokenHeader token_header = 4;`, right?

tyxia (commented):
Thank you very much for contributing to this filter!

When I introduced this filter, I just follow the philosophy-- `keep it simple and add things until it is necessary`. It is super glad to see people are using this filter making it better. Thanks

source/extensions/filters/http/gcp_authn/gcp_authn_filter.cc
erikjoh: Wouldn't there also be some value in being consistent with the naming in `extensions.filters.http.jwt_authn.v3.JwtHeader` (which uses `value_prefix` for this)?
I could imagine cases where gcp_authn and jwt_authn are used together in a client/server pair and in that case some consistency might make it slightly more user friendly.

tyxia: I didn't check the `jwtHeader`. Yea, I agree it is good to be consistent with the naming there.

rulex123: For the sake of consistency, I left `value_prefix` alone :-)

source/extensions/filters/http/gcp_authn/gcp_authn_filter.cc
```diff
@@ -1,6 +1,11 @@
 #include "source/extensions/filters/http/gcp_authn/gcp_authn_filter.h"
 
+#include <iostream>
 #include <memory>
+#include <stdexcept>
```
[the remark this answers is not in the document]
rulex123: hehe nope, well spotted. I removed now.

source/extensions/filters/http/gcp_authn/gcp_authn_filter.cc
```diff
@@ -54,7 +66,8 @@ Http::FilterHeadersStatus GcpAuthnFilter::decodeHeaders(Http::RequestHeaderMap&
       if (token != nullptr) {
         // If token is found in the cache, we add the token string to the request directly and
         // continue the filter chain iteration.
-        addTokenToRequest(hdrs, token->jwt_);
+        addTokenToRequest(hdrs, token->jwt_, filter_config_->has_token_header(),
```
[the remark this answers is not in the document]
rulex123: Ty for this suggestion, I like the idea of saving us one function arg. Updated the code.

test/extensions/filters/http/gcp_authn/gcp_authn_filter_test.cc
rulex123: this was some leftover from a previous iteration I think. Removed now.

rulex123:
I think I have addressed all feedback, thank you @mattklein123 and @tyxia for reviewing! I have also merged main so we should get CI to run this time around :-)

mattklein123 (approved):
Thanks!

tyxia (approved):
Thank you!

PR #23882, opened 2022-11-08
[WIP] listener: enable to add/remove addresses without draining the whole listener.

### LINKED ISSUE #23431: Add address or remove address to a multiple addresses listener without draining the unchanged address connections
Currently, when adding a new address or removing an existing address from a multiple addresses listener, it will lead to a new listener generated, and draining the old listener, then all the existing connections will be drained. 

There is a requirement from the istio side https://github.com/istio/istio/pull/41248. For headless service, Istio will create a listener for each endpoint original. Now with the multiple addresses listener, they can create a single listener for all the endpoints. But when the endpoints are changed, Istio needs to add or remove address to the multiple addresses listener and look for not draining all the existing connections which are on the unchanged address.

For adding address, it should be able done by enabling ConnectionHandleImpl to add `ActiveListener` for a new address to an existing listener. 

For removing address, it should be able done by enabling ConnectionHandleImpl to stop and remove `ActiveListener` for a specific address on a listener. And ListenerManager will be able to drain an `ActiveListener` on a specific address instead of the whole listener.

soulxu:
Commit Message: listener: enable to add/remove addresses without draining the whole listener.
Additional Description:
Risk Level: high
Testing: TODO
Docs Changes: TODO
Release Notes: TODO
Fixes #23431

### DIFF
--- a/envoy/network/connection_handler.h
+++ b/envoy/network/connection_handler.h
@@ -76,8 +76,10 @@ public:
    * Stop listeners using the listener tag as a key. This will not close any connections and is used
    * for draining.
    * @param listener_tag supplies the tag passed to addListener().
+   * @param addresses will be stopped. If not specified, the whole listener will be stopped.
    */
-  virtual void stopListeners(uint64_t listener_tag) PURE;
+  virtual void stopListeners(uint64_t listener_tag,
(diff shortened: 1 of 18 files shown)


soulxu:
still working on it

PR #23893, opened 2022-11-08
xds: Add TTL support to the KeyValueStore xDS delegate

abeyad:
https://github.com/envoyproxy/envoy/pull/22921 added the KeyValueStore based xDS delegate, but the TTLs were not set on the resources processed by the delegate.

Now that support has been added for TTLs in the KeyValueStore interface[1][2], this commit ensures the TTL of the resource gets processed by the KeyValueStore based delegate.

[1] https://github.com/envoyproxy/envoy/pull/22884 
[2] https://github.com/envoyproxy/envoy/pull/23318


alyssawilk (approved):

alyssawilk:
Awesome, thanks for following through on this one!  Throwing over to @wbpcode for !google review

Issue #20086, opened 2022-02-23
support ECDS in config_dump


zirain:
*Title*: *support ECDS in config_dump*

*Description*:
>is there any plan to support ECDS in config_dump

yanjunxiang-google:
I think the config_dump support for ECDS is still pending. Isn't it?

zirain:
@yanjunxiang-google do you plan to working on this?

yanjunxiang-google:
FYI, I am looking at this now, will send out PR soon.

zirain:
> FYI, I am looking at this now, will send out PR soon.

ok, thanks

zema-MSFT:
> FYI, I am looking at this now, will send out PR soon.

hi, Yanjun, could you share the PR for this fix? is there any ETA for this feature fix? currently it's painful for us to debug issues when applying ECDS. Thanks!

zirain:
@zema-MSFT https://github.com/envoyproxy/envoy/pull/23902

zema-MSFT:
> @zema-MSFT #23902

thanks

zirain:
fixed by https://github.com/envoyproxy/envoy/pull/23902

zirain:
I test #23902  with istio master branch, got following output:
```json
{
   "@type": "type.googleapis.com/envoy.admin.v3.EcdsConfigDump",
   "ecds_filters": [
    {
     "ecds_filter": {
      "@type": "type.googleapis.com/envoy.config.core.v3.TypedExtensionConfig",
      "name": "default.httpbin-rate-limiting"
     },
     "last_updated": "1970-01-01T00:00:00Z"
    },
    {
     "version_info": "2022-12-01T02:42:30Z/10",
     "ecds_filter": {
      "@type": "type.googleapis.com/envoy.config.core.v3.TypedExtensionConfig",
      "name": "default.display-metadata",
      "typed_config": {
       "@type": "type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm",
       "config": {
        "name": "default.display-metadata",
        "vm_config": {
         "runtime": "envoy.wasm.runtime.v8",
         "code": {
          "local": {
           "filename": "/var/lib/istio/data/a53784b18caf61bdcce2bb73d45d6bae13e81f9173295992e4612db7be72b439/064f7cd90a62e9345e77031bffcaf949c07d76fefe63be0ff969a1579de5ae57.wasm"
          }
         },
         "environment_variables": {
[...]

PR #23902, opened 2022-11-08
Adding ECDS config dump support.

yanjunxiang-google:
For issue: https://github.com/envoyproxy/envoy/issues/20086

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/filter/config_discovery_impl.h
```diff
@@ -490,7 +501,42 @@ class FilterConfigProviderManagerImpl : public FilterConfigProviderManagerImplBa
...
+  }
+
 private:
+  // Data structure to hold data related to filter configuration.
+  struct FilterConfigData {
+    std::string last_version_info_;
+    SystemTime last_updated_;
+    envoy::config::core::v3::TypedExtensionConfig filter_config_;
+  };
+  absl::flat_hash_map<std::string, struct FilterConfigData> filter_config_store_;
```
adisuissa, line 537: I wonder how much memory this may end up using, especially if not used.
@jmarantz from your experience on the admin component, is memory consumption of saving raw protobufs something that we should take into consideration?
If so, will explicitly enabling/disabling the ECDS filter tracking the right way to go?

adisuissa (commented):
Thanks!
Overall the API LGTM.
2 high level comments/questions:
1. Memory consumption could be non-negligible here. Is there anything we can do to only impact those that want to use this feature?
2. Should there be a way to filter for a specific ECDS config (e.g., using https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--config_dump?name_regex=)?

api/envoy/admin/v3/config_dump_shared.proto
```diff
@@ -370,3 +370,43 @@ message EndpointsConfigDump {
...
+    // updated or acknowledged resource, this field should be empty.
+    // [#not-implemented-hide:]
+    UpdateFailureState error_state = 4;
+
+    // The client status of this resource.
+    // [#not-implemented-hide:]
+    ClientResourceStatus client_status = 5;
+  }
+
+  // The  ECDS filter configs.
```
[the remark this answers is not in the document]
yanjunxiang-google: Done for the comments change!

source/common/filter/config_discovery_impl.h
kyessenov: This is a substantial concern for Wasm filters (which can be O(10MB) each). Although it's probably marginal on top of v8 overhead, but it does add O(10MB) at rutnime.

source/common/filter/config_discovery_impl.cc
```diff
@@ -131,6 +131,9 @@ void FilterConfigSubscription::onConfigUpdate(
   last_type_url_ = type_url;
   last_version_info_ = version_info;
   last_factory_name_ = factory_name;
+  filter_config_provider_manager_.storeFilterConfigs(filter_config, last_type_url_,
```
kyessenov, line 134: Could we avoid duplicating `filter_config`? I think `message` is the same content?

kyessenov: Just to add more context: I think there is one subscription on the main thread, so we can probably reuse that one for the config dump. That helps reducing overhead with more copies of the config.

source/common/filter/config_discovery_impl.h
```diff
@@ -387,9 +392,42 @@ class FilterConfigProviderManagerImplBase : Logger::Loggable<Logger::Id::filter>
...
+
+  ProtobufTypes::MessagePtr dumpEcdsFilterConfigs(const Matchers::StringMatcher& name_matcher) {
+    auto config_dump = std::make_unique<envoy::admin::v3::EcdsConfigDump>();
+    for (const auto& subscription : subscriptions_) {
+      const auto& ecds_filter = subscription.second.lock();
+      if (!name_matcher.match(ecds_filter->name())) {
+        continue;
+      }
+      // Put the filter info into a typed extension proto.
+      envoy::config::core::v3::TypedExtensionConfig filter_config;
```
yanjunxiang-google, line 415: If we directly pack lastConfig() into config_dump any filed, then config dump doesn't have the filter name info. It makes it hard to know what's the filter name. Thus we first put the lastConfig() + name into this typed extension proto.

source/common/filter/config_discovery_impl.h
```diff
@@ -387,9 +392,42 @@ class FilterConfigProviderManagerImplBase : Logger::Loggable<Logger::Id::filter>
                                          absl::string_view type_url) const;
   void validateProtoConfigTypeUrl(const std::string& type_url,
                                   const absl::flat_hash_set<std::string>& require_type_urls) const;
+  virtual const std::string getConfigDumpType() const PURE;
 
 private:
+  void setupEcdsConfigDumpCallbacks(Server::Admin& admin) {
+    if (config_tracker_entry_ == nullptr) {
+      config_tracker_entry_ = admin.getConfigTracker().add(
+          getConfigDumpType(), [this](const Matchers::StringMatcher& name_matcher) {
```
yanjunxiang-google, line 401: the config dump callbacks are stored in a keyed hashmap, thus the keys has to be different to avoid one override the other. Thus we need getConfigDumpType() to get different strings for each filter type, like HTTP filter, TCP listener filter,...

test/integration/extension_discovery_integration_test.cc
```diff
@@ -775,5 +856,68 @@ TEST_P(ExtensionDiscoveryIntegrationTest, ReloadBoth) {
...
+  test_server_->waitForGaugeGe("listener_manager.workers_started", 1);
+  EXPECT_EQ(test_server_->server().initManager().state(), Init::Manager::State::Initialized);
+  BufferingStreamDecoderPtr response;
+  EXPECT_EQ("200", request("admin", "GET", "/config_dump", response));
+  EXPECT_EQ("application/json", contentType(response));
+  Json::ObjectSharedPtr json = Json::Factory::loadFromString(response->body());
+  size_t index = 0;
+  const std::string expected_types[] = {"type.googleapis.com/envoy.admin.v3.BootstrapConfigDump",
+                                        "type.googleapis.com/envoy.admin.v3.ClustersConfigDump",
+                                        "type.googleapis.com/envoy.admin.v3.EcdsConfigDump", // HTTP
```
yanjunxiang-google, line 881: If both HTTP ECDS filter and listener ECDS filters are configured, then there will be two EcdsConfigDump entries in the /config_dump.  Since the data are stored in two different objects, there is no way to dump them into one EcdsConfigDump proto.
1) For listener ECDS filter: https://github.com/envoyproxy/envoy/blob/b0d79665240cdfe3d87280283d28ac30334542e9/source/server/listener_manager_impl.h#L111

2) For HTTP ECDS filter:
https://github.com/envoyproxy/envoy/blob/b0d79665240cdfe3d87280283d28ac30334542e9/source/extensions/filters/network/http_connection_manager/config.cc#L224

source/common/filter/config_discovery_impl.h
yanjunxiang-google: Understood

source/common/filter/config_discovery_impl.cc
yanjunxiang-google: Yes, that's a great idea! The latest code is directly to dump the data in the already existing subscriptions_ hash map: https://github.com/envoyproxy/envoy/blob/b0d79665240cdfe3d87280283d28ac30334542e9/source/common/filter/config_discovery_impl.h#L392

With this, there is no new memory consumption added by this PR.

yanjunxiang-google:
@envoyproxy/api-shepherds PTAL

test/integration/listener_extension_discovery_integration_test.cc
```diff
@@ -537,5 +554,122 @@ TEST_P(ListenerExtensionDiscoveryIntegrationTest, DestroyDuringInit) {
...
+  BufferingStreamDecoderPtr response;
+  EXPECT_EQ("200", request("admin", "GET", "/config_dump", response));
+  EXPECT_EQ("application/json", contentType(response));
+  Json::ObjectSharedPtr json = Json::Factory::loadFromString(response->body());
+  size_t index = 0;
+  const std::string expected_types[] = {"type.googleapis.com/envoy.admin.v3.BootstrapConfigDump",
+                                        "type.googleapis.com/envoy.admin.v3.ClustersConfigDump",
+                                        "type.googleapis.com/envoy.admin.v3.EcdsConfigDump",
+                                        "type.googleapis.com/envoy.admin.v3.ListenersConfigDump",
+                                        "type.googleapis.com/envoy.admin.v3.SecretsConfigDump"};
```
adisuissa, line 581: Same as above.

source/common/filter/config_discovery_impl.h
```diff
@@ -564,6 +603,9 @@ class TcpListenerFilterConfigProviderManagerImpl
           TcpListenerDynamicFilterConfigProviderImpl> {
 public:
   absl::string_view statPrefix() const override { return "tcp_listener_filter."; }
+
+protected:
+  const std::string getConfigDumpType() const override { return "ecds_filter_tcp_listener"; }
```
kyessenov, line 608: Suggestion: `statPrefix` seems like it accomplishes the same goal of partitioning configs in the stats hierarchy. Maybe reuse?

source/common/filter/config_discovery_impl.h
```diff
@@ -387,9 +392,41 @@ class FilterConfigProviderManagerImplBase : Logger::Loggable<Logger::Id::filter>
...
+    auto config_dump = std::make_unique<envoy::admin::v3::EcdsConfigDump>();
+    for (const auto& subscription : subscriptions_) {
+      const auto& ecds_filter = subscription.second.lock();
+      if (!name_matcher.match(ecds_filter->name())) {
+        continue;
+      }
+      // Put the filter info into a typed extension proto.
+      envoy::config::core::v3::TypedExtensionConfig filter_config;
+      filter_config.set_name(ecds_filter->name());
+      filter_config.mutable_typed_config()->PackFrom(*ecds_filter->lastConfig());
```
kyessenov, line 417: Is lastConfig always non-null?

source/common/filter/config_discovery_impl.h
```diff
@@ -387,9 +392,41 @@ class FilterConfigProviderManagerImplBase : Logger::Loggable<Logger::Id::filter>
...
+          getConfigDumpType(), [this](const Matchers::StringMatcher& name_matcher) {
+            return dumpEcdsFilterConfigs(name_matcher);
+          });
+    }
+  }
+
+  ProtobufTypes::MessagePtr dumpEcdsFilterConfigs(const Matchers::StringMatcher& name_matcher) {
+    auto config_dump = std::make_unique<envoy::admin::v3::EcdsConfigDump>();
+    for (const auto& subscription : subscriptions_) {
+      const auto& ecds_filter = subscription.second.lock();
```
kyessenov, line 410: Maybe check that weak_ptr lock is non-null for defensive programming?

test/integration/extension_discovery_integration_test.cc
```diff
@@ -775,5 +855,69 @@ TEST_P(ExtensionDiscoveryIntegrationTest, ReloadBoth) {
...
+  size_t index = 0;
+  const std::string expected_types[] = {
+      "type.googleapis.com/envoy.admin.v3.BootstrapConfigDump",
+      "type.googleapis.com/envoy.admin.v3.ClustersConfigDump",
+      "type.googleapis.com/envoy.admin.v3.EcdsConfigDump", // HTTP
+      "type.googleapis.com/envoy.admin.v3.EcdsConfigDump", // TCP Listener
+      "type.googleapis.com/envoy.admin.v3.ListenersConfigDump",
+      "type.googleapis.com/envoy.admin.v3.ScopedRoutesConfigDump",
+      "type.googleapis.com/envoy.admin.v3.RoutesConfigDump",
+      "type.googleapis.com/envoy.admin.v3.SecretsConfigDump"};
```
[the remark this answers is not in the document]
yanjunxiang-google: The thing is that ECDS config dump is only setup after ECDS filter is actually configured, which is here.  And we don't want to duplicate the ECDS filter configurations into other generic files as mentioned.

test/integration/extension_discovery_integration_test.cc
```diff
@@ -775,5 +855,69 @@ TEST_P(ExtensionDiscoveryIntegrationTest, ReloadBoth) {
   codec_client_->close();
 }
 
+TEST_P(ExtensionDiscoveryIntegrationTest, BasicSuccessConfigDump) {
```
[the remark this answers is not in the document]
yanjunxiang-google: This test, "BasicSuccessConfigDump", has only one Listener ECDS filter configured, which is to test the basic case.

"TwoSubscriptionsDifferentNameWithConfigDump" has two listener filter configured. This test can give us a little bit more assurance that the function dumpEcdsFilterConfigs() can handle multiple ECDS filtersconfigured.

Sorry, maybe I don't fully understand the comments.  Does this answer your question?

test/integration/extension_discovery_integration_test.cc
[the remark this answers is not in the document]
adisuissa: Hmmm... 
Ok, so why is the ECDS test checking for "ClustersConfigDump"?
Will it make sense to only check that the 2 "Ecds" types are part of the config dump?

test/integration/extension_discovery_integration_test.cc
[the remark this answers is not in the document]
adisuissa: Maybe I'm misunderstanding something here.
I see the use of `sendXdsResponse2` which I assumed was for a second stream, no?

test/integration/extension_discovery_integration_test.cc
```diff
@@ -264,6 +306,24 @@ class ExtensionDiscoveryIntegrationTest : public Grpc::GrpcClientIntegrationPara
     ecds_stream_->sendGrpcMessage(response);
   }
 
+  void sendXdsResponse2(const std::string& name, const std::string& version,
```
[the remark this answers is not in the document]
yanjunxiang-google: Ok, let me change the name into sendListenerFilterEcdsResponse(). These two functions have significantly different parameters. Let's just keep them separated.

test/integration/extension_discovery_integration_test.cc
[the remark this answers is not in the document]
yanjunxiang-google: Yes, we can do something like below to filter out others to only keep ECDS config dump.
EXPECT_EQ("200", request("admin", "GET", "/config_dump?resource=ecds_filters", response));

However, I feel the full dump can give us the idea where and how these ECDS config dumps are stored and dumped.  BTW, these configure dumps are dumped in the order of the keys associated with it.  For example, we intentionally set the keys to something like "ecds_filter_http", ecds_filter_tcp_listener_filter" to keep them together. If we put them as "http_ecds" and "listener_ecds", they will be spreaded around with other type of config dumps, and maybe hard to find.

source/common/filter/config_discovery_impl.h
yanjunxiang-google: A couple of issue if we reuse statPrefix as the config dump key:

1) Is that naming confusing?
2) Upstream UpstreamHttpFilterConfigProviderImpl has identical stats prefix with HttpFilterConfigProviderImpl. If we go that route, we need to modify them to avoid one config dump override the other.
3)  stats prefix are named as "http", or "tcp", if we use them as config dump keys, it will cause ECDS config dump spreaded around and mixed with other types since config dumps are dumped in the alphabetic order of the keys.

test/integration/extension_discovery_integration_test.cc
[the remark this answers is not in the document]
yanjunxiang-google: Oh, sorry, I miss read the comments.

Anyway, the tests are written for below purpose:

1) The BasicSuccenssConfigDump in listener_extensions_discovery_integration_test.cc , covers the basic  ECDS filter config dump case, only one ECDS filter is configured. 

2)  The TwoSubscriptionsDifferentNameWithConfigDump in listener_extensions_discovery_integration_test.cc , covers the config dump case for two  ECDS filters with same type .

3) The BasicSuccenssConfigDump in extensions_discovery_integration_test.cc , covers the config dump test for two ECDS filters with different type case: one is HTTP, the other is Listener filter.

source/common/filter/config_discovery_impl.h
yanjunxiang-google: Good catched. Added a check.

source/common/filter/config_discovery_impl.h
yanjunxiang-google: Good idea. Done!

source/common/filter/config_discovery_impl.h
yanjunxiang-google: Yeah, filed an issue to track 2: https://github.com/envoyproxy/envoy/issues/24010

kyessenov (commented):
Code LGTM, please wait for other to approve.

yanjunxiang-google:
> Code LGTM, please wait for @adisuissa to approve.

Thanks @kyessenov !

I pushed one more commits to add the ECDS config_dump documentation. Also added another test to verify ECDS config_dump works well with filter name based regex lookup.  The change is small.   PTAL!

adisuissa (commented):
Thanks for working on this! Left a minor comment on the test and one nit, but otherwise LGTM.

test/integration/extension_discovery_integration_test.cc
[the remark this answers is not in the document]
yanjunxiang-google: Thanks for the comments.

Let me change the test name of (3) to reflect it's testing two ECDS types (HTTP filter and Listener filter).  The reason we need this test is that we what to see if both types of ECDS filters are configured, the config dump actually looks like below, i.e, two ECDS filters are dumped back-to-back after "cluster" and consecutively:

  // * ``bootstrap``: :ref:`BootstrapConfigDump <envoy_v3_api_msg_admin.v3.BootstrapConfigDump>`
  // * ``clusters``: :ref:`ClustersConfigDump <envoy_v3_api_msg_admin.v3.ClustersConfigDump>`
  // * ``ecds_filter_http``: :ref:`EcdsConfigDump <envoy_v3_api_msg_admin.v3.EcdsConfigDump>`
  // * ``ecds_filter_tcp_listener``: :ref:`EcdsConfigDump <envoy_v3_api_msg_admin.v3.EcdsConfigDump>`
  // * ``endpoints``:  :ref:`EndpointsConfigDump <envoy_v3_api_msg_admin.v3.EndpointsConfigDump>`
  // * ``listeners``: :ref:`ListenersConfigDump <envoy_v3_api_msg_admin.v3.ListenersConfigDump>`
  // * ``scoped_routes``: :ref:`ScopedRoutesConfigDump <envoy_v3_api_msg_admin.v3.ScopedRoutesConfigDump>`

If we split these in two separate tests, like one for HTTP, one for ECDS, we will only see one EcdsConfigDump here in each test.  Please check the added documentation below:

 // Currently ECDS is supported in HTTP and listener filters. Note, ECDS configuration for
  // either HTTP or listener filter will only be dumped if it is actually configured.
  
For 2), it is test with only one ECDS filter typed configured(here is Listener, but it's same for HTTP), but there are multiple such filters, the EcdsConfigDump are working fine. So what we saw one EcdsConfigDump entry like below in Gete /config_dump. But inside EcdsConfigDump, there are two filters got dumped.

  // * ``bootstrap``: :ref:`BootstrapConfigDump <envoy_v3_api_msg_admin.v3.BootstrapConfigDump>`
  // * ``clusters``: :ref:`ClustersConfigDump <envoy_v3_api_msg_admin.v3.ClustersConfigDump>`
  // * ``ecds_filter_tcp_listener``: :ref:`EcdsConfigDump <envoy_v3_api_msg_admin.v3.EcdsConfigDump>`
  // * ``endpoints``:  :ref:`EndpointsConfigDump <envoy_v3_api_msg_admin.v3.EndpointsConfigDump>`
  // * ``listeners``: :ref:`ListenersConfigDump <envoy_v3_api_msg_admin.v3.ListenersConfigDump>`
  // * ``scoped_routes``: :ref:`ScopedRoutesConfigDump <envoy_v3_api_msg_admin.v3.ScopedRoutesConfigDump>`

This is to verify the for loop in the added function dumpEcdsFilterConfigs() work as expected.

Also, the reason we add the two ECDS types test in extension_discovery_integration_test.cc, not in listener _extension_discovery_integration_test.cc is that it's much easier to add a listener ECDS filter configuration into a test already has HCM and HTTP ECDS filter, which is the case of extension_discovery_integration_test.cc . And it needs much bigger changes to add HCM and HTTP ECDS filter into a test has Listener ECDS filter (listener _extension_discovery_integration_test.cc).

source/common/filter/config_discovery_impl.h
```diff
@@ -387,9 +392,43 @@ class FilterConfigProviderManagerImplBase : Logger::Loggable<Logger::Id::filter>
                                          absl::string_view type_url) const;
   void validateProtoConfigTypeUrl(const std::string& type_url,
                                   const absl::flat_hash_set<std::string>& require_type_urls) const;
+  virtual const std::string getConfigDumpType() const PURE;
```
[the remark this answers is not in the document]
yanjunxiang-google: done

test/integration/extension_discovery_integration_test.cc
```diff
@@ -857,6 +857,7 @@ TEST_P(ExtensionDiscoveryIntegrationTest, ReloadBoth) {
 
 // ECDS config dump test with one listener ECDS filter and one HTTP ECDS filter.
 TEST_P(ExtensionDiscoveryIntegrationTest, ConfigDumpWithTwoSubscriptionTypes) {
+  DISABLE_IF_ADMIN_DISABLED; // Uses admin interface.
```
yanjunxiang-google, line 860: With https://github.com/envoyproxy/envoy/pull/23903, the ADMIN interface is disabled in compile_time_options CI build. Have to disable these tests in this case since config_dump requires admin interface.

yanjunxiang-google:
@envoyproxy/api-shepherds  could you please approve this PR?

api/envoy/admin/v3/config_dump.proto
```diff
@@ -32,6 +32,8 @@ message ConfigDump {
   //
   // * ``bootstrap``: :ref:`BootstrapConfigDump <envoy_v3_api_msg_admin.v3.BootstrapConfigDump>`
   // * ``clusters``: :ref:`ClustersConfigDump <envoy_v3_api_msg_admin.v3.ClustersConfigDump>`
+  // * ``ecds_filter_http``: :ref:`EcdsConfigDump <envoy_v3_api_msg_admin.v3.EcdsConfigDump>`
+  // * ``ecds_filter_tcp_listener``: :ref:`EcdsConfigDump <envoy_v3_api_msg_admin.v3.EcdsConfigDump>`
```
adisuissa, line 36: Should `ecds_filter_upstream_http` and `ecds_filter_udp_listener` also be added here, or is it because they are not plumbed at the moment?

yanjunxiang-google: That's a good question.  ecds_filter_udp_listener is not plumbed yet so we should leave it out.

ecds_filter_upstream_http leveraged most of the ecds_filter_http code, so theoretically it should work automatically.  However, I don't see any Upstream HTTP filter related test code thus guess it is WIP.  Let's leave it out until the feature is ready and related tests are in place.

yanjunxiang-google:
@envoyproxy/api-shepherds Any comments on the API? Otherwise, could you please approve?

test/integration/extension_discovery_integration_test.cc
```diff
@@ -215,13 +259,11 @@ class ExtensionDiscoveryIntegrationTest : public Grpc::GrpcClientIntegrationPara
...
     sendLdsResponse("initial");
 
-    // Wait for ECDS stream.
-    auto& ecds_upstream = getEcdsFakeUpstream();
-    result = ecds_upstream.waitForHttpConnection(*dispatcher_, ecds_connection_);
-    RELEASE_ASSERT(result, result.message());
-    result = ecds_connection_->waitForNewStream(*dispatcher_, ecds_stream_);
-    RELEASE_ASSERT(result, result.message());
-    ecds_stream_->startGrpcStream();
+    waitForEcdsStream(getEcdsFakeUpstream(), ecds_connection_, ecds_stream_);
```
adisuissa, line 262: Previously, for LEDS, I remember I had a race between the 2 incoming ECDS streams. When registering for 2 resources, sometimes resource-A was received on stream-1, and sometimes resource-B was received on stream-1. Did you encounter something similar on the test that uses 2 streams?

One way to check (though not guaranteed) is to run the test 1000 times and see if it is flaky.

test/integration/extension_discovery_integration_test.cc
```diff
@@ -775,5 +855,79 @@ TEST_P(ExtensionDiscoveryIntegrationTest, ReloadBoth) {
...
+  initialize();
+
+  test_server_->waitForCounterGe("listener_manager.lds.update_success", 1);
+  EXPECT_EQ(test_server_->server().initManager().state(), Init::Manager::State::Initializing);
+  registerTestServerPorts({"http"});
+
+  // Send configuration update for HTTP ECDS filter.
+  sendXdsResponse("foo", "1", denyPrivateConfig());
+  // Send configuration update for listener ECDS filter.
+  sendListenerFilterEcdsResponse("bar", "2", 7);
```
adisuissa, line 876: Are they both sending xDS responses, but one on stream1 and the other on stream2?
If so, I suggest rewriting the method names/generalizing them, so it will be more clear.

adisuissa (commented):
Thanks!
/lgtm api

test/integration/extension_discovery_integration_test.cc
yanjunxiang-google: In this case, the the two ECDS subscription are on two completely different connections. Each subscription has it's own cluster and upstream. I guess that's the reason it doesn't race.  BTW the logic is also in https://github.com/envoyproxy/envoy/blob/dede92e6e5888cdf27320bf3b6bc9987657d6e22/test/integration/listener_extension_discovery_integration_test.cc#L130, which had been in code base for a while.

I run the 1000 tests anyway, and doesn't flaky.

test/integration/extension_discovery_integration_test.cc
yanjunxiang-google: done

kyessenov (approved):

alyssawilk:
I believe this was the other half of breaking coverage for main
Again I'd be inclined to roll back and roll forward when it has sufficient testing of new code
https://github.com/envoyproxy/envoy/pull/24320

yanjunxiang-google:
> I believe this was the other half of breaking coverage for main Again I'd be inclined to roll back and roll forward when it has sufficient testing of new code #24320

When working on #23902 , we found the code : UpstreamHttpFilterConfigProviderManagerImpl in config_discovery_impl.h is not tested, 
https://github.com/envoyproxy/envoy/blob/5b9183f80d6c7ab38877b8f8f48b152509f49d3e/source/common/filter/config_discovery_impl.h#L578

which is causing the test coverage in that file is low.

alyssawilk:
Coverage passed on main before your PR, and your PR made additions to a file.  I have never seen this then fail without the reason being your new code lowered overall coverage.  I'm sorry about this mix up and happy to point you at the coverage results when you try to reinstate it and help you sort out how to do it.  The greater problem was this PR was unintentionally merged without the correct approvals so we have to fix that up as well.

alyssawilk:
yeah looks like you were penalized for 2 comments (stupid sorry) but also didn't test an early return: https://storage.googleapis.com/envoy-postsubmit/main/coverage/source/common/filter/config_discovery_impl.h.gcov.html

yanjunxiang-google:
Hi, @alyssawilk , I quickly created this PR to address the test flaky issue in config_discovery_impl.h: https://github.com/envoyproxy/envoy/pull/24362

Issue #23823, opened 2022-11-03
allow compiling out the admin code


alyssawilk:
Largely for envoy mobile, where it cuts down on code footprint, but also solves some cross platform issues with the profiler code and tcmalloc deps

cc @envoyproxy/envoy-mobile-maintainers

PR #23903, opened 2022-11-08
build: making admin optional

alyssawilk:
Adding a build option to compile out Envoy's admin code.
The main API change is making the admin APIs OptRef instead of explicit references, which requires update to any code which adds config trackers for the admin console.

Risk Level: medium
Testing: new CI build with new flags
Docs Changes: n/a
Release Notes: not currently included
Platform Specific Features:
Fixes https://github.com/envoyproxy/envoy/issues/23823


source/extensions/stat_sinks/hystrix/hystrix.cc
```diff
@@ -283,11 +283,12 @@ HystrixSink::HystrixSink(Server::Configuration::ServerFactoryContext& server,
       upstream_rq_2xx_(stat_name_pool_.add("upstream_rq_2xx")),
       upstream_rq_4xx_(stat_name_pool_.add("upstream_rq_4xx")),
       upstream_rq_5xx_(stat_name_pool_.add("upstream_rq_5xx")) {
-  Server::Admin& admin = server_.admin();
-  ENVOY_LOG(debug,
-            "adding hystrix_event_stream endpoint to enable connection to hystrix dashboard");
-  admin.addHandler("/hystrix_event_stream", "send hystrix event stream",
-                   MAKE_ADMIN_HANDLER(handlerHystrixEventStream), false, false);
+  if (server.admin().has_value()) {
```
RyanTheOptimist, line 286: nit: early return

envoy/server/admin.h
```diff
@@ -279,6 +279,11 @@ class Admin {
    */
   static RequestPtr makeStaticTextRequest(absl::string_view response_text, Http::Code code);
   static RequestPtr makeStaticTextRequest(Buffer::Instance& response_text, Http::Code code);
+
+  /**
+   * Closes the listening socket for the admin.
+   */
+  virtual void closeSocket() PURE;
```
RyanTheOptimist, line 286: I can't seem to figure out why this new pure virtual method is needed. Oh is it because we're using `std::unique_ptr<Admin> admin_` instead of `std::unique_ptr<AdminImpl> admin_` in server.h? Do we need to make that change?

source/server/config_validation/server.h
```diff
@@ -71,7 +70,9 @@ class ValidationInstance final : Logger::Loggable<Logger::Id::main>,
                      Filesystem::Instance& file_system);
 
   // Server::Instance
-  Admin& admin() override { return *admin_; }
+  OptRef<Admin> admin() override {
+    return makeOptRefFromPtr(reinterpret_cast<Envoy::Server::Admin*>(admin_.get()));
```
RyanTheOptimist, line 74: I'm surprised to see reinterpret_cast here instead of static_cast. Does static_cast work here?

test/extensions/http/header_formatters/preserve_case/preserve_case_formatter_integration_test.cc
```diff
@@ -1,5 +1,7 @@
 #include "envoy/extensions/http/header_formatters/preserve_case/v3/preserve_case.pb.h"
 
+#include "source/extensions/filters/http/common/pass_through_filter.h"
```
RyanTheOptimist, line 3: Out of curiosity, how did you find this to be missing?

RyanTheOptimist (commented):
88 files? My goodness :)

Looks great other than a few nits.

source/exe/main_common.cc
```diff
@@ -186,12 +186,14 @@ bool MainCommonBase::run() {
 
 void MainCommonBase::adminRequest(absl::string_view path_and_query, absl::string_view method,
                                   const AdminRequestFn& handler) {
+  RELEASE_ASSERT(server_->admin().has_value(),
+                 "adminRequest called with admin support compiled out");
```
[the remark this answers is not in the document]
jmarantz: Why would we want to assert at all? Can't we just call the handler with an http error code? Maybe a 501?

test/integration/base_integration_test.h
```diff
@@ -41,6 +41,14 @@
   } while (0)
 #endif
 
+#ifndef ENVOY_ADMIN_FUNCTIONALITY
+#define DISABLE_IF_ADMIN_DISABLED return
```
jmarantz, line 45: nit: RETURN_IF_ADMIN_DISABLED may be more obvious at call-sites. If I read this in the middle of a function I might not know what DISABLE meant.

Also -- why not use this pattern in production code also? It's somewhat nice in a way :)

source/common/config/config_provider_impl.cc
```diff
@@ -50,14 +50,16 @@ bool ConfigSubscriptionInstance::checkAndApplyConfigUpdate(const Protobuf::Messa
...
-ConfigProviderManagerImplBase::ConfigProviderManagerImplBase(Server::Admin& admin,
+ConfigProviderManagerImplBase::ConfigProviderManagerImplBase(OptRef<Server::Admin> admin,
                                                              const std::string& config_name) {
-  config_tracker_entry_ = admin.getConfigTracker().add(
-      config_name,
-      [this](const Matchers::StringMatcher& name_matcher) { return dumpConfigs(name_matcher); });
-  // ConfigTracker keys must be unique. We are asserting that no one has stolen the key
-  // from us, since the returned entry will be nullptr if the key already exists.
-  RELEASE_ASSERT(config_tracker_entry_, "");
+  if (admin.has_value()) {
```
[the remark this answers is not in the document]
jmarantz: Why not just use the ifdef?

Or do you also support (or plan to support) disabling admin at runtime via a runtime or command-line flag while still compiling it in?

source/exe/main_common.cc
```diff
@@ -186,12 +186,15 @@ bool MainCommonBase::run() {
 
 void MainCommonBase::adminRequest(absl::string_view path_and_query, absl::string_view method,
                                   const AdminRequestFn& handler) {
+  ASSERT(server_->admin().has_value(), "adminRequest called with admin support compiled out");
```
jmarantz, line 189: ok having this assert in in debug mode seems OK. But if we are not in debug mode should we return a 501 or something?

Seems like the fall-through case is going to jus not populate response_headers at all which might be confusing.

Alternatively -- I know you are not going to like this, but this is really a C++ API for folks that instantiate main_common themselves and can call methods on it. You could ifdef out the existence of this method entirely so calling it would be a build failure rather than a crash. But it's up to you.

alyssawilk: I don't love it, but I'll take it over a custom error path no one actually wants but I'd otherwise feel obliged to write tests for =P

jmarantz: Great point!

jmarantz (dismissed):

test/integration/base_integration_test.h
jmarantz: sure makes sense.

envoy/server/admin.h
RyanTheOptimist: I believe you, but I don't yet understand how to connect all the dots :) Why does the server need an API pointer now when before it had an impl pointer?

RyanTheOptimist (approved):
LGTM, though I'm still curious about the AdminImpl => Admin change which necessitated the closeSocket method being added to the interface.

alyssawilk:
Before the server had an admin pointer because it knew about the admin class.  With this change we stick the impl behind the "is admin compiled in" flags, so the class tracks the interface (since the impl may be compiled out) and the interface needs close socket.  alternately we could stick close socket behind an ifdef and down cast from the interface to the impl, but this seemed much cleaner to me.

RyanTheOptimist:
> Before the server had an admin pointer because it knew about the admin class. With this change we stick the impl behind the "is admin compiled in" flags, so the class tracks the interface (since the impl may be compiled out) and the interface needs close socket. alternately we could stick close socket behind an ifdef and down cast from the interface to the impl, but this seemed much cleaner to me.

AH! Yeah, ok that makes perfect sense (of course :>). Conditionally compiling would also seem fine but so does this. Thanks for the explanation. Still LGTM!

RyanTheOptimist (dismissed):

RyanTheOptimist (approved):

PR #23904, opened 2022-11-08
Use a named struct instead of a pair in key_value_store_base.h

RyanTheOptimist:
std::pair tends to result in fairly unreadable code so use a named struct instead.

Also make members private instead of public as per the style guide.

Risk Level: Low
Testing: Existing tests
Docs Changes: N/A
Release Notes: N/A


alyssawilk (dismissed):

alyssawilk (approved):
hah!  I was surprised it didn't require test updates

Issue #23575, opened 2022-10-19
[stats]: lazy init stats to save RAM (and CPU)


stevenzzzz:
*Title*: stats: lazy init stats to save RAM (and CPU)

*Description*:

With lots of clusters and route-tables in a cloud proxy, we are seeing tons of RAM been spent on stats while most of the stats are never inc-ed due to traffic pattern(or long tail). We are thinking that we can lazy init cluster stats() so that the RAM is only allocated when it's required.

To achieve that we need to have finer grained stats group, e.g. configUpdateStats() are frequently updated by config management server, while upstream_xxx are only required when there is traffic for the cluster, for this sub-group we can save RAM by lazy init it.

Here is an example of cluster stats, when there is no traffic, Envoy is still paying the RAM cost to hold all the 0 stats, as well as the CPU that's burnt on collecting such stats.

![6H22NSL5KWcwJH8](https://user-images.githubusercontent.com/3444556/196794847-bfe7e26c-21e0-46f4-800d-8fc4fa177245.png)

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

stevenzzzz:
cc @jmarantz

stevenzzzz:
This is the subgroups I plan to have for clusters.

/**

All cluster config update related stats.
*/
#define ALL_CLUSTER_CONFIG_UPDATE_STATS(COUNTER, GAUGE, HISTOGRAM,
TEXT_READOUT, STATNAME)
COUNTER(assignment_stale)
COUNTER(assignment_timeout_received)
COUNTER(update_attempt)
COUNTER(update_empty)
COUNTER(update_failure)
COUNTER(update_no_rebuild)
COUNTER(update_success)
GAUGE(version, NeverImport)
/**

All cluster endpoints related stats.
*/
#define ALL_CLUSTER_ENDPOINT_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT,
STATNAME)
GAUGE(max_host_weight, NeverImport)
COUNTER(membership_change)
GAUGE(membership_degraded, NeverImport)
GAUGE(membership_excluded, NeverImport)
GAUGE(membership_healthy, NeverImport)
GAUGE(membership_total, NeverImport)
/**

All cluster loadbalancing related stats.
*/
#define ALL_CLUSTER_LB_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT,
STATNAME)
[...]

PR #23523, opened 2022-10-17
Subgroup cluster stats so that later we can lazy init some of the subgroups to save RAM/GCU.

stevenzzzz:
Commit Message: split the per-cluster stats into subgroups, so that later we can lazy-init them to save RAM. 

Additional Description:
See more description and live example in https://github.com/envoyproxy/envoy/issues/23575.

With lots of clusters and route-tables in a cloud proxy, we are seeing tons of RAM been spent on stats while most of the stats are never inc-ed due to traffic pattern(or long tail). We are thinking that we can lazy init cluster stats() so that the RAM is only allocated when it's required. 

To achieve that we need to have finer grained stats group, e.g. configUpdateStats() are frequently updated by config management server, while upstream_xxx are only required when there is traffic for the cluster, for this sub-group we can save RAM by lazy init it.  

This is the subgroups I plan to have for clusters. 

/**
 * All cluster config update related stats.
 */
#define ALL_CLUSTER_CONFIG_UPDATE_STATS(COUNTER, GAUGE, HISTOGRAM, \
                                        TEXT_READOUT, STATNAME)    \
  COUNTER(assignment_stale)                                        \
  COUNTER(assignment_timeout_received)                             \
  COUNTER(update_attempt)                                          \
  COUNTER(update_empty)                                            \
  COUNTER(update_failure)                                          \
  COUNTER(update_no_rebuild)                                       \
  COUNTER(update_success)                                          \
  GAUGE(version, NeverImport)

/**
 * All cluster endpoints related stats.
 */
#define ALL_CLUSTER_ENDPOINT_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT, \
                                   STATNAME)                                \
  GAUGE(max_host_weight, NeverImport)                                       \
  COUNTER(membership_change)                                                \
  GAUGE(membership_degraded, NeverImport)                                   \
  GAUGE(membership_excluded, NeverImport)                                   \
  GAUGE(membership_healthy, NeverImport)                                    \
  GAUGE(membership_total, NeverImport)

/**
 * All cluster loadbalancing related stats.
 */
#define ALL_CLUSTER_LB_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT, \
                             STATNAME)                                \
  COUNTER(lb_healthy_panic)                                           \
  COUNTER(lb_local_cluster_not_ok)                                    \
  COUNTER(lb_recalculate_zone_structures)                             \
  COUNTER(lb_subsets_created)                                         \
[...]

PR #23907, opened 2022-11-08
[stats]: Subgroup cluster stats() into lb-stats, endpoint-stats, config-update-stats and "the rest"(will be renamed to upstream-stats)

stevenzzzz:
Commit Message: Subgroup cluster stats() into lb-stats, endpoint-stats, config-update-stats and "the rest"(will be renamed to upstream-stats)
Additional Description:
See more description and live example in https://github.com/envoyproxy/envoy/issues/23575.

With lots of clusters and route-tables in a cloud proxy, we are seeing tons of RAM been spent on stats while most of the stats are never inc-ed due to traffic pattern(or long tail). We are thinking that we can lazy init cluster stats() so that the RAM is only allocated when it's required.

To achieve that we need to have finer grained stats group, e.g. configUpdateStats() are frequently updated by config management server, while upstream_xxx are only required when there is traffic for the cluster, for this sub-group we can save RAM by lazy init it.

This is the subgroups I plan to have for clusters.

/**

All cluster config update related stats.
*/
#define ALL_CLUSTER_CONFIG_UPDATE_STATS(COUNTER, GAUGE, HISTOGRAM,
TEXT_READOUT, STATNAME)
COUNTER(assignment_stale)
COUNTER(assignment_timeout_received)
COUNTER(update_attempt)
COUNTER(update_empty)
COUNTER(update_failure)
COUNTER(update_no_rebuild)
COUNTER(update_success)
GAUGE(version, NeverImport)
/**

All cluster endpoints related stats.
*/
#define ALL_CLUSTER_ENDPOINT_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT,
STATNAME)
GAUGE(max_host_weight, NeverImport)
COUNTER(membership_change)
GAUGE(membership_degraded, NeverImport)
GAUGE(membership_excluded, NeverImport)
GAUGE(membership_healthy, NeverImport)
GAUGE(membership_total, NeverImport)
/**

All cluster loadbalancing related stats.
*/
#define ALL_CLUSTER_LB_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT,
STATNAME)
COUNTER(lb_healthy_panic)
COUNTER(lb_local_cluster_not_ok)
COUNTER(lb_recalculate_zone_structures)
COUNTER(lb_subsets_created)
COUNTER(lb_subsets_fallback)
COUNTER(lb_subsets_fallback_panic)
COUNTER(lb_subsets_removed)
COUNTER(lb_subsets_selected)
COUNTER(lb_zone_cluster_too_small)
COUNTER(lb_zone_no_capacity_left)
COUNTER(lb_zone_number_differs)
COUNTER(lb_zone_routing_all_directly)
COUNTER(lb_zone_routing_cross_zone)
COUNTER(lb_zone_routing_sampled)
GAUGE(lb_subsets_active, Accumulate)
/**

All cluster stats. https://github.com/see stats_macros.h
*/
#define ALL_CLUSTER_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT, STATNAME)
COUNTER(bind_errors)
COUNTER(original_dst_host_invalid)
COUNTER(retry_or_shadow_abandoned)
COUNTER(upstream_cx_close_notify)
COUNTER(upstream_cx_connect_attempts_exceeded)
COUNTER(upstream_cx_connect_fail)
COUNTER(upstream_cx_connect_timeout)
COUNTER(upstream_cx_connect_with_0_rtt)
COUNTER(upstream_cx_destroy)
COUNTER(upstream_cx_destroy_local)
COUNTER(upstream_cx_destroy_local_with_active_rq)
COUNTER(upstream_cx_destroy_remote)
COUNTER(upstream_cx_destroy_remote_with_active_rq)
COUNTER(upstream_cx_destroy_with_active_rq)
COUNTER(upstream_cx_http1_total)
COUNTER(upstream_cx_http2_total)
COUNTER(upstream_cx_http3_total)
COUNTER(upstream_cx_idle_timeout)
COUNTER(upstream_cx_max_duration_reached)
COUNTER(upstream_cx_max_requests)
COUNTER(upstream_cx_none_healthy)
COUNTER(upstream_cx_overflow)
COUNTER(upstream_cx_pool_overflow)
COUNTER(upstream_cx_protocol_error)
COUNTER(upstream_cx_rx_bytes_total)
COUNTER(upstream_cx_total)
COUNTER(upstream_cx_tx_bytes_total)
COUNTER(upstream_flow_control_backed_up_total)
COUNTER(upstream_flow_control_drained_total)
COUNTER(upstream_flow_control_paused_reading_total)
COUNTER(upstream_flow_control_resumed_reading_total)
COUNTER(upstream_internal_redirect_failed_total)
COUNTER(upstream_internal_redirect_succeeded_total)
COUNTER(upstream_rq_cancelled)
COUNTER(upstream_rq_completed)
COUNTER(upstream_rq_maintenance_mode)
COUNTER(upstream_rq_max_duration_reached)
COUNTER(upstream_rq_pending_failure_eject)
COUNTER(upstream_rq_pending_overflow)
COUNTER(upstream_rq_pending_total)
COUNTER(upstream_rq_0rtt)
COUNTER(upstream_rq_per_try_timeout)
COUNTER(upstream_rq_per_try_idle_timeout)
COUNTER(upstream_rq_retry)
COUNTER(upstream_rq_retry_backoff_exponential)
COUNTER(upstream_rq_retry_backoff_ratelimited)
COUNTER(upstream_rq_retry_limit_exceeded)
COUNTER(upstream_rq_retry_overflow)
COUNTER(upstream_rq_retry_success)
COUNTER(upstream_rq_rx_reset)
COUNTER(upstream_rq_timeout)
COUNTER(upstream_rq_total)
COUNTER(upstream_rq_tx_reset)
COUNTER(upstream_http3_broken)
GAUGE(upstream_cx_active, Accumulate)
GAUGE(upstream_cx_rx_bytes_buffered, Accumulate)
GAUGE(upstream_cx_tx_bytes_buffered, Accumulate)
GAUGE(upstream_rq_active, Accumulate)
GAUGE(upstream_rq_pending_active, Accumulate)
HISTOGRAM(upstream_cx_connect_ms, Milliseconds)
HISTOGRAM(upstream_cx_length_ms, Milliseconds)

Risk Level: LOW, should be a no-behavior-change CL.

Testing: N/A existing stats test should prove that there is no behavior change.
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


envoy/upstream/upstream.h
```diff
@@ -565,12 +565,39 @@ class PrioritySet {
 };
 
 /**
- * All cluster stats. @see stats_macros.h
+ * All cluster config update related stats.
+ * See https://github.com/envoyproxy/envoy/issues/23575 for details. Stats from ClusterInfo::stats()
+ * will be split into subgroups "config-update", "lb", "endpoint" and "upstream", roughly based on
```
jmarantz, line 570: This comment should describe the current state of affairs assuming this PR is checked in, not express a plan.

AFAICT stats_ no longer exists for upstream; it's now fully partitioned so we can just document that.

envoy/upstream/upstream.h
```diff
@@ -585,14 +612,15 @@ class PrioritySet {
   COUNTER(lb_zone_routing_all_directly)                                                            \
   COUNTER(lb_zone_routing_cross_zone)                                                              \
   COUNTER(lb_zone_routing_sampled)                                                                 \
-  COUNTER(membership_change)                                                                       \
+  GAUGE(lb_subsets_active, Accumulate)
+
+/**
+ * All cluster stats. @see stats_macros.h
+ */
+#define ALL_CLUSTER_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT, STATNAME)                       \
```
jmarantz, line 620: I think maybe we should rename this macro and related call-sites because this does not represent all the stats anymore.

envoy/upstream/upstream.h
```diff
@@ -991,6 +1031,21 @@ class ClusterInfo : public Http::FilterChainFactory {
    */
   virtual TransportSocketMatcher& transportSocketMatcher() const PURE;
 
+  /**
+   * @return ClusterConfigUpdateStats& strongly named config update stats for this cluster.
```
jmarantz, line 1035: I realize this phrase 'strongly named' pre-existed, but I'm not sure what it means.

It's possible I came up with phrase and i still don't know what it means :)

Do you? If not, can we just remove this here & below?

For ClusterLbStats maybe use 'load-balancer-related stats for this cluster'.

source/common/upstream/cluster_manager_impl.h
```diff
@@ -326,6 +326,13 @@ class ClusterManagerImpl : public ClusterManager,
   initializeSecondaryClusters(const envoy::config::bootstrap::v3::Bootstrap& bootstrap) override;
 
   const ClusterStatNames& clusterStatNames() const override { return cluster_stat_names_; }
```
jmarantz, line 328: do we still use ClusterStatNames anywhere?

jmarantz (commented):
looks good basically; just a few nits.

Can you merge main?

envoy/upstream/upstream.h
stevenzzzz: I did that in another CL. See https://github.com/envoyproxy/envoy/pull/23921/files

envoy/upstream/upstream.h
stevenzzzz: I decided to rename that field in the next PR. 
I think the comment here is kinda accurate, updated the "upstream" one to reflect the "current" status.

envoy/upstream/upstream.h
jmarantz: can we focus on just making this one coherent? and getting it in?

envoy/upstream/upstream.h
stevenzzzz: Done.

source/common/upstream/cluster_manager_impl.h
stevenzzzz: I am not sure. 
I saw it only used to instantiate the ClusterStats in upstream_impl.cc.

stevenzzzz (commented):
Thanks! PTAL.

jmarantz:
confused about how this stacks on the other PR -- this works on its own and appears coherent, right? Also that one is a draft and this one looks close to being ready.

stevenzzzz:
@jmarantz let's commit this one and ignore the other one in the chain.  (https://github.com/envoyproxy/envoy/pull/23523)

source/common/upstream/cluster_manager_impl.h
jmarantz: can you comment it out and see what breaks?

I'd like this PR to just be a repartitioning without adding any overlap.

test/common/upstream/load_balancer_fuzz_base.h
```diff
@@ -42,8 +41,8 @@ class LoadBalancerFuzzBase {
   // These public objects shared amongst all types of load balancers will be used to construct load
   // balancers in specific load balancer fuzz classes
   Stats::IsolatedStoreImpl stats_store_;
-  ClusterStatNames stat_names_;
-  ClusterStats stats_;
+  ClusterLbStatNames lb_stat_names_;
```
jmarantz, line 44: hm. I guess you didn't need to change the variable name from stat_names_ to lb_stat_names_ here. You could have just left it at stat_names_ (still changing the type. Ditto for lb_stats_. That would make this PR a lot smaller I think.

stevenzzzz: IIUC, there is a statsname for every StatStruct. So we have now 4 {Lb, Endpoint, ConfigUpdate, }StatNames definitions.

test/common/upstream/load_balancer_impl_test.cc
```diff
@@ -1959,14 +1960,12 @@ TEST_P(LeastRequestLoadBalancerTest, SingleHost) {
   // Host weight is 1.
   {
     EXPECT_CALL(random_, random()).WillOnce(Return(0)).WillOnce(Return(2)).WillOnce(Return(3));
-    stats_.max_host_weight_.set(1UL);
```
jmarantz, line 1962: is this change related?

BTW it's fine to set stats in a test. It's only in prod that I have a concern. In any case this change probably should go in a different PR, no?

test/common/upstream/zone_aware_load_balancer_fuzz_base.cc
```diff
@@ -48,7 +48,6 @@ void ZoneAwareLoadBalancerFuzzBase::setupZoneAwareLoadBalancingSpecificLogic() {
   // Having 3 possible weights, 1, 2, and 3 to provide the state space at least some variation
   // in regards to weights, which do affect the load balancing algorithm. Cap the amount of
   // weights at 3 for simplicity's sake
-  stats_.max_host_weight_.set(3UL);
```
jmarantz, line 51: here too.

Actually if the reason you woke to this is that (a) these stats were no longer accessible from this point and so coudn't be written and (b) that had no impact on the tests, then maybe send just the deletion of these max_host_weight_set() calls into a separate PR? That should be easy to review (deleting a bunch of superfluous test lines).

One concern I might have is if setting these values wakes up some code we need to be testable.

jmarantz (commented):
lgtm modulo minor rename request, plus the superflous changes to set host weight.

thank you!

stevenzzzz (commented):
thanks!

jmarantz:
great. this is OK with me but you might want to throw the removal of stats_.max_host_weight_.set into a separate PR which should be easy to review & merge first. That will avoid having a senior maintainer ask the same question :)

this lgtm to me other than that, and your format fail :)

jmarantz (approved):
Thanks Xin.

Stephen please let me know if you need more context or reach out to the PR author direclty.

envoy/upstream/upstream.h
stevenzzzz: done.

test/common/upstream/zone_aware_load_balancer_fuzz_base.cc
stevenzzzz: removing the max_host_weight_ references in tests with PR: https://github.com/envoyproxy/envoy/pull/23936

jmarantz:
I think we should wait for a non-google maintainer to review. If Stephan can't do it we can get a non-Google non-senior maintainer and then Yan can give final approval.

jmarantz:
Also PTAL @ clang-tidy failure

source/common/upstream/upstream_impl.cc
```diff
@@ -832,9 +832,9 @@ void MainPrioritySetImpl::updateCrossPriorityHostMap(const HostVector& hosts_add
   }
 }
 
-ClusterStats ClusterInfoImpl::generateStats(Stats::Scope& scope,
-                                            const ClusterStatNames& stat_names) {
-  return ClusterStats(stat_names, scope);
+ClusterTrafficStats ClusterInfoImpl::generateStats(Stats::Scope& scope,
+                                                   const ClusterTrafficStatNames& stat_names) {
+  return ClusterTrafficStats(stat_names, scope);
```
yanavlasov, line 837:
```suggestion
  return {stat_names, scope};
```

test/common/upstream/load_balancer_simulation_test.cc
```diff
@@ -107,8 +107,7 @@ TEST(DISABLED_LeastRequestLoadBalancerWeightTest, Weight) {
 class DISABLED_SimulationTest : public testing::Test {
```
yanavlasov, line 107: I know this is not your change, but we need to mollify preexisting clang-tidy

```suggestion
class DISABLED_SimulationTest : public testing::Test { // NOLINT(readability-identifier-naming)
```

yanavlasov (changes requested):
/wait

test/common/upstream/load_balancer_simulation_test.cc
stevenzzzz: Thanks!

stevenzzzz (commented):
thanks!

envoy/upstream/upstream.h
```diff
@@ -565,12 +565,36 @@ class PrioritySet {
...
+  COUNTER(membership_change)                                                                       \
+  GAUGE(membership_degraded, NeverImport)                                                          \
+  GAUGE(membership_excluded, NeverImport)                                                          \
+  GAUGE(membership_healthy, NeverImport)                                                           \
+  GAUGE(membership_total, NeverImport)
+
+/**
+ * All cluster loadbalancing related stats.
+ */
+#define ALL_CLUSTER_LB_STATS(COUNTER, GAUGE, HISTOGRAM, TEXT_READOUT, STATNAME)                    \
```
ggreenway, line 597: I think these should be further sub-divided by functionality. lb_subset, zone_aware, and then all the rest. Both of those features are unused in many configs so they could all be omitted.

ggreenway (changes requested):

envoy/upstream/upstream.h
stevenzzzz: IIUC, the zone_aware stats are the "base" for subset_lb loadbalancers, due to the following: 
```
class EdfLoadBalancerBase : public ZoneAwareLoadBalancerBase
```
so the lb-related stats (subset, and zone-aware) will be created/required at the same time.

ggreenway: But I think the zone-awareness isn't used in many cases, and those stats will remain zero'd (at least I think that's how it works). So if they are separate and lazy-created, I think they won't get created in many configs.

stevenzzzz: our of the 5 lb-impls used in subset_lb, only RingHashLoadBalancer and MagLevLoadBalancer are NOT inherited from zoneawareLoadBalancer. But I see your point there tho. It requires some refactoring work: we'd better instantiate those stats objs from within each lb-object, which triggers even more delta.

do you mind if I do it incrementally?

ggreenway: Incrementally sounds good. Do you want to land this as-is and do it in a subsequent PR?

stevenzzzz: yeah. it'd be much cleaner.

ggreenway (approved):

stevenzzzz:
Thanks for working on my PR!

Issue #23901, opened 2022-11-08
constant view of Buffer:Instance or Buffer::OwnedImpl


JuniorHsu:
*Title*: constant view of `Buffer:Instance` or `Buffer::OwnedImpl`

*Description*:
We have use case to traverse the `Buffer::OwnedImpl` twice but we don't want to copy the buffer since the first traversal is constant operation.

Hence, it is ideal if we're able to traverse the "view" of the buffer at the first time.

To the best of my knowledge we don't have `OwnedImplView` to access the `SliceDeque` under the buffer.

Only way is probably to getRawSlices to `move` out the data and create new `OwnedImpl` to `addBufferFragment` bases on those slices and eventually `move` the "rawSlices" back to the original `OwnedImpl`, which is not handy.
https://github.com/envoyproxy/envoy/blob/e4ae2259add5a327c4d67de18a82df5d78a801b4/envoy/buffer/buffer.h#L237

phlax:
cc @zuercher @antoniovicente

PR #23908, opened 2022-11-09
thrift_proxy: decode passthrough data by protocol parsing only for payload_to_metadata filter

JuniorHsu:
…yload_to_metadata filter

Commit Message:
This PR fixes two issues:
(a) the decoder actually `drain` the buffer so the router couldn't get the data to send.
we copy to buffer to parse for now as we're waiting for https://github.com/envoyproxy/envoy/issues/23901

(b) passthrough data is not the whole thrift request, 
therefore we introduce `DecoderStateMachine::runPassthroughData` to decode the protocol part only.

Additional Description:
Risk Level: low
Testing: unit test
Docs Changes: filter rst

### DIFF
--- a/docs/root/configuration/other_protocols/thrift_filters/payload_to_metadata_filter.rst
+++ b/docs/root/configuration/other_protocols/thrift_filters/payload_to_metadata_filter.rst
@@ -62,7 +62,7 @@ This filter is designed to support payload passthrough. Performing payload to me
 can do deserialization once, and pass the metadata to other filters. This means that load balancing
 decisions, consumed from log and routing could all use payload information with a single parse.
 Also notably performing the parsing in payload passthrough buffer will mean deserialization once
-and not re-serializing, which is the most performant outcome. Currently there is an redundant buffer
+and not re-serializing, which is the most performant outcome. Currently there is a redundant buffer
 copy until we have `BufferView <https://github.com/envoyproxy/envoy/issues/23901>`_.
 
 If any of the filter chain doesn't support payload passthrough, a customized non-passthrough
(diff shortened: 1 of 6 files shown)


source/extensions/filters/network/thrift_proxy/decoder.h
```diff
@@ -88,7 +88,14 @@ class DecoderStateMachine : public Logger::Loggable<Logger::Id::thrift> {
   ProtocolState currentState() const { return state_; }
 
   /**
-   * Set the current state. Used for testing only.
+   * Consumes whole passthrough data without the body start.
```
zuercher, line 91: nit: I think body here is confusing. Maybe "message start" instead of "body start"?

zuercher (approved):
One minor comment on the comment, but otherwise looks good.

source/extensions/filters/network/thrift_proxy/decoder.h
JuniorHsu: nice suggestion. i addressed this. thanks!

docs/root/configuration/other_protocols/thrift_filters/payload_to_metadata_filter.rst
```diff
@@ -62,7 +62,8 @@ This filter is designed to support payload passthrough. Performing payload to me
 can do deserialization once, and pass the metadata to other filters. This means that load balancing
 decisions, consumed from log and routing could all use payload information with a single parse.
 Also notably performing the parsing in payload passthrough buffer will mean deserialization once
-and not re-serializing, which is the most performant outcome.
+and not re-serializing, which is the most performant outcome. Currently there is an redundant buffer
```
rgs1, line 65: nit: a redundant

test/extensions/filters/network/thrift_proxy/filters/payload_to_metadata/payload_to_metadata_filter_test.cc
```diff
@@ -75,8 +90,12 @@ class PayloadToMetadataTest : public testing::Test {
   //     f10: set
   //   }
   // }
-  void writeMessage(MessageMetadata& metadata, Buffer::Instance& buffer,
-                    std::string string_value = "two") {
+  void writeMessage(std::string string_value = "two") {
```
rgs1, line 93: thank you for cleaning this up

rgs1 (approved):
LGTM, just a nit (but you can ignore) too. Thanks!

PR #23913, opened 2022-11-09
[backport/1.21] wasm: update Wasmtime to v0.39.1

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
@PiotrSikora @mattklein123 @kfaseela @ingwonsong 

we have to decide whether to take this PR or not - its pretty invasive for a release patch (updates bazel, toolchains, and a lot of wasm related stuff)

otoh - the branch is failing (multiple iirc) CVEs, which this resolves - and i also had issues just running some of the wasm tests without at least some of the commits added here (not sure why this was only triggering locally for me)

mattklein123:
I don't have an opinion on whether we take it or not. Since WASM is alpha we could also just disable WASM on this branch if it's causing a bunch of issue? Maybe that would be simpler?

/wait

phlax:
mho is that it would be simpler to update than to start disabling things

im also leaning more towards adding this over not - mostly because of the issues running tests on the existing branch - its probably a small risk - but my fear is that a broken test gets triggered when we actually release and stuff fails

phlax:
cc @keith

mattklein123 (dismissed):

phlax:
there is one last error with this in the coverage build

```console
ld.lld: error: main.cc:(.rodata._ZN6spdlog17pattern_formatter12handle_flag_INS_7details13scoped_padderEEEvcNS2_12padding_infoE+0x0): relocation R_X86_64_PC32 out of range: 2827914925 is not in [-2147483648, 2147483647]
```

which i think implies a mismatch in `-fpic` settings somewhere in the build - ill try and track it down by comparing to 1.22

mattklein123 (dismissed):

phlax:
thanks to some help from @PiotrSikora this is now passing pretty confidently

its likely that not all of the commits/changes here are strictly necessary - but it is fairly well ci-tested and i think the selection of changes are not bad in terms of what is required/desirable to update wasm and gives it ~parity with the 1.22 branch

once this (and release PR) land we can publish 1.21.6

phlax:
arggh - this has the flakey diskspace issue now - ill backport _that_ fix ...

phlax:
actually its not diskspace - i think it just timed out - maybe dockerhub issue - but i will increase the time to be sure

mattklein123 (approved):

PR #23915, opened 2022-11-09
[balsa] Fix BalsaParser::isHttp11().

bencebeky:
HTTP/1.1 request line and response line are defined at https://www.rfc-editor.org/rfc/rfc7230.html#section-3.1. HTTP-version is case sensitive.

Note that expectations for codec_->protocol() added to existing Http1ServerConnectionImplTest tests only test request protocol parsing. A new Http1ClientConnectionImplTest test is added for response protocol parsing.

Also remove an unused local variable from
Http1ServerConnectionImplTest.Http10MultipleResponses.

Commit Message: Fix BalsaParser::isHttp11().
Additional Description:
Risk Level: low, changed code protected by existing default-false runtime flag.
Docs Changes: n/a
Release Notes: n/a


diannahu (approved):
This new version is great!

source/common/http/http1/balsa_parser.cc
```diff
@@ -143,7 +143,12 @@ ParserStatus BalsaParser::getStatus() const { return status_; }
 uint16_t BalsaParser::statusCode() const { return headers_.parsed_response_code(); }
 
 bool BalsaParser::isHttp11() const {
-  return absl::EndsWith(headers_.first_line(), Http::Headers::get().ProtocolStrings.Http11String);
+  if (framer_.is_request()) {
+    return absl::EndsWith(headers_.first_line(), Http::Headers::get().ProtocolStrings.Http11String);
+  } else {
+    return absl::StartsWith(headers_.first_line(),
```
birenroy, line 149: Is it worth adding a unit test that exercises this case?

birenroy (approved):
Looks good! Just one question about a Balsa unit test.

source/common/http/http1/balsa_parser.cc
bencebeky: Good question.

Four existing tests fail if I change the request case to `return true`.  18 existing tests (plus two new expectations) fail if I change the request case to `return false`.  Many of the failures are at https://github.com/envoyproxy/envoy/blob/3b8beabfca22af97f9626ed399bfc14e88413137/test/common/http/http1/codec_impl_test.cc#L231.  It seems like both cases are sufficiently covered.

KBaichoo (approved):
Looks good, thank you!

PR #23917, opened 2022-11-09
build: removing forced cares and apple dns deps

alyssawilk:
Making clusters lazily create DNS resolvers so Envoys with, say, only static clusters don't need any DNS dependency.

Risk Level: medium
Testing: n/a
Docs Changes: n/a
Release Notes: inline
Part of https://github.com/envoyproxy/envoy-mobile/issues/2629


alyssawilk:
cc @jmarantz @jpsim

lizan (approved):

Issue #22478, opened 2022-08-01
CI sync job races


phlax:
If 2 PRs are landed in quick succession the "sync" job fails checking out and trying to commit back to the `envoy-filter-example` repo

wrowe:
Do azure pipelines support serializing a build step, across builds? @lizan would you be familiar?

phlax:
> Do azure pipelines support serializing a build step, across builds?

i dont think so

tbh im wondering about the downstream repo using a submodule for this given it has bazel also - sure it should use bazel to pull in the envoy repo

switching that wouldnt resolve by itself - but we could potentially use a PR (that gets updated) or somesuch to update the downstream rather than checkout/commit/push which is prone to racing

another option is adding a retry/rebase if the push fails

phlax:
this has been resolved

PR #23932, opened 2022-11-10
Ci

alyssawilk:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/envoy_mobile_mirror.sh
+++ b/ci/envoy_mobile_mirror.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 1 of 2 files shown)


ci/envoy_mobile_mirror.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -e
```
phlax, line 3: heads up that this pattern already causes a race condition with the filter example (#22478 )

PR #23933, opened 2022-11-10
feat: add runtime guard to handle 5xx from ext_authz response as denied instead of error

### LINKED ISSUE #14001: ext_authz(http): Envoy turns AuthorizationResponse status 500 into 403 and drops body
*Title*: ext_authz(http): Envoy turns AuthorizationResponse status 500 into 403 and drops body

*Description*:
1. User makes an HTTP request against Envoy.
1. Envoy makes AuthorizationRequest to ext_authz filter's http server.
1. ext_authz server replies with 503 and a body.
1. Envoy returns 403 and no body.

*Repro steps*:
This came up as part of a project so it's easier to include instructions for that project:
```sh
# Clone and checkout branch made for reproducing this issue (see last commit)
git clone git@github.com:arrikto/oidc-authservice.git
cd oidc-authservice
git checkout feature-envoy-ext-authz-repo

# Run the E2E test to get a K3d cluster up with everything running.
# The test will fail, but it's expected.
make bin/deps
make e2e

# Port-forward the service locally
export KUBECONFIG=$HOME/.k3d/kubeconfig-e2e-test-cluster.yaml
kubectl port-forward -n istio-system svc/istio-ingressgateway 8080:80

# Make a request and see that it returns a 403
curl -v http://localhost:8080
```

*Config*:
Here is the config dump taken from port 15000 of the istio ingressgateway proxy:
[config_dump.log](https://github.com/envoyproxy/envoy/files/5532915/config_dump.log)

*Logs*:
[access_log.log](https://github.com/envoyproxy/envoy/files/5532921/access_log.log)

imrenagi:
Signed-off-by: Imre Nagi <imre.nagi2812@gmail.com>

Commit Message: feat: add runtime guard to handle 5xx from ext_authz response as denied instead of error. This is to introduce fix for #14001

Additional Description:
Risk Level:
Testing:
* add several unit tests and integration test. the test is also done manually by using this runtime guard:
```
layered_runtime:
  layers:
  - name: static_layer
    static_layer:
      envoy.reloadable_features.ext_authz_http_service_5xx_is_denied_instead_of_error: true
```
Docs Changes: -
Release Notes: -
Platform Specific Features: -

Fixes #14001 

Runtime guard:
* `envoy.reloadable_features.envoy.reloadable_features.ext_authz_http_service_5xx_is_denied_instead_of_error`. Setting this to `true` will make envoy to return 5xx status code and body from ext_authz server as it is to the client. The status returned will be `CheckStatus::Denied`.

Deprecated:
* By adding the runtime guard, we are trying to deprecate the old behaviour of ext_authz filter when receiving non 200 status code.  Currently when ext_authz server return non 200 status code, it will change status code to 403 (by default) or value specified for `status_on_error` and set the body as empty.


imrenagi:
cc @dio

test/extensions/filters/http/ext_authz/ext_authz_test.cc
```diff
@@ -2042,6 +2042,86 @@ TEST_P(HttpFilterTestParam, DeniedResponseWith403) {
...
+                    .counterFromString("upstream_rq_5xx")
+                    .value());
+  EXPECT_EQ(1U, decoder_filter_callbacks_.clusterInfo()
+                    ->statsScope()
+                    .counterFromString("upstream_rq_500")
+                    .value());
+}
+
+// Test that a error response results in the connection closing with a 403 response to the client.
+TEST_P(HttpFilterTestParam, ErrorWithResponse5xx) {
```
dio, line 2096: Sorry, why do we need this one?

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
```diff
@@ -725,6 +725,32 @@ TEST_P(ExtAuthzHttpIntegrationTest, DefaultCaseSensitiveStringMatcher) {
...
+      fake_upstreams_.back()->waitForHttpConnection(*dispatcher_, fake_ext_authz_connection_);
+  RELEASE_ASSERT(result, result.message());
+  FakeStreamPtr ext_authz_request;
+  result = fake_ext_authz_connection_->waitForNewStream(*dispatcher_, ext_authz_request);
+  RELEASE_ASSERT(result, result.message());
+  result = ext_authz_request->waitForEndStream(*dispatcher_);
+  RELEASE_ASSERT(result, result.message());
+
+  Http::TestResponseHeaderMapImpl ext_authz_response_headers{{":status", "500"}};
+  ext_authz_request->encodeHeaders(ext_authz_response_headers, true);
```
dio, line 746: Can we do `encodeData` as well? Hence we are sure the body from ext_authz service is returned back to the client.

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
```diff
@@ -725,6 +725,32 @@ TEST_P(ExtAuthzHttpIntegrationTest, DefaultCaseSensitiveStringMatcher) {
   ASSERT_TRUE(header_entry.empty());
 }
 
+TEST_P(ExtAuthzHttpIntegrationTest, DeniedResponseWhenAuthzReturn5xx) {
```
dio, line 728: Also, we need to test the old behavior? (sending error, 403)

dio (commented):
Thanks! Flushing my comments:

ggreenway:
Is there any value to retain the old behavior with a config knob (not runtime guard)? Does anyone depend on the old behavior?

test/extensions/filters/http/ext_authz/ext_authz_test.cc
imrenagi: It is to test the old behavior. Since the old behaviour returns `CheckStatus::Error`, I assume that we also need this test here?

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
imrenagi: sure. will add.

test/extensions/filters/http/ext_authz/ext_authz_test.cc
dio: Does this actually call the old behavior?

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
imrenagi: i tried to add this, but I cant make the test works. any pointer that I can follow for this?

dio:
I'm interested to the following statement from @yanniszark:

https://github.com/envoyproxy/envoy/issues/14001#issuecomment-742852892

For example, when the authz service sends 5xx. What do we expect from the client's perspective? Also from debugability PoV.

imrenagi:
@ggreenway not sure whether there are people depending on the old behavior.

@mszlgr @erplsf just wondering whether you are relying on old behavior discussed in #14001?

test/extensions/filters/http/ext_authz/ext_authz_test.cc
imrenagi: let me recheck

ggreenway (changes requested):
I think that changing this behavior by default, especially after it has behaved this way for so long, is risky. I can imagine cases where the auth service returns an error with some data intended for internal-only use, which would now be sent back to the client.

I think this should have a config knob to enable, and leave the default behavior as it currently is.

Also, to be consistent, the new behavior should probably cover the error case of 
```
  const auto thread_local_cluster = cm_.getThreadLocalCluster(cluster);
  if (thread_local_cluster == nullptr) {
    // TODO(dio): Add stats related to this.
    ENVOY_LOG(debug, "ext_authz cluster '{}' does not exist", cluster);
    callbacks_->onComplete(std::make_unique<Response>(errorResponse()));
```

and return a 5xx error in that case.

And the grpc implementation should probably also be made consistent with this new behavior, to the extent possible. The grpc async-client will call `onFailure()` for a non-OK grpc response.

/wait

Issue #23919, opened 2022-11-09
Changing the upstream destination in a filter


rjhuijsman:
*Title*: Changing the upstream destination in a filter

*Description*:
I have requests coming into an Envoy gateway (set up via Istio on Kubernetes). I'd like to write a filter to...
1. Inspect the `host` or `:authority` headers and do some arbitrary business logic (do push-ups, make a charitable donation, [...], you name it - anything that's not a built-in Envoy feature yet).
2. Rewrite the `host` and `:authority` headers to point at a different Kubernetes service than the original destination.
3. Have Envoy's route the request to the rewritten Kubernetes service, rather than to the original.

I've performed that experiment using an [`ExternalProcessor` filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter). I've configured its `mutation_rules` to set `allow_all_routing: true`. I've injected that filter (using Istio) as follows:

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: respect-external-filter
  namespace: istio-system
spec:
  configPatches:
    - applyTo: EXTENSION_CONFIG
      patch:
        operation: ADD
        value:
          name: my-cool-external-filter
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor
            failure_mode_allow: false
            async_mode: false
            request_attributes:
              - user
            response_attributes:
              - server
            # Need to set 'allow_all_routing' so that we can modify the
            # 'host'/':authority' headers.
            # https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/common/mutation_rules/v3/mutation_rules.proto
            mutation_rules:
              allow_all_routing: true
            # Only process request headers.
            processing_mode:
              request_header_mode: "SEND"
              response_header_mode: "SKIP"
              request_body_mode: "NONE"
              response_body_mode: "NONE"
              request_trailer_mode: "SKIP"
              response_trailer_mode: "SKIP"
            grpc_service:
              envoy_grpc:
                cluster_name: "outbound|18080||my-cool-external-filter.default.svc.cluster.local"
    - applyTo: HTTP_FILTER
      match:
        listener:
          filterChain:
            filter:
              name: envoy.filters.network.http_connection_manager
              subFilter:
                name: envoy.filters.http.router
      patch:
        operation: INSERT_BEFORE
        value:
[...]

PR #23937, opened 2022-11-10
Expand comment on `ExternalProcessor.mutation_rules`

rjhuijsman:
This is a comments-changes-only PR.

The `ExternalProcessor` has a `mutation_rules` field, whose purpose is to let users modify otherwise-restricted headers, such as `"host"` and `":authority"`. Odds are that users who are modifying such headers are trying to influence the routing decisions that Envoy makes, such as was the case in https://github.com/envoyproxy/envoy/issues/23919.

However, setting only `"host"` or `":authority"` is not sufficient to change such routing decisions; users must also set a `clear_route_cache` field.

This PR aims to make it harder to overlook the `clear_route_cache` field, like I did in https://github.com/envoyproxy/envoy/issues/23919, by improving the comments on the `mutation_rules` field.


mattklein123 (approved):
Thanks!

PR #23939, opened 2022-11-10
[file_system_http_cache] Add proto helpers

ravenblackx:
The simplest of these helpers exist more to get around the envoy CI constraint that doesn't want us to use proto.SerializeToString; allowing this one file to perform specifically the required serialization operations makes more sense than allowing it more broadly in several files.

The only helper function here performing any cache-specific logic is updateProtoFromHeadersAndMetadata.

This is part 2 of #23175

Signed-off-by: Raven Black <ravenblack@dropbox.com>

Commit Message: [file_system_http_cache] Add proto helpers
Additional Description: This is a small support class as part of of file_system_http_cache implementation.
Risk Level: None, code is not yet used.
Testing: Unit tested.
Docs Changes: n/a
Release Notes: n/a


jmarantz (commented):
at a high level this looks good modulo some nits and one potential performance issue.

CODEOWNERS
```diff
@@ -190,8 +190,9 @@ extensions/filters/http/oauth2 @derekargueta @snowp
 # Config Validators
 /*/extensions/config/validators/minimum_clusters @adisuissa @htuch
 # File system based extensions
-/*/extensions/common/async_files @mattklein123 @ravenblack
-/*/extensions/filters/http/file_system_buffer @mattklein123 @ravenblack
+/*/extensions/common/async_files @mattklein123 @ravenblackx
+/*/extensions/filters/http/file_system_buffer @mattklein123 @ravenblackx
+/*/extensions/http/cache/file_system_http_cache @mattklein123 @ravenblackx
```
jmarantz, line 195: IIRC extensions need to have owners that are envoy maintainers. Probably you should make me the maintainer-owner of the file-system-http-cache directory?

It also looks like a bug below that @tyxia is an owner of gcp_authn even if he created it (at least until he becomes a maintainer).

@mattklein123 do I have the semantics of that right?

jmarantz (commented):
/wait

CODEOWNERS
tyxia: Yea, I noticed this error (actually throughout this file) when I added code owner. I was thinking it will be fine as long as a maintainer (yanalvlasov) co-owns that file.

Going forwarding, I am wondering if the pattern that author(s) + one maintainer co-own the file is ok for CODEOWNER here?

jmarantz: @tyxia that sgtm but I think there's something mechanical in play here flagging that file has having errors if there are non-maintainers listed at all.

tyxia: @jmarantz  I looked into this mechanical thing a bit:
1) The error message (unknown error: make sure xx exists and has write access) you see on the web page here is Github Repo CODEOWNERS syntax . More details can be found [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax) 

2) Envoy has its own policy and check in check_format.py. For example, the extension require at least 2 owners in and requires at least one maintainer OWNER. 
You can find the checks in this [function](https://github.com/envoyproxy/envoy/blob/c9d2507ab6eb9c69a958a0472404c8ee169a4080/tools/code_format/check_format.py#L1110)

So it is really the Github mechanical thing. We could raise the PR to Github community though like [this](https://github.com/orgs/community/discussions/29982)

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,145 @@
...
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+namespace {
+
+const absl::flat_hash_set<Http::LowerCaseString> headersNotToUpdate() {
+  CONSTRUCT_ON_FIRST_USE(
+      absl::flat_hash_set<Http::LowerCaseString>,
+      // Content range should not be changed upon validation
+      Http::Headers::get().ContentRange,
```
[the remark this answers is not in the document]
ravenblackx: Now proposed in PR #23977

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,145 @@
...
+      // We don't update the cached response on a Vary; we just delete it
+      // entirely. So don't bother copying over the Vary header.
+      Http::CustomHeaders::get().Vary);
+}
+} // namespace
+
+void updateProtoFromHeadersAndMetadata(CacheFileHeader& entry, const CacheFileHeader& response) {
+  *entry.mutable_metadata_response_time() = response.metadata_response_time();
+  absl::flat_hash_set<Http::LowerCaseString> updated_header_fields;
+  for (const auto& incoming_response_header : response.headers()) {
```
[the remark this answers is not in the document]
ravenblackx: Thanks for citation. All auto removed.

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,145 @@
...
+    Http::LowerCaseString key{incoming_response_header.key()};
+    if (headersNotToUpdate().contains(key)) {
+      continue;
+    }
```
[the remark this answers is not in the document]
ravenblackx: Removed in favor of PR #23977

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,145 @@
...
+    if (!updated_header_fields.contains(key)) {
+      auto it = entry.mutable_headers()->begin();
+      while (it != entry.mutable_headers()->end() && Http::LowerCaseString{it->key()} != key) {
```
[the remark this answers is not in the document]
ravenblackx: Removed in favor of PR #23977

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,145 @@
...
+  auto trailers = Http::ResponseTrailerMapImpl::create();
+  for (const auto& t : trailer.trailers()) {
+    trailers->addCopy(Http::LowerCaseString(t.key()), t.value());
+  }
+  return trailers;
+}
+
+ResponseMetadata metadataFromHeaderProto(const CacheFileHeader& header) {
+  ResponseMetadata metadata;
+  SystemTime epoch_time;
```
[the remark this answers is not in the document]
ravenblackx: Redone more clearly. (It default-initializes to zero, so no, but you're right it wasn't clear.)

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.h
```diff
@@ -0,0 +1,105 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+/**
+ * Update an existing CacheFileHeader with new values from an updateHeaders operation.
+ * - Retains the order of existing header fields if they have only one value.
+ * - If an existing header had more than one value, the first one's order is retained,
```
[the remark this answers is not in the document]
ravenblackx: Removed in favor of PR #23977

source/extensions/filters/http/cache/cache_entry_utils.cc
```diff
@@ -28,6 +28,58 @@ std::ostream& operator<<(std::ostream& os, CacheEntryStatus status) {
...
+  //    Race conditions would not be possible because we are always processing up-to-date data.
+  // 2. No key collision for etag. Therefore, if etag matches it's the same resource.
+  // 3. Backend is correct. etag is being used as a unique identifier to the resource
+
+  // use other header fields provided in the new response to replace all instances
+  // of the corresponding header fields in the stored response
+
+  // `updatedHeaderFields` makes sure each field is only removed when we update the header
+  // field for the first time to handle the case where incoming headers have repeated values
+  absl::flat_hash_set<Http::LowerCaseString> updatedHeaderFields;
```
jmarantz, line 64: style nit: updated_header_fields

CODEOWNERS
jmarantz: OK I think we can just not worry about these warnings for now, as they don't reflect Envoy policy.

Thanks for investigating, Tianyu!

source/extensions/filters/http/cache/cache_entry_utils.cc
```diff
@@ -28,6 +28,58 @@ std::ostream& operator<<(std::ostream& os, CacheEntryStatus status) {
...
+  new_headers.iterate(
+      [&headers_to_update, &updatedHeaderFields](
+          const Http::HeaderEntry& incoming_response_header) -> Http::HeaderMap::Iterate {
+        Http::LowerCaseString lower_case_key{incoming_response_header.key().getStringView()};
+        absl::string_view incoming_value{incoming_response_header.value().getStringView()};
+        if (headersNotToUpdate().contains(lower_case_key)) {
```
jmarantz, line 70: perf nit: capture the map returned by headersNotToUpdate() outside, rather than calling it in the loop, as it has to do some kind of thread-safe access to lazy-init the map.

jmarantz (commented):
This looks good modulo nits. I couldn't find that n^2 loop I commented on earlier; did that get rewritten out?

source/extensions/filters/http/cache/cache_entry_utils.cc
ravenblackx: Ah, I guess I shouldn't click "resolve conversation" as it apparently prevents you from seeing my replies! This bit is in PR #23977, and is moving a block of existing code that I didn't write. As such I'd appreciate if any cleanup of that code could be separate from moving it, and also separate from this whole chain of PRs.

source/extensions/filters/http/cache/cache_entry_utils.cc
ravenblackx: Also PR #23977 (as below).

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,94 @@
...
+    t->set_value(std::string{trailer.value().getStringView()});
+    return Http::HeaderMap::Iterate::Continue;
+  });
+  return file_trailer;
+}
+
+size_t headerProtoSize(const CacheFileHeader& proto) { return proto.SerializeAsString().size(); }
+
+Buffer::OwnedImpl bufferFromProto(const CacheFileHeader& proto) {
+  return Buffer::OwnedImpl{proto.SerializeAsString()};
```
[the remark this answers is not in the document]
ravenblackx: It's the full set of http headers not the small block of sizes (that's why I named the other one FixedHeaderBlock rather than just Header, to help distinguish the two types that are right next to each other a lot.)

I think the least-bad way to serialize a proto without an intermediate string is SerializeToOStream, which I think would be a pretty hefty chunk of extra code to make an ostream capable of writing to a Buffer::OwnedImpl. There's also SerializeToArray but that would require pre-allocating the buffer to the correct size, and measuring a proto isn't that much faster than serializing. How would you feel about a TODO for this? :)

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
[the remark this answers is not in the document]
toddmgreer: works for me

source/extensions/filters/http/cache/cache_entry_utils.cc
jmarantz: sure that makes sense; github review does not make it easy for me to tell when code is moved vs written. Sometimes when writing PRs I call out moves as comments I add to my own PR but they can get lost with new commits.

+1 to letting the initiator of a comment resolve it (most of the time). Sometimes if there are non-controversial suggestions that I just execute as indicated in the PR I'll  go ahead and resolve myself, but if it's just a response in a dialog it makes more sense to let the initiator resolve (IMO).

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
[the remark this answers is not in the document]
jmarantz: sure that's fine, or even just a note to that this will contain the full serialized headers -- I think even full headers is relatively bounded due to default limits in Envoy, so this might be OK, especially relative to the speed of file i/o.

jmarantz:
lgtm for me; please fix clang-tidy. 

@toddmgreer can you do a pass?

source/extensions/filters/http/cache/cache_entry_utils.cc
ravenblackx: I'm slowly getting my head around the github comment system - that the interface you see first won't let you post comments grouped, and that 'resolve' pseudo-disappears it immediately (I'm used the Google review format where a resolve-like button means "mark as resolved" and doesn't disappear anything - Github not keeping track of whether there's any comments I'm still supposed to be addressing is a pain!)

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,96 @@
...
+}
+
+CacheFileHeader makeCacheFileHeaderProto(const Key& key,
+                                         const Http::ResponseHeaderMap& response_headers,
+                                         const ResponseMetadata& metadata) {
+  CacheFileHeader file_header;
+  *file_header.mutable_key() = key;
+  TimestampUtil::systemClockToTimestamp(metadata.response_time_,
+                                        *file_header.mutable_metadata_response_time());
+  response_headers.iterate([&file_header](const Http::HeaderEntry& header) {
```
jmarantz, line 27: nit: you could factor out this iteration using a template function so you don't have to repeat it for trailers.

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
```diff
@@ -0,0 +1,96 @@
...
+    t->set_value(std::string{trailer.value().getStringView()});
+    return Http::HeaderMap::Iterate::Continue;
+  });
+  return file_trailer;
+}
+
+size_t headerProtoSize(const CacheFileHeader& proto) { return proto.ByteSizeLong(); }
+
+Buffer::OwnedImpl bufferFromProto(const CacheFileHeader& proto) {
+  // TODO(ravenblack): consider proto.SerializeToOStream and an ostream impl to Buffer.
```
jmarantz, line 50: or zero-copy.

jmarantz (commented):
lgtm with small nits

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
ravenblackx: It sounds like you were thinking of a larger factoring out than I've done, but I think what I've done here is the most factoring out that doesn't involve a really weird template function (because add_headers vs add_trailers, and either way it needs to be a lambda anyway for the captured destination)

source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.cc
ravenblackx: Changed to zero copy since it's clearly nicer than ostream.

jmarantz (approved):
thanks!

not sure if CI failure is real, and also I'd like @toddmgreer to look.

jmarantz:
@toddmgreer PTAL. THanks!

toddmgreer (approved):
Everything previously discussed seems to have been covered. Looks good.

Sorry for the slow response. I was out of state with family for a while.

PR #23943, opened 2022-11-10
build: fixing brokend dead strip

alyssawilk:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -159,7 +159,7 @@ build:libc++ --action_env=BAZEL_LINKOPTS=-lm:-pthread
 build:libc++ --define force_libcpp=enabled
 
 # Optimize build for binary size reduction.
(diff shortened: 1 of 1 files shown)


alyssawilk:
cc @jpsim @Augustyniak

jpsim:
Is there more context here? How can you tell it was broken before and how can you tell that this fixes the issue?

keith:
if nothing else `--dead_strip` isn't a valid arg to any linker, but definitely not at the same time as `--gc-sections` which is ELF only

keith:
I imagine we don't need either of those since bazel should be flipping those for us and solve the platform dependent issue

jpsim (approved):

Issue #9023, opened 2019-11-14
when `network.rbac` filter is used in combination with `tcp_proxy`, connection to the upstream is established in any case, even if RBAC decision is `Deny`


yskopets:
*Title*: when `network.rbac` filter is used in combination with `tcp_proxy`, connection to the upstream is established in any case, even if RBAC decision is `Deny`

*Description*:

* apparently, `network.rbac` filter doesn't prevent `tcp_proxy` from establishing a connection to an upstream in case when RBAC decision is `Deny`
* I'm wondering, is it a good/intuitive default behaviour ? 

*Repro steps*:

* configure Envoy listener with a combination of `network.rbac` and `tcp_proxy` filters

  ```yaml
  static_resources:
    listeners:
    - name: listener_0
      address:
        socket_address:
          protocol: TCP
          address: 0.0.0.0
          port_value: 10000
      filter_chains:
      - filters:
        - name: envoy.filters.network.rbac
          typed_config:                   
            "@type": type.googleapis.com/envoy.config.filter.network.rbac.v2.RBAC
            rules: {}    # DENY everything
            stat_prefix: backend
        - name: envoy.tcp_proxy
          typedConfig:
            '@type': type.googleapis.com/envoy.config.filter.network.tcp_proxy.v2.TcpProxy
            statPrefix: backend
            cluster: backend
  ```
* use `netcat` as a simple backend implementation (most importantly, it responds before reading  request stream):
  ```bash
  nc -lk -s 127.0.0.1 -p 8081 -e echo -e HTTP/1.1 200 OK\r\nContent-Length: 17\r\n\r\n{"version":"v1"}\n
   ```
* make a request to `envoy`
   ```
   curl localhost:10000
   ```
* apparently, despite the RBAC policy being `deny everything`, connection to the `upstream` is established anyway and a response from it is passed back to the `downstream`

mattklein123:
The issue is that in the network RBAC filter the check is being done in `onData()` and it should be done in `onNewConnection()`. This is so broken that I can't imagine anyone is actually using this. At least I hope not. :scream:

yskopets:
@mattklein123 I could take this issue

rshriram:
The check needs to be done in both. Not just on new connection, as we have codecs that can decode the payloads

PiotrSikora:
@yangminzhu don't we need this fixed for Istio 1.5?

yangminzhu:
[...]

PR #23944, opened 2022-11-10
network: delay server newConnection until after transport socket connect

kyessenov:

Opening for discussion. This PR makes some initial server connection processing steps more explicit by making the following guarantees:
* transport socket must receive `onConnect` for the underlying socket, and then issue `Connected` event when ready to read/write (e.g. after TLS handshake)
* connection must read but wait to write before receiving `Connected` event.
* `onNewConnection` is issued when transport socket connects, not when the underlying socket connects.

This PR addresses violations I found, but it's hard to estimate the downstream impact. I think we'd have to guard this for sure. I ran some sanity checks against Istio tests.
* `tcp_proxy` should disable reads `onNewConnection`, not during initialization (otherwise, SSL socket can't read the handshake).
* `sni_dynamic_forward_proxy` must also disable reads while fetching DNS.
* ALTS server transport socket `onConnected` was never called before, so we have to change it to do nothing now.
* a few tests seem to rely on the fact that tcp_proxy opens upstream connection before finishing handshake, will need to find another way.

https://github.com/envoyproxy/envoy/issues/9023
cc @mattklein123 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


kyessenov:
I'm thinking of splitting the first change (server onConnect) out and turning it into a debug assertion. That way we can catch downstream problems without any significant functional change.

PR #23946, opened 2022-11-10
Start slow start window on first successful HC, make slow start re-enterable

### LINKED ISSUE #18351: Start slow start mode window on first successful HC
*Title*: *Start slow start mode window on first successful HC*

*Description*:

In the [ first iteration of slow start mode ](https://github.com/envoyproxy/envoy/pull/13176) endpoint that has an active HC configured enters slow start mode upon passing first successful HC if its cluster membership duration is within configured slow start window.

Instead, slow start window calculation should start when endpoint passes first HC. More details in this [thread](https://github.com/envoyproxy/envoy/pull/13176#discussion_r716841013).

nezdolik:
Signed-off-by: Kateryna Nezdolii <nezdolik@spotify.com>

New changes to "slow start" logic:
* When active HC is enabled per cluster, start "slow start" calculation after first passing HC (drop cluster membership duration condition).
* When no active HC is enabled per cluster, enter "slow start" immediately (drop cluster membership duration condition).
* Endpoint can re-enter "slow start" if active HC is configured per cluster, on each "unhealthy -> healthy" state transition.

Those changes make "slow start" in Envoy to align more with [ELB slow start](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#slow-start-mode) feature.

Commit Message:
Additional Description:
Risk Level: Medium
Testing: Done
Docs Changes: Done
Release Notes: Not Done
Platform Specific Features:
Fixes #18351


docs/root/intro/arch_overview/upstream/load_balancing/slow_start.rst
```diff
@@ -10,7 +10,7 @@ This could be undesirable for services that require warm up time to serve full p
 Slow start mode is a mechanism that affects load balancing weight of upstream endpoints and can be configured per upstream cluster.
 Currently, slow start is supported in :ref:`Round Robin <envoy_v3_api_field_config.cluster.v3.Cluster.RoundRobinLbConfig.slow_start_config>` and :ref:`Least Request <envoy_v3_api_field_config.cluster.v3.Cluster.LeastRequestLbConfig.slow_start_config>` load balancer types.
 
-Users can specify a :ref:`slow start window parameter<envoy_v3_api_field_config.cluster.v3.Cluster.SlowStartConfig.slow_start_window>` (in seconds), so that if endpoint "cluster membership duration" (amount of time since it has joined the cluster) is within the configured window, it enters slow start mode.
+Users can specify a :ref:`slow start window parameter<envoy_v3_api_field_config.cluster.v3.Cluster.SlowStartConfig.slow_start_window>` (in seconds), which is the duration of slow start mode per endpoint.
```
mattklein123, line 13: Can you add a release note for this behavior change?

envoy/upstream/host_description.h
```diff
@@ -201,6 +201,12 @@ class HostDescription {
    * @return timestamp in milliseconds of when host was created.
    */
   virtual MonotonicTime creationTime() const PURE;
+
+  /**
+   * @return timestamp in milliseconds of when host has transitioned from unhealthy to
```
mattklein123, line 206: (nit, unit is opaque)

```suggestion
   * @return timestamp of when host has transitioned from unhealthy to
```

envoy/upstream/upstream.h
```diff
@@ -220,6 +220,12 @@ class Host : virtual public HostDescription {
    */
   virtual void setOutlierDetector(Outlier::DetectorHostMonitorPtr&& outlier_detector) PURE;
 
+  /**
+   * Set the timestamp of when the host has transitioned from unhealthy to healthy state via an
+   * active healchecking.
+   */
+  virtual void setLastHcPassTime(MonotonicTime&& last_hc_pass_time) PURE;
```
mattklein123, line 227: nit: realistically it will likely be faster to pass this by value as it's just an integer

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -274,6 +275,9 @@ void HealthCheckerImplBase::ActiveHealthCheckSession::handleSuccess(bool degrade
   num_unhealthy_ = 0;
 
   HealthTransition changed_state = HealthTransition::Unchanged;
+  if (first_check_ || host_->healthFlagGet(Host::HealthFlag::FAILED_ACTIVE_HC)) {
+    host_->setLastHcPassTime(time_source_.monotonicTime());
+  }
```
mattklein123, line 280: I'm wondering if we should actually do this in the block below when we reach the healthy threshold? It seems like we should start slow start when the host actually becomes healthy? wdyt?

mattklein123 (changes requested):
Thanks for working on this. I think this makes a lot of sense. Some small comments and also it looks like tidy is still broken. Thank you.

/wait

source/common/upstream/health_checker_base_impl.cc
nezdolik: yes, did not realise that hosts have `Host::HealthFlag::FAILED_ACTIVE_HC` flag set during initialisation

nezdolik:
need to increase test coverage

mattklein123 (changes requested):
Thanks makes sense to me. 

/wait

changelogs/current.yaml
```diff
@@ -19,6 +19,12 @@ behavior_changes:
     this kind of update will be rejected. The runtime flag ``envoy.reloadable_features.enable_update_listener_socket_options`` can be used for revert this behavior.
 - area: build
   change: |
+    moved the strict_dns, original_dst, and logical_dns clusters to extensions. If you use these clusters and override extensions_build_config.bzl you will now need to include it explicitly.
+    moved the static, strict_dns and original_dst clusters to extensions. If you use these clusters and override extensions_build_config.bzl you will now need to include it explicitly.
+- area: loadbalancing
+  change: |
+    When active HC is enabled per cluster, slow start calculation now starts after first passing HC. Cluster membership duration condition is dropped from slow start calculation.
+    Endpoint can now re-enter slow start if active HC is configured per cluster, on each "unhealthy -> healthy" state transition.
```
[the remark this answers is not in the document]
nezdolik: yeap, fixed.

nezdolik:
There was a failure in unrelated test, merged latest main. Let's see if it helps

mattklein123:
> @mattklein123 so far the CI has been complaining about overall low test coverage for class RealHostDescription in logical_host.h. That class is not used anywhere in code and in tests. I have added test for RealHostDescription, but had to change visibility for the library that included it (now format check if complaining because of changed visibility) Wonder if better approach would have been to remove that class altogether?

Yes please remove if it's dead code! Thank you.

/wait

test/extensions/clusters/logical_dns/logical_dns_cluster_test.cc
```diff
@@ -139,6 +139,9 @@ class LogicalDnsClusterTest : public Event::TestUsingSimulatedTime, public testi
     EXPECT_EQ("", data.host_description_->locality().zone());
     EXPECT_EQ("", data.host_description_->locality().sub_zone());
     EXPECT_EQ("foo.bar.com", data.host_description_->hostname());
+    EXPECT_FALSE(data.host_description_->lastHcPassTime());
```
nezdolik, line 142: i was wrong, class `RealHostDescription` is used in one [function](https://github.com/envoyproxy/envoy/blob/main/source/extensions/clusters/common/logical_host.cc#L16). So i reverted previous commit that changes visibility level and instead increased coverage in existing test. Those functions had 0% coverage. There were few more (mostly setters), but this should be good enough to pass coverage threshold.

mattklein123 (approved):
Thanks!

Issue #23699, opened 2022-10-27
per-client idle timeout for tcp conn pool


JuniorHsu:
*Title*: *per-client idle timeout for tcp conn pool*

*Description*:
Thrift proxy uses tcp_conn_pool to manage connections.
If a host has `n` downstream host and the max host-to-host QPS is `q_max`,
the connections on the upstream host would be up to `n * q_max` and those connections are closed only if the pool idle timeout is triggered. However, pool idle timeout requires 0 active streams on the client, which is not going to happen to online systems.
If average host-to-host is `q_ave`, then we wastes `n * (q_max - q_ave)` connections at most.
We notice `downstream_cx_active` almost monotonously increases until we don't hot restart envoy or service.

Therefore we propose a per-client idle timeout.
While we return the connection to the pool, the idle timeout is active until the connection is reused.
If the timeout fires, we closes the connection and let the pool knows.

Here's very brief pseudo code
```c++
diff --git a/source/common/tcp/conn_pool.cc b/source/common/tcp/conn_pool.cc
index 0ece4a46b1..fa2a58c401 100644
--- a/source/common/tcp/conn_pool.cc
+++ b/source/common/tcp/conn_pool.cc
@@ -53,6 +53,7 @@ void ActiveTcpClient::clearCallbacks() {
   }
   callbacks_ = nullptr;
   tcp_connection_data_ = nullptr;
+  // enable client idle timeout
   parent_.onStreamClosed(*this, true);
   parent_.checkForIdleAndCloseIdleConnsIfDraining();
 }
diff --git a/source/common/tcp/conn_pool.h b/source/common/tcp/conn_pool.h
index f9a8885591..3ce2f2635e 100644
--- a/source/common/tcp/conn_pool.h
+++ b/source/common/tcp/conn_pool.h
@@ -204,6 +204,7 @@ public:

   void onPoolReady(Envoy::ConnectionPool::ActiveClient& client,
                    Envoy::ConnectionPool::AttachContext& context) override {
+    // disable client idle timeout
     ActiveTcpClient* tcp_client = static_cast<ActiveTcpClient*>(&client);
     tcp_client->readEnableIfNew();
     auto* callbacks = typedContext<TcpAttachContext>(context).callbacks_;
```

As for the new config field, I'm not sure which is better:
[...]

PR #23947, opened 2022-11-11
TCP: per-client idle timeout for tcp conn pool

JuniorHsu:

Commit Message:

While we return the connection to the pool, the idle timeout starts ticking until the connection is reused. If the timeout fires, we closes the connection and let the pool knows.

Additional Description:
Risk Level: mid
Testing: unit test, had a e2e test
[Optional Fixes #Issue] #23699


JuniorHsu:
This is WIP as the unit test is not done yet.
Default idle timeout is 1 hour.
We did an e2e test on a ~40 QPS downstream host.
After running 5 hours, the memory usage is 4% less than the host without idle timeout.
The connection reuse rate  `1 - (cx_total/rq_total)` is about the same (99.7%)

JuniorHsu:
cc @alyssawilk @zuercher @rgs1 @fishcakez @PeterL328

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -1184,6 +1184,14 @@ message Cluster {
   // If ``connection_pool_per_downstream_connection`` is true, the cluster will use a separate
   // connection pool for every downstream connection
   bool connection_pool_per_downstream_connection = 51;
+
+  // The idle timeout for connections. The idle timeout is defined as the period in which
+  // there are no active stream. When the idle timeout is reached the connection will be closed.
+  //
+  // .. warning::
+  //   Disabling this timeout has a highly likelihood of yielding connection leaks due to lost TCP
```
alyssawilk, line 1192: Is this then on by default?  If so we should comment on the default timeout and runtime guard the behavioral change.

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -1184,6 +1184,14 @@ message Cluster {
...
+  //   FIN packets, etc.
+  google.protobuf.Duration tcp_pool_idle_timeout = 57;
```
alyssawilk, line 1194: If this is TCP only I think it should be in tcp-specific config, kind of like the upstream http protocol options.
I think it makes sense as a general field, though we'd need to put some thought into how it interacts with prefetching connections.

envoy/upstream/upstream.h
```diff
@@ -811,6 +811,11 @@ class ClusterInfo : public Http::FilterChainFactory {
    */
   virtual const absl::optional<std::chrono::milliseconds> idleTimeout() const PURE;
 
+  /**
```
alyssawilk, line 814: how does tihs differ from idleTimeout?  At a glance it looks to me like it's possible all we need is a tcp options parallel to the http protocol options so we can set idleTimeout for non-tcp-upstreams. WDIT?

alyssawilk (commented):
thanks for tackling this!
A couple of high level thoughts before I dive into details
/wait

alyssawilk:
FWIW I'm fine with you de-drafting this and just putting NEEDED in the test field.  We have slack machinery for reminding us to do reviews and I might otherwise lost track of it :-)

api/envoy/config/cluster/v3/cluster.proto
JuniorHsu: Yes I'll add a runtime guard. Thanks for pointing this out.

api/envoy/config/cluster/v3/cluster.proto
JuniorHsu: tcp pool is used by several L7 protocol like thrift so I'm with you to put as a general field. 

Default idle timeout (with runtime guard unblock) would be an hour, and the timer starts ticking only if no one holds the connection. Does prefetching in this context mean creating at most 3 more connections for every new connection? If yes, I wouldn't worry about time performance for  1 hour idle timeout. That hurts only if we don't need that much connection every moment in previous hour.

envoy/upstream/upstream.h
JuniorHsu: I assumed you were saying `non-http-upstreams`

`idleTimeout` here is the same idea with `tcpPoolIdleTimeout` so we can use `idleTimeout` for both http codec and tcp pool.

`idleTimeout` could be what http protocol options say or what cluster proto says, as we need separated configs. Does it make sense?

JuniorHsu:
> FWIW I'm fine with you de-drafting this and just putting NEEDED in the test field. We have slack machinery for reminding us to do reviews and I might otherwise lost track of it :-)

Thanks for your great input! I'm going to work on unit test as I put NEEDED 😅

JuniorHsu:
On the other hand, in our test environment, 1 hour tcp idle timeout save 10% envoy memory after running 3 days
cx_total is pretty much the same 2316 vs 2294 for one upstream and didn't observe significant difference with p99 round trip time

envoy/upstream/upstream.h
JuniorHsu: I eventually keep it separated since i'm not sure if we can have a cluster has both idleTimeout for http and tcpPoolIdleTimeout for tcp pool. happy to make it to one if it's not possible to have both

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -1184,6 +1184,14 @@ message Cluster {
   // If ``connection_pool_per_downstream_connection`` is true, the cluster will use a separate
   // connection pool for every downstream connection
   bool connection_pool_per_downstream_connection = 51;
+
+  // The idle timeout for connections. The idle timeout is defined as the period in which
+  // there are no active stream. When the idle timeout is reached the connection will be closed.
+  //
+  // .. warning::
+  //   Disabling this timeout has a highly likelihood of yielding connection leaks due to lost TCP
+  //   FIN packets, etc.
```
mattklein123, line 1193: Is there a default if not set? If so how does one disable? Can we document that?

mattklein123 (changes requested):
Small API question, thanks.

api/envoy/config/cluster/v3/cluster.proto
JuniorHsu: great point. the is guard by reloadable flags. If reloadable is on, default is an hour, and setting 0 to disable it. I'll document it. Thanks

mattklein123 (approved):
API LGTM, thanks.

api/envoy/config/cluster/v3/cluster.proto
alyssawilk: sorry @mattklein123 can I get your thoughts on this comment?
We have an HTTP equivalent and it's in the http-specific upstream config.  If we're going to leave this here I think we either have to deprecate the HTTP specific option, or at least mention the fact this field is ignored for HTTP upstreams.

mattklein123: Oh I see sorry I missed this. Yeah can we just have a TCP protocol specific options much like we do for HTTP?

api/envoy/config/cluster/v3/cluster.proto
```diff
@@ -1184,6 +1184,9 @@ message Cluster {
   // If ``connection_pool_per_downstream_connection`` is true, the cluster will use a separate
   // connection pool for every downstream connection
   bool connection_pool_per_downstream_connection = 51;
+
+  // TCP protocol options that are applied to TCP connections from pool.
+  core.v3.TcpProtocolOptions tcp_protocol_options = 57;
```
alyssawilk, line 1189: sorry I think I wasn't clear on how this works for HTTP :-/  For HTTP, we have this:

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto

the deal is that you can stuff largely any proto in typed_extension_protocol_options  here:
https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/cluster/v3/cluster.proto#L937

so don't need to declare them in the cluster proto, and folks not needing that proto type don't need to pull it in.  So I think we want to treat TCP as we do HTTP - have a standalone proto (which can live in its own file) and it's added via that type extension config field.  Make sense?

/wait

JuniorHsu: @alyssawilk oh thanks for letting me know. It takes some time to make me aware most of the protocol_options field in cluster.proto are deprecated
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto

HTTP proto structure is like
```
api/envoy/extensions/upstreams/http $ tree .
.
├── generic
│   └── v3
│       ├── BUILD
│       └── generic_connection_pool.proto
├── http
│   └── v3
│       ├── BUILD
│       └── http_connection_pool.proto
├── tcp
│   └── v3
│       ├── BUILD
│       └── tcp_connection_pool.proto
└── v3
    ├── BUILD
    └── http_protocol_options.proto
```

current tcp is 
```
api/envoy/extensions/upstreams/tcp $ tree
.
└── generic
    └── v3
        ├── BUILD
        └── generic_connection_pool.proto
```

So i'm going to add a v3 folder at the root of tcp and put the idle timeout into new standalone tcp_protocol_options.proto. the sturcture will be 
```
api/envoy/extensions/upstreams/tcp $ tree
.
├── generic
│   └── v3
│       ├── BUILD
│       └── generic_connection_pool.proto
└── v3
    ├── BUILD
    └── tcp_protocol_options.proto
```

sounds like a plan? @alyssawilk

alyssawilk: I'm going to let @mattklein123 as I think it's more of an API shephard call :-)
cc @envoyproxy/api-shepherds

mattklein123: Yup this sounds good to me. Thank you.

JuniorHsu: Updated. fwiw i hit
` This change appears to add visibility rules. Please get senior maintainer approval to add an exemption to check_visibility tools/code_format/check_format.py` since I make the new proto public visibility

source/extensions/upstreams/tcp/BUILD
```diff
@@ -0,0 +1,25 @@
...
+licenses(["notice"])  # Apache 2
+
+envoy_extension_package()
+
+envoy_cc_extension(
+    name = "config",
+    srcs = ["config.cc"],
+    hdrs = ["config.h"],
+    # This is core Envoy config.
+    visibility = ["//visibility:public"],
```
JuniorHsu, line 16: This triggers `This change appears to add visibility rules. Please get senior maintainer approval to add an exemption to check_visibility tools/code_format/check_format.py` and blocks ci

Could you help to move forward? @mattklein123 @alyssawilk 
Thanks

mattklein123: I think this is fine as it maps to what we do for HTTP but I will defer to Alyssa who did the HTTP work related to this.

alyssawilk (commented):
Looking good!  Please check out the format issue on CI as well
/wait

api/envoy/extensions/upstreams/tcp/v3/tcp_protocol_options.proto
```diff
@@ -0,0 +1,48 @@
...
+  // The timeout is guarded by
+  // ``envoy.reloadable_features.envoy_reloadable_features_tcp_pool_idle_timeout``.
+  //
+  // If the above runtime flag is enabled and idle timeout is not specified,
+  // this defaults to 1 hour. To disable idle timeouts, explicitly set this to 0.
+  //
+  // .. warning::
+  //   Disabling this timeout has a highly likelihood of yielding connection leaks due to lost TCP
+  //   FIN packets, etc.
+  google.protobuf.Duration idle_timeout = 1;
```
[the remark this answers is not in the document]
JuniorHsu: I saw default is 1 hour everywhere else but happy to make it 10 minutes

source/extensions/upstreams/tcp/BUILD
JuniorHsu: Alyssa agrees on slack so I'm going to tweek `tools/code_format/config.yaml`

api/envoy/extensions/upstreams/tcp/v3/tcp_protocol_options.proto
```diff
@@ -0,0 +1,48 @@
...
+//     - name: some_service
+//       connect_timeout: 5s
+//       typed_extension_protocol_options:
+//         envoy.extensions.upstreams.tcp.v3.TcpProtocolOptions:
+//           "@type": type.googleapis.com/envoy.extensions.upstreams.tcp.v3.TcpProtocolOptions
+//           idle_timeout: 10m
+//        .... [further cluster config]
+
+message TcpProtocolOptions {
+  // The idle timeout for connections. The idle timeout is defined as the period in which
```
alyssawilk, line 35: for connection -> for the connection

api/envoy/extensions/upstreams/tcp/v3/tcp_protocol_options.proto
```diff
@@ -0,0 +1,48 @@
...
+  // there are no active stream. When the idle timeout is reached, the connection will be closed.
```
alyssawilk, line 36: defined as the period in which there connection is not associated with a downstream connection

api/envoy/extensions/upstreams/tcp/v3/tcp_protocol_options.proto
```diff
@@ -0,0 +1,48 @@
...
+  //
+  // The timeout is guarded by
```
alyssawilk, line 38: I think we can skip this - it's in the release notes and is envoy-specific

docs/root/faq/configuration/timeouts.rst
```diff
@@ -155,6 +155,15 @@ TCP
   is the amount of time that the TCP proxy will allow a connection to exist with no upstream
   or downstream activity. The default idle timeout if not otherwise specified is *1 hour*.
 
+* The TCP protocol :ref:`idle_timeout <envoy_v3_api_field_extensions.upstreams.tcp.v3.TcpProtocolOptions.idle_timeout>`
+  is defined in a :ref:`TcpProtocolOptions <envoy_v3_api_msg_extensions.upstreams.tcp.v3.TcpProtocolOptions>`
+  used by all TCP connections from pool. The idle timeout is the time at which
+  a upstream connection will be terminated if there are no active stream. The timeout is guarded by
```
alyssawilk, line 161: again if the connection is not in use

source/common/upstream/upstream_impl.cc
```diff
@@ -1106,6 +1108,19 @@ ClusterInfoImpl::ClusterInfoImpl(
     idle_timeout_ = std::chrono::hours(1);
   }
 
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.tcp_pool_idle_timeout")) {
```
alyssawilk, line 1111: doesn't applying the runtime guard here mean it won't actually reload unless you reload the cluster?  I think we may want to instead create the idle alarm based on the runtime guard such that the flag takes effect on connection creation

test/common/tcp/conn_pool_test.cc
```diff
@@ -403,6 +421,110 @@ TEST_F(TcpConnPoolImplTest, DrainConnections) {
   EXPECT_TRUE(conn_pool_->isIdle());
 }
 
+/**
```
alyssawilk, line 424: Is it possible to get an e2e test as well?  If there's no existing e2e tests with idle connection reuse we could configure it so we reclaim eagerly (1s or less), turn up connection preconnect so that a downstream request kicks off 2 upstream connections and then wait on the stat for the upstream connection to be reclaimed?

alyssawilk (commented):
Awesome.  A few more thoughts (sorry for not picking them up last round)

source/common/upstream/upstream_impl.cc
JuniorHsu: You mean we'll reload the cluster only if two `ClusterInfoImpl` objects around CDS/guard_flip are not equal? Yeah happy to defer the timer decision to connection creation.

test/common/tcp/conn_pool_test.cc
JuniorHsu: Does e2e in the context mean integration test? We did test in our environment for days.

alyssawilk: e2e means /test/integration/tcp_proxy_integration_test.cc :-)  I'm glad you tested internally (that's actually pretty awesome) but it'd be good to have it also end to end regression tested.
I think if you set your new configuration option to a short timeout you can adapt one of the tests which does Preconnect and verify the connection gets reclaimed by waiting on cluster.cluster_0.upstream_cx_destroy

alyssawilk:
@ggreenway this looks good modulo the missing integration test.  Would you be game for second pass?

api/envoy/extensions/upstreams/tcp/v3/tcp_protocol_options.proto
```diff
@@ -0,0 +1,45 @@
...
+//       connect_timeout: 5s
+//       typed_extension_protocol_options:
+//         envoy.extensions.upstreams.tcp.v3.TcpProtocolOptions:
+//           "@type": type.googleapis.com/envoy.extensions.upstreams.tcp.v3.TcpProtocolOptions
+//           idle_timeout: 10m
+//        .... [further cluster config]
+
+message TcpProtocolOptions {
+  // The idle timeout for the connection. The idle timeout is defined as the period in which
+  // there connection is not associated with a downstream connection. When the idle timeout is
```
ggreenway, line 36:
```suggestion
  // the connection is not associated with a downstream connection. When the idle timeout is
```

changelogs/current.yaml
```diff
@@ -225,6 +225,9 @@ new_features:
 - area: http
   change: |
     added :ref:`append_x_forwarded_port <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.append_x_forwarded_port>` to append the ``x-forwarded-port`` header to HTTP upstream requests.
+- area: tcp
+  change: |
+    added :ref:`idle_timeout <envoy_v3_api_field_extensions.upstreams.tcp.v3.TcpProtocolOptions.idle_timeout>` to support per client idle timeout for tcp connection pool. The timeout is guarded by ``envoy.reloadable_features.tcp_pool_idle_timeout`` and timeout defaults to 10 minutes if runtime flag is enabled.
```
ggreenway, line 230: I think this should also be mentioned in section `minor_behavior_changes` because we're adding a timeout that's configured by default, where there wasn't one before.

docs/root/faq/configuration/timeouts.rst
```diff
@@ -155,6 +155,15 @@ TCP
   is the amount of time that the TCP proxy will allow a connection to exist with no upstream
   or downstream activity. The default idle timeout if not otherwise specified is *1 hour*.
 
+* The TCP protocol :ref:`idle_timeout <envoy_v3_api_field_extensions.upstreams.tcp.v3.TcpProtocolOptions.idle_timeout>`
+  is defined in a :ref:`TcpProtocolOptions <envoy_v3_api_msg_extensions.upstreams.tcp.v3.TcpProtocolOptions>`
+  used by all TCP connections from pool. The idle timeout is the time at which
+  a upstream connection will be terminated if the connection is not in use. The timeout is guarded by
```
ggreenway, line 161: `the connection is not in use`. I think it would be helpful to define here (as you did in other places in the docs) exactly what that means, as one reasonable interpretation of `not in use` is `no data being transferred in either direction`, but that is a covered by the tcpproxy idle timeout, not this one.

source/common/tcp/conn_pool.cc
```diff
@@ -55,6 +63,7 @@ void ActiveTcpClient::clearCallbacks() {
   tcp_connection_data_ = nullptr;
   parent_.onStreamClosed(*this, true);
   parent_.checkForIdleAndCloseIdleConnsIfDraining();
+  setIdleTimer();
```
ggreenway, line 66: Can the previous line result in this connection getting closed? If so, maybe move this line up.

source/common/tcp/conn_pool.h
```diff
@@ -125,13 +126,20 @@ class ActiveTcpClient : public Envoy::ConnectionPool::ActiveClient {
   }
   virtual void clearCallbacks();
 
+  // // Called if the underlying connection is idle over the cluster's tcpPoolIdleTimeout()
```
ggreenway, line 129:
```suggestion
  // Called if the underlying connection is idle over the cluster's tcpPoolIdleTimeout()
```

ggreenway (changes requested):
Looks good overall

/wait

source/common/tcp/conn_pool.h
JuniorHsu: :facepalm: thanks for catching this...

source/common/tcp/conn_pool.cc
JuniorHsu: Nice catch. Thanks!

changelogs/current.yaml
JuniorHsu: I move to `minor_behavior_changes` and put a see also here in new feature

alyssawilk (approved):

test/integration/tcp_proxy_integration_test.cc
```diff
@@ -682,6 +682,8 @@ TEST_P(TcpProxyIntegrationTest, TestIdletimeoutWithNoData) {
   tcp_client->waitForDisconnect();
 }
 
+// TODO(kuochunghsu) Debug apple CI failure.
+#if !defined(__APPLE__)
```
ggreenway, line 686: What was the failure? I'd like to have some understanding of it before disabling a test on a platform.

ggreenway (changes requested):
/wait-any

test/integration/tcp_proxy_integration_test.cc
JuniorHsu: I don't have apple build on hand so here's what I can tell.

The log simply shows timeout for apple platform but that might be testing environment issue.
While I wrote the e2e test without preconnections, 
I grab a connection from pool, close the connection to return it to pool, the idle timer starts ticking but never fire alarms. 
I was hinted to use preconnections, and those timers fire alarms as expected.

Not a thread issue since everything seems on the same thread
```
[ RUN      ] IpVersions/TcpProxyIntegrationTest.TestPerClientIdletimeout/IPv4
[2022-12-15 17:33:08.949][32224][error][pool] [source/common/tcp/conn_pool.cc:116] [C4] setIdleTimer
[2022-12-15 17:33:08.949][32224][error][pool] [source/common/tcp/conn_pool.cc:116] [C5] setIdleTimer
[2022-12-15 17:33:08.949][32224][error][pool] [source/common/tcp/conn_pool.cc:116] [C7] setIdleTimer
[2022-12-15 17:33:08.949][32224][error][pool] [source/common/tcp/conn_pool.cc:109] [C7] disableIdleTimer
[2022-12-15 17:33:08.950][32224][error][pool] [source/common/tcp/conn_pool.cc:109] [C7] disableIdleTimer
[2022-12-15 17:33:08.950][32224][error][pool] [source/common/tcp/conn_pool.cc:116] [C7] setIdleTimer
[2022-12-15 17:33:09.050][32224][error][pool] [source/common/tcp/conn_pool.cc:102] [C4] per client idle timeout
[2022-12-15 17:33:09.050][32224][error][pool] [source/common/tcp/conn_pool.cc:109] [C4] disableIdleTimer
[2022-12-15 17:33:09.050][32224][error][pool] [source/common/tcp/conn_pool.cc:102] [C5] per client idle timeout
[2022-12-15 17:33:09.050][32224][error][pool] [source/common/tcp/conn_pool.cc:109] [C5] disableIdleTimer
```

Hence I have a hunch there's a profound testing glitch in specific platform which is beyond my knowledge.
If this is a concern, I am happy to make idle time out happen without specific platform.

ggreenway: Are you able to determine for sure whether it's an issue with the test, or an issue with the feature?

I tried running 1000 iterations of the test on my mac and saw no failures. If we can't reproduce it locally, this could be a challenge to figure out.

JuniorHsu: It's a testing issue since I could reproduce locally (finger crossed) and `TCPActiveClient` is never created unless I put `
  test_server_->waitForCounterGe("cluster.cluster_0.upstream_cx_total", 1);` to wait. PR is coming

JuniorHsu: I `git push` the update. Let's see if ci is happy

JuniorHsu: Nice. Both fails last time.
https://dev.azure.com/cncf/envoy/_build/results?buildId=123667&view=logs&j=fa3d3e18-6969-5713-c3e7-3581195704fd&t=ba370bef-ca49-52f5-0c93-7c0c0f27c465&l=38777

```
[ RUN      ] IpVersions/TcpProxyIntegrationTest.TestPerClientIdletimeout/IPv4
[external/com_google_absl/absl/flags/internal/flag.cc : 115] RAW: Restore saved value of envoy_reloadable_features_no_extension_lookup_by_name to: true
[external/com_google_absl/absl/flags/internal/flag.cc : 115] RAW: Restore saved value of envoy_reloadable_features_runtime_initialized to: false
[       OK ] IpVersions/TcpProxyIntegrationTest.TestPerClientIdletimeout/IPv4 (398 ms)
...
[ RUN      ] IpVersions/TcpProxyIntegrationTest.TestPerClientIdletimeout/IPv6
[external/com_google_absl/absl/flags/internal/flag.cc : 115] RAW: Restore saved value of envoy_reloadable_features_runtime_initialized to: false
[external/com_google_absl/absl/flags/internal/flag.cc : 115] RAW: Restore saved value of envoy_reloadable_features_no_extension_lookup_by_name to: true
[       OK ] IpVersions/TcpProxyIntegrationTest.TestPerClientIdletimeout/IPv6 (383 ms)

```

JuniorHsu:
@ggreenway tests pass, might to take another look? Thanks.

PR #23949, opened 2022-11-11
http: add new early header mutation extension and simple implementation

### LINKED ISSUE #23786: optional early header mutation before tracing and routing
Inspired by the #23550. It would be helpful to support early header mutation before tracing and routing. 

Especially, by this way, it's possible to bridge different tracing system. For example, envoy enabled tracer B. 

Then when a request with trace headers of tracer A is coming, we can create trace headers of tracer B according to the exist A's headers before the tracing is start.

It's also could be used to do some custom modification to the request headers before routing to avoid unnecessary route entry refresh. For example, if we want to remove a specific prefix of all requests' path and route it based on the result path of modification, we need two searches of route table in previous implementation (one for initial search, one for refresh after modification).

wbpcode:
Commit Message: http: add new early header mutation extension and simple regex implementation
Additional Description:

To close #23786

Risk Level: Low. New optional feature.
Testing: unit test, integration test.
Docs Changes: n/a.
Release Notes: added.


api/envoy/config/common/mutation_rules/v3/mutation_rules.proto
```diff
@@ -84,3 +85,19 @@ message HeaderMutationRules {
...
+      [(validate.rules).string = {min_len: 1 well_known_regex: HTTP_HEADER_NAME strict: false}];
+
+  // Optional new header name. If not empty, the final header name after mutation will be replaced with this name.
+  string rename = 2 [
+    (validate.rules).string = {well_known_regex: HTTP_HEADER_NAME strict: false ignore_empty: true}
+  ];
+
+  // Regex match and rewrite operations that will be applied to the header value.
+  type.matcher.v3.RegexMatchAndSubstitute regex_rewrite = 3
+      [(validate.rules).message = {required: true}];
```
mattklein123, line 102: I think it might be cleaner to have this inside of a oneof so it's clearly one operation per mutation? And then make the oneof required to be set via PGV? wdyt?

api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto
```diff
@@ -579,6 +579,15 @@ message HttpConnectionManager {
   // [#extension-category: envoy.http.original_ip_detection]
   repeated config.core.v3.TypedExtensionConfig original_ip_detection_extensions = 46;
 
+  // The configuration for the early header mutation extensions.
+  //
+  // When configured the extensions will be called before any routing, tracing, or any filter processing.
+  // Each extension will be applied in the order they are configured.
+  // The same header will be mutated by multiple extensions, then the last extension will win.
```
mattklein123, line 586:
```suggestion
  // If the same header is mutated by multiple extensions, then the last extension will win.
```

api/envoy/extensions/http/early_header_mutation/regex_mutation/v3/regex_mutation.proto
```diff
@@ -0,0 +1,25 @@
...
+// [#protodoc-title: Custom early header mutation extension]
+// [#extension: envoy.http.early_header_mutation.regex_mutation]
+
+// This extension allows for custom early header mutation by the regex match and rewrite.
+// It is configured with a list of
+// :ref:`header mutation <envoy_v3_api_msg_config.common.mutation_rules.v3.HeaderMutation>`.
+message RegexMutation {
+  repeated config.common.mutation_rules.v3.HeaderMutation header_mutations = 1
+      [(validate.rules).repeated = {min_items: 1}];
+}
```
mattklein123, line 25: Honestly I'm on the fence as to whether this should be an extension or not. It seems like there could just be a list of early header mutations that live within the HCM config? @envoyproxy/api-shepherds @alyssawilk wdyt? It seems like a fair amount of complexity and it's not clear to me that it really needs full customization for any known use case.

If we do keep the extension, I would not call this the RegexMutation. I would call it something more generic that can just operate on the HeaderMutation structure which could eventually do other things per my oneof comment.

mattklein123 (commented):
Thanks for working on this. A few high level comments to get started before the full review.

api/envoy/extensions/http/early_header_mutation/regex_mutation/v3/regex_mutation.proto
alyssawilk: I still don't grok why we can't do this as a filter and just clear the route cache, but if we're going to add it I'd prefer we add it as an extension as I can only imagine it growing over time

mattklein123: I think the issue is that a) there may be significant perf cost for reloading the route and b) some other stuff like tracing pivots on this before filters run.

An extension is OK with me but let's try to make the default version more generic.

mattklein123: yeah agree that's an interesting option to explore. I'm not sure how hard that would be to implement @wbpcode? In general I agree less extension points is better if we don't need it.

api/envoy/config/common/mutation_rules/v3/mutation_rules.proto
wbpcode: Sorry, may be my comment is not clear enough.
`rename` is used if we just want to set a new header with the value from the regex substitution (with the original header value as input) and key from the `rename`.
Both original header and new header will be kept.

Seems `oneof` couldn't achieve this target.

api/envoy/extensions/http/early_header_mutation/regex_mutation/v3/regex_mutation.proto
wbpcode: > yeah I guess I'd rather have an HCM option to delay route selection/tracing and let the filters do it and let this be a filter thing than invent another header modifier (but also again not blocking the plan, just suggesting alternates)

As I know, the route  (route-level upgrade map) is necesssary for HTTP filter chain creation. So, it's almost impossible to delay the route selection.

But I agree adding another header modifier is not best choice. I just couldn't figure out a better choice.

wbpcode: I am thinking if is possible to add a special flag to the current http filter to tell it's a filter should run at very early.

And then we can extend the downstream `FilterManager` to add the methods to handle these early filters. We can add some limits to these filters, such as never stop, only for decoder, only for headers, no route dependency, etc.

And finally, we can implement a common filter to do the header mutations. This filter could run as early filter, normal downstream filter and even upstream filter. And the substitution formatter could be used in the filter. This filter even could be used to replace all `request_headers_to_add`/`response_headers_to_add`.

By this way, users could get a unified header modifier. 

But seems it's too complex for simple header mutation. (And I am not very sure if is possible to implement it as my above envision. Some pre-researchings are necessary.)

wbpcode: Delaying route selection is almost impossible, especially considering that we may need enable different filters based on the route (very common requirement in the multi-tenant gateway) in the future.

Refactoring filter manager is too complex. It won't have better performance and users need understand the new concept of `early filter`.

May simple new extension point is an acceptable option. orz

But as @mattklein123 said, we can make the default extension more general (may be substitution formatter could be used here.)

mattklein123: I'm OK with keeping the new extension point, but can we make it more general and potentially just allow the standard array of header options? Thank you.

/wait

wbpcode:
`HeaderValueOption` is used for this early header mutation.
I am thinking should we add regex substitution to `HeaderValueOption`? By this way, we can achieve flexible data extraction and formatting. 🤔

wbpcode:
friendly ping cc @mattklein123

api/envoy/extensions/http/early_header_mutation/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,23 @@
...
+
+import "udpa/annotations/status.proto";
+
+option java_package = "io.envoyproxy.envoy.extensions.http.early_header_mutation.header_mutation.v3";
+option java_outer_classname = "HeaderMutationProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/http/early_header_mutation/header_mutation/v3;header_mutationv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Custom early header mutation extension]
```
mattklein123, line 15: nit: I would remove "Custom" and just call this something generic, like "Default early header mutator" or something like that.

api/envoy/extensions/http/early_header_mutation/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,23 @@
...
+// [#extension: envoy.http.early_header_mutation.header_mutation]
+
+// This extension allows for custom early header mutation by the regex match and rewrite.
+message HeaderMutation {
+  repeated config.core.v3.HeaderValueOption headers_to_append = 1;
+
+  repeated string headers_to_remove = 2;
```
mattklein123, line 22: They way I would model this is as a sequence of operations that happen in order. So I would make a sub-message like:

```
message Mutation {
  oneof mutation_type {
    ...
  }
}
```

Then I would have a repeated list of `Mutation` objects. This will a) allow us to add other things in the future and b) make ordering very clear.

mattklein123 (changes requested):
A few API comments. Generally LGTM at a high level. Thanks.

/wait

api/envoy/extensions/http/early_header_mutation/header_mutation/v3/header_mutation.proto
wbpcode: got it.

wbpcode:
friendly ping cc @mattklein123

changelogs/current.yaml
```diff
@@ -216,6 +216,12 @@ new_features:
 - area: tracing
   change: |
     added support for setting the hostname used when sending spans to a Datadog collector using the :ref:`collector_hostname <envoy_v3_api_field_config.trace.v3.DatadogConfig.collector_hostname>` field.
+- area: http
+  change: |
+    added support of :ref:`early header mutation <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.early_header_mutation_extensions>` to the HTTP connection manager.
+- area: http
+  change: |
+    added support of :ref:`regex mutation <envoy_v3_api_msg_extensions.http.early_header_mutation.header_mutation.v3.HeaderMutation>` for HCM :ref:`early header mutation <envoy_v3_api_field_extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.early_header_mutation_extensions>`.
```
mattklein123, line 224: update text to not refer to regex anymore?

envoy/http/early_header_mutation.h
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Interface class for early header mutation extensions.
+ */
+class EarlyHeaderMutation {
+public:
+  virtual ~EarlyHeaderMutation() = default;
+
+  /**
+   * Mutate the request headers at very early.
```
mattklein123, line 23: nit: Not very clear. Can you be more clear about when this happens?

source/common/http/conn_manager_utility.cc
```diff
@@ -78,7 +78,16 @@ ServerConnectionPtr ConnectionManagerUtility::autoCreateCodec(
 ConnectionManagerUtility::MutateRequestHeadersResult ConnectionManagerUtility::mutateRequestHeaders(
     RequestHeaderMap& request_headers, Network::Connection& connection,
     ConnectionManagerConfig& config, const Router::Config& route_config,
-    const LocalInfo::LocalInfo& local_info) {
+    const LocalInfo::LocalInfo& local_info, const StreamInfo::StreamInfo& stream_info) {
+
+  if (!config.earlyHeaderMutationExtensions().empty()) {
```
mattklein123, line 83: nit: this empty check isn't needed. The for loop will just not execute below if empty.

test/per_file_coverage.sh
```diff
@@ -79,6 +79,7 @@ declare -a KNOWN_LOW_COVERAGE=(
 "source/server/admin:97.4"
 "source/server/admin:profiler-lib:83"
 "source/server/config_validation:74.8"
+"source/extensions/http/early_header_mutation/header_mutation:95.2" # Death tests don't report LCOV
```
mattklein123, line 82: Why are we needing death tests for this feature?

mattklein123 (changes requested):
LGTM with small comments, thanks.

/wait

changelogs/current.yaml
wbpcode: right, thanks.

api/envoy/extensions/http/early_header_mutation/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,21 @@
...
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/http/early_header_mutation/header_mutation/v3;header_mutationv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Default early header mutation extension]
+// [#extension: envoy.http.early_header_mutation.header_mutation]
+
+// This extension allows for early header mutation by the substitution formatter.
+message HeaderMutation {
+  repeated config.common.mutation_rules.v3.HeaderMutation mutations = 1;
```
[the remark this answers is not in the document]
wbpcode: got it.

envoy/http/early_header_mutation.h
wbpcode: right. got it.

test/per_file_coverage.sh
wbpcode: Just for unset enum of `oneof`. It's ok for me to remove it. 🤔

envoy/http/early_header_mutation.h
wbpcode: comment updated.

mattklein123 (approved):
LGTM, thanks!

test/per_file_coverage.sh
mattklein123: Ah OK that's fine. No problem.

alyssawilk: https://storage.googleapis.com/envoy-pr/9d19ddc/coverage/source/extensions/http/early_header_mutation/header_mutation/index.html

Yep, the only uncovered lines are PANICs.  Bummer that goes below the limit and 100% fine to adjust for that :-)

PR #23957, opened 2022-11-11
Use getaddrinfo to parse IPv6 addresses to allow link local addresses

fcrisciani:
Commit Message: use only `getaddrinfo` to parse IPv6
Risk Level: low
Testing: unit tests

### DIFF
--- a/test/common/network/utility_test.cc
+++ b/test/common/network/utility_test.cc
@@ -140,6 +140,7 @@ TEST(NetworkUtility, ParseInternetAddress) {
   EXPECT_THROW(Utility::parseInternetAddress("fffff::"), EnvoyException);
   EXPECT_THROW(Utility::parseInternetAddress("/foo"), EnvoyException);
 
+  // TODO(#24326): make windows getaddrinfo more strict. See below example.
 #ifndef WIN32
   EXPECT_THROW(Utility::parseInternetAddress("[::]"), EnvoyException);
   EXPECT_THROW(Utility::parseInternetAddress("[::1]:1"), EnvoyException);
@@ -234,6 +235,7 @@ TEST(NetworkUtility, ParseInternetAddressAndPort) {
   EXPECT_THROW(Utility::parseInternetAddressAndPort("::1"), EnvoyException);
   EXPECT_THROW(Utility::parseInternetAddressAndPort("::"), EnvoyException);
 
+  // TODO(#24326): make windows getaddrinfo more strict. See above example.
 #ifndef WIN32
   EXPECT_THROW(Utility::parseInternetAddressAndPort("[[::]]:1"), EnvoyException);
 #endif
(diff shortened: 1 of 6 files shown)


test/common/network/utility_test.cc
```diff
@@ -150,8 +153,10 @@ TEST(NetworkUtility, ParseInternetAddress) {
   EXPECT_EQ(nullptr, Utility::parseInternetAddressNoThrow("0:0:0:0"));
   EXPECT_EQ(nullptr, Utility::parseInternetAddressNoThrow("fffff::"));
   EXPECT_EQ(nullptr, Utility::parseInternetAddressNoThrow("/foo"));
+#ifndef WIN32
   EXPECT_EQ(nullptr, Utility::parseInternetAddressNoThrow("[::]"));
   EXPECT_EQ(nullptr, Utility::parseInternetAddressNoThrow("[::1]:1"));
+#endif
```
yanavlasov, line 159: I think in the NoThrow case it makes sense to have the `#else` clause to show what happens on Windows.

yanavlasov (dismissed):
/wait

test/common/network/utility_test.cc
fcrisciani: done.

wrowe: Same logic above at 142 (note an extra empty line there before #if), below at 237 & 251?

We might want to record a bug to track a different PR to the getaddrinfo call wherein our wrapper enforces a little more strictness, feel free to assign me to it.

wrowe (changes requested):
Otherwise makes sense to me

test/common/network/utility_test.cc
fcrisciani: created the TODO with the bug number and added an extra space before the compiler conditions.

wrowe (changes requested):
Thanks for creating the issue! My original request wasn't as much for consistent whitespace (but that's also goodness, TY), but to illustrate all the exceptions so we can document the broken behavior and proceed to fix them in OsSysCalls, as another PR.

test/common/network/utility_test.cc
```diff
@@ -139,8 +139,11 @@ TEST(NetworkUtility, ParseInternetAddress) {
   EXPECT_THROW(Utility::parseInternetAddress("0:0:0:0"), EnvoyException);
   EXPECT_THROW(Utility::parseInternetAddress("fffff::"), EnvoyException);
   EXPECT_THROW(Utility::parseInternetAddress("/foo"), EnvoyException);
+
+#ifndef WIN32
   EXPECT_THROW(Utility::parseInternetAddress("[::]"), EnvoyException);
   EXPECT_THROW(Utility::parseInternetAddress("[::1]:1"), EnvoyException);
+#endif
```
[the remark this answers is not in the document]
fcrisciani: I'm happy to add it but it's the same as line 160-164, do you want to add here and everywhere else the same parsing issue?

PR #23966, opened 2022-11-12
lb: minor opt: only try to perform locality weighted balancing when it's configured

wbpcode:
Commit Message: lb: minor opt: only try to perform locality weighted balancing when it's configured
Additional Description:

Minor optimization to make the behavior of LB clearer.

Risk Level: Mid. Very core code update.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


source/common/upstream/load_balancer_impl.cc
```diff
@@ -635,21 +636,23 @@ ZoneAwareLoadBalancerBase::hostSourceToUse(LoadBalancerContext* context, uint64_
...
-  } else {
-    locality = host_set.chooseHealthyLocality();
-  }
+  if (locality_weighted_balancing_) {
+    absl::optional<uint32_t> locality;
+    if (host_availability == HostAvailability::Degraded) {
+      locality = host_set.chooseDegradedLocality();
+    } else {
+      locality = host_set.chooseHealthyLocality();
+    }
```
wbpcode, line 645: Is this a behavior change?

Nope. Because only when the `locality_weighted_lb_config` is set the `choose*Locality` methods may return valid `locality`. 

Why we do this?

1. Make the logic clearer. In the previous implementation, there is a very obscure logic, the host set will calculate the locality wights only when the `locality_weighted_lb_config` is set and `choose*Locality` may return a valid value in that condition.
The new `locality_weighted_balancing_` flag will make it clearer.

2. In the future, we may want always calcuate the locality weight for the lb policy extensions. This flag will guard the current behavior not to be changed.

test/common/upstream/subset_lb_test.cc
```diff
@@ -1993,6 +1994,7 @@ TEST_F(SubsetLoadBalancerTest, EnabledScaleLocalityWeights) {
       },
       host_set_, {50, 50});
 
+  common_config_.mutable_locality_weighted_lb_config();
```
wbpcode, line 1997: This change not just fix the test for new implementation. But also make the test closer to the reality.

Because the `localityWeightAware` can only be enabled when the `locality_weighted_lb_config` is set.

wbpcode:
Although this PR only changed a little code, It's very core code update. So, a careful review is necessary to avoid I missed something. Thanks.

source/common/upstream/load_balancer_impl.cc
wbpcode: get it.

wbpcode: But the else clause may be broken soon. Because when the `LoadBalancingPolicyConfig` (extension lb policy) is used, the 
locality weight will always be calculated. This is also a reason to add this if check. This if check will ensure behavior here 
not be changed when the `LoadBalancingPolicyConfig` (extension lb policy) is used.

https://github.com/envoyproxy/envoy/blob/21d0f004a8188e49a39e76ec4dc4ae996ca5cb10/source/common/upstream/upstream_impl.cc#L1770-L1776

mattklein123 (commented):
LGTM but needs format fix.

/wait

mattklein123 (approved):

PR #23967, opened 2022-11-12
lb api: moving load balancing policy specific configuration to extension configuration

wbpcode:
Commit Message: lb api: load balancing policy specific configuration to extension configuration
Additional Description:
Risk Level: low. api only.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
This is part of https://github.com/envoyproxy/envoy/issues/20634.

This PR moved the `zone_aware_config` and `locality_weighted_lb_config` from `common_lb_config`. to least request, round rabin and random's extension configuration.

And `consistent_hashing_lb_config` is moved to maglev's extension configuration.

wbpcode:
cc @markdroth

is it possible for us to use `ConsistentHashingLbConfig` in the ring hash directly? Or to use `envoy.extensions.load_balancing_policies.common.v3.SlowStartConfig` in the configuration of least request or round ranbin directly?

I know we cann't update any API directly according to our API version policy even it's mared `not-implemented-hide`.
But seems that it would be better to use `envoy.extensions.load_balancing_policies.common.v3.SlowStartConfig` rather than `config.cluster.v3.Cluster.SlowStartConfig`.
And the ring hash should shared the same `ConsistentHashingLbConfig` with the maglev.

So, if it's possible to change the API directly (as a special case), it would be great.

api/envoy/extensions/load_balancing_policies/least_request/v3/least_request.proto
```diff
@@ -56,4 +57,10 @@ message LeastRequest {
   // Configuration for slow start mode.
   // If this configuration is not set, slow start will not be not enabled.
   config.cluster.v3.Cluster.SlowStartConfig slow_start_config = 3;
```
markdroth, line 59: I guess this should also switch to `common.v3.SlowStartConfig`.  Technically, that's a breaking change, but I suspect no one is actually using this yet, so that's probably okay.

@temawi, please confirm that this change would not break anything for grpc-java.

api/envoy/extensions/load_balancing_policies/least_request/v3/least_request.proto
```diff
@@ -22,7 +23,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // This configuration allows the built-in LEAST_REQUEST LB policy to be configured via the LB policy
 // extension point. See the :ref:`load balancing architecture overview
 // <arch_overview_load_balancing_types>` for more information.
-// [#extension: envoy.load_balancing_policies]
```
markdroth, line 25: Why remove these extension comments?  I think they're needed to generate the right documentation.

api/envoy/extensions/load_balancing_policies/maglev/v3/maglev.proto
```diff
@@ -19,11 +21,12 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
...
 // <arch_overview_load_balancing_types>` and :ref:`Maglev<arch_overview_load_balancing_types_maglev>` for more information.
-// [#extension: envoy.load_balancing_policies]
 message Maglev {
   // The table size for Maglev hashing. Maglev aims for "minimal disruption" rather than an absolute guarantee.
   // Minimal disruption means that when the set of upstream hosts change, a connection will likely be sent to the same
   // upstream as it was before. Increasing the table size reduces the amount of disruption.
   // The table size must be prime number limited to 5000011. If it is not specified, the default is 65537.
   google.protobuf.UInt64Value table_size = 1 [(validate.rules).uint64 = {lte: 5000011}];
+
+  common.v3.ConsistentHashingLbConfig consistent_hashing_lb_config = 2;
```
markdroth, line 31: How about just inlining the two fields into this config?  That's what we did for the ring_hash config.  And that way, we don't need the `ConsistentHashingLbConfig` message at all.

api/envoy/extensions/load_balancing_policies/round_robin/v3/round_robin.proto
```diff
@@ -18,9 +19,14 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // This configuration allows the built-in ROUND_ROBIN LB policy to be configured via the LB policy
 // extension point. See the :ref:`load balancing architecture overview
 // <arch_overview_load_balancing_types>` for more information.
-// [#extension: envoy.load_balancing_policies]
 message RoundRobin {
   // Configuration for slow start mode.
   // If this configuration is not set, slow start will not be not enabled.
   config.cluster.v3.Cluster.SlowStartConfig slow_start_config = 1;
```
markdroth, line 25: Same comment as above w.r.t. changing this to use `common.v3.SlowStartConfig`.

api/envoy/extensions/load_balancing_policies/least_request/v3/least_request.proto
```diff
@@ -56,4 +57,10 @@ message LeastRequest {
   // Configuration for slow start mode.
   // If this configuration is not set, slow start will not be not enabled.
   config.cluster.v3.Cluster.SlowStartConfig slow_start_config = 3;
+
+  oneof locality_config_specifier {
```
markdroth, line 61: Looks like we're repeating this `oneof` in 3 different places.  Would it make sense to define a new message in common.proto that contains this oneof, and then just have a single field of that new message type in each of these three places?

markdroth (commented):
Thanks for doing this!

markdroth:
Sorry, didn't see your questions above until after I'd reviewed.  Looks like we're thinking along the same lines, though.

> is it possible for us to use `ConsistentHashingLbConfig` in the ring hash directly? 

In my review, I had actually suggested the opposite, which is just inlining the fields in the maglev config, just like we did in the ring hash config.  That having been said, I am open to instead changing the ring_hash config to use `ConsistentHashingLbConfig` if we think it's likely that either (a) we will want to be able to add new fields to `ConsistentHashingLbConfig` in the future or (b) we think there will be more hash-based LB policies in the future that will want to share these fields.

In terms of API breakage policy, we can definitely add a new `ConsistentHashingLbConfig` field to the ring hash config if we want; the only question would be what to do about the existing inlined fields.  Technically, we can't remove them, but we can mark them as deprecated.  In practice, I suspect nothing would break if we simply removed them, since gRPC does not use those fields, but we don't really lose anything by leaving them there either.

> Or to use `envoy.extensions.load_balancing_policies.common.v3.SlowStartConfig` in the configuration of least request or round ranbin directly?

As per my review comments, this one is a bit trickier.  Technically, changing the type of the field is a breaking change.  However, I don't think gRPC uses these fields at all, and Envoy hasn't implemented these new extensions yet, so I don't think anything would break if we just made this change.

I'd be okay with allowing this as a one-off exception to our breakage policy, if none of the other @envoyproxy/api-shepherds objects.

api/envoy/extensions/load_balancing_policies/least_request/v3/least_request.proto
wbpcode: These comments are wrong. The [#extension] comment should be complete name of extension.
I will ensure all these are correct when I implement these extensions.

api/envoy/extensions/load_balancing_policies/least_request/v3/least_request.proto
wbpcode: Sounds great.

wbpcode:
Thanks for you review comments. @markdroth 

> (a) we will want to be able to add new fields to ConsistentHashingLbConfig in the future or (b) we think there will be more hash-based LB policies in the future that will want to share these fields.

This is what I am worried about.

I will create new commit to address your comments.

api/envoy/extensions/load_balancing_policies/common/v3/common.proto
```diff
@@ -0,0 +1,116 @@
...
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/load_balancing_policies/common/v3;commonv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Common configuration for two or more load balancing policy extensions]
+// [#not-implemented-hide:]
+
+// Configuration for :ref:`zone aware routing
+// <arch_overview_load_balancing_zone_aware_routing>`.
+message ZoneAwareLbConfig {
```
markdroth, line 25: Would it make sense to nest `ZoneAwareLbConfig` and `LocalityWeightedLbConfig` inside of `LocalityLbConfig`?

markdroth (commented):
This looks great to me!

api/envoy/extensions/load_balancing_policies/common/v3/common.proto
wbpcode: > Looks like we're repeating this oneof in 3 different places. Would it make sense to define a new message in common.proto that contains this oneof, and then just have a single field of that new message type in each of these three places?

It's used to address previous comment. As said in the comment, the `oneof` repeat three times and this change eliminate this repeating.

markdroth: Right, the `LocalityLbConfig` message is good.  What I'm suggesting here is to move the `ZoneAwareLbConfig` and `LocalityWeightedLbConfig` messages to be inside of `LocalityLbConfig`, since they're not used anywhere else.

wbpcode: Hah, get it.

wbpcode:

A LGTM is necessary for this API only PR.

wbpcode:

@htuch is out for vacation

wbpcode:
cc @snowp could you give a LGTM to this API only PR when you have free time? Thanks.

api/envoy/extensions/load_balancing_policies/ring_hash/v3/ring_hash.proto
```diff
@@ -71,5 +78,15 @@ message RingHash {
...
+  // ..note::
+  //   This is deprecated and please use :ref:`consistent_hashing_lb_config
+  //   <envoy_v3_api_field_extensions.load_balancing_policies.ring_hash.v3.RingHash.consistent_hashing_lb_config>` instead.
+  google.protobuf.UInt32Value hash_balance_factor = 5 [
+    deprecated = true,
+    (validate.rules).uint32 = {gte: 100},
+    (envoy.annotations.deprecated_at_minor_version) = "3.0"
+  ];
+
+  common.v3.ConsistentHashingLbConfig consistent_hashing_lb_config = 6;
```
snowp, line 91: Docs for this here and for the other lib configs?

snowp (changes requested):
Makes sense to me, just one comment

snowp (approved):
Thanks!

PR #23990, opened 2022-11-14
postgres: support for upstream SSL

### LINKED ISSUE #19527: postgres: Support StartTLS upstream and plain-text downstream

cpakulski:
Commit Message:
postgres: support for upstream SSL

Additional Description:
Postgres filter can negotiate upstream SSL connection with Postgres server and enable upstream encryption.
Upon receiving the initial postgres request, the filter buffers the received packet (without sending it) and sends to the upstream server a request to establish SSL connection. When the server agrees, the postgres filter enables upstream STARTTLS transport socket and sends the previously buffered initial packet. From now on, the connection to upstream is encrypted and the filter can read the postgres payloads in clear-text.

If the server does not agree for SSL or converting STARTTLS transport socket to secure mode fails, depending on the configuration, the filter may continue in clear-text or may tear down the connection.   

Risk Level: Low
Testing: unit, integration and manual tests.
Docs Changes: yes.
Release Notes: yes
Fixes #19527


cpakulski:
@lizan. Friendly ping 😊.

api/contrib/envoy/extensions/filters/network/postgres_proxy/v3alpha/postgres_proxy.proto
```diff
@@ -20,6 +20,21 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // [#extension: envoy.filters.network.postgres_proxy]
 
 message PostgresProxy {
+  // Upstream SSL operational modes.
+  enum SSLMode {
+    // Do not encrypt upstream connection to the server.
+    SSL_DISABLE = 0;
```
lizan, line 26: no need SSL_ prefix for them. Just use `DISABLED`, `REQUIRED`.

api/contrib/envoy/extensions/filters/network/postgres_proxy/v3alpha/postgres_proxy.proto
```diff
@@ -20,6 +20,21 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
...
+
+    // Establish upstream SSL connection to the server. If the server does not
+    // accept the request for SSL connection, the session is terminated.
+    SSL_REQUIRE = 1;
+
+    // Try to establish upstream SSL connection to the server. If the server does not
+    // accept the request for SSL or Envoy cannot successfully establish secure connection
+    // to upstream, the session continues in clear-text.
+    SSL_PREFER = 2;
```
lizan, line 35: What's the use case or motivation of this one? From a security perspective this is no better than disabled SSL, and subject to MITM to disable SSL completely.

cpakulski: It came from #19527. But I agree with you that it would be easy to mount MITM. 
So the remaining options are only REQUIRE and DISABLE. Enum is not truly necessary and boolean can be used. I will refactor the code. Thanks!

lizan: We can leave it as an enum to be future proof.

cpakulski: Done. Enums are used.

api/contrib/envoy/extensions/filters/network/postgres_proxy/v3alpha/postgres_proxy.proto
cpakulski: Done.

lizan (approved):

fabriziomello:
Will run Postgres regression tests over this commit to make sure we didn't break anything.

cpakulski:
Thanks @fabriziomello ! I hope it passed 😀.

PR #23995, opened 2022-11-15
api: update warning

alyssawilk:
I don't remember if there's a canonical way to declare features alpha, but this is
1) now supported by default and
2) a large enough change I'd like to be able to bugfix in the clear rather than wait months for security releases.

Risk Level: low
Testing: n/a
Docs Changes: yes
Release Notes: no


alyssawilk:
cc @paul-r-gall once this is approved you should update your API to the same text

htuch (approved):

htuch:
I think we can use the `work_in_progress` annotation at field level if the API is likely to be unstable. But I don't think that's the case here, rather alpha reflects the implementation status, in which case I think it's fine to just document it as done here. If that's the case, feel free to merge.

PR #23996, opened 2022-11-15
build: More sizeopt fixes

alyssawilk:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -159,8 +159,9 @@ build:libc++ --action_env=BAZEL_LINKOPTS=-lm:-pthread
 build:libc++ --define force_libcpp=enabled
 
 # Optimize build for binary size reduction.
-build:sizeopt -c opt --copt -Os --linkopt=-Wl,--gc-sections
-build:macossizeopt -c opt --copt -Os
+build:sizeopt-sections -c opt --copt -Os --linkopt=-Wl,--gc-sections
+build:sizeopt -c opt --copt -Os
+build:sizeopt-strip -c opt --copt -Os --linkopt=-Wl,-dead_strip
 
 # Test options
 build --test_env=HEAPCHECK=normal --test_env=PPROF_PATH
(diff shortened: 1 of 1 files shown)


alyssawilk:
cc @jpsim

alyssawilk:
I assume if we did this we could remove release-common in E-M bazelrc and do custom options in build:release-ios and build:release-android, WDYT?

jpsim:
The other way to do this would be

```
build:sizeopt -c opt --copt -Os
build:gcsections --linkopt=-Wl,--gc-sections
```

where we'd only pass the `gcsections` config to platforms where it works (Linux + Android) but not by default (iOS, macOS, Windows).

I don't feel strongly here. Doing it the way you're proposing in this PR would probably require the fewest changes to the places where we use `sizeopt` today.

.bazelrc
```diff
@@ -160,6 +160,7 @@ build:libc++ --define force_libcpp=enabled
 
 # Optimize build for binary size reduction.
 build:sizeopt -c opt --copt -Os --linkopt=-Wl,--gc-sections
+build:macossizeopt -c opt --copt -Os
```
jpsim, line 163: This isn't macOS-specific, it's non-ELF, which is all apple platforms and Windows.

```suggestion
build:nonelfsizeopt -c opt --copt -Os
```

Don't love that name either. Naming wise it's definitely easier to flip the defaults with `sizeopt` and `gcsections`,

alyssawilk:
ok, let's priotitize unbreak, then I'll work on replacing where we use sizeopt.
also adding dead strip without the double dash which is supposed to work but now I won't break anything testing.

jpsim (approved):

PR #24000, opened 2022-11-15
bazel: remove unnecessary sizeopt configs

keith:
Bazel automatically applies `--gc-sections` when supported:

https://github.com/bazelbuild/bazel/blob/8bd2d97fd143b4b8a6fc96ce300820e6e422c03a/tools/cpp/unix_cc_configure.bzl#L620-L625

and `-dead_strip` on darwin:

https://github.com/bazelbuild/bazel/blob/4caae75b49e815ad2cf1d805f316bc374f03f2ae/tools/osx/crosstool/cc_toolchain_config.bzl#L2159-L2173

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -159,9 +159,7 @@ build:libc++ --action_env=BAZEL_LINKOPTS=-lm:-pthread
 build:libc++ --define force_libcpp=enabled
 
 # Optimize build for binary size reduction.
-build:sizeopt-sections -c opt --copt -Os --linkopt=-Wl,--gc-sections
 build:sizeopt -c opt --copt -Os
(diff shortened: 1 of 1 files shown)


alyssawilk (approved):

alyssawilk:
should https://github.com/envoyproxy/envoy-mobile/issues/180 be closed then?

keith:
this part is still true unfortunately https://github.com/envoyproxy/envoy-mobile/issues/180#issuecomment-505646254, but if that changes, like with https://github.com/envoyproxy/envoy-mobile/pull/2497, at least the option will be there when it can apply at all

jpsim (approved):

Issue #23809, opened 2022-11-02
quic: require_client_certificate is ignored


kyessenov:
Quic implementation seems to not support mTLS. Server silently accepts `require_client_certificate` but a client without a cert can connect regardless.

kyessenov:
cc @RyanTheOptimist

alyssawilk:
cc @RyanTheOptimist can do we do a quick PR rejecting require_client_certificate if we don't support?

RyanTheOptimist:
Sure that sounds reasonable!

danzh2010:
https://github.com/envoyproxy/envoy/pull/24011 early failed the require_client_certificate configuration, but didn't actually address this issue. Could someone reopen this?

DavidSchinazi:
Our team is looking into implementing this next year. Stay tuned for updates.

/assign @rickyp

zargor:
Would it be there some workaround in the meantime?
Such as some extension or else?

DavidSchinazi:
There is currently no workaround, sorry. But we're still on track to build this this year.

zargor:
How about the plans for this feature? Thanks.
Found this [PR](https://github.com/envoyproxy/envoy/pull/40017) though.

grugnog:
@DavidSchinazi I was curious if you have any updates here or were able to look at the [PR](https://github.com/envoyproxy/envoy/pull/40017) from @agrawroh?

DavidSchinazi:
Hi! My team had a change of priorities, so this is no longer on our roadmap, sorry. About a year ago, we looked at that PR and it didn't seem to fully implement mTLS. We've lost context on this since though.

/unassign @DavidSchinazi

PR #24011, opened 2022-11-15
Throw if client certs are required when using QUIC

RyanTheOptimist:
Since TLS client auth is not currently support with QUIC, add a throw EnvoyException if require_client_certificates is true.

Fixes: #23809

Risk Level: Low
Testing: New unit tests
Docs Changes: None
Release Notes: None


alyssawilk (commented):
Thanks for picking this up.  can we throw on config ingress instead?  That way we'd reject the config rather than having a live data plane crash
/wait

RyanTheOptimist:
> Thanks for picking this up. can we throw on config ingress instead? That way we'd reject the config rather than having a live data plane crash /wait

Oh, I thought this was on config ingress as it was creating the Factory. But clearly not. :) Were you thinking this should go in QuicServerTransportSocketConfigFactory::createTransportSocketFactory()? or something else?

RyanTheOptimist:
> > Thanks for picking this up. can we throw on config ingress instead? That way we'd reject the config rather than having a live data plane crash /wait
> 
> Oh, I thought this was on config ingress as it was creating the Factory. But clearly not. :) Were you thinking this should go in QuicServerTransportSocketConfigFactory::createTransportSocketFactory()? or something else?

Hm. After talking to Dan and digging around the code, I still think this is config ingress. That being said to make this a bit clearer I moved the check from the QuicServerTransportSocketFactory constructor to the createTransportSocketFactory() method. I also change the code from RELEASE_ASSERT to throw which seems to be the right way to reject a config, based on some spelunking I did.

I wonder if we should tweak STYLE.md. I read the following advice as suggesting I should treat this invalid config as a fatal error (and hence RELEASE_ASSERT).

  Although we strongly recommend that any type of startup error leads to a fatal error, since this
  is almost always a result of faulty configuration which should be caught during a canary process,
    
In any case, if this is not the right approach I'd welcome any pointers to code I should cargo cult from :)

alyssawilk (approved):
Awesome, thanks.
cc @envoyproxy/api-shepherds - this rejects prior valid config but to avoid a crash so I think necessary

adisuissa:
Did this previously resulted in a crash or just a feature that was silently ignored?
If this was always crashing, then it implies that no config that supplied that field could have worked, and so this change will not break users.
If, on the other hand, this was silently ignored, then a previously working config may now be rejected. In this case I suggest adding a runtime guard, so users can gradually move the configs.

RyanTheOptimist:
> Did this previously resulted in a crash or just a feature that was silently ignored? If this was always crashing, then it implies that no config that supplied that field could have worked, and so this change will not break users. If, on the other hand, this was silently ignored, then a previously working config may now be rejected. In this case I suggest adding a runtime guard, so users can gradually move the configs.

Ah, yes good point. Previously this did not result in a crash, just silently ignoring the config. As per your suggestion, I switched to a runtime guard. Thanks!

Issue #23585, opened 2022-10-20
Should we include the Envoy Node info in `RateLimitQuotaUsageReports`


tyxia:
Question from @bbassingthwaite : In our use case, the RQLS management server needs to know the Envoy node information in order to understand where the usage report is coming from. Should the first RateLimitQuotaUsageReports that Envoy sends contain the Node similar to other xDS APIs?

tyxia:
Hi @bbassingthwaite, for this question, first let me describe our design a bit.

The relationship of different objects here is : 
- A single report could include quota usage for multiple Buckets (i.e., 1 `RateLimitQuotaUsageReports` <-> many `BucketQuotaUsage`) 
- A single `BucketQuotaUsage` includes the usage info for a group of clients (could be single client per configuration) as the rate limiting reports and assignment is at granularity of each bucket and we group clients into each bucket. This is because you may want your high priority clients (e.g., production clients) to have more of your capacity then lower priority clients (e.g., development) and let them share the same rate limiting decision.
- Also, the way requests/clients group into each bucket can be defined in various ways. e.g., client IP, request header, URL path, etc.

So, the `RateLimitQuotaUsageReports` here is not specified at the granularity of Envoy [Node](https://github.com/envoyproxy/envoy/blob/daa9840d0be7aa5e730a525c979d990bb9d03a8b/api/envoy/config/core/v3/base.proto#L142). And I want to know why the Envoy Node info is needed in your case first.

bbassingthwaite:
Hi @tyxia thanks for the response. The RQLS management server will receive reports from many Envoy nodes across various regions, clusters and tenants. In our use case, the Envoy Node contains all of this information and allows us to distinguish the tenant/region/cluster and apply the appropriate rate limiting policies. 

We could encode this information in the bucket but it's quite redundant and would need to live on every bucket. If `RateLimitQuotaUsageReports` include the Node on the first request, we could simply store it and know that every subsequent report that came in would be for that tenant/region/cluster. It would also reduce the size of the RateLimitQuotaUsageReports since we can send that information once and not need to include it on every report.

This behaviour is quite common across the xDS APIs, such as ADS, ALS, Metrics Service, Trace Service, etc...

tyxia:
@bbassingthwaite  Thanks for sharing the information and use case from your side!
For our(i.e., Google) cases, we currently don't need this information as I describe above because we are operating on quota bucket basis and don't really care about which Envoy Node. We could use it later (e.g., logging).

[...]

PR #24014, opened 2022-11-15
rate_limit_quota: [API] Support Envoy node info in `RateLimitQuotaUsageReports`

tyxia:
Even though our (Google) RLQS server doesn't need this, there are needs/use cases from other RLQS customers that the RQLS server will receive reports from many envoy nodes across various regions, clusters and tenants. And the customers will use this Envoy node which contains all of this information to distinguish the tenant/region/cluster and to apply the appropriate rate limiting policies. 

This PR addresses #23585. 

Risk Level: LOW
Testing: N/A

### DIFF
--- a/api/envoy/service/rate_limit_quota/v3/rlqs.proto
+++ b/api/envoy/service/rate_limit_quota/v3/rlqs.proto
@@ -64,7 +64,6 @@ message RateLimitQuotaUsageReports {
   // .. note::
   //   Note that the first report sent for a ``BucketId`` indicates to the RLQS server that
   //   the RLQS client is subscribing for the future assignments for this ``BucketId``.
-  // [#next-free-field: 5]
   message BucketQuotaUsage {
     // ``BucketId`` for which request quota usage is reported.
     BucketId bucket_id = 1 [(validate.rules).message = {required: true}];
--- a/api/envoy/service/rate_limit_quota/v3/rlqs.proto
+++ b/api/envoy/service/rate_limit_quota/v3/rlqs.proto
@@ -2,6 +2,7 @@ syntax = "proto3";
 
 package envoy.service.rate_limit_quota.v3;
 
+import "envoy/config/core/v3/base.proto";
 import "envoy/type/v3/ratelimit_strategy.proto";
 
 import "google/protobuf/duration.proto";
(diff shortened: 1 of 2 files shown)


api/envoy/service/rate_limit_quota/v3/rlqs.proto
```diff
@@ -78,6 +80,9 @@ message RateLimitQuotaUsageReports {
 
     // Requests throttled.
     uint64 num_requests_denied = 4;
+
+    // An optional info for the list of Envoy instances that send the usage report ``BucketQuotaUsage``.
```
htuch, line 84: I'm not sure if you want to send this on every request, this message can get substantially large CC @yanavlasov

tyxia: Yes, You are right. This is designed/recommended to be only on first report (similar to the domain name [here](https://github.com/envoyproxy/envoy/blob/879ff9cb7e6cd522138f11e3b10f96dd5ae8f67d/api/envoy/service/rate_limit_quota/v3/rlqs.proto#L95) )

I was on the fence of adding this recommendation to the code comment as I was thinking this eventually is really up to the customer who set this field.  But on second thought, let me add this to the comment to make it clear.

tyxia:
/retest

deflake

tyxia:
@htuch  Thanks for the review! PTAL.

(The CI failure is caused by CI/CD infra that one agent is not responding... I will re-test it before submitting this PR).

htuch (commented):
/lgtm api

htuch:
@tyxia do you want to fold in the implementation PR into this one? I'm pretty sure that combined this is a small PR.

htuch:
Yep, I suggest merging in the implementation with this PR and let's review together.

tyxia:
/wait

I will update this PR once other parts of RLQS are ready.

tyxia:
I will work on this once other parts of impl are merged. Then I will merge corresponding impl into this proto change, as Harvey suggested.

zuercher:
Is this PR still in progress?

tyxia:
> Is this PR still in progress?

Yes, this PR is currently on-hold to wait for other pieces of project to be merged first. But I will revisit this PR soon.

PR #24015, opened 2022-11-15
transport socket: trigger onConnect for downstream transport sockets

kyessenov:

Commit Message: Trigger downstream transport socket `onConnect`. Server connections start out as socket connected, but not necessarily transport socket connected. The server socket may need this event to trigger the transport socket connected event (e.g. raw_buffer). 
Some server downstream transport sockets may share implementation with the upstream transport socket and not expect this event, in which case these custom downstream transport sockets may need to ignore this event after this change.

Risk Level: medium (for downstream integrations, no functional change in this repo)
Testing: done
Docs Changes: none
Release Notes:
Issue: https://github.com/envoyproxy/envoy/issues/9023

### DIFF
--- a/source/extensions/transport_sockets/alts/tsi_socket.cc
+++ b/source/extensions/transport_sockets/alts/tsi_socket.cc
@@ -11,13 +11,14 @@ namespace TransportSockets {
 namespace Alts {
 
 TsiSocket::TsiSocket(HandshakerFactory handshaker_factory, HandshakeValidator handshake_validator,
-                     Network::TransportSocketPtr&& raw_socket)
+                     Network::TransportSocketPtr&& raw_socket, bool downstream)
     : handshaker_factory_(handshaker_factory), handshake_validator_(handshake_validator),
-      raw_buffer_socket_(std::move(raw_socket)) {}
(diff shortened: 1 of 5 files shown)


kyessenov:
Split out of https://github.com/envoyproxy/envoy/pull/23944.

source/extensions/transport_sockets/alts/tsi_socket.cc
```diff
@@ -361,7 +362,9 @@ void TsiSocket::closeSocket(Network::ConnectionEvent) {
 
 void TsiSocket::onConnected() {
   ASSERT(!handshake_complete_);
-  doHandshakeNext();
+  if (!downstream_) {
+    doHandshakeNext();
+  }
```
mattklein123, line 367: Can you add some comments on why we defer this now? It's not immediately clear to me why we need this change here but not other transport sockets? Thank you.

/wait

kyessenov: This was never called on the downstream. Other sockets don't do work here AFAICT.

mattklein123: I see, OK.

mattklein123 (approved):
LGTM, thanks.

PR #24023, opened 2022-11-16
attr: support upstream.cluster

zirain:
cc @kyessenov 

Commit Message:
Additional Description: support `source.cluster` for common attributes
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/common/expr/context.cc
+++ b/source/extensions/filters/common/expr/context.cc
@@ -246,9 +246,9 @@ absl::optional<CelValue> UpstreamWrapper::operator[](CelValue key) const {
         info_.upstreamInfo().value().get().upstreamTransportFailureReason());
   } else if (value == UpstreamClusterName) {
     auto upstreamClusterInfo = info_.upstreamClusterInfo();
-    if (upstreamClusterInfo.has_value() && upstreamClusterInfo.value() != nullptr) {
+    //if (upstreamClusterInfo.has_value() && upstreamClusterInfo.value() != nullptr) {
       return CelValue::CreateStringView(upstreamClusterInfo.value()->name());
-    }
+    //}
   }
 
   auto ssl_info = info_.upstreamInfo().value().get().upstreamSslConnection();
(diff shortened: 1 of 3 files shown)


kyessenov:
I think this is reasonable but it would be the first instance we're exposing xDS symbols into https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes. Maybe we need some top-level namespace like `xds.cluster_name` to distinguish configuration from raw observations? Thoughts @wjtracey ?

zirain:
> I think this is reasonable but it would be the first instance we're exposing xDS symbols into https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes. Maybe we need some top-level namespace like `xds.cluster_name` to distinguish configuration from raw observations? Thoughts @wjtracey ?

glad to see some top-level design

kyessenov:
@zirain How about the following proposal: we add a top-level `xds` object with the configuration metadata about the request/connection.  This clearly marks the difference between the configuration and the data plane attributes, which I think is a very desirable property.

We can promote majority of the Wasm-specific attributes this way:

| Attribute | Description |
| ------------- | ----------------- |
| xds.node | Boostrap node protobuf |
| xds.cluster_name | Upstream cluster name |
| xds.cluster_metadata | Upstream cluster metadata |
| xds.listener_direction | Enum as int value of listener direction |
| xds.listener_name | Listener name |
| xds.listener_metadata | Listener metadata |
| xds.route_name | Route name |
| xds.route_metadata | Route metadata |
| xds.upstream_host_metadata | Upstream host metadata |

I would suggest defining a fake attribute bag for `xds`, then registering it in Wasm and the default contexts. We need to document them in https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes as part of the PR.

zirain:
sound good to me, if no objection, I can refactor my implement.

zirain:
as talk on slack, @kyessenov is familiar with CEL, and will take over this PR.

PR #24024, opened 2022-11-16
[Bandwidth-Limit] Fix bandwidth delay calculation logic

zhiyong-gayang:
The original bandwidth delay use envoy histogram metric which include the data moving time, when payload is large, the original logic is not accurate, fix this bug by only calculate the waiting time for available tokens.

Risk Level: Low

### DIFF
--- a/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
+++ b/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
@@ -75,7 +75,7 @@ message BandwidthLimit {
   //   If set true, the following 4 trailers will be added, prefixed by ``response_trailer_prefix``:
   //   * bandwidth-request-delay-ms: delay time in milliseconds it took for the request stream transfer including request body transfer time and the time added by the filter.
   //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer including response body transfer time and the time added by the filter.
-  //   * bandwidth-request-filter-delay-ms: delay time in milliseconds that added by the filter.
+  //   * bandwidth-request-filter-delay-ms: delay time in milliseconds in request stream transfer added by the filter.
   //   * bandwidth-response-filter-delay-ms: delay time in milliseconds that added by the filter.
   //   * If :ref:`enable_mode <envoy_v3_api_field_extensions.filters.http.bandwidth_limit.v3.BandwidthLimit.enable_mode>` is ``DISABLED`` or ``REQUEST``, the trailers will not be set.
   //   * If both the request and response delay time is 0, the trailers will not be set.
--- a/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
+++ b/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
@@ -77,7 +77,6 @@ message BandwidthLimit {
   //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer including response body transfer time and the time added by the filter.
   //   * bandwidth-request-filter-delay-ms: delay time in milliseconds in request stream transfer added by the filter.
   //   * bandwidth-response-filter-delay-ms: delay time in milliseconds that added by the filter.
-  //
   //   If :ref:`enable_mode <envoy_v3_api_field_extensions.filters.http.bandwidth_limit.v3.BandwidthLimit.enable_mode>` is ``DISABLED`` or ``REQUEST``, the trailers will not be set.
   //   If both the request and response delay time is 0, the trailers will not be set.
   //
--- a/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
+++ b/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
@@ -74,7 +74,7 @@ message BandwidthLimit {
   //
   //   If set true, the following 4 trailers will be added, prefixed by ``response_trailer_prefix``:
   //   * bandwidth-request-delay-ms: delay time in milliseconds it took for the request stream transfer including request body transfer time and the time added by the filter.
-  //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer including response body transfer time the time added by the filter.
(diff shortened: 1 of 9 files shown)


zhiyong-gayang:
> @zhiyong-gayang thanks for the PR. Just to confirm you are trying to set the trailer to the time that filter had added to the body transfer time. Right now the time in the trailer is both the time it took to receive the body from the peer including the time added by the filter.
> 
> I think the [doc](https://github.com/envoyproxy/envoy/blob/1cf5603dc5239c92e5bc38ef321f59ccf6eabc6e/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto#L76) is somewhat ambiguous about what the value signifies, however I would read it as the total time of proxying the request (not just the time added by the filter).
> 
> Given that this is externally observable change I would suggest adding a new trailer that just has the time added by the filter. Adding other maintainers for comments: @nitgoy @mattklein123 @tonya11en
> 
> /wait-any

The two trailers were added to measure the delays that added by filter, which is the total waiting time due to bandwidth throttling, but was not aware of the body receiving time, I'm ok to add a new one if the current one is still useful.
How about `bandwidth-request-filter-delay-ms` and `bandwidth-response-filter-delay-ms`?

nitgoy:
> 
> I think the [doc](https://github.com/envoyproxy/envoy/blob/1cf5603dc5239c92e5bc38ef321f59ccf6eabc6e/api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto#L76) is somewhat ambiguous about what the value signifies, however I would read it as the total time of proxying the request (not just the time added by the filter).
> 
@yanavlasov From the trailer names, the intention seems to be to provide the 'delay' added just by the bandwidth filter. I believe that the cumulative transfer time can already be retrieved from other places. In that sense, the doc might also need an update to make it more clear.

> Given that this is externally observable change I would suggest adding a new trailer that just has the time added by the filter. Adding other maintainers for comments: @nitgoy @mattklein123 @tonya11en
> 
The word 'delay' in the name can be a little confusing if we keep the current behavior for them. Not sure if it's worth introducing a breaking change to avoid that confusion. Other than that I agree to add new trailer.

yanavlasov:
New trailers is the best approach here, to avoid user visible behavior change. The names you are suggesting look good to me.

yanavlasov (approved):

yanavlasov:
Waiting for API approval.

zhiyong-gayang:
@lizan could you help take a look this PR? thanks.

yanavlasov:
It looks like CI is struck. Can you please merge main? I have also pinged maintainers for API approval.

api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
```diff
@@ -72,9 +72,11 @@ message BandwidthLimit {
   //
   // .. note::
   //
-  //   * If set true, the response trailers ``bandwidth-request-delay-ms`` and ``bandwidth-response-delay-ms`` will be added, prefixed by ``response_trailer_prefix``.
-  //   * bandwidth-request-delay-ms: delay time in milliseconds it took for the request stream transfer.
-  //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer.
+  //   * If set true, the following 4 trailers will be added, prefixed by ``response_trailer_prefix``.
+  //   * bandwidth-request-delay-ms: delay time in milliseconds it took for the request stream transfer including request body transfer time and the time added by the filter.
+  //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer including response body transfer time the time added by the filter.
```
adisuissa, line 77:
```suggestion
  //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer including response body transfer time and the time added by the filter.
```

api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
```diff
@@ -72,9 +72,11 @@ message BandwidthLimit {
...
+  //   * bandwidth-request-filter-delay-ms: delay time in milliseconds that added by the filter.
+  //   * bandwidth-response-filter-delay-ms: delay time in milliseconds that added by the filter.
```
adisuissa, line 79: The bullets after this one should not be bullet points, or be separated from the above 4.

adisuissa (commented):
Thanks, left a couple of nits from the doc side (you can look at the current state [here](https://storage.googleapis.com/envoy-pr/61a1b3b/docs/api-v3/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto.html?highlight=bandwidth_limit#extensions-filters-http-bandwidth-limit-v3-bandwidthlimit))

api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
```diff
@@ -72,9 +72,11 @@ message BandwidthLimit {
   //
   // .. note::
   //
-  //   * If set true, the response trailers ``bandwidth-request-delay-ms`` and ``bandwidth-response-delay-ms`` will be added, prefixed by ``response_trailer_prefix``.
-  //   * bandwidth-request-delay-ms: delay time in milliseconds it took for the request stream transfer.
-  //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer.
+  //   * If set true, the following 4 trailers will be added, prefixed by ``response_trailer_prefix``.
+  //   * bandwidth-request-delay-ms: delay time in milliseconds it took for the request stream transfer including request body transfer time and the time added by the filter.
+  //   * bandwidth-response-delay-ms: delay time in milliseconds it took for the response stream transfer including response body transfer time the time added by the filter.
+  //   * bandwidth-request-filter-delay-ms: delay time in milliseconds that added by the filter.
```
nitgoy, line 78:
```suggestion
  //   * bandwidth-request-filter-delay-ms: delay time in milliseconds in request stream transfer added by the filter.
```

Similarly for the response trailer description

api/envoy/extensions/filters/http/bandwidth_limit/v3/bandwidth_limit.proto
zhiyong-gayang: updated.

Issue #23844, opened 2022-11-04
Support for Month, Year rate limit units for business subscriptions


renuka-fernando:
*Title*: *Support for Month, Year rate limit units for business subscriptions*

*Description*:
It is possible to have business cases with per Month, per Year rate limits (for global rate limiting). These rate limit policies are not to protect upstream from attacks like DDoS, but to have some business advantages like a monthly subscription or yearly subscription for a customer to use the upstream.

Following are some sample subscriptions for an upstream.
- 10k requests per Month
- 1M requests per Year

The existing implementation only supports the following units. For a business subscription, it is worth having month and year as units.

- SECOND
- MINUTE
- HOUR
- DAY

[optional *Relevant Links*:]
https://github.com/envoyproxy/data-plane-api/blob/5ba669d8a54e7c4bbbba3a6445ebd1a09d9e9dc0/envoy/service/ratelimit/v3/rls.proto#L80-L95

renuka-fernando:
@tyxia

tyxia:
@renuka-fernando  Let's continue our offline discussion here so that anyone has opinion about this request can chime in and comment.

For us, `month` and `year` unit are too coarse-grained unit and we want to control rate liming more tightly on a more frequent basis.

With that said, I am open to add this in the API for your business need. But out of curiosity: 1) It seems that it will be huge number to report for `month` and `year` unless the traffic to your upstream is small. and 2) you will have your Envoy instance  to run for month or year with same configuration?

renuka-fernando:
@tyxia thank you for your comment.

> 1) It seems that it will be a huge number to report for `month` and `year` unless the traffic to your upstream is small

Yes, the traffic to the upstream will be small, and planning to control it with another rate limit policy applied together with this monthly/yearly subscription.

For example:
Route: /weather/v1/location - (10K Per Month + 20 Per Min)

So downstream clients can't use the whole 10K within a short period of time. it is controlled with "20 Per Min", the second rate limit policy (as a spike arrest rate limit policy).

> 2) you will have your Envoy instance run for a month or year with the same configuration?

Yes, the Route exists for months and years. But maybe have some changes like adding new Routes. Deprecate this Route and later delete it. Update the subscription from 10K Per Month to 20K Per Month.

renuka-fernando:
Hi @tyxia,
Thank you very much for working on this.

tyxia:
> Hi @tyxia, Thank you very much for working on this.

[...]

PR #24031, opened 2022-11-16
rate_limit_quota: [API] Add time unit `month` and `year`

tyxia:
Even though for us (who introduced the rate_limit_quota feature), `month` and `year` unit are too coarse-grained unit as our traffic will be huge and we will update the Envoy config on more frequent basis, there is business need/case for monthly and yearly subscription from @renuka-fernando side.  
More detailed discussion can be found #23844.

### DIFF
--- a/api/envoy/type/v3/ratelimit_unit.proto
+++ b/api/envoy/type/v3/ratelimit_unit.proto
@@ -28,4 +28,10 @@ enum RateLimitUnit {
 
   // The time unit representing a day.
   DAY = 4;
+
+  // The time unit representing a month.
+  MONTH = 5;
+
+  // The time unit representing a year.
+  YEAR = 6;
 }


renuka-fernando:
I think we have to update this function too. But Month and Year do not have exact seconds, it varies from month to month, year to year, but I wish we do not want exact seconds here. @tyxia wdyt?

https://github.com/envoyproxy/envoy/blob/5c56e45a041a4446411a8dccb5262b12323ff9be/source/extensions/filters/http/ratelimit/ratelimit_headers.cc#L67-L82

api/envoy/type/v3/ratelimit_unit.proto
```diff
@@ -28,4 +28,10 @@ enum RateLimitUnit {
 
   // The time unit representing a day.
   DAY = 4;
+
+  // The time unit representing a month.
+  MONTH = 5;
```
adisuissa, line 33: Should these fields be processed by some code?
I'm assuming the code changes aren't too large, so I suggest adding the code as part of this PR.

tyxia:
> I think we have to update this function too. But Month and Year do not have exact seconds, it varies from month to month, year to year, but I wish we do not want exact seconds here. @tyxia wdyt?
> 
> https://github.com/envoyproxy/envoy/blob/5c56e45a041a4446411a8dccb5262b12323ff9be/source/extensions/filters/http/ratelimit/ratelimit_headers.cc#L67-L82

I think there is misunderstanding here. So do you want to include the month and year to existing rate limiting feature (api/envoy/service/ratelimit/v3/rls.proto)  or the new global rate limiting feature I am working on? The PR here is for new rate limiting feature will not use the code you mentioned above.

api/envoy/type/v3/ratelimit_unit.proto
tyxia: Currently there is no impl code for this yet. But let's wait for the clarification from @renuka-fernando above first.

renuka-fernando:
> > I think we have to update this function too. But Month and Year do not have exact seconds, it varies from month to month, year to year, but I wish we do not want exact seconds here. @tyxia wdyt?
> > https://github.com/envoyproxy/envoy/blob/5c56e45a041a4446411a8dccb5262b12323ff9be/source/extensions/filters/http/ratelimit/ratelimit_headers.cc#L67-L82
> 
> I think there is misunderstanding here. So do you want to include the month and year to existing rate limiting feature (api/envoy/service/ratelimit/v3/rls.proto) or the new global rate limiting feature I am working on? Or both?
> 
> The PR here is for new rate limiting feature and will not use the code you mentioned above.

I am planning to use the existing rate limit feature (with the Envoy Rate-Limit service: https://github.com/envoyproxy/ratelimit). I think what you refer to as the new global rate limit feature is RLQS, that @markdroth mentioned in this comment https://github.com/envoyproxy/ratelimit/pull/368#pullrequestreview-1142536075.

It is grateful if it is possible to have these units in the existing RLS feature too.
So both if possible 🙂.

tyxia:
> > > I think we have to update this function too. But Month and Year do not have exact seconds, it varies from month to month, year to year, but I wish we do not want exact seconds here. @tyxia wdyt?
> > > https://github.com/envoyproxy/envoy/blob/5c56e45a041a4446411a8dccb5262b12323ff9be/source/extensions/filters/http/ratelimit/ratelimit_headers.cc#L67-L82
> > 
> > 
> > I think there is misunderstanding here. So do you want to include the month and year to existing rate limiting feature (api/envoy/service/ratelimit/v3/rls.proto) or the new global rate limiting feature I am working on? Or both?
> > The PR here is for new rate limiting feature and will not use the code you mentioned above.
> 
> I am planning to use the existing rate limit feature (with the Envoy Rate-Limit service: https://github.com/envoyproxy/ratelimit). I think what you refer to as the new global rate limit feature is RLQS, that @markdroth mentioned in this comment [envoyproxy/ratelimit#368 (review)](https://github.com/envoyproxy/ratelimit/pull/368#pullrequestreview-1142536075).
> 
> It is grateful if it is possible to have these units in the existing RLS feature too. So both if possible 🙂.

Thanks for confirmation! The changes to existing RLS will be done in a separate PR.

api/envoy/type/v3/ratelimit_unit.proto
tyxia: @adisuissa , just heard back from @renuka-fernando that he prefer to have it in both rateLimiting designs.
So maybe we can proceed with this one (new RLQS) which doesn't have the impl yet as its baseline imp PR is pending review ?(also, imp for this particular part might not be prioritized to early phase). WDYT?

adisuissa: SGTM.

adisuissa (approved):
LGTM.
/lgtm api

adisuissa:
As discussed offline, this should be followed by a PR to update the previously defined [Unit](https://github.com/envoyproxy/envoy/blob/a1f8fa57132e286d8a8c26b87a1f6e8773348175/api/envoy/service/ratelimit/v3/rls.proto#L80) enum.

PR #24035, opened 2022-11-17
jwt_authn: adjust remote_jwks.async_fetch refetch timer

qiwzhang:

This is just an improvement to async_fetch feature of remote jwks.

*  For a successful fetch,  adjust its re_fetch time to be the jwks_cache_duration - 5s.   This is to avoid of triggering on-demand fetching.   During authentication,  if jwks is not fetched, or is expired,  it will trigger an on-demand fetching.  If async_fetch can refetch the jwks a little bit earlier, it will avoid the on-demand fetch. 

* For a failed fetch,  trigger a re_fetch after 1 second.  Getting jwks is important.  We should keep retrying.  This retrying is after [remote_jwks.retry_policy](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto#L366).  

Risk Level: Low, added more frequence refetch for a failed remote jwks uri. 
Testing:  unit-tested
Docs Changes: None
Release Notes: Yes

### DIFF
--- a/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
+++ b/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
@@ -314,7 +314,7 @@ message RemoteJwks {
   config.core.v3.HttpUri http_uri = 1;
 
   // Duration after which the cached JWKS should be expired. If not specified, default cache
-  // duration is 5 minutes.
+  // duration is 10 minutes.
   google.protobuf.Duration cache_duration = 2;
 
   // Fetch Jwks asynchronously in the main thread before the listener is activated.
--- a/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
+++ b/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
@@ -377,7 +377,7 @@ message JwksAsyncFetch {
   // Default is false.
   bool fast_listener = 1;
 
-  // The duration to refetch after a failed fetch. If not specified, default is 1 secnod.
+  // The duration to refetch after a failed fetch. If not specified, default is 1 second.
   google.protobuf.Duration failed_refetch_duration = 2;
 }
 
--- a/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
+++ b/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
@@ -376,6 +376,9 @@ message JwksAsyncFetch {
   // If true, it is activated without waiting for the initial fetch to complete.
   // Default is false.
   bool fast_listener = 1;
+
+  // The duration to refetch after a failed fetch. If not specified, default is 1 secnod.
+  google.protobuf.Duration failed_refetch_duration = 2;
 }
 
 // This message specifies a header location to extract JWT token.
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -34,7 +34,7 @@ minor_behavior_changes:
     add ``MONTH`` and ``YEAR`` to the unit of time for rate limit.
 - area: jwt_authn
   change: |
-    adjust the refetch time for remote_jwks async_fetch feature, 1 second after a failed fetch, refetch 5 seconds before jwks cache duration for a good fetch.
+    adjust the refetch time for remote_jwks async_fetch feature. For a good fetch, refetch 5 seconds before jwks cache duration. For a failed fetch, refetch time can be specified by :ref:`failed_refetch_duration <envoy_v3_api_field_extensions.filters.http.jwt_authn.v3.JwksAsyncFetch.failed_refetch_duration>` with default 1 second.
 
 bug_fixes:
 # *Changes expected to improve the state of the world and are unlikely to have negative effects*
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -32,6 +32,9 @@ minor_behavior_changes:
 - area: rate_limit
   change: |
     add ``MONTH`` and ``YEAR`` to the unit of time for rate limit.
+- area: jwt_authn
+  change: |
+    adjust the refetch time for remote_jwks async_fetch feature, 1 second after a failed fetch, refetch 5 seconds before jwks cache duration for a good fetch.
 
 bug_fixes:
 # *Changes expected to improve the state of the world and are unlikely to have negative effects*
--- a/source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
+++ b/source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
@@ -12,13 +12,13 @@ namespace JwtAuthn {
 namespace {
 
 // Default cache expiration time in 5 minutes.
-constexpr int DefaultCacheExpirationSec = 600;
+constexpr std::chrono::seconds DefaultCacheExpirationSec{600};
 
 // Number of seconds to refetch before a cached jwks is expired.
-constexpr int RefetchBeforeExpiredSec = 5;
+constexpr std::chrono::seconds RefetchBeforeExpiredSec{5};
 
 // Number of seconds to refetch after a failed fetch.
-constexpr int RefetchAfterFailedSec = 1;
+constexpr std::chrono::seconds RefetchAfterFailedSec{1};
 
 } // namespace
 
@@ -41,9 +41,8 @@ JwksAsyncFetcher::JwksAsyncFetcher(const RemoteJwks& remote_jwks,
   // and not need to block request processing but on-demand fetch is done in the worker
   // thread and block request processing.
   refetch_duration_ = getCacheDuration(remote_jwks);
-  std::chrono::seconds before_expire(RefetchBeforeExpiredSec);
-  if (refetch_duration_ > before_expire) {
-    refetch_duration_ = refetch_duration_ - before_expire;
+  if (refetch_duration_ > RefetchBeforeExpiredSec) {
+    refetch_duration_ = refetch_duration_ - RefetchBeforeExpiredSec;
   }
 
   refetch_timer_ = context_.mainThreadDispatcher().createTimer([this]() -> void { fetch(); });
@@ -63,7 +62,7 @@ std::chrono::seconds JwksAsyncFetcher::getCacheDuration(const RemoteJwks& remote
   if (remote_jwks.has_cache_duration()) {
     return std::chrono::seconds(DurationUtil::durationToSeconds(remote_jwks.cache_duration()));
   }
-  return std::chrono::seconds(DefaultCacheExpirationSec);
+  return DefaultCacheExpirationSec;
 }
 
 void JwksAsyncFetcher::fetch() {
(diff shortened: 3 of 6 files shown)


source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
```diff
@@ -12,7 +12,13 @@ namespace JwtAuthn {
 namespace {
 
 // Default cache expiration time in 5 minutes.
-constexpr int PubkeyCacheExpirationSec = 600;
+constexpr int DefaultCacheExpirationSec = 600;
```
lizan, line 15: style, for all constants:
```suggestion
constexpr std::chrono::seconds DefaultCacheExpiration{600};
```

qiwzhang: Done

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
```diff
@@ -12,7 +12,13 @@ namespace JwtAuthn {
 namespace {
 
 // Default cache expiration time in 5 minutes.
```
suniltheta, line 14: Default cache expiration time in ~~5~~ 10 minutes ?

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
```diff
@@ -21,15 +27,25 @@ JwksAsyncFetcher::JwksAsyncFetcher(const RemoteJwks& remote_jwks,
...
-      context_.mainThreadDispatcher().createTimer([this]() -> void { fetch(); });
+  // Next refetch time should be: jwks_cache_duration - RefetchBeforeExpiredSec.
+  // Using RefetchBeforeExpiredSec is to avoid on_demand fetch.
+  // During the authentication, if jwks is not fetched or is expired, it will trigger
+  // an on-demand fetch. But async_fetch is preferred as it is done in the main thread
+  // and not need to block request processing but on-demand fetch is done in the worker
+  // thread and block request processing.
+  refetch_duration_ = getCacheDuration(remote_jwks);
+  if (refetch_duration_ > RefetchBeforeExpiredSec) {
+    refetch_duration_ = refetch_duration_ - RefetchBeforeExpiredSec;
```
suniltheta, line 45: I am trying to understand the behavior here.

Let's say we configure the `remote_jwks.cache_duration` to 6 seconds then the `refetch_timer_` will be run for every for every 1 seconds (6 seconds - 5 seconds).

If we configure `remote_jwks.cache_duration` to 5 seconds then the `refetch_timer_` will be run for every 5 seconds. From documentation they will not be able to know this change in frequency.


Let's assume for practical purpose we would typically not set a cache duration of 10-5 seconds. But I am thinking how a user would make sense of this difference in behavior. Perhaps we should make that `RefetchBeforeExpiredSec` configurable, and let user know that if `cache_duration` is low (for Eg: 6 seconds), then also configure `RefetchBeforeExpiredSec` to something low (Eg: 1 second) so that fetch happens for every 5 seconds instead of 1 second

suniltheta (commented):
couple of clarifications

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
suniltheta: Need to fix the documentation also. https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto.html?highlight=jwt#extensions-filters-http-jwt-authn-v3-remotejwks

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
```diff
@@ -86,6 +101,7 @@ void JwksAsyncFetcher::onJwksSuccess(google::jwt_verify::JwksPtr&& jwks) {
 void JwksAsyncFetcher::onJwksError(Failure) {
   ENVOY_LOG(warn, "{}: failed", debug_name_);
   handleFetchDone();
+  refetch_timer_->enableTimer(RefetchAfterFailedSec);
```
suniltheta, line 104: checking if this will do a retry back off and where the retry_policy is implemented for RemoteJwks

qiwzhang: OnJwksError will be called from after retry failed in [jwks_fetcher](https://github.com/envoyproxy/envoy/blob/9f80d2c2c4a561dfb8b65a7fb27589ea1bc0abeb/source/extensions/filters/http/common/jwks_fetcher.cc#L68)

So this retry is on top of this [retry_policy](https://github.com/envoyproxy/envoy/blob/287ccb9daafb3df3f42f6d83bbffd965c49cddb9/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto#L366)

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
qiwzhang: Logically, you are correct,  there is a big jump between 1s retry for 6s cache_duration vs 4s retry for 4s cache_duration.  But I feel that normally cache_duration is around 10 minute or even hours,  rarely around 5 second range.    So spend too much time on such rare corner case is over kill.   We don't want to make thing too complicated with so many knots.  What do you think?

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
qiwzhang: Thanks for catching. Will update

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
suniltheta: understood

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
```diff
@@ -376,6 +376,9 @@ message JwksAsyncFetch {
   // If true, it is activated without waiting for the initial fetch to complete.
   // Default is false.
   bool fast_listener = 1;
+
+  // The duration to refetch after a failed fetch. If not specified, default is 1 secnod.
```
suniltheta, line 380: nit: typo in second

changelogs/current.yaml
```diff
@@ -32,6 +32,9 @@ minor_behavior_changes:
 - area: rate_limit
   change: |
     add ``MONTH`` and ``YEAR`` to the unit of time for rate limit.
+- area: jwt_authn
+  change: |
+    adjust the refetch time for remote_jwks async_fetch feature, 1 second after a failed fetch, refetch 5 seconds before jwks cache duration for a good fetch.
```
suniltheta, line 37: should we update this based on `failed_refetch_duration` ?

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
```diff
@@ -11,8 +12,22 @@ namespace HttpFilters {
...
-// Default cache expiration time in 5 minutes.
-constexpr int PubkeyCacheExpirationSec = 600;
+// Default cache expiration time in 10 minutes.
+constexpr std::chrono::seconds DefaultCacheExpirationSec{600};
+
+// Number of seconds to refetch before a cached jwks is expired.
+constexpr std::chrono::seconds RefetchBeforeExpiredSec{5};
+
+// Number of seconds to refetch after a failed fetch.
+constexpr std::chrono::seconds RefetchAfterFailedSec{1};
```
suniltheta, line 22: this gets renamed to `DefaultRefetchAfterFailedSec` ?

And 

`// Default number of seconds before we refetch after a failed fetch` something on these lines?

test/extensions/filters/http/jwt_authn/filter_integration_test.cc
```diff
@@ -43,6 +43,9 @@ std::string getAsyncFetchFilterConfig(const std::string& config_str, bool fast_l
   auto& provider0 = (*proto_config.mutable_providers())[std::string(ProviderName)];
   auto* async_fetch = provider0.mutable_remote_jwks()->mutable_async_fetch();
   async_fetch->set_fast_listener(fast_listener);
+  // Set failed_refetch_duration to a big value to disable failed refetch.
+  // as it interfares the two FailedAsyncFetch integration tests.
```
suniltheta, line 47: nit typo: interfere with the

test/extensions/filters/http/jwt_authn/jwks_async_fetcher_test.cc
```diff
@@ -193,7 +191,9 @@ TEST_P(JwksAsyncFetcherTest, TestGoodFetchAndRefresh) {
   EXPECT_EQ(out_jwks_array_.size(), 1);
 
   // Expect refresh timer is enabled.
-  EXPECT_CALL(*timer_, enableTimer(expected_duration_, nullptr));
+  const std::chrono::milliseconds expected_refetch_time =
+      JwksAsyncFetcher::getCacheDuration(config_) - std::chrono::seconds(5);
```
suniltheta, line 195: should we create a local variable RefetchBeforeExpiredSec for `std::chrono::seconds(5)` ?
So that we know how that value of 5 is set.

test/extensions/filters/http/jwt_authn/jwks_async_fetcher_test.cc
```diff
@@ -225,7 +225,8 @@ TEST_P(JwksAsyncFetcherTest, TestNetworkFailureFetchAndRefresh) {
   EXPECT_EQ(out_jwks_array_.size(), 0);
 
   // Expect refresh timer is enabled.
-  EXPECT_CALL(*timer_, enableTimer(expected_duration_, nullptr));
+  const std::chrono::milliseconds expected_refetch_time = std::chrono::seconds(1);
```
suniltheta, line 228: possible to add a test where this value comes from config?

test/extensions/filters/http/jwt_authn/filter_integration_test.cc
```diff
@@ -43,6 +43,9 @@ std::string getAsyncFetchFilterConfig(const std::string& config_str, bool fast_l
...
+  async_fetch->mutable_failed_refetch_duration()->set_seconds(60);
```
suniltheta, line 48: should we set this value to 10 minutes so that it mimics the existing default of cache_duration?

suniltheta (commented):
some minor comments

api/envoy/extensions/filters/http/jwt_authn/v3/config.proto
qiwzhang: Done

source/extensions/filters/http/jwt_authn/jwks_async_fetcher.cc
qiwzhang: Done

test/extensions/filters/http/jwt_authn/filter_integration_test.cc
qiwzhang: Done

test/extensions/filters/http/jwt_authn/filter_integration_test.cc
qiwzhang: Done

changelogs/current.yaml
qiwzhang: Done

test/extensions/filters/http/jwt_authn/jwks_async_fetcher_test.cc
qiwzhang: Done

test/extensions/filters/http/jwt_authn/jwks_async_fetcher_test.cc
qiwzhang: Thanks.  added,  it actually discovered a bug.

suniltheta (approved):
LGTM!

lizan (approved):

PR #24053, opened 2022-11-17
feat(hds): allow to configure upstream bind address for HDS

lukidzi:
Commit Message: feat(hds): allow to configure upstream bind address for hds
Additional Description: Sometimes when an application doesn't bind to localhost and the machine uses iptables for traffic redirection, HDS might need to override the source address for iptables to redirect traffic to the application instead of getting into the loop.
Risk Level: Low
Testing: small unit test configuration, upstream cluster tests binding functionality, manual testing
Docs Changes: API docs
Release Notes: 
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/upstream/health_discovery_service.cc
+++ b/source/common/upstream/health_discovery_service.cc
@@ -343,7 +343,7 @@ HdsCluster::HdsCluster(Server::Configuration::ServerFactoryContext& server_conte
   socket_match_hash_ = RepeatedPtrUtil::hash(cluster_.transport_socket_matches());
 
   info_ = info_factory.createClusterInfo(
-      {server_context, cluster_, bind_config_, stats_, ssl_context_manager_, added_via_api_, tls});
+      {server_context, cluster_, bind_config, stats_, ssl_context_manager_, added_via_api_, tls});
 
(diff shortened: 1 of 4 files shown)


jmarantz:
aside from the CI failure, @lizan do you have comments on this PR?

source/common/upstream/health_discovery_service.cc
```diff
@@ -393,7 +394,7 @@ void HdsCluster::update(envoy::config::cluster::v3::Cluster cluster,
     if (socket_match_hash_ != socket_match_hash) {
       socket_match_hash_ = socket_match_hash;
       update_cluster_info = true;
-      info_ = info_factory.createClusterInfo({server_context_, cluster_, bind_config_, stats_,
+      info_ = info_factory.createClusterInfo({server_context_, cluster_, bind_config, stats_,
```
lizan, line 397: with this change, we no longer need `bind_config_` member in HdsCluster?

lukidzi: Yes, you are right, they are defined in `Cluster`. I'll change it

lizan (approved):

Issue #21595, opened 2022-06-06
Tag extraction rules missing for many stats


ggreenway:
There are many stats for which default tag extraction rules should be written. The following list was generated from #21397. 

For each match below, rules should be added to `source/common/config/well_known_names.cc`, and the TODO should be deleted.

The list was generated with `grep -R -A 3 "TODO(ggreenway): add tag extraction rules." test/*`.

- [ ] test/extensions/filters/udp/dns_filter/dns_filter_integration_test.cc:    // TODO(ggreenway): add tag extraction rules.
test/extensions/filters/udp/dns_filter/dns_filter_integration_test.cc-    // Missing stat tag-extraction rule for stat 'dns_filter.my_prefix.local_a_record_answers' and
test/extensions/filters/udp/dns_filter/dns_filter_integration_test.cc-    // stat_prefix 'my_prefix'.

- [ ] test/extensions/filters/network/rbac/integration_test.cc:    // TODO(ggreenway): add tag extraction rules.
test/extensions/filters/network/rbac/integration_test.cc-    // Missing stat tag-extraction rule for stat 'tcp.shadow_denied' and stat_prefix 'tcp.'.

- [ ] test/extensions/filters/network/local_ratelimit/local_ratelimit_integration_test.cc:    // TODO(ggreenway): add tag extraction rules.
test/extensions/filters/network/local_ratelimit/local_ratelimit_integration_test.cc-    // Missing stat tag-extraction rule for stat
test/extensions/filters/network/local_ratelimit/local_ratelimit_integration_test.cc-    // 'http_local_rate_limiter.http_local_rate_limit.rate_limited' and stat_prefix
test/extensions/filters/network/local_ratelimit/local_ratelimit_integration_test.cc-    // 'http_local_rate_limiter'.

- [ ] test/extensions/filters/network/connection_limit/connection_limit_integration_test.cc:    // TODO(ggreenway): add tag extraction rules.
test/extensions/filters/network/connection_limit/connection_limit_integration_test.cc-    // Missing stat tag-extraction rule for stat
test/extensions/filters/network/connection_limit/connection_limit_integration_test.cc-    // 'connection_limit.connection_limit_stats.limited_connections' and stat_prefix
test/extensions/filters/network/connection_limit/connection_limit_integration_test.cc-    // 'connection_limit_stats'.

- [ ] test/extensions/filters/network/redis_proxy/redis_proxy_integration_test.cc:    // TODO(ggreenway): add tag extraction rules.
[...]

PR #24054, opened 2022-11-17
stats: properly extract the http.ext_authz stat_prefix

turettn:
Signed-off-by: Neal Turett <neal.turett@datadoghq.com>

## Standard fields

Commit Message: stats: properly extract the http.ext_authz stat_prefix
Risk Level: low
Testing:  added unit tests to confirm extraction works properly
Docs Changes: N/A
Release Notes: N/A

Fixes part of #21595.

### Normal behavior
When adding a stat_prefix to a filter, in general the emitted prometheus stat doesn't change, and instead a tag is added.  For example, adding a stat_prefix to a local rate limiter turns:
```
envoy_http_local_rate_limit_enabled{} 6
```
into
```
envoy_http_local_rate_limit_enabled{envoy_local_http_ratelimit_prefix="puppies"} 6
```

Likewise, adding a stat_prefix to a http_connection_manager turns:
```
envoy_http_downstream_cx_delayed_close_timeout{} 0
```
into
```
envoy_http_downstream_cx_delayed_close_timeout{envoy_http_conn_manager_prefix="admin"} 0
```

### Current behavior
It appears the stat_prefix extractor for http.ext_authz metrics was overlooked.  Added a stat_prefix to http.ext_authz turns:
```
envoy_http_ext_authz_ok{envoy_http_conn_manager_prefix="http"} 35459
```
into
```
envoy_http_ext_authz_foo_ok{envoy_http_conn_manager_prefix="http"} 35459
```

### Corrected behavior
This PR adds a tag extractor to so that the the output follows the pattern all the other stat_prefix supporting filters do:
```
envoy_http_ext_authz_ok{envoy_auth_prefix="foo",envoy_http_conn_manager_prefix="http"} 35459
```

### DIFF
--- a/source/common/config/well_known_names.cc
+++ b/source/common/config/well_known_names.cc
@@ -132,6 +132,11 @@ TagNameValues::TagNameValues() {
   // internal state of the regex which performs better.
   addRe2(HTTP_CONN_MANAGER_PREFIX, R"(^listener\..*?\.http\.((<TAG_VALUE>)\.))", ".http.");
 
+  // Extract ext_authz stat_prefix field
+  // cluster.[<cluster>.]ext_authz.[<ext_authz_prefix>.]*
+  // http.[<http_conn_mgr_prefix>.]ext_authz.[<ext_authz_prefix>.]*
+  addRe2(EXT_AUTHZ_PREFIX, R"(^(?:cluster|http)\.(?:[^\.]+\.)?ext_authz\.((<TAG_VALUE>)\.))");
+
   // http.(<stat_prefix>.)*
   addTokenized(HTTP_CONN_MANAGER_PREFIX, "http.$.**");
 
--- a/source/common/config/well_known_names.h
+++ b/source/common/config/well_known_names.h
@@ -137,6 +137,8 @@ public:
   const std::string RDS_ROUTE_CONFIG = "envoy.rds_route_config";
   // Request route given by the Router http filter
   const std::string ROUTE = "envoy.route";
+  // Stats prefix for the ext_authz HTTP filter
+  const std::string EXT_AUTHZ_PREFIX = "envoy.auth_prefix";
   // Listener manager worker id
   const std::string WORKER_ID = "envoy.worker_id";
   // Stats prefix for the Thrift Proxy network filter
--- a/test/common/stats/tag_extractor_impl_test.cc
+++ b/test/common/stats/tag_extractor_impl_test.cc
@@ -430,6 +430,15 @@ TEST(TagExtractorTest, DefaultTagExtractors) {
   redis_prefix.name_ = tag_names.REDIS_PREFIX;
   redis_prefix.value_ = "my_redis_prefix";
   regex_tester.testRegex("redis.my_redis_prefix.response", "redis.response", {redis_prefix});
+
+  // ExtAuthz Prefix
+  Tag ext_authz_prefix;
+  ext_authz_prefix.name_ = tag_names.EXT_AUTHZ_PREFIX;
+  ext_authz_prefix.value_ = "authpfx";
+  regex_tester.testRegex("http.http_prefix.ext_authz.authpfx.denied", "http.ext_authz.denied",
+                         {listener_http_prefix, ext_authz_prefix});
+  regex_tester.testRegex("cluster.grpc_cluster.ext_authz.authpfx.ok", "cluster.ext_authz.ok",
+                         {grpc_cluster, ext_authz_prefix});
 }
 
 TEST(TagExtractorTest, ExtractRegexPrefix) {


jmarantz:
On the surface this seems fine. However it's actually a breaking change, of sorts, and I wonder if we should runtime protect it.

@htuch @alyssawilk @ggreenway @krajshiva @yanavlasov WDYT?

In particular, these kinds of changes in the recent past have broken our internal monitoring and tests. They are fixable but I wonder what the best policy is for how to roll these out?

htuch:
My $0.02 is that this is a category of stability we have not done a great job at documenting historically - it's not data plane or config/control plane API but o11y naming stability. It would be nice to write up a policy here. In the meantime, I think sticking to bug fixes (which this is, not a gratuitous renaming for example) and treating this with a runtime guard seems a reasonable way to balance the competing interests.

alyssawilk:
SGTM (x2)

turettn:
Thanks everyone!  I will look at adding a runtime guard and figuring out why `test/integration/integration_test_shard_1/2` is failing on MacOS CI.

source/common/config/well_known_names.cc
```diff
@@ -132,6 +134,13 @@ TagNameValues::TagNameValues() {
   // internal state of the regex which performs better.
   addRe2(HTTP_CONN_MANAGER_PREFIX, R"(^listener\..*?\.http\.((<TAG_VALUE>)\.))", ".http.");
 
+  // Extract ext_authz stat_prefix field
+  // cluster.[<cluster>.]ext_authz.[<ext_authz_prefix>.]*
+  // http.[<http_conn_mgr_prefix>.]ext_authz.[<ext_authz_prefix>.]*
+  if (Runtime::runtimeFeatureEnabled("envoy.restart_features.ext_authz_prefix_handling")) {
+    addRe2(EXT_AUTHZ_PREFIX, R"(^(?:cluster|http)\.(?:[^\.]+\.)?ext_authz\.((<TAG_VALUE>)\.))");
```
jmarantz, line 141: From a performance perspective, the parenthesized condition at the start of the regex blocks an optimization, forcing this regex to be evaluated on every stat.

The optimization looks for regexes that start with `^foo.` and puts those in a map from "foo"->list(matchers), and only considers those matchers on stats starting with "foo".

I think you can easily rewrite the above regex into two of them that will enable this optimization. WDYT?

jmarantz (commented):
Nice, thank you!

There's a clang-tidy issue you need to dive into, and I have a performance suggestion.

changelogs/current.yaml
```diff
@@ -32,6 +32,10 @@ minor_behavior_changes:
 - area: rate_limit
   change: |
     add ``MONTH`` and ``YEAR`` to the unit of time for rate limit.
+- area: stats
+  change: |
+    Properly extract the ext_authz stat-prefix as a tag when exporting prometheus metrics.  This behavior can be reverted by setting runtime guard ``envoy.restart_features.ext_authz_prefix_handling`` to false.
```
jmarantz, line 37: @alyssawilk 
Should we be default this to false initially first? I haven't yet read the new policy :)

jmarantz (commented):
lgtm generally! I have a specific question for Alyssa but otherwise think it makes sense to have @ggreenway do the senior maintainer review for this.

source/common/config/well_known_names.cc
```diff
@@ -132,6 +134,14 @@ TagNameValues::TagNameValues() {
   // internal state of the regex which performs better.
   addRe2(HTTP_CONN_MANAGER_PREFIX, R"(^listener\..*?\.http\.((<TAG_VALUE>)\.))", ".http.");
 
+  // Extract ext_authz stat_prefix field
+  if (Runtime::runtimeFeatureEnabled("envoy.restart_features.ext_authz_prefix_handling")) {
```
ggreenway, line 138: Unless something has changed since last time I investigated this, the runtime guard will not work because this is resolved early during startup before the runtime is initialized/loaded/configured. For past changes, we've just documented the change and not added a runtime guard at all, because there wasn't a good way to add it.

turettn: Got it - I'll rip all the runtime guard bits out then.

changelogs/current.yaml
```diff
@@ -32,6 +32,9 @@ minor_behavior_changes:
 - area: rate_limit
   change: |
     add ``MONTH`` and ``YEAR`` to the unit of time for rate limit.
+- area: stats
```
ggreenway, line 35: Please move to `behavior_changes` section, and document the old and new names, similar to https://github.com/envoyproxy/envoy/blob/3752119b94ae4a5ff337364f585f54305406dc7f/changelogs/1.24.0.yaml#L8

source/common/config/well_known_names.h
```diff
@@ -137,6 +137,8 @@ class TagNameValues {
   const std::string RDS_ROUTE_CONFIG = "envoy.rds_route_config";
   // Request route given by the Router http filter
   const std::string ROUTE = "envoy.route";
+  // Stats prefix for the ext_authz HTTP filter
+  const std::string EXT_AUTHZ_PREFIX = "envoy.auth_prefix";
```
ggreenway, line 141:
```suggestion
  const std::string EXT_AUTHZ_PREFIX = "envoy.ext_authz_prefix";
```

source/common/config/well_known_names.cc
```diff
@@ -132,6 +132,12 @@ TagNameValues::TagNameValues() {
   // internal state of the regex which performs better.
   addRe2(HTTP_CONN_MANAGER_PREFIX, R"(^listener\..*?\.http\.((<TAG_VALUE>)\.))", ".http.");
 
+  // Extract ext_authz stat_prefix field
+  // cluster.[<cluster>.]ext_authz.[<ext_authz_prefix>.]*
+  addRe2(EXT_AUTHZ_PREFIX, R"(^cluster\.(?:[^\.]+\.)?ext_authz\.((<TAG_VALUE>)\.))");
```
ggreenway, line 137: Should these be `addTokenized` instead? @jmarantz ? Similar to `addTokenized(GRPC_BRIDGE_METHOD, "cluster.*.grpc.*.$.**");` earlier in the list.

ggreenway (changes requested):
Delete https://github.com/envoyproxy/envoy/blob/3752119b94ae4a5ff337364f585f54305406dc7f/test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc#L33. It should not be needed with this change.

/wait

source/common/config/well_known_names.cc
jmarantz: Yes that sounds good to me; that will be the fastest option.

jmarantz (commented):
lgtm modulo Greg's comments.

jmarantz (approved):

turettn:
@jmarantz I just fixed the merge conflict in the changelog, and that dismissed your review - sorry about that.

I think it's ready to go 🤞 - please let me know if there's anything else I can do.

jmarantz:
CI still has issues.

/wait

turettn:
So, that was interesting.  Removing `skip_tag_extraction_rule_check_ = true` added some checks that started throwing (in my opinion) false negatives.  The GoogleGrpc test integration uses the cluster name as the cluster stat_prefix, and then the tests automatically check to ensure that no stat_prefixes end up in the resulting tags.

Effectively, this means that your gRPC cluster name can not be a string that would appear in the final tag name - in this case, `cluster_name = ext_authz` is not compatible with the pattern `*.ext_authz.<ext_authz_prefix>.ok`.

I changed the gRPC cluster name to something else, since it seemed like the lowest-impact, most local change. 
 `//test/common/stats:all` and `//test/extensions/filters/http/ext_authz:all` pass locally, so I threw it at CI again.

ggreenway:
> So, that was interesting. Removing `skip_tag_extraction_rule_check_ = true` added some checks that started throwing (in my opinion) false negatives. The GoogleGrpc test integration uses the cluster name as the cluster stat_prefix, and then the tests automatically check to ensure that no stat_prefixes end up in the resulting tags.
> 
> Effectively, this means that your gRPC cluster name can not be a string that would appear in the final tag name - in this case, `cluster_name = ext_authz` is not compatible with the pattern `*.ext_authz.<ext_authz_prefix>.ok`.
> 
> I changed the gRPC cluster name to something else, since it seemed like the lowest-impact, most local change. `//test/common/stats:all` and `//test/extensions/filters/http/ext_authz:all` pass locally, so I threw it at CI again.

That seems like the best choice. I did the same in some other tests for the same reason.

ggreenway (approved):
LGTM

ggreenway:
@turettn something went wonky with CI. Can you merge main into this PR, mostly to re-trigger CI? I can't make that failed job run.

@jmarantz can you take a final look at this?

/wait-any

jmarantz (approved):

turettn:
@ggreenway @jmarantz 

GitHub seems to be expecting 5 CI checks (`android_build`, `java_helloworld`, `kotlin_baseline_app`, `kotlin_experimental_app`, and `kotlin_helloworld`) that never even start.  I don't think this is related to any of my changes...

If you'd like to do anything on my end (e.g. merge from main) please let me know.

jmarantz:
Yes let's get this in. How? I'm not sure! But can you try merging main and we'll see if that helps?

turettn:
Merging from main appears to have gotten everything green - thanks for your patience getting this PR to a mergeable state!

jmarantz:
Happy to finally push the merge button & thanks for your patience!

PR #24003, opened 2022-11-15
[balsa] Run HttpProtocolIntegrationTests with BalsaParser as well.

bencebeky:
Set up HttpProtocolIntegrationTest and its derived test classes, as well as Http2BufferWatermarksTest (a class derived from HttpIntegrationTest) to run with BalsaParser in addition to http-parser if either the upstream or the downstream protocol is HTTP/1.1.  Disable BalsaParser for now for a few tests that are failing.

Also bump shard counts for the following two tests:

//test/integration:overload_integration_test
//test/extensions/filters/http/health_check:health_check_integration_test

This is because the increased number of test cases resulted in timeouts (against the five minute limit) of TSAN runs.

Tracking issue: https://github.com/envoyproxy/envoy/issues/21245

Signed-off-by: Bence Béky <bnc@google.com>

Commit Message: [balsa] Run HttpProtocolIntegrationTests with BalsaParser as well.
Additional Description:
Risk Level: low, test-only change
Testing: //test/integration:all
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: n/a

PR #24058, opened 2022-11-17
[balsa] Fix 304WithBody test and add new one.

bencebeky:
Add //test/common/http/http1:codec_impl_test test case for 304 response
with body for both BalsaParser and http-parser.  In this test, the body
is dispatched to the parser in the same Parser::execute() call as the
headers.

Modify ProtocolIntegrationTest.304WithBody by appending '\r\n' to the
end of the "body" of the 304 response, necessary for BalsaParser to
signal an error.

Tracking issue: #21245

Also include #24003 (GitHub does not support dependent PRs)

Commit Message: [balsa] Fix 304WithBody test and add new one.
Additional Description:
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/http1/balsa_parser.cc
+++ b/source/common/http/http1/balsa_parser.cc
@@ -226,8 +226,8 @@ void BalsaParser::OnRequestFirstLineInput(absl::string_view /*line_input*/,
 
 void BalsaParser::OnResponseFirstLineInput(absl::string_view /*line_input*/,
                                            absl::string_view /*version_input*/,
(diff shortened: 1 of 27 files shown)


bencebeky:
Now that https://github.com/envoyproxy/envoy/pull/24003 landed, abandoning in favor of a new PR with clean history.  The only purpose of this temporary PR was to see if presubmits pass, and if not, be able to start working on that while waiting for https://github.com/envoyproxy/envoy/pull/24003.

PR #24060, opened 2022-11-17
sni proxy: disable read while waiting for DNS

kyessenov:

Commit Message: Explicitly disable read during async DNS call. 
Additional Description: tcp_proxy implicitly disables read before this filter during `initialize`, but we cannot always assume it is going to do that. Context: https://github.com/envoyproxy/envoy/pull/23944
Risk Level: low
Testing: regression


soulxu (commented):
LGTM, thanks

KBaichoo (approved):

Issue #23985, opened 2022-11-14
CI: Coverage is currently broken


KBaichoo:
**If you are reporting *any* crash or *any* potential security issue, *do not*
open an issue in this repo. Please report the issue via emailing
envoy-security@googlegroups.com where the issue will be triaged appropriately.**

CI: Coverage pages 404

*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

404 when going to GCS bucket for coverage report using the link given by CI.

*Repro steps*:
> Go to  recent coverage CI output to see the uploaded coverage report. For example: https://github.com/envoyproxy/envoy/pull/23935, we'd get linked to https://storage.googleapis.com/envoy-pr/87dbffb/coverage/index.html which encounters the error below:

```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Details>No such object: envoy-pr/87dbffb/coverage/index.html</Details>
</Error>
```

phlax:
i have followed this up quite far already - the problem is essentially bazel - the result of the `bazel coverage ...` command/s is returning empty files - not sure why that is but that is how far i got (#23930)

KBaichoo:
Yea looking at the gcs bucket I see:
```
[08:01:51] kbaichoo:87dbffb:0 $ls -al coverage/
total 8
drwxr-x--- 2 kbaichoo primarygroup 4096 Nov 14 20:00 .
drwxr-x--- 5 kbaichoo primarygroup 4096 Nov 14 20:01 ..
-rw-r----- 1 kbaichoo primarygroup    0 Nov 14 20:00 coverage.dat
```

phlax:
at that point we should loads of html files

if you track back the bazel command is outputting the empty dat file

phlax:
i suspect the issue is with `bazel/coverage/collect_cc_coverage.sh` not being updated in line with bazel changes or somesuch

phlax:
tracking trough ci history seems to confirm https://github.com/envoyproxy/envoy/pull/23678 as the culprit

cc @keith

phlax:
and reverting that commit reveals that its not just the upload broken but coverage itself.

PR #24062, opened 2022-11-18
Reduce VHDS overhead in RDS update

yurykats:
If there is no VHDS, do not allocate a map and do not clear/reset RDS vhosts during RDS update.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/router/route_config_update_receiver_impl.cc
+++ b/source/common/router/route_config_update_receiver_impl.cc
@@ -20,9 +20,10 @@ namespace {
 
 // Resets 'route_config::virtual_hosts' by merging VirtualHost contained in
 // 'rds_vhosts' and 'vhds_vhosts'.
-void rebuildRouteConfigVirtualHosts(const VirtualHostMap& rds_vhosts,
-                                    const VirtualHostMap& vhds_vhosts,
-                                    envoy::config::route::v3::RouteConfiguration& route_config) {
+void rebuildRouteConfigVirtualHosts(
+    const RouteConfigUpdateReceiverImpl::VirtualHostMap& rds_vhosts,
+    const RouteConfigUpdateReceiverImpl::VirtualHostMap& vhds_vhosts,
+    envoy::config::route::v3::RouteConfiguration& route_config) {
   route_config.clear_virtual_hosts();
   for (const auto& vhost : rds_vhosts) {
     route_config.mutable_virtual_hosts()->Add()->CopyFrom(vhost.second);
--- a/source/common/router/route_config_update_receiver_impl.cc
+++ b/source/common/router/route_config_update_receiver_impl.cc
@@ -81,15 +81,11 @@ bool RouteConfigUpdateReceiverImpl::onVhdsUpdate(
     const VirtualHostRefVector& added_vhosts, const std::set<std::string>& added_resource_ids,
     const Protobuf::RepeatedPtrField<std::string>& removed_resources,
     const std::string& version_info) {
-  std::unique_ptr<std::map<std::string, envoy::config::route::v3::VirtualHost>>
-      vhosts_after_this_update;
+  std::unique_ptr<VirtualHostMap> vhosts_after_this_update;
   if (vhds_virtual_hosts_ != nullptr) {
-    vhosts_after_this_update =
-        std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>(
-            *vhds_virtual_hosts_);
+    vhosts_after_this_update = std::make_unique<VirtualHostMap>(*vhds_virtual_hosts_);
   } else {
-    vhosts_after_this_update =
-        std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>();
+    vhosts_after_this_update = std::make_unique<VirtualHostMap>();
   }
   const bool removed = removeVhosts(*vhosts_after_this_update, removed_resources);
   const bool updated = updateVhosts(*vhosts_after_this_update, added_vhosts);
--- a/source/common/router/route_config_update_receiver_impl.cc
+++ b/source/common/router/route_config_update_receiver_impl.cc
@@ -59,17 +59,25 @@ bool RouteConfigUpdateReceiverImpl::onRdsUpdate(const Protobuf::Message& rc,
   new_route_config->CopyFrom(rc);
   const uint64_t new_vhds_config_hash =
       new_route_config->has_vhds() ? MessageUtil::hash(new_route_config->vhds()) : 0ul;
-  VirtualHostMap rds_virtual_hosts;
-  for (const auto& vhost : new_route_config->virtual_hosts()) {
-    rds_virtual_hosts.emplace(vhost.name(), vhost);
-  }
-  if (vhds_virtual_hosts_ != nullptr && !vhds_virtual_hosts_->empty()) {
-    // If there are vhosts supplied by VHDS, merge them with RDS vhosts.
-    rebuildRouteConfigVirtualHosts(rds_virtual_hosts, *vhds_virtual_hosts_, *new_route_config);
+  if (new_route_config->has_vhds()) {
+    // When using VHDS, stash away RDS vhosts, so that they can be merged with VHDS vhosts in
+    // onVhdsUpdate.
+    if (rds_virtual_hosts_ == nullptr) {
+      rds_virtual_hosts_ =
+          std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>();
+    } else {
+      rds_virtual_hosts_->clear();
+    }
+    for (const auto& vhost : new_route_config->virtual_hosts()) {
+      rds_virtual_hosts_->emplace(vhost.name(), vhost);
+    }
+    if (vhds_virtual_hosts_ != nullptr && !vhds_virtual_hosts_->empty()) {
+      // If there are vhosts supplied by VHDS, merge them with RDS vhosts.
+      rebuildRouteConfigVirtualHosts(*rds_virtual_hosts_, *vhds_virtual_hosts_, *new_route_config);
+    }
   }
   base_.updateConfig(std::move(new_route_config));
   base_.updateHash(new_hash);
-  rds_virtual_hosts_ = std::move(rds_virtual_hosts);
   vhds_configuration_changed_ = new_vhds_config_hash != last_vhds_config_hash_;
   last_vhds_config_hash_ = new_vhds_config_hash;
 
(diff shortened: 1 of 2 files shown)


source/common/router/route_config_update_receiver_impl.cc
```diff
@@ -78,10 +81,11 @@ bool RouteConfigUpdateReceiverImpl::onVhdsUpdate(
     const VirtualHostRefVector& added_vhosts, const std::set<std::string>& added_resource_ids,
     const Protobuf::RepeatedPtrField<std::string>& removed_resources,
     const std::string& version_info) {
-
   auto vhosts_after_this_update =
-      std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>(
-          *vhds_virtual_hosts_);
+      vhds_virtual_hosts_ == nullptr
```
jmarantz, line 85: WDYT of making this a normal conditional rather than a ternary.

It probably doesn't make a difference in terms of what it compiles to, but doing that would have 2 (minor advantages):  slightly easier to read (IMO), and it would be easier to verify that you hit both branches in tests by looking at coverage.

jmarantz (commented):
Looks fine but I guess someone that knows this code should look also. 

But can you ensure there's a unit test that covers both cases?

yurykats:
There is existing test coverage in vhds_test and rds_impl_test.

source/common/router/route_config_update_receiver_impl.cc
yurykats: Done

jmarantz (approved):
thanks!

jmarantz:
Oh sorry you need to fix DCO.

source/common/router/route_config_update_receiver_impl.cc
```diff
@@ -63,7 +63,10 @@ bool RouteConfigUpdateReceiverImpl::onRdsUpdate(const Protobuf::Message& rc,
   for (const auto& vhost : new_route_config->virtual_hosts()) {
     rds_virtual_hosts.emplace(vhost.name(), vhost);
   }
-  rebuildRouteConfigVirtualHosts(rds_virtual_hosts, *vhds_virtual_hosts_, *new_route_config);
+  if (vhds_virtual_hosts_ != nullptr && !vhds_virtual_hosts_->empty()) {
+    // If there are vhosts supplied by VHDS, merge them with RDS vhosts.
+    rebuildRouteConfigVirtualHosts(rds_virtual_hosts, *vhds_virtual_hosts_, *new_route_config);
```
adisuissa, line 68: Clarifying question/suggestion: if I read [rebuildRouteConfigVirtualHosts](https://github.com/envoyproxy/envoy/blob/456e4497a2fc3d50446d52e61cd5d496f4e1da3f/source/common/router/route_config_update_receiver_impl.cc#L23) correctly, it clears the virtual_hosts (vhosts) from the given `new_route_config`, then adds the vhosts from `rds_virtual_hosts`, which were already in `new_route_config`, and then adds the hosts from `vhds_virtual_hosts_` (both lists are added into `new_route_config.virtual_hosts`).
If the if-condition holds, it means that to get to the same result, only the vhosts in `vhds_virtual_hosts_` need to be added, no?

If that's true, then instead of clearing the vhosts in `new_route_config` + re-adding the `rds_virtual_hosts` vhosts + adding the `vhds_virtual_hosts_` vhosts, this can be replaced by just adding the `vhds_virtual_hosts_` vhosts.

adisuissa (commented):
Change LGTM. Left a minor question/suggestion.

adisuissa:
Please [fix DCO](https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#fixing-dco).
Please check CI failures, as they seem related to the code-changes.

source/common/router/route_config_update_receiver_impl.cc
yurykats: rebuildRouteConfigVirtualHosts is called from 2 places -- onRdsUpdate and onVhdsUpdate -- to combine vhost lists coming from both places. The original author(s) went for a simple route of clearing all vhosts and then putting rds and vhds hosts back in, instead of figuring out what exactly needs to be added/updated/removed. I don't think it's worth re-implementing the whole vhost merging logic. The point of this PR is to optimize the common path when vhosts are coming from RDS only, and there is no VHDS.

source/common/router/route_config_update_receiver_impl.cc
```diff
@@ -63,7 +63,10 @@ bool RouteConfigUpdateReceiverImpl::onRdsUpdate(const Protobuf::Message& rc,
   for (const auto& vhost : new_route_config->virtual_hosts()) {
     rds_virtual_hosts.emplace(vhost.name(), vhost);
```
stevenzzzz, line 64: let's avoid this copy by moving hosts out of new_route_config, the rebuild is just simply append existing vhds_virtual_hosts to existing rds_v-hosts, which could be also be done 0-copy through move.

yurykats: I don't think to mess with the VHDS handling here. Both new_route_config and rds_virtual_hosts are used to rebuild vhosts in onVhdsUpdate. new_route_config is also cached for other potential uses some place else.

source/common/router/route_config_update_receiver_impl.h
```diff
@@ -42,9 +42,7 @@ class RouteConfigUpdateReceiverImpl : public RouteConfigUpdateReceiver {
       : config_traits_(optional_http_filters,
                        factory_context.messageValidationContext().dynamicValidationVisitor()),
         base_(config_traits_, proto_traits, factory_context), last_vhds_config_hash_(0ul),
-        vhds_virtual_hosts_(
-            std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>()),
-        vhds_configuration_changed_(true) {}
+        vhds_virtual_hosts_(nullptr), vhds_configuration_changed_(true) {}
```
jmarantz, line 45: you can omit nullptr inits for unique_ptr.

source/common/router/route_config_update_receiver_impl.cc
```diff
@@ -78,10 +81,16 @@ bool RouteConfigUpdateReceiverImpl::onVhdsUpdate(
     const VirtualHostRefVector& added_vhosts, const std::set<std::string>& added_resource_ids,
     const Protobuf::RepeatedPtrField<std::string>& removed_resources,
     const std::string& version_info) {
-
-  auto vhosts_after_this_update =
-      std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>(
-          *vhds_virtual_hosts_);
+  std::unique_ptr<std::map<std::string, envoy::config::route::v3::VirtualHost>>
```
jmarantz, line 84: WDYT about making a nickname for this complicated map type?

source/common/router/route_config_update_receiver_impl.h
yurykats: Done

source/common/router/route_config_update_receiver_impl.cc
yurykats: Done

source/common/router/route_config_update_receiver_impl.cc
adisuissa: This is just a perf suggestion, as the O(N) re-additions of RDS routes isn't needed, but as you point, this is not the main point of the PR.

adisuissa (approved):
LGTM, thanks!
Left a small perf suggestion, but up to you.

source/common/router/route_config_update_receiver_impl.h
```diff
@@ -34,6 +34,8 @@ class ConfigTraitsImpl : public Rds::ConfigTraits {
   ProtobufMessage::ValidationVisitor& validator_;
 };
 
+using VirtualHostMap = std::map<std::string, envoy::config::route::v3::VirtualHost>;
```
jmarantz, line 37: this is fine; or maybe you want to sink this inside RouteConfigUpdateReceiverImpl ?

source/common/router/route_config_update_receiver_impl.cc
```diff
@@ -78,10 +81,12 @@ bool RouteConfigUpdateReceiverImpl::onVhdsUpdate(
...
-
-  auto vhosts_after_this_update =
-      std::make_unique<std::map<std::string, envoy::config::route::v3::VirtualHost>>(
-          *vhds_virtual_hosts_);
+  std::unique_ptr<VirtualHostMap> vhosts_after_this_update;
+  if (vhds_virtual_hosts_ != nullptr) {
+    vhosts_after_this_update = std::make_unique<VirtualHostMap>(*vhds_virtual_hosts_);
+  } else {
+    vhosts_after_this_update = std::make_unique<VirtualHostMap>();
+  }
```
jmarantz, line 89: naive question: could you skip making the empty map the calls to removeHosts and updateVhosts if vhds_virtual_hosts is null?

just set 'removed' and 'updated' to 0?

I admit I didn't read deeply here so maybe those functions have more side effects.

yurykats: I don't really want to get into changing onVhdsUpdate here, just trying to improve onRdsUpdate :)

source/common/router/route_config_update_receiver_impl.h
yurykats: I'm new to Envoy code base, what's the common practice?

source/common/router/route_config_update_receiver_impl.cc
yurykats: Yes, I don't want to get into VHDS code paths here, just improving RDS update when there is no VHDS.

source/common/router/route_config_update_receiver_impl.h
jmarantz: I've seen both: it all depends on whether that nickname is useful outside that class. At first glance I didn't see any need to have it exposed outside the class.

source/common/router/route_config_update_receiver_impl.cc
stevenzzzz: I think the two copies are just for "merging" rds-pushed vhost and VHDS-pushed vhosts, when either RDS or VHDS are updated. 

Instead of making two copies, it's totally fine to have a range-index indicates how many of the vhosts in the stored RouteConfiguration::vhosts are from RDS/VHDS. 
All the copy around could be move based.

source/common/router/route_config_update_receiver_impl.h
yurykats: Moved

source/common/router/route_config_update_receiver_impl.cc
yurykats: We'll look into avoiding vhost copy in a follow up.

source/common/router/route_config_update_receiver_impl.cc
yurykats: Changed to not keep a copy of RDS vhosts at all, if there is no VHDS. This is the largest memory hog.

htuch:
LGTM. I wanted to verify that the coverage looked good on the new branches, but I got https://storage.googleapis.com/envoy-pr/b5ceb11/coverage/index.html - @phlax do you know what is up here?

phlax:
> @phlax do you know what is up here?

- https://github.com/envoyproxy/envoy/issues/23985
- https://github.com/envoyproxy/envoy/pull/24093

jmarantz (approved):
These changes look fine to me but I feel like someone more familiar with the router system better review. I'll find one.

zuercher (approved):
This looks fine to me. I think my only comment is that `rebuildRouteConfigVirtualHosts` might be better named `mergeRouteConfigVirtualHosts`, but that was probably always true.

Issue #23652, opened 2022-10-25
ext_proc ImmediateResponse headers not respected in ResponseBody response using BUFFERED_PARTIAL with chunked response from server


tcohen01:
### Version
v1.24-latest / v1.24.0

### Description
We wish to send ImmediateResponse in our ext_proc service based on conditions that need to be met on response body content. The response body may come in different formats depending on the webserver, possibly chunked transfer-encoding as well.

However, when using `transfer-encoding: chunked` in the server's response and there's more than one or two chunks, ImmediateResponse's headers are not respected and instead envoy attempts to send ImmediateResponse's body with the original chunked response, this leads to transfer encoding issues as the ImmediateResponse body is not sent in chunked format, leading to errors:
<details>
<summary>CURL Error on chunked ResponseBody ImmediateResponse</summary>

```
curl "localhost:10000/chunked?size=1024&chunksize=512" -v
*   Trying 127.0.0.1:10000...
* Connected to localhost (127.0.0.1) port 10000 (#0)
> GET /chunked?size=1024&chunkedsize=512 HTTP/1.1
> Host: localhost:10000
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< x-content-type-options: nosniff
< date: Tue, 25 Oct 2022 11:56:59 GMT
< x-envoy-upstream-service-time: 0
< server: envoy
< transfer-encoding: chunked
< 
* transfer closed with outstanding read data remaining
* Closing connection 0
curl: (18) transfer closed with outstanding read data remaining
```
</details>
<details>
<summary>Expected CURL Results (Same request, ResponseBody processing mode set to BUFFERED</summary>

```
curl "localhost:10000/chunked?size=1024&chunksize=512" -v
*   Trying 127.0.0.1:10000...
* Connected to localhost (127.0.0.1) port 10000 (#0)
> GET /chunked?size=2048&chunksize=64 HTTP/1.1
> Host: localhost:10000
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 23
< content-type: text/plain
< date: Tue, 25 Oct 2022 12:54:30 GMT
< server: envoy
< 
* Connection #0 to host localhost left intact
Immediate Response Body
```
</details>

Note:
* This bug does not happen in BUFFERED processing mode.
[...]

PR #24064, opened 2022-11-18
Fix PARTIAL_BUFFERED mode to wait until body is delivered before resuming headers

gbrail:
Commit Message: Modify PARTIAL_BUFFERED code path so that it does not resume the plugin flow until either the body has been completely received OR if the body limit has been exceeded.
Additional Description: This way, a processor using this mode can change headers or send an immediate response after examining the body, unless the body is larger than the limit. Also changed filter_test.cc as this a "white box" test that verifies specific responses.
Risk Level: Medium
Testing: Added new integration test and checked using test case supplied by opener of the bug.
Docs Changes:
Release Notes:
Platform Specific Features:
Fixes [#Issue](https://github.com/envoyproxy/envoy/issues/23652)


source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -348,38 +351,19 @@ FilterDataStatus Filter::onData(ProcessorState& state, Buffer::Instance& data, b
...
-      std::tie(terminate, result) = sendStreamChunk(state, data, end_stream);
-
-      if (terminate) {
-        return result;
-      }
     } else {
-      // Keep on running and buffering
       state.enqueueStreamingChunk(data, false, false);
-
-      if (Runtime::runtimeFeatureEnabled(Runtime::defer_processing_backedup_streams) &&
```
snowp, line 362: Do we no longer care about this runtime flag?

source/extensions/filters/http/ext_proc/ext_proc.cc
```diff
@@ -339,6 +339,9 @@ FilterDataStatus Filter::onData(ProcessorState& state, Buffer::Instance& data, b
     if (state.partialBodyProcessed()) {
       // We already sent and received the buffer, so everything else just falls through.
       ENVOY_LOG(trace, "Partial buffer limit reached");
+      // Make sure that we do not accidentally try to modify the headers before
+      // we continue, which will result in them possibly being sent.
+      state.setHeaders(nullptr);
```
snowp, line 344: For my understanding - do we end up restoring the headers in the processingState somewhere so that we are able to process the headers once we are ready to send them?

source/extensions/filters/http/ext_proc/processor_state.cc
```diff
@@ -240,7 +246,25 @@ absl::Status ProcessorState::handleBodyResponse(const BodyResponse& response) {
...
+            return mut_status;
+          }
+        } else {
+          ENVOY_LOG(debug, "Response had header mutations but headers aren't available");
+        }
+      }
       if (common_response.has_body_mutation()) {
+        if (headers_ != nullptr) {
+          // Always reset the content length here to prevent later problems.
+          headers_->removeContentLength();
```
snowp, line 266: Is this code block the same as the one above that handles evaluating the header mutation? If so, is there a way to resuse the code to make it obivous that we're doing the same thing?

snowp (changes requested):
High level this makes sense, a few questions

source/extensions/filters/http/ext_proc/ext_proc.cc
gbrail: In this case, it's more that the behavior that someone previously found was correct when this flag is enabled seems to be the correct behavior in all cases. I'm not actually sure who put that feature flag check in before, but when I looked at the code carefully to diagnose this issue I figured out that we didn't need the test because that was actually what we wanted to do all the time anyway.

source/extensions/filters/http/ext_proc/ext_proc.cc
gbrail: What actually happened here is that the filter, in [en|de]codeHeaders, stashed away a pointer to the headers if it was going to be in buffered or partial buffered mode so that we can modify them in response to the message body. This only works because in these two cases we have paused processing and we know that envoy will not deliver the headers to the client yet, and because Envoy doesn't move or modify the target of that pointer.

I don't really LIKE doing it this way at all, and I feel like I'm relying too much on the current behavior of the proxy. I would much prefer if Envoy would let me use something reference-counted (like a shared_ptr) to refer to the headers object, but I believe that would require a change to the whole filter mechanism.

source/extensions/filters/http/ext_proc/processor_state.cc
gbrail: Well, there are three places where we call applyHeaderMutations and two where we call applyBodyMutations, but each time the context or the parameters are different -- I don't see a further opportunity to "DRY up" that code from where it already is. If it helps we can add more comments, or adjust the log messages to make them more consistent, if you think that helps. Otherwise, if I'm missing the point, I'm happy to look for more opportunities.

gbrail: I'm going to resolve this and the other comments, but please re-open if you have more comments!

stevenzzzz:
hi Greg, I want to understand better why not we allow ext_proc to buffer while headers processing is not returned. 
I think if we return StopAllIterationAndWatermark in {de/en}codeHeaders, it would be a much simpler state machine for ext_proc to maintain.

gbrail:
@stevenzzzz I'm not sure -- are you suggesting that in all cases we return StopAllIterationandWatermark instead of StopIteration? That's worth thinking about, as there are indeed a lot of use cases here. I think that we still need to deal with all the cases of what happens when the whole body comes in versus part of the body, etc. But I'd love to see if we can simplify all things so we can certainly experiment. Do you have a specific example where you think this might help?

gbrail:
@snowp I resolved your comments a while back, and I wasn't sure what, if any, changes were necessary -- could you please take a look?

stevenzzzz:
> @stevenzzzz I'm not sure -- are you suggesting that in all cases we return StopAllIterationandWatermark instead of StopIteration? That's worth thinking about, as there are indeed a lot of use cases here. I think that we still need to deal with all the cases of what happens when the whole body comes in versus part of the body, etc. But I'd love to see if we can simplify all things so we can certainly experiment. Do you have a specific example where you think this might help?

Yes, my thinking is that if since we don't what to do with the body yet until the HandleHeaderResponse is called, we might as well just ask other filters not to send body until then. I think that way it will simplify the state machine quite a bit. 

If we do it that way, then we only need to deal with data in onData IIUC.

stevenzzzz:
hi Greg, do you happen to be on Slack? 

I think we can talk in person for the idea that "onHeaders always return StopAllIterationandWatermark", I think that will simplifies things quite a lot, and is probably the right thing to do. 

We have a bug where when the:ext_proc filter is still waiting for OnResponseHeaders to be called,  upstream filter passes down encodeData(data, end_stream=true), when the data is bigger than the per-stream buffer size, it ends up with a 500 as the buffer size hits the high watermark. 

It's a much simpler state machine if we can streamline decodeHeaders and decodeData() for multiple filters by returning the StopAllIterationandWatermark, which stops upstream (or filters upstream) from calling encodeData.

gbrail:
I understand the question about using StopAllIteration, and I remembered why I don't.

In buffered and partial buffered mode, we want to have the ability for a processor to modify the headers after receiving the body -- imagine wanting to change content-length or content-type, for example, after transforming the contents. That means that we need to hold both header and body iteration in this filter until we know what the processing mode will be so that can happen -- once we let the headers flow to the next filter in the chain, we can't go back and change them.

Returning StopAllIterationAndWatermark from the headers callback would accomplish this and would indeed support many use cases. However, in the case above, we can't proceed with the filter chain until we get the body and let the processor have a crack at that. The only way to resume iteration and allow the body to flow to this filter is to call "continue(Encoding|Decoding)", which will allow the data to flow to this filter, but it will ALSO let the headers now flow to the next filter in the chain, which is not what we want.

What we need in order to do what you suggest is to add a "continue(En|De)CodingForThisFilterOnly" method. Then we could simplify some of this logic. I don't know how complex it is to add such a thing to the filter chain, however, so I haven't done it!

stevenzzzz:
> I understand the question about using StopAllIteration, and I remembered why I don't.
> 
> In buffered and partial buffered mode, we want to have the ability for a processor to modify the headers after receiving the body -- imagine wanting to change content-length or content-type, for example, after transforming the contents. That means that we need to hold both header and body iteration in this filter until we know what the processing mode will be so that can happen -- once we let the headers flow to the next filter in the chain, we can't go back and change them.
> 
> Returning StopAllIterationAndWatermark from the headers callback would accomplish this and would indeed support many use cases. However, in the case above, we can't proceed with the filter chain until we get the body and let the processor have a crack at that. The only way to resume iteration and allow the body to flow to this filter is to call "continue(Encoding|Decoding)", which will allow the data to flow to this filter, but it will ALSO let the headers now flow to the next filter in the chain, which is not what we want.
> 
> What we need in order to do what you suggest is to add a "continue(En|De)CodingForThisFilterOnly" method. Then we could simplify some of this logic. I don't know how complex it is to add such a thing to the filter chain, however, so I haven't done it!

I am not sure where the requirement "the ability for a processor to modify the headers after receiving the body" is from. I think 99.9% of the time we do body manipulation based on header-processing result instead. 

Actually per this section from Filter::onData, the data is only sent to ext-proc after header response is received anyway? 

```
  if (state.callbackState() == ProcessorState::CallbackState::HeadersCallback) {
    ENVOY_LOG(trace, "Header processing still in progress -- holding body data");
    // We don't know what to do with the body until the response comes back.
    // We must buffer it in case we need it when that happens.
    if (end_stream) {
      state.setPaused(true);
      return FilterDataStatus::StopIterationAndBuffer;
    } else {
      // Raise a watermark to prevent a buffer overflow until the response comes back.
      state.setPaused(true);
      state.requestWatermark();
      return FilterDataStatus::StopIterationAndWatermark;
    }
  }
```

KBaichoo:
I've seen use cases (not exactly of this filter) where json body is used to modify the headers.

> I am not sure where the requirement "the ability for a processor to modify
> the headers after receiving the body" is from. I think 99.9% of the time we
> do body manipulation based on header-processing result instead.

From a quick look, I think if we used `StopAllIterationAndWatermark` we wouldn't get invoked on `decodeData` unless `continueDecoding` is invoked which as @gbrail points out will continue iteration of headers beyond this filter. 

Couldn't you just hold the data using StopIteration for this? e.g. if the headers, data are still held by your ext_proc filter we wouldn't have to worry about encode* of course with the caveat of buffering/streaming. AFIAK this is what the filter does now.

> It's a much simpler state machine if we can streamline decodeHeaders and decodeData() for multiple filters by returning the StopAllIterationandWatermark, which stops upstream (or filters upstream) from calling encodeData.


The solution seems like you'd want the filter to apply flow control to the upstream ?

> We have a bug where when the:ext_proc filter is still waiting for OnResponseHeaders to be called, upstream filter passes down encodeData(data, end_stream=true), when the data is bigger than the per-stream buffer size, it ends up with a 500 as the buffer size hits the high watermark.

stevenzzzz:
> I've seen use cases (not exactly of this filter) where json body is used to modify the headers.
> 
> > I am not sure where the requirement "the ability for a processor to modify
> > the headers after receiving the body" is from. I think 99.9% of the time we
> > do body manipulation based on header-processing result instead.
> 
> From a quick look, I think if we used `StopAllIterationAndWatermark` we wouldn't get invoked on `decodeData` unless `continueDecoding` is invoked which as @gbrail points out will continue iteration of headers beyond this filter.

What's the point of calling decodeData if headers response is gonna terminate the stream anyway?

> Couldn't you just hold the data using StopIteration for this? e.g. if the headers, data are still held by your ext_proc filter we wouldn't have to worry about encode* of course with the caveat of buffering/streaming. AFIAK this is what the filter does now.
See the problem you answered below. 
> 
> > It's a much simpler state machine if we can streamline decodeHeaders and decodeData() for multiple filters by returning the StopAllIterationandWatermark, which stops upstream (or filters upstream) from calling encodeData.
> 
> The solution seems like you'd want the filter to apply flow control to the upstream ?
> 
> > We have a bug where when the:ext_proc filter is still waiting for OnResponseHeaders to be called, upstream filter passes down encodeData(data, end_stream=true), when the data is bigger than the per-stream buffer size, it ends up with a 500 as the buffer size hits the high watermark.
by upstream I assume you mean the "filter up the chain pushing data to this filter", yeah, that would probably solve the problem I am dealing with. 

I am just curious what's the right model for flow control, do we need to defend a "inconsiderate filter pushing more than welcomed data"? 
I am also not sure the complexity in ext_proc filter justify the rare use case of "dealing with headers after seeing data", which is basically saying "buffering the whole response".  The filter has been made this way already tho.

stevenzzzz:
I think the current impl actually wants to "pause data plumbing until header processing response is back". 

@gbrail do you mind I open an issue tracking the "always raise watermark from decodeHeaders" ask?

gbrail:
Not today, GitHub!

I still want to merge this -- let me re-merge with main and see what else you all need...

snowp (approved):
Thanks!

PR #24070, opened 2022-11-18
rate_limit: Add time unit month and year

tyxia:
Per #23844 and #24031,  @renuka-fernando wants to have time unit `month` and `year` for his business need

Release note added.

### DIFF
--- a/source/extensions/filters/http/ratelimit/ratelimit_headers.cc
+++ b/source/extensions/filters/http/ratelimit/ratelimit_headers.cc
@@ -75,6 +75,10 @@ uint32_t XRateLimitHeaderUtils::convertRateLimitUnit(
     return 60 * 60;
   case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::DAY:
     return 24 * 60 * 60;
+  case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::MONTH:
+    return 30 * 24 * 60 * 60;
+  case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::YEAR:
+    return 365 * 24 * 60 * 60;
   case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::UNKNOWN:
   default:
     return 0;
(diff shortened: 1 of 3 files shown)


source/extensions/filters/http/ratelimit/ratelimit_headers.cc
```diff
@@ -75,6 +75,10 @@ uint32_t XRateLimitHeaderUtils::convertRateLimitUnit(
     return 60 * 60;
   case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::DAY:
     return 24 * 60 * 60;
+  case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::MONTH:
+    return 30 * 24 * 60 * 60;
+  case envoy::service::ratelimit::v3::RateLimitResponse::RateLimit::YEAR:
```
adisuissa, line 80: As the return type is uint32, I guess there's a (pretty large) limit to the number of years (and maybe months). Could you document that in the related proto fields comments?
Moreover, consider rejecting a config with a large months/years value.

(alternatively the return type can be updated, but not sure if it actually makes sense in this case as I don't think one will set a 150 years quota)

adisuissa (commented):
Thanks for working on this!
Please add release notes as this is a user-facing change.

source/extensions/filters/http/ratelimit/ratelimit_headers.cc
tyxia: It is a good point to consider
The config here is just to specify `unit` but not to specify the value (i.e., num# of that unit or total value). So there is no config specifying the large months/year value. Also, one year is already pretty rare case for rate limiting here as this indicates they are running the envoy instance without update the config.

For this function itself, I think current return type should be fine: max of uint32(4,294,967,295) still has some room for this year's value (31,536,000). For this case, It is more of the caller (rather than callee) site 's responsibility to make sure it doesn't overflow by using larger type.(e.g., when they manipulate the return value of this function )

tyxia (commented):
Thanks for the review! Release Note has been added.

PTAL

source/extensions/filters/http/ratelimit/ratelimit_headers.cc
adisuissa: Yes, I agree it is up to the caller of this function.
I had a mistake when reading this function earlier, thinking it actually does conversion of a given value in some unit to the value in seconds.

adisuissa (approved):
Thanks!
Code changes LGTM!
/lgtm api

PR #24081, opened 2022-11-18
api: document the fact that v3 is forever

mattklein123:
Signed-off-by: Matt Klein <mklein@lyft.com>


api/API_VERSIONING.md
```diff
@@ -82,57 +78,17 @@ implementations within a major version should set explicit values for these fiel
...
-plane to condition on client and major/minor API version support. Currently under discussion, but
-not finalized will be the sunsetting of Envoy client support for deprecated features after a year
-of support within a major version. Please post to https://github.com/envoyproxy/envoy/issues/8416
-any thoughts around this.
+At one point, the Envoy project planned for regular major version updates to the xDS API in order to
+remove technical debt. At this point we recognize that Envoy and the larger xDS ecosystem (gRPC,
+etc.) is too widely used to make version bumps realistic. As such, for practical purposes, the v3
+API is the final major version of the API and will be supported forever. Deprecations will still
+occur as an end-user indication that there is a preferred way to configure a particular feature, but
+no field will ever be removed nor will Envoy ever remove the implementation for any deprecated
```
soulxu, line 86: Should we mention the wrong API behavior could be fixed and the old behavior will be deprecated along with runtime flag?

api/API_VERSIONING.md
```diff
@@ -82,57 +78,17 @@ implementations within a major version should set explicit values for these fiel
...
+field.
+
+**NOTE**: Client implementations are free to output additional warnings about field usage beyond
```
htuch, line 89: I think also client versions are free to stop supporting fields if they want to.

api/API_VERSIONING.md
htuch: I think "wrong" is a bit too broad a term. We'll fix implementation behaviors that don't match the documented spec under runtime guard, is this what is meant?

htuch (commented):
Generally LGTM, strong agreement with the proposed change.

htuch:
Once this is done, we can probably turn down even more of the protoxform version related stuff @phlax

api/API_VERSIONING.md
soulxu: yea, agree with "wrong" isn't accurate. Not sure whether we doc about runtime and API somewhere.  I thought the rule could be:

* If the API doesn't work(It doesn't have any meaningful behavior), then it is just a bug, we fix it directly.
* If the API works with meaningful behavior, but the behavior is different from what the API doc said(or API doc doesn't have a clear description of the behavior), then we fix it with a runtime flag.

mattklein123: I think we cover the use of runtime guarding in Envoy elsewhere (cc @alyssawilk) but happy to eat verbiage here if you can suggest some? In some sense I think runtime guarding is orthogonal to this.

api/API_VERSIONING.md
mattklein123: Added an additional clause, PTAL,

api/API_VERSIONING.md
soulxu: I only found this doc https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#runtime-guarding, we can reference it. But it doesn't talk about runtime flag for API, maybe we can improve that later.

soulxu: add few words there https://github.com/envoyproxy/envoy/pull/24134 if we didn't have such doc somewhere else

htuch (approved):
LGTM, thanks! End of an era!

soulxu:
LGTM, about runtime guard, can be discussed in https://github.com/envoyproxy/envoy/pull/24134

phlax (approved):
lgtm, altho forever seems like a long time!

api/API_VERSIONING.md
```diff
@@ -7,26 +7,22 @@ stability.
...
 
 In everyday discussion and GitHub labels, we refer to the `v2`, `v3`, `vN`, `...` APIs. This has a
 specific technical meaning. Any given message in the Envoy API, e.g. the `Bootstrap` at
 `envoy.config.bootstrap.v3.Bootstrap`, will transitively reference a number of packages in the Envoy
 API. These may be at `vN`, `v(N-1)`, etc. The Envoy API is technically a DAG of versioned package
 namespaces. When we talk about the `vN xDS API`, we really refer to the `N` of the root
 configuration resources (e.g. bootstrap, xDS resources such as `Cluster`). The
-v3 API bootstrap configuration is `envoy.config.bootstrap.v3.Bootstrap`, even
-though it might transitively reference `envoy.service.trace.v2`.
+v3 API bootstrap configuration is `envoy.config.bootstrap.v3.Bootstrap`.
```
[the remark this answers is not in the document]
mattklein123: OK not exactly sure how to fix. I'm going to merge but feel free to raise a clarification PR.

PR #24084, opened 2022-11-19
Validate core retry policy at config parsing time

qiwzhang:

This issue is found by newly added fuzz test from jwt_authn filter.  The root course is 
```
Utility::convertCoreToRouteRetryPolicy 
```
may throw an exception [here](https://github.com/envoyproxy/envoy/blob/main/source/common/http/utility.cc#L1088)

This function is called by `jwt_authn` filter and `gcp_authn` filters in the worker thread.   

The changes:
* change the `throw EnvoyException` to `ENVOY_BUG`.  It should not throw exception if it may be called at worker threads.
* Add a new function `validate` to be called at main thread.
* 

Risk Level: None
Testing:  None
Docs Changes: None
Release Notes:  None

### DIFF
--- a/source/common/http/utility.cc
+++ b/source/common/http/utility.cc
@@ -1067,6 +1067,20 @@ Utility::AuthorityAttributes Utility::parseAuthority(absl::string_view host) {
   return {is_ip_address, host_to_resolve, port};
 }
 
+void Utility::validateCoreRetryPolicy(const envoy::config::core::v3::RetryPolicy& retry_policy) {
+  if (retry_policy.has_retry_back_off()) {
+    const auto& core_back_off = retry_policy.retry_back_off();
+
+    uint64_t base_interval_ms = PROTOBUF_GET_MS_REQUIRED(core_back_off, base_interval);
+    uint64_t max_interval_ms =
+        PROTOBUF_GET_MS_OR_DEFAULT(core_back_off, max_interval, base_interval_ms * 10);
+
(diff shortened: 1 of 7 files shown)


qiwzhang:
`android_tests / java_tests_mac ` test failed. Not sure how to debug it.

source/common/http/utility.cc
```diff
@@ -1067,6 +1067,20 @@ Utility::AuthorityAttributes Utility::parseAuthority(absl::string_view host) {
...
+    uint64_t base_interval_ms = PROTOBUF_GET_MS_REQUIRED(core_back_off, base_interval);
+    uint64_t max_interval_ms =
+        PROTOBUF_GET_MS_OR_DEFAULT(core_back_off, max_interval, base_interval_ms * 10);
+
+    if (max_interval_ms < base_interval_ms) {
+      throw EnvoyException("max_interval must be greater than or equal to the base_interval");
+    }
+  }
+}
+
```
wbpcode, line 1083: Should this be called for `GrpcAccessLogger` which also have a retry policy?

wbpcode: /way-any

qiwzhang: Done

wbpcode (approved):
LGTM. Thanks.

wbpcode:
Defer to cc @lizan for merge or final review.

lizan (approved):

PR #24094, opened 2022-11-20
upstream: verify all statuses of the override host

### LINKED ISSUE #23213: override_host_set conflates different EDS health statuses

wbpcode:
Commit Message: upstream: verify all statuses of the override host
Additional Description:

Fix #23213

Risk Level: Mid. core code update.
Testing: unit.
Docs Changes: n/a.
Release Notes: n/a.

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -47,6 +47,9 @@ bug_fixes:
 - area: http
   change: |
     fixed a bug where Utility::PercentEncoding::encode() encodes some characters incorrectly because it was treating the value as negative.
+- area: upstream
+  change: |
+    fixed a bug that only coarse health status is used for override host selection.
 
 removed_config_or_runtime:
 # *Normally occurs at the end of the* :ref:`deprecation period <deprecated>`
--- a/source/common/upstream/cluster_manager_impl.h
+++ b/source/common/upstream/cluster_manager_impl.h
@@ -565,12 +565,15 @@ private:
       // be shared across its hosts.
       Http::PersistentQuicInfoPtr quic_info_;
 
-      // Expected override host statues. Every bit in the OverrideHostStatus represent an enum value
-      // of Host::Health. The specific correspondence is shown below:
+      // Expected override host statues. Every bit in the HostStatusSet represent an enum value
+      // of envoy::config::core::v3::HealthStatus. The specific correspondence is shown below:
       //
-      // * 0b001: Host::Health::Unhealthy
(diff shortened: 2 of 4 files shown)


wbpcode:
cc @markdroth

RyanTheOptimist (commented):
This change looks good to me, but I wonder if there should be a runtime guard for the behavior change. @yanavlasov what do you think?

wbpcode:
friendly ping cc @yanavlasov

wbpcode:
or should I add a runtime guard directly? cc @RyanTheOptimist

RyanTheOptimist:
Sounds like crickets on maintainers channel. Since neither @yanavlasov nor I are sure, let's go ahead and add a guard. Hopefully it's not too much of a hassle. Sorry about that!

wbpcode:
Get it. Thanks guys.

yanavlasov:
Sounds good. Runtime guard it is.

/wait

RyanTheOptimist (dismissed):

RyanTheOptimist (approved):

PR #24096, opened 2022-11-21
avoid calling ssl alpn function twice

ramaraochavali:
Commit Message: avoid calling ssl alpn function twice
Additional Description: To determine the next protocol, this calls SSL's alpn function twice. This PR avoids that.
Risk Level: Low
Testing:  Existing tests
Docs Changes: N/a
Release Notes: N/a


source/common/http/conn_manager_utility.cc
```diff
@@ -38,8 +38,9 @@ absl::string_view getScheme(absl::string_view forwarded_proto, bool is_ssl) {
 } // namespace
 std::string ConnectionManagerUtility::determineNextProtocol(Network::Connection& connection,
                                                             const Buffer::Instance& data) {
-  if (!connection.nextProtocol().empty()) {
-    return connection.nextProtocol();
+  const std::string& next_protocol = connection.nextProtocol();
```
kyessenov, line 41: `nextProtocol` returns a string value, so this is a reference to a temporary. Can you just use `std::string` for the type?

kyessenov (approved):

PR #24104, opened 2022-11-21
Repo merge clean diffs

alyssawilk:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


RyanTheOptimist (commented):
Exciting! Assuming this works (passes CI), LGTM!

.github/workflows/commands.yml
```diff
@@ -1,14 +0,0 @@
-name: commands
-on:
-  issue_comment:
-    types: [created]
-
-jobs:
-  retest:
-    name: Retest
```
jpsim: FYI this means that `/retest` won't retest failed GitHub Actions jobs.

Where could I look at how Repokitteh is setup so maybe I could teach it to do that?

.github/workflows/coverage.yml
```diff
@@ -21,7 +23,7 @@ jobs:
         with:
           submodules: true
       - name: Add safe directory
-        run: git config --global --add safe.directory /__w/envoy-mobile/envoy-mobile
+        run: git config --global --add safe.directory /__w/envoy/envoy
```
jpsim, line 26: I don't think these are needed anymore because there's no git submodule involved anymore, just the main git repo.

.github/workflows/format.yml
```diff
@@ -54,8 +46,8 @@ jobs:
           submodules: true
       - name: 'Run DrString'
         env:
-          DEVELOPER_DIR: /Applications/Xcode_14.1.app
-        run: ./bazelw run @DrString//:drstring check
+          DEVELOPER_DIR: /Applications/Xcode_14.0.app
```
jpsim, line 49: Why did we revert from Xcode 14.1 to Xcode 14.0 here?

.github/workflows/commands.yml
alyssawilk: I'm fine adding it back, I just was trying to start with CI workflows that verified we didn't break the build :-)
repokitteh files for Envoy are here:  ci/repokitteh/modules

.github/workflows/coverage.yml
alyssawilk: CI broke without them, and told me to add them back.

jpsim: Odd.

.github/workflows/format.yml
jpsim: https://github.com/envoyproxy/envoy/pull/24223

.github/workflows/commands.yml
phlax: related #24156

.github/workflows/format.yml
alyssawilk: argh,  my original pull was before the bump and the . files kept being problematic.  will fix thanks for the catch

PR #24112, opened 2022-11-21
Move simple_http_cache to a consistent path

ravenblackx:
Commit Message: Move simple_http_cache to a consistent path
Additional Description: Per discussion with htuch@ and mattklein123@ [here](https://github.com/envoyproxy/envoy/pull/23175#discussion_r1014260888), this change moves simple_http_cache from:
api: `api/extensions/cache/simple_http_cache`
category: `envoy.extensions.http.cache`
impl: `source/extensions/filters/http/cache/simple_http_cache`
to
api: `api/extensions/http/cache/simple_http_cache`
category: `envoy.extensions.http.cache`
impl: `source/extensions/http/cache/simple_http_cache`

Risk Level: Low. Extension is both WIP and not recommended for production.
Testing: Existing tests still pass.
Docs Changes: Updated paths.
Release Notes: api path changed from `api/extensions/cache/simple_http_cache` to `api/extensions/http/cache/simple_http_cache`, and source code moved, to match extension category.


adisuissa (commented):
This is a breaking change, but is ok due to being WIP.
LGTM, modulo release notes.
/lgtm api

changelogs/current.yaml
```diff
@@ -26,6 +26,9 @@ minor_behavior_changes:
 - area: cache_filter
   change: |
     add a completion callback to updateHeaders interface. Any external cache implementations will need to update to match this new interface. See changes to simple_http_cache in PR#23666 for example.
+- area: cache_filter
```
adisuissa, line 29: nit: I suggest moving it up to `behavior changes`, and explicitly say that "api path of work-in-progress extension changed from..."

adisuissa (commented):
Thanks!

changelogs/current.yaml
ravenblackx: Added clarity, but I think minor behavior changes is the right category: `# *Changes that may cause incompatibilities for some users, but should not for most*`
(Given that it's an extension that's WIP *and* explicitly not production ready, it seems like even "some users" should be an overestimate of the impact.)

adisuissa (dismissed):
/lgtm api
LGTM, thanks!

yanavlasov (approved):

PR #24118, opened 2022-11-21
test: skipping envoy coverage CI for E-M changes

alyssawilk:
If this works, will apply to most Azure workflows.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

co-authored-by phlax


alyssawilk:
cc @phlax I'd like your thoughts on this

I like what Enovy Mobile does to avoid spurious CI work: 
https://github.com/envoyproxy/envoy-mobile/blob/main/.github/workflows/android_tests.yml#L23
but I'm not sure of the AZP way to do if, and I thought tweaking do_ci would be not much hackier but want to get your thoughts before I put more time into it.

If SGTY I'll extend to skipping all tests for mobile (and if you'd like, docs)-only changes

phlax:
my ideal is that we can just run `bazel test //...` and it runs only and all things that should be run

in reality, we are quite far off from that so i guess `git diff` and bail hacks will have to do for now (there are a few problems with this pattern but its the easiest way i think)

what we could probably do is run some early ci that puts vars into the azp namespace - ie `ENVOY_HAS_CHANGES` , `MOBILE_HAS_CHANGES` and then we can get azp to skip specific parts of the CI altogether (ie no git checkouts etc)

phlax:
canonical example for setting vars from inside tasks/steps

```yaml
variables:
- name: one
  value: initialValue 

steps:
  - script: |
      echo ${{ variables.one }} # outputs initialValue
      echo $(one)
    displayName: First variable pass
  - bash: echo "##vso[task.setvariable variable=one]secondValue"
    displayName: Set new variable value
  - script: |
      echo ${{ variables.one }} # outputs initialValue
      echo $(one) # outputs secondValue
    displayName: Second variable pass

```

from https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch

alyssawilk:
cc @phlax so more like this?
I'm less confident this works - not sure how to test other than either submitting and trying a mobile only change, or adding some extra excludes for a test run  WDYT?

.azure-pipelines/pipelines.yml
```diff
@@ -23,6 +23,8 @@ variables:
 - name: isStableBranch
   # A release branch can be either `main` or a `release/v1.x` stable branch
   value: $[or(eq(variables['isMain'], 'true'), eq(variables['isReleaseBranch'], 'true'))]
+- name: nonMobileFilesChanged
+  value: git diff --name-only origin/main | grep -qe common/
```
phlax, line 27:
```suggestion
  value: True
```

.azure-pipelines/pipelines.yml
```diff
@@ -23,6 +23,8 @@ variables:
...
 
 stages:
```
phlax, line 29:
```suggestion
stages:
- stage: change_detection
  jobs:
  - job: format
    dependsOn: []
    pool:
      vmImage: "ubuntu-20.04"
    steps:
    - bash: |
        git diff --name-only origin/main | grep -qe common/ \
          || echo "##vso[task.setvariable variable=nonMobileFilesChanged]False"
      displayName: "Detect repo changes"
      workingDirectory: $(Build.SourcesDirectory)

```

phlax: `git diff --name-only origin/main | grep -qe common/` this doesnt seem right - but i think this is ~what we need

i believe this will effect subsequent stages but not sure i can test further

incidentally the pattern is pretty similar with github actions iirc - although not sure if we can do that once across all jobs

would be good to turn off mobile linting etc when its an envoy only build

phlax: also not 100% about bools in azp yaml

phlax: we might need/want to force other stages to depend on this one

phlax:
running some tests in #24204

.azure-pipelines/pipelines.yml
```diff
@@ -23,8 +23,23 @@ variables:
...
+- stage: change_detection
+  jobs:
+  - job: repo_change
+    dependsOn: []
+    pool:
+      vmImage: "ubuntu-20.04"
+    steps:
+    - bash: |
+        git diff --name-only origin/main | grep -qe common/ \
```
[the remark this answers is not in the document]
phlax: thinking about this we could skip coverage for a whole bunch of changes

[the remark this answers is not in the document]
phlax: a much better (working) suggestion for the git command to the one above:

```bash
git diff-index --quiet upstream/$(System.PullRequest.TargetBranch) :^mobile \
    || ...
```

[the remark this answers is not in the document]
phlax: this is still problematic as it only works on pull requests

for main/release branches we probably want to compare to the last commit (or commit of last landed pr in case of release branches)

phlax:
arggh

testing this further and it would seem doing `echo "##vso[task.setvariable variable=nonMobileFilesChanged]false` only sets the value in the current stage and not in the `task` as it would suggest

reading the docs more thoroughly seems to confirm this altho they also contradict what it does do

this pattern can still be useful if for example we wanted to prevent some or all of the check jobs from running - we can still just check this once and bail, we just cant do it across the entire pipeline

phlax:
hmm, actually does seem possible (see eg https://praveenkumarsreeram.com/2022/05/03/azure-devops-tips-and-tricks-5-how-to-pass-values-between-tasks-in-a-pipeline-using-task-setvariable-logging-command/)

the variable needs to be set with `isoutput=true` and the subsequent references to it need to prefix with the "TaskName" 

ill test this further

PR #24132, opened 2022-11-21
[balsa] Fix 304WithBody test and add new one.

bencebeky:
Add //test/common/http/http1:codec_impl_test test case for 304 response with body for both BalsaParser and http-parser.  In this test, the body is dispatched to the parser in the same Parser::execute() call as the headers.  In this case, the parser treats the response as invalid due to the presence of the body.  Note that multiple requests cannot be parsed in a single Parser::execute() call.

Modify ProtocolIntegrationTest.304WithBody by appending '\r\n' to the end of the "body" of the 304 response, necessary for BalsaParser to signal an error.  This is because the "body" is fed into the parser via a separate Parser::execute() call, so it is in fact treated as a second response.

Commit Message: [balsa] Fix 304WithBody test and add new one.
Additional Description:
Docs Changes: n/a
Release Notes: n/a


test/integration/protocol_integration_test.cc
```diff
@@ -1656,7 +1651,10 @@ TEST_P(ProtocolIntegrationTest, 304WithBody) {
     ASSERT_FALSE(response->complete());
   }
 
-  upstream_request_->encodeData(2, true);
+  // Note that at this point the parser is expecting a new response, and it
+  // signals an error because it cannot parse the received data as a valid
+  // status line.
+  upstream_request_->encodeData("aa\r\n", true);
```
diannahu, line 1657: I think this makes sense and is reasonable, but I also wonder if this kind of diff (IIUC where the legacy parser can signal an error right away, and Balsa needs the \r\n) may come back to bite at some point. I'm guessing this diff is present for a general scenario of an invalid response that doesn't have \r\n, so I think it makes sense for this change to go through and consider the general scenario separately. (And maybe the result of that consideration is that this is NBD.) WDYT? Thanks!

diannahu (approved):

test/integration/protocol_integration_test.cc
bencebeky: Yes, this is definitely worth keeping in mind.  This difference has already come up, for example, at https://github.com/envoyproxy/envoy/pull/22525/files.  Reworking Balsa would requiring not only some engineering work, but also every functional change faces tremendous operational inertia in Google's deployment.  My plan is to proceed with this current behavioral change, and revisit only if a business need arises from Envoy's side.

diannahu: Makes sense, thanks!

diannahu (approved):

test/integration/protocol_integration_test.cc
KBaichoo: +1 on looking at this holistically in #21245

KBaichoo (approved):
Looks good, thank you!

PR #24145, opened 2022-11-22
docker: Don't include su-exec in the distroless image

mkauf:

Risk Level: Low


phlax (approved):
nice, thanks @mkauf, lgtm

PR #24146, opened 2022-11-22
Move Http1ParserImpl and pretty-printer to test_common/utility.h.

bencebeky:
Consolidate these existing implementations.

Commit Message: Move Http1ParserImpl and pretty-printer to test_common/utility.h.
Additional Description:
Risk Level: low, test-only refactor
Testing: test-only change
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/test_common/utility.h
+++ b/test/test_common/utility.h
@@ -143,6 +143,12 @@ private:
   RealTimeSource real_time_source_;
 };
 
+// See https://github.com/envoyproxy/envoy/issues/21245.
+enum class Http1ParserImpl {
+  HttpParser, // http-parser from node.js
+  BalsaParser // Balsa from QUICHE
+};
+
 class TestUtility {
 public:
   /**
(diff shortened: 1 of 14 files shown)


bencebeky:
Kevin: PTAL. This is a test-only refactor with no behavior change. I'm open to alternate naming suggestions, naming is hard.  Note that Http2Impl (oghttp2 and nghttp2) is both for parsing and serializing, whereas Http1ParserImpl (Balsa and http-parser) is only for parsing, hence the name difference. Thank you.

test/test_common/utility.h
```diff
@@ -534,6 +540,11 @@ class TestUtility {
     return message;
   }
 
+  // Allows pretty printed test names.
+  static std::string http1ParserImplToString(Http1ParserImpl impl) {
+    return impl == Http1ParserImpl::HttpParser ? "HttpParser" : "BalsaParser";
```
KBaichoo, line 545: nit: could be done in a follow up, I think https://github.com/envoyproxy/envoy/pull/24146/files#diff-f21b0fd69ad347bf91f43b74096401ae3477f7a9a0a2660b3598a96565116b10L75-L84 which implements similar pretty printing is more robust. 

Pragmatically I don't think we'd get additional enums here, but if we did we'd silently clobber it :(.

KBaichoo (approved):
Looks good, thank you!

test/test_common/utility.h
bencebeky: Thanks for the suggestion.  Following up in #24226.

Issue #18819, opened 2021-10-28
Support HTTP extension (custom) methods


yanavlasov:
Presently H/1 codec has a hardcoded list of HTTP methods it accepts. This prevents applications that use extension methods [(see RFC)](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.1) from working.

bencebeky:
The current HTTP/1 parser is http-parser, which cannot be configured to accept custom methods.  There is ongoing effort tracked at #21245 to migrate to Balsa as HTTP/1 parser.  This can allow custom methods.  When UHV is enabled, it will allow users to configure whether custom methods should be rejected via the restrict_http_methods field at https://github.com/envoyproxy/envoy/blob/140dd100d1c97e7c1bf2ef6113441bed07d556b9/api/envoy/extensions/http/header_validators/envoy_default/v3/header_validator.proto#L139-L142.

bencebeky:
The header validation logic in UHV already supports custom methods:

https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/extensions/http/header_validators/envoy_default/header_validator.cc#L99-L106

However, BalsaParser currently does not skip comparison of method to the hardcoded list regardless of how UHV is configured:

https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L44-L45

vamshi177:
Hi,
Recently, we had a use case requiring support for custom methods in our environments. However, using `http_protocol_options.allow_custom_methods` didn't fully resolve the issue, as the calls were being rejected by the client Envoy at the `http_inspector`, which still uses `http_parser` and doesn't currently support custom methods.

I'm working on a change to enable custom methods in http_parser. Could you please review [this change](https://github.com/envoyproxy/envoy/pull/36245)  Thank you!

PR #24157, opened 2022-11-22
[balsa] Call ParserCallback::onMessageBegin() early.

bencebeky:
When the first byte of the message arrives, check if it is the first character of a valid method name for a request, or if it is 'H' for a response (responses must start with "HTTP").  If so, call ParserCallback::onMessageBegin() immeditately.  If not, signal an error immediately.

This is already the behavior of http-parser.

Add three tests.  Unfortunately the response starting with 'H' case cannot be easily tested, as ParserCallbacks are not directly accessible, and ClientConnectionImpl::onMessageBegin() does not have any relevant side effects that can be easily tested.

Also add comment referencing #18819.  A way to disable method name validation will be needed in the future.

Commit Message: [balsa] Call ParserCallback::onMessageBegin() early.
Additional Description:
Risk Level: low, changed code protected by existing default-false runtime flag.
Testing: //test/common/http/http1:codec_impl_test, //test/integration:idle_timeout_integration_test
Docs Changes: n/a
Release Notes: n/a


source/common/http/http1/balsa_parser.cc
```diff
@@ -268,6 +268,7 @@ void BalsaParser::MessageDone() {
   }
   status_ = convertResult(connection_->onMessageComplete());
   framer_.Reset();
+  on_message_begin_called_ = false;
```
diannahu, line 271: completely out of scope for this change, okay to consider later: Would it be right to reset other members also, e.g., headers_done_ and/or others? IIUC, resetting on_message_begin_called_ sort of suggests that we can reuse BalsaParser for a subsequent message (hm, or maybe just subsequent calls to execute()?).

diannahu (approved):
A nice way to begin execution!

source/common/http/http1/balsa_parser.cc
bencebeky: Yes, I think that makes perfect sense, I'll look into it.

birenroy:
LGTM. Let me know when you'd like us to take another look.

KBaichoo (approved):
Looks good, thank you!

bencebeky:
Test failure still looks unrelated.

/retest

source/common/http/http1/balsa_parser.cc
bencebeky: Thanks for catching this!  I filed #24557 for resetting `headers_done_`.  I couldn't identify other members that would need to be reset.  I'm fixing this, among other things, with test coverage for plenty of weird cases, at #24594.

PR #24165, opened 2022-11-23
contrib-sipproxy: rework sipproxy stats

durd07:
Signed-off-by: Felix Du <durd07@gmail.com>

Commit Message: rework sipproxy stats to improve performance
Additional Description:
Risk Level: Low
Testing: Ut/Mt
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jmarantz:
CI needs to be sorted out. Thanks!

/retest

durd07:
@dorisd0102  Please help to review this PR.
Thanks.

dorisd0102:
LGTM

mattklein123 (changes requested):
/wait

contrib/sip_proxy/filters/network/source/conn_manager.cc
```diff
@@ -31,7 +31,17 @@ TrafficRoutingAssistantHandler::TrafficRoutingAssistantHandler(
...
-  if (cache_manager_[type][key] != val) {
+
+  bool should_update_tra = false;
+  try {
+    if (cache_manager_.at(type, key) != val) {
+      should_update_tra = true;
+    }
+  } catch (std::out_of_range) {
+    should_update_tra = true;
+  }
```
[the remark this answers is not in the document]
durd07: Hi @mattklein123 ,
Thanks for your comments, but why exceptions are not recommended?
I want to implement a similiar method like [std::map::at](https://en.cppreference.com/w/cpp/container/map/at)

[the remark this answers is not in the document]
mattklein123: Because a) we are trying to avoid exceptions at this point and b) this is not a good use of exceptions. Use find to return an iterator.

contrib/sip_proxy/filters/network/source/conn_manager.cc
```diff
@@ -32,13 +32,9 @@ void TrafficRoutingAssistantHandler::updateTrafficRoutingAssistant(
...
-  bool should_update_tra = false;
-  try {
-    if (cache_manager_.at(type, key) != val) {
-      should_update_tra = true;
-    }
-  } catch (std::out_of_range) {
-    should_update_tra = true;
+  bool should_update_tra = true;
+  if (cache_manager_.contains(type, key) && (cache_manager_.at(type, key) == val)) {
+    should_update_tra = false;
```
mattklein123, line 37: You can void a double lookup by using find(). Same elsewhere.

contrib/sip_proxy/filters/network/source/utility.h
```diff
@@ -261,6 +261,7 @@ template <typename T, typename K, typename V> class CacheManager {
 
   /**
    * NOTE: This may throw exception std::out_of_range if type or key doesn't exist.
+   * use contains(type, key) to check exist first.
    */
```
mattklein123, line 265: Rework to always use find. Never us at.

mattklein123 (changes requested):
/wait

contrib/sip_proxy/filters/network/source/utility.h
durd07: @mattklein123 Thanks for your comments.

Firstly I want to implement a `find` method like what does as `std::map`, but I can't figure out the solution because different iterator maybe returned in a nested map, so write a method called `at` which return empty value if not exist.
```
[[ret_type]] find(const T& type, const K& key) {
  auto it = caches_.find(type);
  if (it != caches_.cend()) {
    return it->second.find(key);    // iterator of Cache<K, V>::iterator
   } else {
    return [[ the iterator of caches_ ]]
  }
}
```

jmarantz: drive-by comment: if your goal is to return an optional value if it's in one of several collections, you could return absl::optional<V>, or OptRef<V> (see envoy/common/optref.h) if you want V to be a reference.

durd07: Thanks @jmarantz , I already aware of this, but I don't want to make thing more complex.

durd07:
@mattklein123 @jmarantz 
Thanks for your comments, all of them are addressed.

mattklein123 (approved):
Thanks!

PR #24166, opened 2022-11-23
utility: fix broken URL in checkTransportVersion() utility method

agrawroh:
This PR fixes the broken URL link in `checkTransportVersion()` method inside utility file by pointing it to `/latest` so that it points to the correct page i.e. https://www.envoyproxy.io/docs/envoy/latest/faq/api/envoy_v3

**Commit Message:** Fixes broken URL in `checkTransportVersion()`.
**Additional Description:** This PR fixes the broken URL for the warning that gets printed while using xDS v2.
**Risk Level:** Very Low
**Testing:** N/A
**Docs Changes:** N/A
**Release Notes:** N/A

### DIFF
--- a/source/common/config/utility.h
+++ b/source/common/config/utility.h
@@ -208,13 +208,12 @@ public:
     const auto transport_api_version = api_config_source.transport_api_version();
     ASSERT_IS_MAIN_OR_TEST_THREAD();
     if (transport_api_version != envoy::config::core::v3::ApiVersion::V3) {
-      const ApiVersion version = ApiVersionInfo::apiVersion();
       const std::string& warning = fmt::format(
           "V2 (and AUTO) xDS transport protocol versions are deprecated in {}. "
           "The v2 xDS major version has been removed and is no longer supported. "
           "You may be missing explicit V3 configuration of the transport API version, "
-          "see the advice in https://www.envoyproxy.io/docs/envoy/v{}.{}.{}/faq/api/envoy_v3.",
-          api_config_source.DebugString(), version.major, version.minor, version.patch);
+          "see the advice in https://www.envoyproxy.io/docs/envoy/latest/faq/api/envoy_v3.",
+          api_config_source.DebugString());
       ENVOY_LOG_MISC(warn, warning);
       throw DeprecatedMajorVersionException(warning);
     }


wbpcode:
@agrawroh Thanks. Seem cc @phlax would be better choice to review this PR. Thanks again.

source/common/config/utility.h
```diff
@@ -208,13 +208,12 @@ class Utility {
     const auto transport_api_version = api_config_source.transport_api_version();
     ASSERT_IS_MAIN_OR_TEST_THREAD();
     if (transport_api_version != envoy::config::core::v3::ApiVersion::V3) {
-      const ApiVersion version = ApiVersionInfo::apiVersion();
       const std::string& warning = fmt::format(
           "V2 (and AUTO) xDS transport protocol versions are deprecated in {}. "
           "The v2 xDS major version has been removed and is no longer supported. "
           "You may be missing explicit V3 configuration of the transport API version, "
-          "see the advice in https://www.envoyproxy.io/docs/envoy/v{}.{}.{}/faq/api/envoy_v3.",
-          api_config_source.DebugString(), version.major, version.minor, version.patch);
```
phlax, line 217: @agrawroh im wondering how this is broken

if i browse for example to https://www.envoyproxy.io/docs/envoy/v1.24.0/faq/api/envoy_v3 it works as expected - i think we want the message to point to the correct version of the docs

agrawroh: @phlax For some reason it was pointing to `v3.0.0` when I was using v1.24 and trying to add the ExtAuthZ gRPC filter without setting `transport_api_version`.  

```
2022/10/18 20:17:48 WARN DockerService[Thread-4] : envoy-a371437ef2b74340b34a4431c13bb821 - }
2022/10/18 20:17:48 WARN DockerService[Thread-4] : envoy-a371437ef2b74340b34a4431c13bb821 - . The v2 xDS major version has been removed and is no longer supported. You may be missing explicit V3 configuration of the transport API version, see the advice in https://www.envoyproxy.io/docs/envoy/v3.0.0/faq/api/envoy_v3.
```

phlax: hmm, yep i see its pointing to the api version, not the Envoy version

@wbpcode any ideas on how to fix that ?

wbpcode: cc @phlax IMO, simple `latest` is good enough. 
And the `BUILD_VERSION_NUMBER` would be a althernative. But if it's end with `-dev`, we need to calculate the suitable version first.

phlax: not sure i follow exactly - if it ends with `-dev` then im happy with it just going to `latest` docs

if the binary is (release) versioned then im more concerned it points to the correct version of the docs

we get less now, but historically we got a lot of tickets where there was a mismatch between running version and the docs the user was reading

phlax: ok - lets just go with `latest` on the basis that its better than what we have now

i think we will probably want to revisit this at some point, as the other issue is that there is no guarantee the page exists in `latest` only in the current branch - i have removed/fixed similar links for this reason in the past

phlax:
ci is broken - just waiting on fix to land ...

https://github.com/envoyproxy/envoy/pull/24169

agrawroh:
> ci is broken - just waiting on fix to land ...
> 
> #24169

Ahh, Okay. Thanks for the heads up!

phlax:
should work now

/retest

agrawroh:
@phlax @wbpcode CI Passed this time. Could you please take a look?

phlax (approved):
lgtm, thanks @agrawroh

Issue #23758, opened 2022-10-31
Merge Envoy mobile code into the Envoy repo


alyssawilk:
Almost all files are copied from envoy-mobile repo to envoy/mobile  Source of truth for mobile changes is still envoy-mobile.  @alyssawilk is doing updates until we flip source of truth to envoy/mobile.

TODO: 
- run relevant Envoy workflows when dependencies are bumped (phlax)

Thursday(ish) switch source of truth from envoy-mobile to envoy/mobile
 - copy over PRs which have landed since then
 - notify on various slack channels / notify on PRs in flight

Post-move
- fix up / move github release cron jobs
-  Circleci functionality (e.g. docs) https://github.com/envoyproxy/envoy/pull/24295
- Update references to envoyproxy/envoy-mobile repo in source code and docs (Tackle after Thursday)
- Archive envoyproxy/envoy-mobile repo (Tackle after release workflows moved over)

alyssawilk:
Known missing things
-- swift lint because I can't get it to work.
-- perf (https://github.com/envoyproxy/envoy/pull/24153)
-- release validation 
-- release

-- all cron workflows (weekly release, release, etc)

alyssawilk:
perf turned up in  (https://github.com/envoyproxy/envoy/pull/24153)

alyssawilk:
@jpsim @Augustyniak I think we're at a point we should flip source of truth to the Envoy repo, and do a weekly copy into the E-M repo until we fix up tagging and releases.  I've snagged some time on JP's calendar to talk about it next week on the assumption he'll find a time that works and invite whoever is relevant on the Lyft side.

jpsim:
Some other migration tasks that come to mind:

* Update references to `envoyproxy/envoy-mobile` repo in source code and docs
* Enable `/retest` for GitHub Actions: https://github.com/envoyproxy/envoy/issues/24156
* Archive `envoyproxy/envoy-mobile` repo

Augustyniak:
Related fix: https://github.com/envoyproxy/envoy/pull/24228

jpsim:
> run relevant Envoy workflows when dependencies are bumped (phlax)

Done in https://github.com/envoyproxy/envoy/pull/24407

jpsim:
> fix up / move github release cron jobs

This still needs to be done. We discussed this at last week's Envoy Mobile community call, but didn't come to a clear conclusion.

[...]

PR #24171, opened 2022-11-23
mobile: turning up release validation CI

alyssawilk:
copied from the mobile repo, and edited to run for repo merge.  Only runs for mobile PRs.

Risk Level: low
Testing: first push shows it worked successfully
Docs Changes: n/a
Release Notes: n/a
part of https://github.com/envoyproxy/envoy/issues/23758

### DIFF
--- a/.github/workflows/release_validation.yml
+++ b/.github/workflows/release_validation.yml
@@ -6,9 +6,11 @@ on:
       - main
     paths:
     - 'mobile/**'
+    - '.github/workflows/**'
   pull_request:
     paths:
     - 'mobile/**'
+    - '.github/workflows/**'
 
 jobs:
   validate_swiftpm_example:
--- a/.github/workflows/release_validation.yml
+++ b/.github/workflows/release_validation.yml
@@ -4,7 +4,11 @@ on:
   push:
     branches:
       - main
+    paths:
+    - 'mobile/**'
   pull_request:
+    paths:
+    - 'mobile/**'
 
 jobs:
   validate_swiftpm_example:
--- a/.github/workflows/release_validation.yml
+++ b/.github/workflows/release_validation.yml
@@ -1,0 +1,34 @@
+name: release_validation
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  validate_swiftpm_example:
+    name: validate_swiftpm_example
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -20,6 +20,7 @@
 !\.yapfignore
 !\.zuul
 !\.zuul.yaml
+!\.github/workflows/*
```
phlax, line 23: please dont do this - rather clean up the yaml

phlax (commented):
looks good, altho im a little concerned that we are adding a lot of CI that will still run when not strictly required

the `paths` (and `paths-ignore`) pattern for workflow triggers is cool - i checked re azp to do the same thing - it seems we can altho i think we probably still need to use the pattern we have discussed as it is more about not/triggering steps _within_ a workflow there

i was a little bit confused by the `.gitignore` change - mostly because i could see that the yaml should not pass linting so assumed that it was to prevent linting 

i realise now that it does the opposite - ie ensure those files are included - but its not necessary as there is already:

```gitignore
 15 !\.github
```

checking this confirms

```console
$ touch .github/workflows/asdf
$ git status .github/workflows/
On branch tooling-yamllint2
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.github/workflows/asdf

```


re linting the reason these files were not picked up is due to it only searching for `.yaml` files and not `.yml` - im fixing that now (https://github.com/envoyproxy/pytooling/pull/623)

and will lint the existing files as that lands

phlax:
yaml lint/cleanup pr is #24197

.gitignore
alyssawilk: it's not necessary but it makes it a lot less annoying to change workflow files as it means git add * will add them.

.github/workflows/release_validation.yml
```diff
@@ -0,0 +1,38 @@
+name: release_validation
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+    - 'mobile/**'
```
phlax, line 8: could you add the `.azure-pipelines/**` path also pleaes

phlax: sorry - not azp - the `.github/workflows/*` path

jpsim (approved):

PR #24172, opened 2022-11-23
test: not validating all protos in every test

alyssawilk:
Currently proto_gen_validate pulls all protos into every unit test, and every unit test validates their existence on start up.  This is pretty terrible.  https://github.com/bufbuild/protoc-gen-validate/pull/738 pulls this out for Envoy Mobile proto-gen-validate-less builds, and so breaks all the unit tests.  Changing the Envoy test framework to only validate proto registration for integration tests.

Risk Level: low (should be test only)
Testing: https://github.com/envoyproxy/envoy/pull/24151
Docs Changes: n/a
Release Notes: n/a


test/integration/base_integration_test.cc
```diff
@@ -47,6 +48,7 @@ BaseIntegrationTest::BaseIntegrationTest(const InstanceConstSharedPtrFn& upstrea
       version_(version), upstream_address_fn_(upstream_address_fn),
       config_helper_(version, *api_, config),
       default_log_level_(TestEnvironment::getOptions().logLevel()) {
+  Envoy::Server::validateProtoDescriptors();
```
adisuissa, line 51: Would like to highlight 2 questions regarding this:
1. Is calling this outside of the mutex ok?
2. Could there be a use of one of the validated protos before the BaseIntergrationTest c'tor is executed?

I'm less worried about 2 because this change only impacts tests, and so I guess we will see failures.
For 1, I'm guessing this is only relevant to multi-threaded use-cases, and AFAIK threading only starts after the integration test is instantiated.

Feel free to correct me if I'm wrong on any of the above.

adisuissa (commented):
Thanks, overall I think this change is ok, just highlighted 2 questions regarding one of the test changes and I'd like to hear your thoughts.

test/integration/base_integration_test.cc
alyssawilk: 1) I think we're largely checking things that are registered at or before start up, so this is fine
2) yes, but then the test will fail with a slightly less helpful warning so if the goal is to fail (it's an ASSERT not RELEASE_ASSERT) we will achieve the goal just less helpfully :-)

adisuissa (approved):
LGTM pending CI, thanks!

PR #24176, opened 2022-11-23
Add file size to DirectoryEntry

ravenblackx:
Commit Message: Add file size to DirectoryEntry
Additional Description: Directory [find] iterator is potentially useful for file system cache and, in future, serving content directly from file system. The iterator already has access to file size from `stat()` or its Windows equivalent, but it is not accessible to the caller. Using the current available APIs, the caller would have to make an additional call to `stat` for each file they wish to know the size of. This change exposes the file size through the existing API, sparing those additional OS calls in the case that the size is useful, at the cost of a small value-copy.
In passing, fixes a mistaken use of `errno` that should have been the error number from the os-call's returned structure, and updates a throw-test to not require a nominally private function to be public.
Risk Level: Low. Simple change, existing fields should be unchanged.
Testing: Tests updated to match new values, one test added to validate that file size is returned.
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Updated both win32 and posix versions.


envoy/filesystem/filesystem.h
```diff
@@ -169,6 +169,9 @@ struct DirectoryEntry {
   // target. For example, if name_ is a symlink to a directory, its file type will be Directory.
   FileType type_;
 
+  // The file size in bytes.
```
jmarantz, line 172: nit: I don't feel a compelling need to document this variable any more than it's well-chosen name :) 

No worries if you prefer to leave it though.

source/common/filesystem/posix/directory_iterator_impl.cc
```diff
@@ -42,39 +42,39 @@ void DirectoryIteratorImpl::nextEntry() {
   }
 
   if (entry == nullptr) {
-    entry_ = {"", FileType::Other};
+    entry_ = {"", FileType::Other, 0};
   } else {
     const std::string current_path(entry->d_name);
     const std::string full_path(directory_path_ + "/" + current_path);
-    entry_ = {current_path, fileType(full_path, os_sys_calls_)};
+    entry_ = makeEntry(entry->d_name);
```
jmarantz, line 49: since you are computing full_path in makeEntry you can drop the 2 temps above.

source/common/filesystem/posix/directory_iterator_impl.cc
```diff
@@ -42,39 +42,39 @@ void DirectoryIteratorImpl::nextEntry() {
...
-  if (S_ISDIR(stat_buf.st_mode)) {
-    return FileType::Directory;
+    if (file_type == FileType::Other) {
+      throw EnvoyException(fmt::format("unable to stat file: '{}' ({})", full_path, errno));
+    }
+  } else if (S_ISDIR(stat_buf.st_mode)) {
+    file_type = FileType::Directory;
   } else if (S_ISREG(stat_buf.st_mode)) {
-    return FileType::Regular;
+    file_type = FileType::Regular;
```
jmarantz, line 75: pre-existing, but should we assign file_type in the 'else' or throw an exception or something?

source/common/filesystem/posix/directory_iterator_impl.cc
```diff
@@ -42,39 +42,39 @@ void DirectoryIteratorImpl::nextEntry() {
...
+        file_type = FileType::Regular;
       }
     }
-    throw EnvoyException(fmt::format("unable to stat file: '{}' ({})", full_path, errno));
-  }
-
-  if (S_ISDIR(stat_buf.st_mode)) {
-    return FileType::Directory;
+    if (file_type == FileType::Other) {
+      throw EnvoyException(fmt::format("unable to stat file: '{}' ({})", full_path, errno));
```
jmarantz, line 70: pre-existing, but can makeEntry be called during a request? It seems like it can, in which case we should switch from throwing an exception to returning an absl::StatusOr. It is not permitted to throw exceptions on worker threads in Envoy.

Actually throwing them is OK if they crash the program (for some definition of OK). But we disallow *catch* in worker threads.

This is beyond the scope of this PR but would it be worth it to throw in a TODO while you are here?

test/common/filesystem/directory_test.cc
```diff
@@ -48,6 +48,15 @@ class DirectoryTest : public testing::Test {
     }
   }
 
+  void addFileWithContents(absl::string_view file_name, absl::string_view contents) {
+    const std::string full_path = absl::StrCat(dir_path_, "/", file_name);
+    {
+      std::ofstream file(full_path);
+      file << contents;
```
jmarantz, line 55: should we check for errors here, even with (say) an EXPECT_* macro?

jmarantz (commented):
lgtm with a few nits.

envoy/filesystem/filesystem.h
ravenblackx: I'd rather leave it because I don't know what to do with the spacing otherwise, haha. A blank line between variables is weird, but with no blank line the comment above the *other* variable seems to cover both.

source/common/filesystem/posix/directory_iterator_impl.cc
ravenblackx: Done.

source/common/filesystem/posix/directory_iterator_impl.cc
ravenblackx: Added TODO. (I too thought throwing an exception was a weird choice here.)

source/common/filesystem/posix/directory_iterator_impl.cc
ravenblackx: It's assigned at construct-time (because it's used both for the 'else' and for the condition in the first if block). Added an else-comment to clarify that the fallthrough is intentional.

test/common/filesystem/directory_test.cc
ravenblackx: Added test-asserts, and one in the pre-existing similar spot too.

envoy/filesystem/filesystem.h
```diff
@@ -169,6 +169,9 @@ struct DirectoryEntry {
...
+  uint64_t size_bytes_;
```
jmarantz, line 173: should we include size_bytes_ in the operator== impl? Actually it is not obvious why you previously needed one of those. Would the default operator not work?

WIth this explicit definition, now it will be different of course as this one ignores the size.

Also, what is this operator== used for?

jmarantz (commented):
lgtm modulo minor remaining nits.

envoy/filesystem/filesystem.h
ravenblackx: Yikes, good catch. The == is at least used by the test, so fixing it revealed that I wasn't testing what I thought I was testing. It also seems to be used in a != operator for the iterator, for the `!= end()` comparison, which didn't really care about the size part. Fixed.

ravenblackx: Oh, and there is no default == operator apparently. I too thought there was for pod types, but apparently not until C++20 and even then you still have to specify it with `=default`. I guess we've both been mixing languages too much!

ravenblackx: I *think* keeping the `==` (but fixed) is the least-confusing option for this. It's unfortunate that iterators use `!=` as the 'is at the end' check. We *could* override that to be specifically a test for end-iterator but that would be way more confusing than having an otherwise-unnecessary == operator.

And having the == operator be only as specific as it needs to be for iterator end detection would also just be more confusing than helpful, and would require rewriting the tests significantly. So I think what we have now (a handwritten default == operator) is the least horrible option.

test/common/filesystem/directory_test.cc
```diff
@@ -50,7 +50,7 @@ class DirectoryTest : public testing::Test {
       const std::string full_path = dir_path_ + "/" + file_name;
       {
         const std::ofstream file(full_path);
-        ASSERT_TRUE(file) << "failed to open test file";
+        EXPECT_TRUE(file) << "failed to open test file";
```
jmarantz, line 53: nit: any reason not to use the filesystem itself here, rather than dropping to iostream? It's got an API for returning the error text which would be nice here.

up to you though, I'm OK leaving it like this, but generally it ruffles my feathers to swallow error messages like this.

ravenblackx: I guess depending on an adjacent thing for the unit tests of a thing is a little dubious? The main reason I did it this way is because it was already done this way and I'm trying to not change everything I look at and think yuck about, because that would take [even more] forever.

jmarantz: I'm not sure why ... that's part of the subsystem under test. 

But I also don't feel that strongly about it.

jmarantz (approved):
/assign-from envoyproxy/senior-maintainers

envoy/filesystem/filesystem.h
jmarantz: sounds good

jmarantz (approved):

envoy/filesystem/filesystem.h
```diff
@@ -169,15 +169,19 @@ struct DirectoryEntry {
   // target. For example, if name_ is a symlink to a directory, its file type will be Directory.
   FileType type_;
 
+  // The file size in bytes for regular files. Should not be relied on for directories,
+  // symlinks, or FileType::Other.
+  uint64_t size_bytes_;
```
snowp, line 174: Any way we can make this less of a footgun? One thing that comes to mind is wrapping it in a call that ASSERTs that `type_` is `Regular`? Or perhaps making it a absl::optional?

snowp (dismissed):
Overall lgtm, just one suggestion

envoy/filesystem/filesystem.h
ravenblackx: Asserting that `type_` is `Regular` doesn't actually solve the problem, because on Windows with a symlink the type is regular and the size is unknown (or maybe it's the size of the linked path, I forget - either way, not the size of the file).

Using `absl::optional` is a good idea though, to make it a hard failure instead of a quiet failure if size is used when it shouldn't be.

source/common/filesystem/win32/directory_iterator_impl.cc
```diff
@@ -34,27 +34,30 @@ DirectoryIteratorImpl& DirectoryIteratorImpl::operator++() {
...
-    entry_ = {std::string(find_data.cFileName), fileType(find_data)};
+    entry_ = makeEntry(find_data);
   }
 
   return *this;
 }
 
-FileType DirectoryIteratorImpl::fileType(const WIN32_FIND_DATA& find_data) const {
+DirectoryEntry DirectoryIteratorImpl::makeEntry(const WIN32_FIND_DATA& find_data) {
+  ULARGE_INTEGER file_size;
```
alyssawilk, line 46: should this all be in the final else?

test/common/filesystem/directory_test.cc
```diff
@@ -15,6 +15,17 @@
 namespace Envoy {
 namespace Filesystem {
 
+// NOLINTNEXTLINE(readability-identifier-naming)
+void PrintTo(const DirectoryEntry& entry, std::ostream* os) {
```
alyssawilk, line 19: sorry if I'm missing something obvious but why NOLINTNEXTLINE? Can you not just have Envoy standard naming?

test/common/filesystem/directory_test.cc
```diff
@@ -43,11 +54,26 @@ class DirectoryTest : public testing::Test {
...
   }
 
+  void addFileWithContents(absl::string_view file_name, absl::string_view contents) {
+    const std::string full_path = absl::StrCat(dir_path_, "/", file_name);
+    {
+      std::ofstream file(full_path);
+      EXPECT_TRUE(file) << "failed to open test file";
+      file << contents;
+      file.close();
+      EXPECT_TRUE(file) << "failed to write to test file";
```
alyssawilk, line 72: sorry ignorance of ostream but if you fail to write will file become null/false?

envoy/filesystem/filesystem.h
```diff
@@ -169,15 +170,19 @@ struct DirectoryEntry {
   // target. For example, if name_ is a symlink to a directory, its file type will be Directory.
   FileType type_;
 
+  // The file size in bytes for regular files. nullopt for FileType::Directory and FileType::Other,
+  // and, on Windows, also nullopt for symlinks, and on Linux nullopt for broken symlinks.
+  absl::optional<uint64_t> size_bytes_;
+
   bool operator==(const DirectoryEntry& rhs) const {
-    return name_ == rhs.name_ && type_ == rhs.type_;
+    return name_ == rhs.name_ && type_ == rhs.type_ && size_bytes_ == rhs.size_bytes_;
```
alyssawilk, line 178: sorry if I'm missing but did we test this new equivalence?

alyssawilk (commented):
Looks good overall!
/wait

envoy/filesystem/filesystem.h
ravenblackx: We did now.

Issue #24154, opened 2022-11-22
Improve memory efficiency of config data structures


yurykats:
When XDS config is converted from protobuf into C++ data structures, those data structures are not always as memory efficient as the protobufs, leading to large memory consumption. 

Optional sub-messages are often stored as full blown objects, even when the config for them isn't present.

An example of that is RetryPolicy in RDS:
* Route holds an instance of RetryPolicyImpl class: https://github.com/envoyproxy/envoy/blob/aa694434a64871c309cfc95a207bbb67494e6c2a/source/common/router/config_impl.h#L1012
* VitrualHost holds an optional instance of the RetryPolicy protobuf: https://github.com/envoyproxy/envoy/blob/aa694434a64871c309cfc95a207bbb67494e6c2a/source/common/router/config_impl.h#L316

Each of them are 200+ bytes. 

Another example are HeaderParsers. There are 2 (request/response) in each Route and each VirtualHost. Even if no headers are present in the config, RDS ends up creating HeaderParser instances: https://github.com/envoyproxy/envoy/blob/aa694434a64871c309cfc95a207bbb67494e6c2a/source/common/router/header_parser.cc#L254

A better approach would be to keep a unique_ptr member and only initialize it when config is present. If not, a reference to a default ConstSingleton object can be used.

yurykats:
no stalebot

PR #24182, opened 2022-11-23
Reduce VirtualHost memory utilization by avoiding CatchAllVirtualCluster when not needed

yurykats:
Reduce VirtualHost memory utilization by avoiding CatchAllVirtualClusterinstance when not needed.

Currently an instance of CatchAllVirtualCluster is always created for each vhost, but is only used when there are other virtual clusters (and none of them matches). If there are no virtual clusters to begin with, CatchAllVirtualCluster is a pure overhead.

This PR switches an instance to a unique_ptr and only creates one when there are virtual clusters present.

Removed vcluster stats from xfcc_integration_test, since the test does not setup vclusters and the stats are no longer created by default (they were never updated before). vcluster stats are tested in router_test, tag_extractor_impl_test and http_integration.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


adisuissa:
Assigning Stephan (and myself, on behalf of Yury's request)
/assign @zuercher

source/common/router/config_impl.h
```diff
@@ -315,7 +315,7 @@ class VirtualHostImpl : public VirtualHost, Logger::Loggable<Logger::Id::router>
   const bool include_attempt_count_in_response_;
   absl::optional<envoy::config::route::v3::RetryPolicy> retry_policy_;
   absl::optional<envoy::config::route::v3::HedgePolicy> hedge_policy_;
-  const CatchAllVirtualCluster virtual_cluster_catch_all_;
+  std::unique_ptr<CatchAllVirtualCluster> virtual_cluster_catch_all_;
```
adisuissa, line 318: Why was the const removed?

source/common/router/config_impl.cc
```diff
@@ -1733,10 +1731,14 @@ VirtualHostImpl::VirtualHostImpl(
...
-        VirtualClusterEntry(virtual_cluster, *vcluster_scope_,
-                            factory_context.routerContext().virtualClusterStatNames()));
+  if (!virtual_host.virtual_clusters().empty()) {
+    for (const auto& virtual_cluster : virtual_host.virtual_clusters()) {
+      virtual_clusters_.push_back(
+          VirtualClusterEntry(virtual_cluster, *vcluster_scope_,
+                              factory_context.routerContext().virtualClusterStatNames()));
+    }
+    virtual_cluster_catch_all_ = std::make_unique<CatchAllVirtualCluster>(
+        *vcluster_scope_, factory_context.routerContext().virtualClusterStatNames());
```
adisuissa, line 1741: IIUC vcluster_scope_ is passed and is not const (i.e., its state may be updated in the `CatchAllVirtualCluster` c'tor). I'm not sure if ordering of operations is important, but I guess to make it consistent with the previous code, this line can be executed at the beginning of the if block.

adisuissa (commented):
Thanks, left a couple of minor comments.

source/common/router/config_impl.h

source/common/router/config_impl.cc

adisuissa (dismissed):
LGTM, thanks!

adisuissa (commented):
Can you please check CI, the failure of `//test/integration:xfcc_integration_test` seems consistent.

yurykats:
Removed vcluster stats from xfcc_integration_test, since the test does not setup vclusters and the stats are no longer created by default (they were never updated before).
vcluster stats are tested in router_test, tag_extractor_impl_test and http_integration.

adisuissa:
> Removed vcluster stats from xfcc_integration_test, since the test does not setup vclusters and the stats are no longer created by default (they were never updated before). vcluster stats are tested in router_test, tag_extractor_impl_test and http_integration.

I think that is a valid claim, but want to another input. @jmarantz IIUC this PR removes some stats that currently appear, and as such it is a use-facing change. Assuming that these stats are always 0 in this case because vhost isn't explicitly configured (@yurykats correct me if I'm wrong), is it ok to remove them? If so, other than updating the release notes, is there anything else to consider or update?

jmarantz:
I don't have a strong opinion about this particular stat. I'm not sure what monitoring might break for various Envoy deployments.  I feel like @zuercher would be able to provide insight or redirect to someone else if needed.

yurykats:
> We need to add a release note that describes that change, however.

Thanks for the review, added to release notes.

adisuissa (approved):
LGTM, thanks!

jmarantz:
Removing myself as my review is not needed on this PR and I didn't actually review it :)

zuercher (approved):

PR #24184, opened 2022-11-23
jwt_authn: fix jwt cache bug with requirement_with_audiences

qiwzhang:

If jwt_cache is enabled, it may break [ProviderWithAudiences](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto#L398) feature. 

The `ProviderWithAudiences` may need to check different audiences. But when a cached jwt is used, the audiences will not be checked.  

Risk Level:  Low
Testing:  Unit-test
Docs Changes: No
Release Notes: Yes

### DIFF
--- a/source/extensions/filters/http/jwt_authn/authenticator.cc
+++ b/source/extensions/filters/http/jwt_authn/authenticator.cc
@@ -163,23 +163,29 @@ void AuthenticatorImpl::startVerify() {
   curr_token_ = std::move(tokens_.back());
   tokens_.pop_back();
 
+  bool use_jwt_cache = false;
+  Status status;
   if (provider_.has_value()) {
     jwks_data_ = jwks_cache_.findByProvider(*provider_);
     jwt_ = jwks_data_->getJwtCache().lookup(curr_token_->token());
     if (jwt_ != nullptr) {
-      handleGoodJwt(/*cache_hit=*/true);
-      return;
+      jwks_cache_.stats().jwt_cache_hit_.inc();
+      use_jwt_cache = true;
+    } else {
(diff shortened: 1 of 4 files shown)


jmarantz:
Please merge main. Thank you!

/wait

source/extensions/filters/http/jwt_authn/authenticator.cc
```diff
@@ -163,23 +163,29 @@ void AuthenticatorImpl::startVerify() {
...
 
-  ENVOY_LOG(debug, "{}: Parse Jwt {}", name(), curr_token_->token());
-  owned_jwt_ = std::make_unique<::google::jwt_verify::Jwt>();
-  Status status = owned_jwt_->parseFromString(curr_token_->token());
-  jwt_ = owned_jwt_.get();
+  if (!use_jwt_cache) {
+    ENVOY_LOG(debug, "{}: Parse Jwt {}", name(), curr_token_->token());
+    owned_jwt_ = std::make_unique<::google::jwt_verify::Jwt>();
+    status = owned_jwt_->parseFromString(curr_token_->token());
+    jwt_ = owned_jwt_.get();
```
lizan, line 183: `jwt_` becomes a dangling pointer after this block?

lizan (changes requested):

source/extensions/filters/http/jwt_authn/authenticator.cc
qiwzhang: Not, its object is owned by `owned_jwt_` which will be freed after the whole Authentication.

For jwt cache hit   `jwt_`  points to one owned in the cache.

lizan (approved):

PR #24198, opened 2022-11-24
network: support hashable address for ConnectionHandlerImpl

soulxu:
Commit Message: network: support hashable address for ConnectionHandlerImpl
Additional Description:
The string output of `Network::Address::Instance` doesn't guarantee it should be the correct output for the void hash key. For example, the envoy internal address' string output is `envoy://{address_id}/{endpoint_id}`. But actually, we only use the `address_id` as the hash key on the listener side.

This PR adds a wrapper `Network::Address::AddressKey` for `InstanceConstSharedPtr`, enabling to use `Network::Address::InstanceConstSharedPtr` as a hash key for absl::Hash. Then we can remove the usage of address string as hash key.

This PR changes the hash maps in `ConnectionHandlerImpl` to use the `Network::Address::AddressKey` as the hash key. the parameter of `ConnectionHandlerImpl::getBalancedHandlerByAddress` is changed for enabling query based on `Network::Address::InstanceConstSharedPtr`.

There are more places that need to replace the address string hash key, those will be done in the separated PRs.

Risk Level: high
Testing: unittests
Docs Changes: n/a
Release Notes: n/a


jmarantz:
@kyessenov can you take a pass? I think this is worth having @mattklein123 do a senior maintainer pass eventually.

@soulxu can you talk, in the description, about the motivation for this? Is this to make things faster by not having to hash the string? Do you have a microbenchmark showing what we'll eventually get after the last phase of this?

soulxu:
> @kyessenov can you take a pass? I think this is worth having @mattklein123 do a senior maintainer pass eventually.
> 
> @soulxu can you talk, in the description, about the motivation for this? Is this to make things faster by not having to hash the string? Do you have a microbenchmark showing what we'll eventually get after the last phase of this?

@jmarantz thanks for the review! that is a good question. Actually, there are several times we mentioned should make the Address object hashable instead of using the string. The most recent one is https://github.com/envoyproxy/envoy/pull/23760#discussion_r1014155485

I think the major reason is the string output of the address object isn't a guarantee of the hash key for the address object. One of example is the internal address' string output is 'evnoy://{address_id}/{endpoint_id}', but actually, we only use the `address_id` as the hash key. https://github.com/envoyproxy/envoy/pull/24198/files#diff-fe538327af200510b78cadc759f2c3af989c822a370d3f8a005b92bc59568aa5L134

I will update this in the description.

soulxu:
> I think the major reason is the string output of the address object isn't a guarantee of the hash key for the address object.(The string output is major for human readable I think) One example is the envoy internal address' string output is `envoy://{address_id}/{endpoint_id}`, but actually, we only use the `address_id` as the hash key. https://github.com/envoyproxy/envoy/pull/24198/files#diff-fe538327af200510b78cadc759f2c3af989c822a370d3f8a005b92bc59568aa5L134

@kyessenov do we use `address_id` + `endpoint_id` for hash the host in the upstream? if yes, that is confusing too, that seems like we can't have a single interface to hash the object since downstream and upstream hash with a different keys.

kyessenov:
@soulxu Sorry I think I messed up something in your comment. Was on a flaky internet and hit refresh at a wrong time.

soulxu:
> @soulxu Sorry I think I messed up something in your comment. Was on a flaky internet and hit refresh at a wrong time.

no worries, thanks for the feedback!

Seems it should be fine, since listener is the only place using address to hash today. Even if other places need a different ways to hash the address in the future, we can define another `AddressKey`.

envoy/network/address.h
```diff
@@ -230,6 +231,39 @@ class Instance {
...
+ * Wrapper for InstanceConstSharedPtr, enable to use InstanceConstSharedPtr
+ * hash key for absl::Hash.
+ */
+struct AddressKey {
+  AddressKey(InstanceConstSharedPtr address) : address_(address) {}
+
+  friend bool operator==(const AddressKey& lhs, const AddressKey& rhs) {
+    // For the envoy internal address, only compare the address id and
+    // ignore the endpoint id.
+    if (lhs.address_->type() == Type::EnvoyInternal &&
```
kyessenov, line 251: I think it would be simpler to also hash endpoint_id. Semantically, this is similar to a "port" on an address, and it's reasonable to expect that:
* upstream specifies an internal "port" to an internal address.
* downstream listens on all "ports" for an internal address.

The same code path should work for non-internal address if we used actual ports, and it's better to avoid exceptions.

Since we don't support individual "endpoint_id" on downstream, we can add a check that denies internal downstream listen address with a non-empty endpoint_id.

envoy/network/address.h
```diff
@@ -230,6 +231,39 @@ class Instance {
    * @return SocketInterface to be used with the address.
    */
   virtual const Network::SocketInterface& socketInterface() const PURE;
+
+  /**
+   * For support the absl::Hash.
+   */
+  virtual void abslHashValue(absl::HashState state) const PURE;
```
kyessenov, line 238: IMHO we should support hashing directly without the indirect `AddressKey`. None of the current code hashes directly, and using structured hashing and equality is strictly better long term.

test/common/network/address_impl_test.cc
```diff
@@ -331,6 +360,26 @@ TEST(PipeInstanceTest, Basic) {
   EXPECT_EQ(nullptr, address.envoyInternalAddress());
 }
 
+TEST(InternalInstanceTest, Hashable) {
+  EXPECT_TRUE(absl::VerifyTypeImplementsAbslHashCorrectly({
```
kyessenov, line 364: You should have a test that calls this on all types of Addresses simultaneously instead of grouping them by type.

kyessenov (commented):
Apologies for dropping the ball on this review. I think this is a good idea but this is a really core code so would definitely want some senior reviewer as well.

cc @envoyproxy/senior-maintainers

envoy/network/address.h
soulxu: This is due to I want to hash the shared pointer of address instance object `InstanceConstSharedPtr `, otherwise I thought it should expensive to store each address instance object copy in the hash map.

envoy/network/address.h
soulxu: > I think it would be simpler to also hash endpoint_id. Semantically, this is similar to a "port" on an address, and it's reasonable to expect that:
> 
> * upstream specifies an internal "port" to an internal address.
> * downstream listens on all "ports" for an internal address.
> 
> The same code path should work for non-internal address if we used actual ports, and it's better to avoid exceptions.

got it, +1 for hash all type addresses similar 

> 
> Since we don't support individual "endpoint_id" on downstream, we can add a check that denies internal downstream listen address with a non-empty endpoint_id.

got it, I can submit a separated PR for this, I guess we need a runtime flag for this, right?

source/common/network/address_impl.cc
```diff
@@ -280,12 +280,17 @@ Ipv6Instance::Ipv6Instance(const std::string& address, const SocketInterface* so
...
+
+Ipv6Instance::Ipv6Instance(const std::string& address, uint32_t port, uint32_t scope_id,
+                           const SocketInterface* sock_interface, bool v6only)
     : InstanceBase(Type::Ip, sockInterfaceOrDefault(sock_interface)) {
   throwOnError(validateProtocolSupported());
   sockaddr_in6 addr_in;
   memset(&addr_in, 0, sizeof(addr_in));
   addr_in.sin6_family = AF_INET6;
   addr_in.sin6_port = htons(port);
+  addr_in.sin6_scope_id = scope_id;
```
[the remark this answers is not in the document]
soulxu: This is for unittest to be able to specify a scope id https://github.com/envoyproxy/envoy/pull/24198/files#diff-eccdb64fd1344e2f120b02d8373331a7c4cf5b6ef4550a560076405d1ad9eb10R254, so I added new construct method here.

test/common/network/address_impl_test.cc
soulxu: got it, will add

soulxu:
> Apologies for dropping the ball on this review. I think this is a good idea but this is a really core code so would definitely want some senior reviewer as well.
> 
> cc @envoyproxy/senior-maintainers

thanks for the review again!

soulxu:
/wait

according to https://github.com/envoyproxy/envoy/pull/24198#discussion_r1040147664, this PR https://github.com/envoyproxy/envoy/pull/24438 is added.

In theory, we should wait for that PR merge first.

envoy/network/address.h
soulxu: @kyessenov gentle ping

envoy/network/address.h
kyessenov: OK, that makes sense.

source/common/network/address_impl.cc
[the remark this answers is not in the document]
kyessenov: Thanks.

source/common/network/address_impl.h
```diff
@@ -312,6 +330,13 @@ class PipeInstance : public InstanceBase {
   }
   absl::string_view addressType() const override { return "default"; }
 
+  void abslHashValue(absl::HashState state) const override {
+    absl::HashState::combine(absl::HashState::combine_contiguous(std::move(state),
+                                                                 pipe_.address_.sun_path,
+                                                                 sizeof(pipe_.address_.sun_path)),
```
kyessenov, line 336: I think this should be `address_length_`?

kyessenov (commented):
Generally, LGTM, apart from minor address_length_ comment.
I think this needs a second set of eyes though.

envoy/network/address.h
kyessenov: I think it's fine to strip endpoint_id when searching for a matching listener if we have to have one big registry. Alternatively, we'd just keep the map separate for internal addresses. I don't recall which way is more natural at this point.

soulxu: got it, Iet me try to strip endpoint id, thanks

source/common/network/address_impl.h
soulxu: address_length_ is only set for abstract namespace. For non abstract namespace, it doesn't set 
https://github.com/envoyproxy/envoy/blob/357bf0016ffa0102ed1fc3c51687eb23b9d2305d/source/common/network/address_impl.cc#L381-L389

if use the `sizeof(pipe_.address_.sun_path)`, it will be safe if the `pipe_.address_.sun_path` set to zero by `memset`. But yea, there are PipeInstance's constructs method accept `sockaddr_un* address` as the parameter. It can guarantee the memory is set as zero.

Maybe I can use address_length_ for abstract namespace. The use `strlen` for non abstract namespace.

soulxu:
> Generally, LGTM, apart from minor address_length_ comment. I think this needs a second set of eyes though.

thanks for the review, sorry for reply the comment partially previously. Since I got covoid19, I will slow down work few days.

source/common/network/address_impl.h
kyessenov: Yeah, I was looking at equality or addressAsString I think, and it's possible that usage is wrong there.

kyessenov:
@soulxu Sorry to hear, hope it's nothing serious. I'll be on and off in the next couple of weeks, too.

soulxu:
> @soulxu Sorry to hear, hope it's nothing serious. I'll be on and off in the next couple of weeks, too.

It is all good here, thanks!

I change ConnectionHandleImpl::findByAddres to generate a temp address without endpoint id. Also change the pipe address hash method to use the actual path size. Would you like to review  again, then I can catch the senior maintainer for next review.

Thanks
Alex

source/extensions/listener_managers/listener_manager/connection_handler_impl.cc
```diff
@@ -290,8 +289,10 @@ void ConnectionHandlerImpl::setListenerRejectFraction(UnitFloat reject_fraction)
 Network::InternalListenerOptRef
 ConnectionHandlerImpl::findByAddress(const Network::Address::InstanceConstSharedPtr& address) {
   ASSERT(address->type() == Network::Address::Type::EnvoyInternal);
-  if (auto listener_it =
-          internal_listener_map_by_address_.find(address->envoyInternalAddress()->addressId());
+  Network::Address::InstanceConstSharedPtr internal_instance =
+      std::make_shared<Network::Address::EnvoyInternalInstance>(
+          address->envoyInternalAddress()->addressId());
```
kyessenov, line 294: random thought, feel free to ignore: a heterogeneous lookup with value address instead of const-shared-ptr address would be slightly faster.

kyessenov (approved):
Thanks for coming back to the PR after the break.

source/common/network/address_impl.h
```diff
@@ -312,6 +330,15 @@ class PipeInstance : public InstanceBase {
   }
   absl::string_view addressType() const override { return "default"; }
 
+  void abslHashValue(absl::HashState state) const override {
+    absl::HashState::combine(
+        absl::HashState::combine_contiguous(std::move(state), pipe_.address_.sun_path,
+                                            pipe_.address_.sun_path[0] == '\0'
```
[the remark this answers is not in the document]
soulxu: yea, `abstract_namespace` is better, let me update.

source/extensions/listener_managers/listener_manager/connection_handler_impl.cc
kyessenov: OK, I don't remember off-hand the magic, but I recall that absl hash maps can accept string_view and string at the same time without a temporary.

envoy/network/address.h
```diff
@@ -230,6 +231,32 @@ class Instance {
...
```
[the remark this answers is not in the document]
mattklein123: > My concern with this though is this may cause different hashing of upstream connections than we currently have and thrashing during rollout if folks expect consistent hashing across instances. cc @htuch @mattklein123 for if this is an actual problem and if so how we would do it safely.

The fact that we are already using absl::Hash for this means that there is no guarantee of stability between restarts (though in practice stability is likely), so it's probably OK to make this change. If we really want stability in this path we should be using one of the stable hashes in here: https://github.com/envoyproxy/envoy/blob/main/source/common/common/hash.h. (Like we do for the consistent hash load balancers for this reason.)

So I guess this is probably OK? Perhaps we should open an issue to migrate this use of hash to a stable hash though?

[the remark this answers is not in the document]
alyssawilk: I'd think absl::Hash should be pretty stable across restarts for the same IP address and backend sets. If we think stability is currently likely and this may change it, how about a runtime guard and release note as a compromise?  We could also then leave the guard false until all the code is guarded, so have reviewable code segments but not have inconsistent hashing within the binary.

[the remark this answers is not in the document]
htuch: Thanks for raising this @alyssawilk. 

I think the PR is mainly affecting how we hash for STL/ABSL data structures, rather than in places like `hash_policy.cc`, where presumably we still keep using string representation. I think this should also hold for the affinity balancers and ring construction. Maybe I'm missing some place where this matters though, so it can't hurt to runtime guard.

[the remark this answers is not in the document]
soulxu: Actually I hash on the InstanceConstSharePtr, that why I added `struct AddressKey` below https://github.com/envoyproxy/envoy/pull/24198/files/97988a05a5450639759221f626aac84a7b9d2a64#diff-131ae3cdf1b8289789a6da638553a5748c9c007d28ff04a3f3c8fcb4f533e58cR245

So I guess any code without explicitly using the `AddressKey`, then won't use this hash code, that means all the existing code won't be changed by this PR?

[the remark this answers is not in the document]
yanavlasov: absl::Hash is not deterministic across process invocations. It uses ASLR as a seed to randomize hash generation. https://github.com/abseil/abseil-cpp/blob/570dc02b0b70bd4fb37efd79021890fa9edc283f/absl/hash/internal/hash.h#L1120

So changing the input into the hash function should not matter with respect to hash stability across process restarts.

envoy/network/address.h
```diff
@@ -230,6 +231,32 @@ class Instance {
    * @return SocketInterface to be used with the address.
    */
   virtual const Network::SocketInterface& socketInterface() const PURE;
+
+  /**
+   * For support the absl::Hash.
```
alyssawilk, line 236: -> For supporting absl::Hash

source/common/network/address_impl.cc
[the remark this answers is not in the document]
alyssawilk: sorry I'm not clear on this either - if there was no way to specify a scope id, why do we need to regression test the scope ID in the hash?

envoy/network/address.h
```diff
@@ -230,6 +231,32 @@ class Instance {
...
+   */
+  virtual void abslHashValue(absl::HashState state) const PURE;
+};
+
+/**
+ * Wrapper for InstanceConstSharedPtr, enable to use InstanceConstSharedPtr
+ * hash key for absl::Hash.
+ */
+struct AddressKey {
+  AddressKey(InstanceConstSharedPtr address) : address_(address) {}
```
alyssawilk, line 246: why do we need the constructor and member variable? it it just for tests because it appears to not be used outside of them.  I'd prefer to avoid it, but if it's unavoidable at least comment it's test only?

envoy/network/address.h
toddmgreer: Instance::abslHashValue's documentation should say something like:
// Overrides must ensure that everything that affects operator== is combined into state, and vice-versa.
// See https://abseil.io/docs/cpp/guides/hash#making-hashable-types.

envoy/network/address.h
soulxu: Not sure I understand your question. The AddressKey is a wrap of `InstanceConstSharedPtr`, I want to store the shared ptr in the key, not the object value that is too expensive, and I can't override the `operator==` for shared ptr. That is why I added the AddressKey. The member variable `address_` is the actual thing as the key. The constructor can implicitly convert `InstanceConstSharedPtr` to the AddressKey, which is more convenient. Let me know if I didn't answer your question.

envoy/network/address.h
soulxu: got it, thanks!

soulxu: > Instance::abslHashValue's documentation should say something like: // Overrides must ensure that everything that affects operator== is combined into state, and vice-versa. // See https://abseil.io/docs/cpp/guides/hash#making-hashable-types.

I need to use `InstanceConstSharedPtr` as the hash key. Not the actual address object. Then I added `struct AddressKey` below, the `operator==` is defined for AddressKey also.

source/common/network/address_impl.h
```diff
@@ -144,6 +144,11 @@ class Ipv4Instance : public InstanceBase {
   // given address if not.
   static absl::Status validateProtocolSupported();
 
+  void abslHashValue(absl::HashState state) const override {
+    absl::HashState::combine(std::move(state), ip_.ipv4_.address_.sin_addr.s_addr,
```
toddmgreer, line 148: The abslHashValue overrides would be more obviously correct if they used the same expressions as ther coresponding operator==. (e.g. ip_.ipv4_.address() and ip_.port())

source/common/network/address_impl.h
```diff
@@ -312,6 +330,14 @@ class PipeInstance : public InstanceBase {
   }
   absl::string_view addressType() const override { return "default"; }
 
+  void abslHashValue(absl::HashState state) const override {
```
toddmgreer, line 333: PipeInstance::operator== is based on asString(). Can you document why we know that equal PipeInstances will have equal HashStates and unequal PipeInstances will have unequal HashStates?

source/common/network/address_impl.h
soulxu: Emm... you are right. It is fine with IPv4 and IPv6, but I have wrong operator== for pipe and internal address, let me fix it, thanks!

source/common/network/address_impl.h
```diff
@@ -358,6 +384,11 @@ class EnvoyInternalInstance : public InstanceBase {
   socklen_t sockAddrLen() const override { return 0; }
   absl::string_view addressType() const override { return "envoy_internal"; }
 
+  void abslHashValue(absl::HashState state) const override {
+    absl::HashState::combine(std::move(state), internal_address_.address_id_,
```
toddmgreer, line 388: Can you document why we know that equal EnvoyInternalInstance will have equal HashStates and unequal EnvoyInternalInstance will have unequal HashStates?

source/common/network/address_impl.h
soulxu: I'm wrong here, I should have a different `operator==` here with different method name here for the hash.

source/common/network/address_impl.h
soulxu: Yea, I will fix that as above said.

mobile/library/common/network/synthetic_address_impl.h
```diff
@@ -55,6 +55,10 @@ class SyntheticAddressImpl : public Instance {
   }
 
 private:
+  void abslHashValue(absl::HashState state) const override {
```
toddmgreer, line 58: Every instance of SyntheticAddressImpl will hash to the same value, even though operator== considers them all unequal. This violates the requirements of absl::Hash and might lead to inefficiencies if these get inserted into absl containers.

I can't tell whether this function will actually be used. If so, it would be safest to follow the rules. Maybe replace `address_` with `this`?

source/common/network/address_impl.h
soulxu: I will add doc.

mobile/library/common/network/synthetic_address_impl.h
soulxu: I wasn't sure this will be used or not. If it won't be used I can put a `not implement` panic here. @alyssawilk may know this.

alyssawilk: please no panics in production code.  I don't think it's is used today but it's the sort of thing that I could imagine causing a runtime crash in the data plane if it became used so I'd encourage doing what Todd suggested and maybe adding an ENVOY_BUG if @jpsim or @Augustyniak are confident it should be unused.

soulxu: Got it, learned that, thanks

envoy/network/address.h
toddmgreer: I don't think `AddressKey` is necessary--absl containers support heterogenous lookup for this sort of use case. See https://abseil.io/tips/144.

IIUC (please confirm), the containers that store `InstanceConstSharedPtr` are fundamentally sets, but we need to be able to compute hashes, compare equality, and do lookups based on a subset of the pointee's state. Using maps instead of sets is one way of handling that, but I think it would be more efficient and cleaner to use sets with heterogenous lookup enabled.

soulxu: Emm...do you mean we can hash on `InstanceConstSharedPtr` directly? Is that means we hash on the memory address? If yes, is that safe?

yanavlasov: What @toddmgreer is suggesting is that you store the `InstanceConstSharedPtr` in the hash map and add custom hash and comparator classes in the `absl::flat_hash_map` template parameters. You can move the == operator and the AbslHashValue to these custom classes.

Moreover you can add methods to the comparator class to compare (as well as the require tag) `Network::Address::InstanceConstSharedPtr` with `Network::Address::Instance` and in this way you would not need to change the `getBalancedHandlerByAddress` method parameter type. You'll be bale to make lookups in the hash map by using the `Network::Address::Instance` type. This is what's called heterogeneous lookup in the ling the Todd have provided.

soulxu: Thanks for explaining those to me again. I did a test, it works, I will update the PR. The sample I wrote is as below:
```
#include <iostream>
#include <string>
#include <memory>
#include <assert.h>
#include "absl/container/flat_hash_map.h"

// Base class
class MyType {
public:
	virtual ~MyType() {}
	virtual bool operator==(const MyType& rhs) const = 0;
	virtual int getType() const = 0;
};

// One sub type
class MySubType1 : public MyType {
public:
	MySubType1(int i) : i_(i) {}
	~MySubType1() override {}
	bool operator==(const MyType& rhs) const override {
		auto sub_rhs = dynamic_cast<const MySubType1&>(rhs);
		return i_ == sub_rhs.i_;
	}

	int getType() const override { return 1; }

	template <typename H>
	friend H AbslHashValue(H h, const MySubType1& m);

	int i_;
};

bool operator==(const MySubType1& lhs, const MySubType1& rhs) {
  return lhs.i_ == rhs.i_;
}

template <typename H>
H AbslHashValue(H h, const MySubType1& m) {
  return H::combine(std::move(h), m.i_);
}

// Another sub type.
class MySubType2 : public MyType {
public:
	MySubType2(std::string s) : s_(s) {}
	~MySubType2() override {}
	bool operator==(const MyType& rhs) const override {
		auto sub_rhs = dynamic_cast<const MySubType2&>(rhs);
		return s_ == sub_rhs.s_;
	}

	int getType() const override { return 2; }

	template <typename H>
	friend H AbslHashValue(H h, const MySubType2& m);

	std::string s_;
};

bool operator==(const MySubType2& lhs, const MySubType2& rhs) {
  return lhs.s_ == rhs.s_;
}

template <typename H>
H AbslHashValue(H h, const MySubType2& m) {
  return H::combine(std::move(h), m.s_);
}

using MyTypeSharedPtr = std::shared_ptr<MyType>;

// The custom hash MyTypeSharedPtr 
struct MyTypeHash {
	using is_transparent = void;
	std::size_t operator()(const MyTypeSharedPtr& s) const noexcept
    {
		if (s->getType() == 1) {
			return absl::Hash<MySubType1>{}(dynamic_cast<const MySubType1&>(*s));
		}

        return absl::Hash<MySubType2>{}(dynamic_cast<const MySubType2&>(*s));
    }
	std::size_t operator()(const MyType& s) const noexcept
    {
        if (s.getType() == 1) {
			return absl::Hash<MySubType1>{}(dynamic_cast<MySubType1 const&>(s));
		}

        return absl::Hash<MySubType2>{}(dynamic_cast<MySubType2 const&>(s));
    }
};

// The custom comparator for MyTypeSharedPtr 
struct MyTypeEqual {
 using is_transparent = void;

 bool operator()(const MyTypeSharedPtr& lhs, const MyTypeSharedPtr& rhs) const
 {
    return *lhs == *rhs;
 }

 bool operator()(const MyTypeSharedPtr& lhs, const MyType& rhs) const {
	return *lhs == rhs;
 }

 bool operator()(const MyType& lhs, const MyTypeSharedPtr& rhs) const {
	return lhs == *rhs;
 }
 bool operator()(const MyType& lhs, const MyType& rhs) const {
	return lhs == rhs;
 }
};

int main(int argc, const char** argv) {
	absl::flat_hash_map<MyTypeSharedPtr, int, MyTypeHash, MyTypeEqual> map;

	std::shared_ptr<MyType> t1 = std::make_shared<MySubType1>(1);
	map.insert({t1, 1});

	MyType& tr1 = *t1;
	auto iter = map.find(tr1);
	assert(iter != map.end());
	assert(iter->second == 1);

	std::shared_ptr<MyType> t2 = std::make_shared<MySubType1>(1);
	map.insert_or_assign(t2, 2);
	auto iter2 = map.find(t2);
	assert(iter2->first.get() != t2.get());
	assert(iter2->second == 2);

	std::shared_ptr<MyType> t3 = std::make_shared<MySubType1>(2);
	map.insert_or_assign(t3, 3);
	auto iter3 = map.find(t3);
	assert(iter3->first.get() == t3.get());
	assert(iter3->second == 3);
}
```

soulxu: @toddmgreer Another suggestion from you is using absl::flat_hash_set instead of absl::flat_hash_map, since the value is `PerAddressActiveListenerDetails`, it is include address also, then I can use the heterogenous lookup to query by the address. Did I understand correctly?

One issue is we have other places use the address to hash also, like https://github.com/envoyproxy/envoy/blob/4fb3eeb1a12b9005133639e96425fef16816c2a4/source/extensions/listener_managers/listener_manager/listener_impl.h#L425
https://github.com/envoyproxy/envoy/blob/4fb3eeb1a12b9005133639e96425fef16816c2a4/source/extensions/listener_managers/listener_manager/listener_impl.h#L510

Seems like I should create custom function and comparator for each of them. Is that something we want?

zuercher:
Is this PR still in progress?

soulxu:
> Is this PR still in progress?

@zuercher, sorry, I recently focus on IoUring support and other internal things. Plan to continue this next quarter, also https://github.com/envoyproxy/envoy/pull/23882 since this PR is ahead refactor for it.

Are you looking for this change? or just cleanup stall PR?

zuercher:
Just checking in since it seemed stalled.

PR #24204, opened 2022-11-24
[TESTING] Ci skip

phlax:
PLEASE IGNORE

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# ARBITRARY DOCS CHANGE
+
 # set SPHINX_SKIP_CONFIG_VALIDATION environment variable to true to skip
 # validation of configuration examples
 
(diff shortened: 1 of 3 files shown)


docs/build.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# ARBITRARY DOCS CHANGE
```
phlax, line 3: cc @alyssawilk - testing with a docs only chnage - ill try mobile after

phlax: hmm, so first problem i see is that `check` stage is required for pr to land - i dont think we can make that conditional - i guess we could supress the individual jobs in the `check` stage

phlax: doesnt seem to have worked anyway - wierd - i will debug

phlax: arrg - it has `--quiet` which will not work - ill PR to fix that

PR #24210, opened 2022-11-25
listener: enable socket_options for multiple addresses

soulxu:
Commit Message: listener: enable socket_options for multiple addresses
Additional Description:
This PR enables specifying socket_options for each additional address. Since when using dual-stack, the IPv4 and IPv6 address have different socket option, then we need a way to specify different socket option for multiple addresses.

notes for implementation:
* To simplify the implementation, the new listener will only clone the old listener's socket when all the addresses' socket_options are equal when updating the listener. This is the same as when updating additional addresses, we only clone the socket when all the addresses are the same.

* For updating socket options for additional addresses, it is also under the runtime guard protection of "envoy.reloadable_features.enable_update_listener_socket_options". This is same as https://github.com/envoyproxy/envoy/pull/23189

Risk Level: high
Testing: unittest and integration test
Docs Changes: API doc
Release Notes: new feature

### DIFF
--- a/source/server/listener_impl.h
+++ b/source/server/listener_impl.h
@@ -318,12 +318,9 @@ public:
   }
   void addSocketFactory(Network::ListenSocketFactoryPtr&& socket_factory);
   void setSocketAndOptions(const Network::SocketSharedPtr& socket);
-  OptRef<const Network::Socket::OptionsSharedPtr> listenSocketOptions(const Network::Address::InstanceConstSharedPtr& address) {
-    auto iter = listen_socket_options_map_.find(address->asString());
-    if (iter != listen_socket_options_map_.end()) {
-      return iter->second;
-    }
-    return absl::nullopt;
+  const Network::Socket::OptionsSharedPtr& listenSocketOptions(uint32_t address_index) {
+    ASSERT(listen_socket_options_list_.size() > address_index);
+    return listen_socket_options_list_[address_index];
   }
   const std::string& versionInfo() const { return version_info_; }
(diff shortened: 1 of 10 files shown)


source/server/listener_impl.h
```diff
@@ -461,9 +466,11 @@ class ListenerImpl final : public Network::ListenerConfig,
   void checkIpv4CompatAddress(const Network::Address::InstanceConstSharedPtr& address,
                               const envoy::config::core::v3::Address& proto_address);
 
-  void addListenSocketOptions(const Network::Socket::OptionsSharedPtr& options) {
-    ensureSocketOptions();
-    Network::Socket::appendOptions(listen_socket_options_, options);
+  void addListenSocketOptions(uint32_t address_index,
+                              const Network::Socket::OptionsSharedPtr& options) {
```
mattklein123, line 470: nit: I think you can avoid passing an index here and actually just pass an already initialized options structure that you drive from the loop inside `buildListenSocketOptions`. It would be nice to not have to reason about the indexes in multiple places.

mattklein123 (changes requested):
Thanks, LGTM with small comments.

/wait

source/server/listener_impl.h
soulxu: If I move line 472 `ensureSocketOptions(address_index);` into the `buildListenSocketOptions`, then this method seems useless. In the end. In the end, it looks like I spread this two lines code into the `buildListenSocketOptions`

mattklein123 (approved):
Thanks!

PR #24212, opened 2022-11-25
ci: Trigger mobile workflows on workflow change

phlax:
If the workflow has changed but is not tested, it might break next time it actually is tested

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -6,9 +6,11 @@ on:
       - main
     paths:
     - 'mobile/**'
+    - '.github/workflows/**'
   pull_request:
     paths:
     - 'mobile/**'
+    - '.github/workflows/**'
 
 jobs:
   androidbuild:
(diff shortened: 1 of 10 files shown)


phlax:
just to flag this is the kind of problem i was referring to with the "`git diff` and bail" pattern

with bazel, you have a hermetic principle and everything sticks to it you can trust your inputs and bazel automatically blows the cache if it detects any change in them

with this type of pattern you are effectively managing the inputs manually which means that in all of the places you want to control program flow according to change you need to capture exactly the correct inputs/changes, and supposing you manage to do that, you then have to maintain them as different parts of the system change

phlax:
a similar example would be where the mobile part of the repo uses tooling from the envoy part of the repo - the tooling might change in the envoy part, which wont be detected as a mobile change and so wont get tested until the mobile part itself changes

(we might check and see that it doesnt pose any problem, but there is not guarantee that it will continue to do so)

alyssawilk (approved):
very good call, thanks!

PR #24221, opened 2022-11-28
ci: Add pipeline change detection

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

part of https://github.com/envoyproxy/envoy/issues/23758

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -25,7 +25,75 @@ variables:
   value: $[or(eq(variables['isMain'], 'true'), eq(variables['isReleaseBranch'], 'true'))]
 
 stages:
+- stage: env
+  #
+  # Note to use the outputs created here - this `stage` must be set as a dependency in the relevant stage
+  # eg:
+  #
+  # stage: mystage
+  # dependsOn: ["env"]
+  #
+  # To use a variable as a condition for a `stage` you can specify it as follows:
+  #
+  # condition: ne(dependencies.env.outputs['repo.changed.mobileOnly'], 'true')
+  #
+  # To use a variable as a condition for a `job` you can specify it as follows:
+  #
+  # job:
+  #   condition: ne(stageDependencies.env.repo.outputs['changed.mobileOnly'], 'true')
+  #
+  # Note the difference in name resolution when used in stages and jobs.
+  #
+  # NB: Avoid using _inclusive_ lists to trigger CI - eg _dont_ only trigger mobile CI if `mobile/` changed.
+  #  Instead use _exclusive_ bools - eg _do_ suppress other CI if only `mobile/` changed
+  #
+  # To use a variable in a `step` you can specify it as follows:
+  #
+  # stage: my_next_stage
+  #   jobs:
+  #     job: another_job
+  #     variables:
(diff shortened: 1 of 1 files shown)


.azure-pipelines/pipelines.yml
```diff
@@ -25,6 +25,73 @@ variables:
   value: $[or(eq(variables['isMain'], 'true'), eq(variables['isReleaseBranch'], 'true'))]
 
 stages:
+- stage: change_detection
```
phlax, line 28: perhaps rename this to `environment` as we can also use it to detect eg VERSION

.azure-pipelines/pipelines.yml
```diff
@@ -25,6 +25,73 @@ variables:
...
+  # dependsOn: ["change_detection"]
+  #
+  # To use a variable as a condition for a `stage` you can specify it as follows:
+  #
+  # condition: ne(dependencies.change_detection.outputs['repo.changed.mobileOnly'], 'true')
+  #
+  # To use a variable as a condition for a `job` you can specify it as follows:
+  #
+  # job:
+  #   condition: eq(stageDependencies.change_detection.repo.outputs['changed.mobileOnly'], 'true')
```
phlax, line 42: ne

alyssawilk:
hah I was about to ask if it was easier for you to pick this up than suggest all the fixes.  Thanks so much for tackling it!  LMK when this is ready as I'd love to get something in this week

phlax:
i have not added any actual exclusions to this PR - just the framework for detecting changes - as when we add any exclusions we need to think carefully about how it will run on `main` or other release branches and on tagged commits/releases

im inclined to not exclude anything when a PR lands - this does risk PRs passing and then failing when they land - but i would rather that than commits passing and then releases not passing and i think we need to ensure every change is tested _somewhere_

phlax:
testing PR is #24204

phlax:
this is probably gtg altho we may need to iterate - testing is hard because the PR introduces change itself

alyssawilk:
I'd be inclined to land with at least one CI we think is affected (i.e. https://github.com/envoyproxy/envoy/pull/24204/files) as I'm not sure landing this as-is is going to tell us anything. WDYT?

phlax:
> WDYT?

adding a change here is probably not going to help as the PR contains change itself

i think land this and then run the test PR a couple of times, and then clarify what should/not be excluded

phlax:
> adding a change here is probably not going to help as the PR contains change itself

hmm, actually thinking about it - yep i see - we need at least one thing to be excluded to test - ill add ...

alyssawilk:
yeah I'll also say now we have this (or almost) I am so excited to say not run all the Envoy tests for docs only changes.  This is going to be pretty awesome :-D

phlax:
k - ive added a condition which hopefully will only run coverage iff either:

- commit is postsubmit (on any branch)
- the PR is neither docs-only nor mobile-only

should work

alyssawilk (approved):
Let's see how it goes :-)

PR #24223, opened 2022-11-28
ci: run DrString against Xcode 14.1

jpsim:
Commit Message: ci: run DrString against Xcode 14.1
Additional Description: Matching the Xcode version used for other `mobile/**` CI jobs.
Risk Level: None
Testing: Previously set on envoyproxy/envoy-mobile
Docs Changes: None
Release Notes: None
Platform Specific Features: Only runs on changes to `mobile/**`

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -48,7 +48,7 @@ jobs:
           submodules: true
       - name: 'Run DrString'
         env:
-          DEVELOPER_DIR: /Applications/Xcode_14.0.app
+          DEVELOPER_DIR: /Applications/Xcode_14.1.app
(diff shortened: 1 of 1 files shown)


alyssawilk (approved):

alyssawilk:
so I'm OK merging as-is, but it might be worth picking up https://github.com/envoyproxy/envoy/pull/24212 just to make sure it runs cleanly.  theoretically it should as it works for downstream E-M but I had enough things break in weird ways t can't hurt to double check.

alyssawilk (approved):
There have been so many changes to the yaml files today it's hard to not be behind ;-)

PR #24197, opened 2022-11-24
deps: Bump `envoy.code.check` -> 0.3.4 (and cleanup yaml)

phlax:
Fix #24195 

Signed-off-by: Ryan Northey <ryan@synca.io>

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #24224, opened 2022-11-28
ci: run SwiftLint on `mobile/**` changes

jpsim:
Commit Message: ci: run SwiftLint on `mobile/**` changes
Additional Description:
Risk Level: Low, CI job, only runs when on `mobile/**` changes
Testing: Local
Docs Changes: None
Release Notes: None
Platform Specific Features: only runs when on `mobile/**` changes
part of https://github.com/envoyproxy/envoy/pull/24224

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -5,10 +5,10 @@ on:
     branches:
       - main
     paths:
-    - 'mobile/**'
+      - 'mobile/**'
   pull_request:
     paths:
-    - 'mobile/**'
+      - 'mobile/**'
 
 jobs:
   formatall:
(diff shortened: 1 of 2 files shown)


.github/workflows/format.yml
```diff
@@ -5,10 +5,10 @@ on:
     branches:
       - main
     paths:
-    - 'mobile/**'
+      - 'mobile/**'
```
phlax, line 9: could you fix the linting here - we should be checking this but currently arent - there is #24197 pending to fix

jpsim:
Confirmed SwiftLint job ran successfully when I forced it to run in https://github.com/envoyproxy/envoy/actions/runs/3565522319/jobs/5990798974

abeyad (approved):

jpsim:
FYI I didn't do this, maybe some automation did? Repokitteh maybe? It's more aggressive than I'd like, I didn't need this PR to ping multiple people 🙃.

image

alyssawilk:
Envoy codeowners results in auto-assigning reviewers.
in the Envoy repo we generally ignore reviewers and only review if a PR is assigned via the assignees field.

abeyad:
> FYI I didn't do this, maybe some automation did? Repokitteh maybe? It's more aggressive than I'd like, I didn't need this PR to ping multiple people 🙃.
> 
> image

My understanding is that one of the github plugins automatically adds whoever is in the OWNERS set of the affected code to the "reviewers" list automatically.  But technically only the people in the "Assignees" list are responsible for the code review.

alyssawilk:
want to just do a main merge so it'll run due to .workflow changes?

PR #24231, opened 2022-11-28
Ignore *.s?? intead of *.sw* to make sure git does not ignore .swift 

RyanTheOptimist:
Ignore *.s?? instead of *.sw* to make sure git does not ignore .swift files

Part of #23758


keith (approved):
lol

alyssawilk (approved):

PR #24232, opened 2022-11-28
process: adding mobile maintainers as reviewers to mobile PRs

alyssawilk:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/repokitteh.star
+++ b/repokitteh.star
@@ -25,6 +25,13 @@ use(
       "auto_assign": True,
     },
     {
+      "owner": "envoyproxy/mobile-maintainers!",
+      "path": "(mobile/)",
+      "label": "mobile",
+      "github_status_label": "changes to Envoy Mobile",
+      "auto_assign": True,
+    },
+    {
       "owner": "envoyproxy/api-watchers",
       "path": "(api/envoy/|docs/root/api-docs/)",
     },


jpsim (approved):
+1 based on my cursory reading of how the other assignments are being done in this file.

repokitteh.star
```diff
@@ -24,6 +24,13 @@ use(
       "github_status_label": "any API change",
       "auto_assign": True,
     },
+    {
+      "owner": "envoyproxy/mobile-maintainers!",
```
phlax, line 28:
```suggestion
      "owner": "envoyproxy/mobile-maintainers",
```

not sure if this is correct 8/

alyssawilk: dependency and api shephards have the ! so I assume it's useful?  API watchers does not but it doesn't auto-assign
cc @itayd

itayd: iirc, and from looking at the code (https://github.com/envoyproxy/envoy/blob/main/ci/repokitteh/modules/ownerscheck.star),  '!' means this requires approval and not only an FYI. @htuch actually rewrote the entire owners thing so he might know better.

alyssawilk: I think that's "allow global approval" because dependency-shepherds has a ! and doesn't  have the approval problem.  Anyway if no one knows and @phlax prefers no ! I'll remove.  We can always fix up later if it's a problem.

phlax: im happy either way - just wasnt sure whether it was correct - reading the code here https://github.com/envoyproxy/envoy/blob/d1be114907c9513b1b45a5eef5b0a88b6188ec7c/ci/repokitteh/modules/ownerscheck.star#L148-L155 its not immediately obvious what it is doing - i can dig through the code and test to clarify

phlax: follow up here #24241

jpsim (approved):
Happy to start with this and re-add the `!` if we determine we need it.

Issue #24183, opened 2022-11-23
Flag --incompatible_config_setting_private_default_visibility will break Envoy in Bazel 7.0


keertk:
Incompatible flag `--incompatible_config_setting_private_default_visibility` will be enabled by default in the next major release (Bazel 7.0), thus breaking Envoy. Please migrate to fix this and unblock the flip of this flag.

The flag is documented here: https://github.com/bazelbuild/bazel/issues/12933.

Please check the following CI builds for build and test results:

- [Ubuntu 20.04, OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1341#0184a207-fe4f-4610-8136-0e86d2201bef)
- [Ubuntu 20.04, OpenJDK 11](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1341#0184a207-fe4c-405a-a54f-05dbff615c64)

Never heard of incompatible flags before? We have [documentation](https://docs.bazel.build/versions/master/backward-compatibility.html) that explains everything.
If you have any questions, please file an issue in https://github.com/bazelbuild/continuous-integration.

wbpcode:
Hi, @keith could you take a look for this when you have free time? Thanks.

keith:
WIP here https://github.com/envoyproxy/envoy/pull/24237

meteorcloudy:
The fix didn't get into downstream because https://buildkite.com/bazel/envoy/builds/2118#01849d85-5a46-4688-9dc4-2e30d6659b1b is failing and prevented the update of last green commit for Envoy.

keith:
https://github.com/envoyproxy/envoy/pull/24270

keith:
looking at the build failure, interesting that it's hit there but not on our CI, maybe a compiler version difference

keith:
the actual build failure is https://github.com/envoyproxy/envoy/issues/23891

botengyao:
We can keep track of this issue here: https://github.com/google/tcmalloc/issues/148

keith:
bumping the clang version for bazel CI here https://github.com/bazelbuild/continuous-integration/pull/1502

keith:
fixed by https://github.com/envoyproxy/envoy/pull/24296

PR #24237, opened 2022-11-29
bazel: fix config_setting visibility issues

keith:
This pulls in a googleurl patch that's waiting on upstream, and updates rules_go for compatibility with this flag. This also flips it early so we don't regress before it's flipped upstream.

Fixes https://github.com/envoyproxy/envoy/issues/24183


keith:
thanks! was just a draft in case of more surprises

.bazelrc
```diff
@@ -31,6 +31,9 @@ build --action_env=PATH --host_action_env=PATH
 build --enable_platform_specific_config
 build --test_summary=terse
 
+# TODO(keith): Remove once this is the default
+build --incompatible_config_setting_private_default_visibility
```
meteorcloudy, line 35: You probably have to enable `--incompatible_enforce_config_setting_visibility` as well to make sure this actually take effect.

keith: https://github.com/envoyproxy/envoy/pull/24270

PR #24243, opened 2022-11-29
Reduce Route/VirtualHost memory utilization by avoiding RateLimitPolicy instances when not needed

yurykats:
Reduce Route/VirtualHost memory utilization by avoiding RateLimitPolicy instances when not needed.

Currently an instance of RateLimitPolicyImpl is created in each vhost and route, even if no rate limiting is configured. This PR switches instances to unique_ptr and only creates RateLimitPolicyImpl objects when actually configured. Otherwise a singletong default policy is used.

Similar change to Retry, Hedge and InternalRedirect policies was made in PR #24182.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


adisuissa (approved):
LGTM, thanks!

zuercher (approved):
thanks!

PR #24249, opened 2022-11-29
build: Set -gdwarf-4 for clang and gcc

ggreenway:

Commit Message: This makes the clang debug information be compatible with the gdb in the envoyproxy build container.
Additional Description:
Risk Level: Very low
Testing: Can use gdb against tests in the build container and get expected debug info
Docs Changes: none
Release Notes: none
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/bazel/envoy_internal.bzl
+++ b/bazel/envoy_internal.bzl
@@ -51,9 +51,12 @@ def envoy_copts(repository, test = False):
                # debugging info detailing some 1600 test binaries would be wasteful.
                # targets listed in order from generic to increasing specificity.
                # Bazel adds an implicit -DNDEBUG for opt targets.
-               repository + "//bazel:opt_build": [] if test else ["-ggdb3"],
+               #
+               # The GDB currently in the CI build container chokes on dwarf-5, so
+               # force dwarf-4.
+               repository + "//bazel:opt_build": [] if test else ["-ggdb3", "-gdwarf-4"],
                repository + "//bazel:fastbuild_build": [],
-               repository + "//bazel:dbg_build": ["-ggdb3"],
+               repository + "//bazel:dbg_build": ["-ggdb3", "-gdwarf-4"],
                repository + "//bazel:windows_opt_build": [] if test else ["-Z7"],
                repository + "//bazel:windows_fastbuild_build": [],
                repository + "//bazel:windows_dbg_build": [],
--- a/.bazelrc
+++ b/.bazelrc
@@ -65,7 +65,6 @@ build:sanitizer --test_tag_filters=-no_san
 build:clang --action_env=BAZEL_COMPILER=clang
 build:clang --action_env=CC=clang --action_env=CXX=clang++
 build:clang --linkopt=-fuse-ld=lld
-build:clang --copt=-gdwarf-4
 
 # Flags for Clang + PCH
 build:clang-pch --spawn_strategy=local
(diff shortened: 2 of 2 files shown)


ggreenway:
References that show the types of errors, and where I got the source for the fix: 

https://github.com/opencollab/llvm-toolchain-integration-test-suite/commit/5d3f1d29e1ced6478d12d51462a54f36fd2e6677
https://github.com/opencollab/llvm-toolchain-integration-test-suite/issues/74

keith (commented):
it looks like this flag enables `-g`, so i'm not sure we can add it globally?

ggreenway:
Oh that's a good point. Is there an easy way to make this apply to clang-debug builds?

keith:
maybe we can get it in here?

https://github.com/envoyproxy/envoy/blob/a479ec095f514268b2506e0430cd38b175f96872/bazel/envoy_internal.bzl#L54-L56

ggreenway:
`repository + "//bazel:opt_build": [] if test else ["-ggdb3"],` isn't that already setting `-g` for opt builds? Is it just `fastbuild` that we want to not have `-g`?

keith (approved):
i think this does what we want. reading through clang at least this shouldn't affect -ggdb3 so both should combine correctly

bazel/envoy_internal.bzl
```diff
@@ -51,9 +51,12 @@ def envoy_copts(repository, test = False):
...
                # targets listed in order from generic to increasing specificity.
                # Bazel adds an implicit -DNDEBUG for opt targets.
-               repository + "//bazel:opt_build": [] if test else ["-ggdb3"],
+               #
+               # The GDB currently in the CI build container chokes on dwarf-5, so
+               # force dwarf-4.
+               repository + "//bazel:opt_build": [] if test else ["-ggdb3", "-gdwarf-4"],
                repository + "//bazel:fastbuild_build": [],
-               repository + "//bazel:dbg_build": ["-ggdb3"],
+               repository + "//bazel:dbg_build": ["-ggdb3", "-gdwarf-4"],
```
lizan, line 59: For debug build, it may still produce some compatibility issue as this flag isn't passed to dependencies.

lizan (commented):
Shall we just upgrade GDB version in build container?

ggreenway:
> Shall we just upgrade GDB version in build container?

How easy is that to do? Right now we are using the ubuntu-provided version, and there isn't a newer one available for the ubuntu version we're on.

bazel/envoy_internal.bzl
ggreenway: Oh, that explains why I still get some errors I think. In practice this change at least let me do normal debugging in envoy code and tests, so it seems like a big improvement over gdb being completely broken.

lizan:
@ggreenway I found why, when we upgrade base image to Ubuntu-20.04 https://github.com/envoyproxy/envoy-build-tools/pull/177 , we stopped using toolchain-r PPA, which contains GDB 10.2, before that PR build container has GDB 10.2 but the recent image is GDB 9.2. We should definitely bring 10.2 back

ggreenway:
> @ggreenway I found why, when we upgrade base image to Ubuntu-20.04 [envoyproxy/envoy-build-tools#177](https://github.com/envoyproxy/envoy-build-tools/pull/177) , we stopped using toolchain-r PPA, which contains GDB 10.2, before that PR build container has GDB 10.2 but the recent image is GDB 9.2. We should definitely bring 10.2 back

Ok, that seems like a much better solution. I'll close this PR.

Issue #24156, opened 2022-11-22
CI github actions are not associated with repokitteh `/retest`


tyxia:
*Title*: *The new eng-flow work is not associated with repokitteh retest*

*Description*:

Envoy CD/CI has new eng-flow tests like `python_test` or `objc_helloworld` but they are flaky sometimes and could be failed due to test infra issue. (e.g. https://github.com/envoyproxy/envoy/actions/runs/3525319287/jobs/5911893897)

Re-running the test can fix them sometimes (at least both of failures are fixed by re-run), but unfortunately it is not associated with repokitteh retest that the re-run can be triggered by `/retest` and there is no `rerun` job button for people who are not maintainers. https://screenshot.googleplex.com/B2fsiGJSZX8EZdj

Ideally, people should be able to rerun those eng-flow tests via `/retest` comment like normal envoy presubmits

tyxia:
cc @phlax 

Just FYI, I am not sure if it is something that you are in charge of.

phlax:
i think we need this - tracking through the github actions is a PITA

ive changed the title to make it clearer the issue - altho im now wondering if `/retest` works with any of the github actions

@itayd is this something we can address?

itayd:
If there is an endpoint in github's API that allows to retry an action that shouldn't be an issue. I guess this might be it? https://docs.github.com/en/rest/actions/workflow-runs#re-run-a-job-from-a-workflow-run

jpsim:
There isn't a single endpoint, no. Here's how we handle this for the Envoy Mobile repo: https://github.com/jpsim/retest/blob/c158dec0a7f67cb85f8367468dc8a9a75308bb7f/retest.ts#L34-L64

jpsim:
`/retest` now works for GitHub Actions as of #24251 

![9BFA06BB-0571-4EDC-81FB-7E32C8393D4D](https://user-images.githubusercontent.com/474794/204816925-5d3de352-f741-41cf-8a50-0ac88fcb8b44.jpeg)

However it would still be better to move it to Repokitteh which would probably make it kick in faster, since the existing action has a 10-20s delay to spin up the GitHub Actions job that processes the comment.

phlax:
should we close this ticket @jpsim ?

jpsim:
Sure let's close this and we can re-open a new issue if moving the functionality to Repokitteh ever becomes a priority.

jpsim:
[...]

PR #24251, opened 2022-11-29
workflows: copying envoy mobile command workflow

alyssawilk:
Allow retest for E-M workflows

Part of https://github.com/envoyproxy/envoy/issues/23758 workaround until we have https://github.com/envoyproxy/envoy/issues/24156

### DIFF
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -1,0 +1,13 @@
+name: commands
+on:
+  issue_comment:
+    types: [created]
+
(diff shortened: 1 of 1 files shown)


jpsim:
Unfortunately, GitHub Actions that operate on comments run the main branch version of the action, so we won't be able to verify that this works until the PR is merged into `main`.

jpsim (approved):
This _should_ work. We can test once it's merged on `main`

PR #24257, opened 2022-11-30
upstream: not parse bootstrap upstream config when cluster upstream config specified

soulxu:
Commit Message: upstream: not parse bootstrap upstream config when cluster upstream config specified
Additional Description:

Envoy will reject when specifying an IPv6 address in bootstrap binding config and an IPv4 address in cluster binding config, this is due to both bootstrap and cluster binding config parsed. This config should be accepted, and the bootstrap binding config should be ignored when choice upstream bind config for the endpoint.

Risk Level: low
Testing: unittest
Docs Changes: n/a
Release Notes: bug fixes


mattklein123 (approved):
Thanks LGTM

PR #24258, opened 2022-11-30
route: support route list for matcher action

### LINKED ISSUE #24238: Support route list for route matcher action
*Title*: Support route list in route matcher action

*Description*:

Now only a single `Route` can be used in router matcher as this code shows: https://github.com/envoyproxy/envoy/blob/4e40d6ad5f98cd3c86f1398bb8990b1d88bdfe92/source/common/router/config_impl.cc#L1880-L1883

But we may need a list of routes under a branch of the matcher tree. `matcher_list` is similar, but not flexible like `RouteMatch `. For example, `RouteMatch ` supports `uri_template`, `headers` and so on.

I wrote an implementation for this: https://github.com/envoyproxy/envoy/compare/main...tony612:route-match-action?expand=1#diff-c1e4a3798005eb4f69d0ea87905f7117810f3c2ef2fd3e06df2d75f3724b5fd7R228

With this new code, we can write config like below, which will be easier than config with only `v3.Route` as matcher's action.

```
route_config:
  virtual_hosts:
  - name: default
    domains: ["*"]
    matcher:
      matcher_tree:
        input:
          name: req-headers
          typed_config:
            "@type": type.googleapis.com/envoy.type.matcher.v3.HttpRequestHeaderMatchInput
            header_name: :path
        prefix_match_map:
          map:
            "/service-1/":
              action:
                name: route
                typed_config:
                  "@type": type.googleapis.com/envoy.config.route.v3.RouteMatchAction
                  routes:
                  - match:
                      path_match_policy:
                        name: envoy.path.match.uri_template.uri_template_matcher
                        typed_config:
                          "@type": type.googleapis.com/envoy.extensions.path.match.uri_template.v3.UriTemplateMatchConfig
                          path_template: "/service-1/foo/{country}/{lang}"
                    route:
                      cluster: svc1
                      path_rewrite_policy:
                        name: envoy.path.rewrite.uri_template.uri_template_rewriter
                        typed_config:
                          "@type": type.googleapis.com/envoy.extensions.path.rewrite.uri_template.v3.UriTemplateRewriteConfig
                          path_template_rewrite: "/foo/{lang}/{country}"
                  - match:
                      headers:
                      - name: "x-custom"
                        exact_match: "foo"
                    route:
                      cluster: svc1
                      timeout: 5s
```

In my implementation, I added a new message `RouteMatchAction`, which is a wrapper of the route list(like VirtualHost but simpler) and maybe more fields can be added in the future.

If this looks good, I can add tests and send a PR.

tony612:
Commit Message: route: support route list for matcher action
Additional Description:
Risk Level: Low
Testing: unit test
Docs Changes:
Release Notes: 
Platform Specific Features:

Fixes #24238


api/envoy/config/route/v3/route_components.proto
```diff
@@ -224,6 +224,13 @@ message FilterAction {
   google.protobuf.Any action = 1;
 }
 
+// This can be used in route matcher :ref:`VirtualHost.matcher<envoy_v3_api_field_config.route.v3.VirtualHost.matcher>`.
+// When the matcher matches, routes will be matched and run.
+message RouteMatchAction {
+  // The list of routes that will be matched and run, in order. The first route that matches will be used.
+  repeated Route routes = 1;
+}
```
mattklein123, line 232: Please add a release note, also add a documentation example for this wherever we document how to use the matching system.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -224,6 +224,13 @@ message FilterAction {
   google.protobuf.Any action = 1;
 }
 
+// This can be used in route matcher :ref:`VirtualHost.matcher<envoy_v3_api_field_config.route.v3.VirtualHost.matcher>`.
+// When the matcher matches, routes will be matched and run.
+message RouteMatchAction {
```
mattklein123, line 229: I would probably call this RouteListAction?

mattklein123 (changes requested):
Thanks LGTM with a small comment.

/wait

mattklein123:
Please do not force push as it makes reviews more complicated. Just merge main and add commits. Thanks.

docs/root/intro/arch_overview/advanced/matching/matching_api.rst
```diff
@@ -179,8 +179,8 @@ HTTP Routing Integration
 ########################
 
 The matching API can be used with HTTP routing, by specifying a match tree as part of the virtual host
-and specifying a Route as the resulting action. See examples in the above sections for how the match
-tree can be configured.
+and specifying a Route or RouteListAction as the resulting action. See :ref:`the examples
```
mattklein123, line 182: nit: please ref link Route and RouteListAction. Though now that I'm looking at this perhaps just s/Route/RouteList to make it consistent? Sorry for the churn.

mattklein123 (changes requested):
Thanks looks good just a small comment, sorry for the back and forth.

/wait

api/envoy/config/route/v3/route_components.proto
```diff
@@ -224,6 +224,13 @@ message FilterAction {
   google.protobuf.Any action = 1;
 }
 
+// This can be used in route matcher :ref:`VirtualHost.matcher <envoy_v3_api_field_config.route.v3.VirtualHost.matcher>`.
+// When the matcher matches, routes will be matched and run.
+message RouteListAction {
```
mattklein123, line 229: Sorry maybe you missed my comment before. WDYT about renaming RouteListAction -> RouteList so it more naturally matches "Route"? As long as we are adding a new thing I think that is better?

/wait

tony612: Ah, I misunderstand the meaning of the second half of the comment. Yeah, this makes sense. I already updated it. Please take another look. Thank you!

mattklein123 (approved):
Thanks!

PR #24266, opened 2022-11-30
mobile: import envoy-mobile commits (2022-11-30)

jpsim:
Imports the following commits from the Envoy Mobile repo:

* https://github.com/envoyproxy/envoy-mobile/pull/2712
* https://github.com/envoyproxy/envoy-mobile/pull/2709


alyssawilk (approved):
I'm fine with this but should we just wait until y'all are ready to switch source of truth and just do one final one?

jpsim:
> I'm fine with this but should we just wait until y'all are ready to switch source of truth and just do one final one?

We (Lyft) are planning to switch our apps to point to this version of Envoy Mobile today, but to do that we want to pull in some changes that landed yesterday.

PR #24267, opened 2022-11-30
Refactor: Filter Manager

KBaichoo:
Commit Message: Rename overloaded term (dual filter), Remove unnecessary duplicated bit in base class.
Additional Description: N/A
Risk Level: low
Testing: ran filter tests
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/filter_manager.cc
+++ b/source/common/http/filter_manager.cc
@@ -535,7 +535,7 @@ void FilterManager::decodeHeaders(ActiveStreamDecoderFilter* filter, RequestHead
     ENVOY_STREAM_LOG(trace, "decode headers called: filter={} status={}", *this,
                      (*entry)->filter_context_.config_name, static_cast<uint64_t>(status));
 
-    (*entry)->decode_headers_called_ = true;
+    (*entry)->processed_headers_ = true;
 
     const auto continue_iteration = (*entry)->commonHandleAfterHeadersCallback(status, end_stream);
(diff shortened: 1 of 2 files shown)


source/common/http/filter_manager.cc
```diff
@@ -1120,7 +1120,7 @@ void FilterManager::encodeMetadata(ActiveStreamEncoderFilter* filter,
     // If the filter pointed by entry hasn't returned from encodeHeaders, stores newly added
     // metadata in case encodeHeaders returns StopAllIteration. The latter can happen when headers
     // callbacks generate new metadata.
-    if (!(*entry)->encode_headers_called_ || (*entry)->stoppedAll()) {
+    if (!(*entry)->processed_headers_ || (*entry)->stoppedAll()) {
```
kyessenov, line 1123: Sorry, can you explain why this is safe? Can `decodeHeaders, encodeMetadata` happen in a sequence for dual filters?

KBaichoo: dual filter -> encoder-decoder filters. 

The `processed_headers_` bit is on the wrapper for a filters. So an encoder-decoder filter would have both a encoder-wrapper and a decoder-wrapper which would each have their own seperate `processed_headers_` bit.

kyessenov: Thanks.

kyessenov (approved):

PR #24277, opened 2022-11-30
http_11_proxy: fixing a bug with buggy proxies

alyssawilk:
By the HTTP/1.1 spec, all headers are supposed to end with CRLF (\r\n)
Turns out the HTTP/1.1 spec also call out some broken legacy servers only return LF (\n")
Removing a performance optimization that was breaking on proxies with this bug.

Risk Level: low
Testing: new unit test
Docs Changes: n/a
Release Notes: not included

### DIFF
--- a/source/extensions/transport_sockets/http_11_proxy/connect.cc
+++ b/source/extensions/transport_sockets/http_11_proxy/connect.cc
@@ -15,7 +15,8 @@ namespace Extensions {
 namespace TransportSockets {
 namespace Http11Connect {
 
-bool UpstreamHttp11ConnectSocket::isValidConnectResponse(absl::string_view data, bool& error, unsigned long& len) {
+bool UpstreamHttp11ConnectSocket::isValidConnectResponse(absl::string_view data, bool& error,
+                                                         unsigned long& len) {
   SelfContainedParser parser;
 
   len = parser.parser().execute(data.data(), data.length());
@@ -83,8 +84,8 @@ Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer) 
       return {Network::PostIoAction::Close, 0, false};
     }
     if (!is_valid_connect_response) {
-      ENVOY_CONN_LOG(trace, "Incomplete CONECT header: {} bytes received",
-                     callbacks_->connection(), peek_data.size());
+      ENVOY_CONN_LOG(trace, "Incomplete CONECT header: {} bytes received", callbacks_->connection(),
+                     peek_data.size());
       return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
     }
 
@@ -95,7 +96,8 @@ Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer) 
     }
     buffer.drain(len);
 
-    ENVOY_CONN_LOG(trace, "Successfully stripped {} bytes of CONNECT header", callbacks_->connection(), len);
+    ENVOY_CONN_LOG(trace, "Successfully stripped {} bytes of CONNECT header",
+                   callbacks_->connection(), len);
     need_to_strip_connect_response_ = false;
   }
   return transport_socket_->doRead(buffer);
--- a/source/extensions/transport_sockets/http_11_proxy/connect.cc
+++ b/source/extensions/transport_sockets/http_11_proxy/connect.cc
@@ -15,15 +15,13 @@ namespace Extensions {
 namespace TransportSockets {
 namespace Http11Connect {
 
-bool UpstreamHttp11ConnectSocket::isValidConnectResponse(absl::string_view data, bool& error,
-                                                         unsigned long& len) {
+bool UpstreamHttp11ConnectSocket::isValidConnectResponse(absl::string_view data,
+                                                         bool& headers_complete,
+                                                         size_t& bytes_processed) {
   SelfContainedParser parser;
 
-  len = parser.parser().execute(data.data(), data.length());
-  if (parser.parser().getStatus() == Http::Http1::ParserStatus::Error ||
-      parser.parser().statusCode() != 200) {
-    error = true;
-  }
+  bytes_processed = parser.parser().execute(data.data(), data.length());
+  headers_complete = parser.headersComplete();
 
   return parser.parser().getStatus() != Http::Http1::ParserStatus::Error &&
          parser.headersComplete() && parser.parser().statusCode() == 200;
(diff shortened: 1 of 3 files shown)


alyssawilk:
cc @Augustyniak

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -67,27 +68,34 @@ Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer)
...
+
+    if (error) {
+      ENVOY_CONN_LOG(trace, "Response does not appear to be a successful CONNECT upgrade",
+                     callbacks_->connection());
+      return {Network::PostIoAction::Close, 0, false};
     }
-    result = callbacks_->ioHandle().read(buffer, index + 4);
-    if (!result.ok() || result.return_value_ != index + 4) {
-      ENVOY_CONN_LOG(trace, "failed to drain CONNECT header", callbacks_->connection());
+    if (!is_valid_connect_response && len == 200) {
```
RyanTheOptimist, line 81: Should this be MAX_RESPONSE_HEADER_SIZE instead of 200? (I thought 200 meant the status code initially) For that matter, should it be >= MAX_RESPONSE_HEADER_SIZE? (I confess I'm not quite sure what the

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -15,14 +15,15 @@ namespace Extensions {
 namespace TransportSockets {
 namespace Http11Connect {
 
-bool UpstreamHttp11ConnectSocket::isValidConnectResponse(Buffer::Instance& buffer) {
+bool UpstreamHttp11ConnectSocket::isValidConnectResponse(absl::string_view data, bool& error, unsigned long& len) {
```
RyanTheOptimist: It looks like Parse::execute() returns a size_t so we should probably use that here. (Though that method should probably be modified to return uint64_t as per current guidance, but that's way outside the scope of this PR.

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -15,14 +15,15 @@ namespace Extensions {
...
-    buffer.drain(parsed);
+
+  len = parser.parser().execute(data.data(), data.length());
+  if (parser.parser().getStatus() == Http::Http1::ParserStatus::Error ||
+      parser.parser().statusCode() != 200) {
+    error = true;
   }
+
   return parser.parser().getStatus() != Http::Http1::ParserStatus::Error &&
          parser.headersComplete() && parser.parser().statusCode() == 200;
```
RyanTheOptimist, line 29: It looks like we're repeating this same expression twice and I'm not sure I understand the difference between the return value and the `error` outparam. Could we perhaps eliminate `error`?

edit:
Oh, I see. The difference is headersComplete(). I see. Ok, how 'bout we do this:

```
bool UpstreamHttp11ConnectSocket::isValidConnectResponse(absl::string_view data, bool& headers_complete, unsigned long& header_bytes_processed) {
  SelfContainedParser parser;

  header_bytes_processed = parser.parser().execute(data.data(), data.length());
  headers_complete = parser.headersComplete();
  return parser.parser().getStatus() != Http::Http1::ParserStatus::Error &&
           parser.headersComplete() && parser.parser().statusCode() == 200;
}
```

Then below we can do something like:
```
if (!is_valid_connect_response) {
  if (headers_complete) {
    // invalid response 
  } else if (bytes_processed > x) {
    // failed to receive response quickly enought
  }
  // incomplete response
}
```

WDYT?

source/extensions/transport_sockets/http_11_proxy/connect.cc
RyanTheOptimist: nit: Consider renaming `len` to `bytes_parsed` or some such for enhanced clarity.

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -67,27 +68,34 @@ Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer)
...
     absl::string_view peek_data(peek_buf, result.return_value_);
-    size_t index = peek_data.find("\r\n\r\n");
-    if (index == absl::string_view::npos) {
-      if (result.return_value_ == MAX_RESPONSE_HEADER_SIZE) {
-        ENVOY_CONN_LOG(trace, "failed to receive CONNECT headers within {} bytes",
-                       callbacks_->connection(), MAX_RESPONSE_HEADER_SIZE);
-        return {Network::PostIoAction::Close, 0, false};
-      }
-      return Network::IoResult{Network::PostIoAction::KeepOpen, 0, false};
+    unsigned long len = 0;
```
RyanTheOptimist, line 72: nit: bytes_processed, I think. (or header_bytes processed, perhaps)

source/extensions/transport_sockets/http_11_proxy/connect.h
```diff
@@ -19,7 +19,7 @@ namespace Http11Connect {
 class UpstreamHttp11ConnectSocket : public TransportSockets::PassthroughSocket,
                                     public Logger::Loggable<Logger::Id::connection> {
 public:
-  static bool isValidConnectResponse(Buffer::Instance& buffer);
+  static bool isValidConnectResponse(absl::string_view data, bool& error, unsigned long& len);
```
RyanTheOptimist, line 22: nit: Please comment the meaning of the parameters vs the return value.

test/extensions/transport_sockets/http_11_proxy/connect_test.cc
```diff
@@ -386,25 +382,32 @@ TEST_F(SocketFactoryTest, CreateSocketReturnsNullWhenInnerFactoryReturnsNull) {
 }
 
 TEST(ParseTest, TestValidResponse) {
+  unsigned long len;
```
RyanTheOptimist, line 385: Do we have test coverage of the incomplete case?
Can we add tests for all the termination variants? \r\n\r\n \r\n\n \n\r\n \n\n ?
It might also be a good idea to add a test in which there are response headers in addition to the response firstline, just to make sure we've got this nailed.

RyanTheOptimist (commented):
Why can't our standards be simple?

Augustyniak:
@alyssawilk I have confirmed that this patch makes PAC file to work on my Samsung device - awesome!

source/extensions/transport_sockets/http_11_proxy/connect.cc
```diff
@@ -67,27 +67,36 @@ Network::IoResult UpstreamHttp11ConnectSocket::doRead(Buffer::Instance& buffer)
...
+    bool is_valid_connect_response =
+        isValidConnectResponse(peek_data, headers_complete, bytes_processed);
+
+    if (!headers_complete) {
+      if (peek_data.size() == MAX_RESPONSE_HEADER_SIZE) {
         ENVOY_CONN_LOG(trace, "failed to receive CONNECT headers within {} bytes",
                        callbacks_->connection(), MAX_RESPONSE_HEADER_SIZE);
         return {Network::PostIoAction::Close, 0, false};
       }
+      ENVOY_CONN_LOG(trace, "Incomplete CONECT header: {} bytes received", callbacks_->connection(),
```
Augustyniak, line 81: CONECT -> CONNECT

RyanTheOptimist (approved):
LGTM!

PR #24280, opened 2022-11-30
Filter Manager: Abort 1xxHeader Filter Chain

KBaichoo:
Commit Message: Have the 1xx filter chain abort if a send local reply occurs during iteration.
Additional Description: This brings it inline with other encode* calls.
Risk Level: low
Testing: added integration test
Docs Changes: n/a
Release Notes: included
Platform Specific Features:


source/common/http/filter_manager.h
```diff
@@ -1005,6 +1005,10 @@ class FilterManager : public ScopeTrackedObject,
       // Used to indicate that we're processing the final [En|De]codeData frame,
       // i.e. end_stream = true
       static constexpr uint32_t LastDataFrame = 0x80;
+
+      // Masks for filter call state.
+      static constexpr uint32_t IsDecodingMask = DecodeHeaders | DecodeData | DecodeTrailers;
+      static constexpr uint32_t IsEncodingMask = EncodeHeaders | Encode1xxHeaders | EncodeData | EncodeTrailers;
```
RyanTheOptimist, line 1011: Nice cleanup!

test/integration/multiplexed_integration_test.cc
```diff
@@ -1463,6 +1463,35 @@ TEST_P(MultiplexedIntegrationTest, EmptyTrailers) {
...
+                                                                        {":scheme", "http"},
+                                                                        {":authority", "host"},
+                                                                        {"expect", "100-continue"}},
+                                         10);
+
+  waitForNextUpstreamRequest();
+  // This will trigger local reply response.
+  upstream_request_->encode1xxHeaders(Http::TestResponseHeaderMapImpl{{":status", "100"}});
+  ASSERT_TRUE(response->waitForEndStream());
+  ASSERT_TRUE(response->complete());
```
RyanTheOptimist, line 1492: To confirm, this test fails w/o the changes in this PR, right?

RyanTheOptimist (approved):
LGTM, assuming I understand the test behavior correctly.

test/integration/multiplexed_integration_test.cc
KBaichoo: Yes, we'd hit an assert and fail without the changes.

RyanTheOptimist (dismissed):

PR #24289, opened 2022-12-01
mobile: import envoy-mobile commits (2022-12-01)

jpsim:
Imports the following commits from the Envoy Mobile repo:

- https://github.com/envoyproxy/envoy-mobile/pull/2704
- https://github.com/envoyproxy/envoy-mobile/pull/2713


jpsim:
> though again if you don't need these specific changes I'd be very mildly inclined to wait until we flip source of truth.

Lyft now points to Envoy as its source of truth, so I'd like to keep Envoy up to date with EM changes daily for as long as changes land in EM.

PR #24294, opened 2022-12-01
envoy: creating a stripped main base without validation server

alyssawilk:
Splitting main common base into a third class, to allow for a main base without the validation server

Risk Level: Medium (added fast fail for main common base validate to mittigate)
Testing: updated tests
Docs Changes: n/a
Release Notes: did not add


alyssawilk:
@abeyad this one's still a draft so please do ignore for now

jpsim:
I can confirm that this removes some stuff from the iOS framework:

```console
# Before
$ size Envoy.framework/Envoy
508M  Envoy.framework/Envoy
$ nm Envoy.framework/Envoy | grep Server14validateConfig
                 U __ZN5Envoy6Server14validateConfigERKNS0_7OptionsERKNSt3__110shared_ptrIKNS_7Network7Address8InstanceEEERNS0_16ComponentFactoryERNS_6Thread13ThreadFactoryERNS_10Filesystem8InstanceE
0000000000005548 T __ZN5Envoy6Server14validateConfigERKNS0_7OptionsERKNSt3__110shared_ptrIKNS_7Network7Address8InstanceEEERNS0_16ComponentFactoryERNS_6Thread13ThreadFactoryERNS_10Filesystem8InstanceE
00000000000c6128 b __ZZN5Envoy6Server14validateConfigERKNS0_7OptionsERKNSt3__110shared_ptrIKNS_7Network7Address8InstanceEEERNS0_16ComponentFactoryERNS_6Thread13ThreadFactoryERNS_10Filesystem8InstanceEE7flogger
000000000008c6d8 s l___func__._ZN5Envoy6Server14validateConfigERKNS0_7OptionsERKNSt3__110shared_ptrIKNS_7Network7Address8InstanceEEERNS0_16ComponentFactoryERNS_6Thread13ThreadFactoryERNS_10Filesystem8InstanceE
# After
$ size Envoy.framework/Envoy 
504M  Envoy.framework/Envoy
$ nm Envoy.framework/Envoy | grep validateConfig
```

I can do a more thorough review when it's ready.

jmarantz:
Per IM, this requires some scrutiny of how we use the MainCommon/MainCommonBase APIs internally to ensure we have a path to being successful in the new model.

source/exe/main_common_base.h
```diff
@@ -0,0 +1,101 @@
...
+
+class ProdComponentFactory : public Server::ComponentFactory {
+public:
+  // Server::DrainManagerFactory
+  Server::DrainManagerPtr createDrainManager(Server::Instance& server) override;
+  Runtime::LoaderPtr createRuntime(Server::Instance& server,
+                                   Server::Configuration::Initial& config) override;
+};
+
+class MainCommonBase {
```
KBaichoo, line 33: Can this class be documented? e.g. especially differences between MainCommon and MainCommonBase

KBaichoo (commented):
Few questions: Can you shed light on why we're splitting these e.g. what's the problem we're trying to solve? What the reason is we're removing validation from `MainCommonBase`? Thank you.

alyssawilk:
this should now be a no-op for everything but envoy mobile.

source/exe/BUILD
```diff
@@ -93,6 +118,14 @@ envoy_cc_library(
     ],
 )
 
+envoy_cc_library(
+    name = "envoy_stripped_main_base_lib",
+    deps = [
+        ":stripped_main_base_lib",
+        "//source/common/version:version_linkstamp",
```
abeyad, line 125: nit: from a quick glance looking at the `version_linkstamp` build target, it's not clear to me what it does, and so it's not clear to me the difference between `stripped_main_base_lib` and `envoy_stripped_main_base_lib`.. perhaps a comment would help?

source/exe/stripped_main_base.h
```diff
@@ -0,0 +1,86 @@
...
+public:
+  // Server::DrainManagerFactory
+  Server::DrainManagerPtr createDrainManager(Server::Instance& server) override;
+  Runtime::LoaderPtr createRuntime(Server::Instance& server,
+                                   Server::Configuration::Initial& config) override;
+};
+
+// This is the common main between Envoy and Envoy mobile.
+// It is stripped down to functionality required by Envoy Mobile: anything
+// server-specific should live in MainCommonBase or MainCommon which remind
```
abeyad, line 35: nit: s/remind/remain

jpsim (approved):
+1 I still confirmed that the validation symbols disappear from the EM binary with this change, and from what I can tell there's no functional impact to Envoy Mobile.

abeyad (approved):
Left some minor nits, otherwise LGTM!  It was hard to compare that the code moved from MainCommonBase to StrippedMainBase was correct, but I'm assuming the tests would catch that.

KBaichoo (approved):

source/exe/BUILD
```diff
@@ -150,25 +149,21 @@ envoy_cc_library(
...
-    srcs = [
-        "main_common.cc",
-    ],
-    hdrs = [
-        "main_common.h",
-    ],
+    srcs = ["main_common.cc"],
+    hdrs = ["main_common.h"],
     deps = [
-        "//source/server/config_validation:server_lib",
```
alyssawilk: cc @phlax I think this version should not have passed format checks.  Did we lose buildifier on our format checks?

phlax: i believe it works the same as it always did

afaict the queue was cancelled before the check was run

it the `./tools/code_format/check_format.py` script that runs it i think

alyssawilk:
@KBaichoo Ci should be sorted, PTAL!

KBaichoo (approved):

PR #24295, opened 2022-12-01
mobile: add CircleCI docs generation CI job

jpsim:
This is the CI job that's responsible for building the public website docs at https://envoymobile.io.

Commit Message: mobile: add CircleCI docs generation CI job
Additional Description: This is the CI job that's responsible for building the public website docs at https://envoymobile.io.
Risk Level: Low
Testing: In Envoy Mobile repo.
Docs Changes: Yes :)
Release Notes: None.
Platform Specific Features: Only relevant to mobile.

Part of https://github.com/envoyproxy/envoy/issues/23758

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,32 @@
+version: 2.1
+
+executors:
+  ubuntu-build:
+    description: "A regular build executor based on ubuntu image"
+    docker:
+      - image: envoyproxy/envoy-build-ubuntu:0a02a76af5951bf7f4c7029c0ea6d29d96c0f682
+    # TODO(mattklein123): Get xlarge class enabled
+    resource_class: medium
+    working_directory: /source
+
+jobs:
+  docs:
+    executor: ubuntu-build
+    steps:
+      - checkout
+      - run: mobile/docs/build.sh
+      - add_ssh_keys:
+          fingerprints:
+            - "a2:f7:59:f0:01:8b:91:31:ab:0c:3f:9f:25:4c:1e:e5"
+      - run: mobile/docs/publish.sh
+      - store_artifacts:
+          path: generated/docs
+
+workflows:
+  version: 2
+  all:
+    jobs:
+      - docs:
+          filters:
+            tags:
+              only: /^v.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       - run: mobile/docs/build.sh
       - add_ssh_keys:
           fingerprints:
-            - "a2:f7:59:f0:01:8b:91:31:ab:0c:3f:9f:25:4c:1e:e5"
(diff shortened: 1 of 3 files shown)


jpsim:
@alyssawilk do you think you could set up the CircleCI webhook here, mirroring the one defined in Envoy Mobile? https://github.com/envoyproxy/envoy-mobile/settings/hooks/246245000

jpsim:
Ideally we'll migrate this to GitHub Actions, but I'd like to get it up and running as it's defined in Envoy Mobile first.

phlax:
> Ideally we'll migrate this to GitHub Actions

i was wondering about this - the way we currently publish Envoy docs is with GCS - this is also what lets us use `/docs` etc on PRs so we can always check outputs etc without having to mentally parse rst

iiuc this set up requires using azp (cc @lizan for confirmation on this) so im wondering whether it makes sense to use that for this job specifically

alyssawilk:
Theoretically I've done the webhook so my high level question is do we want to limit this to only pushes containing mobile files or changes to circileci as we did with all the yaml?

jpsim:
> Theoretically I've done the webhook

ok I'll push an empty commit to hopefully trigger it.

> do we want to limit this to only pushes containing mobile files or changes to circileci as we did with all the yaml?

I'm inclined to build on every change for now, the job is reliable in my experience, and pretty fast, finishing in ~2 minutes: https://app.circleci.com/pipelines/github/envoyproxy/envoy-mobile/12852/workflows/98b67571-3236-4b62-aac7-f20c9696f584/jobs/13177

alyssawilk:
SG.  from the UI I think we need to land this before I can configure it on app.circleci.com so want lmk if you want to merge as is or have me point to your branch first to test

jpsim:
> have me point to your branch first to test

If it's possible to point to this branch so we can validate this before we merge, that'd be great.

.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  docs:
+    executor: ubuntu-build
+    steps:
+      - checkout
+      - run: mobile/docs/build.sh
+      - add_ssh_keys:
+          fingerprints:
+            - "a2:f7:59:f0:01:8b:91:31:ab:0c:3f:9f:25:4c:1e:e5"
```
alyssawilk, line 20: this is one of the few things that differed per project so not sure if it'll work but I'm fine landing and seeing :-P

alyssawilk (approved):

.circleci/config.yml
jpsim: Presumably this means publishing won't work.

jpsim:
Docs were successfully pushed in https://github.com/envoy-mobile/envoy-mobile.github.io/commit/9f451fbaa897bd278f38b8e33a8d92da0d2b5ff1

alyssawilk (approved):

PR #24298, opened 2022-12-01
Add fstat to OsSysCalls

ravenblackx:
Commit Message: Add fstat to OsSysCalls
Additional Description: Just another os call wrapper. This functionality will be useful in file system cache implementation.
Risk Level: Very low, not changing any production code paths.
Testing: Added a unit test covering the new function and several others.
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Identical implementation for Posix and Windows.


test/common/api/os_sys_calls_test.cc
```diff
@@ -1,9 +1,65 @@
+#include <fcntl.h>
```
jmarantz, line 1: do you need this? Aren't we working entirely with the OSSysCalls abstraction?

jmarantz (commented):
nice. lgtm with one nit.

jmarantz:
format issues also

/wait

test/common/api/os_sys_calls_test.cc
ravenblackx: Yeah, it won't build on Windows without it because `O_CREAT` and `O_READWR` aren't defined without that header.

jmarantz (approved):
thanks!

source/common/api/win32/os_sys_calls_impl.cc
```diff
@@ -114,8 +114,14 @@ SysCallIntResult OsSysCallsImpl::ioctl(os_fd_t sockfd, unsigned long control_cod
 }
 
 SysCallIntResult OsSysCallsImpl::close(os_fd_t fd) {
-  const int rc = ::closesocket(fd);
-  return {rc, rc != -1 ? 0 : ::WSAGetLastError()};
+  int rc = ::closesocket(fd);
+  const int socket_err = ::WSAGetLastError();
```
alyssawilk, line 118: this doesn't appear to be called out in the release note.  is it tested in the new test?

test/common/api/os_sys_calls_test.cc
```diff
@@ -1,9 +1,68 @@
...
 #include "source/common/api/os_sys_calls_impl.h"
 
+#include "test/test_common/environment.h"
+
 #include "gtest/gtest.h"
 
 namespace Envoy {
 
+// Test happy path for `open`, `pwrite`, `pread`, `fstat`, `close`, `stat` and `unlink`.
+TEST(OsSyscallsTest, OpenPwritePreadFstatCloseStatUnlink) {
```
alyssawilk, line 15: ugh I really want to break these out but you can't really fully test read without write/open, ditto fstat.   so yeah can leave as is but if fully testing close requires tweaking anything let's at least add line breaks between the different sections.  if close is tested we can merge as-is

source/common/api/win32/os_sys_calls_impl.cc
ravenblackx: Discussed in Slack; in existing code close is only called where closesocket behavior is fine. My hope is to later change this back and make it *explicitly* closesocket (for all platforms), and remove everything I've done here after implementing [fstat, pread, pwrite, tmpfiles] in `Envoy::Filesystem::File` and migrating `AsyncFileManager` to using that. But getting it in here allows me to make progress elsewhere while that chain of changes is in flight.

test/common/api/os_sys_calls_test.cc
ravenblackx: Close is [kind of] tested, in that this new test tests the new path, and existing tests implicitly cover the old path. But it turns out Windows doesn't like to *unlink* a file that was created without explicit windows-specific permissions flags, so that shot the auto-merge. :(

alyssawilk (approved):

ravenblackx:
/retest
(unrelated flakes)

ravenblackx:
/retest
(Different unrelated flakes.)

ravenblackx:
/retest
Hoping to get luckier on that first flake. #24469 should land eventually if not!

PR #22392, opened 2022-07-26
[PoC] io: add iouring support

soulxu:
Commit Message: io: add iouring support
Additional Description:
Risk Level: low, it should be enabled by default.
Testing: unit tests and integration tests are expected
Docs Changes: TODO
Release Notes: TODO
Platform Specific Features: Only linux platform support iouring
Runtime guard: (TODO) A runtime flag should be added for this feature.

Signed-off-by: He Jie Xu <hejie.xu@intel.com>
Signed-off-by: Zhi Hao Xie <zhihao.xie@intel.com>
Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@intel.com>

PR #24305, opened 2022-12-02
[PoC] io: support iouring - version 2

soulxu:
Commit Message: [PoC] support iouring - version 2
Additional Description:
The new iteration for https://github.com/envoyproxy/envoy/pull/22392

design doc: https://docs.google.com/document/d/17M15J6vh7ZzoJPCY65j6HNoLSqNEoWnQ3Oa9oKIru0o/edit?usp=sharing

Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:

Signed-off-by: He Jie Xu [hejie.xu@intel.com](mailto:hejie.xu@intel.com)
Signed-off-by: Zhi Hao Xie [zhihao.xie@intel.com](mailto:zhihao.xie@intel.com)
Signed-off-by: Dmitry Rozhkov [dmitry.rozhkov@intel.com](mailto:dmitry.rozhkov@intel.com)


soulxu:
@KBaichoo I submitted a PR for this version although I didn't thought it is ready yet, but yea, welcome and appreciate if you have and feedback or any help here.

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,486 @@
...
+  void onCancel(int32_t result) override;
+private:
+  os_fd_t fd_;
+
+  IoUringHandler& io_uring_handler_;
+  IoUringWorker& parent_;
+
+  sockaddr_storage remote_addr_;
+  socklen_t remote_addr_len_{sizeof(remote_addr_)};
+  os_fd_t connection_fd_{INVALID_SOCKET};
```
KBaichoo, line 36: What's the difference between fd_ and connection_fd_?

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,486 @@
...
+
+  Request* accept_req_{nullptr};
+  Request* cancel_req_{nullptr};
+  Request* close_req_{nullptr};
+
+  bool is_disabled_{false};
+  bool is_pending_accept_{false};
+  bool is_closing_{false};
```
KBaichoo, line 44: I've seen elsewhere in the codebase enums for these state transitions, it might make sense here instead of accept/is closing

source/common/network/socket_interface_impl.h
```diff
@@ -27,11 +41,20 @@ class SocketInterfaceImpl : public SocketInterfaceBase {
   };
 
   static IoHandlePtr makePlatformSpecificSocket(int socket_fd, bool socket_v6only,
-                                                absl::optional<int> domain);
+                                                absl::optional<int> domain,
+                                                Io::IoUringFactory* io_uring_factory = nullptr);
+
+  // TODO (soulxu): making those configurable if needed.
```
KBaichoo, line 47: +1

source/common/network/io_uring_socket_handle_impl.cc
```diff
@@ -0,0 +1,546 @@
...
+  if (result.return_value_ != 0) {
+    throw EnvoyException(fmt::format("getsockname failed for '{}': ({}) {}", fd_, result.errno_,
+                                     errorDetails(result.errno_)));
+  }
+  return Address::addressFromSockAddrOrThrow(ss, ss_len, socket_v6only_);
+}
+
+Address::InstanceConstSharedPtr IoUringSocketHandleImpl::peerAddress() {
+  // TODO(rojkov): This is a copy-paste from Network::IoSocketHandleImpl.
+  // Unification is needed.
```
KBaichoo, line 328: +1

KBaichoo (commented):
This is a great start. At a high level: Is there a design doc on this so I can better understand from a high level? Also, can you fix dco so we can see how many tests pass? Thank you

source/common/io/io_uring.h
```diff
@@ -97,24 +116,115 @@ class IoUring {
...
+  virtual void disableSocket(os_fd_t fd) PURE;
+  virtual IoUringSocket& addAcceptSocket(os_fd_t fd, IoUringHandler& handler) PURE;
+  virtual IoUringSocket& addServerSocket(os_fd_t fd, IoUringHandler& handler, uint32_t read_buffer_size) PURE;
+  virtual IoUringSocket& addClientSocket(os_fd_t fd, IoUringHandler& handler, uint32_t read_buffer_size) PURE;
+  virtual void closeSocket(os_fd_t fd) PURE;
+  virtual std::unique_ptr<IoUringSocket> removeSocket(os_fd_t) PURE;
+  virtual Event::Dispatcher& dispatcher() PURE;
+  virtual IoUringSocket& getIoUringSocket(os_fd_t fd) PURE;
+
+  virtual Request* submitAcceptRequest(IoUringSocket& socket, sockaddr_storage* remote_addr,
```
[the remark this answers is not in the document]
soulxu: This is the `user data` pass to the kernel when submit the io uring request, then it will be deallocate here https://github.com/envoyproxy/envoy/pull/24305/files#diff-5a267b3ed3d6b9001c2748d9940424e4582512c97271f77664dd0e126af81160R146, Then I use this to track the request. let me think about if there is better way to manager those pointers in next iteration.

source/common/io/io_uring_factory.cc
```diff
@@ -0,0 +1,41 @@
...
+
+OptRef<IoUringWorker> IoUringFactoryImpl::getIoUringWorker() const {
+  auto ret = tls_.get();
+  if (ret == absl::nullopt) {
+    return absl::nullopt;
+  }
+  return ret.ref();
+}
+
+OptRef<IoUring> IoUringFactoryImpl::get() const {
```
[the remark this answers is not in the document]
soulxu: yea, will do

source/common/io/io_uring_worker.h
soulxu: `fd_` is the socket we listening on. The `connection_fd_` store the socket which returns by accepted. https://github.com/envoyproxy/envoy/pull/24305/files/2013f76b1f070b655e2e7e90f60d74c760e6c0f3#diff-5a267b3ed3d6b9001c2748d9940424e4582512c97271f77664dd0e126af81160R27

Then IoUringSocketHandle will return `connection_fd_` when the upper layer code invokes the IoUringSocketHandleImpl::accept

source/common/io/io_uring_worker.h
soulxu: +1, that is what I'm thinking too, will do in the next iteration

soulxu:
> This is a great start. At a high level: Is there a design doc on this so I can better understand from a high level? Also, can you fix dco so we can see how many tests pass? Thank you

We will prepare a doc at the end of this week or next week. (we want to prepare the performance result), and hoping to collect high-level feedback from the community. 

Let me fix the DCO, in my local, the envoy has 81 integration tests, and we have 25 failed. and there were 17 tests supposed to fail. 

I'm thinking jump into the next iteration (after getting feedback from the community to ensure we are in the right direction), with better quality code, and with the unit tests, then I hope all the corner cases can be fixed.

source/common/io/io_uring.h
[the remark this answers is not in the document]
soulxu: Maybe I can add `Request` as a member of IoUringAcceptSocket, since there always only one accept request at one time. Then I needn't allocate/deallocate memory for it.

soulxu:
Hi, @KBaichoo @mattklein123 @yanavlasov @ggreenway @alyssawilk 

I and @zhxie worked out the second version of io_uring support. Although it isn't perfect. But we think it is time to get high-level feedback from the community to ensure we are in the right direction. We summarize the design and put some benchmark results in here https://docs.google.com/document/d/17M15J6vh7ZzoJPCY65j6HNoLSqNEoWnQ3Oa9oKIru0o/edit?usp=sharing

Compared to the first version https://github.com/envoyproxy/envoy/pull/22392, it can pass most of the integration tests. Envoy has total 81 integration tests, we tested the second version in local dev env, and only left 25 tests failed (17 tests are supposed to fail, most of them due to not implemented features. We only implement the read/write/readv/writev for TCP in this version). We think the failed tests should be some corner case we didn't care about that right in this PoC.

We also change the design since some issues can't be fixed in the previous version, and the previous version becomes very hard to maintain and lose readability by adding more implementation. The detail of the design is described in the doc.

We still focus on PoC functionality in this version, we didn't do any optimization on the code yet, but it is available for some benchmarking. The benchmark result seems pretty good. We also see a case that didn't gain any performance, which is still under investigation. All the details are in the doc also.

If the direction is right, we plan to jump into the next iteration. In the next iteration, we expect the code to be ready for review and with unit tests, then we expect the left tests can be fixed with high-quality code. We still want to put all the code into a single PR in the next iteration, to ensure we get workable code in the end. Then we will separate the PR into multiple small PRs and ask the maintainers to do the official review.

Then one thing we need help which is how to test io_uring on the CI, I guess we need to run both socket API and io_uring version on the CI. which needs to figure out in the next step.

Anyway appreciate any feedback and review, let us know whether we are in the right direction, and thanks in advance!

mattklein123:
I looked at the doc and it looks reasonable to me. I'm excited to see this work land!

As I mentioned previously, I believe if this code is compiled in (either as an extension or with ifdef support) we should use it if the kernel supports the io_uring operations we need without any additional configuration. We shouldn't make the user manually enable this to get the perf boost. You might want to add a section to the doc around configuration, runtime guarding, etc. to make sure we are on the same page there. Thank you.

source/common/network/io_uring_socket_handle_impl.h
```diff
@@ -0,0 +1,125 @@
...
+  Event::FileReadyCb cb_;
+
+  OptRef<Io::AcceptedSocketParam> accepted_socket_param_{absl::nullopt};
+  OptRef<Io::ReadParam> read_param_{absl::nullopt};
+  OptRef<Io::WriteParam> write_param_{absl::nullopt};
+  IoUringSocketType io_uring_socket_type_{IoUringSocketType::Unknown};
+
+  std::unique_ptr<IoHandle> shadow_io_handle_;
+  bool enable_server_socket_{true};
+  bool enable_client_socket_{true};
```
soulxu, line 121: note for the reviewer: the shade iohandle is just for debugging, it is enabled to debug accept/server/client socket separately. Like you can disable server/client socket to use normal iohandle, then debug io_uring accept socket.

soulxu:
> I looked at the doc and it looks reasonable to me. I'm excited to see this work land!
> 
> As I mentioned previously, I believe if this code is compiled in (either as an extension or with ifdef support) we should use it if the kernel supports the io_uring operations we need without any additional configuration. We shouldn't make the user manually enable this to get the perf boost. You might want to add a section to the doc around configuration, runtime guarding, etc. to make sure we are on the same page there. Thank you.

Matt, Thanks! I added a section for the configuration. We can enable io_uring by default, but it should be protected by a runtime flag for a few releases before we think it is stable enough.

mattklein123:
> Matt, Thanks! I added a section for the configuration. We can enable io_uring by default, but it should be protected by a runtime flag for a few releases before we think it is stable enough.

OK this makes sense at a high level. I would default the runtime flag to false initially, then true after some testing, then we can remove it later.

mattklein123:
Sure that's fine, though in general I would hope we could autodetect when it is appropriate to use or not.

soulxu:
> Sure that's fine, though in general I would hope we could autodetect when it is appropriate to use or not.

got it, thanks!

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+      WriteParam param{result};
+      io_uring_handler_.onWrite(param);
+    }
+  }
+
+  uint64_t write(Buffer::Instance& buffer) override {
+    ENVOY_LOG(trace, "write, size = {}, fd = {}, slices = {}", buffer.length(), fd_,
+              buffer.getRawSlices().size());
+    uint64_t buffer_limit = 1024 * 1024;
+    uint64_t writable_size = buffer_limit - write_buf_.length();
```
KBaichoo, line 309: what prevents this from underflowing?

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+    if (writable_size <= 0) {
```
KBaichoo, line 310: since the above underflows, this would only occur if writable_size == 0.

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+      is_full_ = false;
+      WriteParam param{result};
+      io_uring_handler_.onWrite(param);
+    }
+  }
+
+  uint64_t write(Buffer::Instance& buffer) override {
+    ENVOY_LOG(trace, "write, size = {}, fd = {}, slices = {}", buffer.length(), fd_,
+              buffer.getRawSlices().size());
+    uint64_t buffer_limit = 1024 * 1024;
```
KBaichoo, line 308: would be great if this was configurable / used Envoy limits

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+    iov_->iov_len = read_buffer_size_;
+    read_req_ = parent_.submitReadRequest(*this, iov_);
+  }
+
+  void submitWriteRequest() {
+    ASSERT(write_req_ == nullptr);
+    ASSERT(iovecs_ == nullptr);
+
+    Buffer::RawSliceVector slices = write_buf_.getRawSlices();
+    iovecs_ = new struct iovec[slices.size()];
```
KBaichoo, line 362: What's the reason we need to use new directly here?

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+#include "source/common/common/logger.h"
+#include "source/common/io/io_uring.h"
+#include "source/common/io/io_uring_impl.h"
+
+#include "io_uring.h"
+
+namespace Envoy {
+namespace Io {
+
+class IoUringAcceptSocket : public IoUringSocket, protected Logger::Loggable<Logger::Id::io> {
```
KBaichoo, line 13: could also make this inherit from another entity if some of the methods e.g. writev, etc. don't make sense for this entity.

KBaichoo (commented):
Left some comments on the doc as well, here's another pass.

This looks very promising!

source/common/io/io_uring.h
```diff
@@ -97,24 +116,122 @@ class IoUring {
...
+struct ReadParam {
+  Buffer::Instance& pending_read_buf_;
+  int32_t result_;
+};
+
+struct WriteParam {
+  int32_t result_;
+};
+
+class IoUringSocket : public Event::DeferredDeletable {
```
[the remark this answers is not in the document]
soulxu: got it, I will move those to `envoy/...` and add comments.

source/common/io/io_uring.h
```diff
@@ -97,24 +116,122 @@ class IoUring {
...
+  virtual void onConnect(int32_t) { PANIC("not implemented"); };
+  virtual void onRead(int32_t) { PANIC("not implemented"); };
+  virtual void onWrite(int32_t, Request*) { PANIC("not implemented"); };
+};
+
+class IoUringWorker : public ThreadLocal::ThreadLocalObject {
+public:
+  virtual ~IoUringWorker() = default;
+
+  virtual void start() PURE;
```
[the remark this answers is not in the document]
soulxu: I think it can be started on creation, I add this method in the beginning, in the end I didn't find it is useful.

source/common/io/io_uring.h
```diff
@@ -97,24 +116,122 @@ class IoUring {
...
+public:
+  virtual ~IoUringSocket() = default;
+
+  virtual os_fd_t fd() const PURE;
+  virtual void start() PURE;
+  virtual void close() PURE;
+  virtual void enable() PURE;
+  virtual void disable() PURE;
+
+  virtual uint64_t write(Buffer::Instance&) { PANIC("not implemented"); }
```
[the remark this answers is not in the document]
soulxu: got it

source/common/io/io_uring_impl.cc
```diff
@@ -64,19 +50,72 @@ bool IoUringImpl::isEventfdRegistered() const { return SOCKET_VALID(event_fd_);
...
 
+  // move the current injected completions into temp vector in case
+  // the event callback will inject new completion, then we should
+  // go through those new completion in next round.
+  current_injected_completions_.swap(injected_completions_);
+
   eventfd_t v;
-  int ret = eventfd_read(event_fd_, &v);
-  RELEASE_ASSERT(ret == 0, "unable to drain eventfd");
+  while (true) {
```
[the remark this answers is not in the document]
soulxu: yea, it could be a bounded loop.

source/common/io/io_uring_impl.cc
```diff
@@ -64,19 +50,72 @@ bool IoUringImpl::isEventfdRegistered() const { return SOCKET_VALID(event_fd_);
...
+  injected_completions_.emplace_back(fd, user_data, result);
+  ENVOY_LOG(trace, "inject completion, fd = {}, req = {}, num injects = {}", fd,
+            fmt::ptr(user_data), injected_completions_.size());
+}
+
+void IoUringImpl::removeInjectedCompletion(os_fd_t fd) {
+  ENVOY_LOG(debug, "remove injected completion, fd = {}", fd);
+  for (auto iter = injected_completions_.begin(); iter != injected_completions_.end();) {
+    if (iter->fd_ == fd) {
+      iter = injected_completions_.erase(iter);
```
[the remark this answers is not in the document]
soulxu: got it, thanks!

source/common/io/io_uring_impl.h
```diff
@@ -11,7 +9,16 @@ namespace Io {
 
 bool isIoUringSupported();
 
-class IoUringImpl : public IoUring, public ThreadLocal::ThreadLocalObject {
+struct InjectedCompletion {
+  InjectedCompletion(os_fd_t fd, void* user_data, int32_t result)
+      : fd_(fd), user_data_(user_data), result_(result) {}
+
+  os_fd_t fd_;
```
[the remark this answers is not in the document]
soulxu: got it

source/common/io/io_uring_worker.h
soulxu: Emm...sorry, I'm not sure I get what you mean. Do you mean using another base class here? And that base class hasn't the methods like `writev`?

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+      ::close(fd_);
+    }
+  }
+
+  // IoUringSocket
+  os_fd_t fd() const override { return fd_; }
+
+  void start() override {
+    ENVOY_LOG(debug, "start, fd = {}", fd_);
+    if (read_req_ == nullptr) {
```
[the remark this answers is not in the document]
soulxu: I think in most the cases, after we create the socket, then we will start to read/write the data. If we submit the read request first after creating the socket, then when we want to read later, the data may already be ready. I thought this will be more efficient.

Otherwise, we can submit the read req lazily. We can submit the read request when the `Iohandle::read` is invoked, but that means we need to return -EAGAIN on the first `Iohandle::read` invoke, and wait for iouring to get read finished. After the read request is done, we trigger the read event again, then wait for the up layer invoke `ioHandle::read` again to fetch the data.

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+  void onCancel(int32_t result) override {
+    cancel_req_ = nullptr;
+    ENVOY_LOG(debug, "cancel request done, result = {}, fd = {}", result, fd_);
+
+    if (is_closing_ && read_req_ == nullptr && write_req_ == nullptr && close_req_ == nullptr) {
+      close_req_ = parent_.submitCloseRequest(*this);
+    }
+  }
+
+  void onClose(int32_t result) override {
```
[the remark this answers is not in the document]
soulxu: In theory, the `close` could be failed https://man7.org/linux/man-pages/man2/close.2.html. But I think it should be rare case. Currently, in Envoy, we didn't take care the `close`'s return value also. I think it should be fine to ignore that.

source/common/io/io_uring_worker.h
```diff
@@ -0,0 +1,509 @@
...
+  void close() override {
+    is_closing_ = true;
+    if (read_req_ != nullptr) {
+      ASSERT(cancel_req_ == nullptr);
+      cancel_req_ = parent_.submitCancelRequest(*this, read_req_);
+      return;
+    }
+
+    if (write_req_ != nullptr) {
+      return;
```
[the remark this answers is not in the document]
soulxu: All the requests can be canceled. We should let the `write` request finished first before close. The envoy code still assumes the underlayer is socket API. There could be a close after a write. For socket API, if the `write` returns success, then it means the write is done. So we should mimic that behavior.

source/common/io/io_uring_worker.h
soulxu: emm...good question. There should be no limit. Initially, I just thought the buffer should have a limit. But the envoy up layer code should already limit the buffer size. There won't be a case for unlimited huge size to write.

source/common/io/io_uring_worker.h
soulxu: emm... no, I should be able to use unique ptr here.

source/server/server.cc
```diff
@@ -614,15 +614,26 @@ void InstanceImpl::initialize(Network::Address::InstanceConstSharedPtr local_add
...
+    sock = const_cast<Network::SocketInterface*>(Network::socketInterface(sock_name));
+  }
+
+  if (bootstrap_.default_socket_interface().empty() || sock == nullptr) {
+    auto factory =
+        Registry::FactoryRegistry<Server::Configuration::BootstrapExtensionFactory>::getFactory(
+            "envoy.extensions.network.socket_interface.default_socket_interface");
+    bootstrap_extensions_.push_back(factory->createBootstrapExtension(
+        *factory->createEmptyConfigProto(), serverFactoryContext()));
+    sock = dynamic_cast<Network::SocketInterface*>(factory);
```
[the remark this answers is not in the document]
soulxu: let me try.

PR #4723, opened 2018-10-15
Add typedMetadata for RouteEntry&ClusterInfo

stevenzzzz:
For an explanation of how to fill out the fields, please see the relevant section 
in PULL_REQUESTS.md

*Description*: Add typedMetadata interface for RouteEntry and ClusterInfo. 
design doc : https://docs.google.com/document/d/1AKA_5iMSR6WZaDbjQfLFbK9maO_la_MZgVtYdD9gWtA/edit?usp=sharing
*Risk Level*: LOW (new interface)
*Testing*: unit test.
*Docs Changes*: Api level.
*Release Notes*:
[Optional Fixes #Issue] #4513 
[Optional *Deprecated*:]

PR #24309, opened 2022-12-02
Reduce Route memory utilization by avoiding Metadata instances when not needed

yurykats:
Reduce Route memory utilization by avoiding Metadata instances when not needed

Currently an instance of raw Metadata and an instance of TypedMetadata are created in each route, even if no metadata is configured. This PR switches instances to unique_ptr and only creates Metadata objects when actually configured. Otherwise a singleton default metadata is used.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/router/config_impl.h
```diff
@@ -1061,8 +1075,8 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
   HeaderParserPtr request_headers_parser_;
   HeaderParserPtr response_headers_parser_;
   uint32_t retry_shadow_buffer_limit_{std::numeric_limits<uint32_t>::max()};
-  envoy::config::core::v3::Metadata metadata_;
-  Envoy::Config::TypedMetadataImpl<HttpRouteTypedMetadataFactory> typed_metadata_;
+  std::unique_ptr<envoy::config::core::v3::Metadata> metadata_;
+  std::unique_ptr<RouteTypedMetadata> typed_metadata_;
```
adisuissa, line 1079: nit: should these be const?

source/common/router/config_impl.cc
```diff
@@ -534,6 +533,10 @@ RouteEntryImplBase::RouteEntryImplBase(const VirtualHostImpl& vhost,
                           validator),
       route_name_(route.name()), time_source_(factory_context.mainThreadDispatcher().timeSource()),
       random_value_header_name_(route.route().weighted_clusters().header_name()) {
+  if (route.has_metadata()) {
+    metadata_ = std::make_unique<envoy::config::core::v3::Metadata>(route.metadata());
+    typed_metadata_ = std::make_unique<RouteTypedMetadata>(route.metadata());
+  }
```
adisuissa, line 539: I find it interesting that both wrap the same metadata field.
I guess this is intentional (after reading this [comment thread](https://github.com/envoyproxy/envoy/pull/4723/files#r225213097)), but just wondering whether there should be some ordering between these, or at least both should not store the same metadata.

source/common/router/config_impl.h
```diff
@@ -658,8 +658,22 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
...
+      return *metadata_;
+    }
+    return DefaultMetadata::get();
+  }
+  using RouteTypedMetadata = Envoy::Config::TypedMetadataImpl<HttpRouteTypedMetadataFactory>;
+  const Envoy::Config::TypedMetadata& typedMetadata() const override {
+    if (typed_metadata_ != nullptr) {
+      return *typed_metadata_;
+    }
+    static RouteTypedMetadata* defaultTypedMetadata =
```
adisuissa, line 673: nit: const.

adisuissa: On a second thought, could this be a ConstSingleton by itself (i.e., not dependent on `DefaultMetadata::get()`)?
I'm trading off another object of empty Metadata object against dependencies between singletons. WDYT?

adisuissa (commented):

source/common/router/config_impl.h
pradeepcrao: ConstSingleton as defined does not allow constructors that need input arguments.

source/common/router/config_impl.cc
yurykats: Yes, I find it strange as well, but it does keep the same metadata in two forms -- as a raw protobuf, and converted to "typed" data structure. I am not aiming to change this here, just avoid creation of both when there is no actual metadata.
To make things even more confusing, RDS keeps a full copy of the entire raw protobuf anyway, so a raw copy of the metadata message here is pure overhead. Again, not for this PR.

source/common/router/config_impl.h
yurykats: What @pradeepcrao said, typed metadata needs a raw metadata instance to be created.

pradeepcrao: Might be possible to do using type traits which will make this cleaner. I can attempt that in a follow-up PR.

source/common/router/config_impl.h

source/common/router/config_impl.h
yurykats: Added const. Leaving the singleton as is.

adisuissa (dismissed):
LGTM, thanks!

zuercher (approved):

adisuissa (approved):
LGTM, thanks!
Note: no changes since last approval by @zuercher (just merge-main conflict that was solved), so this PR should be ok to merge.

PR #24318, opened 2022-12-02
Geoip filter

nezdolik:
This change introduces  IP Geolocation filter. Implementation according to [design doc](https://docs.google.com/document/d/1klskyqKxgv5xBq1Xx38x5i5pUx7dkSztaSaoECJovJA/edit?usp=sharing).

Commit Message: IP Geolocation filter
Additional Description:
Risk Level: Low (new extension)
Testing: unit tests
Docs Changes: Done
Release Notes: tbd
Fixes moby/moby#23736


api/envoy/extensions/filters/http/geoip/v3/geoip.proto
```diff
@@ -0,0 +1,77 @@
...
+    // If set, the header will be used to populate the anonymizing hosting provider associated with the IP address.
+    string anon_hosting = 6 [(validate.rules).string = {well_known_regex: HTTP_HEADER_NAME ignore_empty: true}];
+    // If set, the header will be used to populate the anonymizing TOR exit node associated with the IP address.
+    string anon_tor = 7 [(validate.rules).string = {well_known_regex: HTTP_HEADER_NAME ignore_empty: true}];
+    // If set, the header will be used to populate the anonymizing proxy associated with the IP address.
+    string anon_proxy = 8 [(validate.rules).string = {well_known_regex: HTTP_HEADER_NAME ignore_empty: true}];
+  }
+
+  message GeolocationProvider {
+    // The name of the geoip client to instantiate. The name must match a
```
nezdolik, line 62: todo rename to driver

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -0,0 +1,107 @@
...
+  if (config_->use_xff() && config_->xffNumTrustedHops() > 0) {
+    remote_address =
+        Envoy::Http::Utility::getLastAddressFromXFF(headers, config_->xffNumTrustedHops()).address_;
+  }
+  // If `config_->use_xff() == false` or xff header has not been populated for some reason.
+  if (!remote_address) {
+    remote_address = callbacks_->streamInfo().downstreamAddressProvider().remoteAddress();
+  }
+  // todo add rest of headers. Does driver supports complex lookup?
+  if (config_->geoCountryHeader()) {
```
nezdolik, line 76: todo extract into method

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -0,0 +1,115 @@
...
+  setupForGeolocationHeader(geo_city_header_, geo_headers_to_add, &GeolocationHeadersToAdd::city);
+  setupForGeolocationHeader(geo_region_header_, geo_headers_to_add, &GeolocationHeadersToAdd::region);
+  setupForGeolocationHeader(geo_asn_header_, geo_headers_to_add, &GeolocationHeadersToAdd::asn);
+  setupForGeolocationHeader(geo_anon_vpn_header_, geo_headers_to_add, &GeolocationHeadersToAdd::anon_vpn);
+  setupForGeolocationHeader(geo_anon_hosting_header_, geo_headers_to_add, &GeolocationHeadersToAdd::anon_hosting);
+  setupForGeolocationHeader(geo_anon_tor_header_, geo_headers_to_add, &GeolocationHeadersToAdd::anon_tor);
+  setupForGeolocationHeader(geo_anon_proxy_header_, geo_headers_to_add, &GeolocationHeadersToAdd::anon_proxy);
+}
+
+void GeoipFilterConfig::setupForGeolocationHeader(absl::optional<std::string>& geo_header, const GeolocationHeadersToAdd& geo_headers_to_add,
```
nezdolik, line 36: todo function param can be reduced to string

nezdolik:
@mattklein123 @alyssawilk wdyt if we split work on geolocation filter into 2 PRs? First PR contains the filter implementation and second PR will contain implementation of geolocation provider (Maxmind). Or would you prefer to have all changes in one large PR? Am proposing to split, as current PR is already big.

alyssawilk:
throwing over to raven for first pass review.  Not sure who will be assigned to second pass but I think I can speak for all of us in saying splitting up into smaller PRs is generally helpful :-)

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -0,0 +1,132 @@
...
+}
+
+void GeoipFilter::lookupGeolocationData(
+    const absl::optional<std::string>& geo_header,
+    const Network::Address::InstanceConstSharedPtr& remote_address,
+    const absl::optional<std::string>& (Driver::*func)(
+        const Network::Address::InstanceConstSharedPtr& address) const,
+    Http::RequestHeaderMap& headers) {
+  if (geo_header) {
+    auto lookupResult = ((*driver_).*func)(remote_address);
```
ravenblackx, line 92: Are we confident that this is going to be fast enough to not leave the worker thread blocked on I/O for any significant amount of time? It seems likely that this would be risky to extend to other geoip sources for that reason. Might it be better to implement with callbacks and handle asynchronously?

test/extensions/filters/http/geoip/config_test.cc
```diff
@@ -0,0 +1,165 @@
...
+               GeoipFilterPeer::geoCityHeader(*filter).value() == "x-geo-city" &&
+               !GeoipFilterPeer::geoCountryHeader(*filter) &&
+               !GeoipFilterPeer::geoRegionHeader(*filter) &&
+               !GeoipFilterPeer::geoAsnHeader(*filter) &&
+               !GeoipFilterPeer::geoAnonVpnHeader(*filter) &&
+               !GeoipFilterPeer::geoAnonHostingHeader(*filter) &&
+               !GeoipFilterPeer::geoAnonTorHeader(*filter) &&
+               !GeoipFilterPeer::geoAnonProxyHeader(*filter);
+      };
+  EXPECT_CALL(filter_callback, addStreamDecoderFilter(Truly(has_expected_config)));
```
ravenblackx, line 59: This is *okay*, but in the event of a test failure it would have completely unhelpful output. Consider defining a bunch of matchers (I think they could be one-liners) instead of one globby function, and doing 
```
MATCHER_P(HasUseXff, expected, "") {return GeoipFilterPeer::useXff(*arg);}

EXPECT_CALL(filter_callback, addStreamDecoderFilter(AllOf(
    HasUseXff(false),
    HasXffNumTrustedHops(0),
    [etc.]
)));
```

Makes even more sense when you can also reuse this in the other `has_expected_config` case[s] below.

source/extensions/filters/http/geoip/geoip_provider_config.h
```diff
@@ -0,0 +1,154 @@
...
+  virtual ~Driver() = default;
+
+  /**
+   *  Looks up a city that is associated with the given IP address.
+   *
+   *  @param   address IP address (IPv4, IPv6) to perform lookup for.
+   *  @returns city for the IP address or absl::nullopt if address is not found in the geolocation
+   * database.
+   */
+  virtual const absl::optional<std::string>&
```
ravenblackx, line 25: Seems unsafe for all these functions to be returning references - if there's any kind of dynamic database reload possible, or if the results for an implementation aren't all held in memory at once, the reference could become invalid before it's used. (References wouldn't work with an RPC-based provider, for example.)

ravenblackx (changes requested):

api/envoy/extensions/filters/http/geoip/v3/maxmind.proto
```diff
@@ -0,0 +1,25 @@
+syntax = "proto3";
```
ravenblackx, line 1: This file should go in a location that matches its type and the path to the implementation. For example, extensions that go in the cache filter live in `envoy/extensions/http/cache` not in `envoy/extensions/filters/http/cache`.

Since this provider is not http specific at all, I would think `envoy/extensions/geoip_providers` would be the place.

ravenblackx:
> throwing over to raven for first pass review. Not sure who will be assigned to second pass but I think I can speak for all of us in saying splitting up into smaller PRs is generally helpful :-)

Agreeing that maxmind in a separate PR makes sense. I get why maxmind.proto is in this one though, because it's doing cross-referencing and you don't want to need API reviewers to act on the same file twice. It sure would be nice if we could land the API as a first review, with no implementation at all, and then do the implemention afterwards, but those integration checks hate that idea.

api/envoy/extensions/filters/http/geoip/v3/geoip.proto
```diff
@@ -0,0 +1,81 @@
...
+
+option java_package = "io.envoyproxy.envoy.extensions.filters.http.geoip.v3";
+option java_outer_classname = "GeoipProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/geoip/v3;geoipv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: Geoip]
+// Geoip :ref:`configuration overview <config_http_filters_geoip>`.
```
ravenblackx, line 18: After you fix the current docs CI failure, I think you're going to run into one here as well - you must add that doc file if you're including a reference to it.

source/extensions/filters/http/geoip/geoip_provider_config.h
nezdolik: good point, missed it

nezdolik:
Thanks for the detailed review @ravenblackx,  i have switched filter code to an async version. The doc checks will still fail on missing documentation for maxmind extension. I will fix it shortly.

source/extensions/filters/http/geoip/geoip_filter.cc
nezdolik: Agree, async fits much better for this use case

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -0,0 +1,157 @@
...
+  }
+  config_->incTotal(geo_header.value());
+  if (allLookupsComplete()) {
+    ENVOY_LOG(debug, "Geoip filter: finished decoding geolocation headers");
+    decoder_callbacks_->continueDecoding();
+  }
+}
+
+bool GeoipFilter::allLookupsComplete() {
+  return all_lookups_invoked_ && num_lookups_to_perform_ == 0;
```
nezdolik, line 151: For now i think those two flags are enough to track the state of lookups completion per request. But i may have missed some edge cases

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -0,0 +1,157 @@
...
+  ASSERT(num_lookups_to_perform_ > 0);
+  --num_lookups_to_perform_;
+  if (lookup_result) {
+    request_headers_->setCopy(Http::LowerCaseString(geo_header.value()), lookup_result.value());
+    config_->incHit(geo_header.value());
+  }
+  config_->incTotal(geo_header.value());
+  if (allLookupsComplete()) {
+    ENVOY_LOG(debug, "Geoip filter: finished decoding geolocation headers");
+    decoder_callbacks_->continueDecoding();
```
ravenblackx, line 146: I think this doesn't work great with asynchronous callbacks - num_lookups_to_perform_ being updated by the callback directly means it would need to be mutex-guarded or atomic to prevent overlapping operations occurring on different threads. There's also an issue that the filter 'this' can potentially be destroyed while an operation is in flight.

You could resolve this (and other raciness issues so you'd no longer need `all_lookups_invoked_`) by having the callback go through the dispatcher on decoder_callbacks_, and capture a weak_ptr to something owned by the filter. Using the dispatcher effectively makes the filter single-threaded again (i.e. the filter part of the callback can't be called while decodeHeaders is still processing), and capturing a weak_ptr means you can abandon the rest of the callback if the filter has been destroyed before the callback is called.

I don't think there's a way to do this safely where the driver takes only a reference to the callback object - it must take a shared_ptr or a std::function so that it's safe for the filter to be destroyed without causing the driver's async callback to lose its necessary context.

My suggestion would be to pass the driver a unique_ptr to an additional LookupCallbacks implementation that contains a reference to the dispatcher and a weak_ptr that leads back to the filter. When called, that object could then `dispatcher_->dispatch([the_weak_ptr, callback_params]() {if (auto p = the_weak_ptr.lock()) p->onLookupComplete(callback_params);})`

test/extensions/filters/http/geoip/geoip_filter_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  request_headers.addCopy("x-forwarded-for", "10.0.0.1,10.0.0.2");
+  expectStats("x-geo-region");
+  Network::Address::InstanceConstSharedPtr remote_address =
+      Network::Utility::parseInternetAddress("1.2.3.4");
+  filter_callbacks_.stream_info_.downstream_connection_info_provider_->setRemoteAddress(
+      remote_address);
+  EXPECT_EQ(Http::FilterHeadersStatus::StopIteration,
+            filter_->decodeHeaders(request_headers, false));
+  EXPECT_CALL(filter_callbacks_, continueDecoding());
+  filter_->onLookupComplete(dummy_region_, "x-geo-region");
```
ravenblackx, line 144: Could you make dummy_driver_ also SaveArg<2> (or 3), the callback, and then instead of directly calling onLookupComplete, call it through the captured callback? That would get you better coverage of how the filter actually works. Current testing doesn't verify that the driver gets a correct callback, nor that calling the driver's callback does anything.

(Might make more sense to not use SaveArg, and instead make a custom function that captures all the args into a specified dummy_driver_ member variable struct, so you can save the different callbacks to different slots without having to copy-paste 8 near-duplicate sets of SaveArg.)

test/extensions/filters/http/geoip/geoip_filter_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  filter_->onLookupComplete(empty_response_, "x-geo-city");
+
+  EXPECT_EQ(1, request_headers.size());
+  EXPECT_EQ("dummy_region",
+            request_headers.get(Http::LowerCaseString("x-geo-region"))[0]->value().getStringView());
+  EXPECT_EQ("1.2.3.4:0", captured_address_->asString());
+}
+
```
ravenblackx, line 273: Combination of the other comments - please add a test `NoCrashIfFilterDestroyedBeforeCallbackCalled`, in which filter_->decodeHeaders is called and queues up some driver callbacks, then `filter_.clear()`, then call the callbacks.

source/extensions/filters/http/geoip/geoip_filter.h
```diff
@@ -0,0 +1,126 @@
...
+                           const LookupCallbacks& callbacks,
+                           const absl::optional<std::string>& geo_header) const);
+  bool allLookupsComplete();
+  // Allow the unit test to have access to private members.
+  friend class GeoipFilterPeer;
+  GeoipFilterConfigSharedPtr config_;
+  Http::StreamDecoderFilterCallbacks* decoder_callbacks_{};
+  DriverSharedPtr driver_;
+  bool all_lookups_invoked_{false};
+  Envoy::Http::RequestHeaderMap* request_headers_ = nullptr;
```
ravenblackx, line 119: I think the Envoy way to do this now is
`OptRef<Http::RequestHeaderMap> request_headers_;`

test/extensions/filters/http/geoip/geoip_filter_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  EXPECT_EQ(Http::FilterHeadersStatus::StopIteration,
+            filter_->decodeHeaders(request_headers, false));
+  EXPECT_CALL(filter_callbacks_, continueDecoding());
+  filter_->onLookupComplete(dummy_region_, "x-geo-region");
+  EXPECT_EQ(2, request_headers.size());
+  expectHeader(request_headers, "x-geo-region", "dummy_region");
+  EXPECT_EQ("10.0.0.1:0", captured_address_->asString());
+}
+
+TEST_F(GeoipFilterTest, GeoHeadersOverridenForIncomingRequest) {
```
ravenblackx, line 150: Overriden -> Overridden

test/extensions/filters/http/geoip/geoip_filter_test.cc
```diff
@@ -0,0 +1,278 @@
...
+    EXPECT_CALL(*dummy_driver_, getIsAnonymousPublicProxy(_, _, _))
+        .WillRepeatedly(DoAll(SaveArg<0>(&captured_address_)));
+  }
+
+  void expectStats(const std::string& geo_header) {
+    EXPECT_CALL(stats_, counter(absl::StrCat("prefix.geoip.", geo_header, ".total")));
+    EXPECT_CALL(stats_, counter(absl::StrCat("prefix.geoip.", geo_header, ".hit")));
+  }
+
+  void expectHeader(const Http::TestRequestHeaderMapImpl& request_headers,
```
ravenblackx, line 73: Could you make this a `MATCHER_P`? (It would make it easier for the next person to find and move into a common place, if you don't want to put it somewhere common already, and you get better line numbers on failure with a proper EXPECT_THAT rather than a helper function.)

Also, right now you'd get an exception thrown on the EXPECT_EQ if the EXPECT_TRUE fails, and not very helpful output. As a MATCHER_P you could easily `*result_listener << "bla bla bla" << request_headers` on each type of failure.

test/extensions/filters/network/redis_proxy/command_splitter_impl_test.cc
```diff
@@ -73,7 +73,7 @@ class RedisCommandSplitterImplTest : public testing::Test {
 
   Event::SimulatedTimeSystem time_system_;
   InstanceImpl splitter_{std::make_unique<NiceMock<MockRouter>>(route_),
-                         store_,
+                         *store_.rootScope(),
```
ravenblackx, line 76: This looks like a change that belongs in a different PR.

ravenblackx (changes requested):

source/extensions/filters/http/geoip/geoip_filter.cc
nezdolik: I may have misinterpreted the docs, was thinking that http filter instance is created per stream/request and that no guarding is needed.

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -0,0 +1,157 @@
...
+}
+
+void GeoipFilter::setDecoderFilterCallbacks(Http::StreamDecoderFilterCallbacks& callbacks) {
+  decoder_callbacks_ = &callbacks;
+}
+
+void GeoipFilter::onLookupComplete(const absl::optional<std::string>& lookup_result,
+                                   const absl::optional<std::string>& geo_header) {
+  if (!request_headers_) {
+    ENVOY_LOG(debug, "Geoip filter: no header to decorate in reguest");
```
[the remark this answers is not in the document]
nezdolik: agree

source/extensions/filters/http/geoip/geoip_filter.cc
ravenblackx: No, you have that part right. The issue is that the driver isn't part of the filter, and (being asynchronous) may have its own thread. So now there's the filter's thread and the driver's thread (or threads! There's multiple calls to the driver in flight at the same time), both/all touching about the same variables. If the callbacks are applied through the dispatcher then they end up back on the filter's thread which therefore would eliminate any need to guard the variables. (But does still need to be careful that the filter instance hasn't been destroyed by the time the dispatcher gets around to dispatching.)

wbpcode:
/wait 

Mart it as waiting. CI need to be resolved and previous comments need to be addressed. Feel free to ping the maintainer when It's readly for a new review.

source/extensions/filters/http/geoip/geoip_filter.cc
nezdolik: Yeap, fair. Thanks @ravenblackx and apologies for delay, I have been OOO entire February. Am now catching up and continuing this work.

CODEOWNERS
```diff
@@ -291,6 +291,8 @@ extensions/filters/http/oauth2 @derekargueta @snowp
 /*/extensions/load_balancing_policies/round_robin @wbpcode @UNOWNED
 # Early header mutation
 /*/extensions/http/early_header_mutation/header_mutation @wbpcode @UNOWNED
+# IP Geolocation
+/*/extensions/filters/http/geoip @nezdolik @mattklein123
```
nezdolik, line 295: @mattklein123 i will update the ownership once it's clear who is sponsoring the change

ravenblackx:
I see a CI test failure looking like this:
```
  Expected arg #0: (has use xff true) and (has xff num trusted hops 1) and (has geo header "x-geo-city") and (has geo header "x-geo-region") and (has geo headers size 2) and (has anon geo headers size 1) and (has geo anon header "x-anon-vpn")
           Actual: 0x17ebbf8a2088
```
We get good human-readable output first, but unhelpful "actual". I don't think we can reasonably fix the "actual" because the parameter type being matched is a shared_ptr to a superclass, but I believe you could make this failure output more human-readable by altering the matchers to have additional output, e.g.
```
MATCHER_P(HasUseXff, expected, "") {
  auto filter = std::static_pointer_cast<GeoipFilter>(arg);
  if (GeoipFilterPeer::useXff(*filter) == expected) {
    return true;
  }
  *result_listener << "expected useXff=" << expected << " but was " << GeoipFilterPeer::useXff(*filter);
}
```
(But I'm not sure this will help - this aspect of matchers never behaves quite how I expect.)

nezdolik:
@ravenblackx after updating matchers according to your recommendation the fail case now looks like this:
```test/extensions/filters/http/geoip/config_test.cc:131: EXPECT_CALL(filter_callback, addStreamDecoderFilter( AllOf(HasUseXff(false), HasXffNumTrustedHops(1), HasGeoHeader("x-geo-country"), HasGeoHeader("x-geo-region"), HasGeoHeadersSize(2), HasAnonGeoHeadersSize(1), HasGeoAnonHeader("x-anon-vpn"))))...
  Expected arg #0: (has use xff false) and (has xff num trusted hops 1) and (has geo header "x-geo-country") and (has geo header "x-geo-region") and (has geo headers size 2) and (has anon geo headers size 1) and (has geo anon header "x-anon-vpn")
           Actual: 16-byte object <90-30 DF-3F FE-06 00-00 80-30 DF-3F FE-06 00-00>, expected useXff=0 but was 1
         Expected: to be called once
           Actual: never called - unsatisfied and active
```

api/envoy/extensions/filters/http/geoip/v3/geoip.proto
```diff
@@ -21,64 +21,74 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
-    string name = 1 [(validate.rules).string = {min_bytes: 1}];
+    // supported geoip driver.
+    string name = 1 [(validate.rules).string = {min_len: 1}];
 
     // Geoip driver specific configuration which depends on the driver being instantiated.
-    // See the geoip drivers for examples:
-    //
-    // - :ref:`MaxMindConfig <envoy_api_msg_config.geoip.v3.MaxMindConfig>`
-    // [#extension-category: envoy.geoip_providers]
     oneof config_type {
```
ravenblackx, line 72: Why is this in a oneof? An Any type is kind of an implicit version of the same thing.

source/extensions/filters/http/geoip/geoip_filter.cc
```diff
@@ -71,49 +69,31 @@ Http::FilterHeadersStatus GeoipFilter::decodeHeaders(Http::RequestHeaderMap& hea
...
-    const Network::Address::InstanceConstSharedPtr& remote_address,
-    void (Driver::*func)(const Network::Address::InstanceConstSharedPtr& address,
-                         const LookupCallbacks& callbacks,
-                         const absl::optional<std::string>& geo_header) const) {
-  if (geo_header) {
-    ++num_lookups_to_perform_;
-    ((*driver_).*func)(remote_address, *this, geo_header);
-  }
+  // Stop the iteration for headers for the current filter and the filters following.
+  return state_ == State::Complete ? Http::FilterHeadersStatus::Continue
```
ravenblackx, line 95: I think there's no way state_ can be State::Complete here now that you're using the dispatcher, except in unit tests if the mock dispatcher is allowed to just call the callback immediately *and* you have the driver also call its callback immediately. Having a production code branch that's only relevant for tests is best avoided - I'd recommend capturing at least one of the callbacks in unit tests, and calling it after decodeHeaders. That way the code here can be always StopIteration, and the test better reflects actual behavior.

(Otherwise you also need a bunch of unnecessary complexity around *not* calling continueDecoding from the callback if you're still inside decodeHeaders, etc.)

Also, I think with this change `state_` is no longer necessary at all. This is the only place where it's read from, so all the writes to it become meaningless.

source/extensions/filters/http/geoip/geoip_filter.h
```diff
@@ -0,0 +1,99 @@
...
+  void incHit(absl::string_view geo_header) {
+    incCounter(stat_name_set_->getBuiltin(absl::StrCat(geo_header, ".hit"), unknown_hit_));
+  }
+  void incTotal(absl::string_view geo_header) {
+    incCounter(stat_name_set_->getBuiltin(absl::StrCat(geo_header, ".total"), unknown_hit_));
+  }
+
+  bool use_xff() const { return use_xff_; }
+  uint32_t xffNumTrustedHops() const { return xff_num_trusted_hops_; }
+  const absl::flat_hash_set<std::string>& geoHeaders() const { return geo_headers_; }
```
ravenblackx, line 39: Looks like you need a #include for flat_hash_set, from clangtidy errors.

test/extensions/filters/http/geoip/config_test.cc
```diff
@@ -0,0 +1,218 @@
...
+  return false;
+}
+
+MATCHER_P(HasGeoHeader, expected_header, "") {
+  auto filter = std::static_pointer_cast<GeoipFilter>(arg);
+  auto geo_headers = GeoipFilterPeer::geoHeaders(*filter);
+  bool header_found = false;
+  for (auto header : geo_headers) {
+    if (header == expected_header) {
+      header_found = true;
```
ravenblackx, line 54: Can just return true here, not need header_found, and add some *result_listener output before return false after the loop.

test/extensions/filters/http/geoip/config_test.cc
```diff
@@ -0,0 +1,218 @@
...
+                   << GeoipFilterPeer::useXff(*filter);
+  return false;
+}
+
+MATCHER_P(HasGeoHeader, expected_header, "") {
+  auto filter = std::static_pointer_cast<GeoipFilter>(arg);
+  auto geo_headers = GeoipFilterPeer::geoHeaders(*filter);
+  bool header_found = false;
+  for (auto header : geo_headers) {
+    if (header == expected_header) {
```
ravenblackx, line 53: `if (testing::Matches(expected_header)(header))` is better - using `==` requires expected_header to be a string, versus using `Matches` allows e.g. `EXPECT_THAT(bla, HasGeoHeader(HasSubstr("x")))`

(Same for most of these matchers - if there's an `==` prefer `Matches`)

For single comparisons, prefer `testing::ExplainMatchResult(expected_header, header, result_listener)` which would also allow for any additional output from the other matcher to come through.

test/extensions/filters/http/geoip/geoip_filter_test.cc
```diff
@@ -0,0 +1,345 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Geoip {
+namespace {
+
+MATCHER_P2(HasExpectedHeader, expected_header, expected_value, "") {
+  auto request_headers = static_cast<Http::TestRequestHeaderMapImpl>(arg);
+  EXPECT_TRUE(request_headers.has(expected_header));
```
ravenblackx, line 30: In general using EXPECT inside a matcher is likely to result in some confusing behaviors on test failure. Better to do `if (fail condition) {*result_listener<<explanation; return false;}`

In this case because it's using get, the expected_header param can't be a matcher, but still worth using `testing::ExplainMatchResult` for the comparison of expected_value so that sub-matchers are an option.

ravenblackx (changes requested):

phlax:
apologies - ive cancelled CI - please hold off pushing until we have progressed the releases that are currently going through

test/extensions/filters/http/geoip/geoip_filter_test.cc
ravenblackx: You changed it to not need the `EXPECT_TRUE`, but forgot to delete the now unnecessary `EXPECT_TRUE`. :)

ravenblackx (changes requested):
Still got a couple of prior comments to resolve too.

nezdolik:
tsan ci failure seems to be unrelated, will merge master and check if it helps

CODEOWNERS
nezdolik: @ravenblackx should i put you instead of Matt as an owner?

ravenblackx: I guess. It's not really my field, but it would make sense since having reviewed some of it I now have the best context to review the rest of it.

ravenblackx (approved):

ravenblackx (approved):

nezdolik:
cc @sc0ttbeardsley i have started working on plugging the geoip provider for this filter in a separate change. If your company uses some other geolocation provider, you will need to submit a patch for it.

nezdolik:
@mattklein123 could you please take the final look? :)

api/envoy/extensions/filters/http/geoip/v3/geoip.proto
```diff
@@ -0,0 +1,92 @@
...
+  // determining the origin client's IP address. The default is zero if this option
+  // is not specified. See the documentation for
+  // :ref:`config_http_conn_man_headers_x-forwarded-for` for more information.
+  uint32 xff_num_trusted_hops = 2;
+
+  // Configuration for geolocation headers to add to request.
+  GeolocationHeadersToAdd geo_headers_to_add = 3 [(validate.rules).message = {required: true}];
+
+  // Geolocation provider specific configuration.
+  GeolocationProvider provider = 4 [(validate.rules).message = {required: true}];
```
mattklein123, line 91: This should use `TypedExtensionConfig` I think?

api/envoy/extensions/filters/http/geoip/v3/geoip.proto
```diff
@@ -0,0 +1,92 @@
...
+  // trusted client address from `x-forwarded-for` header.
+  // Otherwise, the immediate downstream connection source address will be used.
+  bool use_xff = 1;
+
+  // The number of additional ingress proxy hops from the right side of the
+  // :ref:`config_http_conn_man_headers_x-forwarded-for` HTTP header to trust when
+  // determining the origin client's IP address. The default is zero if this option
+  // is not specified. See the documentation for
+  // :ref:`config_http_conn_man_headers_x-forwarded-for` for more information.
+  uint32 xff_num_trusted_hops = 2;
```
mattklein123, line 85: I would refactor this into an XffConfig message, with the presence of the message indicating that XFF should be consulted, and the num_trusted_hops inside of it. This will be a bit clearer from an API perspective.

api/envoy/extensions/filters/http/geoip/v3/geoip.proto
```diff
@@ -0,0 +1,92 @@
...
+    // If set, the header will be used to populate the region ISO code associated with the IP address.
+    string region = 3
+        [(validate.rules).string = {well_known_regex: HTTP_HEADER_NAME ignore_empty: true}];
+
+    // If set, the header will be used to populate the ASN associated with the IP address.
+    string asn = 4
+        [(validate.rules).string = {well_known_regex: HTTP_HEADER_NAME ignore_empty: true}];
+
+    // If set, the IP address will be checked if it belongs to any type of anonymization network (e.g. VPN, public proxy etc)
+    // and header will be populated with the check result.
```
mattklein123, line 45: Can you specify the concrete values? Is it true/false? Something else? Same for all of the ones that aren't obviously strings.

docs/root/configuration/http/http_filters/geoip_filter.rst
```diff
@@ -0,0 +1,46 @@
...
+  name: envoy.filters.http.geoip
+  typed_config:
+    "@type": type.googleapis.com/envoy.extensions.filters.http.geoip.v3.Geoip
+    use_xff: true
+    xff_num_trusted_hops: 1
+    geo_headers_to_add:
+      country: "x-geo-country"
+      region: "x-geo-region"
+    provider:
+        name: "envoy.geoip_providers.maxmind"
```
mattklein123, line 45: This doesn't actually exist. Is this going to be added in some future PR?

mattklein123 (changes requested):
Thanks a few high level API comments. Can you also add a release note? Or do we want to wait on that until the maxmind provider is added per my other comment?

/wait

docs/root/configuration/http/http_filters/geoip_filter.rst
nezdolik: That's correct. I have initially suggested to split this work into implementing filter itself and then implementing the provider.

phlax:
not sure what is causing the docker fail - but ticket is here #26634 - ive reported upstream also https://github.com/moby/moby/issues/45344

nezdolik:
@mattklein123 regarding release note, i think is better to wait until provider is implemented.

mattklein123 (approved):
Thanks!

PR #24320, opened 2022-12-02
ci: Fix flaky coverage limit

phlax:
this can currently fail in builds that dont change anything

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim (commented):
+1 to unblock PRs but if we're at 96.0 already this just kicks the can very slightly down the road. Shouldn't we give more of a buffer like `95.0`?

jpsim (approved):

phlax:
> Shouldn't we give more of a buffer like 95.0

probably yes - altho there is an intention of constantly pushing it up - not sure what leeway we should add

cc @alyssawilk

jpsim:
I think Alyssa's out today so let's unblock at 96.0 and resume the discussion about how much buffer we should have next week.

alyssawilk:
if we're seeing flakes I think we improve coverage not drop the bar 1% :-(

jpsim:
> if we're seeing flakes I think we improve coverage not drop the bar 1% :-(

I don't think it was flaky, I think there were two PRs which individually were over 96.1 that when they were both merge to `main` went under 96.1.

alyssawilk:
yeah I wasn't responding to this PR, which was a fix to a deterministic failure, I was responding to @jmarantz suggesting we drop a full point.
Do we know which 2 PRs they were so we can ask for more  unit testing?

jpsim:
> I was responding to @jmarantz suggesting we drop a full point.

That was me in https://github.com/envoyproxy/envoy/pull/24320#pullrequestreview-1202759861

> Do we know which 2 PRs they were so we can ask for more unit testing?

I don't know which PRs reduced code coverage, but a motivated person could bisect `main` with code coverage with higher accuracy to see what changes increased vs decreased coverage.

Personally, I don't think it's realistic to say that no change can lower the code coverage amount, but that's my 2 cents.

alyssawilk:
oh we regularly allow tweaks due to refactors that say remove code (and thus lower coverage)
what I don't like about this is it lowers coverage for each individual directory rather than exempting whatever was causing the underlying problem, which means we permanently lower the bar for the whole project.  I'll try to throw together a change I like better :-)

Issue #24269, opened 2022-11-30
[gcp_authn] How to configure authn cluster


a1lu:
Hi,
I am trying to get the [GCP Authn Filter working](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/gcp_authn_filter) working with gRPC requests.
I copied the authn examples from the docs but I don't know how to configure the gcp_authn endpoint. The example uses `0.0.0.0:8000` like the fake service and envoy itself.

* In envoy logs I find this error related to gcp_authn
```
[2022-11-30 16:09:58.397][8][error][init] [source/extensions/filters/http/gcp_authn/gcp_authn_impl.cc:81] Response status is not OK, status: 503
```

* In the logs of the receiving service I can see the error 401 for my calls (as expected).
* I create a envoy container in google cloud run using the base image `envoyproxy/envoy-distroless:v1.23-latest`

My envoy.yaml:

```
static_resources:
  listeners:
  - name: listener_0
    address:
      socket_address: { address: 0.0.0.0, port_value: 80 }
    filter_chains:
    - filters:
      - name: envoy.filters.network.http_connection_manager
        typed_config:
          "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
          codec_type: auto
          stat_prefix: ingress_http
          route_config:
            name: local_route
            virtual_hosts:
            - name: local_service
              domains: ["*"]
              routes:
              - match: { prefix: "/" }
                route:
                  cluster: my_service
                  auto_host_rewrite: true
                  max_stream_duration:
                    grpc_timeout_header_max: 0s
          http_filters:
          - name: envoy.filters.http.gcp_authn
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.gcp_authn.v3.GcpAuthnFilterConfig
              http_uri:
                uri: "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=[AUDIENCE]"
                cluster: "gcp_authn"
                timeout:
                  seconds: 10
          - name: envoy.filters.http.router
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
  clusters:
  - name: my_service
    type: strict_dns
    connect_timeout: 20s
    typed_extension_protocol_options:
      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
[...]

PR #24328, opened 2022-12-02
gcp_authn: polish the gcp_authn doc

tyxia:
The original configuration in the doc is from the integration test with fake server, which might be a bit confusing. See #24269.
Thus, update the configuration with production use case.


docs/root/configuration/http/http_filters/_include/gcp-authn-filter-configuration.yaml
```diff
@@ -57,20 +57,19 @@ static_resources:
...
-  # Cluster for fake metadata server.
+  # Cluster for GCE metadata server
   - name: gcp_authn
-    load_assignment:
-      cluster_name: gcp_authn
+    type: STRICT_DNS
+    connect_timeout
+      seconds: 5
+    dns_lookup_family: V4_ONLY
+    load_assignment {
```
a1lu, line 66:
```suggestion
    load_assignment:
```

tyxia:
/assign @phlax

Could you please take a look at this PR? Thanks!

phlax (commented):
@tyxia iiuc this change to the docs is to use:

- an actual GCP authn server
- which is HTTP1 

from a docs pov the changes seem fine - im just a little apprehensive about a `google.internal` example address - is the only configurable address ?

a1lu:
@phlax The google docs state this metadata server as the only way to query identity tokens, if you cannot use the gcloud SDK.
See:
https://cloud.google.com/run/docs/securing/service-identity#identity-tokens
https://cloud.google.com/run/docs/container-contract#metadata-server

phlax:
thanks for explanation @a1lu - if this is the publicly documented way to configure this im fine with that

tyxia:
> @tyxia iiuc this change to the docs is to use:
> 
> * an actual GCP authn server
> * which is HTTP1
> 
> from a docs pov the changes seem fine - im just a little apprehensive about a `google.internal` example address - is the only configurable address ?
Currently yes, as stated in gcp_authn [doc](https://github.com/tyxia/envoy/blob/ed2559e74545ff2f3cee9ec690fc57b5a5798db5/docs/root/configuration/http/http_filters/gcp_authn_filter.rst#gcp-authentication-filter), this filter is used to fetch authentication tokens from [Google Compute Engine(GCE) metadata server](https://cloud.google.com/compute/docs/metadata/overview). In the future, i could imagine there will have other addresses for metadata server but still will be gcp environment.

phlax:
@tyxia seems like you will need a `main` merge - ci is a bit in flux atm - apologies for incovenience

Issue #24217, opened 2022-11-26
consider host as active when active health check passes with both active/passive HC are enabled


ramaraochavali:
When we have both active and passive health checks enabled, when host is ejected out of pool by passive health check failure, it will come back to pool only after ejection time is over, even though in between active health checks pass. Consider a case a where the ejection time is 2 min, the host wont be considered again for load balancing even if active health checks detect it healthy after say 30s.

I understand this is done to cover up for cases where active health check may not be going to actual server, Can this made configurable behaviour for services like Redis that wants 
- Early ejection based on passive health checks (with high ejection times)
- Come back to pool based on active health check success (instead of waiting for ejection timeout)?

Any existing config knobs that help this case or any down sides of doing this?

cc: @wbpcode @mattklein123 @snowp

mattklein123:
Seems reasonable to me.

ramaraochavali:
Should we have api config to drive this or runtime override is good enough?

ramaraochavali:
> I would probably just do it always with a runtime guard and see if anyone complains?

+ 1. This is what I was thinking as well. I will push a change.

bernardoVale:
Hi, I was reading the release notes and bumped into this new behavior.

If I understood correctly if an API endpoint starts failing on a specific host and gets ejected by the outlier but the health check remains healthy, Envoy will uneject the host on the next health check evaluation.

Isn't the example above one of the values of using an outlier detector?

ramaraochavali:
This change will not wait for the outlier detection time i.e. if a host is ejected because of outlier detection, it will be ejected for base_ejectoin time * iteration
- with out this change, envoy will wait for that duration before it brings it back to load balancer
- with this change, if a health check passes before that interval, it will consider host as healthy and brings it back.

PR #24331, opened 2022-12-03
prefer active health check during outlier ejection wait

ramaraochavali:
Commit Message:prefer active health check during outlier ejection wait
Additional Description:
Risk Level:Low
Testing:Added
Docs Changes:
Release Notes: Updated

Fixes https://github.com/envoyproxy/envoy/issues/24217


snowp (changes requested):
This seems right to me but is there any way to test this interacting with actual outlier detection? For example, are we going to hit any asserts in that code due to the flag being cleared outside of the outlier detection code?

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -294,6 +294,14 @@ void HealthCheckerImplBase::ActiveHealthCheckSession::handleSuccess(bool degrade
       // no longer be excluded.
       host_->healthFlagClear(Host::HealthFlag::EXCLUDED_VIA_IMMEDIATE_HC_FAIL);
       host_->healthFlagClear(Host::HealthFlag::FAILED_ACTIVE_HC);
+      // When the host is ejected by outlier detection and is waiting for outlier detection timer to
+      // bring it back to the pool after ejection period is over, If active health check succeeds,
+      // we should treat this host as healthy.
```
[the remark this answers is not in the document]
ramaraochavali: Sure. Will change

ramaraochavali:
>  but is there any way to test this interacting with actual outlier detection?

I validated the code and did not find any problem. However, if you have a test in mind let me know - I can add. Otherwise I will take a look at outlier detection tests and see if it is possible to add there.

snowp:
I think perhaps just an outlier detection test where you manually flip the flag out of band with outlier detection might be sufficient? I think that would simulate what happens if something else resets that flag

/wait-any

source/common/upstream/outlier_detection_impl.cc
```diff
@@ -282,29 +282,41 @@ DetectorImpl::~DetectorImpl() {
...
   }
+
+  if (cluster.healthChecker() != nullptr) {
+    cluster.healthChecker()->addHostCheckCompleteCb(
+        [this](HostSharedPtr host, HealthTransition changed_state) {
+          if (Runtime::runtimeFeatureEnabled(
+                  "envoy.reloadable_features.successful_active_health_check_uneject_host") &&
+              changed_state == HealthTransition::Changed &&
+              host->coarseHealth() == Host::Health::Healthy) {
+            unejectHost(host);
```
snowp, line 316: Seems like a consequence of this is that not only do we uneject the host on hc success, we also reset any outlier detector counters etc on a successful hc (when the host was previously unhealthy)? This seems in line with considering active hc as an authoritative check, but seems like a bigger change than I first anticipated. We probably want this behavior documented in the outlier detention docs and reflected in the change log

snowp (changes requested):
Thanks, a few more comments

test/common/upstream/outlier_detection_impl_test.cc
```diff
@@ -352,6 +352,73 @@ TEST_F(OutlierDetectorImplTest, BasicFlow5xxViaHttpCodes) {
...
+  loadRq(hosts_[0], 1, 500);
+  EXPECT_TRUE(hosts_[0]->healthFlagGet(Host::HealthFlag::FAILED_OUTLIER_CHECK));
+
+  EXPECT_EQ(1UL, outlier_detection_ejections_active_.value());
+
+  // Interval that doesn't bring the host back in.
+  time_system_.setMonotonicTime(std::chrono::milliseconds(9999));
+  EXPECT_CALL(*interval_timer_, enableTimer(std::chrono::milliseconds(10000), _));
+  interval_timer_->invokeCallback();
+  EXPECT_FALSE(hosts_[0]->outlierDetector().lastUnejectionTime());
```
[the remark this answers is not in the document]
ramaraochavali: No specific reason. Can remove it

source/common/upstream/outlier_detection_impl.cc
ramaraochavali: Makes sense. Updated

ramaraochavali:
@snowp Can you PTAL when you get chance?

ramaraochavali:
@snowp PTAL when you get chance

source/common/upstream/outlier_detection_impl.cc
snowp, line 316: Will this cause a regular transition from healthy -> unhealthy via active hc cause the outlier detector to count this as an unejection causing the stats to get messed up? Should this only respond to health transitions that clear the outlier check flag?

In general this approach of indirectly responding to a health change via the hc callback seems a bit brittle, though I'm not sure what the better approach would be. Maybe @mattklein123 has any ideas?

ramaraochavali: > Will this cause a regular transition from healthy -> unhealthy via active hc cause the outlier detector to count this as an unejection causing the stats to get messed up? Should this only respond to health transitions that clear the outlier check flag?

I do not think. But we can restrict when only respond to health transitions that clear the outlier check flag to be 100 % safe. I will change it like that.

> In general this approach of indirectly responding to a health change via the hc callback seems a bit brittle, though I'm not sure what the better approach would be. Maybe @mattklein123 has any ideas?

I could not think of any better approach. Let us see if @mattklein123 has any ideas

ramaraochavali:
@snowp PTAL when you get chance

source/common/upstream/outlier_detection_impl.cc
```diff
@@ -282,29 +282,45 @@ DetectorImpl::~DetectorImpl() {
...
   }
+
+  if (cluster.healthChecker() != nullptr) {
+    cluster.healthChecker()->addHostCheckCompleteCb(
+        [this](HostSharedPtr host, HealthTransition changed_state) {
+          // If the host is ejected by outlier detection and active health check succeeds,
+          // we should treat this host as healthy.
+          if (Runtime::runtimeFeatureEnabled(
+                  "envoy.reloadable_features.successful_active_health_check_uneject_host") &&
+              changed_state == HealthTransition::Changed &&
```
snowp, line 316: I don't think we want to gate on this? Since the health checker doesn't care about FAILED_OUTLIER_CHECK it would only set HealthTransition::Changed if we were already failing active hc and are now succeeding

I guess `!host->healthFlagGet(Host::HealthFlag::FAILED_ACTIVE_HC) && host->healthFlagGet(Host::HealthFlag::FAILED_OUTLIER_CHECK)`  would be sufficient to say "we got a hc indicating that active hc is good, but the outlier detector has flagged this host so we should clear that out"? 

I think ideally we'd have an integration test operating between the active hc and the outlier detector, seems a bit tricky to make sure that things are working as expected just through unit tests

ramaraochavali: >  I guess !host->healthFlagGet(Host::HealthFlag::FAILED_ACTIVE_HC) && host->healthFlagGet(Host::HealthFlag::FAILED_OUTLIER_CHECK) would be sufficient to say "we got a hc indicating that active hc is good, but the outlier detector has flagged this host so we should clear that out"?

But we want this behaviour of whether active health check success should clear of pending outlier detection failure should be guarded by run time flag in case some one has some issues with it. But we are defaulting it to true.

KBaichoo:
friendly ping @snowp 

Think this is waiting on your input

source/common/upstream/outlier_detection_impl.cc
snowp: Sorry yeah this should be gated by a feature flag, I was thinking about the `HealthTransition::Changed` check. I would think that we want to respond to every health check result from the active health checker and reset any ejected hosts regardless of whether we transitioned the active health of the host?

ramaraochavali: I see what you mean. Yes. That can be removed

ramaraochavali:
@snowp PTAL when you get chance

test/common/upstream/outlier_detection_impl_test.cc
```diff
@@ -352,6 +353,79 @@ TEST_F(OutlierDetectorImplTest, BasicFlow5xxViaHttpCodes) {
...
+  // Interval that doesn't bring the host back in.
+  time_system_.setMonotonicTime(std::chrono::milliseconds(9999));
+  EXPECT_CALL(*interval_timer_, enableTimer(std::chrono::milliseconds(10000), _));
+  interval_timer_->invokeCallback();
+
+  // Trigger the health checker callbacks and validate host is unejected.
+  EXPECT_CALL(checker_, check(hosts_[0]));
+  EXPECT_CALL(*event_logger_,
+              logUneject(std::static_pointer_cast<const HostDescription>(hosts_[0])));
+  health_checker->runCallbacks(hosts_[0], HealthTransition::Changed);
```
snowp, line 401: Can you do a test case where you use other transition values?

snowp (changes requested):
Sorry for the slow turnaround here, just one thing

test/common/upstream/outlier_detection_impl_test.cc
ramaraochavali: Done

snowp (approved):
LGTM, thanks!

@mattklein123 mind doing the second review?

mattklein123 (changes requested):
LGTM but we should wait to merge this till after the release (today) due to risk. Once the release is cut can you merge main which will require a release note fix? Thank you.

/wait

mattklein123 (approved):

mattklein123 (changes requested):
Sorry I think you have a merge issue on the release notes.

/wait

ramaraochavali:
Ah. Sorry. Did n't realize that. Fixed now.

mattklein123 (approved):

PR #24345, opened 2022-12-05
[1.22] ci: Bump ubuntu images and disable CodeQL

phlax:
Ubuntu 18.04 is going EOL in a few months so we need to bump these deps for any branches that will remain

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
@lizan @keith not sure what we can do about this - the ubuntu 18 image is EOL in march/april and neither azp nor github will continue providing that image

for azp its probably not such a big issue as we use the build container but codeql is failing due to glibc version issues

phlax:
pushing this through as its required and would be good to include in this release

PR #24347, opened 2022-12-05
ci: Fix change detection (part 3)

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -451,6 +451,19 @@ stages:
         GCS_ARTIFACT_BUCKET: $(GcsArtifactBucket)
       condition: always()
 
+  - job: complete
+    displayName: "linux_x64"
+    dependsOn: ["bazel", "coverage"]
+    pool:
+      vmImage: "ubuntu-20.04"
+    condition: |
+      and(
+        in(dependencies.bazel.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
+        in(dependencies.coverage.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'))
+    steps:
+    - bash: |
+        echo "checks complete"
+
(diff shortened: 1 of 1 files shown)


phlax:
@alyssawilk the idea with this Pr is that we remove all the requirements on checks from branch protection and replace them with just the `complete` job - that way the other jobs can either run or be skipped without failing the pipeline

.azure-pipelines/pipelines.yml
```diff
@@ -451,6 +451,19 @@ stages:
...
       condition: always()
 
+  - job: complete
+    displayName: "linux_x64"
+    dependsOn: ["bazel", "coverage"]
+    pool:
+      vmImage: "ubuntu-20.04"
+    condition: |
+      and(
+        in(dependencies.bazel.result, 'Succeeded', 'SucceededWithIssues', 'Skipped'),
```
alyssawilk, line 461: Can we comment about this job and how it works inline?  I would assume that SucceededWithIssues would be bad and not grouped with the other two.
/wait

phlax: i think we need `SucceededWithIssues` - not sure exactly what it does but this condition is functionally equivalent to using the `succeeded()` queue time var

that is my interpretation of:

`succeeded()` `==` `in(variables['Agent.JobStatus'], 'Succeeded', 'SucceededWithIssues')`

here https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#succeeded

phlax: comment added

alyssawilk (approved):

Issue #24353, opened 2022-12-05
restore coverage limits


alyssawilk:
Lowered in 
https://github.com/envoyproxy/envoy/pull/24320 to fix main

tyxia:
Hi @alyssawilk , I have sent out a PR for your review.

Looking at the CI result in this PR https://github.com/envoyproxy/envoy/pull/24351, especially the coverage report [here](https://dev.azure.com/cncf/envoy/_build/results?buildId=122519&view=logs&j=bbe4b42d-86e6-5e9c-8a0b-fea01d818a24&t=e00c5a13-c6dc-5e9a-6104-69976170e881), looks like the problem is `Code coverage for source/extensions/clusters/common is lower than limit of 96.6 (90.3)` rather than main coverage (`Overall coverage rate:
  lines......: 96.1% (121370 of 126334 lines)
  functions..: 76.1% (25767 of 33838 functions)
Code coverage 96.1 is good and higher than limit of 96.0`)

alyssawilk:
That has a green coverage result with the lower limits so to be clear while it's a step in the right direction it doesn't resolve the bug.

tyxia:
@alyssawilk  I think this issue can now be closed after #24355.   Could you help me to close it ?(Looks like I don't see a button for me to close it)
Thanks!

alyssawilk:
Sorry this was incorrectly closed: we fixed the explicit breakage but the filter which caused the breakage was still added without sufficient testing and that has not yet been fixed.  @tyxia can you please make sure it's addressed?
"source/extensions/filters/http/rate_limit_quota:94" # TODO(tyxia) WIP. Improve coverage as development continues

tyxia:
Sure. actually rate_limit_quota's current coverage is actually [94.8](https://storage.googleapis.com/envoy-pr/b46a61c/coverage/source/extensions/filters/http/rate_limit_quota/index.html) slightly higher than 94 listed there.

I thought the goal (why it was closed) was to bring the overall coverage back to 961, which is now 96.3
Is the goal to bring rate_limit_quota itself to 96.1 ?

alyssawilk:
the goal is to remove the exception by bringing coverage up to the bar of 96.6.

tyxia:
Got it. Thanks!

alyssawilk:
Marking as no stalebot, and ccing @krajshiva

krajshiva:
Thanks! will prioritize this cleanup.

[...]

PR #24355, opened 2022-12-05
coverage: Improve test coverage

tyxia:
Remove the boilerplate code that was added as TODO for future PRs and was not tested.
Also, restore the overall coverage back to 96.1%

This is a part of effort to address #24353


yanavlasov:
Can you adjust the coverage threshold value for the filter, now the untested code was removed?

/wait-any

alyssawilk:
not just the filter but overall coverage and close off that bug, if this is sufficient

RyanTheOptimist:
It seems like this leaves us teetering on the edge of 96.1. Why does source/extensions/filters/http/rate_limit_quota have such an incredibly low coverage limit? Looking at coverage output, it seems like it would be pretty trivial to write a unit test for XRateLimitHeaderUtils::convertRateLimitUnit and RateLimitFilterConfig::createRouteSpecificFilterConfigTyped in order to get the coverage for this directory up to normal.

https://storage.googleapis.com/envoy-pr/9cb3b98/coverage/source/extensions/filters/http/ratelimit/ratelimit_headers.cc.gcov.html
https://storage.googleapis.com/envoy-pr/9cb3b98/coverage/source/extensions/filters/http/ratelimit/config.cc.gcov.html

tyxia:
> It seems like this leaves us teetering on the edge of 96.1. Why does source/extensions/filters/http/rate_limit_quota have such an incredibly low coverage limit? Looking at coverage output, it seems like it would be pretty trivial to write a unit test for XRateLimitHeaderUtils::convertRateLimitUnit and RateLimitFilterConfig::createRouteSpecificFilterConfigTyped in order to get the coverage for this directory up to normal.
> 
> https://storage.googleapis.com/envoy-pr/9cb3b98/coverage/source/extensions/filters/http/ratelimit/ratelimit_headers.cc.gcov.html https://storage.googleapis.com/envoy-pr/9cb3b98/coverage/source/extensions/filters/http/ratelimit/config.cc.gcov.html

rate_limit_qutoa is a WIP that only partially implemented/merged due to large amount of code it requires and the spirit of small CLs.  But these already merged code is needed so that I can continue the development as Github doesn't really support chained CLs.  Since it is unfinished work, it has some boilerplate code required by compilation and for my future PRs but not tested.

The links you mentioned are for `rateLimit` which has nothing to do with `rate_limit_quota` here. The coverage for `rate_limit_quota` was at 86.2%. I added some more so it should be over 90% (Also some of line are coverage tool's false positive )

For the WIP, I feel it might be more worthy to continue the next phase development so that the code will be covered in next PRs.
Please let me know your thoughts

RyanTheOptimist:
> rate_limit_qutoa is a WIP that only partially implemented/merged due to large amount of code it requires and the spirit of small CLs. But these already merged code is needed so that I can continue the development as Github doesn't really support chained CLs. Since it is unfinished work, it has some boilerplate code required by compilation and for my future PRs but not tested.
> 
> The links you mentioned are for `rateLimit` which has nothing to do with `rate_limit_quota` here.  The coverage for `rate_limit_quota` was at 86.2%. I added some more so it should be over 90% (Also some of line are coverage tool's false positive )

Oh, sheesh. Yes, you're right. Is the correct [coverage for this PR is here](https://storage.googleapis.com/envoy-pr/9cb3b98/coverage/source/extensions/filters/http/rate_limit_quota/index.html)? That seems to show 86.2% not 90%. Is that what you expect? In any case, can you update test/per_file_coverage.sh to the correct currently value (either 86.2% or 90% or whatever it is)?

> For the WIP, I feel it might be more worthy to continue the next phase development so that the code will be covered in next PRs. Please let me know your thoughts

If we have a directory that is below the limit, we can easily lower the threshold for that directory in test/per_file_coverage.sh, as we've done. But the problem with that is that this counts against the *global* project coverage limits. So by keeping rate_limit_quota below the limit, we're making it more likely that some other PR might cause us to drop down below the *global* limit. This causes pain for the entire project. Can we remove the untested methods from the tree until we're ready to use them (in follow up PRs)?

tyxia:
Hi @RyanTheOptimist, I updated the test coverage, `rate_limit_quota` is now at 95% https://storage.googleapis.com/envoy-pr/761a7cc/coverage/index.html. I updated the limit to 94% in case there are any flaky issues.
The only lines left are dead code caused by exhaustive switch statement. I will think about updating this in next PR.

PTAL. Thanks!

RyanTheOptimist (dismissed):
Great! Thanks for doing this.

tyxia:
The failed coverage is caused by `Code coverage for source/common/common/posix is lower than limit of 94.5 (92.7)`, which was removed in #24358. Thus, I think this is not ready to be removed? I added it back in this PR.

tyxia:
I figured that in order to unblock the presubmit test faster, I probably should have a separate PR to fix posix test coverage rather than include it in this PR which has many other things. So I send out #24463

RyanTheOptimist:
Hm. Even though #24463 has landed, this PR still includes a change to source/common/common/posix. Is that expected?

RyanTheOptimist (approved):

yanavlasov (approved):

alyssawilk:
Unfortunately  I'm not quite sure the coverage results worked: 
Checking per-extension coverage
/source/test/run_envoy_bazel_coverage.sh: line 121: ./test/per_file_coverage.sh: Permission denied
Per-extension coverage passed.
I don't think it was your PR but I'm not sure if the number here is correct as CI functionally didn't run in at least one pass.  I'm fixing in https://github.com/envoyproxy/envoy/pull/24529 and we'll see how it goes

phlax:
for ref the permissions were changed in 614d03aa9b96d492271eda2a670390996e466179

alyssawilk:
thankfully all good, coverage runs restored cleanly :-)

PR #24358, opened 2022-12-05
test: updating coverage numbers

alyssawilk:
Fixes https://github.com/envoyproxy/envoy/issues/11987
Fixes https://github.com/envoyproxy/envoy/issues/12002
Fixes https://github.com/envoyproxy/envoy/issues/11991


test/per_file_coverage.sh
```diff
@@ -4,85 +4,61 @@
...
+"source/common/secret:95.0"
+"source/common/signal:87.2" # Death tests don't report LCOV
 "source/common/singleton:95.7"
-"source/common/stats:96.5"
-"source/common/tcp:90.6"
+"source/common/tcp:90.8"
 "source/common/thread:0.0" # Death tests don't report LCOV
-"source/common/tracing:96.1"
-"source/common/upstream:96.0"
+"source/common/tracing:96.14"
```
RyanTheOptimist, line 26: nit: This is higher precision than we use elsewhere. Did you mean 96.4, by any chance?

RyanTheOptimist (commented):
Impressive! Thanks for doing this. I know it's still a draft, but one tiny comment.

RyanTheOptimist (approved):

test/per_file_coverage.sh
alyssawilk: ugh I think it's supposed to be 91.4 but hey it passes and isn't worse?

PR #24361, opened 2022-12-05
mobile: check for pending exceptions after JNI call

Augustyniak:
Commit Message: Check for pending exceptions after some of the Java calls from within the JNI layer. At Lyft, we noticed that we have a bunch of crashes related to us not checking for exceptions after on_request_headers or/and on_response_headers methods.
Additional description: I tried to use `ENVOY_LOG_EVENT_TO_LOGGER` to log when we detect a pending exception but ran into issues with deadlocks. Got rid of the extra logging to unblock my change.
Risk Level: Low, mostly additive change. New code executes for when the library crashes previously.
Testing: Integration test
Docs Changes:
Release Notes:
Platform Specific Features:

### DIFF
--- a/mobile/library/common/jni/jni_interface.cc
+++ b/mobile/library/common/jni/jni_interface.cc
@@ -279,8 +279,8 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_recordHistogramValue(JNIEnv* en
 
 // JvmCallbackContext
 
-static void pass_headers(const char* method, const Envoy::Types::ManagedEnvoyHeaders& headers,
-                         jobject j_context) {
+static void passHeaders(const char* method, const Envoy::Types::ManagedEnvoyHeaders& headers,
+                        jobject j_context) {
   JNIEnv* env = get_env();
   jclass jcls_JvmCallbackContext = env->GetObjectClass(j_context);
   jmethodID jmid_passHeader = env->GetMethodID(jcls_JvmCallbackContext, method, "([B[BZ)V");
@@ -319,7 +319,7 @@ static void* jvm_on_headers(const char* method, const Envoy::Types::ManagedEnvoy
   jni_log("[Envoy]", "jvm_on_headers");
   JNIEnv* env = get_env();
   jobject j_context = static_cast<jobject>(context);
-  pass_headers("passHeader", headers, j_context);
+  passHeaders("passHeader", headers, j_context);
 
   jclass jcls_JvmCallbackContext = env->GetObjectClass(j_context);
   jmethodID jmid_onHeaders =
@@ -514,7 +514,7 @@ static void* jvm_on_trailers(const char* method, envoy_headers trailers,
 
   JNIEnv* env = get_env();
   jobject j_context = static_cast<jobject>(context);
-  pass_headers("passHeader", trailers, j_context);
+  passHeaders("passHeader", trailers, j_context);
 
   jclass jcls_JvmCallbackContext = env->GetObjectClass(j_context);
   jmethodID jmid_onTrailers =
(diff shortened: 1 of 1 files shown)


mobile/library/common/jni/jni_utility.cc
```diff
@@ -66,6 +66,16 @@ void jni_delete_const_global_ref(const void* context) {
   jni_delete_global_ref(const_cast<void*>(context));
 }
 
+bool exception_check(JNIEnv* env) {
```
alyssawilk, line 69: I'd be inclined to ABSL_MUST_USE_RESULT for this.  or at least make it clear we're clearing

alyssawilk (commented):
cool, thanks for tackling this!
/wait

mobile/library/common/jni/jni_interface.cc
```diff
@@ -278,22 +279,23 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_recordHistogramValue(JNIEnv* en
 
 // JvmCallbackContext
 
-static void pass_headers(const char* method, envoy_headers headers, jobject j_context) {
+static void pass_headers(const char* method, const Envoy::Types::ManagedEnvoyHeaders& headers,
```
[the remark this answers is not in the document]
Augustyniak: >I thought these all had to be simple (C) types. can you clue me in on why it's OK use ManagedEnvoyHeaders?

the usage of c types is a self-imposed limitations in many places in our apps. Decided to use c++ type in here to make mem management simpler.

>edit: nm I think I see why. if this isn't called from Java layers can we use standard passHeaders naming to make that clear?
OK

mobile/library/common/jni/jni_utility.cc
Augustyniak: I do not think that it always have to be handled explicitly. Per ur suggestion, will change method's name to imply that it clears any pending exceptions

mobile/library/common/jni/jni_interface.cc
```diff
@@ -327,24 +328,50 @@ static void* jvm_on_headers(const char* method, envoy_headers headers, bool end_
...
+
+    env->DeleteLocalRef(jcls_object_array);
+    env->DeleteLocalRef(jcls_int);
+
+    return result;
+  } else {
+    env->DeleteLocalRef(j_stream_intel);
+    env->DeleteLocalRef(jcls_JvmCallbackContext);
+
+    return result;
```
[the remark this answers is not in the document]
Augustyniak: thank you @RyanTheOptimist  :)

mobile/library/common/jni/jni_interface.cc
```diff
@@ -306,19 +308,18 @@ static void pass_headers(const char* method, envoy_headers headers, jobject j_co
...
-  release_envoy_headers(headers);
 }
 
 // Platform callback implementation
 // These methods call jvm methods which means the local references created will not be
 // released automatically. Manual bookkeeping is required for these methods.
 
-static void* jvm_on_headers(const char* method, envoy_headers headers, bool end_stream,
-                            envoy_stream_intel stream_intel, void* context) {
+static void* jvm_on_headers(const char* method, const Envoy::Types::ManagedEnvoyHeaders& headers,
```
alyssawilk, line 317: this one isn't called by jvm directly either.  optional nit onHeadersFromJvm or something

alyssawilk (approved):

mobile/library/common/jni/jni_interface.cc
Augustyniak: it's calling a method from the java world tho which I think is why it has a `jvm` prefix (`pass_headers - now renamed to `passHeaders` - method did not have `jvm` prefix as it was/is not calling into `jvm`). 

I do not have a strong opinion in here but will keep the things they are until we figure out new naming patterns for this file - ideally in a separate PR.

PR #24363, opened 2022-12-05
mobile: allow using a light listener manager

alyssawilk:
This enables using Envoy Mobile from the real downstream listener manger to a lightweight listener manager, allowing compiling out all the downstream listener code.

The real listener is still by default. This should be largely a no-op for everything but the c++ CI.

Risk Level: low 
Testing: c++ ci regression tests new code paths.
Docs Changes: n/a
Release Notes: n/a

Part of https://github.com/envoyproxy/envoy-mobile/issues/2711


.github/workflows/android_tests.yml
```diff
@@ -110,5 +110,5 @@ jobs:
             --test_output=all \
             --build_tests_only \
             $([ -z $GITHUB_TOKEN ] || echo "--config=remote-ci-linux-clang") \
-            --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
+             --copt=-DALLOW_TCP_LISTENERS --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
```
jpsim, line 113: Can this be moved to a `.bazelrc` config?

envoy/server/options.h
```diff
@@ -266,6 +266,11 @@ class Options {
    * responsibility of the caller to handle the duplicates.
    */
   virtual const Stats::TagVector& statsTags() const PURE;
+
+  /**
+   * @return the type of listener manager to create.
+   */
+  virtual const std::string& listenerManager() const PURE;
```
jpsim, line 273: Can this return an enum instead of a string where the valid value space is ambiguous?

jpsim:
> The real listener is still used if Envoy Mobile is compiled with --copt=-DALLOW_TCP_LISTENERS as evidenced by the kotlin build.

Why does Kotlin need to build with the real listener?

jpsim:
> Release Notes: maybe should have for E-M?

We have release notes in https://github.com/envoyproxy/envoy/blob/main/mobile/docs/root/intro/version_history.rst

.github/workflows/android_tests.yml
alyssawilk: The goal here is to have E-M build without TCP listeners, but we exempt it for this CI due to the proxying tests.  will comment inline.

envoy/server/options.h
alyssawilk: We can, but I don't think Envoy should have to know the name of the E-M listener, and this allows other to do overrides, so I think the string is appropriate.

alyssawilk:
Also adding in Snow for the changes to Envoy.  Mind taking a look?  Very minimal PR.

mobile/library/common/extensions/listener_managers/api_listener_manager/api_listener_manager.h
```diff
@@ -0,0 +1,60 @@
...
+namespace Envoy {
+namespace Server {
+
+/**
+ * Implementation of a lightweight ListenerManager for Envoy Mobile.
+ * This does not handle downstream TCP / UDP connections but only the API listener.
+ */
+class ApiListenerManagerImpl : public ListenerManager, Logger::Loggable<Logger::Id::config> {
+public:
+  ApiListenerManagerImpl(Instance& server);
```
snowp, line 19: explicit

snowp (dismissed):
lgtm, just the one nit

.github/workflows/android_tests.yml
jpsim: But then these tests will fail when run locally?

Why not add `build test --copt=-DALLOW_TCP_LISTENERS` to the `.bazelrc` so it works both on CI and locally?

envoy/server/options.h
jpsim: ok

jpsim:
From what I can tell, this increases the binary size when building with `./bazelw build //test/performance:test_binary_size --config=sizeopt`.

* Before: 32870384
* After: 32891472

Augustyniak:
A discussion related to this PR happened in Slack. The current hypothesis is that https://github.com/envoyproxy/envoy/pull/24294 broke objc hello world app CI job which was disabled in https://github.com/envoyproxy/envoy/pull/24363 to 'unblock' main (make it green again). 

We think that your PR may fix objc hello world app issue but to test it we would need to re-enable the aforementioned job. Could we verify whether your PR fixes the objc hello world app CI job?

alyssawilk:
Ok CI reinstated and passing (needed a _different_ force-register)

mobile/envoy_build_config/BUILD
```diff
@@ -14,6 +14,7 @@ envoy_cc_library(
     deps = [
         "extension_registry_platform_additions",
         "@envoy//source/common/network:socket_lib",
+        "@envoy//source/common/quic:quic_transport_socket_factory_lib",
```
Augustyniak, line 17: This may be missing envoy_select_enable_http3 usage https://github.com/envoyproxy/envoy/pull/24496/files#diff-4517a4a52636c1c8e390c1d9653aa000a67fc1070d977ebc114ca9eacac2ec49R48

mobile/envoy_build_config/BUILD
```diff
@@ -46,7 +45,10 @@ envoy_cc_library(
         "@envoy_mobile//library/common/extensions/filters/http/socket_tag:config",
         "@envoy_mobile//library/common/extensions/listener_managers/api_listener_manager:api_listener_manager_lib",
         "@envoy_mobile//library/common/extensions/retry/options/network_configuration:config",
-    ],
+    ] + select({
+        "@envoy//bazel:disable_http3": [],
```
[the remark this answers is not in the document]
alyssawilk: it fails because it wants @envoy// in fron tof the bazel:disable_http3
looks like we don't directly use any of the base select macros

.github/workflows/android_tests.yml
alyssawilk: because then all tests will always run with TCP listeners, and we won't have regression testing for, say, the C++ tests running without TCP listeners.

What I'm hearing is you'd prefer I add it to the bazelrc, and explicitly not allow TCP listeners on all the non-legacy CI instead?

Augustyniak:
>Why does Kotlin need to build with the real listener?

It's needed by Android proxy tests which spam 2 instances of Envoy Mobile, one of which acts as a "fake" proxy and uses the default/complex listener to handle incoming requests.

mobile/envoy_build_config/extension_registry.cc
```diff
@@ -81,6 +82,7 @@ void ExtensionRegistry::registerFactories() {
   Envoy::Network::forceRegisterGetAddrInfoDnsResolverFactory();
   Envoy::Extensions::RequestId::forceRegisterUUIDRequestIDExtensionFactory();
   Envoy::Server::forceRegisterDefaultListenerManagerFactoryImpl();
+  Envoy::Server::forceRegisterApiListenerManagerFactoryImpl();
```
Augustyniak, line 85: should we force register in here only if `DUSE_API_LISTENER` is set?

mobile/envoy_build_config/BUILD
```diff
@@ -44,6 +44,7 @@ envoy_cc_library(
         "@envoy_mobile//library/common/extensions/filters/http/platform_bridge:config",
         "@envoy_mobile//library/common/extensions/filters/http/route_cache_reset:config",
         "@envoy_mobile//library/common/extensions/filters/http/socket_tag:config",
+        "@envoy_mobile//library/common/extensions/listener_managers/api_listener_manager:api_listener_manager_lib",
```
Augustyniak, line 47: similar to my other comment - should we include this dep only if the new compiler/build flag is set?

Augustyniak (approved):
Left a comment with regard to including the new api listener based on the value of the introduced flag. 

It's not a hard blocker but it's worth noting that the size of the binary will actually go up by a tiny bit until we do not remove the old listener and if we always include the new listener as part of the build. Based on JP's numbers for the size increase I am _not_ worried about the size increase tho (as it appears to be tiny).

mobile/envoy_build_config/extension_registry.cc
alyssawilk: I was going to do that but I think it requires a bazel flag and additional plumbing to be useful given the library include

mobile/envoy_build_config/BUILD
alyssawilk: I think it depends.  if the long term plan is to have a bazel flag to have it optional, we can add it now. if long term we want to say E-M is a client library not a proxy, I'm inclined to not do the busy-work which I'd then have to unwind, and accept the additional include in the interim (we can remove it in-housE)

alyssawilk: actually having thought a bit more I think it is sensible to have a flag but I would like to add it when we flip the default, so have one for "enable TCP listener" where right now we have compile support for "enable API listener"

.github/workflows/android_tests.yml
jpsim: Hmm, yeah I was thinking a config that we could pass when running these tests specifically, but then it doesn't make much difference if you have to pass a config or a `--copt` because either way `./bazelw test ...`  won't be sufficient.

I'm ok keeping this as-is then.

jpsim (approved):

PR #24365, opened 2022-12-05
Implement shadow streaming

paul-r-gall:

Commit Message: Implement streaming for shadow requests.
Additional Description: Shadow/mirror requests will now be initialized on decodeHeaders. The router filter will copy and dispatch incoming data and trailers to shadow streams as they arrive. In order to prevent OOMs on shadow streams, shadow requests can now also apply backpressure downstream. Part of the goal is to decouple shadow request buffer limits from retry buffer limits.
Risk Level: High
Testing: unit testing in router_test.cc, and substantial integration testing in shadow_policy_integration_test.cc
Docs Changes: None
Release Notes: None
Optional Runtime guard: envoy.reloadable_features.streaming_shadow -- off by default.


paul-r-gall:
windows release had a test timeout.

/retest

envoy/http/async_client.h
```diff
@@ -163,13 +170,52 @@ class AsyncClient {
      */
     virtual void reset() PURE;
 
+    /***
+     * Add destructor callback.
+     */
+    virtual void setDestructorCallback(StreamDestructorCallbacks callback) PURE;
```
alyssawilk, line 176: I think we can be a bit more clear in comments
"Register a destructor to be called on [stream deletion? client deletion? end of stream?]. The callback must persist beyond the lifetime of the stream or be unregistered via removeDestructorCallback.
Is it illegal to call set if something is already set?

envoy/http/async_client.h
```diff
@@ -163,13 +170,52 @@ class AsyncClient {
...
+
+    /***
+     * Remove previously set destructor callback.
+     */
+    virtual void removeDestructorCallback() PURE;
+
+    /***
+     * Add watermark callbacks.
+     */
+    virtual void setWatermarkCallbacks(DecoderFilterWatermarkCallbacks& callbacks) PURE;
```
alyssawilk, line 186: ditto.  there's a lot of callbacks so maybe include when it'll be called and what should be done.

envoy/http/async_client.h
```diff
@@ -163,13 +170,52 @@ class AsyncClient {
...
+  public:
+    /***
+     * Take ownership of data, and sends it to the underlying stream.
+     * @param data owned buffer to pass to upstream.
+     * @param end_stream whether to end the stream.
+     */
+    virtual void captureAndSendData(Buffer::InstancePtr&& data, bool end_stream) PURE;
+
+    /***
+     * Take ownership of data, and sends it to the underlying stream.
```
alyssawilk, line 213: data?

envoy/http/async_client.h
```diff
@@ -318,6 +364,10 @@ class AsyncClient {
       sampled_ = sampled;
       return *this;
     }
+    RequestOptions& setEndStream(bool end_stream) {
+      end_stream_immediately_ = end_stream;
```
alyssawilk, line 368: it's unintuitive to me that we have to latch end stream immediately rather than ending the stream immediately.  can we make more clear what'll happen and when?  Is this really just a boolean to indicate if the stream is headers only?

source/common/http/async_client_impl.h
```diff
@@ -466,25 +498,68 @@ class AsyncStreamImpl : public AsyncClient::Stream,
   friend class AsyncClientImplUnitTest;
 };
 
-class AsyncRequestImpl final : public AsyncClient::Request,
-                               AsyncStreamImpl,
-                               AsyncClient::StreamCallbacks {
+class AsyncRequestSharedImpl : public virtual AsyncClient::Request,
```
alyssawilk, line 501: sorry is this just splitting out a base class and having one subclass for old and one for new?
I assume that just doing the streaming vs not inline was uglier?
If we're going this route we should definitely comment what the structure is and add TODOs for clean up after runtime guard is gone.

alyssawilk (commented):
Very cool addition!  Started a going through but I think I'm going to throw this over to @KBaichoo for first pass.  Here's what I have so far

source/common/http/async_client_impl.h
paul-r-gall: Right so now there's two separate methods in AsyncClient: `send` which expects a full request, and returns something that can only be cancelled (AsyncClient::Request*), and `startRequest` which expects headers only, and returns a whole stream that can be interacted with (AsyncClient::OngoingRequest*). The implementations of these two have a lot in common, hence the shared base with a protected constructor.

envoy/http/async_client.h
```diff
@@ -163,13 +170,52 @@ class AsyncClient {
...
+   * An interface for ongoing asynchronous requests.
+   */
+  class OngoingRequest : public virtual Request, public virtual Stream {
+  public:
+    /***
+     * Take ownership of data, and sends it to the underlying stream.
+     * @param data owned buffer to pass to upstream.
+     * @param end_stream whether to end the stream.
+     */
+    virtual void captureAndSendData(Buffer::InstancePtr&& data, bool end_stream) PURE;
```
KBaichoo, line 210: There are other ways to consume data from a buffer, e.g. taking the underlying slices which might be a better approach.

envoy/http/async_client.h
```diff
@@ -349,6 +401,18 @@ class AsyncClient {
   virtual Request* send(RequestMessagePtr&& request, Callbacks& callbacks,
                         const RequestOptions& options) PURE;
 
+  /**
+   * Starts a new request asynchronously with the given headers.
```
KBaichoo, line 405: would be useful to document the difference of this approach vs start below for a caller

envoy/router/shadow_writer.h
```diff
@@ -27,6 +27,18 @@ class ShadowWriter {
    */
   virtual void shadow(const std::string& cluster, Http::RequestMessagePtr&& request,
                       const Http::AsyncClient::RequestOptions& options) PURE;
+
+  /**
+   * Initialize shadowing a request. Differs from the above in that additional
+   * data can be passed to the returned handle after the headers have been sent.
+   * @param cluster supplies the cluster name to shadow to.
+   * @param headers supplies the headers for initializing the shadow.
+   * @param timeout supplies the shadowed request timeout.
```
KBaichoo, line 36: It's not just a timeout, it's request options instead. Can you update this and the declaration above?

source/common/router/shadow_writer_impl.cc
```diff
@@ -38,5 +45,24 @@ void ShadowWriterImpl::shadow(const std::string& cluster, Http::RequestMessagePt
   thread_local_cluster->httpAsyncClient().send(std::move(request), *this, shadow_options);
 }
 
+Http::AsyncClient::OngoingRequest*
+ShadowWriterImpl::streamingShadow(const std::string& cluster, Http::RequestHeaderMapPtr&& headers,
+                                  const Http::AsyncClient::RequestOptions& options) {
+  const auto thread_local_cluster = cm_.getThreadLocalCluster(cluster);
```
KBaichoo, line 51: Not a blocker, but it seems like we can share a lot of the code between both streaming and shadow above e.g. when preparing the request, headers, etc. Only really differs how we use the async client

source/common/router/router.cc
```diff
@@ -738,6 +778,47 @@ void Filter::sendNoHealthyUpstreamResponse() {
                              StreamInfo::ResponseCodeDetails::get().NoHealthyUpstream);
 }
 
+/*
+ * Helper class to handle a known number of buffer copies.
+ */
+class ManyCopiedBuffer {
```
KBaichoo, line 784: Can this be placed with the other buffer code?

source/common/router/router.cc
```diff
@@ -738,6 +778,47 @@ void Filter::sendNoHealthyUpstreamResponse() {
...
+      : original_buffer_(original_buffer), num_copies_(num_copies) {}
+
+  // Get an unowned reference to the next copy of the buffer. The copy is owned by this class.
+  Buffer::Instance& nextBuffer() {
+    ASSERT(num_copies_ > -1);
+    if (num_copies_ == 0) {
+      --num_copies_;
+      return original_buffer_;
+    } else {
+      copy_ = Buffer::OwnedImpl(original_buffer_);
```
KBaichoo, line 796: we're losing information such as any accounts the original buffer had, we should preserve it (at least for the shadow client use case).

It might make more sense to have the shadow stream itself wired with this.

source/common/router/router.cc
```diff
@@ -694,6 +704,36 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
       *this, std::move(generic_conn_pool), can_send_early_data, can_use_http3);
   LinkedList::moveIntoList(std::move(upstream_request), upstream_requests_);
   upstream_requests_.front()->acceptHeadersFromRouter(end_stream);
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.streaming_shadow")) {
+    // start the shadow streams.
+    for (const auto& shadow_policy_wrapper : active_shadow_policies_) {
+      const auto& shadow_policy = shadow_policy_wrapper.get();
+      const absl::optional<absl::string_view> cluster_name =
```
KBaichoo, line 711: nit: s/cluster_name/shadow_cluster_name/

source/common/router/router.cc
```diff
@@ -694,6 +704,36 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
       *this, std::move(generic_conn_pool), can_send_early_data, can_use_http3);
   LinkedList::moveIntoList(std::move(upstream_request), upstream_requests_);
   upstream_requests_.front()->acceptHeadersFromRouter(end_stream);
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.streaming_shadow")) {
+    // start the shadow streams.
+    for (const auto& shadow_policy_wrapper : active_shadow_policies_) {
```
KBaichoo, line 709: Seems like we clobber `shadow_headers_` if we have multiple policies to just be based on a single shadow header from one policy?

https://github.com/envoyproxy/envoy/pull/24365/files#diff-e3584a650478334f7373c0df344c85d499b5123f5237ce2bbabaad5ee53b899eR683

source/common/router/router.cc
```diff
@@ -694,6 +704,36 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
       *this, std::move(generic_conn_pool), can_send_early_data, can_use_http3);
   LinkedList::moveIntoList(std::move(upstream_request), upstream_requests_);
   upstream_requests_.front()->acceptHeadersFromRouter(end_stream);
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.streaming_shadow")) {
```
KBaichoo, line 707: might make sense to latch this for a given FC e.g. for the router, what happens if it gets flipped halfway through a stream?

KBaichoo (commented):
Thanks for working on this. Here's a first pass.

envoy/http/async_client.h
paul-r-gall: Could you clarify? We need to copy the original buffer into each of the shadows, and ownership needs to be retained by the new request.

source/common/router/router.cc
paul-r-gall: You have to make one copy per shadow, otherwise you get at least 2 problems: 
1. Ownership issues (shadow stream may outlive the main stream)
2. Upstream filters for one shadow may make header modifications that are undesired for other shadows.

source/common/router/router.cc
paul-r-gall: Reasonable, but FWIW the existing shadowing approach doesn't do any accounting either. Would you be ok if we postponed that to a future PR?

envoy/http/async_client.h
paul-r-gall: Clarified semantics.

envoy/http/async_client.h
paul-r-gall: Clarified semantics.

source/common/router/router.cc
paul-r-gall: Done; latched on router construction.

envoy/http/async_client.h
```diff
@@ -163,13 +170,52 @@ class AsyncClient {
...
+
     /***
      * @returns if the stream has enough buffered outbound data to be over the configured buffer
      * limits
      */
     virtual bool isAboveWriteBufferHighWatermark() const PURE;
   };
 
+  /***
+   * An interface for ongoing asynchronous requests.
```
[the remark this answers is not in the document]
paul-r-gall: Added clarifying comment.

envoy/http/async_client.h
paul-r-gall: Fixed -- removed the end stream field and documented that for a headers only request, the `send` method should be used instead of `startRequest`.

envoy/http/async_client.h
```diff
@@ -163,13 +168,57 @@ class AsyncClient {
...
+     */
+    virtual void setDestructorCallback(StreamDestructorCallbacks callback) PURE;
+
+    /***
+     * Remove previously set destructor callback.
+     */
+    virtual void removeDestructorCallback() PURE;
+
+    /***
+     * Register a callback to be called when high/low watermark events occur on the stream. This
```
alyssawilk, line 184: this is for upstream write buffer watermarks yeah?  maybe worth clarifying since watermarks are confusing AF to everyone, self included =P

envoy/http/async_client.h
```diff
@@ -163,13 +168,57 @@ class AsyncClient {
...
     /***
      * @returns if the stream has enough buffered outbound data to be over the configured buffer
      * limits
      */
     virtual bool isAboveWriteBufferHighWatermark() const PURE;
   };
 
+  /***
+   * An in-flight HTTP request to which additional data and trailers can be sent. Must be terminated
+   * by sending trailers or data with end_stream.
```
alyssawilk, line 204: I assume there's also a reset option?

envoy/http/async_client.h
alyssawilk: ping.  ownership of trailers I assume?  implicitly ends stream too.

envoy/http/filter.h
```diff
@@ -429,11 +429,32 @@ class StreamFilterCallbacks {
...
+
+  /**
+   * Called when the buffer for a decoder filter or any buffers the filter sends data to go over
+   * their high watermark.
+   *
+   * In the case of a filter such as the router filter, which spills into multiple buffers (codec,
+   * connection etc.) this may be called multiple times. Any such filter is responsible for calling
+   * the low watermark callbacks an equal number of times as the respective buffers are drained.
+   */
+  virtual void onDecoderFilterAboveWriteBufferHighWatermark() PURE;
```
alyssawilk, line 444: nit: mind a newline here?

source/common/http/async_client_impl.h
```diff
@@ -93,7 +98,22 @@ class AsyncStreamImpl : public AsyncClient::Stream,
...
                   const AsyncClient::StreamOptions& options);
-  ~AsyncStreamImpl() override { router_.onDestroy(); }
+  ~AsyncStreamImpl() override {
+    router_.onDestroy();
+    if (destructor_callback_.has_value()) {
+      (*destructor_callback_)();
+    }
+  }
+
+  void setDestructorCallback(AsyncClient::StreamDestructorCallbacks callback) override {
```
alyssawilk, line 108: optional: assert not already set  (and update comment to indicate disallowed)

source/common/http/async_client_impl.h
```diff
@@ -93,7 +98,22 @@ class AsyncStreamImpl : public AsyncClient::Stream,
...
+    destructor_callback_.emplace(callback);
+  }
+  void removeDestructorCallback() override { destructor_callback_.reset(); }
+
+  void setWatermarkCallbacks(DecoderFilterWatermarkCallbacks& callbacks) override {
```
alyssawilk, line 113: ditto

alyssawilk, line 113: can't remember - can we have async H2 clients, such that the downstream connection may be over the high watermark when a stream is created?  If so we need to check for high watermark and call high watermark callbacks N times when we set a new callback. and unit test.  optional bonus points for e2e.

source/common/http/async_client_impl.h
```diff
@@ -93,7 +98,22 @@ class AsyncStreamImpl : public AsyncClient::Stream,
...
+    watermark_callbacks_.emplace(callbacks);
+  }
+  void removeWatermarkCallbacks() override { watermark_callbacks_.reset(); }
```
alyssawilk, line 116: ditto here - if we have a shadow request from cluster A to cluster B, cluster B is slow and causes back-up so we pause reading from the client, shadow request is reset remotely, so we kill off that client, will we resume reading from the client and spooling to A?

alyssawilk (commented):
doing one drive by pass before I head out for winter break.  I'm out 'til the 3rd so if you get Kevin's LGTM you can try for Snow or Matt

test/common/http/async_client_impl_test.cc
```diff
@@ -203,6 +203,57 @@ TEST_F(AsyncClientImplTest, Basic) {
...
+  auto headers = std::make_unique<TestRequestHeaderMapImpl>();
+  HttpTestUtility::addDefaultHeaders(*headers);
+  TestRequestHeaderMapImpl headers_copy = *headers;
+
+  std::unique_ptr<Buffer::OwnedImpl> data = std::make_unique<Buffer::OwnedImpl>("test data");
+  Buffer::OwnedImpl data_copy("test data");
+
+  auto trailers = std::make_unique<TestRequestTrailerMapImpl>();
+  trailers->addCopy("some", "trailer");
+  TestRequestTrailerMapImpl trailers_copy = *trailers;
```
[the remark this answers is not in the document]
paul-r-gall: Added for some, but the headers_copy have extra stuff added to them.

source/common/router/router.cc
```diff
@@ -738,6 +786,47 @@ void Filter::sendNoHealthyUpstreamResponse() {
                              StreamInfo::ResponseCodeDetails::get().NoHealthyUpstream);
 }
 
+/*
+ * Helper class to handle a known number of buffer copies.
+ */
+class ManyCopiedBuffer {
+public:
+  ManyCopiedBuffer(Buffer::Instance& original_buffer, int num_copies)
+      : original_buffer_(original_buffer), num_copies_(num_copies) {}
```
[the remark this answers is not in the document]
paul-r-gall: Done -- made it so that you can call `next*Buffer()` `num_copies_` times.

source/common/router/router.cc
paul-r-gall: Done.

envoy/http/async_client.h
paul-r-gall: Done.

source/common/http/async_client_impl.h
paul-r-gall: Yes, this can happen, but the connection only will call high/low watermark callbacks once per stream.

source/common/http/async_client_impl.h
paul-r-gall: Done.

test/common/http/async_client_impl_test.cc
```diff
@@ -203,6 +203,57 @@ TEST_F(AsyncClientImplTest, Basic) {
...
+
+  EXPECT_CALL(stream_encoder_, encodeHeaders(HeaderMapEqualRef(&headers_copy), false));
+  EXPECT_CALL(stream_encoder_, encodeData(BufferEqual(&data_copy), false));
+  EXPECT_CALL(stream_encoder_, encodeTrailers(HeaderMapEqualRef(&trailers_copy)));
+
+  AsyncClient::OngoingRequest* request =
+      client_.startRequest(std::move(headers), callbacks_, AsyncClient::RequestOptions());
+  EXPECT_NE(request, nullptr);
+
+  request->captureAndSendData(std::move(data), false);
```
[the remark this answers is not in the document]
paul-r-gall: Done in the following watermark test.

test/integration/buffer_accounting_integration_test.cc
```diff
@@ -108,6 +110,11 @@ class Http2BufferWatermarksTest
 
     for (uint32_t idx = 0; idx < num_responses; ++idx) {
       responses.emplace_back(codec_client_->makeRequestWithBody(header_map, request_body_size));
+      absl::SleepFor(absl::Milliseconds(20));
```
KBaichoo, line 113: rm sleep? the waitForGaugeEq should have timeout you can set.

test/integration/buffer_accounting_integration_test.cc
```diff
@@ -358,6 +366,61 @@ TEST_P(Http2BufferWatermarksTest, ShouldTrackAllocatedBytesToUpstream) {
...
+        auto* mirror_policy = hcm.mutable_route_config()
+                                  ->mutable_virtual_hosts(0)
+                                  ->mutable_routes(0)
+                                  ->mutable_route()
+                                  ->add_request_mirror_policies();
+        mirror_policy->set_cluster("cluster_1");
+      });
+  initialize();
+
+  buffer_factory_->setExpectedAccountBalance(request_body_size, num_requests);
```
KBaichoo, line 396: should the account balance be 2x? Since it's many copied to the original and the mirror

source/common/http/async_client_impl.h
```diff
@@ -100,20 +100,37 @@ class AsyncStreamImpl : public virtual AsyncClient::Stream,
                   const AsyncClient::StreamOptions& options);
   ~AsyncStreamImpl() override {
     router_.onDestroy();
+    for (uint32_t i = 0; i < high_watermark_calls_; ++i) {
+      watermark_callbacks_->get().onDecoderFilterBelowWriteBufferLowWatermark();
```
KBaichoo, line 104: do we have a test covering this case?

KBaichoo (commented):
Thank you for adding this!

source/common/http/async_client_impl.h
KBaichoo: let's add `ENVOY_BUG` then?

I think it could happen if we watermarked, and got a close or reset from shadow upstream. Could you add a test for this scenario?

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
-INSTANTIATE_TEST_SUITE_P(IpVersions, ShadowPolicyIntegrationTest,
-                         testing::ValuesIn(TestEnvironment::getIpVersionsForTest()),
-                         TestUtility::ipTestParamsToString);
+INSTANTIATE_TEST_SUITE_P(
+    IpVersionsAndStreaming, ShadowPolicyIntegrationTest,
+    testing::Combine(testing::ValuesIn(TestEnvironment::getIpVersionsForTest()), testing::Bool()),
+    [](const ::testing::TestParamInfo<ShadowPolicyIntegrationTest::ParamType>& params) {
+      return absl::StrCat(std::get<0>(params.param) == Network::Address::IpVersion::v4 ? "IPv4"
+                                                                                       : "IPv6",
+                          "_", std::get<1>(params.param) ? "streaming_shadow" : "delayed_shadow");
```
KBaichoo, line 106: delayed_show -> buffered_shadow?

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+    });
+
+TEST_P(ShadowPolicyIntegrationTest, RequestMirrorPolicyWithDownstreamReset) {
+  if (!streaming_shadow_) {
+    GTEST_SKIP() << "Not applicable for non-streaming shadows.";
+  }
+  autonomous_upstream_ = false;
+  /*
+  TestScopedRuntime scoped_runtime;
+  scoped_runtime.mergeValues({{"envoy.reloadable_features.remember_shadow", "true"}});*/
```
KBaichoo, line 116: dead code rm

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+  EXPECT_EQ(test_server_->counter("cluster.cluster_0.upstream_rq_completed")->value(), 0);
+}
+
+TEST_P(ShadowPolicyIntegrationTest, RequestMirrorPolicyWithMainUpstreamReset) {
+  if (!streaming_shadow_) {
+    GTEST_SKIP() << "Not applicable for non-streaming shadows.";
+  }
+  autonomous_upstream_ = false;
+  /*
+  TestScopedRuntime scoped_runtime;
```
KBaichoo, line 172: dead code, rm

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+
+  codec_client_ = makeHttpConnection(lookupPort("http"));
+  Http::TestRequestHeaderMapImpl request_headers = default_request_headers_;
+  request_headers.addCopy("potato", "salad");
+  std::pair<Http::RequestEncoder&, IntegrationStreamDecoderPtr> result =
+      codec_client_->startRequest(request_headers, false);
+  auto response = std::move(result.second);
+
+  FakeHttpConnectionPtr fake_upstream_connection_0;
+  FakeStreamPtr upstream_request_0;
```
KBaichoo, line 185: nit: can these be named to say which is the shadow_request, main_request, etc.?

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+}
+
+TEST_P(ShadowPolicyIntegrationTest, RequestMirrorPolicyWithShadowUpstreamReset) {
+  if (!streaming_shadow_) {
+    GTEST_SKIP() << "Not applicable for non-streaming shadows.";
+  }
+  autonomous_upstream_ = false;
+  /*
+  TestScopedRuntime scoped_runtime;
+  scoped_runtime.mergeValues({{"envoy.reloadable_features.remember_shadow", "true"}});*/
```
KBaichoo, line 228: dead code rm

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+  FakeHttpConnectionPtr fake_upstream_connection_1;
+  FakeStreamPtr upstream_request_1;
+  ASSERT_TRUE(fake_upstreams_[1]->waitForHttpConnection(*dispatcher_, fake_upstream_connection_1));
+  ASSERT_TRUE(fake_upstream_connection_1->waitForNewStream(*dispatcher_, upstream_request_1));
+  ASSERT_TRUE(upstream_request_0->waitForHeadersComplete());
+  ASSERT_TRUE(upstream_request_1->waitForHeadersComplete());
+
+  // Send upstream reset on shadow request.
+  upstream_request_1->encodeResetStream();
+  ASSERT_TRUE(upstream_request_1->waitForReset());
```
KBaichoo, line 253: might make sense to wait for the Envoy to recv the  reset, e.g. by checking the stat reset for for shadow.

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+  EXPECT_EQ(test_server_->counter("cluster.cluster_1.upstream_rq_completed")->value(), 1);
+}
+
+TEST_P(ShadowPolicyIntegrationTest, RequestMirrorPolicyWithDownstreamTimeout) {
+  if (!streaming_shadow_) {
+    GTEST_SKIP() << "Not applicable for non-streaming shadows.";
+  }
+  autonomous_upstream_ = false;
+  /*
+  TestScopedRuntime scoped_runtime;
```
KBaichoo, line 285: rm deadcode

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,323 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+  EXPECT_EQ(test_server_->counter("cluster.cluster_1.upstream_rq_completed")->value(), 0);
+}
+
+TEST_P(ShadowPolicyIntegrationTest, RequestMirrorPolicyWithShadowBackpressure) {
+  if (!streaming_shadow_) {
+    GTEST_SKIP() << "Not applicable for non-streaming shadows.";
+  }
+  autonomous_upstream_ = false;
+  /*
+  TestScopedRuntime scoped_runtime;
```
KBaichoo, line 347: rm deadcode

KBaichoo (commented):
Thanks @paul-r-gall ! This generally looks great

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done

test/integration/shadow_policy_integration_test.cc
paul-r-gall: D

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/shadow_policy_integration_test.cc
paul-r-gall: Done.

test/integration/buffer_accounting_integration_test.cc
paul-r-gall: Done.

source/common/http/async_client_impl.h
KBaichoo: sorry for being unclear, since this can happen in the case discussed above, then it's not an envoy bug if we have high wm calls > 0.

test/integration/shadow_policy_integration_test.cc
```diff
@@ -85,15 +89,324 @@ class ShadowPolicyIntegrationTest : public testing::TestWithParam<Network::Addre
...
+  EXPECT_TRUE(response->complete());
+
+  cleanupUpstreamAndDownstream();
+
+  EXPECT_EQ(test_server_->counter("http.config_test.downstream_flow_control_paused_reading_total")
+                ->value(),
+            1);
+  EXPECT_EQ(test_server_->counter("cluster.cluster_0.upstream_cx_total")->value(), 1);
+  EXPECT_EQ(test_server_->counter("cluster.cluster_1.upstream_cx_total")->value(), 1);
+  // Main cluster saw no reset; shadow cluster saw remote reset.
```
KBaichoo, line 406: stale comment? copy-pasted?

source/common/router/router.cc
```diff
@@ -694,6 +704,46 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
...
+            Http::createHeaderMap<Http::RequestHeaderMapImpl>(*shadow_headers_)));
+        config_.shadowWriter().shadow(std::string(shadow_cluster_name.value()), std::move(request),
+                                      options);
+      } else {
+        Http::AsyncClient::OngoingRequest* shadow_stream = config_.shadowWriter().streamingShadow(
+            std::string(shadow_cluster_name.value()), std::move(shadow_headers), options);
+        if (shadow_stream != nullptr) {
+          shadow_streams_[shadow_stream] = end_stream;
+          shadow_stream->setDestructorCallback(
+              [this, shadow_stream]() { shadow_streams_.erase(shadow_stream); });
```
KBaichoo, line 741: what's the guarantee we have that we won't call this dtor under iterating this map? Can we have this as an assertion, or something programmatic?

KBaichoo (commented):
Few final questions on my end, thank you!

KBaichoo:
/assign @alyssawilk 

for senior maintainer pass

/wait

source/common/http/async_client_impl.h
alyssawilk: +1 to extensive watermark testing :-)
I would expect for shared H2 connections you could end up destroying in high watermark state

source/common/http/async_client_impl.h
```diff
@@ -466,25 +522,64 @@ class AsyncStreamImpl : public AsyncClient::Stream,
...
+public:
+  AsyncOngoingRequestImpl(RequestHeaderMapPtr&& request_headers, AsyncClientImpl& parent,
+                          AsyncClient::Callbacks& callbacks,
+                          const AsyncClient::RequestOptions& options)
+      : AsyncRequestSharedImpl(parent, callbacks, options),
+        request_headers_(std::move(request_headers)) {
+    ASSERT(request_headers_);
+  }
+  void captureAndSendData(Buffer::InstancePtr&& data, bool end_stream) override {
+    request_data_ = std::move(data);
```
alyssawilk, line 557: why move this to a class member variable?
if there's any expectations of persistence, another send may overwrite.

source/common/router/router.cc
```diff
@@ -282,6 +282,16 @@ Filter::~Filter() {
   // Upstream resources should already have been cleaned.
   ASSERT(upstream_requests_.empty());
   ASSERT(!retry_state_);
+
+  // Unregister from shadow stream notifications.
+  for (auto& [shadow_stream, ended] : shadow_streams_) {
+    if (!ended) {
+      // we have not yet sent an end stream to the shadows; cancel them.
```
alyssawilk, line 289: nit: We

source/common/router/router.cc
```diff
@@ -694,6 +704,46 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
...
+        // without waiting.
+        Http::RequestMessagePtr request(new Http::RequestMessageImpl(
+            Http::createHeaderMap<Http::RequestHeaderMapImpl>(*shadow_headers_)));
+        config_.shadowWriter().shadow(std::string(shadow_cluster_name.value()), std::move(request),
+                                      options);
+      } else {
+        Http::AsyncClient::OngoingRequest* shadow_stream = config_.shadowWriter().streamingShadow(
+            std::string(shadow_cluster_name.value()), std::move(shadow_headers), options);
+        if (shadow_stream != nullptr) {
+          shadow_streams_[shadow_stream] = end_stream;
```
alyssawilk, line 739: isn't this in an end_stream = false branch?  why set this?
or rather why keep them around once the stream is ended rather than remove when complete?

source/common/router/shadow_writer_impl.cc
alyssawilk: yeah would it work to have the cluster and host setting in a shared function perhaps?

alyssawilk (commented):
very cool. not looking at tests yet
/wait

envoy/router/shadow_writer.h
```diff
@@ -23,10 +23,22 @@ class ShadowWriter {
...
   virtual void shadow(const std::string& cluster, Http::RequestMessagePtr&& request,
                       const Http::AsyncClient::RequestOptions& options) PURE;
+
+  /**
+   * Initialize shadowing a request. Differs from the above in that additional
+   * data can be passed to the returned handle after the headers have been sent.
+   * @param cluster supplies the cluster name to shadow to.
+   * @param headers supplies the headers for initializing the shadow.
+   * @param options supplies the request options for the underlying asynchronous request.
+   * @return OngoingRequest* pointer which can have additional data and trailers sent to it.
```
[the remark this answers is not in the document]
paul-r-gall: I was just mirroring the methods in AsyncClient, which all return pointers.

source/common/router/router.cc
```diff
@@ -777,22 +828,23 @@ Http::FilterDataStatus Filter::decodeData(Buffer::Instance& data, bool end_strea
...
-    // If we are going to buffer for retries or shadowing, we need to make a copy before encoding
-    // since it's all moves from here on.
-    if (!upstream_requests_.empty()) {
-      Buffer::OwnedImpl copy(data);
-      upstream_requests_.front()->acceptDataFromRouter(copy, end_stream);
+    callbacks_->addDecodedData(many_copied_buffer.nextBuffer(), true);
+  }
+  if (streaming_shadows_ && !shadow_streams_.empty()) {
+    for (auto& [shadow_stream, ended] : shadow_streams_) {
+      shadow_stream->captureAndSendData(many_copied_buffer.nextBufferOwned(), end_stream);
```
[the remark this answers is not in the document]
paul-r-gall: Intermediate class is just to make sure that the final time we need the data, we actually drain the original buffer.

source/common/http/async_client_impl.h
```diff
@@ -411,15 +453,27 @@ class AsyncStreamImpl : public AsyncClient::Stream,
   void encodeTrailers(ResponseTrailerMapPtr&& trailers) override;
   ResponseTrailerMapOptRef responseTrailers() const override { return {}; }
   void encodeMetadata(MetadataMapPtr&&) override {}
-  void onDecoderFilterAboveWriteBufferHighWatermark() override { ++high_watermark_calls_; }
+  void onDecoderFilterAboveWriteBufferHighWatermark() override {
+    ++high_watermark_calls_;
```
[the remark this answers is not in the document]
paul-r-gall: Yes -- in UpstreamRequest upstream_flow_control_backed_up_total_ is incremented (and decremented).

source/common/http/async_client_impl.h
```diff
@@ -411,15 +453,27 @@ class AsyncStreamImpl : public AsyncClient::Stream,
...
     --high_watermark_calls_;
+    if (watermark_callbacks_.has_value()) {
+      watermark_callbacks_->get().onDecoderFilterBelowWriteBufferLowWatermark();
+    }
   }
   void addDownstreamWatermarkCallbacks(DownstreamWatermarkCallbacks&) override {}
   void removeDownstreamWatermarkCallbacks(DownstreamWatermarkCallbacks&) override {}
-  void setDecoderBufferLimit(uint32_t) override {}
-  uint32_t decoderBufferLimit() override { return 0; }
+  void setDecoderBufferLimit(uint32_t limit) override {
```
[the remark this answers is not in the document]
paul-r-gall: The goal here is to have exactly the same behavior as before if no buffer limit was specified in options.

source/common/router/router.cc
paul-r-gall: That makes sense, fixed.

source/common/http/async_client_impl.h
paul-r-gall: Fixed.

source/common/http/async_client_impl.h
```diff
@@ -93,7 +98,40 @@ class AsyncStreamImpl : public AsyncClient::Stream,
...
+    }
+    if (destructor_callback_.has_value()) {
+      (*destructor_callback_)();
+    }
+  }
+
+  void setDestructorCallback(AsyncClient::StreamDestructorCallbacks callback) override {
+    ASSERT(!destructor_callback_);
+    destructor_callback_.emplace(callback);
+  }
```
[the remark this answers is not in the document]
paul-r-gall: Done

envoy/http/async_client.h
```diff
@@ -258,6 +320,10 @@ class AsyncClient {
 
     envoy::config::core::v3::Metadata metadata;
 
+    Buffer::BufferMemoryAccountSharedPtr account_{nullptr};
```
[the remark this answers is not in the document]
paul-r-gall: HTTP Metadata is ignored.

source/common/router/shadow_writer_impl.cc
paul-r-gall: Done.

envoy/router/shadow_writer.h
```diff
@@ -34,7 +34,8 @@ class ShadowWriter {
    * @param cluster supplies the cluster name to shadow to.
    * @param headers supplies the headers for initializing the shadow.
    * @param options supplies the request options for the underlying asynchronous request.
-   * @return OngoingRequest* pointer which can have additional data and trailers sent to it.
+   * @return OngoingRequest* pointer owned by the AsyncClient which can have additional data and
```
alyssawilk, line 37: fair point on consistency re pointer.  can it be null?  let' comment either way.

source/common/http/async_client_impl.h
alyssawilk: if it's not possible then remove the code handling it and leave it an envoy bug?  bonus points for a comment explaining it'll be done in clean up

source/common/http/async_client_impl.h
```diff
@@ -93,7 +98,40 @@ class AsyncStreamImpl : public AsyncClient::Stream,
 public:
   AsyncStreamImpl(AsyncClientImpl& parent, AsyncClient::StreamCallbacks& callbacks,
                   const AsyncClient::StreamOptions& options);
-  ~AsyncStreamImpl() override { router_.onDestroy(); }
+  ~AsyncStreamImpl() override {
+    router_.onDestroy();
+    ENVOY_BUG(high_watermark_calls_ == 0, "Excess high watermark calls after async stream ended.");
+    for (uint32_t i = 0; i < high_watermark_calls_; ++i) {
+      watermark_callbacks_->get().onDecoderFilterBelowWriteBufferLowWatermark();
```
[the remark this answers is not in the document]
alyssawilk: ping?

source/common/router/router.cc
```diff
@@ -841,9 +838,15 @@ Http::FilterDataStatus Filter::decodeData(Buffer::Instance& data, bool end_strea
...
-      ended = end_stream;
+    for (auto* shadow_stream : shadow_streams_) {
+      if (end_stream) {
+        shadow_stream->removeDestructorCallback();
+        shadow_stream->removeWatermarkCallbacks();
+      }
+      shadow_stream->sendData(many_copied_buffer.nextBuffer(), end_stream);
+    }
+    if (end_stream) {
+      shadow_streams_.clear();
```
alyssawilk, line 849: So having told you to do this I have another question I want to ask because I'm second guessing =P
if we haven't spooled all the data to a shadow stream and the client disconnects, do we want to cancel, or not?  I'm inclined to say it'd better to be consistent (send all the data read, which is what we do now) but I wanted to check
If we leave as-is let's at least comment here that clearing the shadow streams doesn't indicate they're done - they're self owning and will spool data to the upstream on their own, we just stop the assocation.

Also sanity check, if they're disassociated and the upstream stops reading, will they eventually time out?

source/common/router/router.h
```diff
@@ -657,8 +657,8 @@ class Filter : Logger::Loggable<Logger::Id::router>,
 
   Network::TransportSocketOptionsConstSharedPtr transport_socket_options_;
   Network::Socket::OptionsSharedPtr upstream_options_;
-  // map from shadow stream to whether or not we've sent end stream to it.
-  absl::flat_hash_map<Http::AsyncClient::OngoingRequest*, bool> shadow_streams_;
+  // Set of active shadow stream.
```
alyssawilk, line 660: set of ongoing shadow streams which have not yet received end stream?

alyssawilk (commented):
so now we have streaming shadows worth testing!

timeouts (make sure shadow requests will eventually die in all cases) ideally at least 1 e2e test
disconnects of upstream main request and of shadowed request during spooled data (e2e) 
disconnect for shadow request after the main request is disassociated (unit test?)
internal redirects  (lack thereof)
upstream main filters going over the configured buffer limits and so closing the connection while processing data and upstream shadown filters going over configured buffer limits (crib from HittingDecoderFilterLimit)
definitely want to make sure we have at least unit tests of adding a new stream when the connection is above high watermark, as well as removing a shadow stream and ensuring the connection resumes (which you probably have but I haven't dug out)
Also I'd encourage explicitly testing filters which add head and body data to ensure the two are distinct (there should be code you can lift in the filter integration tests)

source/common/router/router.cc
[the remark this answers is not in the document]
alyssawilk: sorry I'm probably missing something but if we have an original buffer (which normally just goes upstream) and we need to copy that buffer for each shadow stream can't we just do a copy for each shadow, then hand the original upstream?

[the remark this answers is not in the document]
paul-r-gall: I want the main stream to take priority, so I want the data sent to the main stream first. I could do it in the opposite order, but that seems unintuitive.

source/common/http/async_client_impl.h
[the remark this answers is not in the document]
paul-r-gall: We could always stream infinite data via the AsyncStream API -- just keep calling `sendData`. I could set the default to be kBufferLimitForRetry (currently 64KiB). Another option (which I maybe like more) is I could pipe through the `retry_shadow_buffer_limit_` from the main Router filter to here instead of using the downstream filter chain `decoderBufferLimit`.

source/common/router/router.cc
paul-r-gall: Not totally sure what you mean -- these are the scenarios I can think of:
1. We see `end_stream`, and then the client disconnects before the main upstream has properly sent a response. In this case, the shadow streams are already disassociated from the main stream, and they'll continue on their way, respecting the timeouts associated with the shadow cluster.
2. The client disconnects before sending `end_stream`. Here, we'll (eventually) destroy the router filter, which will cancel all the shadow streams.

paul-r-gall:
> removing a shadow stream and ensuring the connection resumes

@alyssawilk do you mean a scenario where the shadow stream is blocked, and then the shadow stream times out, thereby allowing the main stream to continue?

paul-r-gall:
I believe that header/body addition is covered by the pre-existing test `ClusterFilterOverridesRouterFilter` in shadow_policy_integration_test.cc.

source/common/http/async_client_impl.h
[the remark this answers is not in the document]
alyssawilk: but previously didn't we only do shadow streams if we were willing to buffer the whole body?  So there was an implicit limit where now there may not be?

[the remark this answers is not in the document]
paul-r-gall: right; my question is "What should that limit be?".

source/common/router/router.cc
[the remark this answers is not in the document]
mattklein123: I guess my vote would be to keep it simple as Alyssa mentioned and then if we really do so perf issues we can revisit? In the grand scheme of things I'm doubtful the ordering will have a material impact on overall performance.

KBaichoo:
friendly ping @alyssawilk 

I think this is waiting on your input

source/common/http/async_client_impl.h
```diff
@@ -93,7 +98,40 @@ class AsyncStreamImpl : public AsyncClient::Stream,
 public:
   AsyncStreamImpl(AsyncClientImpl& parent, AsyncClient::StreamCallbacks& callbacks,
                   const AsyncClient::StreamOptions& options);
-  ~AsyncStreamImpl() override { router_.onDestroy(); }
+  ~AsyncStreamImpl() override {
+    router_.onDestroy();
+    // UpstreamRequest::onCleanup() is guaranteed to reset the high watermark calls.
```
alyssawilk, line 103: sorry I'm not finding onCleanup - do we have the right function name or is it PEBKAC?

alyssawilk (commented):
LG modulo remaining nits.  Though I'll say I'd be surprised if something didn't come up in your canary so I appreciate the default-off :-)

alyssawilk:
@snowp you game for a non-google pass?

snowp:
Yeah I'll give this a look today

source/common/buffer/buffer_impl.h
```diff
@@ -817,6 +817,47 @@ class OwnedImpl : public LibEventInstance {
   };
 };
 
+/*
+ * Helper class to handle a known number of buffer copies.
```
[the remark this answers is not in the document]
snowp: +1, a clear use case spelled out would be helpful

source/common/buffer/buffer_impl.h
```diff
@@ -817,6 +817,47 @@ class OwnedImpl : public LibEventInstance {
...
+      return copy_;
+    }
+  }
+
+  // Get an owned pointer to the next copy of the buffer.
+  Buffer::InstancePtr nextBufferOwned() {
+    ASSERT(num_copies_ > 0);
+    if (num_copies_ == 1) {
+      auto copy = std::make_unique<Buffer::OwnedImpl>();
+      copy->move(original_buffer_);
```
snowp, line 846: This is not actually a copy right? It's draining the original buffer?

source/common/http/async_client_impl.h
snowp: `UpstreamRequest::cleanUp` perhaps?

source/common/http/async_client_impl.h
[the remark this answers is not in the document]
snowp: Is the question here just around what happens if setDecoderBufferLimit is called more than once? I would think it should override the previous value if we expect multiple calls, or ASSERT if we don't expect multiple calls

source/common/router/router.cc
```diff
@@ -777,22 +825,29 @@ Http::FilterDataStatus Filter::decodeData(Buffer::Instance& data, bool end_strea
...
+  if (streaming_shadows_ && !shadow_streams_.empty()) {
+    for (auto* shadow_stream : shadow_streams_) {
+      if (end_stream) {
+        shadow_stream->removeDestructorCallback();
+        shadow_stream->removeWatermarkCallbacks();
+      }
+      shadow_stream->sendData(many_copied_buffer.nextBuffer(), end_stream);
+    }
+    if (end_stream) {
+      shadow_streams_.clear();
```
snowp, line 849: Why clear the streams at this point? Wont it all get torn down when the filter is destroyed?

source/common/router/router.cc
```diff
@@ -819,6 +874,15 @@ Http::FilterTrailersStatus Filter::decodeTrailers(Http::RequestTrailerMap& trail
   if (!upstream_requests_.empty()) {
     upstream_requests_.front()->acceptTrailersFromRouter(trailers);
   }
+  if (streaming_shadows_) {
+    for (auto* shadow_stream : shadow_streams_) {
```
snowp, line 878: Are you able to just iterate over shadow_streams_ without the first check? Wouldn't it be empty if disabled?

source/common/router/shadow_writer_impl.cc
```diff
@@ -11,31 +11,57 @@
 namespace Envoy {
 namespace Router {
 
+namespace {
+
+std::string newHost(absl::string_view host) {
```
snowp, line 16: Maybe give this a name that better describes what it does? An example could be helpful as well

snowp (changes requested):
Nice this is great work, a few comments

source/common/router/router.cc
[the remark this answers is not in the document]
paul-r-gall: Removed this in the most recent commit.

source/common/router/router.cc
```diff
@@ -698,6 +705,46 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
...
+      }
+      auto shadow_headers = Http::createHeaderMap<Http::RequestHeaderMapImpl>(*shadow_headers_);
+      auto options = Http::AsyncClient::RequestOptions()
+                         .setTimeout(timeout_.global_timeout_)
+                         .setParentSpan(callbacks_->activeSpan())
+                         .setChildSpanName("mirror")
+                         .setSampled(shadow_policy.traceSampled())
+                         .setIsShadow(true)
+                         .setBufferAccount(callbacks_->account())
+                         .setBufferLimit(callbacks_->decoderBufferLimit());
```
paul-r-gall, line 725: @alyssawilk @snowp what do you think the default limit should be here? I was thinking either `callbacks_->decoderBufferLimit()` or we could be more aligned to previous behavior and do `max(retry_shadow_buffer_limit_, 1)`.

source/common/http/async_client_impl.h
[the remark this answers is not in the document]
paul-r-gall: Made it an ENVOY_BUG. The only place that this could be called is from the router filter, and there's no calls in router.cc.

source/common/router/router.cc
paul-r-gall: It's more of a semantic thing -- I want `shadow_streams_` to only contain streams to which have not seen `end_stream=true`.

source/common/router/router.cc
paul-r-gall: Removed.

source/common/router/router.cc
```diff
@@ -750,7 +797,8 @@ Http::FilterDataStatus Filter::decodeData(Buffer::Instance& data, bool end_strea
   // a backoff timer.
   ASSERT(upstream_requests_.size() <= 1);
 
-  bool buffering = (retry_state_ && retry_state_->enabled()) || !active_shadow_policies_.empty() ||
+  bool buffering = (retry_state_ && retry_state_->enabled()) ||
+                   (!active_shadow_policies_.empty() && !streaming_shadows_) ||
```
snowp, line 801: Looking at what this checks gate, it seems like it would clear out `active_shadow_policies_` in the case where we're streaming shadow and (retries or internal redirects) are enabled - is that going to cause a problem?

source/common/router/router.cc
snowp: Neither are perfect choices IMO, maybe retry_shadow_buffer_limit so that users can still impact how much memory is used for shadowing without changing the whole decoder buffer?

snowp (changes requested):

paul-r-gall:
@snowp resolved both of the above issues. I decided to use `retry_shadow_buffer_limit_` for the buffer limit in the shadow.

alyssawilk:
main merge should fix the verify examples so @snowp PTAL even if that hasn't been resolved yet

source/common/router/router.cc
```diff
@@ -715,14 +716,17 @@ Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers,
...
+          Http::AsyncClient::RequestOptions()
+              .setTimeout(timeout_.global_timeout_)
+              .setParentSpan(callbacks_->activeSpan())
+              .setChildSpanName("mirror")
+              .setSampled(shadow_policy.traceSampled())
+              .setIsShadow(true)
+              .setBufferAccount(callbacks_->account())
+              // A buffer limit of 1 is set in the case that retry_shadow_buffer_limit_ == 0,
+              // because a buffer limit of zero on async clients is interpreted as no buffer limit.
+              .setBufferLimit(1 > retry_shadow_buffer_limit_ ? 1 : retry_shadow_buffer_limit_);
```
snowp, line 729: Would the `retry_shadow_buffer_limit_ == 0` case result in shadowing being effectively disabled unless the body happens to fit in a 1 byte buffer? 

If there's no way to disable buffering altogether for the async client maybe add a TODO about that, setting it to 1 seems like a little bit of a hack (but okay for now)

snowp (approved):
Thanks, this LGTM!

jpsim:
@paul-r-gall it looks like this broke Envoy Mobile, or at least Lyft's integration of Envoy Mobile in our iOS apps. Requests are failing where they previously didn't. I'll share more information about the failures we're seeing shortly, but for now I'd like to revert. https://github.com/envoyproxy/envoy/pull/25184

PR #24368, opened 2022-12-05
Add setRequestDecoder to ResponseEncoder interface

pksohn:
Commit Message: Add setRequestDecoder to ResponseEncoder interface
Additional Description: This change was introduced in #23648 but was moved into a separate PR given its more general scope.
Risk Level: Low
Testing: No tests added.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/envoy/http/codec.h
+++ b/envoy/http/codec.h
@@ -45,6 +45,7 @@ const char MaxResponseHeadersCountOverrideKey[] =
     "envoy.reloadable_features.max_response_headers_count";
 
 class Stream;
+class RequestDecoder;
 
 /**
  * Error codes used to convey the reason for a GOAWAY.
@@ -165,6 +166,12 @@ public:
    * error.
    */
   virtual bool streamErrorOnInvalidHttpMessage() const PURE;
+
+  /**
+   * Set a new request decoder for this ResponseEncoder.
+   * @param decoder new request decoder.
+   */
+  virtual void setRequestDecoder(RequestDecoder& decoder) PURE;
 };
 
 /**
--- a/envoy/http/codec.h
+++ b/envoy/http/codec.h
@@ -168,7 +168,9 @@ public:
   virtual bool streamErrorOnInvalidHttpMessage() const PURE;
 
   /**
-   * Set a new request decoder for this ResponseEncoder.
+   * Set a new request decoder for this ResponseEncoder. This is helpful in the case of an internal
+   * redirect, in which a new request decoder is created in the context of the same downstream
+   * request.
    * @param decoder new request decoder.
    */
   virtual void setRequestDecoder(RequestDecoder& decoder) PURE;
--- a/source/common/http/conn_manager_impl.cc
+++ b/source/common/http/conn_manager_impl.cc
@@ -1767,6 +1767,8 @@ void ConnectionManagerImpl::ActiveStream::recreateStream(
   connection_manager_.doEndStream(*this, /*check_for_deferred_close*/ false);
 
   RequestDecoder& new_stream = connection_manager_.newStream(*response_encoder, true);
+  // This doesn't currently work for HTTP/1 as the H/1 ResponseEncoder doesn't hold the active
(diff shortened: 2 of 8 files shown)


envoy/http/codec.h
```diff
@@ -165,6 +166,12 @@ class ResponseEncoder : public virtual StreamEncoder {
    * error.
    */
   virtual bool streamErrorOnInvalidHttpMessage() const PURE;
+
+  /**
+   * Set a new request decoder for this ResponseEncoder.
+   * @param decoder new request decoder.
```
alyssawilk, line 172: think it's worth a comment on why this is a useful association?

source/common/http/http1/codec_impl.h
```diff
@@ -156,6 +156,8 @@ class ResponseEncoderImpl : public StreamEncoderImpl, public ResponseEncoder {
     return stream_error_on_invalid_http_message_;
   }
 
+  void setRequestDecoder(Http::RequestDecoder& /*decoder*/) override {}
```
alyssawilk, line 159: or maybe why HTTP/1.1 doesn't need this

alyssawilk (commented):
LGTM modulo one of two explanatory comments :-)

envoy/http/codec.h
pksohn: added, ptal!

source/common/http/http1/codec_impl.h
pksohn: Added. H/1 will need more work to actually reset the decoder on internal redirect, but it seems like we don't need to tackle that now?

source/common/http/conn_manager_impl.cc
```diff
@@ -1767,6 +1764,12 @@ void ConnectionManagerImpl::ActiveStream::recreateStream(
   connection_manager_.doEndStream(*this, /*check_for_deferred_close*/ false);
 
   RequestDecoder& new_stream = connection_manager_.newStream(*response_encoder, true);
+  // Set the new RequestDecoder on the ResponseEncoder. Even though all of the decoder callbacks
+  // have already been called at this point, the encoder still needs the new decoder for deferred
+  // logging in some cases.
+  // This doesn't currently work for HTTP/1 as the H/1 ResponseEncoder doesn't hold the active
+  // stream's pointer to the RequestDecoder.
```
danzh2010, line 1771: For my own knowledge, how does H1 codec decode request headers in that case?

source/common/http/http1/codec_impl.h
danzh2010: We will need to make it work for H1 to defer logging I guess? Can you add a TODO?

source/common/http/conn_manager_impl.cc
pksohn: H1 has an [ActiveStream](https://github.com/envoyproxy/envoy/blob/28ac8d08d71b41544969d23415247dc53ce919c1/source/common/http/http1/codec_impl.h#L465-L466) that has a handle on both a ResponseEncoder and RequestDecoder. So e.g. on receiving headers the connection goes through the active request to get to the decoder, like [here](https://github.com/envoyproxy/envoy/blob/28ac8d08d71b41544969d23415247dc53ce919c1/source/common/http/http1/codec_impl.cc). It seems like the H/2 and H/3 codecs have the decoding path go through the response encoder -> request decoder partly to simplify multiplexing, so it makes sense that H/1 (which doesn't have streams) is organized a bit differently. I could be wrong about that last part.

alyssawilk (approved):

source/common/http/conn_manager_impl.cc
danzh2010: Get it! Thanks for the link!

danzh2010 (approved):
LGTM. Please update the PR description to reflect the recreateStream() change.

mattklein123 (approved):

PR #24385, opened 2022-12-06
Remove unecessary `this->` usages in mobile/...engine_builder.cc

caschoener:
For some reason this file uses `this->` to reference class members.  Let's ctrl-f to remove them all!

Tests work with all usages removed.


RyanTheOptimist (approved):
Thank you for doing this!

RyanTheOptimist:
Looks like it needs the format to be fixed, though.

PR #24386, opened 2022-12-06
build(deps): bump abseil-cpp to latest version

jpsim:
This also allows us to remove the mobile override and its patch because the new version of abseil-cpp pulls in https://github.com/google/cctz/pull/237.

Fixes https://github.com/envoyproxy/envoy-mobile/issues/136.

Commit Message:
Additional Description:
Risk Level: Moderate, bumping a core dependency
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
Fixes: https://github.com/envoyproxy/envoy-mobile/issues/136


alyssawilk (dismissed):
LGTM, ci-willing :-)

jpsim:
> ci-willing :-)

CI was not willing :(

Had to `2022-12-06` -> `2022-12-05`

jpsim:
I've spun up a Linux EC2 instance to try to reproduce the asan failures. Maybe that'll let me bisect the abseil changes and pinpoint the change that's causing this asan failure.

image

jpsim:
I filed https://github.com/abseil/abseil-cpp/issues/1339 to report the undefined behavior error reported in sanitizer CI jobs.

derekmauro:
> I filed abseil/abseil-cpp#1339 to report the undefined behavior error reported in sanitizer CI jobs.

https://github.com/envoyproxy/envoy/blob/8486a0ca5b26a1728415f56912dd8645c64a47d3/source/common/http/status.cc#L43-L65

The bug in the reinterpret_cast in this code. The reinterpret_cast assumes the data is pointer aligned. Instead the data needs to be read/written using memcpy or something endian-neutral if it is sent over the wire.

jpsim:
> The bug is in the reinterpret_cast in this code. The reinterpret_cast assumes the data is pointer-aligned. Instead the data needs to be read/written using memcpy or something endian-neutral if it is sent over the wire.

Correct me if I'm wrong, but wouldn't that incur the cost of a copy where the current code doesn't?

derekmauro:
> > The bug is in the reinterpret_cast in this code. The reinterpret_cast assumes the data is pointer-aligned. Instead the data needs to be read/written using memcpy or something endian-neutral if it is sent over the wire.
> 
> Correct me if I'm wrong, but wouldn't that incur the cost of a copy where the current code doesn't?

A copy is already being made in `storePayload()`. If you really want to play games, perhaps you could `new` a copy of the object in `storePayload()`, and then construct the `absl::Cord` using `absl::MakeCordFromExternal()`.

jpsim:
@yanavlasov could you please take a look?

source/common/http/status.cc
```diff
@@ -41,9 +42,13 @@ struct PrematureResponsePayload : public EnvoyStatusPayload {
 };
 
 template <typename T> void storePayload(absl::Status& status, const T& payload) {
-  absl::Cord cord(absl::string_view(reinterpret_cast<const char*>(&payload), sizeof(payload)));
+  const size_t payload_size = sizeof(payload);
+  char* buffer = new char[payload_size];
+  safeMemcpyUnsafeDst(buffer, &payload);
+  absl::Cord cord(absl::string_view(buffer, payload_size));
```
derekmauro, line 48: The changes to this function do not solve any problem. I think the only change needed is something like this:

```
template <typename T> void storePayload(absl::Status& status, const T& payload) {
  const T* allocated = new T(payload);
  absl::Cord cord = absl::MakeCordFromExternal(absl::string_view(&allocated, sizeof(allocated)), [allocated]() { delete allocated; });
  status.SetPayload(EnvoyPayloadUrl, std::move(cord));
}
```

Then `getPayload` should not have to change at all since the object will already be correctly aligned.

source/common/http/status.cc
```diff
@@ -57,7 +62,9 @@ template <typename T = EnvoyStatusPayload> const T& getPayload(const absl::Statu
       auto data = cord.TryFlat();
       ASSERT(data.has_value()); // EnvoyPayloadUrl cords are flattened ahead of time
       ASSERT(data.value().length() >= sizeof(T), "Invalid payload length");
-      payload = reinterpret_cast<const T*>(data.value().data());
+      char* buffer = new char[sizeof(T)];
```
derekmauro, line 65: Note that this buffer is being leaked.

derekmauro (changes requested):

source/common/http/status.cc
derekmauro: It is safe to cast to `const char *`.

jpsim: ok, trying that now in 68c52cfacc20ecff8f9e242c76d24ae616565256

jpsim: Looks like the tests pass with the latest changes, including with the address and undefined behavior sanitizers.

source/common/http/status.cc
```diff
@@ -41,7 +41,10 @@ struct PrematureResponsePayload : public EnvoyStatusPayload {
 };
 
 template <typename T> void storePayload(absl::Status& status, const T& payload) {
-  absl::Cord cord(absl::string_view(reinterpret_cast<const char*>(&payload), sizeof(payload)));
+  const T* allocated = new T(payload);
+  const absl::string_view sv =
+      absl::string_view(reinterpret_cast<const char*>(allocated), sizeof(allocated));
+  absl::Cord cord = absl::MakeCordFromExternal(sv, [allocated]() { delete allocated; });
```
alyssawilk, line 47: trusting proto folks and our tests for correctness here :-P

alyssawilk (approved):

yanavlasov (approved):

PR #24389, opened 2022-12-06
Remove uneccessary `this->` from mobile engine builder

caschoener:
This file uses `this->` whenever referencing class members but it doesn't need to. Let's just ctrl-f and delete them all.

Unit tests still work!


RyanTheOptimist:
Why did you close https://github.com/envoyproxy/envoy/pull/24385 and open a new PR with basically the same changes?

caschoener:
That one wasn't on a branch, it was on main.  So I think I had to close it?

My ssh agent was only working 50% of the time locally so some commands got lost.

caschoener:
> Do you know if there's a clang-format config option for this?

After skimming clang format docs I don't see anything, also the google3 style guide doesn't seem to mention it.

PR #24392, opened 2022-12-06
ci: changing android workflow wto always run but not alwasy perform t…

alyssawilk:
…ests

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/android_tests.yml
+++ b/.github/workflows/android_tests.yml
@@ -4,13 +4,7 @@ on:
   push:
     branches:
     - main
-    paths:
-    - 'mobile/**'
-    - '.github/workflows/**'
   pull_request:
-    paths:
-    - 'mobile/**'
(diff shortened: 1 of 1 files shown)


jpsim:
What if we extracted this to a `mobile/tools/should_run_ci.sh` check that passed a `--job` parameter so we could reuse this in more places and consolidate the logic?

```
./mobile/tools/should_run_ci.sh --job kotlintestsmac
```

I'm happy to tackle this if you're supportive of the direction.

alyssawilk:
sure, SGTM, and thanks for offering to pick it up (this is about as much workflow logic as I know how to handle =P)

PR #24394, opened 2022-12-06
downstream: refactoring code to remove listener hard deps

alyssawilk:
Risk Level: low (simply moving code around)
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/envoy/server/factory_context.h
+++ b/envoy/server/factory_context.h
@@ -288,6 +288,13 @@ public:
 };
 
 using FilterChainFactoryContextPtr = std::unique_ptr<FilterChainFactoryContext>;
+using FilterChainsByName = absl::flat_hash_map<std::string, Network::DrainableFilterChainSharedPtr>;
+
+class FilterChainBaseAction : public Matcher::Action {
+public:
+  virtual const Network::FilterChain* get(const FilterChainsByName& filter_chains_by_name,
+                                          const StreamInfo::StreamInfo& info) const PURE;
+};
 
 /**
  * An implementation of FactoryContext. The life time should cover the lifetime of the filter chains
--- a/source/extensions/matching/actions/format_string/config.cc
+++ b/source/extensions/matching/actions/format_string/config.cc
@@ -12,8 +12,9 @@ namespace Matching {
 namespace Actions {
 namespace FormatString {
 
-const Network::FilterChain* ActionImpl::get(const Server::FilterChainsByName& filter_chains_by_name,
-                                            const StreamInfo::StreamInfo& info) const {
+const Network::FilterChain*
+ActionImpl::get(const Server::Configuration::FilterChainsByName& filter_chains_by_name,
+                const StreamInfo::StreamInfo& info) const {
   const std::string name =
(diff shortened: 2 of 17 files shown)


envoy/server/factory_context.h
```diff
@@ -288,6 +288,13 @@ class FilterChainFactoryContext : public virtual FactoryContext {
 };
 
 using FilterChainFactoryContextPtr = std::unique_ptr<FilterChainFactoryContext>;
+using FilterChainsByName = absl::flat_hash_map<std::string, Network::DrainableFilterChainSharedPtr>;
+
+class FilterChainBaseAction : public Matcher::Action {
```
alyssawilk, line 293: in retrospect I should have commented this up when moving it.  If there end up being no other comments I'm inclined to do in the immediate follow-up to spare the 4 hour CI process.

alyssawilk:
I think clang tidy is one of those ones we have to force merge around:

/build/tmp/_bazel_envoybuild/b570b5ccd0454dc9af9f65ab1833764d/execroot/envoy/./source/common/common/hash.h:13:10: error: 'xxhash.h' file not found [clang-diagnostic-error]
#include "xxhash.h"

https://github.com/envoyproxy/envoy/blob/main/source/common/common/BUILD#L142

source/extensions/matching/actions/format_string/config.cc
```diff
@@ -12,8 +12,9 @@ namespace Matching {
 namespace Actions {
 namespace FormatString {
 
-const Network::FilterChain* ActionImpl::get(const Server::FilterChainsByName& filter_chains_by_name,
-                                            const StreamInfo::StreamInfo& info) const {
+const Network::FilterChain*
+ActionImpl::get(const Server::Configuration::FilterChainsByName& filter_chains_by_name,
+                const StreamInfo::StreamInfo& info) const {
```
RyanTheOptimist, line 17: Out of curiosity, why did this line change?  It *looks* identical but just re-wrapped. Did clang-format dislike the old formatting, I guess?

source/server/listener_stats.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include "envoy/stats/scope.h"
+#include "envoy/stats/stats_macros.h"
+
+namespace Envoy {
+namespace Server {
+
+#define ALL_LISTENER_STATS(COUNTER, GAUGE, HISTOGRAM)                                              \
```
RyanTheOptimist, line 9: Consider adding a brief comment here.

envoy/server/factory_context.h
RyanTheOptimist: SGTM! I did have one additional comment but it was to add a comment, so feel free to address both in a follow up instead of waiting on CI again!

RyanTheOptimist (approved):

source/extensions/matching/actions/format_string/config.cc
alyssawilk: Server::FilterChainsByName ->  Server::Configuration::FilterChainsByName
because it was declared in the envoy/ file which has a different namespace

Issue #23821, opened 2022-11-03
LinkCreatesNamedFile test failure


jwendell:
This test is consistently failing in my environment: https://github.com/envoyproxy/envoy/blob/35bf11a0f14a996d4e95b5302b0a953e0beba9f6/test/extensions/common/async_files/async_file_handle_thread_pool_test.cc#L140

Here's my attempt to explain why:

The function `manager_->createAnonymousFile()` relies on the underlying filesystem supporting the `O_TMPFILE` flag: https://github.com/envoyproxy/envoy/blob/35bf11a0f14a996d4e95b5302b0a953e0beba9f6/source/extensions/common/async_files/async_file_manager_thread_pool.cc#L135

The fallback code (a few lines below) creates a file and deletes it, without closing it, returning its file descriptor. I'm reaching this fallback code, because my filesystem doesn't support the `O_TMPFILE` flag.

The problem happens when a hard link is created later with `linkat()`: https://github.com/envoyproxy/envoy/blob/35bf11a0f14a996d4e95b5302b0a953e0beba9f6/source/extensions/common/async_files/async_file_context_thread_pool.cc#L49

It fails with ENOENT, as stated in the manual page:
```
ENOENT An attempt was made to link to a /proc/self/fd/NN file
              corresponding to a file that has been deleted.
```
As seen here: https://man7.org/linux/man-pages/man2/link.2.html

cc @ravenblackx

ravenblackx:
What sort of environment is it that exhibits this failure? (Thinking maybe an NTFS filesystem on Linux?)

I would suggest this could be fixed by having the fallback behavior create a named temporary file (unfortunately), keeping track of that file in the AsyncFileContext object, moving it in lieu of linkat or deleting it during close. It should maybe be a distinct AsyncFileContext (though perhaps a subclass of AsyncFileContextThreadPool) since the more common case works without that information. That variant could also potentially make this able to function in Windows.

(Unfortunate because it could potentially make for inconsistent behavior with any code that looks at directory contents.)

jwendell:
The env is a docker container running in a k8s cluster probably running in an old kernel.

ravenblackx:
I have a plan to address this, involving migrating `AsyncFileManager` to using `Envoy::Filesystem::File`, after making `File` have an implementation for temp files and all the other required operations.
I'd assign this issue to me but I don't think I have permission to do that.

ravenblackx:
Still working on it, bot!

PR #24398, opened 2022-12-06
[WIP] Add capability to create tmpfile to envoy/filesystem

ravenblackx:
Commit Message: Add capability to create tmpfile to envoy/filesystem
Additional Description: Currently `AsyncFileManager` uses `OsSysCall` to perform file operations. A constraint of this is that it is using file descriptors, which in Windows makes it impossible to `pread`/`pwrite` (there is no equivalent in Windows that works with file descriptors), and awkward to `close` (doesn't know whether to `closesocket` or `close` for a file descriptor). Migrating off `OsSysCall` and onto `Envoy::Filesystem::File` would simplify those things (because `File` uses a `HANDLE` in Windows) and also reduce the amount of special handling in AsyncFileManager.

This is a first step towards that migration, to be followed with `pread`, `pwrite`, `rename` and `stat`.

#23821 is related.
Risk Level: Low. Change is mostly separated from existing code by a switch case on a new enum value. The only change to the taken code path is adding FILE_SHARE_DELETE to the properties of an opened file in Windows, which should have no practical impact at this time (it is necessary in order to allow rename operations later).
Testing: Fully covered except one line, a bit awkwardly, as that branch of the tmpfile creation can't be covered naturally, as reaching it depends on the behavior of the test tmpdir file system.
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Per-platform implementation.


yanavlasov (approved):

tyxia:
just a small nit: You may want to remove `[WIP]` when it is ready for review/submitted to main.
thanks

ravenblackx:
Yeah, I only noticed I hadn't removed it after it was too late, unfortunately.

Issue #23718, opened 2022-10-27
high cpu usage on tls handshack backpressure from http health check timeout


JuniorHsu:
*Title*: *high cpu usage on tls handshack backpressure from http health check timeout*

*Description*:
We notice a host is slow to response to admin port and cpu usage is maxed out on our docker limit.
flamegraph shows > 25% of cpu usages on tls handshake for http health checker timeout.

`Envoy::Upstream::HttpHealthCheckerImpl::HttpActiveHealthCheckSession::onTimeout` tries to close the socket but `ConnectionCloseType::NoFlush` is still try to write bytes out on
https://github.com/envoyproxy/envoy/blob/489a41f248ebb02bcba3e30788a6c4ba943b678e/source/common/network/connection_impl.cc#L143-L147 which needs tls handshake.

We should have a new flag like `ConnectionCloseType::NoFlushAndWrite` (a bad naming I admit) to avoid the back pressure given the cpu is already busy.

*Config*:
The issue is mitigated after we increase timeout to 2 seconds
```json

    "health_checks": [
      {
        "healthy_threshold": 1,
        "http_health_check": {
          "path": "/_/_/health/"
        },
        "initial_jitter": "0s",
        "interval": "5s",
        "interval_jitter": "10s",
        "timeout": "1s",
        "tls_options": {
          "alpn_protocols": [
            "http/1.1"
          ]
        },
        "unhealthy_threshold": 1
      }
```

cc @mattklein123 @adisuissa @fishcakez @PeterL328

mattklein123:
I suppose we could also just stop trying to write pending data in the NoFlush case? Seems like if we are already not caring about flushing and this is causing issues we can just remove it backed by runtime guard?

cc @alyssawilk @ggreenway

ggreenway:
Yeah, I think it makes sense to have a `ConnectionCloseType::Abort` that just closes everything doing the minimum amount of work, with no gracefulness even attempted.

mattklein123:
> Yeah, I think it makes sense to have a ConnectionCloseType::Abort that just closes everything doing the minimum amount of work, with no gracefulness even attempted.

I think my question above was whether NoFlush should just be changed to be the equivalent of your "abort" ?

JuniorHsu:
i have a diff to ci to see if we have test breakage. stay tuned

ggreenway:
[...]

PR #24401, opened 2022-12-06
connection: prevent writing pending data for health checkers by introducing ConnectionCloseType::Abort

JuniorHsu:

Commit Message:
Please check #23718

Additional Description:
Risk Level: low
Testing: unit test in connection and health checker
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue] #23718
[Optional Deprecated:]


source/common/network/connection_impl.cc
```diff
@@ -140,7 +140,7 @@ void ConnectionImpl::close(ConnectionCloseType type) {
   const bool delayed_close_timeout_set = delayed_close_timeout_.count() > 0;
   if (data_to_write == 0 || type == ConnectionCloseType::NoFlush ||
       !transport_socket_->canFlushClose()) {
-    if (data_to_write > 0) {
+    if (data_to_write > 0 && type != ConnectionCloseType::NoFlush) {
       // We aren't going to wait to flush, but try to write as much as we can if there is pending
       // data.
       transport_socket_->doWrite(*write_buffer_, true);
```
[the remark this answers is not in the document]
JuniorHsu: thanks for a quick update! i was trying to see if we have any current unit breakage. 
i'm not super sure what you mean about the else block.
we want to prevent `doWrite`  for NoFlush even if we have data, right?

test/common/network/connection_impl_test.cc
```diff
@@ -2880,6 +2880,19 @@ TEST_F(PostCloseConnectionImplTest, NoReadAfterCloseFlushWriteWriteData) {
...
 }
 
+// Test that close(ConnectionCloseType::Abort) won't write and flush pending data.
+TEST_F(PostCloseConnectionImplTest, CloseAbort) {
+  InSequence s;
+  initialize();
+  writeSomeData();
+
+  // Connection abort. We have data written above in writeSomeData(),
+  // it won't be written and flushed due to Abort.
```
soulxu, line 2890: super nit, s/Abort/abort

soulxu (dismissed):
LGTM, thanks

test/common/network/connection_impl_test.cc
JuniorHsu: hmm `abort` might fail grammar  in some sense. I use ``ConnectionCloseType::Abort`` if you don't mind. Thanks for catching this!

changelogs/current.yaml
```diff
@@ -107,6 +107,9 @@ bug_fixes:
 - area: jwt_authn
   change: |
     fix a bug that jwt_cache breaks the :ref:`provider_and_audiences <envoy_v3_api_field_extensions.filters.http.jwt_authn.v3.JwtRequirement.provider_and_audiences>` JWT requirement.
+- area: health_checker
+  change: |
+    prevent writing pending data for health checkers by introducing ``ConnectionCloseType::Abort``.
```
snowp, line 112: For my understanding, why does this matter? What would happen previously? Should we include some detail in the release note?

snowp (changes requested):
Overall lgtm, just one comment re release note. Is this something we'd want to runtime guard?

changelogs/current.yaml
JuniorHsu: The detail is in #23718. 
tl;dr there's cascading cpu backpressue for health checker handshake while cpu is throttled by docker cgroup limit. More and more HC requests are timeout-ed and closed before handshake, but we still try to TLS handshake.

IMO runtime guard is **not** needed for now since it doesn't matter to give up health checker's data.

However it needs some deliberation to replace as much `NoFlush` with `Abort` as possible in the future
https://github.com/envoyproxy/envoy/issues/23718#issuecomment-1341338352

I'll mention the detail in the release note in next commit.

envoy/network/connection.h
```diff
@@ -69,8 +69,9 @@ class ConnectionCallbacks {
 enum class ConnectionCloseType {
   FlushWrite, // Flush pending write data before raising ConnectionEvent::LocalClose
   NoFlush,    // Do not flush any pending data and immediately raise ConnectionEvent::LocalClose
```
soulxu, line 71: It would be better to mention that this type will give a try to write left data.

JuniorHsu: addressed. thanks!

soulxu (commented):
thanks, LGTM

yanavlasov (approved):

JuniorHsu:
coverage ci failure should be unrelated given I didn't change the folder
```
Code coverage for source/extensions/http/early_header_mutation is lower than limit of 96.6 (95.2)
Code coverage for source/extensions/filters/network/http_connection_manager is lower than limit of 96.6 (96.4)
```

JuniorHsu:
merge to top which mitigate coverage issue

JuniorHsu:
ci passed and we got approved. could we merge this? Thanks. @snowp @yanavlasov

yanavlasov:
I'll ping @snowp for LGTM and then submit.

PR #24402, opened 2022-12-06
swift: stop using loopback for direct responses

alyssawilk:
This changes swift's direct responses from being proxied upstream and handled by the "fake" (real) listener, to being handled locally, and bypassing the "do not forward local replies" logic in the onLocalReply classes

Commit Message: na/a
Additional Description: n/a
Risk Level: medium (mobile only)
Testing: existing tests
part of https://github.com/envoyproxy/envoy/issues/24428

### DIFF
--- a/mobile/library/common/config/config.cc
+++ b/mobile/library/common/config/config.cc
@@ -267,7 +267,6 @@ const char* config_template = R"(
             - name: remote_service
               domains: ["*"]
               routes:
-#{fake_remote_responses}
               - match: { prefix: "/" }
                 direct_response: { status: 404, body: { inline_string: "not found" } }
                 request_headers_to_remove:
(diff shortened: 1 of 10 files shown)


mobile/library/common/config/config.cc
```diff
@@ -267,7 +267,6 @@ const char* config_template = R"(
             - name: remote_service
               domains: ["*"]
               routes:
-#{fake_remote_responses}
```
Augustyniak, line 270: oh, I thought that we would be able to remove a bigger portion of this fake listener as part of this step. What's preventing us from doing that (apologies if you have explained it in the community sync meeting)?

Augustyniak (approved):

mobile/library/common/config/config.cc
alyssawilk: I was going to do that as a follow-up after I removed listener support so we can make sure it doesn't break Lyft tests first :-)

alyssawilk: so the plan is land this, then https://github.com/envoyproxy/envoy/pull/24363 (working on a c-opt for legacy tests like the kotlin tests) then remove the listener entirely

PR #24403, opened 2022-12-06
Consistent rust matcher

snowp:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


snowp:
Alright so this basically green now except for a drop in coverage, a few notes:
- To get gcc working this switched the linker over to lld - is this ok? I imagine we might need to reach out and see if people using gold actually need it
- There seems to be a thing on Windows where rust_library depending on a cc_library with `linkstatic = False` causes the cc_library's transitive deps to ignore `--dynamic_mode`, but I was able to work around this with the provided patch + custom macro, so I think we're good here
- There is an expected drop in coverage as the coverage tool isnt able to compare both Rust and C++ coverage data - this would have significant implications on our coverage tooling if we started adopting more Rust
- Cargo deps likely needs the same third party dep treatment as any other third party dep, so likely we'd need some process change there
- There are known complications with adding Rust to the EM iOS build + known binary size increase by adding the Rust stdlib. Is this going to prevent us from being able to add this to Envoy core?

PR #24408, opened 2022-12-07
bazel: test vendored llvm toolchain

keith:


phlax:
@keith i think this may still be worth investigating - im aware @lizan mentioned that it is the same statically compiled llvm bins that we currently use in the docker container - but im also wondering whether it would still save a bit of rbe/ci time as only the used bins rather than the whole suite would be getting transferred around.

it also would potentially make the llvm toolchain authoritative here, which i think is also a good idea

keith:
Yea I think it would have the major benefit that if folks didn't care about their compiler version there would be 0 setup. I  got slightly less motivated to push it when I heard that it sounds like gcc support is required by some, so it wouldn't eliminate as much maintenance as I hoped

mattklein123:
> Yea I think it would have the major benefit that if folks didn't care about their compiler version there would be 0 setup. I got slightly less motivated to push it when I heard that it sounds like gcc support is required by some, so it wouldn't eliminate as much maintenance as I hoped

I would like to hear why GCC is actually required for anyone if the entire thing could be vendored? IMO we should push this.

keith:
maybe @lizan can provide more context, but it sounded like the redhat folks specifically were reliant on gcc

phlax:
another nail in the coffin of our current way of provisioning llvm/clang ...

currently we pull the prebuilt x86_64 binaries from the llvm github project releases, but they have stopped building this from 15+

https://github.com/llvm/llvm-project/releases

https://discourse.llvm.org/t/llvm-15-0-0-release/65099

phlax:
@keith is there any way we can land this for llvm and not lose gcc ? apologies if this is a naive question

keith:
it's a good question, i think we could but I'm not yet sure how, i think we could probably introduce some new flag for forcing gcc and do it that way. or maybe env var. note that this toolchain also uses the llvm binary downloads so if those do go away it wouldn't work for this either. since you posted that comment some appeared to have trickled in. also I have LLVM push access so theoretically i could build and upload the ones we care about at least 😬

phlax:
great - `debian-bullseye-slim` please!

phlax:
re bins - unless im misreading something - the only recent ubuntu ones are for ppcle64

phlax:
@keith im circling this again as im trying to trim down the docker build images and would like to shift clang tools into bazel

im wondering if we could add this toolchain without immediately using to compile - just to use the tooling, and then hopefully switch compiler when possible

phlax:
see #29415 for updated PR

PR #24418, opened 2022-12-07
grpc_http1_bridge: prevent excessive data removal

kmcbride:
When leveraging the Protobuf upgrade functionality there is the possibility of data loss for responses that are processed in multiple chunks, as the frame header to be removed is seemingly only present in the first chunk.

This is a fix for an issue that we've seen in practice: clients fail to decode responses >16KB because 5 bytes of data is missing every subsequent ~16KB.


source/extensions/filters/http/grpc_http1_bridge/http1_bridge_filter.cc
```diff
@@ -72,6 +72,7 @@ Http::FilterDataStatus Http1BridgeFilter::encodeData(Buffer::Instance& data, boo
   } else {
     // Buffer until the complete request has been processed.
     if (do_framing_) {
+      do_framing_ = false;
```
snowp, line 75: Could you add a comment explaining why we set this to false here? The grpc reverse filter has its own variable to guard this, but changing `do_framing_` seems fine as well

kmcbride:
@snowp thanks for the review. The latest commit should hopefully address the comments.

I'm not entirely sure what's going on with the failing coverage job because the [error message](https://dev.azure.com/cncf/envoy/_build/results?buildId=123027&view=logs&j=5035ce29-be3c-5fcf-dc50-abb6a265f8b1&t=548c3e59-0417-5971-4f1b-f8545005c291) seems unrelated to any changes I've made.

snowp (approved):
Thanks!

kmcbride:
@snowp CI seems stuck 🙁

snowp:
Can you try pushing an empty commit? I don't think we have a way to force these to re-run

kmcbride:
@snowp I think this PR needs to be re-approved because I had to push a commit that fixed a merge conflict in the changelog.

snowp (approved):
Thanks!

PR #24421, opened 2022-12-07
tls: allow multiple certs with the same SAN

ggreenway:

Commit Message: Augments the recently added (#22036) support for SNI-based certificate selection to allow names to be duplicated across different certificates. The first cert in the list with a given name is used for that name.
Additional Description:
Risk Level: low
Testing: added tests
Docs Changes: changed
Release Notes: The note from #22036, which has not yet released, is adequate
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


ggreenway:
@LuyaoZhong after some discussions with people, it seemed that restricting to no duplicate SANs was quite cumbersome in some cases. Any thoughts or objections to this change?

source/extensions/transport_sockets/tls/context_impl.cc
```diff
@@ -889,10 +889,8 @@ void ServerContextImpl::populateServerNamesMap(TlsContext& ctx, int pkey_id) {
     auto sn_match = server_names_map_.try_emplace(sn_pattern, pkey_types_map).first;
     auto pt_match = sn_match->second.find(pkey_id);
     if (pt_match != sn_match->second.end()) {
-      throw EnvoyException(fmt::format(
-          "Failed to load certificate chain from {}, at most one "
-          "certificate of a given type may be specified for each DNS SAN entry or Subject CN: {}",
-          ctx.cert_chain_file_path_, sn_match->first));
+      // When there are duplicate names, prefer the earlier one.
+      return;
```
mattklein123, line 893: I think we should at least warn about this? This seems like something the operator should know about?

ggreenway: From some conversations I've had, I think it's common enough (and intentional) that I'm not sure it deserves a warning. I think in most cases that will just be noise in the logs.

ggreenway: It's probably not the entire cert that won't be used; just one of the SANs.

With some more tracking, I could figure out if any cert is completely shadowed for all SANs by other certs. That seems worth warning about. But it adds complexity, and I don't know that it's needed.

mattklein123: Oh I see, OK. Yeah maybe just add a TODO/comment about that.

LuyaoZhong:
> @LuyaoZhong after some discussions with people, it seemed that restricting to no duplicate SANs was quite cumbersome in some cases. Any thoughts or objections to this change?

This change looks good to me, raising exception is overkill, actually this is what I do for now in my case in downstream. I think the pain is if we allow loading duplicate SANs certs, how do we define selection criteria? I have no idea about that, so ignoring duplicate certs is good enough for now.

mattklein123 (approved):

test/extensions/transport_sockets/tls/ssl_socket_test.cc
```diff
@@ -1340,6 +1341,72 @@ TEST_P(SslSocketTest, MultiCertPreferExactSniMatch) {
...
+      private_key:
+        filename: "{{ test_rundir }}/test/extensions/transport_sockets/tls/test_data/san_multiple_dns_1_key.pem"
+    - certificate_chain:
+        filename: "{{ test_rundir }}/test/extensions/transport_sockets/tls/test_data/san_dns_rsa_1_cert.pem"
+      private_key:
+        filename: "{{ test_rundir }}/test/extensions/transport_sockets/tls/test_data/san_dns_rsa_1_key.pem"
+)EOF";
+
+    TestUtilOptions test_options(client_ctx_yaml, server_ctx_yaml, true, version_);
+    testUtil(test_options.setExpectedSni("server1.example.com"));
```
soulxu, line 1406: It will be great to ensure the first cert is used.

soulxu (approved):
LGTM, thanks

test/extensions/transport_sockets/tls/ssl_socket_test.cc
LuyaoZhong: It is already verified by hash matching, see L1390-L1391

soulxu: thanks for point that out

LuyaoZhong (approved):

LuyaoZhong (changes requested):
This should be land after https://github.com/envoyproxy/envoy/pull/24483.

Issue #23774, opened 2022-11-01
The size of the buffer in the filter chain is set inconsistently, which causes problems with grpc transcoder


ravenblackx:
When a stream is http2, the maximum size of the buffer in the filter chain is derived (surprisingly) from `initial_stream_window_size`

When it's not http2, the maximum size of the buffer comes from `per_connection_buffer_limit_bytes`, which sounds more reasonable from the name, but turns out to still actually be surprising, because that value is intended to be "the maximum size of a single read" not "the maximum size of the stream buffer".

There appears to be no way to directly set the maximum size of the stream buffer.

An undesirable side-effect of this behavior is that `grpc_json_transcoder` caps a message size at whichever of those values is controlling the buffer size, erroring with e.g. `transcoder's internal buffer size exceeds the configured limit: 4210688 > 4194304` if that limit is exceeded.

A quick workaround for this is to include an `http_buffer` filter in the chain with the desired buffer size, but that also adds other behaviors that aren't necessarily desirable (i.e. it prevents streaming of the request/response even if the transcoder isn't handling a request), and clearly should not be necessary.

I propose either of two reasonable fixes for this:
1. Add a configuration option at the top level for setting the stream buffer size (which if unset will default to the current behavior, and must be greater or equal to the value from the current behavior)
2. Add a maximum message size field to grpc_json_transcoder, which calls setDecoderBufferLimit and setEncoderBufferLimit (or two fields so the maximum size can be set independently per direction) if set (and if greater than the current buffer limit).

I'd be happy to implement either of these; some discussion of which is more desirable first would be nice, or maybe someone has a better idea.

(Some prior discussion [in slack](https://envoyproxy.slack.com/archives/C78HA81DH/p1666991566358279).)

KBaichoo:
I'm leaning more in favor of the second solution since it will push the complexity just to where it's needed -- in this case the grpc transcoder filter which might expect a given size for transcoding.

cc @nareddyt

nareddyt:
This is surprising behavior for me too. I always assumed the buffer size would come from `per_connection_buffer_limit_bytes`, but I read the discussion on slack and understand why it's not the case.

As for the proposed solution, I would **not** prefer 2.

`setDecoderBufferLimit` and `setEncoderBufferLimit` change the buffer limits for the entire filter chain. It is nothing specific to gRPC transcoder config. It seems odd to have configuration in a single filter to mutate limits in the chain.

[...]

PR #24426, opened 2022-12-07
Replace set..coderBufferLimit with set..coderBufferLimitLowerBound

ravenblackx:
Commit Message: Replace set..coderBufferLimit with set..coderBufferLimitLowerBound
Additional Description: Per #23774 the current behavior of setDecoderBufferLimit and setEncoderBufferLimit can be surprising if two filters both try to use it. This more reasonable implementation recognizes that there's no reason for a filter to want a *smaller* buffer, so if two filters both ask for a size, whichever requires the larger size should take priority. Making these function only act to increase the buffer-limit solves this, while renaming the functions makes it clear that the change has happened.
Risk Level: Low. There are few calls to these functions.
Testing: Tests modified slightly to cover the change.
Docs Changes: Updated flow_control.md
Release Notes: Yes, in changelogs.

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -27,6 +27,13 @@ behavior_changes:
 
 minor_behavior_changes:
 # *Changes that may cause incompatibilities for some users, but should not for most*
+- area: http_filters
+  change: |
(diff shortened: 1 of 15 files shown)


ravenblackx:
Making tests work with this change isn't work the effort.

Issue #24372, opened 2022-12-06
[contrib] kafka: record-distributing Kafka consumer proxy for multiple upstream clusters (mesh-filter new feature)


adamkotwasinski:
*Title*: *kafka: record-distributing Kafka consumer proxy for multiple upstream clusters (mesh-filter new feature)*

*Description*:
To follow up with Kafka producer proxy, I'd like to have Envoy act as a proxy for _multiple_ Kafka upstreams.
The consumer would want to point to Envoy instance, and with a `poll()` it would receive records from all matching upstreams.
Initially we could start with the same thing as Kafka producer handler does (policies based on topic names).

![consumer-distributing-proxy](https://user-images.githubusercontent.com/17969711/205775470-d99f484a-254a-43b7-a446-cf1534ca0619.png)

To provide record distribution (load sharing), Envoy would start N (librdkafka) consumers matching downstreams' requirements and connect them to upstream clusters. Envoy would then continuously poll for records (similar to what it does with producers now) and if there is "interest" (someone wants our records) then it'd immediately answer. Limiting the number of messages stored will be included.

On the protocol level this means the following requests from "simple" consumer flow need to be handled: `ListOffsets` (to handle consumer initially asking "where am I?") and `Fetch` (to do the real work).

Future work would then add another "proper proxy" mode where we would get into internals of upstream connection and then translate downstream FetchRequests into upstream ones (and here we might not need librdkafka anymore, as we get to a lower level).

*Relevant Links*:
* [current mesh filter](https://www.envoyproxy.io/docs/envoy/v1.22.0/configuration/listeners/network_filters/kafka_mesh_filter)

adamkotwasinski:
On PTO until mid-Jan, so don't get angry with me, repokitteh ;)

adamkotwasinski:
Bad bot - this issue had a PR merged just a few days ago `:)`

adamkotwasinski:
Bad bot - this issue had a PR merged just a few days ago `:)`

adamkotwasinski:
hey, sorry I was away and now I'm back a bit sick
btw. bad bot - issue mentioned `2 weeks ago`

adamkotwasinski:
Fetch request interactions:
```mermaid
graph TD;
    FRH["FechRequestHolder"]
    KafkaMeshFilter <-.-> |"in-flight-reference\n(finish/abandon)"| FRH
    KafkaMeshFilter --> RP["RequestDecoder+RequestProcessor"]
[...]

PR #24431, opened 2022-12-07
kafka: add upstream-kafka-consumer

adamkotwasinski:
Commit Message: kafka: add upstream-kafka-consumer
Additional Description: Add the upstream Kafka consumer (analogous to already existing upstream Kafka producer). As it is performing blocking operations (`consume`) it needs to run in a dedicated worker thread to consume the records. Compared to the producer it is a bit more complex, as it does not always need to consume the records - first it needs to check whether there is anyone interested in these topics (== if there are unfinished `Fetch` requests received by the mesh filter). Part of https://github.com/envoyproxy/envoy/issues/24372.
Risk Level: Low
Testing: unit tests, integration tests with full stack
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/contrib/kafka/filters/network/source/mesh/BUILD
+++ b/contrib/kafka/filters/network/source/mesh/BUILD
@@ -111,6 +111,18 @@ envoy_cc_library(
 )
 
 envoy_cc_library(
+    name = "inbound_record_lib",
+    srcs = [
+    ],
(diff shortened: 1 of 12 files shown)


mattklein123 (approved):

adamkotwasinski:
Thank you!

PR #24438, opened 2022-12-08
listener: reject specifying an endpoint id for envoy internal address

soulxu:
Commit Message: listener: reject specifying an endpoint id for envoy internal address 
Additional Description:
According to the discussion at https://github.com/envoyproxy/envoy/pull/24198#discussion_r1040147664, specifying endpoint id for listener is meanless and we want to have a consistent hash for all type addresses.

Risk Level: low
Testing: unittest
Docs Changes: API doc
Release Notes: minor behavior change


soulxu:
cc @kyessenov 

In the end, I didn't add a runtime flag, since there should be no one using the endpoint id in the listener. but let me know if we should protect this change with runtime flag.

Thanks

kyessenov (approved):
Thanks, agree about the runtime flag (it's not used in the implementation anyways and the control plane can easily adapt).

soulxu:
cc @mattklein123 for api and senior maintainer review

PR #24447, opened 2022-12-08
ci: move mobile docs job to Github Actions

jpsim:
This is the only use of CircleCI in this repo, so removing this helps us remove it altogether.

As part of this change, also stop publishing docs when tags are pushed.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim:
We can remove the CircleCI webhook and then we can remove the config file itself.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,33 @@
...
+    branches:
+    - main
+    tags:
+    - v*
+  pull_request:
+
+jobs:
+  docs:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 10
```
phlax, line 14: we might want to give it a longer timeout as it seems like contention can slow things down

jpsim:
> not sure if its any different to now but im concerned that if we tag an Envoy branch we will trigger a envoy-mobile release build (and i guess vice versa - how we tag a mobile branch without triggering an Envoy release)

Yes, when Envoy releases are tagged, this will build and publish docs to https://envoymobile.io.

I don't think it's super important at this stage in the Envoy Mobile project to publish docs when releases are tagged, so I'll remove that for now.

mobile/docs/publish.sh
```diff
@@ -13,10 +13,10 @@ DOCS_DIR=generated/docs
 CHECKOUT_DIR=../envoy-mobile-docs
 BUILD_SHA="$(git rev-parse HEAD)"
 
-if [ -n "$CIRCLE_TAG" ]
+if [ "$GITHUB_REF_TYPE" == "tag" ]
```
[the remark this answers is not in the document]
jpsim: Not anymore as of b96b0d56e297567bc93f43112acea39abe75fb84.

phlax (approved):
lgtm, thanks @jpsim

phlax:
i think ive seen the flake that has happened before:

```console
 ERROR: /private/var/tmp/_bazel_runner/38952b625e1e7c72284b0f9be4a35639/external/build_bazel_rules_swift_local_config/BUILD:9:22: While resolving toolchains for target @build_bazel_rules_swift_local_config//:toolchain: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type.
```

not sure what triggers that

jpsim:
I filed an issue to track the toolchain resolution flake: https://github.com/bazelbuild/rules_swift/issues/950

.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+
+jobs:
+  docs:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 20
+    container:
+      image: envoyproxy/envoy-build-ubuntu:b0ff77ae3f25b0bf595f9b8bba46b489723ab446
+    steps:
+    - uses: actions/checkout@v1
```
lizan, line 16: Doesn't `checkout@v3` work? Too old git in the image?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Add safe directory
+      run: git config --global --add safe.directory /__w/envoy/envoy
```
lizan, line 18: use `GITHUB_WORKSPACE` instead of hardcoding?

jpsim: Seeing if `checkout@v3` removes the need for this.

phlax: if it doesnt you can specify `safe.directory` with v3 directly i believe

jpsim: Still needed it but your `GITHUB_WORKSPACE` suggestion worked.

.github/workflows/docs.yml
jpsim: `checkout@v3` works.

jpsim:
This needs a final approval so I can merge.

phlax (approved):
lgtm, thanks @jpsim

PR #24454, opened 2022-12-08
stats: setting default value for returned values

adisuissa:
Commit Message: stats: setting default value for returned values
Additional Description:
Our internal msan test fails here, because the 2 data-members could be directly fetched without these values being explicitly initialized. Initializing the values solves the issue.
There should not be any impact, as the explicit values are the also the default ones.

Risk Level: Low
Testing: N/A.
Docs Changes: N/A.
Release Notes: N/A.

### DIFF
--- a/source/common/stats/histogram_impl.h
+++ b/source/common/stats/histogram_impl.h
@@ -70,7 +70,7 @@ private:
   std::vector<uint64_t> computed_buckets_;
   uint64_t sample_count_{0};
   double sample_sum_{0};
-  const Histogram::Unit unit_;
+  const Histogram::Unit unit_{Histogram::Unit::Unspecified};
 };
 
 class HistogramImplHelper : public MetricImpl<Histogram> {
--- a/source/common/stats/histogram_impl.h
+++ b/source/common/stats/histogram_impl.h
@@ -68,8 +68,8 @@ private:
   ConstSupportedBuckets& supported_buckets_;
   std::vector<double> computed_quantiles_;
   std::vector<uint64_t> computed_buckets_;
-  uint64_t sample_count_;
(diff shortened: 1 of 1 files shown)


KBaichoo (dismissed):

source/common/stats/histogram_impl.h
```diff
@@ -68,8 +68,8 @@ class HistogramStatisticsImpl final : public HistogramStatistics, NonCopyable {
   ConstSupportedBuckets& supported_buckets_;
   std::vector<double> computed_quantiles_;
   std::vector<uint64_t> computed_buckets_;
-  uint64_t sample_count_;
-  double sample_sum_;
+  uint64_t sample_count_{0};
+  double sample_sum_{0};
```
jmarantz, line 72: should we also init unit_  if we are initing the other scalars?

pradeepcrao: It is initialized in member initializer list of both constructors, but doing it here makes it easier to see the default value.

adisuissa: I don't think it can be initialized here as it is defined as `const`.

adisuissa: I can move the initialization of `sample_count_` and `sample_sum_` to the MIL, if you prefer.

KBaichoo: I think you can add a default here even though it's const.

adisuissa: TIL, thanks!
Done.

KBaichoo (approved):

alyssawilk (approved):

jmarantz (approved):

PR #24456, opened 2022-12-08
Remove inheritance from base_client_integration_test to engine_builder

caschoener:
Currently the unit test fixture inherits from the engine builder. This is mostly just a readability win and also allows setting up multiple builders within one test.

### DIFF
--- a/mobile/library/cc/engine_builder.h
+++ b/mobile/library/cc/engine_builder.h
@@ -16,8 +16,6 @@ namespace Envoy {
 namespace Platform {
 
 class EngineBuilder {
-  friend Envoy::BaseClientIntegrationTest;
-
 public:
   EngineBuilder(std::string config_template);
   EngineBuilder();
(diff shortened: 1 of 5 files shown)


mobile/library/cc/engine_builder.h
```diff
@@ -66,8 +66,6 @@ class EngineBuilder {
   std::string generateConfigStr() const;
 
   EngineSharedPtr build();
-
-protected:
```
abeyad, line 70: I'm weary of making these two functions available to any user of the EngineBuilder class.  Can we keep it protected (or private) and make BaseClientIntegrationTest a friend class of EngineBuilder? https://en.cppreference.com/w/cpp/language/friend

Friendship is not inherited in C++, but we can get around that by adding functions to BaseClientIntegrationTest that call into the EngineBuilder's protected functions, and have ClientIntegrationTest and XdsIntegrationTest call into those.

caschoener: Learn something new about c++ every day!  Sounds good.

mobile/library/cc/engine_builder.h
```diff
@@ -16,6 +16,8 @@ namespace Envoy {
 namespace Platform {
 
 class EngineBuilder {
+  friend Envoy::BaseClientIntegrationTest;
```
abeyad, line 19: can we move the friend declaration to the protected section?

abeyad (approved):

abeyad:
this has merge conflicts with main that need to be resolved

abeyad:
Added @jpsim for cross-company review

jpsim (approved):

PR #24463, opened 2022-12-09
test_coverage: fix posix test coverage

tyxia:
Coverage test is now failing with message `Code coverage for source/common/common/posix is lower than limit of 94.5 (92.7)`. Change it back to 92.7 which is the original coverage limits.


tyxia:
/assign @RyanTheOptimist 


+ Ryan as well in case Alyssa is OOO.

phlax (approved):
thanks @tyxia im gonna land this unbreak CI, lgtm

PR #24467, opened 2022-12-09
Add max_request_body_size and max_response_body_size to grpc_json_transcoder

ravenblackx:
Commit Message: Add max_request_body_size and max_response_body_size to grpc_json_transcoder
Additional Description: See #23774 for justification.
Risk Level: Low. All existing tests pass unchanged, change only takes effect if the new field is populated.
Testing: Added integration and unit tests.
Docs Changes: Yes, generated from proto.
Release Notes: Yes, in current.yaml


ravenblackx:
Thanks gcov for ensuring that I unit test my comments:
image
image

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -270,4 +270,12 @@ message GrpcJsonTranscoder {
   // Proto enum values are supposed to be in upper cases when used in JSON.
   // Set this to true if your JSON request uses non uppercase enum values.
   bool case_insensitive_enum_parsing = 14;
+
+  // The maximum size of a body to be transcoded, in bytes. A body exceeding this size will
+  // provoke a ``HTTP 413 Request Entity Too Large`` if the request body was oversized,
+  // or ``HTTP 500 Internal Server Error`` response with a descriptive message if the response
+  // body was oversized.
+  //
+  // If unset or zero, the current stream buffer size is used.
```
adisuissa, line 279: I think a primitive type cannot be unset, and if this is intended, then you can convert it to a WKT (`google.protobuf.UInt32Value`).
Also should there be a maximal value? (maybe not, but want to ensure that this is correct).

CC'ing @KBaichoo and @nareddyt to review from a filter's point-of-view and ensuring that the default behavior is correct.

changelogs/current.yaml
```diff
@@ -142,6 +142,11 @@ removed_config_or_runtime:
     removed ``envoy.reloadable_features.do_not_await_headers_on_upstream_timeout_to_emit_stats`` and legacy code paths.
 
 new_features:
+- area: grpc_json_transcoder
+  change: |
+    added ``max_message_size`` field, which can either increase or decrease the size of messages that
```
adisuissa, line 147:
```suggestion
    added ``max_body_size`` field, which can either increase or decrease the size of messages that
```

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.h
```diff
@@ -103,11 +103,21 @@ class JsonTranscoderConfig : public Logger::Loggable<Logger::Id::config>,
    */
   bool convertGrpcStatus() const;
 
+  /**
+   * @param buffer_size the size of the buffer, returned if config did not specify a size.
+   * @return the maximum size of the message body.
+   */
+  uint32_t maxBodySize(uint32_t buffer_size) const {
+    return max_body_size_ == 0 ? buffer_size : max_body_size_;
```
adisuissa, line 111: Should this be max(buffer_size, max_body_size_)?

adisuissa (commented):
Thanks. Left a high-level comment, and asking @KBaichoo and @nareddyt to take a look.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -277,5 +277,5 @@ message GrpcJsonTranscoder {
   // body was oversized.
   //
   // If unset or zero, the current stream buffer size is used.
-  uint32 max_message_size = 15;
+  uint32 max_body_size = 15;
```
qiwzhang, line 280: We need to split it into two fields:  `max_request_body_size` and max_response_body_size`

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
```diff
@@ -241,17 +240,17 @@ class GrpcJsonTranscoderIntegrationTest
 class GrpcJsonTranscoderIntegrationTestWithSizeLimit1024
     : public GrpcJsonTranscoderIntegrationTest {
 protected:
-  uint32_t maxMessageSize() const override { return 1024; }
+  uint32_t maxBodySize() const override { return 1024; }
```
qiwzhang, line 243: This is not good.  It basically duplicating the integration tests 4 times.  it is already very slow,  now it is 4 time slower.    Just simply testing smaller max_body_size did not impact the all tests.   This is not good.

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.h
ravenblackx: No, max_(request|response)_body_size_ can validly argue for accepting a smaller message size than the size of the stream buffer.

changelogs/current.yaml
ravenblackx: Done (but now `max_request_body_size` and `max_response_body_size` per qiwzhang's request).

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
ravenblackx: Done.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
ravenblackx: Unsure whether to WKT it to be explicit that unset is a special case (and then constrain it so zero is a config error), or just do what I've done; I've removed the word "unset" for now so it's unambiguous.

ravenblackx: I don't think there's any maximum - grpc itself has a soft constraint at, I think, 4 or 6MB per message, but that constraint can be overridden, so I don't think there's any justification for a hard limit here. (There's no hard limit on setDecoderBufferLimit, which when the value is high this is essentially a proxy for.)

test/extensions/filters/http/grpc_json_transcoder/grpc_json_transcoder_integration_test.cc
qiwzhang: I see.  Thanks

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -985,7 +987,8 @@ bool JsonTranscoderFilter::maybeConvertGrpcStatus(Grpc::Status::GrpcStatus grpc_
 }
 
 bool JsonTranscoderFilter::decoderBufferLimitReached(uint64_t buffer_length) {
-  uint32_t max_size = per_route_config_->maxBodySize(decoder_callbacks_->decoderBufferLimit());
+  uint32_t max_size =
+      per_route_config_->maxRequestBodySize(decoder_callbacks_->decoderBufferLimit());
```
qiwzhang, line 991: Hmm,  I don't understand the purpose of this change.  Basically,  it newly added `max_request_body_size` is too small, smaller than decodeBufferLimit(), it will reject the request.    Do we really need this feature?  
 This is making the new config `max_request_body_size` dangerous,  not easy to config.  

I prefer not to have change here.

ravenblackx: If newly added `max_request_body_size` is smaller *than the size of the request*, it will reject the request, which is presumably what anyone setting that setting would want. It doesn't reject the request just because it's smaller than `decoderBufferLimit`.

`decoderBufferLimit` is horrifically unpredictable - it's different depending whether you use http or http2, unless another filter overrides it. Which makes the behavior of this filter without this setting unpredictable and surprising, which is how #23774 ended up being an issue (someone in our organization migrated a service from http to http2, and unexpectedly the grpc transcoding started failing for large messages).

ravenblackx: To put it another way, I was aiming for the principle of least surprise. If the user of the filter explicitly sets a maximum request size of 5K, it would be surprising to receive requests of size 10K.
And (before this PR) if the user of the filter expects all grpc to be transcoded, it would be surprising to have 2MB messages rejected and have no reasonable way to configure it so they are not rejected!

qiwzhang: But with this:  users will generally put int a very big number to avoid the rejections.  this will break flow-control, envoy will use a lot of memory if we have a fast sender and slow receiver.

qiwzhang: I guess flow-control is a different issue.  I am fine then.  Please add some comment in the new fields that a large `max_body_size` may lead envoy to use too much memory if there are too many concurrent requests

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -430,6 +432,17 @@ void JsonTranscoderFilter::initPerRouteConfig() {
   per_route_config_ = route_local ? route_local : &config_;
 }
 
+void JsonTranscoderFilter::maybeExpandBufferSize() {
```
qiwzhang, line 435: maybeExpandBufferLimit?

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
ravenblackx: I think for *most* users it doesn't really matter if they set this very large, because the default buffer limit is already much larger than normal grpc messages, and putting a very large number here won't change any actual behavior under those circumstances because unless the message is larger than the original buffer limit, nothing will behave differently. (Except if there is malicious very-large-messages traffic, but setting limits to large numbers is explicitly allowing for that sort of thing!)

The problem being addressed here is that right now there is no coherent way to fix the behavior of this filter for those users who *do* want to deliver a larger message - the only way to configure a larger buffer limit with http2 is to add a buffer filter which happens to do that, which is essentially exploiting a side-effect of that filter *and* adding the unwanted effects of what that filter is supposed to do.

Anyway, requested comment about large values is added.

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -990,12 +1007,14 @@ bool JsonTranscoderFilter::decoderBufferLimitReached(uint64_t buffer_length) {
 }
 
 bool JsonTranscoderFilter::encoderBufferLimitReached(uint64_t buffer_length) {
-  if (buffer_length > encoder_callbacks_->encoderBufferLimit()) {
+  uint32_t max_size =
+      per_route_config_->maxResponseBodySize(encoder_callbacks_->encoderBufferLimit());
```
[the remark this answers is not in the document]
ravenblackx: Moved and commented.

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
ravenblackx: Done (but maybeExpandBufferLimits, even, since there's two of them.)

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
ravenblackx: Among other changes, I have now changed it to WKT with explicitly forbidden zero, because it makes some of the code more readable to use `absl::optional` and then `value_or` rather than using zero as a sentinel value. (Though it makes the tests more of a pain!)

qiwzhang (dismissed):

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -430,6 +436,17 @@ void JsonTranscoderFilter::initPerRouteConfig() {
   per_route_config_ = route_local ? route_local : &config_;
 }
 
+void JsonTranscoderFilter::maybeExpandBufferLimits() {
+  uint32_t max_request_size = per_route_config_->max_request_body_size_.value_or(0);
```
adisuissa, line 440: nit: const here and below

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -972,11 +991,15 @@ bool JsonTranscoderFilter::maybeConvertGrpcStatus(Grpc::Status::GrpcStatus grpc_
 }
 
 bool JsonTranscoderFilter::decoderBufferLimitReached(uint64_t buffer_length) {
-  if (buffer_length > decoder_callbacks_->decoderBufferLimit()) {
+  // The limit is either the configured maximum request body size, or, if not configured,
+  // the default buffer limit.
+  uint32_t max_size =
```
adisuissa, line 996: nit: const

source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
```diff
@@ -990,12 +1013,16 @@ bool JsonTranscoderFilter::decoderBufferLimitReached(uint64_t buffer_length) {
 }
 
 bool JsonTranscoderFilter::encoderBufferLimitReached(uint64_t buffer_length) {
-  if (buffer_length > encoder_callbacks_->encoderBufferLimit()) {
+  // The limit is either the configured maximum response body size, or, if not configured,
+  // the default buffer limit.
+  uint32_t max_size =
```
adisuissa, line 1018: nit: const

adisuissa (commented):
Thanks, small nits but otherwise LGTM.

/lgtm api

qiwzhang (approved):

mattklein123 (approved):

PR #24479, opened 2022-12-09
[Filesystem::File] Make truncate apply on open in Windows, as it does in Posix

ravenblackx:
Commit Message: [Filesystem::File] Make truncate apply on open in Windows, as it does in Posix
Additional Description: In Posix, opening a file with O_TRUNC causes its size to become zero on open. In the previous implementation for Windows, file size truncation occurs after write completes. As more file operations may be added to this abstraction, this inconsistency is likely to cause surprises. This change makes the Windows behavior of opening with truncate more similar to the Posix behavior.
Risk Level: Should be small since this reflects how it already works in Posix.
Testing: If it doesn't break any existing tests it should be good!
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Windows-only change.


daixiang0:
sorry that I am not familiar with Windows, could we add tests to cover it?

phlax:
@wrowe would you mind taking a look

wrowe (approved):
Much more appropriate, TY for the patch.

wrowe:
@mattklein123 I see you self-assigned this, so holding on the merge till day's end in case you had additional feedback.

wrowe:
> Right now it kind of has coverage in that tests verify that truncation happens _at some point_, which until the addition of pread/pwrite is the only aspect that can reasonably be tested. (And is the only aspect that any dependencies rely upon right now.)
> (#24480 is that change, still draft because it depends on this and chains on #24398)

Looks like #24480 will be ready to retest once this merge lands, after updating to main.

PR #24480, opened 2022-12-09
Add pread/pwrite to Envoy::Filesystem::File

ravenblackx:
Commit Message: Add pread/pwrite to Envoy::Filesystem::File
Additional Description: Currently AsyncFileManager uses OsSysCall to perform file operations. A constraint of this is that it is using file descriptors, which in Windows makes it impossible to pread/pwrite (there is no equivalent in Windows that works with file descriptors), and awkward to close (doesn't know whether to closesocket or close for a file descriptor). Migrating off OsSysCall and onto Envoy::Filesystem::File would simplify those things (because File uses a HANDLE in Windows) and also reduce the amount of special handling in AsyncFileManager.

This is a second step towards that migration, following #24398, and to be followed with support for rename and stat.

#23821 is related.
Risk Level: Low. Change is mostly adding new functions, though some care is warranted around the new `KeepExistingData` flag, which, in defaulting to false, should not change any pre-existing behavior.
Testing: 100% coverage of new code.
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: Yes, platform specific implementations.


ravenblackx:
(A small amount of this change is inherited from #24479, which could be reviewed first for simplicity.)

mattklein123 (approved):

mattklein123:
Oops looks like this needs a main merge.

/wait

PR #24475, opened 2022-12-09
Rollback #22036 due to crash

KBaichoo:
Signed-off-by: Kevin Baichoo <kbaichoo@google.com>

Commit Message: Rollback #22036 due to crash
Additional Description: See https://github.com/envoyproxy/envoy/pull/22036#discussion_r1044579979
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:

PR #24483, opened 2022-12-11
tls: SNI-based cert selection during TLS handshake (#22036)

LuyaoZhong:
Envoy supports selecting certs by selecting filter chain based on SNI. But it is possible that we access different services via one filter chain, which requires SNI-based cert selection in one single filter chain during handshake.

This change is merged by https://github.com/envoyproxy/envoy/pull/22036 and reverted by https://github.com/envoyproxy/envoy/pull/24475.


LuyaoZhong:
resubmit PR due to rollback by https://github.com/envoyproxy/envoy/pull/24475,  don't merge until https://github.com/envoyproxy/envoy/pull/22036/files#r1044579979 is addressed.

ggreenway:
@LuyaoZhong can you break this into two commits? One being the revert of the revert, which should be the original change you made, and the 2nd commit for the fix/change that is different from the original PR? You'll have to force-push to the PR branch, but that's fine in this case.

LuyaoZhong:
> @LuyaoZhong can you break this into two commits? One being the revert of the revert, which should be the original change you made, and the 2nd commit for the fix/change that is different from the original PR? You'll have to force-push to the PR branch, but that's fine in this case.

Actually I think the root cause of crash is not missing nullptr checking, since it gets checked in earlier place. I reply on https://github.com/envoyproxy/envoy/pull/22036/files#r1044579979. And I would like @yanavlasov and @KBaichoo can provide me more context about the crash case.

And I'm OK with adding checking as a second safeguard. Will append a commit for that change.

LuyaoZhong:
> > @LuyaoZhong can you break this into two commits? One being the revert of the revert, which should be the original change you made, and the 2nd commit for the fix/change that is different from the original PR? You'll have to force-push to the PR branch, but that's fine in this case.
> 
> Actually I think the root cause of crash is not missing nullptr check, since it gets checked in earlier place. I reply on https://github.com/envoyproxy/envoy/pull/22036/files#r1044579979. And I would like @yanavlasov and @KBaichoo can provide me more context about the crash case.
> 
> And I'm OK with adding checking as a second safeguard. Will append a commit for that change.

After dig the code, I got that when user has its own handshaker to provide certificates, it might trigger the bug.

KBaichoo:
Thank you @LuyaoZhong! Yes, I think the problem was with using a custom handshaker, this looks good. Can you add a test we could add to capture this?

KBaichoo (approved):
That's a fair point, thank you for fixing. It LGTM.
> @KBaichoo I think this should be tested in concrete handshaker extension test cases, but there is no any custom handshaker extension in envoy source code


cc @yanavlasov

ggreenway:
There are at least some tests that involve custom handshakes [here](https://github.com/envoyproxy/envoy/blob/main/test/extensions/transport_sockets/tls/handshaker_test.cc#L179). Can this be modified to add a test case for this?

test/extensions/transport_sockets/tls/handshaker_factory_test.cc
```diff
@@ -204,6 +205,98 @@ TEST_F(HandshakerFactoryTest, HandshakerContextProvidesObjectsFromParentContext)
...
+  std::unique_ptr<Extensions::TransportSockets::Tls::ContextManagerImpl> context_manager_;
+  envoy::extensions::transport_sockets::tls::v3::DownstreamTlsContext tls_context_;
+  TestScopedRuntime scoped_runtime_;
+  Ssl::HandshakerCapabilities capabilities_;
+};
+
+// TlsCertificate messages can be empty when handshaker provides certificates.
+TEST_F(HandshakerFactoryDownstreamTest, ServerHandshakerProvidesCertificates) {
+  HandshakerFactoryImplForDownstreamTest handshaker_factory;
+  Ssl::HandshakerCapabilities cap{cap.provides_certificates = true};
```
ggreenway, line 276:
```suggestion
  Ssl::HandshakerCapabilities cap{provides_certificates = true};
```

gcc is failing on this line

source/extensions/transport_sockets/tls/context_manager_impl.cc
```diff
@@ -18,7 +18,8 @@ namespace Tls {
 ContextManagerImpl::ContextManagerImpl(TimeSource& time_source) : time_source_(time_source) {}
 
 ContextManagerImpl::~ContextManagerImpl() {
-  KNOWN_ISSUE_ASSERT(contexts_.empty(), "https://github.com/envoyproxy/envoy/issues/10030");
+  // KNOWN_ISSUE_ASSERT(contexts_.empty(), "https://github.com/envoyproxy/envoy/issues/10030");
```
ggreenway, line 21: clang-tidy is complaining that since this is trivial you should use `~ContextManagerImpl() = delete` instead. I think you can do that and delete the comment here and the commented out assert, since the new test would trigger this assert.

ggreenway (changes requested):
This looks good; thanks for writing the test. There's a gcc error and a clang-tidy error to fix to make CI pass, but other than that I think this is ready to merge.

/wait

ggreenway (approved):

source/extensions/transport_sockets/tls/context_manager_impl.cc
LuyaoZhong: Done. Fixed CI issues, and add one more line code to make sure the test cover the updated code path.

PR #24486, opened 2022-12-12
ci: fix check for forks in `mobile_docs` job

jpsim:
The previous logic meant that we were never installing the deploy key.

E.g.: https://github.com/envoyproxy/envoy/actions/runs/3673844011/jobs/6211361666#step:7:7

Commit Message:
Additional Description:
Risk Level: No Envoy code changes, moderate risk of keeping this CI job broken in some cases.
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #24487, opened 2022-12-12
ci: fix check for forks in `mobile_docs` job

jpsim:
This is the same as https://github.com/envoyproxy/envoy/pull/24486 but from a forked repo to validate that the check to see if the job is running from a fork is working correctly.


jpsim:
Appears to have worked: https://github.com/envoyproxy/envoy/pull/24486#issuecomment-1346611883

PR #24491, opened 2022-12-12
Add fstat and stat capabilities to async_files

ravenblackx:
Commit Message: Add fstat and stat capabilities to async_files
Additional Description: For the sake of filesystem caching, it is useful to be able to measure an open file, or a non-open file, so that the size impact of deleting it can be easily accounted for in the cache size.
Risk Level: Very low. AsyncFileManager is only used in filters tagged as not-yet-safe, and this shouldn't change any live paths anyway.
Testing: 100% coverage.
Docs Changes: n/a
Release Notes: n/a
Platform Specific Features: AsyncFileHandle is already not supported in Windows.


ravenblackx:
Note status_helper changes could be reviewed separately/first in #24498

ravenblackx:
status_helper change now merged in.

mattklein123 (approved):

Issue #24373, opened 2022-12-06
Missing SDS resources for 62ab872c-038e-4311-90bc-74a77cda9f53-webserver-certs in onConfigUpdate()


AmitKatyal-Sophos:
*Title*: *Missing SDS resources error during config update*

*Description*:
We are using the XDS ADS with incremental mode (DeltaStream) instead of stow and observing an error. We have configured the the source of config for the LDS, CDS, RDS and SDS dynamic resources as ADS. We are using the go control plane snapshot cache to set the new configuration snapshot. In case of the change in the listener filter chain configuration, for example, one of the filter chain is being deleted, the corresponding SDS resource entry is also removed from the configuration. 
However, on SDS config update from the go control plane to the envoy proxy, envoy returns the below error.

*'typeUrl [type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret](http://type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret) error :code:13 message:\"Missing SDS resources for d0232574-dc99-4599-9b92-6b27f10a520d-mtls-certs in onConfigUpdate()'*

*Attaching below the response from @adisuissa  on the slack channel:*

It seems that the error comes from: https://github.com/envoyproxy/envoy/blob/d142c9d55ae9aab34e9924aa25f20bd27635e060/source/common/secret/sds_api.cc#L150
What I think happens here, is that Envoy deletes the subscription, and doesn't expect an update to the resource to be returned.
I guess the way to improve this is to do something like in RDS (https://github.com/envoyproxy/envoy/blob/5b9183f80d6c7ab38877b8f8f48b152509f49d3e/source/common/rds/rds_route_config_subscription.cc#L122). Please open an issue so we can track this.
[sds_api.cc](https://github.com/envoyproxy/envoy/blob/d142c9d55ae9aab34e9924aa25f20bd27635e060/source/common/secret/sds_api.cc)
<https://github.com/[envoyproxy/envoy](https://github.com/envoyproxy/envoy)|envoyproxy/envoy>envoyproxy/envoy | Added by [GitHub](https://envoyproxy.slack.com/services/B01UCCXGF2T)
[rds_route_config_subscription.cc](https://github.com/envoyproxy/envoy/blob/5b9183f80d6c7ab38877b8f8f48b152509f49d3e/source/common/rds/rds_route_config_subscription.cc)

*Repro steps*:
[...]

PR #24492, opened 2022-12-12
Fix for the SDS exception during the removal of the SDS resource

AmitKatyal-Sophos:
### Commit Message: 
Fix for the SDS exception during the removal of the SDS resource
### Additional Description: 
During the deletion of SDS resource, go-control-plane is sending the delta response with zero resources to add and the deleted resource to be removed. However, the existing implementation was throwing an exception, if the number of resources to be added is zero.   
Risk Level:
### Testing: 
Not done as I was not able to compile the envoy on my MAC. Tried compiling the source code with ubuntu VM using the docker CI but the build is stuck with below logs. However, I could successfully compile the envoy proxy for MAC and Arm architecture.

image
 
Docs Changes: NA
Release Notes: NA
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/secret/sds_api.cc
+++ b/source/common/secret/sds_api.cc
@@ -81,7 +81,9 @@ void SdsApi::onWatchUpdate() {
 
 void SdsApi::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& resources,
                             const std::string& version_info) {
-  validateUpdateSize(resources.size());
+  if (!validateUpdateSize(resources.size())) {
+    return;
+  }
   const auto& secret = dynamic_cast<const envoy::extensions::transport_sockets::tls::v3::Secret&>(
       resources[0].get().resource());
 
@@ -133,7 +135,9 @@ void SdsApi::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& resou
 
 void SdsApi::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& added_resources,
                             const Protobuf::RepeatedPtrField<std::string>&, const std::string&) {
-  validateUpdateSize(added_resources.size());
+  if (!validateUpdateSize(added_resources.size())) {
+    return;
+  }
   onConfigUpdate(added_resources, added_resources[0].get().version());
 }
 
@@ -144,14 +148,16 @@ void SdsApi::onConfigUpdateFailed(Envoy::Config::ConfigUpdateFailureReason reaso
   init_target_.ready();
 }
 
-void SdsApi::validateUpdateSize(int num_resources) {
+bool SdsApi::validateUpdateSize(int num_resources) {
   if (num_resources == 0) {
-    throw EnvoyException(
-        fmt::format("Missing SDS resources for {} in onConfigUpdate()", sds_config_name_));
+    ENVOY_LOG_MISC(debug, "Missing SDS resource for {} in onConfigUpdate()", sds_config_name_);
+    init_target_.ready();
+    return false;
   }
   if (num_resources != 1) {
     throw EnvoyException(fmt::format("Unexpected SDS secrets length: {}", num_resources));
   }
+  return true;
 }
 
 void SdsApi::initialize() {
--- a/source/common/secret/sds_api.h
+++ b/source/common/secret/sds_api.h
@@ -92,7 +92,7 @@ protected:
   Api::Api& api_;
 
 private:
-  void validateUpdateSize(int num_resources);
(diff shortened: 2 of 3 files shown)


AmitKatyal-Sophos:
Can someone please review the changes ? Thanks.

AmitKatyal-Sophos:
@tonya11en could you please review the changes ? Thanks.

AmitKatyal-Sophos:
@lizan  could someone please help me with the review ?

source/common/secret/sds_api.cc
```diff
@@ -133,7 +135,9 @@ void SdsApi::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& resou
 
 void SdsApi::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& added_resources,
                             const Protobuf::RepeatedPtrField<std::string>&, const std::string&) {
-  validateUpdateSize(added_resources.size());
+  if (!validateUpdateSize(added_resources.size())) {
+    return;
+  }
   onConfigUpdate(added_resources, added_resources[0].get().version());
 }
```
snowp, line 142: How is the removal process going to work with this? Seems like we just ignore the update if we get a removal without an addition? 

Is go-control-plane sending two distinct updates, first to remove (which is ignored) then another to add?

source/common/secret/sds_api.cc
```diff
@@ -144,14 +148,16 @@ void SdsApi::onConfigUpdateFailed(Envoy::Config::ConfigUpdateFailureReason reaso
   init_target_.ready();
 }
 
-void SdsApi::validateUpdateSize(int num_resources) {
+bool SdsApi::validateUpdateSize(int num_resources) {
   if (num_resources == 0) {
-    throw EnvoyException(
-        fmt::format("Missing SDS resources for {} in onConfigUpdate()", sds_config_name_));
+    ENVOY_LOG_MISC(debug, "Missing SDS resource for {} in onConfigUpdate()", sds_config_name_);
+    init_target_.ready();
```
snowp, line 154: Why this call?

snowp (changes requested):
Thanks, some questions to get this started

source/common/secret/sds_api.cc
AmitKatyal-Sophos: As per the comments on the issue to follow the RDS handling. For now, I have fixed it as per the RDS handling. 

https://github.com/envoyproxy/envoy/issues/24373

https://github.com/envoyproxy/envoy/blob/5b9183f80d6c7ab38877b8f8f48b152509f49d3e/source/common/rds/rds_route_config_subscription.cc#L120-L134

source/common/secret/sds_api.cc
AmitKatyal-Sophos: Again, modified as per the RDS handling.

source/common/secret/sds_api.cc
AmitKatyal-Sophos: Please let me know if we need to handle it differently. Thanks.

KBaichoo:
PTAL @snowp 

I think this is waiting on your input

source/common/secret/sds_api.cc
snowp: This would mean that receiving an empty SDS update would complete cluster warming, which I'm not sure if we want? I can't speak to why RDS does this, but this seems like an unnecessary change to make if we're just trying to have SDS support resource removals

cc @adisuissa to see if this makes sense from a general xDS implementation pov

source/common/secret/sds_api.cc
snowp: It seems to me like both SDS and RDS aren't really supporting xDS removals and just ignore the update in that case, so while this change would probably fix your exception issue it doesn't seem ideal? 

cc @adisuissa here as well, I'm not sure where we are with support for resource removal for the different resource types

source/common/secret/sds_api.cc
AmitKatyal-Sophos: @snowp Sorry for the rate reply & thanks for the feedback!
So, do you suggest that I should discard this PR ?

I am new to the envoy but would be happy to contribute with proper handling of the resource removal.
@adisuissa, If you already working on it then please share the details & can start with the smaller task.

phlax:
ping @snowp seems this is waiting for review

AmitKatyal-Sophos:
Any suggestion on how to go about it ?

PR #24493, opened 2022-12-12
Multi envoy unit test

caschoener:
Added support for running base_client_integration_tests with multiple envoy engines, defaulting to one engine for backwards compatibility.  And added a new test `(ClientIntegrationTest, MultipleEngines)` using the new functionality to spin up 20 instances and send a single request to each.


mobile/test/common/integration/base_client_integration_test.cc
```diff
@@ -158,41 +163,53 @@ std::shared_ptr<Platform::RequestHeaders> BaseClientIntegrationTest::envoyToMobi
   return std::make_shared<Platform::RequestHeaders>(builder.build());
 }
 
-void BaseClientIntegrationTest::threadRoutine(absl::Notification& engine_running) {
-  setOnEngineRunning([&]() { engine_running.Notify(); });
-  engine_ = build();
+void BaseClientIntegrationTest::threadRoutine(absl::BlockingCounter& engine_running) {
+  builder_.setOnEngineRunning([&]() { engine_running.DecrementCount(); });
+  for (auto i = num_engines_for_test_; i--;) {
+    multi_engines_.push_back(builder_.build());
```
caschoener, line 169: Note that this branch doesn't currently separate out the builder() but I've tried that in a separate branch and it didn't change the error produced.

mobile/test/common/integration/base_client_integration_test.h
```diff
@@ -61,17 +62,24 @@ class BaseClientIntegrationTest : public BaseIntegrationTest {
   envoyToMobileHeaders(const Http::TestRequestHeaderMapImpl& request_headers);
   Event::ProvisionalDispatcherPtr dispatcher_ = std::make_unique<Event::ProvisionalDispatcher>();
   envoy_http_callbacks bridge_callbacks_;
-  ConditionalInitializer terminal_callback_;
-  callbacks_called cc_{0, 0, 0, 0, 0, 0, 0, "", &terminal_callback_, {}};
+  std::shared_ptr<ConditionalInitializer> terminal_callback_;
```
abeyad, line 65: this doesn't need to be a shared_ptr, it can be a reference, i.e. `ConditionalInitializer&`.  You can assign `terminal_callback = multi_terminal_callbacks[0];` and it'll store the first entry in the array by reference to `ConditionalInitializer& terminal_callback`.

mobile/test/common/integration/base_client_integration_test.h
```diff
@@ -61,17 +62,24 @@ class BaseClientIntegrationTest : public BaseIntegrationTest {
...
+  std::vector<std::shared_ptr<ConditionalInitializer>> multi_terminal_callbacks_;
+  std::shared_ptr<callbacks_called> cc_;
```
abeyad, line 67: same for here, this can be a reference, not a shared_ptr.  then, all the calls below can stay `cc_.` instead of `cc_->`.

mobile/test/common/integration/client_integration_test.cc
```diff
@@ -325,20 +325,62 @@ TEST_P(ClientIntegrationTest, DirectResponse) {
...
+    multi_streams_[i]->close(trailers);
+
+    multi_terminal_callbacks_[i]->waitReady();
+
+    ASSERT_EQ(multi_cc_[i]->on_headers_calls, 1);
+    ASSERT_EQ(multi_cc_[i]->status, "200");
+    ASSERT_EQ(multi_cc_[i]->on_data_calls, 2);
+    ASSERT_EQ(multi_cc_[i]->on_complete_calls, 1);
+    ASSERT_EQ(multi_cc_[i]->on_header_consumed_bytes_from_response, 27);
+    ASSERT_EQ(multi_cc_[i]->on_complete_received_byte_count, 67);
```
abeyad, line 379: awesome, looks like you got the multi-engine data request validation working too!

mobile/test/common/integration/base_client_integration_test.h
caschoener: I tried that but got the following error, I think it's because references cannot be initialized as null? ```test/common/integration/base_client_integration_test.cc:86:28: error: constructor for 'Envoy::BaseClientIntegrationTest' must explicitly initialize the reference member 'terminal_callback_'
BaseClientIntegrationTest::BaseClientIntegrationTest(Network::Address::IpVersion ip_version,
                           ^
./test/common/integration/base_client_integration_test.h:60:27: note: declared here
  ConditionalInitializer& terminal_callback_;
                          ^
test/common/integration/base_client_integration_test.cc:86:28: error: constructor for 'Envoy::BaseClientIntegrationTest' must explicitly initialize the reference member 'cc_'
BaseClientIntegrationTest::BaseClientIntegrationTest(Network::Address::IpVersion ip_version,
```

Comment (caschoener) [reply mobile/test/common/integration/base_client_integration_test.h]: We settled on using unique and raw pointers

File: mobile/test/common/integration/client_integration_test.cc (hunk 4 above)
Comment (caschoener) [reply mobile/test/common/integration/client_integration_test.cc:379]: Yup! Finally something just worked first try, you love to see it

File: mobile/test/common/integration/base_client_integration_test.cc [hunk 5]
```diff
@@ -162,15 +175,25 @@ std::shared_ptr<Platform::RequestHeaders> BaseClientIntegrationTest::envoyToMobi
 }
 
 void BaseClientIntegrationTest::threadRoutine(absl::Notification& engine_running) {
-  builder_.setOnEngineRunning([&]() { engine_running.Notify(); });
-  engine_ = builder_.build();
+  for (auto i = 0; i < num_engines_for_test_; i++) {
```
abeyad, line 178: nit: prefer int to auto here

abeyad (approved):

abeyad:
added @jpsim for cross company review

caschoener:
Just pushed some changes to fix a tsan-detected data race with logger contexts.  Basically, when creating or deleting a loggercontext we edit this global pointer current_context, but current_context can be read by other threads.  The simplest solution is to just add a lock for accessing the context object.

Adding a lock has potential for performance concerns so I'll talk with Josh Marantz about this as he created the original code.

source/common/common/logger.cc
```diff
@@ -158,16 +158,20 @@ void DelegatingLogSink::setTlsDelegate(SinkDelegate* sink) { *tlsSink() = sink;
 SinkDelegate* DelegatingLogSink::tlsDelegate() { return *tlsSink(); }
 
 static Context* current_context = nullptr;
+static absl::Mutex current_context_mutex;
```
RyanTheOptimist, line 161: nit: Let's add ABSL_GUARDED_BY(current_context_mutex) to current_context.

source/common/common/logger.cc
```diff
@@ -195,13 +199,15 @@ void Context::activate() {
 }
 
 bool Context::useFineGrainLogger() {
+  absl::MutexLock context_lock(&current_context_mutex);
```
RyanTheOptimist, line 202: nit: ReaderMutexLock here and in other read-only uses.

RyanTheOptimist (commented):
Woo hoo!

source/common/common/logger.cc
abeyad, line 202: we are only reading `current_context` here, so I believe this can be absl::ReaderMutexLock instead.  Same for the few functions below that don't assign `current_context`.

abeyad:
> Adding a lock has potential for performance concerns so I'll talk with Josh Marantz about this as he created the original code.

Yes good point, it would be great to chat with Josh or Alyssa (who is OOO until new years) about potential performance issues with this.  I would think that the lock access gets elided in most cases so the performance implication would be minimal, but I'm not an expert enough in this to know for sure and would be great if someone else could validate this.

> One alternative would be to break the logger context up into individual variables and only do atomic operations on them. But the current design stores all previous contexts as a stack so this would be somewhat complex.

I think this needs a redesign.  Something like what Alyssa mentioned about maintaining a map of loggers instead of saving the last one in a global pointer variable.

source/common/common/logger.cc
```diff
@@ -158,16 +158,20 @@ void DelegatingLogSink::setTlsDelegate(SinkDelegate* sink) { *tlsSink() = sink;
...
 
 Context::Context(spdlog::level::level_enum log_level, const std::string& log_format,
                  Thread::BasicLockable& lock, bool should_escape, bool enable_fine_grain_logging)
     : log_level_(log_level), log_format_(log_format), lock_(lock), should_escape_(should_escape),
-      enable_fine_grain_logging_(enable_fine_grain_logging), save_context_(current_context) {
+      enable_fine_grain_logging_(enable_fine_grain_logging) {
+  absl::MutexLock context_lock(&current_context_mutex);
+  save_context_ = current_context;
   current_context = this;
   activate();
```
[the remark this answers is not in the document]
RyanTheOptimist: Fair point. We need lines 168 and 169 (saving/setting current_context) to be under the lock but maybe not activate. That being said, it seems safer to do it that way, lest we end up with some inconsistent state.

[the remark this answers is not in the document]
caschoener: getFineGrainLogContext is actually a function in another file which doesn't take this lock.  I guess I tend to agree with ryan, if I read his comment correctly, that it seems safer to hold the lock while running `activate()`. 

This also keeps consistency with the other `activate()` call in the deletion, which seems like it requires the lock.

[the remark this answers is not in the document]
abeyad: Right, but locks in C++ are not re-entrant.  So when calling `activate()`, we are still holding onto the lock guarded by `current_context_mutex`, and then in `activate()`, we call `getFineGrainLogContext()` which tries to acquire the lock guarded by the same mutex again.  Since it's not re-entrant, won't we deadlock?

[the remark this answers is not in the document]
caschoener: Chatted offline, getFineGrainLogContext doesn't do what it sounds like and is ok

abeyad:
Hey @snowp , added you as a reviewer since based on git history, it looks like you've worked in the logging code before.  Please feel free to re-assign if you feel more comfortable with someone else doing the review, thanks!

source/common/common/logger.cc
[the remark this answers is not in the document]
abeyad: Sorry it was my bad, I misread `getFineGrainLogContext()` as `getFineGrainLogFormat()` and thought it was in the same logger.cc file guarded by the same mutex, but that's not the case, so the question of re-entrant locks doesn't arise.

mobile/test/common/integration/base_client_integration_test.h
```diff
@@ -61,17 +63,24 @@ class BaseClientIntegrationTest : public BaseIntegrationTest {
   envoyToMobileHeaders(const Http::TestRequestHeaderMapImpl& request_headers);
   Event::ProvisionalDispatcherPtr dispatcher_ = std::make_unique<Event::ProvisionalDispatcher>();
   envoy_http_callbacks bridge_callbacks_;
-  ConditionalInitializer terminal_callback_;
-  callbacks_called cc_{0, 0, 0, 0, 0, 0, 0, "", &terminal_callback_, {}};
+  ConditionalInitializer* terminal_callback_;
+  std::vector<std::unique_ptr<ConditionalInitializer>> multi_terminal_callbacks_;
+  callbacks_called* cc_;
+  std::vector<std::unique_ptr<callbacks_called>> multi_cc_;
```
snowp, line 69: Maybe `callbacks_called_` and `per_engine_callbacks_called_`? The acronym is a bit ambiguous

mobile/test/common/integration/base_client_integration_test.h
```diff
@@ -61,17 +63,24 @@ class BaseClientIntegrationTest : public BaseIntegrationTest {
...
   Http::TestRequestHeaderMapImpl default_request_headers_;
   Event::DispatcherPtr full_dispatcher_;
   Platform::StreamPrototypeSharedPtr stream_prototype_;
+  std::vector<Platform::StreamPrototypeSharedPtr> multi_stream_prototypes_;
   Platform::StreamSharedPtr stream_;
+  std::vector<Platform::StreamSharedPtr> multi_streams_;
```
snowp, line 75: `per_engine_stream` or something like that? Not immediately clear that this is one active stream per engine

mobile/test/common/integration/base_client_integration_test.h
snowp, line 69: I guess another option would be to use a struct to hold each per engine object, e.g. `std::vector<PerStreamHandle> multi_engines_` and then access things via `multi_engines_[0].stream`

Not a big deal, might not make things clearer

source/common/common/logger.cc
snowp, line 202: Seems like this one is still not a ReaderMutexLock?

snowp (changes requested):
Thanks, looks reasonable! A few comments

mobile/test/common/integration/base_client_integration_test.h
caschoener: Maybe I'll expand cc in a later diff? Tons of usages which would clutter this one.

caschoener:
Added some code to work around a detected deadlock (https://pastebin.com/nSpN309G).  I really don't see how there could be a deadlock, perhaps it's a false positive.  But it's probably better to drop the lock earlier anyways.

source/common/common/logger.cc
```diff
@@ -157,20 +157,29 @@ void DelegatingLogSink::setTlsDelegate(SinkDelegate* sink) { *tlsSink() = sink;
...
 }
 
 Context::~Context() {
-  current_context = save_context_;
-  if (current_context != nullptr) {
-    current_context->activate();
+  {
+    absl::MutexLock context_lock(&current_context_mutex);
+    current_context = save_context_;
```
jmarantz, line 178: Should we first `ASSERT(current_context == this);` ? This stacked context discipline might behave strangely if you don't have strict nesting of the construct/destruct sequence so it might be better to catch a bad usage with an assert.

caschoener: Great call! in fact when I started working on this I had exactly that issue... I'll add an ASSERT with a comment to explain it.

caschoener:
Josh working towards a version without locks.  Here's a simple one where I just add a parameter "create_logger" to be used so that only one envoy engine creates a logger::context object.  It's inelegant, but the Context object does depend on the `HotRestart->logLock()` which is engine specific.  So any other approach would require a behavior change...

jmarantz:
per IM I don't think either of these (locking, or bool param saying whether to create a logger) is the right approach.

There needs to be one Logger per process (at a time). Not one per engine. Per IM, I believe we should partition the MainCommon* structures into per-process and per-engine variants, as needed, with the LoggerContext being in the per-process partition.

One open question is whether hot-restart is, or will ever be supported in E-M, as there is some interaction with the hot-restart objects to obtain a cross-process mutex.

caschoener:
What Josh said ^^. Abandoning for now

PR #24495, opened 2022-12-12
build: Moving more downstream code into the listener extension

alyssawilk:
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

part of https://github.com/envoyproxy/envoy-mobile/issues/2711


source/server/server.cc
```diff
@@ -64,6 +63,19 @@
...
 namespace Server {
+namespace {
+std::unique_ptr<ConnectionHandler> getHandler(Event::Dispatcher& dispatcher) {
+
+  auto* factory = Config::Utility::getFactoryByName<ConnectionHandlerFactory>(
+      "envoy.connection_handler.default");
+  if (factory) {
+    return factory->createConnectionHandler(dispatcher, absl::nullopt);
+  }
+  ENVOY_LOG_MISC(debug, "Unable to find envoy.connection_handler.default factory");
```
soulxu, line 74: Why not raise an exception? if returns nullptr, it will crash Envoy.

alyssawilk: this should never happen in upstream Envoy.  it can in Envoy mobile if you misconfigure things.

snowp: Would that be due to failing to register the factory? Or is there some config that drives this?

In general EM does crash on missing registration, though I guess just logging might be better?

snowp (approved):
Code move lgtm

source/server/server.cc
Augustyniak: This is hit in Envoy Mobile which makes mobile application crash after we call `addDirectResponse` on `TestEngineBuilder` (as it adds a listener) (iOS only API).

PR #24497, opened 2022-12-12
UHV: Move checks for headers with underscores to UHV when it is enabled

yanavlasov:
Commit Message:
While UHV has implemented checks for header names with underscores, the corresponding checks in codec_impl.cc was not disabled. This PR removes this check from codecs when ENVOY_ENABLE_UHV is defined, such that code in UHV is exercised.
Also added missing plumbing for the stats for header map modification done by UHV.

Risk Level: Low, main build is not affected
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A


alyssawilk (dismissed):
this feels like a lot of complexity for a solution aimed to simplify but I can't think of any improvements :-/
please do get !google sign-off as well.

lizan (approved):
API LGTM and non-Googler sign-off.

envoy/http/header_validator.h
```diff
@@ -106,8 +106,9 @@ class HeaderValidatorStats {
 public:
   virtual ~HeaderValidatorStats() = default;
 
-  virtual void incDroppedHeadersWithUnderscores() PURE;
-  virtual void incRequestsRejectedWithUnderscoresInHeaders() PURE;
+  // These can't be PURE for now, otherwise coverage in non UHV builds is unhappy.
```
alyssawilk, line 109: optional: would it make more sense to leave these pure, remove the guard around the implementation, and add guards around invocation?
If you want to leave as-is I think a TODO to remove base class impl a part of the UHV bug would be good.

alyssawilk (commented):
LGTM modulo outstanding nit.  I'll let Kevin make sure that's sorted one way or another for final stamp before merging since I'm out soon.

envoy/http/header_validator.h
yanavlasov: I agree. I realized I could fix coverage by using these methods in the ifdefed code.

adisuissa (commented):
/lgtm api

RyanTheOptimist (approved):
Coverage changes LGTM

PR #24501, opened 2022-12-12
bazel: download Go SDKs for each supported platform

jpsim:
So we can do cross-compilation on macOS hosts.

This is how the rules_go folks recommend to do this in https://github.com/bazelbuild/rules_go/issues/3402.

Commit Message:
Additional Description:
Risk Level: Low, CI should 
Testing: Tested on CI and with RBE locally on an M1 Mac.
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/bazel/dependency_imports.bzl
+++ b/bazel/dependency_imports.bzl
@@ -27,11 +27,29 @@ def envoy_dependency_imports(go_version = GO_VERSION, jq_version = JQ_VERSION, y
     go_register_toolchains(go_version)
     go_download_sdk(
         name = "go_sdk",
+        goos = "linux",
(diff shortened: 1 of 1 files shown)


phlax:
wierd failure:

```console
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcf in position 1429: invalid continuation byte
```

i think its in a dependency - seems like non-utf8 file somewhere afaict

phlax:
@keith can i pass this to you ?

keith (approved):
Looks good. We do something similar for rust

PR #24512, opened 2022-12-13
Add a feature to disable inclusion of the Envoy Mobile certificate list

RyanTheOptimist:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:

### DIFF
--- a/mobile/bazel/BUILD
+++ b/mobile/bazel/BUILD
@@ -22,6 +22,11 @@ config_setting(
     values = {"define": "include_ifaddrs=true"},
 )
 
+config_setting(
+    name = "exclude_certificates",
+    values = {"define": "exclude_certificates=true"},
+)
+
 exports_files([
     "pom_template.xml",
     "test_manifest.xml",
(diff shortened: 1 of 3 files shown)


RyanTheOptimist:
@alyssawilk PTAL. I'm not sure if we want to add this to and CI configs as the tests probably won't pass.

alyssawilk (approved):
we've got size checks CI to see how large Envoy is.  I'd think we cold at least have a build target to make sure it doesn't regress but given that's highly unlikely I'm OK with it either way.

mobile/bazel/BUILD
```diff
@@ -22,6 +22,11 @@ config_setting(
     values = {"define": "include_ifaddrs=true"},
 )
 
+config_setting(
+    name = "exclude_certificates",
+    values = {"define": "exclude_certificates=true"},
```
jpsim, line 27: Wait, the value defaults to true, shouldn't this default to `false` to match the builder defaults?

Lyft hasn't been able to successfully roll out platform cert verification FYI. There were two different crashes (proguard & use-after-free) and we haven't re-enabled it in our apps since then.

jpsim: I might be misunderstanding how this works.

jpsim: Yup, I was misunderstanding how it worked. Still off by default.

PR #24513, opened 2022-12-13
quic: splitting into client and server

alyssawilk:
Splitting build libraries into client and server variants.
Moving the server variants to be weakly linked in the HCM (via factory) such that they won't be included by E-M

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
https://github.com/envoyproxy/envoy-mobile/issues/2711


source/common/http/codec_client.cc
```diff
@@ -14,7 +14,7 @@
 #include "source/common/http/utility.h"
 
 #ifdef ENVOY_ENABLE_QUIC
-#include "source/common/quic/codec_impl.h"
+#include "source/common/quic/client_codec_impl.h"
```
RyanTheOptimist, line 17: Is there no source/common/http/codec_server.cc to correspond to this current file? I guess not. How come?

RyanTheOptimist (approved):
Awesome! Thanks for doing this. One FYI question.

source/common/http/codec_client.cc
alyssawilk: This is the client side of Envoy, creating upstream QUIC clients :-P
The server bits are in the HCM, and I turned that into factory code to avoid the direct dependency.

RyanTheOptimist (approved):

PR #24519, opened 2022-12-13
router: add x-envoy-is-timeout-retry header

angelahning:
Commit Message: add x-envoy-is-timeout-retry header on retries initiated by hitting the per try timeout. 
Risk Level: Low 
Testing: unit tests + integration tests
Docs Changes: updated docs. 
Release Notes: added. 

Signed-off-by: Angela Ning [aning@squareup.com](mailto:aning@squareup.com)


source/common/router/router.cc
```diff
@@ -5,6 +5,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <iostream>
```
mpuncel, line 8: is this needed?

source/common/router/router.cc
```diff
@@ -1064,7 +1068,7 @@ void Filter::onPerTryTimeoutCommon(UpstreamRequest& upstream_request, Stats::Cou
 void Filter::onStreamMaxDurationReached(UpstreamRequest& upstream_request) {
   upstream_request.resetStream();
 
-  if (maybeRetryReset(Http::StreamResetReason::LocalReset, upstream_request)) {
+  if (maybeRetryReset(Http::StreamResetReason::LocalReset, upstream_request, /*reachPerTryTimeout*/ false)) {
```
mpuncel, line 1071: the comment casing is inconsistent w/ the actual variable name. also should this be `is_per_try_timeout` instead of `reach_per_try_timeout`?

source/common/router/router.cc
```diff
@@ -1804,6 +1808,10 @@ void Filter::doRetry(bool can_send_early_data, bool can_use_http3) {
     downstream_headers_->setEnvoyAttemptCount(attempt_count_);
   }
 
+  if (include_timeout_retry_header_in_request_) {
+    downstream_headers_->setEnvoyIsTimeoutRetry(is_timeout_retry? "true" : "false");
```
mpuncel, line 1812: maybe it would be better to avoid setting the header altogether when it was not a timeout? as opposed to setting it with "false"?

test/integration/http_timeout_integration_test.cc
```diff
@@ -323,6 +323,85 @@ TEST_P(HttpTimeoutIntegrationTest, PerTryTimeoutWithoutGlobalTimeout) {
...
+
+  codec_client_ = makeHttpConnection(makeClientConnection(lookupPort("http")));
+  auto encoder_decoder = codec_client_->startRequest(
+      Http::TestRequestHeaderMapImpl{{":method", "POST"},
+                                     {":path", "/test_retry"},
+                                     {":scheme", "http"},
+                                     {":authority", "squareup.com"},
+                                     {"x-forwarded-for", "10.0.0.1"},
+                                     {"x-envoy-retry-on", "5xx"},
+                                     {"x-envoy-hedge-on-per-try-timeout", use_hedged_retry? "true" : "fase"},
```
mpuncel, line 341: nit: space before the `?`

source/common/router/router.cc
mpuncel: also put a space before the `?`

mpuncel: when I posted the initial comment I thought this was going on _all_ requests, not just retries. It seems reasonable to me to include the header for all retries

source/common/router/router.cc
angelahning: used when i was debugging. Deleted.

source/common/router/router.h
```diff
@@ -573,7 +573,8 @@ class Filter : Logger::Loggable<Logger::Id::router>,
                                                      const Http::HeaderMap& headers) const;
 
   void maybeDoShadowing();
-  bool maybeRetryReset(Http::StreamResetReason reset_reason, UpstreamRequest& upstream_request);
+  bool maybeRetryReset(Http::StreamResetReason reset_reason, UpstreamRequest& upstream_request,
+                       bool is_timeout_retry);
```
mattklein123, line 577: For readability, instead of using a bool and having the comments in various places I would recommend creating a 2 state enum. TimeoutRetry::Yes, TimeoutRetry::No. Same below.

mattklein123 (changes requested):
Thanks, LGTM with a small nit.

/wait

source/common/router/router.h
angelahning: updated!

mattklein123 (approved):
Thanks!

PR #24521, opened 2022-12-14
stats: reject stats by scope to reduce overhead of rejection-matching counters/gauges individually

jmarantz:
Commit Message: This PR reduces the overhead of disabled stats that use matcher-prefix `"scope_name."` by creating a `NullScope` object.

The main advantage of this is that some amount of work can be skipped over with a relatively fast exit-path on the scope. For example, when a stat is created in a disabled scope, without this optimization, a temporary new stat-name must be created by joining the name of the leaf stat with the scope name. With this optimization, that step is skipped.

When I initially wrote this I thought it was going to be more impactful because we would also skip populating a 'remember-disabled" map, but actually that part was already optimized in the case of a prefix-match with a `"."` suffix.

So I think this PR is correct and will save CPU but would be worth seeing whether its benefit is worth the added complexity. I'm guessing not, in the case of Envoy running on servers. But it might be helpful for Envoy Mobile where CPU might be a scarcer resource. I don't expect this optimization to save memory, since we were short-circuiting the map population already for prefix-matches with trailing `"."`.

@Augustyniak would you be able to patch this in and try it? I think you'd first need to convert the matcher-expressions to use prefixes ending in `"."` and then measure *that", which will actually have a pretty big impact. And then patch this PR to see if that helps beyond just optimizing the stats-matcher config?

@yurykats @stevenzzzz FYI

Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/stats/thread_local_store.cc
```diff
@@ -152,7 +152,59 @@ ScopeSharedPtr ThreadLocalStoreImpl::createScope(const std::string& name) {
   return scopeFromStatName(stat_name_storage.statName());
 }
 
+class NullScope : public Scope {
```
pradeepcrao, line 155: Curious to know how this will interact with the stats matcher. What will cause the creation of a `NullScope`

jmarantz: This needs to be thought through better, but in this PR see the implementation of createScope -- it matches the scope name against the matcher.

I think it needs to be a little bit more careful about that. Like scope "foo" should match against matcher-prefix "foo." or regex "foo\..*" or similar, but actually not against literal "foo" or prefix "foo".

Also I'm wondering about matcher mutations. There is some handling of adding a stats matcher *after* having constructed some stats. I don't think there's any handling of mutating the matcher after it has been added.

pradeepcrao: 🤦 I missed that. Makes sense.

Augustyniak:
>@Augustyniak would you be able to patch this in and try it? I think you'd first need to convert the matcher-expressions to use prefixes ending in "." and then measure *that", which will actually have a pretty big impact. And then patch this PR to see if that helps beyond just optimizing the stats-matcher config?

Yes, will be able to help in here. Should be able to report back later this week.

Augustyniak:
@jmarantz a big delay in here on my part but I just want to say that I am still planning to verify the changes with a profiler in here. I have started working on this last week but continue to run into multiple issues with Envoy Mobile set up for Android Studio IDE which prevents me from running a profiler. Will report as soon as I figure these issues out.

jmarantz:
I think there is not an obvious case that this offers enough benefit to warrant the added complexity.

PR #24529, opened 2022-12-14
fixing coverage

alyssawilk:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/run_envoy_bazel_coverage.sh
+++ b/test/run_envoy_bazel_coverage.sh
@@ -118,7 +118,7 @@ fi
 set +e
 if [[ "$VALIDATE_COVERAGE" == "true" ]] && [[ "${FUZZ_COVERAGE}" == "false" ]]; then
   echo "Checking per-extension coverage"
-  output=$(./test/per_file_coverage.sh)
+  output=$(sh ./test/per_file_coverage.sh)
 
   if [ $? -eq 1 ]; then
     echo Per-extension coverage failed:


phlax:
we could also revert the permissions on the file - so it can be run directly from its shebang

alyssawilk:
yeah I could do that as well but I wanted to make sure we didn't inadvertantly break it again.

phlax:
we probably also want  to change:

https://github.com/envoyproxy/envoy/blob/2e6b67115d2e88c97ae020c4bcf5cfa4b4515b79/test/run_envoy_bazel_coverage.sh#L123

to make it `$? -ne 0`

alyssawilk:
I did that, and it didn't make a difference locally.  Which makes _no_ sense to me because if I did echo $? it returned 126, but then skipped the -ne 0 block and said coverage passed.

phlax:
>  if I did echo $? it returned 126

if im not wrong - its the old chestnut of `echo` succeeding 8/

phlax:
does it fail if you do ...

```
RESPONSE=$?
echo $RESPONSE

if [ $RESPONSE -ne 0]; then ...
```

alyssawilk:
I don't think I tested that.  I think I did run without the echo.  no strong feelings here though, I largely just wanted to avoid a second push on something that was working becasue I want to see if the coverage numbers need to be tweaked.

test/run_envoy_bazel_coverage.sh
```diff
@@ -118,9 +118,9 @@ fi
 set +e
 if [[ "$VALIDATE_COVERAGE" == "true" ]] && [[ "${FUZZ_COVERAGE}" == "false" ]]; then
   echo "Checking per-extension coverage"
-  output=$(./test/per_file_coverage.sh)
+  output=$(sh ./test/per_file_coverage.sh)
```
phlax, line 121: i think we still need to remove the `sh ` - there is bash (eg `[[`) syntax in the target file

phlax:
```
$ sh test/per_file_coverage.sh 
test/per_file_coverage.sh: 5: Syntax error: "(" unexpected
```

alyssawilk: sh test/per_file_coverage.sh  runs for me and I prefer better handling of -x :-(

phlax: wierd - for ref this is my shell

```console
$ $SHELL --version
GNU bash, version 5.2.2(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

phlax: > sh test/per_file_coverage.sh runs for me and I prefer better handling of -x :-(

not sure iiuc - but this file cannot run with actual `sh` it has incorrect syntax - im guessing if it works then `sh` is being aliased to bash or something that implements bash syntax (my `sh` is really `dash`)

phlax (dismissed):
lgtm, thanks @alyssawilk

phlax:
v annoying shellcheck - not sure why the change is triggering this - fix can be:

```diff
diff --git a/test/run_envoy_bazel_coverage.sh b/test/run_envoy_bazel_coverage.sh
index 0fe90448b0..6ba361cb3b 100755
--- a/test/run_envoy_bazel_coverage.sh
+++ b/test/run_envoy_bazel_coverage.sh
@@ -119,8 +119,9 @@ set +e
 if [[ "$VALIDATE_COVERAGE" == "true" ]] && [[ "${FUZZ_COVERAGE}" == "false" ]]; then
   echo "Checking per-extension coverage"
   output=$(./test/per_file_coverage.sh)
+  response=$?
 
-  if [ $? -eq 1 ]; then
+  if [ $response -ne 0 ]; then
     echo Per-extension coverage failed:
     echo "$output"
     COVERAGE_FAILED=1
```

phlax (approved):

PR #24532, opened 2022-12-14
ios: fix platform key value store

Augustyniak:
Commit Message:
Additional Description: Fixes a typo kind of bug in the implementation of the iOS platform key value store and adds a test coverage for iOS platform key value store to test other parts of the implementation and avoid regressions.
Risk Level: Low
Testing: Unit tests
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/library/objective-c/BUILD
+++ b/mobile/library/objective-c/BUILD
@@ -7,7 +7,6 @@ exports_files([
 objc_library(
     name = "envoy_engine_objc_lib",
     srcs = [
-        "EnvoyAliases.h",
         "EnvoyConfiguration.h",
         "EnvoyConfiguration.m",
         "EnvoyEngineImpl.m",
@@ -23,7 +22,6 @@ objc_library(
         "EnvoyHTTPFilterFactory.m",
         "EnvoyHTTPStream.h",
         "EnvoyHTTPStreamImpl.m",
-        "EnvoyKeyValueStore.h",
         "EnvoyLogger.h",
         "EnvoyLogger.m",
         "EnvoyNativeFilterConfig.h",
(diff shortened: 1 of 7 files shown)


mobile/library/objective-c/EnvoyEngineImpl.m
```diff
@@ -403,8 +403,8 @@ static void ios_kv_store_save(envoy_data native_key, envoy_data native_value, co
     NSString *key = [[NSString alloc] initWithBytes:native_key.bytes
                                              length:native_key.length
                                            encoding:NSUTF8StringEncoding];
-    NSString *value = [[NSString alloc] initWithBytes:native_key.bytes
-                                               length:native_key.length
+    NSString *value = [[NSString alloc] initWithBytes:native_value.bytes
```
Augustyniak, line 406: this is the actual bug fix.

abeyad: good catch!  looks like it was a copy-paste bug.

abeyad (approved):
Thanks for adding the test Rafal!  I don't know Objective-C, but the fix and the contents of the test LGTM.

mobile/library/objective-c/BUILD
```diff
@@ -49,6 +47,8 @@ objc_library(
     }),
     visibility = ["//visibility:public"],
     deps = [
+        ":envoy_aliases_lib",
```
jpsim, line 50: Doesn't this dep need to go in `envoy_objc_bridge_lib` so that the header is available when building it?

Augustyniak: it's a part of `envoy_objc_bridge_lib ` already - but you are right, it's redundant here I think. Removing

jpsim (approved):

PR #24537, opened 2022-12-14
Reduce Route memory utilization by sharing the most common prefix path matchers

yurykats:
Reduce Route memory utilization by sharing the most common prefix path matchers

An analysis of a massively multi-tenant Envoy deployment shows that 75% of path matchers have either "" or "/" as a prefix. This PR creates a shared instance for both, so that all the routes can re-use the same trivial matcher instead of creating new identical ones.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/common/matchers.cc
```diff
@@ -130,12 +130,26 @@ PathMatcherConstSharedPtr PathMatcher::createExact(const std::string& exact, boo
...
+namespace {
+PathMatcherConstSharedPtr createPrefixPathMatcher(const std::string& prefix, bool ignore_case) {
   envoy::type::matcher::v3::StringMatcher matcher;
   matcher.set_prefix(prefix);
   matcher.set_ignore_case(ignore_case);
   return std::make_shared<const PathMatcher>(matcher);
 }
+static const PathMatcherConstSharedPtr emptyPrefixPathMatcher = createPrefixPathMatcher("", false);
+static const PathMatcherConstSharedPtr slashPrefixPathMatcher = createPrefixPathMatcher("/", false);
+} // namespace
```
pradeepcrao, line 142: nit: blank line before `} // namespace`

test/common/common/matchers_test.cc
```diff
@@ -432,6 +432,30 @@ TEST(PathMatcher, MatchPrefixPathIgnoreCase) {
...
+  const auto matcher3 = Envoy::Matchers::PathMatcher::createPrefix("/", true);
+  EXPECT_TRUE(matcher3->match("/bla"));
+
+  EXPECT_EQ(matcher1, matcher2);
+  EXPECT_EQ(matcher1, matcher3);
+}
+
+TEST(PathMatcher, EmptyPrefixMatcherShared) {
+  const auto matcher1 = Envoy::Matchers::PathMatcher::createPrefix("", false);
+  EXPECT_TRUE(matcher1->match("/bla"));
```
pradeepcrao, line 449: nit: can the match input not begin with a `/`?

pradeepcrao (commented):
LGTM.

source/common/common/matchers.cc
```diff
@@ -130,13 +130,30 @@ PathMatcherConstSharedPtr PathMatcher::createExact(const std::string& exact, boo
...
   envoy::type::matcher::v3::StringMatcher matcher;
   matcher.set_prefix(prefix);
   matcher.set_ignore_case(ignore_case);
   return std::make_shared<const PathMatcher>(matcher);
 }
 
+} // namespace
+
+PathMatcherConstSharedPtr PathMatcher::createPrefix(const std::string& prefix, bool ignore_case) {
+  if (prefix.empty()) {
```
adisuissa, line 144: nit: add comments explaining why this is added this way (preventing future "optimizations" to remove redundant code).

source/common/common/matchers.cc
```diff
@@ -130,13 +130,30 @@ PathMatcherConstSharedPtr PathMatcher::createExact(const std::string& exact, boo
...
+    static const PathMatcherConstSharedPtr emptyPrefixPathMatcher =
+        createPrefixPathMatcher("", false);
+    return emptyPrefixPathMatcher;
```
adisuissa, line 147: I guess that to use `CONSTRUCT_ON_FIRST_USE`, one will need to pass to it a function `createEmptyPrefixPathMatcher()` that essentially does `return createPrefixPathMatcher("", false);`.

I guess it complicates the readability of this code.

adisuissa (commented):
Thanks, LGTM, modulo small nit.
I guess that these are the only 2 "well-used" prefix matchers, or otherwise a solution that uses a cache of prefix path-matchers would be better suited.

adisuissa (approved):
LGTM, thanks!

zuercher (approved):

PR #24542, opened 2022-12-14
Allow creation of ConstSingleton using a factory callback.

pradeepcrao:
Signed-off-by: pcrao <pcrao@google.com>

Commit Message:
Additional Description:
Risk Level: Low
Testing: TODO
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/router/config_impl.h
+++ b/source/common/router/config_impl.h
@@ -688,13 +688,13 @@ public:
     return DefaultMetadata::get();
   }
   using RouteTypedMetadata = Envoy::Config::TypedMetadataImpl<HttpRouteTypedMetadataFactory>;
+  using DefaultRouteTypedMetadata = ConstSingleton<RouteTypedMetadata>;
   const Envoy::Config::TypedMetadata& typedMetadata() const override {
     if (typed_metadata_ != nullptr) {
       return *typed_metadata_;
     }
-    static const RouteTypedMetadata* defaultTypedMetadata =
-        new RouteTypedMetadata(DefaultMetadata::get());
-    return *defaultTypedMetadata;
+    return DefaultRouteTypedMetadata::get(
+        []() { return new RouteTypedMetadata(DefaultMetadata::get()); });
   }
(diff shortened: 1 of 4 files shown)


source/common/router/config_impl.h
```diff
@@ -688,13 +688,13 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
...
   using RouteTypedMetadata = Envoy::Config::TypedMetadataImpl<HttpRouteTypedMetadataFactory>;
+  using DefaultRouteTypedMetadata = ConstSingleton<RouteTypedMetadata>;
   const Envoy::Config::TypedMetadata& typedMetadata() const override {
     if (typed_metadata_ != nullptr) {
       return *typed_metadata_;
     }
-    static const RouteTypedMetadata* defaultTypedMetadata =
-        new RouteTypedMetadata(DefaultMetadata::get());
-    return *defaultTypedMetadata;
+    return DefaultRouteTypedMetadata::get(
```
yurykats, line 696: I personally find this rather confusing, and not very readable -- you can call get() with different parameters, but only the 1st would matter (which is what your unit test demonstrates nicely). There is no way to tell in what order they would be called, so no real way to know what kind of singleton you are getting.

jmarantz: Yury's PR does not depend on this anyway, right?

yurykats: No, it doesn't. And this approach won't work in my case for the exact same reason --  I needed 2 singleton instances of the same type, but created with different parameters.

pradeepcrao: You're right, this can be very confusing and hard to reason about. The only way this could work is if `get()` never takes arguments, and the factory is somehow captured at the point that the variable is declared.

jmarantz: What do you want to do going forward this PR then?

pradeepcrao: I don't have a good solution. I'll close the PR.

PR #24543, opened 2022-12-15
Clean up mobile/library/common/jni/BUILD targets

### LINKED ISSUE #24544: Envoy Mobile jni BUILD targets are confusing
`jni_version.cc` and `jni_utility.cc` are listed in srcs for multiple build targets. This ends up with the potential for duplicate symbols. Here's a dependency graph of some of the targets involved.

Screenshot 2022-12-14 at 5 13 02 PM

`jni_interface.cc` is also listed in srcs for multiple build targets. While this doesn't appear to cause problems it's still a bitunorthodox.

Finally, there is an annoying warning that we should clean up:
```
WARNING: /usr/local/google/home/rch/src/envoy/mobile/library/common/jni/BUILD:95:27: target 'libndk_envoy_jni.jnilib' is both a rule and a file; please choose another name for the rule
```

RyanTheOptimist:

- Remove unnecessary copts now that c++17 is the default.
- Rename libndk_envoy_jni.jnilib -> libndk_envoy_jni_jnilib to squelch warning
- Use ":base_java_jni_lib" instead of "base_java_jni_lib", per standards
- Depend on jni_utility_lib in base_java_jni_lib instead of duplicating the srcs

Fixes #24544

### DIFF
--- a/mobile/library/common/jni/BUILD
+++ b/mobile/library/common/jni/BUILD
@@ -27,10 +27,26 @@ cc_library(
 )
 
 cc_library(
+    name = "jni_interface_lib",
+    srcs = [
+        "jni_interface.cc",
+    ],
+    deps = [
+        ":android_network_utility_lib",
+        ":jni_utility_lib",
+        ":ndk_jni_support",
+        "//library/common:envoy_main_interface_lib",
+        "//library/common/api:c_types",
+        "//library/common/types:managed_types_lib",
+    ],
+    # We need this to ensure that we link this into the .so even though there are no code references.
+    alwayslink = True,
+)
+
+cc_library(
     name = "envoy_jni_lib",
     srcs = [
         "android_jni_interface.cc",
-        "jni_interface.cc",
     ],
     linkopts = [
         "-lm",
@@ -44,6 +60,7 @@ cc_library(
     }),
     deps = [
         ":android_network_utility_lib",
+        ":jni_interface_lib",
         ":jni_utility_lib",
         ":ndk_jni_support",
         "//library/common:envoy_main_interface_lib",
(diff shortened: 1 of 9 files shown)


mobile/library/common/jni/BUILD
```diff
@@ -95,7 +93,7 @@ android_debug_info(
 ## Targets for local execution
 # OS X binary (.jnilib) for NDK testing
 envoy_mobile_so_to_jni_lib(
-    name = "libndk_envoy_jni.jnilib",
```
keith, line 98: pretty sure this name is important for testing on macOS, @jpsim or @Augustyniak can verify

jpsim: Seeing as how CI runs tests that go through the JNI on macOS, I think that's sufficient coverage, but I'll test locally to be sure.

jpsim:
> Remove unnecessary copts now that c++17 is the default.

Does that apply to the NDK toolchain too? With your branch, if I pass `-s` to a bazel test command, I see both `-std=c++11` and `-std=c++17` in the build arguments, which I find puzzling.

keith:
That's probably bazel's default + our .bazelrc, so as long as the 17 is later it should be fine

jpsim:
> That's probably bazel's default + our .bazelrc, so as long as the 17 is later it should be fine

Yes, the 17 is later.

mobile/library/common/jni/BUILD
jpsim: `./bazelw test --define=signal_trace=disabled //test/java/...` still passes for me locally on macOS.

RyanTheOptimist:
Thanks for the double-check, @jpsim and thanks for the comments @keith!

jpsim:
I don't see any issues with this from my testing.

While you're here, maybe we can remove this, which appears to have only been needed for Rust, which Lyft doesn't use in Envoy Mobile anymore: https://github.com/envoyproxy/envoy/blob/eeab187b21e540dbe47b83d023b94f4682fb1179/mobile/library/common/jni/BUILD#L80

PR #24550, opened 2022-12-15
ci: Set fetch depth/tags for `env` job explicitly

### LINKED ISSUE #24551: Speed up AZP CI by switching default git config to `fetchDepth=1` and `fetchTags=false`
Newly created AZP repos have the defaults for these settings reversed

We should update to use the faster shallow checkouts

Some jobs may actually need a full checkout so in these cases we need to set it explicitly before switching the default

i have begun some testing on this here #24550

phlax:
this will allow us to flip the defaults for all of the other jobs and speed up CI

Fix #24551 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -66,6 +66,9 @@ stages:
     pool:
       vmImage: "ubuntu-20.04"
     steps:
+    - checkout: self
+      fetchDepth: 0
+      fetchTags: true
     - bash: |
         set -e
 


phlax:
once this lands i will flip the default values for these

cc @lizan

lizan (approved):

PR #24556, opened 2022-12-15
Remove unused linkopt on libenvoy_jni.so

RyanTheOptimist:

Remove unused linkopt on libenvoy_jni.so

Risk Level: Low
Testing: Existing
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/mobile/library/common/jni/BUILD
+++ b/mobile/library/common/jni/BUILD
@@ -90,10 +90,6 @@ cc_binary(
     }) + select({
         "@envoy//bazel:android": ["-llog"],
         "//conditions:default": [],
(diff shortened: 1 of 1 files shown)


jpsim (approved):
+1 since I'm pretty sure this was only used when Lyft injected Rust code into the Envoy Mobile binary, which we no longer do.

RyanTheOptimist:
> +1 since I'm pretty sure this was only used when Lyft injected Rust code into the Envoy Mobile binary, which we no longer do.

Thanks, @jpsim!

keith (approved):

PR #14493, opened 2020-12-21
tracing: tracer integration with Apache SkyWalking SDK

Shikugawa:
For an explanation of how to fill out the fields, please see the relevant section
in [PULL_REQUESTS.md](https://github.com/envoyproxy/envoy/blob/master/PULL_REQUESTS.md)

Commit Message: Fix #13799. This means integration with external Apache SkyWalking SDK for C++ and destroyed redundant parts of implementations. Hired library is https://github.com/skyAPM/cpp2sky. But it is not released and some of missing parts of implementations which are necessary to achieve clean integration. Currently working private fork of that SDK here. https://github.com/Shikugawa/cpp2sky-envoy. 

Additional Description:

cpp2sky dependency policy check list

### MUST
- [x] Cloud Native Computing Foundation (CNCF) approved license
- [x] Dependencies must not substantially increase the binary size unless they are optional (i.e. confined to specific extensions)
- [x] No duplication of existing dependencies
       Maybe, the dependencies are described [here](https://github.com/SkyAPM/cpp2sky/blob/main/bazel/repositories.bzl)
- [x] Hosted on a git repository and the archive fetch must directly reference this repository. We will NOT support intermediate artifacts built by-hand located on GCS, S3, etc.
- [x] CVE history appears reasonable, no pathological CVE arcs
- [x] Code review (ideally PRs) before merge
- [x] Security vulnerability process exists, with contact details and reporting/disclosure process
- [x] Tests run in CI

### SHOULD
- [ ] > 1 contributor responsible for a non-trivial number of commits
- [ ] High test coverage (also static/dynamic analysis, fuzzing)
      Parts which is used in Envoy have higher test coverage. But gRPC async segment reporter(which is not used in Envoy) has less test.
- [ ] Envoy can obtain advanced notification of vulnerabilities or of security releases
- [ ] Do other significant projects have shared fate by using this dependency?
- [x] Releases (with release notes)
- [x] Commits/releases in last 90 days

Risk Level: Mid
Testing: Unit/Integration
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

PR #24558, opened 2022-12-15
fix: catched all exceptions from cpp2sky to avoid possible runtime exception

wbpcode:
Commit Message: fix: catched all exceptions from cpp2sky to avoid possible runtime exception
Additional Description:

Influence range: this bug was introduced by the #[14493](https://github.com/envoyproxy/envoy/pull/14493). So, all versions after 1.18 have this bug.
How to reproduce: enable skywalking tracer and send a request with illegal sw8 to envoy. The sw8 header should contains 8 fields and the first field should be legal and fouth field shoul be illegal. Then envoy will crash.

Risk Level: low.
Testing: unit.
Docs Changes: n/a.
Release Notes: added.


wbpcode:

seems that @lizan is too busy recently to handle this PR.

snowp (approved):
Seems reasonable to, thanks! Is the plan to eventually fix upstream?

wbpcode:
> Seems reasonable to, thanks! Is the plan to eventually fix upstream?

It's in my TODO list

wilsonwu:
This problem maybe as a vulnerability for cyber attack, suggest pick to nearly 3 release and publish hotfix.

PR #24571, opened 2022-12-15
rate_limit_quota: Implement the on_no_match logic and refactor the code

tyxia:
- Reorganized and moved all matching related code to separate matcher lib {cc|h}.
- Refactored `requestMatching` function : Change the return value from `StatusOr<BucketId>` to 
  `StatusOr<Matcher::ActionPtr>` so that the matching action is moved out and performed on the caller site.
- Implemented `on_no_match` and its action. This can be used for default security policy.  
- Added partial implementation of decodeHeaders: perform request matching; generate bucket id. More stuff (like quota 
   cache) will be added in next PRs.
- Added the unit tests.
 

Signed-off-by: tyxia [tyxia@google.com](mailto:tyxia@google.com)


RyanTheOptimist:
@yanavlasov friendly maintainer on-call ping

source/extensions/filters/http/rate_limit_quota/filter.cc
```diff
@@ -2,28 +2,39 @@
...
                                                               bool) {
-  // TODO(tyxia) Create the rate limit gRPC client and start the stream on the first request.
-  absl::StatusOr<BucketId> match_result = requestMatching(headers);
-
-  // Request is not matched by any matchers. In this case, requests are ALLOWED by default (i.e.,
-  // fail-open) and will not be reported to RLQS server.
+  // First, perform the request matching.
+  absl::StatusOr<Matcher::ActionPtr> match_result = requestMatching(headers);
   if (!match_result.ok()) {
+    // When the request is not matched by any matchers, it is ALLOWED by default (i.e., fail-open)
```
yanavlasov, line 17: We will definitely want to add a counter here to track fail open requests.

yanavlasov (changes requested):
/wait

source/extensions/filters/http/rate_limit_quota/filter.cc
tyxia: Very good point. Is it ok that I add the stats in a separate PR? Because it will involve set up stats infra e.g., Stats::Scope, COUNTER enum. Also, I want to scan through the whole filter and add more stats than just this place (if needed).

I added a TODO here for stats. Please let me know WDYT.

yanavlasov (approved):

PR #24580, opened 2022-12-16
http2: use getStreamUnchecked wherever possible

vehre-x41:
Commit Message: http2: use getStreamUnchecked wherever possible
Additional Description:
Let h2 codec use the non-asserting getStream() function, because it
handles no stream gracefully, but the asserting one blocks fuzzing.

- Use getStreamUnchecked() more often where stream == nullptr is in
    place anyway.
- Add logging for some places.
- Add const getStreamUnchecked().
    
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/http2/codec_impl.cc
+++ b/source/common/http/http2/codec_impl.cc
@@ -1007,7 +1007,6 @@ Http::Status ConnectionImpl::dispatch(Buffer::Instance& data) {
 }
 
 const ConnectionImpl::StreamImpl* ConnectionImpl::getStream(int32_t stream_id) const {
-
   // Delegate to the non-const version.
   return const_cast<ConnectionImpl*>(this)->getStream(stream_id);
 }
@@ -1436,7 +1435,7 @@ int ConnectionImpl::onMetadataReceived(int32_t stream_id, const uint8_t* data, s
   StreamImpl* stream = getStreamUnchecked(stream_id);
   if (!stream || stream->remote_end_stream_) {
     if (!stream) {
-      ENVOY_CONN_LOG(trace, "no stream for stream_id {} while receiving METADATA", connection_,
+      ENVOY_CONN_LOG(debug, "no stream for stream_id {} while receiving METADATA", connection_,
                      stream_id);
     }
     return 0;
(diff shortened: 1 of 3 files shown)


vehre-x41:
I tried to increase coverage. But the error-messages I have added are probably unreachable with the interfaces given. I would need to drill open the classes interface's to much to be able to produce these error messages. (A simple approach I tried, failed due to getStream() not being virtual. But just for testing I don't make it virtual. Too expensive). So what can I do? I tried a rebase. Writing tests for code I can only guess the intention for is not to my liking.

tyxia:
> here

From the report https://storage.googleapis.com/envoy-pr/d093207/coverage/source/common/http/http2/codec_impl.h.gcov.html, looks like the code comments (those highlighted in RED) confused the coverage checker. It is checker's problem, but you can try to simplify(or even remove) the comment to make the coverage checker happy.

source/common/http/http2/codec_impl.cc
```diff
@@ -2011,9 +2024,12 @@ Status ClientConnectionImpl::trackInboundFrames(const nghttp2_frame_hd* hd,
     ENVOY_CONN_LOG(trace, "error reading frame: {} received in this HTTP/2 session.", connection_,
                    result.message());
     if (isInboundFramesWithEmptyPayloadError(result)) {
-      ConnectionImpl::StreamImpl* stream = getStream(hd->stream_id);
+      ConnectionImpl::StreamImpl* stream = getStreamUnchecked(hd->stream_id);
       if (stream) {
```
adisuissa, line 2028: @yanavlasov would like to hear your thoughts as original code author.
Is the returned `stream` always expected to be non-null? If so, should this if statement change to an assert?

Context: the else-clause in this PR is failing coverage report, and @vehre-x41 could not find a way to trigger this path.

yanavlasov: For client codec the stream can never be nullptr here, since it is created right before the codec allocates stream_id. The codec would have failed before this point if it got response for unknown stream_id.

To be  on the defensive side I think the else statement can simply be omitted here. Maybe just add a comment that the `if` statement is defensive.

yanavlasov (approved):

vehre-x41:
Finally managed to get pipelines pass. \o/

RyanTheOptimist:
@yanavlasov friendly maintainer on-call ping

tyxia:
I think we can just merge this PR as Yan already approved it before.

vehre-x41:
I don't know if you can merge it w/o a reviewer approving, but I can't . The merge button is grey'ed out for me.

yanavlasov (approved):

PR #24589, opened 2022-12-16
Reduce Route memory utilization by avoiding ConnectConfig instances when not needed

yurykats:
Reduce Route memory utilization by avoiding ConnectConfig instances when not needed.
    
Currently an instance of `ConnectConfig` (wrapped in `optional`) is created in each route, even if not configured. This PR switches the instance to `unique_ptr` and only creates `ConnectConfig` objects when actually configured.
    
To facilitate the refactoring, `connectConfig()` API changes from `absl::optional<ConnectConfig>&` to `ConnectConfigOptRef`. Most of the PR changes are to reflect that, updating tests and mocks to the new  signature.
    
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/upstreams/http/tcp/upstream_request.cc
+++ b/source/extensions/upstreams/http/tcp/upstream_request.cc
@@ -50,7 +50,7 @@ Envoy::Http::Status TcpUpstream::encodeHeaders(const Envoy::Http::RequestHeaderM
   ASSERT(route_entry != nullptr);
   if (route_entry->connectConfig().has_value()) {
     Buffer::OwnedImpl data;
-    auto& connect_config = route_entry->connectConfig();
+    const auto& connect_config = route_entry->connectConfig();
     if (connect_config->has_proxy_protocol_config() &&
         upstream_request_->connection().has_value()) {
       Extensions::Common::ProxyProtocol::generateProxyProtoHeader(


source/extensions/upstreams/http/tcp/upstream_request.cc
```diff
@@ -51,10 +51,11 @@ Envoy::Http::Status TcpUpstream::encodeHeaders(const Envoy::Http::RequestHeaderM
   ASSERT(route_entry != nullptr);
   if (route_entry->connectConfig().has_value()) {
     Buffer::OwnedImpl data;
-    auto& connect_config = route_entry->connectConfig().value();
-    if (connect_config.has_proxy_protocol_config() && upstream_request_->connection().has_value()) {
+    auto& connect_config = route_entry->connectConfig();
```
adisuissa, line 54: nit: const

source/common/router/config_impl.cc
```diff
@@ -634,7 +634,7 @@ RouteEntryImplBase::RouteEntryImplBase(const VirtualHostImpl& vhost,
       throw EnvoyException(absl::StrCat("Duplicate upgrade ", upgrade_config.upgrade_type()));
     }
     if (upgrade_config.upgrade_type() == Http::Headers::get().MethodValues.Connect) {
-      connect_config_ = upgrade_config.connect_config();
+      connect_config_ = std::make_unique<ConnectConfig>(upgrade_config.connect_config());
```
adisuissa, line 637: I was wondering whether this can be further optimized by surrounding with an `if (upgrade_config.has_value())`, but looking at the for loop around this whole thing I think that at the moment if there are multiple upgrade_configs and the last hasn't got a connect_config, then the `connect_config_` will be set as empty.

adisuissa (commented):
Small nit, but otherwise LGTM, thanks!

source/common/router/config_impl.cc
yurykats: Not sure I follow. This seems to be finding the only Connect config, I didn't change that semantics. It throws out of the loop if it finds duplicate types (so only one Connect is possible), and if no Connects are present, then it will remain empty.  Not sure what you mean by `upgrade_config.has_value()` check.

source/extensions/upstreams/http/tcp/upstream_request.cc
yurykats: Done

adisuissa (approved):
LGTM, thanks!

source/common/router/config_impl.cc
yurykats: Ah, I see, thanks for the clarification. That is indeed orthogonal, and I assume a default ConnectConfig would do in this case (but no idea really). The connect_config field in the proto is documented as "Note that CONNECT support is currently considered alpha in Envoy".

zuercher (approved):

PR #24600, opened 2022-12-16
mobile: add `./remote` helper

jpsim:
Envoy Mobile project developers can use EngFlow's RBE to speed up and offload local builds to EngFlow's remote execution cluster.

Most bazel commands should work by replacing `./bazelw` with `./remote`.

For example:

```console
$ ./remote build --config ios ios_dist
$ ./remote build --config android android_dist
```

With certificates places in the `mobile/tmp/certs` directory (which is ignored by git), it should "just work".

Note that `protoc` commands fail with a "Bad CPU type in executable" error message when using RBE on an M1 Mac with an arm64 bazel binary, so to work around this we tell `bazelw` to use the amd64 version, which appears to work just fine.

Commit Message:
Additional Description:
Risk Level: Low, only used for local development.
Testing: Local on an M1 Mac.
Docs Changes: Not adding docs for this since it's only going to be used by Envoy Mobile maintainers.
Release Notes: N/A
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/remote
+++ b/mobile/remote
@@ -15,7 +15,7 @@ else
 fi
 
 if [[ $OSTYPE == darwin* ]]; then
-  # TODO(jpsim): RBE fails with arm64 versions of Bazel on macOS
+  # TODO(#24605): RBE fails with arm64 versions of Bazel on macOS
   BAZELW_ARCH=amd64 ./bazelw "$@" \
     --tls_client_certificate="$certificate" \
     --tls_client_key="$key" \
--- a/mobile/remote
+++ b/mobile/remote
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -euo pipefail
+
+script_root="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+certificate="$script_root/tmp/certs/engflow.crt"
+key="$script_root/tmp/certs/engflow.key"
+
+if [[ -f "$certificate" && -f "$key" ]]; then
+  echo "Using EngFlow RBE"
+else
+  echo "EngFlow RBE certificate and key not installed in $certificate and $key"
+  echo "Contact Andrés Felipe Barco Santa <andres@engflow.com> for access to these files"
+  exit 1
+fi
+
+if [[ $OSTYPE == darwin* ]]; then
+  # TODO(jpsim): RBE fails with arm64 versions of Bazel on macOS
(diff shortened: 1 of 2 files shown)


mobile/remote
```diff
@@ -0,0 +1,28 @@
...
+  echo "Using EngFlow RBE"
+else
+  echo "EngFlow RBE certificate and key not installed in $certificate and $key"
+  echo "Contact Andrés Felipe Barco Santa <andres@engflow.com> for access to these files"
+  exit 1
+fi
+
+if [[ $OSTYPE == darwin* ]]; then
+  # TODO(jpsim): RBE fails with arm64 versions of Bazel on macOS
+  BAZELW_ARCH=amd64 ./bazelw "$@" \
```
RyanTheOptimist, line 19: If, as per the comment on line 18, RBE fails with amd64, I'm surprised that we're setting BAZELW_ARCH=amd64 here. Can you say more abut this?

RyanTheOptimist (approved):
Nice!

mobile/remote
jpsim: I'll add more context.

What happens is that if you run a `bazel build` command on an M1 Mac host, at some point it tries to execute an arm64 binary of `protoc` on the amd64 remote exec machine, leading to [this error](https://envoy.cluster.engflow.com/invocation/30cc6fcc-8c30-46b0-b616-48af66cec8a9#console):

```
sandbox-exec: execvp() of 'bazel-out/.../com_google_protobuf/protoc' failed: Bad CPU type in executable
```

By forcing the host to use an amd64 binary of bazel (which can execute via Rosetta on an arm64 Mac), an amd64 `protoc` binary is used on the amd64 remote exec machine, preventing the error.

I might be misinterpreting what's happening here, but in the end forcing amd64 works.

jpsim: I'm working with @jayconrod from EngFlow over email to dig more into this.

RyanTheOptimist: Oh, man! I read this like 5 times and FINALLY realized that "amd64" != "arm64". *facepalm*

jpsim: I filed https://github.com/envoyproxy/envoy/issues/24605 to track.

RyanTheOptimist (approved):
Oh weird, PR bot didn't remind me about this and I didn't see it in my dashboard. Re-approving. Sorry about the delay

PR #24601, opened 2022-12-16
bazel: add version check to WORKSPACE

keith:
This ensures that the minimum bazel version you can use is the one in the .bazelversion file. The default function for this doesn't work because it doesn't compare anything after the version so `6.0.0rc3` == `6.0.0rc4`, which isn't realistic, especially for rolling releases.

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -74,6 +74,7 @@ def _envoy_repo_impl(repository_ctx):
     api_version_path = repository_ctx.path(repository_ctx.attr.envoy_api_version)
     version = repository_ctx.read(repo_version_path).strip()
     api_version = repository_ctx.read(api_version_path).strip()
+    fail("hi")
     repository_ctx.file("version.bzl", "VERSION = '%s'\nAPI_VERSION = '%s'" % (version, api_version))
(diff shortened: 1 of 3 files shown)


keith:
RFC to avoid having to debug bazel version issues

phlax (commented):
couple of thoughts

can we call it something like `bazel_version_check` so we disambiguate what is being checked

the other thought is - we have an `@envoy_repo` repository_ctx already - not sure if we want to make it part of that

keith:
> can we call it something like bazel_version_check so we disambiguate what is being checked

done!

> the other thought is - we have an @envoy_repo repository_ctx already - not sure if we want to make it part of that

looks like that isn't evaluated if you just `bazel build envoy`, so I don't think it would work for all cases

bazel/repositories.bzl
```diff
@@ -74,6 +74,7 @@ def _envoy_repo_impl(repository_ctx):
     api_version_path = repository_ctx.path(repository_ctx.attr.envoy_api_version)
     version = repository_ctx.read(repo_version_path).strip()
     api_version = repository_ctx.read(api_version_path).strip()
+    fail("hi")
```
yanavlasov, line 77: Is this leftover debugging?

yanavlasov (commented):
/wait-any

PR #24604, opened 2022-12-16
Recommend filters not decrease buffer limits

ravenblackx:
Commit Message: Recommend filters not decrease buffer limits
Additional Description: Just changing `flow_control.md` to recommend not calling `set(En|De)coderBufferLimit` when it would decrease the buffer limit, as that's likely to conflict with other filters, vs. increasing the limit is unlikely to cause conflicts. This is inspired by #23774, in which it turned out that a code solution to this awkwardness is not reasonable, so a documentation solution is the next best thing.
Risk Level: None, it's documentation.
Testing: No, it's documentation.
Docs Changes: Yes, it is.
Release Notes: n/a

### DIFF
--- a/envoy/http/filter.h
+++ b/envoy/http/filter.h
@@ -662,6 +662,12 @@ public:
   /**
    * This routine may be called to change the buffer limit for decoder filters.
    *
+   * It is recommended (but not required) that filters calling this function should
+   * generally only perform increases to the buffer limit, to avoid potentially
+   * conflicting with the buffer requirements of other filters in the chain, i.e.
+   *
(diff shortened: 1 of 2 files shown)


KBaichoo (commented):
can you update the documentation for in https://github.com/envoyproxy/envoy/blob/abdbc449c61d43b6d0f9624d77bf57ffdb37d465/envoy/http/filter.h#L1002 for both `set(En|De)coderBufferLimit` ? I think callers would stumble upon that more easily.

Thank you!

/wait

KBaichoo (approved):

PR #24614, opened 2022-12-17
Reduce Route memory utilization by avoiding extra copies of matcher strings

yurykats:
Reduce Route memory utilization by avoiding extra copies of matcher strings

Multiple sub-classes of RouteEntryImplBase (such as PrefixRouteEntryImpl) make two copies of the same match criterion string -- one copy is kept inside the PathMatcher they create, and another is kept as a class member. This PR avoids duplication by eliminating the criteria as the class member, and using a reference to PathMather's copy.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/common/matchers.cc
+++ b/source/common/common/matchers.cc
@@ -161,23 +161,5 @@ bool PathMatcher::match(const absl::string_view path) const {
   return matcher_.match(Http::PathUtil::removeQueryAndFragment(path));
 }
 
-const std::string& PathMatcher::criterion() const {
-  switch (matcher_.matcher().match_pattern_case()) {
-  case envoy::type::matcher::v3::StringMatcher::kPrefix:
-    return matcher_.matcher().prefix();
-  case envoy::type::matcher::v3::StringMatcher::kExact:
-    return matcher_.matcher().exact();
-  case envoy::type::matcher::v3::StringMatcher::kSuffix:
-    return matcher_.matcher().suffix();
-  case envoy::type::matcher::v3::StringMatcher::kSafeRegex:
-    return matcher_.matcher().safe_regex().regex();
-  case envoy::type::matcher::v3::StringMatcher::kContains:
-    return matcher_.matcher().contains();
-  case envoy::type::matcher::v3::StringMatcher::MATCH_PATTERN_NOT_SET:
-    break;
-  }
-  PANIC("unexpected");
-}
-
 } // namespace Matchers
 } // namespace Envoy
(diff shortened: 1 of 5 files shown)


source/common/router/config_impl.cc
```diff
@@ -1663,8 +1661,10 @@ PathSeparatedPrefixRouteEntryImpl::matches(const Http::RequestHeaderMap& headers
   }
   absl::string_view sanitized_path = sanitizePathBeforePathMatching(
       Http::PathUtil::removeQueryAndFragment(headers.getPathValue()));
-  if (sanitized_path.size() >= prefix_.size() && path_matcher_->match(sanitized_path) &&
-      (sanitized_path.size() == prefix_.size() || sanitized_path[prefix_.size()] == '/')) {
+  size_t sanitized_size = sanitized_path.size();
+  size_t matcher_size = matcher().size();
```
adisuissa, line 1665: nit: const

source/common/common/matchers.cc
```diff
@@ -161,5 +161,23 @@ bool PathMatcher::match(const absl::string_view path) const {
   return matcher_.match(Http::PathUtil::removeQueryAndFragment(path));
 }
 
+const std::string& PathMatcher::criterion() const {
```
adisuissa, line 164: If possible I think the pattern where all matcher pattern types are explicitly listed here should be avoided, as it introduces another place in the code where one needs to add a case when adding a new matcher type (if any will be added in the future).

Can this instead be done either by directly fetching the correct matcher field in the places where `criterion()` is called? (for example: in `PrefixRouteEntryImpl` the call to `matcher()` will directly return `path_matcher_->matcher_.matcher().prefix()`, [maybe matcher_ will need a const getter])

adisuissa (commented):
Thanks, overall LGTM, left small comments.

source/common/common/matchers.cc
yurykats: Yeah, that sounds better, I agree. Although the accessors become the long chain of "matchers" :) `path_matcher_->matcher().matcher().prefix()`. Will change.

source/common/router/config_impl.cc

adisuissa (approved):
LGTM, thanks!

yanavlasov (approved):

PR #24624, opened 2022-12-19
swift/kotlin: add enable dns cache method

Augustyniak:
Commit Message:
Additional Description:
Risk Level: Low, optional opt-in feature.
Testing: unit tests
Docs Changes:
Release Notes: Updated.
Platform Specific Features: iOS only.
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim:
The config yaml insert seems pretty language/platform agnostic. Is there any reason to not expose similar functions on the C++ & Kotlin engine builders? They all have KV store builders now too.

Augustyniak:
> The config yaml insert seems pretty language/platform agnostic. Is there any reason to not expose similar functions on the C++ & Kotlin engine builders? They all have KV store builders now too.

AFAIK it's not possible until we do not change iOS/Android config builders so that they call into a c++ builder (where the shared logic could live).

Augustyniak:
>They all have KV store builders now too.
I do not want to tie KV store builders API with DNS cache APIs as platform key value store can be potentially used for more than just DNS cache.

Augustyniak:
> I meant can we add a `enableDNSCache()` function to the Kotlin and C++ builders? I didn't mean to move the yaml generation upstream or anything.

Ah, yes - that we could do. I just wanted to verify that Android side of things actually works before merging it and to speed things up I decided to go with iOS first. Going to follow up with Android/c++ next if you are OK with merging iOS first.

jpsim:
> Ah, yes - that we could do. I just wanted to verify that Android side of things actually works before merging it and to speed things up I decided to go with iOS first. Going to follow up with Android/c++ next if you are OK with merging iOS first.

Yes, I'm ok starting with the place we actually want to use this first and following up with the others. It's good to have this written down though so we know there are no known blockers to porting the API over to the other languages.

mobile/library/objective-c/EnvoyConfiguration.m
```diff
@@ -192,6 +194,12 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
 
   [definitions appendFormat:@"%@\n", cert_validator_template];
 
+  if (self.enableDNSCache) {
+    NSString *dns_cache_config =
+        [[NSString alloc] initWithUTF8String:dns_cache_key_value_config_insert];
```
jpsim, line 199: Nit:

```suggestion
    NSString *dns_cache_config = @(dns_cache_key_value_config_insert);
```

mobile/library/swift/EngineBuilder.swift
```diff
@@ -171,6 +172,20 @@ open class EngineBuilder: NSObject {
     return self
   }
 
+  /// Specify whether to enable DNS cache.
+  ///
+  /// Note that DNS cache requires an addition of a key value store named
+  /// 'reserved.platform_store'.
```
jpsim, line 178: Seems odd to have the user provide a "reserved" KV store.

jpsim (approved):
KV store naming question

mobile/library/swift/EngineBuilder.swift
Augustyniak: agree but that requirement was not introduced in this PR. IMO the API for adding key value stores is too flexible - which I think is something that we already discussed in the past. Will need to close it down a bit in the future.

mobile/library/common/config/config.cc
```diff
@@ -97,6 +97,18 @@ const char* socket_tag_config_insert = R"(
...
+const char* persistent_dns_cache_config_insert = R"(
+- &persistent_dns_cache_config
+  config:
+    name: "envoy.key_value.platform"
+    typed_config:
+      "@type": type.googleapis.com/envoymobile.extensions.key_value.platform.PlatformKeyValueStoreConfig
+      key: dns_persistent_cache
+      save_interval:
+        seconds: 15
+      max_entries: 100
```
Augustyniak, line 109: added this for safery

jpsim: good call

mobile/library/common/config/config.cc
```diff
@@ -97,6 +97,18 @@ const char* socket_tag_config_insert = R"(
...
 
+const char* persistent_dns_cache_config_insert = R"(
+- &persistent_dns_cache_config
+  config:
+    name: "envoy.key_value.platform"
+    typed_config:
+      "@type": type.googleapis.com/envoymobile.extensions.key_value.platform.PlatformKeyValueStoreConfig
+      key: dns_persistent_cache
+      save_interval:
+        seconds: 15
```
jpsim, line 108: How does this save interval work? I don't think we get so many DNS resolutions for it to significantly impact performance to persist on every resolution without batching every 15 seconds.

jpsim (approved):

jpsim (approved):

Augustyniak:
going to follow up with c++ changes soon

PR #24626, opened 2022-12-19
WIP Reduce Route memory utilization by avoiding `absl::optional` overhead.

yurykats:
Reduce Route memory utilization by avoiding `absl::optional` overhead.

`std::duration` values wrapped in an `absl::optional` take twice as much memory as the durations themseves -- 16 bytes for each optional milliseconds value. This PR removes the optional overhead by using a sentinel -1 value to represent lack of the duration.

There are 6 optional durations on each Route, and 2 more in RetryPolicy, so this is saving 64 bytes per route.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/router/config_impl.cc
```diff
@@ -486,15 +487,15 @@ RouteEntryImplBase::RouteEntryImplBase(const VirtualHostImpl& vhost,
       cluster_not_found_response_code_(ConfigUtility::parseClusterNotFoundResponseCode(
           route.route().cluster_not_found_response_code())),
       timeout_(PROTOBUF_GET_MS_OR_DEFAULT(route.route(), timeout, DEFAULT_ROUTE_TIMEOUT_MS)),
-      idle_timeout_(PROTOBUF_GET_OPTIONAL_MS(route.route(), idle_timeout)),
+      idle_timeout_(PROTOBUF_GET_MS_OR_DEFAULT(route.route(), idle_timeout, -1)),
```
jmarantz, line 490: can you use a constexpr rather than -1 here?

source/common/router/config_impl.h
```diff
@@ -404,8 +408,10 @@ class RetryPolicyImpl : public RetryPolicy {
   std::vector<uint32_t> retriable_status_codes_;
   std::vector<Http::HeaderMatcherSharedPtr> retriable_headers_;
   std::vector<Http::HeaderMatcherSharedPtr> retriable_request_headers_;
-  absl::optional<std::chrono::milliseconds> base_interval_;
-  absl::optional<std::chrono::milliseconds> max_interval_;
+  // Durations below are optional in the config. -1 sentinel is used to represent a value that
+  // was not provided.
+  std::chrono::milliseconds base_interval_{-1};
```
jmarantz, line 413: suggest
```
static constexpr std::chrono::milliseconds TimeoutUnset = -1;
```
or if std:;chrono::miliseconds cannot be constexpr-constructed then you could use an inline function to supply the Unset value to use as a default.

source/common/router/config_impl.h
```diff
@@ -656,22 +662,26 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
...
+    return grpc_timeout_header_offset_.count() >= 0 ? absl::optional(grpc_timeout_header_offset_)
+                                                    : absl::nullopt;
   }
   absl::optional<std::chrono::milliseconds> maxGrpcTimeout() const override {
-    return max_grpc_timeout_;
+    return max_grpc_timeout_.count() >= 0 ? absl::optional(max_grpc_timeout_) : absl::nullopt;
   }
   absl::optional<std::chrono::milliseconds> grpcTimeoutOffset() const override {
-    return grpc_timeout_offset_;
+    return grpc_timeout_offset_.count() >= 0 ? absl::optional(grpc_timeout_offset_) : absl::nullopt;
```
jmarantz, line 684: how hard would it be to change call-sites to compare against the Unset constant rather than having to convert to a std::optional and then have another conditional at the call-site?

yurykats: `optional` in the return value of the API makes sense to me -- it clear communicates to the caller that this may or may not have a value. If we just return a duration with a magic sentinel, we lose that clarity. We can convert the current callsites, but any future developer would not expect to have to compare the result to `Unset` constant.

jmarantz: fair enough -- that is a clear interface and I'm fine with it.

I was prematurely optimizing that the expanded code would have two conditionals rather than one. What you have is fine.

source/common/router/config_impl.cc

source/common/router/config_impl.h

pradeepcrao: I didn't check, but it would be good to ensure that we have PGV rules that prevent the user from ever setting a sentinal value in the config.

yurykats: I didn't see PGV rules around `Durations`, I assume it's implied that a negative duration is unreasonable.

source/common/router/config_impl.h
```diff
@@ -344,6 +344,8 @@ class VirtualHostImpl : public VirtualHost, Logger::Loggable<Logger::Id::router>
 };
 
 using VirtualHostSharedPtr = std::shared_ptr<VirtualHostImpl>;
+// Sentinel value to represent an unset std::duration.
+static constexpr long UnsetDuration = -1;
```
jmarantz, line 348: looking at the definition in one of our header files, and also to align with Envoy practices, I think this should be int64_t rather than long.

source/common/router/config_impl.h
```diff
@@ -1079,12 +1105,14 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
...
   const Http::Code cluster_not_found_response_code_;
   const std::chrono::milliseconds timeout_;
-  const absl::optional<std::chrono::milliseconds> idle_timeout_;
-  const absl::optional<std::chrono::milliseconds> max_stream_duration_;
-  const absl::optional<std::chrono::milliseconds> grpc_timeout_header_max_;
-  const absl::optional<std::chrono::milliseconds> grpc_timeout_header_offset_;
-  const absl::optional<std::chrono::milliseconds> max_grpc_timeout_;
-  const absl::optional<std::chrono::milliseconds> grpc_timeout_offset_;
+  // Durations below are optional in the config. `UnsetDuration` sentinel is used to represent a
+  // value that was not provided.
```
jmarantz, line 1109: add `This takes less memory than using std::optional.`

jmarantz (commented):
lgtm with minor nits.

source/common/router/config_impl.h

source/common/router/config_impl.h

jmarantz (dismissed):
Thanks!

yurykats:
heh, it didn't like the comment! `ERROR: ./source/common/router/config_impl.h:347: Don't use std::optional; use absl::optional instead`

jmarantz (approved):

ggreenway (commented):
I think this makes the code a bit less readable and more error-prone, because it would be easy to forget a check against Unset and get an unpredictable result.

Could you instead make a utility `OptionalTime` that has only one member (the chrono::milliseconds), but has the same interface (roughly) as absl::optional (`has_value()`, `operator*`, etc), and uses `-1` internally to indicate no value? I think that would give the best of both worlds. @jmarantz wdyt?

/wait-any

jmarantz:
Great idea Greg. I think (Yury is going to kill me for scope creep) this might be a very good pattern to generalize like OptInt<IntegerType, SentinelValue>. I almost might expect abs::optional or std::optional to provide this as a specialization except the API doesn't provide an (obvious) way to specify what the sentinel is.

yurykats:
This has been done before, there are libraries out there that do just that, eg https://github.com/Sedeniono/tiny-optional. What's the Envoy's policy of using something like that?

jmarantz:
Oh that's nice. Can we take the integer version of that and not bring in the part that relies on undefined behavior? That might be a problem for flows where use UBSAN. It might be annoying to disallow the case without a sentinel.

There is a process for adding new deps but I don't recall the details. @yanavlasov any thoughts?

yanavlasov:
I would not go the route of platform specific tricks. We have to consider ARM platforms and UBSAN builds. This may be more trouble than it is worth. I would just make dedicated optional_duration which encapsulates the sentinel value and use that.

yurykats:
I agree, that a more generic sentinel-based optional, a-la tiny_optional, would be helpful in general. I may get back to that, but for these optional timeouts I went a slightly different route -- https://github.com/envoyproxy/envoy/pull/24659 -- which allows to avoid even more overhead if none of them are present.

PR #24627, opened 2022-12-19
Http input matcher on query parameters

aguinet:
Signed-off-by: Adrien Guinet <adrien@reblaze.com>

Commit Message:

Add an HTTP input matcher that returns a named query parameter.

Additional Description: N/A
Risk Level:  Medium
Testing: Unit test that verifies the documented behavior is correct
Docs Changes: API documentation added
Release Notes: N/A
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/http/matching/inputs.cc
```diff
@@ -2,13 +2,47 @@
 
 #include "envoy/registry/registry.h"
 
+#include "source/common/http/utility.h"
+
 namespace Envoy {
 namespace Http {
 namespace Matching {
+
+Matcher::DataInputGetResult
```
tyxia, line 11: nit: The implementation here is thicker so that it is suitable to be put in .cc file even though get() function for `class HttpHeadersDataInputBase` sits in .h file.

I don't have strong preference between .cc and .h file. But can we be consistent on this ---- Either move `HttpHeadersDataInputBase`'s get to .cc OR move it to .h.
WDYT?

test/common/http/matching/inputs_test.cc
```diff
@@ -107,6 +107,29 @@ TEST(MatchingData, HttpResponseTrailersDataInput) {
...
 }
 
+TEST(MatchingData, HttpRequestQueryParamsDataInput) {
+  Network::ConnectionInfoSetterImpl connection_info_provider(
+      std::make_shared<Network::Address::Ipv4Instance>(80),
+      std::make_shared<Network::Address::Ipv4Instance>(80));
+  HttpMatchingDataImpl data(connection_info_provider);
+
+  {
+    HttpRequestQueryParamsDataInput input("username", false /* url_decoded */);
```
tyxia, line 117: format nit: the correct format for annotation should be `/* url_decoded=*/ false`.

Same for the test below

source/common/http/matching/inputs.h
```diff
@@ -139,6 +139,42 @@ class HttpResponseTrailersDataInputFactory
...
+    return [query_param = typed_config.query_param(), url_decoded = typed_config.url_decoded()] {
+      return std::make_unique<HttpRequestQueryParamsDataInput>(query_param, url_decoded);
+    };
+  };
+  ProtobufTypes::MessagePtr createEmptyConfigProto() override {
+    return std::make_unique<envoy::type::matcher::v3::HttpRequestQueryParamMatchInput>();
+  }
+
+private:
+  const std::string name_;
```
tyxia, line 175: This is unused since it is hard coded to `query_params` in name() function

Either remove this or set it to `query_params` and use it in name() function.

source/common/http/matching/inputs.cc
```diff
@@ -2,13 +2,47 @@
...
+
+  Utility::QueryParams params;
+  if (url_decoded_) {
+    params = Http::Utility::parseAndDecodeQueryString(ret->value().getStringView());
+  } else {
+    params = Http::Utility::parseQueryString(ret->value().getStringView());
+  }
+  auto ItParam = params.find(query_param_);
+  if (ItParam == params.end()) {
+    return {Matcher::DataInputGetResult::DataAvailability::NotAvailable, absl::nullopt};
```
tyxia, line 32: I think the return code here doesn't match the expected behavior.
As you can see guide [here](https://github.com/envoyproxy/envoy/blob/abdbc449c61d43b6d0f9624d77bf57ffdb37d465/envoy/matcher/matcher.h#L201),  we should return {AllDataAvailable, absl::nullopt}

test/common/http/matching/inputs_test.cc
```diff
@@ -107,6 +107,29 @@ TEST(MatchingData, HttpResponseTrailersDataInput) {
...
+
+  {
+    HttpRequestQueryParamsDataInput input("user name", true /* url_decoded */);
+    TestRequestHeaderMapImpl request_headers({{":path", "/test?user\%20name=foo\%20bar"}});
+    data.onRequestHeaders(request_headers);
+
+    EXPECT_EQ(input.get(data).data_, "foo bar");
+  }
+}
+
```
tyxia, line 132: Could add some negative test cases? For example the return status is `return {AllDataAvailable, absl::nullopt}`.

Once you fixes the CI/Docs issue, the full CD/CI can run and it may ask for test coverage for the negative coverage

api/envoy/type/matcher/v3/http_inputs.proto
```diff
@@ -60,3 +60,15 @@ message HttpResponseTrailerMatchInput {
...
+// Match input indicates that matching should be done on a specific query parameter.
+// The resulting input string will be the first query parameter for the value
+// 'query_param'.
+// [#extension: envoy.matching.inputs.query_params]
+message HttpRequestQueryParamMatchInput {
+  // The query parameter to match on.
+  string query_param = 1 [(validate.rules).string = {min_len: 1}];
+
+  // Whether the query parameters and their value should be URL-decoded.
+  bool url_decoded = 2;
```
tyxia, line 73: In production, it seems is always safe and legal to use `parseAndDecodeQueryString` for URL (but not for `parseQueryString` doesn't handle percent-encoding)

So, do we need `bool url_decoded` here in the proto? Can we remove this flag and just always use `parseAndDecodeQueryString`
WDYT?

tyxia (changes requested):
Thanks for working on this!

You will need to add `envoy.matching.inputs.query_params` to extensions_metadata.yaml file to fix the CI/docs failure. You can search for the examples like `envoy.matching.inputs.request_headers`

(Somehow, my review comments were published separately in two sections even though I did click `review` in one shot... Sorry for this)

api/envoy/type/matcher/v3/http_inputs.proto
aguinet: I think @tzuryby has more inputs here: what is used from a WAF point of view, decoded or raw query strings?

tzuryby: Parse and decode. 
This is the value the waf will match.

source/common/http/matching/inputs.cc
aguinet: I just saw that `status_code_input` has its separate CC/header and proto files. Maybe I should do the same here?

tyxia: Right in this exact file, [class HttpHeadersDataInputBase](https://github.com/envoyproxy/envoy/blob/34d1e7ddaaffe31b907f13502d96c7a36dacd67b/source/common/http/matching/inputs.h#L25) has get()'s both impl and declare in the header file.

I think `HttpHeadersDataInputBase` is more related here and we should keep consist this input.{cc.h} at least.

api/envoy/type/matcher/v3/http_inputs.proto
aguinet: Fixed here https://github.com/envoyproxy/envoy/pull/24627/commits/5305a220ce918ae6fbac237d031bcb97fed2ff13#diff-8ab5a462143e5ab5a9fa350b07d8a8c86b98a63a648066cba38cca1d1fe76638L71

source/common/http/matching/inputs.cc
aguinet: Thanks for the link to the guide, this makes the semantic much clearer! Fixed here: https://github.com/envoyproxy/envoy/pull/24627/commits/5305a220ce918ae6fbac237d031bcb97fed2ff13#diff-f73e74c7f9395ff0a56e44f6fc5729a30576471e557bca2de38647b8af581e02R29

source/common/http/matching/inputs.h
aguinet: Indeed, sorry about this... Fixed here: https://github.com/envoyproxy/envoy/pull/24627/commits/5305a220ce918ae6fbac237d031bcb97fed2ff13#diff-2df46d0706764a215b8c3bf412d40761a2f7c02f8abb798aa929cad394e721b7L175

test/common/http/matching/inputs_test.cc
aguinet: Added here: https://github.com/envoyproxy/envoy/pull/24627/commits/5305a220ce918ae6fbac237d031bcb97fed2ff13#diff-53f65a7ea6acbe833684b356ca5108be6eda4de221cd37269ac94f823f9f9662R132

aguinet:
Thanks @tyxia for the review! I also added the extension to `extensions_metadata.yaml`!

source/common/http/matching/inputs.cc
aguinet: So in the end I should put the code of the `get` function in `inputs.h`?

tyxia: Yea, that works for me.

tyxia (changes requested):
Thanks for addressing the comments!

I noticed that there is still presubmit failure. Here is the [style guide](https://github.com/envoyproxy/envoy/blob/main/api/STYLE.md#adding-an-extension-configuration-to-the-api) Could you check that guide? 

You can also search `envoy.matching.inputs.request_trailers` throughout the codebase.  I think You will also need to add it to `matching_api.rst` doc file and `extensions_schema.yaml` Also, you need to add it to release note `changelogs/current.yaml` for introducing this new feature.

aguinet:
@tyxia I added a couple of tests to make sure code coverage is fine. In order to test the factory part, I added this file https://github.com/envoyproxy/envoy/blob/a71029666457489c02cd088e099d1a3e5a15bb7b/test/common/http/matching/config_test.cc which basically calls `createDataInputFactoryCb`  and verifies the output is okay.

I didn't find any code that already did this atomically for the existing matching inputs. My understanding is that the factory is tested indirectly in other tests, as this one https://github.com/envoyproxy/envoy/blob/317efa53c6add02fa9533fc8f93f10f590077dca/test/common/http/match_delegate/config_test.cc#L180-L209 .

My question is: should I put this "factory creation test" somewhere else?

Thanks, and happy new year :)

tyxia:
@aguinet  Happy New Year!
(Please expect my slow response as I am OOO early Jan)

This factory part is used by http filter. As you are adding a new HTTP matching method, IMO it is probably better to add the test coverage in `match_delegate/config_test.cc`, where match delegate filter handles match tree evaluating for underly filter. This way is a bit more work than what you are having now though.

With that said, if you create matcher tree with proper InputType, like [example](https://github.com/envoyproxy/envoy/blob/24723a9c2725b81297fe00a77ccc50d3e9c71dd1/test/common/http/match_delegate/config_test.cc#L253), it should cover `createDataInputFactoryCb `. Then you can use the matcher tree in decode path.

lizan (commented):
/lgtm api

KBaichoo:
/wait 

for CI fix, think this is waiting on changes from @aguinet

aguinet:
> @aguinet Happy New Year! (Please expect my slow response as I am OOO early Jan)
> 
> This factory part is used by http filter. As you are adding a new HTTP matching method, IMO it is probably better to add the test coverage in `match_delegate/config_test.cc`, where match delegate filter handles match tree evaluating for underlying filter. This way is a bit more work than what you are having now though.
> 
> With that said, if you create matcher tree with proper InputType, like [example](https://github.com/envoyproxy/envoy/blob/24723a9c2725b81297fe00a77ccc50d3e9c71dd1/test/common/http/match_delegate/config_test.cc#L253), it should cover `createDataInputFactoryCb `. Then you can use the matcher tree later in your test decode path.

Hello @tyxia ,

Thanks for your help and sorry for my late reply. I've pushed https://github.com/envoyproxy/envoy/pull/24627/commits/a321cb42a8e7a606a15c4cfcc0f8d9d8920aaf87 that should implements what you suggests!

tyxia:
@aguinet  Sorry for late response.(It has been some other priorities going on in past two weeks on my end)

There is still coverage issue here. I think I pointed to right direction previously (i.e., add test to match_delegate/config_test.cc ) but didn't point to the example clearly/precisely.

Could you try to add something similar to [WithMacther](https://github.com/envoyproxy/envoy/blob/24723a9c2725b81297fe00a77ccc50d3e9c71dd1/test/common/http/match_delegate/config_test.cc#L48)? Basically, you will provide the static/yaml config but with different typed_config : use "@type": type.googleapis.com/envoy.type.matcher.v3.HttpRequestQueryParamMatchInput rather than HttpRequestHeaderMatchInput.  Then call createFilterFactoryFromProto.
I would expect callstack like `createFilterFactoryFromProto` -> `createTreeMatcher`-> `createDataInput` -> `createDataInputBase` -> and eventually to factory->createDataInputFactoryCb. It should hit the `createDataInputFactoryCb` you added because we are specifying `HttpRequestQueryParamMatchInput`

Can you give it a try? Thanks!

aguinet:
@tyxia thanks for your help! I've implemented it in https://github.com/curiefense/envoy/commit/ca04376cce1aee601907ff369d3206434abf6fc3 . It could be nice to verify that the parsed `query_param`  is the expected one, but I'm not really sure I can do that from the generic created object interface.

test/common/http/match_delegate/config_test.cc
```diff
@@ -149,6 +152,40 @@ TEST(MatchWrapper, DEPRECATED_FEATURE_TEST(WithDeprecatedMatcher)) {
...
+      typed_config:
+        "@type": type.googleapis.com/envoy.type.matcher.v3.HttpRequestQueryParamMatchInput
+        query_param: arg
+    exact_match_map:
+        map:
+            match:
+                action:
+                    name: skip
+                    typed_config:
+                        "@type": type.googleapis.com/envoy.extensions.filters.common.matcher.action.v3.SkipFilter
```
tyxia, line 181: nit: indentation here seems off. Please fix it in this config string

tyxia (dismissed):
Thanks for your work and patience throughout this PR!  All envoy-presubmit checks are green now. 

LGTM, modulo format nit

lizan (commented):
LGTM, can you fix the indentation?

aguinet:
> @tyxia I added a couple of tests to make sure code coverage is fine. In order to test the factory part, I added this file https://github.com/envoyproxy/envoy/blob/a71029666457489c02cd088e099d1a3e5a15bb7b/test/common/http/matching/config_test.cc which basically calls `createDataInputFactoryCb` and verifies the output is okay.
> 
> I didn't find any code that already did this atomically for the existing matching inputs. My understanding is that the factory is tested indirectly in other tests, as this one
> 
> https://github.com/envoyproxy/envoy/blob/317efa53c6add02fa9533fc8f93f10f590077dca/test/common/http/match_delegate/config_test.cc#L180-L209
> .
> 
> My question is: should I put this "factory creation test" somewhere else?
> 
> Thanks, and happy new year :)

aguinet:
@lizan just pushed a commit to fix the indentation! Sorry for the late answer

lizan (approved):

PR #24637, opened 2022-12-20
Clean up byte-swapping operations in grpc/codec.cc

ravenblackx:
Commit Message: Clean up byte-swapping operations in grpc/codec.cc
Additional Description: The review of #23847 suggested we don't like raw byte-shifting operations. This change removes the raw byte-shifting operations from codec.cc (the only non-test code that still does this sort of thing), and also removes some unnecessary double-indexing from the state machine.
Risk Level: Very low, should be effectively a no-op change, just cleanup.
Testing: Continues to pass existing tests.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/grpc/codec.cc
+++ b/source/common/grpc/codec.cc
@@ -16,10 +16,7 @@ Encoder::Encoder() = default;
 
 void Encoder::newFrame(uint8_t flags, uint64_t length, std::array<uint8_t, 5>& output) {
   output[0] = flags;
-  output[1] = static_cast<uint8_t>(length >> 24);
-  output[2] = static_cast<uint8_t>(length >> 16);
-  output[3] = static_cast<uint8_t>(length >> 8);
-  output[4] = static_cast<uint8_t>(length);
+  absl::big_endian::Store32(&output[1], length);
 }
 
 void Encoder::prependFrameHeader(uint8_t flags, Buffer::Instance& buffer) {
@@ -74,7 +71,8 @@ uint64_t FrameInspector::inspect(const Buffer::Instance& data) {
   uint64_t delta = 0;
   for (const Buffer::RawSlice& slice : data.getRawSlices()) {
     uint8_t* mem = reinterpret_cast<uint8_t*>(slice.mem_);
-    for (uint64_t j = 0; j < slice.len_;) {
+    uint8_t* end = mem + slice.len_;
+    while (mem < end) {
       uint8_t c = *mem;
       switch (state_) {
       case State::FhFlag:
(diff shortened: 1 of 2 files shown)


jmarantz:
I feel like the context in the other review was quite different, and that the "before" state of the code in this PR looks more readable to me, and also faster. I don't think we should be having a ton of byte-shifting everyone in Envoy but the codecs seem like a place they might need to be.

ravenblackx:
> I feel like the context in the other review was quite different, and that the "before" state of the code in this PR looks more readable to me, and also faster. I don't think we should be having a ton of byte-shifting everyone in Envoy but the codecs seem like a place they might need to be.

I would be very surprised if the before state was faster - I think the 'write' context will be identical (boiling down to a single assembler instruction when optimized, and it's one easily-readable line now instead of four error-prone lines with casts and bitshifts), and the 'read' context I would actually expect to be faster as well, because previously it was four operations of shifting a byte in a 32 bit space then or-ing it onto another 32 bit space (so at least 8 instructions), and now it's four byte-copy operations plus (most likely) a single assembler instruction for the endian-swap. Plus previously it was doing two index updating operations for each byte read, and now it is doing one. The optimizer can't do anything to optimize the byte-shifts away when they're occurring in a state machine loop like they were, but it can (and does) optimize `absl::big_endian` functions into a single `bswap` instruction.

Any speed difference is likely approximately irrelevant anyway, since it's only on the 'length' part of the packet which is typically a tiny fraction of the data. If we really wanted to optimize for speed here, the right thing to do would be to check if there are four bytes available in the buffer (which there will be in almost every real-world case) and read them in a single operation if there are, and only do the byte-at-a-time state machine loop in the degenerate case.

ravenblackx:
> I would be very surprised if the before state was faster

I fed an approximation of the two versions to [Godbolt](https://godbolt.org/z/Pcdrjbrx5) which confirms my expectations - the two 'write' versions compile to identical assembler instructions, and the version of the 'read' that uses absl compiles the byte-branches to 3+3+3+6 instructions, while the shift-using version compiles to 4+5+5+8 instructions. (Though the 8-bit movs [might be](https://www.agner.org/optimize/instruction_tables.pdf) slightly slower than the shifts and ors.)

jmarantz:
You've convinced me this is fast enough :). But maybe it'd be better to find another reviewer that's familiar with the codecs?

ravenblackx:
Incidentally, the *main* motivation for this change was not speed or readability (though I think the change improves both), but rather a twofer of consistency (if my serialization should not use manual byte-shifts then it makes sense that neither should others) and discoverability (before I wrote those other byte-shifts I searched to see if there was an envoy style alternative and found codec.cc doing byte-shifts, from which I inferred that it was acceptable to do this). Since the review suggested that is not the case, it makes sense to back-propagate that ruling to my "canonical source" so the next time someone is in the same situation they aren't misled.

As for finding a reviewer more familiar with this code - `CODEOWNERS` doesn't suggest anyone in particular (just defaults to all maintainers), and the change history doesn't show any strong patterns either. Maybe alyssawilk or htuch? I don't think I have permissions to reassign.

jmarantz:
I'll abstain on this one; I mildly prefer the previous state but I have no actual objections. I haven't been working much in the grpc code but I see that Kuat has so I'll kick this over.

kyessenov (approved):
I think this is an improvement: I'd expect absl to make a better use of the hardware micro-optimizations for conversions. The only tricky part is the sharing of the storage, but there's a comment there already.

PR #24638, opened 2022-12-20
test: Add clarifying comments to BaseClientIntegrationTest

abeyad:
We clarify that BaseClientIntegrationTest::defaultConfig *only* gets used if `override_builder_config_` is set to true.
The code is very confusing to read without knowing this.


RyanTheOptimist (approved):
Oh, shoot! I totally missed this PR.

THANKS FOR DOING THIS!!!

PR #24642, opened 2022-12-20
[file_system_http_cache] Add eviction

ravenblackx:
Commit Message: [file_system_http_cache] Add eviction
Additional Description: Add the last piece of making file_system_http_cache functional; eviction. This adds a thread for the eviction operation to be performed on, allowing for on-demand eviction rather than periodic. This has advantages and disadvantages
* Pro: when there is no demand for eviction because the cache is below its size limits, the cost is very low.
* Pro: the cache should not get significantly oversized, because a demand for eviction will occur as soon as it is oversized.
* Con: if the cache is constantly being populated and entries rotated out, the cost will be higher than periodic eviction would be.
* Option for improvement: a configuration option for postponing eviction if the last eviction was more recent than (specified duration) would reduce the cost of that con, but reduce the benefit of the second pro.

Risk Level: None, this filter is not in use and is still tagged WIP.
Testing: Almost 100% coverage (updating an atomic more than once because there was a race is not covered).
Docs Changes: Adds `DESIGN.md` to explain the various eviction strategy decisions and other implementation details.
Release Notes: n/a (not yet)


ravenblackx:
/retest

(Doesn't look like I broke that.)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,75 @@
...
+}
+
+bool CacheEvictionThread::waitForSignal() {
+  absl::MutexLock lock(&mu_);
+  // Worth noting here that if `signalled_` is already true, the lock is not released
+  // until idle_ is false again, so waitForIdle will not return until `signalled_`
+  // stays false for the duration of an eviction cycle.
+  idle_ = true;
+  auto cond = [this]() ABSL_EXCLUSIVE_LOCKS_REQUIRED(mu_) { return signalled_; };
+  mu_.Await(absl::Condition(&cond));
```
jmarantz, line 50: nit: you can just use `mu_.Await(absl::Condition)(&signalled_))` without manually making the lambda.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,75 @@
...
+  ASSERT(inserted);
+}
+
+void CacheEvictionThread::removeCache(FileSystemHttpCache& cache) {
+  absl::MutexLock lock(&cache_mu_);
+  bool removed = caches_.erase(&cache);
+  ASSERT(removed);
+  if (caches_.empty()) {
+    terminate();
+    thread_->join();
```
jmarantz, line 28: I feel like it would be more natural to leave the cache thread running in case a cache is added after the last one is removed, and do the join() in the destructor for CacheEvictionThread, where I'd also assert caches_.empty(). WDYT?

The way you have it, if you own a CacheEvictionThread, the following sequence will fail:
```
  CacheEvictionThread cache_eviction_thread(thread_factory);
  FileSystemHttpCache cache;
  cache_eviction_thread.addCache(cache);
  cache_eviction_thread.removeCache(cache);
  cache_eviction_thread.addCache(cache);
  cache_eviction_thread.removeCache(cache);
```

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+CacheEvictionThread::CacheEvictionThread(Thread::ThreadFactory& thread_factory)
+    : os_sys_calls_(Api::OsSysCallsSingleton::get()),
```
jmarantz, line 13: as os_sys_calls_ is not used by this class directly, could we avoid holding onto it, and just pass that directly to the constructor of the FileSystemHttpCache?

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -354,14 +366,15 @@ void FileSystemHttpCache::init() {
...
+  cache_eviction_thread_.addCache(*this);
 }
 
 void FileSystemHttpCache::trackFileAdded(uint64_t file_size) {
   size_count_++;
   size_bytes_ += file_size;
   stats_.size_count_.inc();
   stats_.size_bytes_.add(file_size);
-  // TODO(ravenblack): potentially trigger cache cleanup operation.
+  cache_eviction_thread_.signal();
```
jmarantz, line 377: I'd suggest keeping an approximate track of the size of the cache overall and only signaling the eviction thread when either:
 * the approximate size (ignoring other processes sharing the cache) exceeds a threshold
 * a timer we can tune goes off; which would do eventual cleanup if there were two overlapping Envoy processes sharing the cache config, e.g. for hot-restart or because someone wants to run more than one distinct Envoy server from the same machine.

My concern is that every new file that gets added will trigger a full sweep of the disk, which will impact system performance. I think it'd be good to have in mind an eviction strategy where we do evictions when the file system reaches X bytes, and bring it down to 0.75*X bytes, or similar, so that we aren't thrashing the disk in the extra thread.

I think it's OK to have an extra thread doing the GC but I'd be concerned if that was constantly busy.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +401,77 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+  uint64_t proposed_size_evicted = 0;
+  struct ProposedEviction {
+    std::string name_;
+    uint64_t size_;
+    struct timespec last_touch_;
+    bool operator<(const ProposedEviction& other) const { return last_touch_ < other.last_touch_; }
+  };
+  std::multiset<ProposedEviction> proposed_evictions;
+
+  for (const Filesystem::DirectoryEntry& entry : Filesystem::Directory(std::string{cachePath()})) {
```
jmarantz, line 438: I thought the plan (based on the doc) was to generate a tree structure for the files in the cache. should we be doing a recursive scan through the directories to hit all the files?

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
```diff
@@ -21,13 +21,17 @@ namespace FileSystemHttpCache {
 using ConfigProto =
     envoy::extensions::http::cache::file_system_http_cache::v3::FileSystemHttpCacheConfig;
 
+class CacheEvictionThread;
+
```
jmarantz, line 25: can we have a reference to the doc you wrote in the FileSystemCache class-comment here? Also is that doc up-to-date? If not, can we make it so?

jmarantz (commented):
This looks like a great start!

I haven't looked in detail at the cleaning algo yet but just had a few comments as a first pass to get started.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -71,11 +81,16 @@ absl::string_view FileSystemHttpCache::name() {
...
 
 FileSystemHttpCache::FileSystemHttpCache(
-    Singleton::InstanceSharedPtr owner, ConfigProto config,
-    std::shared_ptr<Common::AsyncFiles::AsyncFileManager>&& async_file_manager,
+    Singleton::InstanceSharedPtr owner, CacheEvictionThread& cache_eviction_thread,
+    ConfigProto config, std::shared_ptr<Common::AsyncFiles::AsyncFileManager>&& async_file_manager,
     Stats::Scope& stats_scope)
     : owner_(owner), config_(config), async_file_manager_(async_file_manager),
-      stats_(generateStats(stats_scope, cachePath())) {}
+      stats_(generateStats(stats_scope, cachePath())),
```
jmarantz, line 88: can we keep some eviction stats? e.g.:
 * how much time is busy vs idle in the eviction thread
 * number of files evicted
 * number of bytes evicted

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
jmarantz: Or if you want to manage this with a singleton (which I generally prefer to avoid but that was already decided I guess) then I think I wouldn't start the thread in the constructor; I'd start the thread when the number of caches transitions from 0->1, and join it when the number of caches transitions from 1->0, and note this can happen multiple times during the lifetime of the system.

It might also be fine to just leave the thread running until shutdown.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: Aha, or not even pass it at all and just get it from the singleton since we're not using dependency injection of it in any tests.

And I definitely hope to remove os_sys_calls in favor of a file-specific abstraction anyway, once I'm done buffing that up. :)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: The reason I terminate the thread on removeCache rather than destructor is that the thread from which the destructor is called differs, which is apparently not an okay place to do thread-joining. Good point about symmetry though, so I've put thread create/terminate+join in addCache/removeCache.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: Nice, thanks.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Added a comment here to clarify that signaling does not start an eviction, only a *maybe* eviction - the eviction thread only does any disk-work if the current cache size exceeds a threshold.

I do have in mind softening options because as you say even with a threshold check, it still might be constantly sweeping if the cache keeps repopulating. Either high/low thresholds (erase to low when it hits high, like watermarks) or a threshold and a time limit (erase only if threshold exceeded *and* it's been n seconds since last erase), but this impl is MVP.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Yes, MVP pass is just doing flat structure though.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
ravenblackx: I'll make a `README.md` in this folder for it (or maybe a different-named markdown file so github doesn't shove a big doc in your face when you don't want it?)

jmarantz: I think a link to the doc is OK; would that be easy?

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: I have included this in DESIGN.md as an unchecked goal and postponed it from this change. We like small PRs, right? :)

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
ravenblackx: Added DESIGN.md, and I simplified the contents vs. the doc to exclude some of the speculative things that at this point feel unlikely, and exclude goals that require only changes to the cache filter not the cache itself.
([the doc](https://docs.google.com/document/d/1DWy1TE8RfV_0_fWs28DARS-GblUNuSRBqFxagOi6QPQ/edit#heading=h.7sus632d57si) in case you want to compare.)

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
jmarantz: Drop a TODO here then?

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +405,78 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+  uint64_t proposed_size_evicted = 0;
+  struct ProposedEviction {
+    std::string name_;
+    uint64_t size_;
+    struct timespec last_touch_;
+    bool operator<(const ProposedEviction& other) const { return last_touch_ < other.last_touch_; }
+  };
+  std::multiset<ProposedEviction> proposed_evictions;
+
+  for (const Filesystem::DirectoryEntry& entry : Filesystem::Directory(std::string{cachePath()})) {
```
jmarantz, line 443: it seems like an interface bug that Filesystem::Directory doesn't allow a string_view. Can we make its interface be a string_view and avoid having to stringify things at every call-site?

This would be a separate PR; I'm not asking for changes on this one for this.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +405,78 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+    if (!isCacheFile(entry)) {
+      continue;
+    }
+    count++;
+    size += entry.size_bytes_.value_or(0);
+    struct stat s;
+    if (os_sys_calls.stat(absl::StrCat(cachePath(), entry.name_).c_str(), &s).return_value_ != -1) {
```
jmarantz, line 450: is cachePath() guaranteed to end in a directory separator?

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +405,78 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+  if (config().has_max_cache_entry_count()) {
+    // capture a value from the atomic because we're going to use it twice and don't want our
+    // value to change in between.
+    uint64_t seen_count = size_count_;
+    count_to_evict = seen_count > config().max_cache_entry_count().value()
+                         ? seen_count - config().max_cache_entry_count().value()
+                         : 0;
+  }
+  if (size_to_evict == 0 && count_to_evict == 0) {
+    return;
```
jmarantz, line 428: can we factor lines 409-429 into a needsCleanup() method?

Then it might be easier to run that prior to signalling, and would also be easy to test.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +405,78 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+
+  for (const Filesystem::DirectoryEntry& entry : Filesystem::Directory(std::string{cachePath()})) {
+    if (!isCacheFile(entry)) {
+      continue;
+    }
+    count++;
+    size += entry.size_bytes_.value_or(0);
+    struct stat s;
+    if (os_sys_calls.stat(absl::StrCat(cachePath(), entry.name_).c_str(), &s).return_value_ != -1) {
+      struct timespec last_touch = std::max(s.st_atim, s.st_ctim);
```
jmarantz, line 451: i've never used it but it might be worth considering the C++17 support for this: https://en.cppreference.com/w/cpp/filesystem/last_write_time

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +405,78 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+          last_touch < proposed_evictions.rbegin()->last_touch_) {
+        // We either haven't evicted enough yet, or this eviction candidate is 'older'
+        // than our current 'youngest' eviction candidate. So add this one to candidates.
+        proposed_evictions.insert(
+            ProposedEviction{entry.name_, entry.size_bytes_.value_or(0), last_touch});
+        proposed_size_evicted += entry.size_bytes_.value_or(0);
+        auto youngest = std::prev(proposed_evictions.end());
+        if (proposed_size_evicted - youngest->size_ >= size_to_evict &&
+            proposed_evictions.size() > count_to_evict) {
+          // We'd still be evicting enough if we don't evict the 'youngest' proposed eviction,
```
jmarantz, line 462: I was expecting a loop here.

Let's say that I need to evict 1M to get to my target size, and the youngest 1000 files are 1k byte each, but then we find an older file that's 1M. We can un-propose evicting all the young 1k files.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +405,78 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+          proposed_evictions.erase(youngest);
+        }
+      }
+    }
+  }
+  size_bytes_ = size;
+  size_count_ = count;
+  for (const ProposedEviction& eviction : proposed_evictions) {
+    if (os_sys_calls.unlink(absl::StrCat(cachePath(), eviction.name_).c_str()).return_value_ !=
+        -1) {
```
jmarantz, line 474: it's worth a comment here that we expect this to fail sometimes, for example if there are two Envoys cleaning the directory at the same time, or if someone manually deletes cache files.

It might be worth a TODO to invalidate our view of how much is consumed if this occurs though.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
jmarantz: sure, sg. Can you add a ref to where that will be here?

jmarantz (commented):
a few more comments; still haven't looked at tests.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: The train of thought was that we generally want envoy threads freed as quickly as possible, and signal-always is probably cheaper *in that thread* than touching two atomics. And then the eviction thread won't actually wake up every time, because (assuming things are busy) many times it will already be awake and the signal won't do anything. But you're probably right that most of the time we should be doing nothing, and touching two atomics is cheap.

On the downside I ended up with two versions of checking if we need to evict, because one place we need to know *how much* to evict, which is more work, so we probably don't want to do that in the envoy worker thread unnecessarily.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Done.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
ravenblackx: Added reference to DESIGN.md on the class description for both FileSystemHttpCache and CacheEvictionThread.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Done.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -1,18 +1,29 @@
...
 #include "source/common/filesystem/directory.h"
 #include "source/common/http/header_map_impl.h"
+#include "source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.h"
 #include "source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h"
 #include "source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.h"
 #include "source/extensions/http/cache/file_system_http_cache/insert_context.h"
 #include "source/extensions/http/cache/file_system_http_cache/lookup_context.h"
 #include "source/extensions/http/cache/file_system_http_cache/stats.h"
 
+static bool operator<(const struct timespec& a, const struct timespec& b) {
```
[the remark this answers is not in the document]
ravenblackx: Used Envoy::SystemTime (which is std::chrono). Probably a little more expensive and still needed a little helper function, but I agree it's a more practical one.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Agree that would be nicer. Or at least making it take a std::string rather than a std::string&, because it just copies it into its own string anyway, so it's having to perform *two* copies here unless the optimizer reasons it away.
(Taking a string_view would reduce it to one copy in this context, but so would taking a std::string and moving it. For convenience *and* performance, having both std::string and absl::string_view constructors would be optimal, because e.g. absl::StrCat(...) passed in as a string_view would perform an additional copy, vs. absl::StrCat(...) passed in as a std::string would require only a move.)

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Done-ish. needsEviction, because Cleanup means too many other things, and I ended up with two of them because here we need to capture the size and in the other place we don't (and doing so anyway would be awkward and confusing).

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: We *might* get "last read or written" from stat, which is preferable behavior over just last written. There doesn't seem to be a C++17 function for that.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Agree. Looped.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Commented and TODO'd, though I'm not sure what you actually meant by "how much is consumed".

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+## Discussions
+
+<a name="thundering-herd"></a>
+### Thundering herd
+
+The current implementation, if there are multiple requests for the same resource before the cache is populated, has only one of them perform an insert operation to the cache, and the rest simply bypass the cache. This can cause the "thundering herd" problem - if requests come in bursts the cache will not protect the upstream from that load.
+
+One possible solution would be to have all requesters for the same cache entry stream as the cache entry is written. However, if we do that, and the inserting stream gets closed prematurely, all the dependent streams would be forced to drop their also-incomplete responses.
+
+Another possible solution is to have secondary requesters wait until the cache entry is populated or abandoned before deciding whether to become cache readers or inserters (or bypass, if it turns out to be uncacheable). The downside of this option is that for large content, the dependent clients won't start streaming at all until the first client *finishes* streaming.
```
jmarantz, line 35: that makes sense, and wasn't something we had considered an issue as video was not a target.

Maybe the initial requester can put the lock object into one of several states:
 * in-progress, size unknown
 * headers-complete; content-length known and less than a threshold
 * headers-complete, content-length known and more than a threshold
 * headers-complete, chunked encoding

In this case we can block secondary requests until we get the headers, and then we can decide: chunked or large content-length means "secondaries don't wait".

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+
+An ideal solution would be to either make an inserting stream non-cancellable (i.e. if the client cancels, the upstream connection continues to stream to populate the cache). This could be achieved either by using the existing stream and adding a "non-cancellable" feature in the core (a bit of a large scale change), or by making insertion not use the existing stream at all, instead creating its own client. The problem with that option is that ideally the client would only pass through the filters that are upstream of the cache filter, and there is currently no mechanism for creating a new "partial filter chain" client like this.
```
jmarantz, line 37: I assume the most common consideration is compressing before you cache. This brings in a lot of corner cases. In this respect, keeping it simple is recommended even if it's not the most effiicent.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
jmarantz: I can see why it might be a bit confusing to join from a different thread than the creator, but I don't see it being invalid in the C++ spec. Is there an Envoy standard?

It *is* UB to join from more than one thread though.

Moreover, I don't see how your change keeps us away from that situation.

Would it be easier to just never join the thread until the singleton is cleaned up? It would just be in a state when the thread was blocked on an Await that never comes. Of course we'd want to lazy-create in that case to avoid using a thread just cause this is linked in.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,77 @@
...
+  idle_ = true;
+  mu_.Await(absl::Condition(&signalled_));
+  signalled_ = false;
+  idle_ = false;
+  return !terminating_;
+}
+
+void CacheEvictionThread::work() {
+  while (waitForSignal()) {
+    absl::MutexLock lock(&cache_mu_);
```
jmarantz, line 60: this lock might be held for a long time while the cache is iterated over. I'm not sure if this is a real problem.

One suggestion would be to have a locked initial loop that makes a local vector of shared_ptr<FileCache> to hold onto them. Then you can release the lock and iterate over the caches.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -1,18 +1,32 @@
...
+#include "source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.h"
 #include "source/extensions/http/cache/file_system_http_cache/cache_file_fixed_block.h"
 #include "source/extensions/http/cache/file_system_http_cache/cache_file_header_proto_util.h"
 #include "source/extensions/http/cache/file_system_http_cache/insert_context.h"
 #include "source/extensions/http/cache/file_system_http_cache/lookup_context.h"
 #include "source/extensions/http/cache/file_system_http_cache/stats.h"
 
+static constexpr Envoy::SystemTime timespecToChrono(const struct timespec& t) {
+  return Envoy::SystemTime{std::chrono::duration_cast<std::chrono::system_clock::duration>(
+      std::chrono::seconds{t.tv_sec} + std::chrono::nanoseconds{t.tv_nsec})};
```
jmarantz, line 17: Feels like this utility should go into source/common/common/utility.cc

jmarantz (commented):
flushing more comments

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
+# File system cache design
+
+## Goals
+
+[x] Cache should be usable by two processes at once (e.g. during hot restart)
```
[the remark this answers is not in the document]
ravenblackx: Fixed (was hard to tell, vscode's preview doesn't support these things!)

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+[x] Cache should evict the least recently used (LRU) entry when full
+[x] Cache should be limited to a specified amount of storage
+[ ] Cache should mitigate thundering herd problem (i.e. if two or more workers request the same cacheable uncached result at the same time, only one worker should hit upstream). See [discussion](#thundering-herd).
```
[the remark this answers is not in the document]
ravenblackx: Added a comment above the checkboxes ("not yet", essentially)

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+[ ] There should be an ability to remove objects from the cache with some kind of API call.
+[ ] Cache should expose counters for eviction stats (files evicted, bytes evicted).
+[ ] Cache should expose counters for timing information (eviction thread idle, eviction thread busy)
```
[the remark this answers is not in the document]
ravenblackx: Latency histograms was already in the list, added size histogram.

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+[ ] Cache should expose gauges for total size stored.
+[ ] Cache should optionally expose histograms for insert and lookup latencies.
+[x] Cache should index by the request route *and* a key generated from headers that may affect the outcome of a request (See [allowed_vary_headers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/cache/v3/cache.proto.html))
+[ ] Cache should allow a tree structure of folders, to avoid filesystem performance issues with overcrowded directories.
```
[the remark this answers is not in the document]
ravenblackx: Clarified and added an internal link.

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Storage design
+
+* The only state stored in memory is that a cache entry is in the process of being written; this allows other requests for the same resource to avoid creating duplicate write operations.
```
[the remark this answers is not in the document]
ravenblackx: Added more detail, yes, not part of the goal.

source/extensions/http/cache/file_system_http_cache/DESIGN.md
ravenblackx: Added paraphrase of this to the doc. Seems like a good idea in the short term, as any of the streaming-capable options seem like a better idea but involve significantly more effort up-front.

source/extensions/http/cache/file_system_http_cache/DESIGN.md
ravenblackx: Right, I think the "make a stream non-cancellable" option would be *vastly* simpler, but has the downside of being a change to the core behavior which is more dangerous, and likely requires a lot of understanding of existing things that I am not yet familiar with. And in the short term, not doing shared streaming at all is definitely the simplest.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: Added a comment about this. Releasing the lock first is no good because when you're not holding the lock, the pointers can become invalid.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Done. Included a bit of a shabby test because there's not really a good test for such a simple thing. (Made some effort to make the test not just a duplicate of the code!)

ravenblackx:
For some reason I can't reply in the thread-joining thread. You're right, I misremembered and described the problem incorrectly. At this point I can't actually remember what the problem was, but there was definitely *something*. I think it's to do with the order in which things are destroyed in tests, rather than in production, that leds to some sort of thread-related test failure. However, it seems some other restructuring has resolved that problem, so it does make sense to use constructor and destructor now.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +404,108 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
   stats_.size_bytes_.set(size_bytes_);
 }
 
+bool FileSystemHttpCache::needsEviction() const {
+  if (config().has_max_cache_size_bytes() &&
+      size_bytes_ > config().max_cache_size_bytes().value()) {
```
jmarantz, line 409: it would seem easy to have two thresholds, and the one we check here should be higher.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +404,108 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+    return true;
+  }
+  if (config().has_max_cache_entry_count() &&
+      size_count_ > config().max_cache_entry_count().value()) {
+    return true;
+  }
+  return false;
+}
+
+bool FileSystemHttpCache::needsEviction(uint64_t& size_to_evict_out,
```
jmarantz, line 419: Can we merge the two implementations for needsEviction and have it return the two eviction-goals as a pair?

We can then just check the value of those for non-zero at the call-site.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +404,108 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
...
+                             : 0;
+  } else {
+    count_to_evict_out = 0;
+  }
+  return size_to_evict_out > 0 || count_to_evict_out > 0;
+}
+
+void FileSystemHttpCache::maybeEvict() {
+  uint64_t size_to_evict, count_to_evict;
+  if (!needsEviction(size_to_evict, count_to_evict)) {
```
jmarantz, line 446: I'm wondering if we need to re-call this here. I don't know if it's possible to call maybeEvict() now unless an eviction is needed. And then you may as well do the directory walk and get the more exact numbers.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,79 @@
...
+#include "source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+CacheEvictionThread::CacheEvictionThread(Thread::ThreadFactory& thread_factory)
+    : thread_(thread_factory.createThread([this]() { work(); })) {}
```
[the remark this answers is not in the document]
ravenblackx: That's already the case - CacheSingleton is only allocated when getCache is called, and the CacheSingleton is deleted when the last cache goes out of scope, because the caches have shared_ptr ownership of the CacheSingleton.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Four thresholds, because there's already two. I'm trying not to touch the API in this PR since then an API review is also required. More thresholds can be another change (again, I thought we liked small changes! More thresholds is in DESIGN.md as a goal now.)

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Done. (A struct rather than a pair to make it nicer to work with.)

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: You'd have to do the directory walk twice in that case, once to find the size and then again to select the files to evict now that you know how much you need to evict.
Also, the pre-check is necessary because the eviction thread has a shot at each cache when woken, so if there's more than one cache and only one needs evicting, there's no need to scan the others.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: The thread can't own the caches with shared_ptr because the caches indirectly own the thread with shared_ptr. It would make a memory leak loop.

The caches are solely owned by filter config, which is necessary as that's the only way to manage removal of unused caches when the config changes.

weak_ptr transitioning to shared_ptr seems like an option, but if you do *that* then you can end up with the CacheEvictionThread being the owner of the last reference to the last cache, which means when that block goes out of scope the CacheEvictionThread destructor is called, which means it tries to call join from the thread itself, and *that* was the problem that led to the non-constructor-based initialization before, which didn't actually help because removeCache is still part of the cache destructor.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
jmarantz: I don't see how to get around it unless you trust your cached value (which seems unwise if there are two processes operating in the same directory).

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
```diff
@@ -388,6 +404,93 @@ void FileSystemHttpCache::trackFileRemoved(uint64_t file_size) {
   stats_.size_bytes_.set(size_bytes_);
 }
 
+FileSystemHttpCache::EvictionsRequired FileSystemHttpCache::needsEviction() const {
+  EvictionsRequired required;
+  if (config().has_max_cache_size_bytes()) {
+    // capture a value from the atomic because we're going to use it twice and don't want our
+    // value to change in between.
+    uint64_t seen_size = size_bytes_;
+    required.size_bytes_ = seen_size > config().max_cache_size_bytes().value()
```
jmarantz, line 413: make a temp for max_cache_size_bytes(); same for the count below.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
jmarantz: I can see how this has gotten complex :)

The high level requirement (from my perspective) is that I think we shouldn't get into a situation where we are blocking the data plane on a cache cleanup.

You might be able to use atomics to help you avoid that, but it requires some careful thinking.

jmarantz:
I think there are still other pending comments besides the ones I just flushed.

/wait

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: Done, but then undone because instead we don't do this any more.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: I think it would be fine to trust the cached values really, but okay, I've redone it so it doesn't trust the cached values, still only does one pass, and now doesn't block for long, at the small cost of putting all the cache filenames in memory briefly during the eviction pass. (Still doing a pre-check to avoid running eviction over cache instances that don't need it.)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: Done by splitting off a shared *part* of the cache instance, which allows the main cache instance to still be deleted by its owner, but the eviction thread to keep alive the part that it's using until it's done with it.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.cc
ravenblackx: (And this should be fine if two instances run at the same time since they'll agree to keep the same files, worst case it would give them a misleading estimate of their cache size afterwards so they might run another eviction pass before it's necessary.)

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,58 @@
...
+
+Another possible solution is to have secondary requesters wait until the cache entry is populated or abandoned before deciding whether to become cache readers or inserters (or bypass, if it turns out to be uncacheable). The downside of this option is that for large content, the dependent clients won't start streaming at all until the first client *finishes* streaming.
+
+An ideal solution would be to either make an inserting stream non-cancellable (i.e. if the client cancels, the upstream connection continues to stream to populate the cache). This could be achieved either by using the existing stream and adding a "non-cancellable" feature in the core (a bit of a large scale change), or by making insertion not use the existing stream at all, instead creating its own client. The problem with that option is that ideally the client would only pass through the filters that are upstream of the cache filter, and there is currently no mechanism for creating a new "partial filter chain" client like this.
+
+_Proposal from jmarantz:_
+
+The lock object could go into one of several states:
```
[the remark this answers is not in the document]
toddmgreer: First, I strongly support the current approach of initially ignoring the problem. The thundering herd issue is complicated, with lots of reasonable approaches and tradeoffs to be considered. There's plenty of value in getting a filesystem-based cache into production, and we shouldn't delay it for this.

Some observations:
1. We must first decide whether to pause a request. The proposed list of states is a good start, and I like the approach of letting the config set this policy, because request coalescing isn't always desirable.
2. We should support optionally starting a paused response before the cache fill is complete, to avoid unreasonably delaying large/streaming responses.
3. The cache fill from upstream should either be gated by the fastest client, or should not be gated (except by how fast the cache can insert). Otherwise, a slow first client will slow down all clients.
4. Upstream connection problems are tricky.
  a. When should we restart an upstream connection?
  b. When should we report errors to clients?
  c. When should we let waiting requests just proceed upstream?
  d. If we resume an upstream request, how do we confirm that it's equal and what do we do if it isn't?
  e. Should we run multiple requests in parallel in case one fails?
5. How do we handle conditional requests?
6. How do we handle range requests?
7. If we support pausing requests before the first request has response headers, how do we handle Vary?
8. All this should be managed by CacheFilter, with minimal requirements placed on HttpCache implementations. However, it might make sense to initially implement it in a particular HttpCache implementation, then move it to CacheFilter once it's settled and tested. That would make it easier to evolve, at the cost of later work to move it.

Some of these are tricky; others are fairly straightforward, but they all involve tradeoffs. Many of them must be decided before any real work can begin on the feature.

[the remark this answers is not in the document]
toddmgreer: Also, a major form of the thundering herd problem (the cache validation thundering herd) can be solved much more simply with support for asynchronously validating soon-to-expire entries, or with stale-while-revalidate. That may be enough for many deployments.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
```diff
@@ -116,12 +121,75 @@ TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, InitialStatsAreSetCorrectly) {
...
+  cache_ = std::dynamic_pointer_cast<FileSystemHttpCache>(
+      http_cache_factory_->getCache(cacheConfig(cfg), context_));
+  waitForEvictionThreadIdle();
+  env_.writeStringToFileForTest(absl::StrCat(cache_path_, "cache-c"), large_file_contents, true);
+  EXPECT_EQ(cache_->stats().size_bytes_.value(), file_contents.size() * 2);
+  EXPECT_EQ(cache_->stats().size_count_.value(), 2);
+  cache_->trackFileAdded(large_file_contents.size());
+  waitForEvictionThreadIdle();
+  EXPECT_EQ(cache_->stats().size_bytes_.value(), large_file_contents.size());
+  EXPECT_EQ(cache_->stats().size_count_.value(), 1);
```
jmarantz, line 187: I see you have a test for it but I am failing to see where size_count_ gets decremented. Can you help me out?

One theory I have is that there's a subsequent GC call which resets size_count_ on cache_eviction_thread.cc:123.

This reminds me, can you also keep counters for how many times eviction runs and test that here?

jmarantz (commented):
This looks great -- mostly just a collection of nits and minor comments, one concern about how size_count_ is maintained.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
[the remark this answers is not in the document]
ravenblackx: Added some log statements. Testing it on a live server is next in my set of things to do, but it'll be much easier to do that if this PR can be merged first. (We have an end-to-end test framework, but there's a whole dependency chain thing going on such that I'm only supposed to update to main-branch envoy.)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+  thread_->join();
+}
+
+void CacheEvictionThread::addCache(std::shared_ptr<CacheShared> cache) {
+  {
+    absl::MutexLock lock(&cache_mu_);
+    bool inserted = caches_.emplace(std::move(cache)).second;
+    ASSERT(inserted);
+  }
+  signal();
```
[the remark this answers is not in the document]
ravenblackx: Commented. (The signal is to perform the initial measuring, which may or may not trigger GC.)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+  // TODO(ravenblack): Add support for directory tree structure.
+  for (const Filesystem::DirectoryEntry& entry :
+       Filesystem::Directory(std::string{cache.cachePath()})) {
+    if (!isCacheFile(entry)) {
+      continue;
+    }
+    cache.size_count_++;
+    cache.size_bytes_ += entry.size_bytes_.value_or(0);
+  }
+  cache.stats_.size_count_.set(cache.size_count_);
```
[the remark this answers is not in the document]
ravenblackx: Done, but kept in this file, as a more meaningful mental separation of what's happening in which threads. (Added comments making that clear.)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+  cache.stats_.size_bytes_.set(cache.size_bytes_);
+  cache.needs_init_ = false;
+}
+
+void CacheEvictionThread::evict(CacheShared& cache) {
+  auto os_sys_calls = Api::OsSysCallsSingleton::get();
```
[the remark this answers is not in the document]
ravenblackx: Done.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+  struct CacheFile {
+    std::string name_;
+    uint64_t size_;
+    Envoy::SystemTime last_touch_;
+    // Reversed < operator because we're comparing timestamps but considering age, i.e.
+    // a greater timestamp is a lesser age.
+    bool operator<(const CacheFile& other) const { return other.last_touch_ < last_touch_; }
+  };
+  // A set of files sorted by last-touch timestamp, highest (i.e. youngest) first.
+  std::multiset<CacheFile> cache_files;
```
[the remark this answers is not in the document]
ravenblackx: Good point. Set made more sense when it was being dynamically updated and mutated, but now that's it's add-all remove-some, vector is better. Done.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+        -1) {
+      // May want to add logging here for cache eviction failure, but it's expected sometimes,
+      // e.g. if another instance of Envoy is performing cleanup at the same time, or some external
+      // operator deleted the file. If it fails we don't reduce the estimated cache size, so another
+      // eviction run will happen sooner.
+      // TODO(ravenblack): might be worth checking the type of the error, or whether the file is
+      // gone - if there's a permissions issue, for example, then the cache might remain oversized
+      // and the eviction thread will be churning, trying and failing to remove a file, which would
+      // be worth logging a warning, versus if the file is already gone then there's no problem.
+      cache.trackFileRemoved(it->size_);
```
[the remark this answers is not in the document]
ravenblackx: trackFileRemoved updates count and size. Clarified function comment to that effect.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+      // work is being performed.
+      absl::MutexLock lock(&cache_mu_);
+      caches = caches_;
+    }
+
+    for (const std::shared_ptr<CacheShared>& cache : caches) {
+      if (cache->needs_init_) {
+        init(*cache);
+      }
+      if (cache->needsEviction()) {
```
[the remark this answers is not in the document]
ravenblackx: Added an unchecked box to DESIGN.md for subsequent PR.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.h
```diff
@@ -0,0 +1,95 @@
...
+   * @return false if terminating.
+   */
+  bool waitForSignal();
+  void terminate();
+
+  absl::Mutex mu_;
+  bool signalled_ ABSL_GUARDED_BY(mu_) = false;
+  bool terminating_ ABSL_GUARDED_BY(mu_) = false;
+
+  absl::Mutex cache_mu_;
```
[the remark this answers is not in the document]
ravenblackx: Added comment and (slightly hacky) enforcement that they are never held together. (Surprised absl annotations doesn't have an annotation specifically for that.)

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
```diff
@@ -116,12 +121,75 @@ TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, InitialStatsAreSetCorrectly) {
...
 
+TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, EvictsOldestFilesUntilUnderCountLimit) {
+  const std::string file_contents = "XXXXX";
+  const uint64_t max_count = 2;
+  ConfigProto cfg = testConfig();
+  cfg.mutable_max_cache_entry_count()->set_value(max_count);
+  env_.writeStringToFileForTest(absl::StrCat(cache_path_, "cache-a"), file_contents, true);
+  env_.writeStringToFileForTest(absl::StrCat(cache_path_, "cache-b"), file_contents, true);
+  {
+    // TODO(ravenblack): replace this with backdating the files when that's possible.
```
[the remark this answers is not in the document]
ravenblackx: Well, there's two parts to the problem, one that it's not taking the API calls via injection, and one that mocking the API calls would be remarkably impractical, would make a hard-to-read test *and* wouldn't really test this very well - which is why I'm simultaneously trying to improve the non-async File class to have the capabilities we need here. Once that's done, we could use the fake filesystem, which would be much better in all three respects.

Hence a TODO - I would really rather not migrate to injected OsApi, and make a whole mess of painful mocks, then migrate *again* to injected FileSystem once that's ready to support this kind of thing.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
```diff
@@ -116,12 +121,75 @@ TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, InitialStatsAreSetCorrectly) {
...
+    absl::Mutex mu;
+    absl::MutexLock lock(&mu);
+    auto cond = []() ABSL_EXCLUSIVE_LOCKS_REQUIRED(mu) { return false; };
+    mu.AwaitWithTimeout(absl::Condition(&cond), absl::Seconds(1));
```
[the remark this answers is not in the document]
ravenblackx: Agree, but not at the cost of mocking stat() repeatedly. :)

Believe me, I am a *very* strong opponent of hard delays in tests, I definitely don't intend to leave this in long-term.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
```diff
@@ -116,12 +121,75 @@ TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, InitialStatsAreSetCorrectly) {
...
+  EXPECT_FALSE(Filesystem::fileSystemForTest().fileExists(absl::StrCat(cache_path_, "cache-a")));
+  EXPECT_FALSE(Filesystem::fileSystemForTest().fileExists(absl::StrCat(cache_path_, "cache-b")));
+  EXPECT_TRUE(Filesystem::fileSystemForTest().fileExists(absl::StrCat(cache_path_, "cache-c")));
+  EXPECT_TRUE(Filesystem::fileSystemForTest().fileExists(absl::StrCat(cache_path_, "cache-d")));
+}
+
+TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, EvictsOldestFilesUntilUnderSizeLimit) {
+  const std::string file_contents = "XXXXX";
+  const std::string large_file_contents = "XXXXXXXXXXXX";
+  const uint64_t max_size = large_file_contents.size();
```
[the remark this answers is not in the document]
ravenblackx: I don't think it's worth it here, there isn't really that much common (except the sleep-hack), and abstracting small pieces away would make it harder to follow what the test is actually doing.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
ravenblackx: The sequence of events here is:
1. trackFileAdded increments the count to 3, and signals eviction thread.
2. eviction thread counts the files, and sets the count to 3 again.
3. eviction thread evicts files, calls trackFileRemoved, decrementing the count (twice).

Added eviction_runs to stats, and included it in the tests.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,189 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Cache {
+namespace FileSystemHttpCache {
+
+namespace {
+bool isCacheFile(const Filesystem::DirectoryEntry& entry) {
+  return absl::StartsWith(entry.name_, "cache-") && entry.type_ == Filesystem::FileType::Regular;
```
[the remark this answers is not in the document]
ravenblackx: Done.

source/extensions/http/cache/file_system_http_cache/DESIGN.md
```diff
@@ -0,0 +1,37 @@
...
+* The cache can be configured with a maximum number of cache entry files, thereby effectively enforcing a maximum number of files per path.
+* A new cache entry that causes the cache to exceed the configured maximum size or maximum number of entries triggers the eviction thread to evict sufficient LRU entries to bring it back below the threshold\[s\] exceeded.
+* Each cache entry file starts with [a fixed structure header followed by a serialized proto](cache_file_header.proto), followed by proto-serialized headers, raw body and proto-serialized trailers.
+* Cache entry files are named `cache-` followed by a stable hash key for the entry.
+* (When implemented) the tree structure of folders is simply one level deep of folders named `cache-0000`, `cache-0001` etc. as four-digit hexadecimal numbers up to the configured number of subdirectories. Cache files are placed in a folder according to a short stable hash of their key. On cache startup, any cache entries found to be in the wrong folder (as would be the case if the number of folders was reconfigured) will simply be removed.
```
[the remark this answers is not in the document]
ravenblackx: It's part of measuring the initial cache size, so does need to be done reasonably early. I have (since this comment) migrated this action to the eviction thread so it doesn't block the main thread that's updating config.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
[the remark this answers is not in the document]
jmarantz: My preference would not be to mock stat(). My preference would be to add stat() to the File interface in envoy/filesystem/filesystem.h and provide a TimeSource to MemfileImpl, which would it would use for saving the timestamp.

This will also work well with the existing usage of simulated time in cache tests.

In any case I'm highly opposed to circumventing the attempts to block spurious usage of things that are equivalent to sleep() in tests. If we do ever decide that's the only way, we should have the code that implement sleep do so with a clear override of the policy, rather than squirreling in a substitute implementation that isn't in the lint yet.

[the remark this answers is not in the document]
ravenblackx: That matches my preference. Which is why I've been making steps in that direction in filesystem.h already. But I'd rather this PR not be blocked by the chain of PRs that will be required to get filesystem into a state where it's useable here. And I'd rather the test be able to do what it needs to do to be useful, in the meantime.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
jmarantz: got it -- I had missed that in my scan of the changes somehow.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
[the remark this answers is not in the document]
jmarantz: I don't want to add use-cases in the codebase of how to deliberately work around the lint check.

Why is a series of PRs needed to add stat support the memory-based filesystem? Can't that just be one PR?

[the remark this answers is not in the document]
ravenblackx: Discussed in slack, and filed #24994 to keep track.

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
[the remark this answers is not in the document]
jmarantz: OK. FWIW it might be worth scanning through the tests for mod_pagespeed's file cache: https://github.com/apache/incubator-pagespeed-mod/blob/master/test/pagespeed/kernel/cache/file_cache_test.cc

Some of these may not be relevant due to different eviction architecture (periodic cleaning rather than triggered by a write).

jmarantz (approved):
@toddmgreer  can you do a final pass and then I'll merge?

test/extensions/http/cache/file_system_http_cache/file_system_http_cache_test.cc
```diff
@@ -116,10 +121,71 @@ TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, InitialStatsAreSetCorrectly) {
...
+}
+
+TEST_F(FileSystemHttpCacheTestWithNoDefaultCache, EvictsOldestFilesUntilUnderCountLimit) {
+  const std::string file_contents = "XXXXX";
+  const uint64_t max_count = 2;
+  ConfigProto cfg = testConfig();
+  cfg.mutable_max_cache_entry_count()->set_value(max_count);
+  env_.writeStringToFileForTest(absl::StrCat(cache_path_, "cache-a"), file_contents, true);
+  env_.writeStringToFileForTest(absl::StrCat(cache_path_, "cache-b"), file_contents, true);
+  // TODO(#24994): replace this with backdating the files when that's possible.
```
jmarantz, line 139: nit: "backdating" is not what I was thinking of. Simulated time is what I was thinking of, with a file-system that pays attention to it.

ravenblackx: Sure. Either way would work, and I don't think it's important to specify accurately which it's going to be in a TODO. :)
(Fake time is just pre-emptively backdating the files, anyway.)

source/extensions/http/cache/file_system_http_cache/config.cc
```diff
@@ -56,9 +59,11 @@ class CacheSingleton : public Envoy::Singleton::Instance {
       cache = it->second.lock();
     }
     if (!cache) {
-      cache = std::make_shared<FileSystemHttpCache>(
-          singleton, std::move(config),
-          async_file_manager_factory_->getAsyncFileManager(config.manager_config()), stats_scope);
+      auto async_file_manager =
```
jmarantz, line 62: nit: use an explicit type here (not obvious from context although I could guess).

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
```diff
@@ -21,21 +22,33 @@ namespace FileSystemHttpCache {
...
+  ~FileSystemHttpCache() override;
 
   // Overrides for HttpCache
   LookupContextPtr makeLookupContext(LookupRequest&& lookup,
                                      Http::StreamDecoderFilterCallbacks& callbacks) override;
   InsertContextPtr makeInsertContext(LookupContextPtr&& lookup_context,
                                      Http::StreamEncoderFilterCallbacks& callbacks) override;
   CacheInfo cacheInfo() const override;
-  const CacheStats& stats() const { return stats_; }
+  const CacheStats& stats() const;
```
jmarantz, line 51: I think it's OK for the implementation to be inline but it's up to you. Same for config().

jmarantz (dismissed):
Looks great, thanks. Sorry for the delay -- been dealing with a lot of stuff the past couple weeks.

jmarantz:
Todd can you take a pass?

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
```diff
@@ -205,17 +214,67 @@ class FileSystemHttpCache : public HttpCache,
       entries_being_written_ ABSL_GUARDED_BY(cache_mu_);
 
   std::shared_ptr<Common::AsyncFiles::AsyncFileManager> async_file_manager_;
-  CacheStats stats_;
 
+  // Stats and config are held in a shared_ptr so that CacheEvictionThread can use
+  // them even if the cache instance has been deleted while it performed work.
+  std::shared_ptr<CacheShared> shared_;
+
+  CacheEvictionThread& cache_eviction_thread_;
```
toddmgreer, line 222: cache_eviction_thread_ needs a comment that it must be after owner_, since destruction of owner_ could destroy the object to which cache_eviction_thread_ refers.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
```diff
@@ -0,0 +1,188 @@
...
+namespace {
+bool isCacheFile(const Filesystem::DirectoryEntry& entry) {
+  return entry.type_ == Filesystem::FileType::Regular && absl::StartsWith(entry.name_, "cache-");
+}
+} // namespace
+
+CacheEvictionThread::CacheEvictionThread(Thread::ThreadFactory& thread_factory)
+    : thread_(thread_factory.createThread([this]() { work(); })) {}
+
+CacheEvictionThread::~CacheEvictionThread() {
```
toddmgreer, line 26: The pattern of calling ~CacheEvictionThread while CacheEvictionThread methods may be executing on another thread is subtly fragile. It's well-defined, but only because CacheEvictionThread is the most-derived class. Otherwise, any virtual methods running in another thread would cause undefined behavior.

Therefore, please either declare CacheSingleton as `final` (with a comment), or call `join` before ~CacheEvectionThread (also with a comment).

toddmgreer (changes requested):

source/extensions/http/cache/file_system_http_cache/config.cc
ravenblackx: Done.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
ravenblackx: Can't be inline without significant rejiggering, since the shared class where these values live is declared after the main class. Since the dependencies go both ways, it's a case where moving stuff out of headers is a good option.

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.h
```diff
@@ -0,0 +1,97 @@
...
+   */
+  void removeCache(std::shared_ptr<CacheShared>& cache);
+
+  /**
+   * Signals the cache eviction thread that it's time to test things.
+   * After receiving a signal, the thread will exit if terminating_ is set.
+   * Otherwise it will call each cache's `maybeEvict` function in an arbitrary
+   * order.
+   */
+  void signal();
```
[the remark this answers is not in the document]
ravenblackx: Moved much of this comment onto 'work' since that's where the described actions actually happen, and clarified both remaining parts.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
ravenblackx: Done (though I'm not sure it's required - it's only after the destructor completes that the class members are destroyed, and at that point you can't use this reference anyway.)

source/extensions/http/cache/file_system_http_cache/cache_eviction_thread.cc
ravenblackx: I'm fairly sure you meant `CacheEvictionThread` as final - done.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
jmarantz: makes sense; thanks.

source/extensions/http/cache/file_system_http_cache/file_system_http_cache.h
toddmgreer: Thank you for the correction. My mental pattern matching misfired and I didn't notice.

toddmgreer (approved):

jmarantz (approved):

PR #24645, opened 2022-12-21
mobile: remove timer & distribution methods from PulseClient

abeyad:
These methods are not used and we'd like to simplify the PulseClient APIs before working on feature parity for the C++ PulseClient API.

Risk Level: Low
Testing: //test/...
Docs Changes: N/A
Release Notes: included


Augustyniak (approved):
Looks good, thank you for the change. Do you have plans to get rid of the gauge too? It has never been used and as far as I understand it's usefullness in the content of mobile apps is limited.

abeyad:
> Looks good, thank you for the change. Do you have plans to get rid of the gauge too? It has never been used and as far as I understand it's usefullness in the content of mobile apps is limited.

Thanks Rafal!  I wasn't sure about gauge, seemed very similar to counter, hence why I kept in this PR, but since you say it's never been used and it's not useful for mobile, I will send a follow up PR now to remove gauge too.

PR #24648, opened 2022-12-21
mobile: remove gauge methods from PulseClient

abeyad:
The `gauge` methods are not useful in the mobile context, and we'd like to simplify the PulseClient APIs before working on feature parity for the C++ PulseClient API.  This is a follow-up to https://github.com/envoyproxy/envoy/pull/24645, where we removed the `timer` and `distribution` methods from PulseClient.

Risk Level: Low
Testing: //test/...
Docs Changes: N/A
Release Notes: included


Augustyniak (approved):
Thanks for doing this!

RyanTheOptimist (approved):

PR #24651, opened 2022-12-21
mobile: Ensure no admin interface in integration test Bootstrap config

abeyad:


mobile/test/common/integration/base_client_integration_test.cc
```diff
@@ -188,6 +188,8 @@ void BaseClientIntegrationTest::createEnvoy() {
   finalizeConfigWithPorts(config_helper_, ports, use_lds_);
 
   if (override_builder_config_) {
+    ASSERT_FALSE(config_helper_.bootstrap().has_admin())
```
Augustyniak, line 191: help me to understand the significance of this assert - we are just making sure that by default the config does not enable admin endpoint as we are afraid that we will accidentally flip the default/enable it, is that correct?

Augustyniak (commented):
one question

RyanTheOptimist (approved):

mobile/test/common/integration/base_client_integration_test.cc
abeyad: that's right... basically if a test that inherits from BaseClientIntegrationTest sets an admin interface, they'll get an error from the bowels of Envoy saying that the admin extension needs to be added to use the admin interface.  But for the EM integration tests, we want the developer to know that they should not depend on the admin interface at all (instead of trying to add it as an extension to overcome the error).

Note that the tests in this PR will fail until https://github.com/envoyproxy/envoy/pull/24607 is landed, so still waiting on approvals for that.

Augustyniak (dismissed):

RyanTheOptimist (approved):

PR #24656, opened 2022-12-21
Make round trip time accessible through streaminfo

DiazAlan:
Signed-off-by: AlanDiaz <diazalan@google.com>

Commit Message: Make round trip time accessible through stream info
Additional Description: Make the last round trip time of the previous handshake accessible through stream info. This will allow us to make it accessible in access logs,
Risk Level: Low
Testing: Added an integration test in filter integration test that adds roundtriptime to the header and checks if it exists
Docs Changes: NA
Release Notes:  NA
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/envoy/network/socket.h
+++ b/envoy/network/socket.h
@@ -107,6 +107,11 @@ public:
    * @return ja3 fingerprint hash of the downstream connection, if any.
    */
   virtual absl::string_view ja3Hash() const PURE;
+
+  /**
+   * @return roundTripTime of the connection
+   */
+  virtual const absl::optional<std::chrono::milliseconds>& roundTripTime() const PURE;
 };
 
 class ConnectionInfoSetter : public ConnectionInfoProvider {
@@ -168,6 +173,11 @@ public:
    * @param JA3 fingerprint.
    */
   virtual void setJA3Hash(const absl::string_view ja3_hash) PURE;
+
+  /**
+   * @param  milliseconds of round trip time of previous connection
+   */
+  virtual void setRoundTripTime(std::chrono::milliseconds roundTripTime) PURE;
 };
 
 using ConnectionInfoSetterSharedPtr = std::shared_ptr<ConnectionInfoSetter>;
--- a/source/common/http/filter_manager.h
+++ b/source/common/http/filter_manager.h
@@ -582,6 +582,9 @@ public:
   absl::string_view ja3Hash() const override {
     return StreamInfoImpl::downstreamAddressProvider().ja3Hash();
   }
+  const absl::optional<std::chrono::milliseconds>& roundTripTime() const override {
+    return StreamInfoImpl::downstreamAddressProvider().roundTripTime();
+  }
 
 private:
   Network::Address::InstanceConstSharedPtr overridden_downstream_remote_address_;
--- a/source/common/network/socket_impl.h
+++ b/source/common/network/socket_impl.h
@@ -67,6 +67,12 @@ public:
   }
   absl::string_view ja3Hash() const override { return ja3_hash_; }
   void setJA3Hash(const absl::string_view ja3_hash) override { ja3_hash_ = std::string(ja3_hash); }
+  const absl::optional<std::chrono::milliseconds>& roundTripTime() const override {
+    return round_trip_time_;
+  }
+  void setRoundTripTime(std::chrono::milliseconds round_trip_time) override {
+    round_trip_time_ = round_trip_time;
+  }
 
 private:
   Address::InstanceConstSharedPtr local_address_;
@@ -79,6 +85,7 @@ private:
   absl::optional<std::string> interface_name_;
   Ssl::ConnectionInfoConstSharedPtr ssl_info_;
   std::string ja3_hash_;
+  absl::optional<std::chrono::milliseconds> round_trip_time_;
 };
 
 class SocketImpl : public virtual Socket {
--- a/source/server/active_tcp_listener.cc
+++ b/source/server/active_tcp_listener.cc
@@ -101,6 +101,12 @@ void ActiveTcpListener::onReject(RejectCause cause) {
 void ActiveTcpListener::onAcceptWorker(Network::ConnectionSocketPtr&& socket,
                                        bool hand_off_restored_destination_connections,
                                        bool rebalanced) {
+  // Get Round Trip Time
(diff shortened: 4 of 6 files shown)


ggreenway (changes requested):
I think it makes sense to expose this, but I'm not sure if it's a good idea to latch the value right after the connection is setup. At least on linux, the TCP_INFO syscall that provides this provides a smoothed RTT, and it may not be very smoothed right after connection setup (very few round trips at that point). Also, the RTT can vary over the life of the connection. If this is being logged on a specific http connection, we'd want to sample the RTT near-in-time to when that particular request happened. Maybe it would make more sense to latch this in `Http::StreamDecoderFilter::onStreamComplete()` or near that time.

/wait-any

envoy/network/socket.h
```diff
@@ -168,6 +173,11 @@ class ConnectionInfoSetter : public ConnectionInfoProvider {
    * @param JA3 fingerprint.
    */
   virtual void setJA3Hash(const absl::string_view ja3_hash) PURE;
+
+  /**
+   * @param  milliseconds of round trip time of previous connection
```
yanavlasov, line 178: Why is it `previous`?

yanavlasov:
> I think it makes sense to expose this, but I'm not sure if it's a good idea to latch the value right after the connection is setup. At least on linux, the TCP_INFO syscall that provides this provides a smoothed RTT, and it may not be very smoothed right after connection setup (very few round trips at that point). Also, the RTT can vary over the life of the connection. If this is being logged on a specific http connection, we'd want to sample the RTT near-in-time to when that particular request happened. Maybe it would make more sense to latch this in `Http::StreamDecoderFilter::onStreamComplete()` or near that time.
> 
> /wait-any

It makes sense. The eventual goal is to log the RTT in the access log. I agree that ideally it should be done at the end of the HTTP request or TCP connection, right before access log is to be emitted. We will need to deal with the fact that H/1 requests can be framed be connection closure and at this point RTT may not be available but this may be too far of an edge case.

Would it be ok if these additional samples are implemented in a follow-up PR, to make it easier to iterate on this change?

ggreenway:
> Would it be ok if these additional samples are implemented in a follow-up PR, to make it easier to iterate on this change?

I think a lot of the code in this PR will have to be changed/reverted to get to that solution, so if we agree that we should change to sampling sometime during the request (beginning or end of request probably) then I'm not sure it makes sense to merge this as-is.

yanavlasov:
> > Would it be ok if these additional samples are implemented in a follow-up PR, to make it easier to iterate on this change?
> 
> I think a lot of the code in this PR will have to be changed/reverted to get to that solution, so if we agree that we should change to sampling sometime during the request (beginning or end of request probably) then I'm not sure it makes sense to merge this as-is.

Hmm, do you have a different approach for making RTT accessible in StreamInfo? I did not think we will need to change much, since both HCM or FilterManager or TCP proxy have access to the Network::Connection that had `ConnectionInfoSetter` and the ' lastRoundTripTime`. It can be optimized a bit by adding the `recordLastRoundTripTime` method to Network::Connection in the ConnectionInfoProvider, but I would not think we will need to redesign the existing code. We still want to record the RTT from TCP handshake in case connection fails right after that.

ggreenway:
> Hmm, do you have a different approach for making RTT accessible in StreamInfo? I did not think we will need to change much, since both HCM or FilterManager or TCP proxy have access to the Network::Connection that had `ConnectionInfoSetter` and the ' lastRoundTripTime`. It can be optimized a bit by adding the `recordLastRoundTripTime` method to Network::Connection in the ConnectionInfoProvider, but I would not think we will need to redesign the existing code. We still want to record the RTT from TCP handshake in case connection fails right after that.

I agree that accessing it via stream_info through the `ConnectionInfoProvider` is the right place to put it. But when it is latched from a different place, the changes in this PR to proxy protocol tests and active_tcp_listener will need to be reverted, which by number-of-lines is about half of the PR.

source/extensions/listener_managers/listener_manager/active_tcp_listener.cc
```diff
@@ -101,6 +101,12 @@ void ActiveTcpListener::onReject(RejectCause cause) {
 void ActiveTcpListener::onAcceptWorker(Network::ConnectionSocketPtr&& socket,
                                        bool hand_off_restored_destination_connections,
                                        bool rebalanced) {
+  // Get Round Trip Time
+  absl::optional<std::chrono::milliseconds> t = socket->lastRoundTripTime();
```
ggreenway, line 105: I think this block should move to `Envoy::Http::FilterManager::onStreamComplete()`, which has access to the `Network::Connection`.

ggreenway (changes requested):

source/extensions/listener_managers/listener_manager/active_tcp_listener.cc
yanavlasov: Why move? In this case RTT will only be available for HTTP requests that end up creating streams. It will not cover the case where stream was not created, i.e. due to a protocol error. Or TCP to CONNECT case.

I think it makes sense to **add** it in the place you are suggesting, but still keeping the code for getting the initial measurement from TCP handshake. WDYT?

ggreenway (dismissed):

source/extensions/listener_managers/listener_manager/active_tcp_listener.cc
DiazAlan: Added to conn_manager_impl right before stream completes.

source/common/http/conn_manager_impl.cc
```diff
@@ -277,6 +277,13 @@ void ConnectionManagerImpl::doDeferredStreamDestroy(ActiveStream& stream) {
   }
 
   stream.completeRequest();
+
+  // Set roundtrip time in connectionInfoSetter before OnStreamComplete
+  absl::optional<std::chrono::milliseconds> t = read_callbacks_->connection().lastRoundTripTime();
```
ggreenway, line 282: Please add a test that validates this happens before `log()` is called.

ggreenway (changes requested):
/wait

yanavlasov (approved):

ggreenway (approved):

PR #24659, opened 2022-12-22
Reduce Route memory utilization by avoiding `absl::optional` overhead for optional timeouts

yurykats:
Reduce Route memory utilization by avoiding `absl::optional` overhead for optional timeouts.

`std::duration` values wrapped in an absl::optional take twice as much memory as the durations themseves -- 16 bytes for each optional milliseconds value.

This PR removes the optional overhead by
1. Packing all timeouts into a separate data structure, sharing a single bool to track which timeout is set. This saves 8 bytes per timeout.
2. Not creating the data structure at all, if none of the optional timeouts are present. This saves another 8 bytes per timeout, for routes that don't configure them.

There are 6 optional timeouts, so this saves either 40 or 90 bytes per route.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/router/config_impl.h
+++ b/source/common/router/config_impl.h
@@ -547,6 +547,60 @@ private:
 };
 using DefaultInternalRedirectPolicy = ConstSingleton<InternalRedirectPolicyImpl>;
 
+// Manages various optional route timeouts. We pack them in a separate data structure for memory
+// efficiency:
+// - Sharing one bool avoids the overhead of `absl::optional`, saving 8 bytes per timeout
+// - If none of the timeouts are configured, we avoid tracking any of the `std::duration` values,
+// saving another 8 bytes per timeout
+class OptionalTimeouts {
+public:
+  OptionalTimeouts(const envoy::config::route::v3::RouteAction& route);
+
+  absl::optional<std::chrono::milliseconds> idleTimeout() const {
+    return has_idle_timeout_ ? absl::optional<std::chrono::milliseconds>(idle_timeout_)
+                             : absl::nullopt;
+  }
+  absl::optional<std::chrono::milliseconds> maxStreamDuration() const {
+    return has_max_stream_duration_
+               ? absl::optional<std::chrono::milliseconds>(max_stream_duration_)
+               : absl::nullopt;
+  }
+  absl::optional<std::chrono::milliseconds> grpcTimeoutHeaderMax() const {
+    return has_grpc_timeout_header_max_
+               ? absl::optional<std::chrono::milliseconds>(grpc_timeout_header_max_)
+               : absl::nullopt;
+  }
+  absl::optional<std::chrono::milliseconds> grpcTimeoutHeaderOffset() const {
+    return has_grpc_timeout_header_offset_
+               ? absl::optional<std::chrono::milliseconds>(grpc_timeout_header_offset_)
+               : absl::nullopt;
+  }
+  absl::optional<std::chrono::milliseconds> maxGrpcTimeout() const {
+    return has_max_grpc_timeout_ ? absl::optional<std::chrono::milliseconds>(max_grpc_timeout_)
+                                 : absl::nullopt;
+  }
+  absl::optional<std::chrono::milliseconds> grpcTimeoutOffset() const {
+    return has_grpc_timeout_offset_
+               ? absl::optional<std::chrono::milliseconds>(grpc_timeout_offset_)
+               : absl::nullopt;
+  }
+
+private:
+  std::chrono::milliseconds idle_timeout_;
+  std::chrono::milliseconds max_stream_duration_;
+  std::chrono::milliseconds grpc_timeout_header_max_;
+  std::chrono::milliseconds grpc_timeout_header_offset_;
+  std::chrono::milliseconds max_grpc_timeout_;
+  std::chrono::milliseconds grpc_timeout_offset_;
+  // Keep small members at the end of class, to reduce alignment overhead.
+  bool has_idle_timeout_ : 1;
+  bool has_max_stream_duration_ : 1;
+  bool has_grpc_timeout_header_max_ : 1;
+  bool has_grpc_timeout_header_offset_ : 1;
+  bool has_max_grpc_timeout_ : 1;
+  bool has_grpc_timeout_offset_ : 1;
+};
+
 /**
  * Base implementation for all route entries.q
  */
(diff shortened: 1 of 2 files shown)


source/common/router/config_impl.h
```diff
@@ -547,6 +547,60 @@ class InternalRedirectPolicyImpl : public InternalRedirectPolicy {
...
+  std::chrono::milliseconds grpc_timeout_header_offset_;
+  std::chrono::milliseconds max_grpc_timeout_;
+  std::chrono::milliseconds grpc_timeout_offset_;
+  // Keep small members at the end of class, to reduce alignment overhead.
+  bool has_idle_timeout_ : 1;
+  bool has_max_stream_duration_ : 1;
+  bool has_grpc_timeout_header_max_ : 1;
+  bool has_grpc_timeout_header_offset_ : 1;
+  bool has_max_grpc_timeout_ : 1;
+  bool has_grpc_timeout_offset_ : 1;
```
stevenzzzz, line 601: const ? if they are from config. 
same apply to the timeouts above

source/common/router/config_impl.h
```diff
@@ -547,6 +547,60 @@ class InternalRedirectPolicyImpl : public InternalRedirectPolicy {
...
+  }
+  absl::optional<std::chrono::milliseconds> grpcTimeoutOffset() const {
+    return has_grpc_timeout_offset_
+               ? absl::optional<std::chrono::milliseconds>(grpc_timeout_offset_)
+               : absl::nullopt;
+  }
+
+private:
+  std::chrono::milliseconds idle_timeout_;
+  std::chrono::milliseconds max_stream_duration_;
```
stevenzzzz, line 590: You can go further define them as unique_ptr<T>, which takes much less space than "long long".

I think you could also just define them as uint32_t, chrono::milliseconds is trivially constructible from int.

source/common/router/config_impl.h
```diff
@@ -1076,12 +1137,7 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
   const Http::LowerCaseString cluster_header_name_;
   ClusterSpecifierPluginSharedPtr cluster_specifier_plugin_;
   const std::chrono::milliseconds timeout_;
-  const absl::optional<std::chrono::milliseconds> idle_timeout_;
-  const absl::optional<std::chrono::milliseconds> max_stream_duration_;
-  const absl::optional<std::chrono::milliseconds> grpc_timeout_header_max_;
-  const absl::optional<std::chrono::milliseconds> grpc_timeout_header_offset_;
-  const absl::optional<std::chrono::milliseconds> max_grpc_timeout_;
-  const absl::optional<std::chrono::milliseconds> grpc_timeout_offset_;
+  std::unique_ptr<const OptionalTimeouts> optional_timeouts_;
```
stevenzzzz, line 1140: const

source/common/router/config_impl.h
yurykats: milliseconds and unique_ptr are both 8 bytes, there is nothing to gain from the unique_ptr.

source/common/router/config_impl.h
yurykats: I need to assign them, so can't be const.

source/common/router/config_impl.h
yurykats: The common patter here is `unique_ptr<const T>`, this is consistent with the rest.

yurykats:
/asssign @jmarantz @ggreenway -- since both of you we reviewing the other approach in https://github.com/envoyproxy/envoy/pull/24626 and have context. Thanks!

source/common/router/config_impl.h
pradeepcrao: These can be assigned in the member initialiser list of the constructor, and be kept const (which is a common and preferred pattern used in the Envoy codebase).

source/common/router/config_impl.h
jmarantz: agreed; but maybe drop a TODO because I'm not sure we really need to represent more than 4 billion milliseconds (about 46 days) for these timeouts; so you could hold them in this structure as uint32_t if you want to save some memory.

Maybe you don't want to do that in this PR since that would be a semantic change.

source/common/router/config_impl.h
jmarantz: that looks a little awkward because we assign 2 member vars for every input arg, but it doesn't look too bad :)

source/common/router/config_impl.h
```diff
@@ -547,6 +547,60 @@ class InternalRedirectPolicyImpl : public InternalRedirectPolicy {
...
+  }
+
+private:
+  std::chrono::milliseconds idle_timeout_;
+  std::chrono::milliseconds max_stream_duration_;
+  std::chrono::milliseconds grpc_timeout_header_max_;
+  std::chrono::milliseconds grpc_timeout_header_offset_;
+  std::chrono::milliseconds max_grpc_timeout_;
+  std::chrono::milliseconds grpc_timeout_offset_;
+  // Keep small members at the end of class, to reduce alignment overhead.
```
jmarantz, line 595: nice! love how this packs.

source/common/router/config_impl.h
stevenzzzz: I vaguely recall there is some optimization that makes unique_ptr 0 cost when it's empty. 

The other way to save would be to use array with mask but probably not worth the extra complexity.

source/common/router/config_impl.h
stevenzzzz: I would suggest to not lose the constness on these timeouts. I dont see the pattern or what it brings us.

stevenzzzz (commented):
LGTM

source/common/router/config_impl.h
yurykats: I have conditional checks for existence there, member initialiser list of the constructor is not super useful.

pradeepcrao:
LGTM

source/common/router/config_impl.h
jmarantz: Right - that's what I was not suggesting doing that in this PR but it might be good to drop a TODO or just queue it onto the issue where you are tracking potential optimizations.

yurykats: @pradeepcrao is already working on a follow up improvement here. 

I don't think a TODO to switch to int32s would be very productive. Given that the current proto config supports int64 values, it would be a hard change to make. There are probably users out there that had set it to int64::max as a way of saying "no timeout".

jmarantz (approved):
ok lgtm from my perspective. This is a win for memory consumption under all situations, and an even bigger win of there are no timeouts set at all. It adds an indirection if there are any timeouts set, consistent with other optional features.

Greg: would definitely need a non-google approver here IMO.

ggreenway (changes requested):
LGTM

Depending on configuration, if many routes have identical timeouts, you could keep a registry of them and share the `OptionalTimeouts` between all routes with identical timeout settings, if you're looking for further memory reductions.

/wait

yurykats:
> Depending on configuration, if many routes have identical timeouts, you could keep a registry of them and share the `OptionalTimeouts` between all routes with identical timeout settings, if you're looking for further memory reductions.

Yes, absolutely. This applies to many different aspects of configuration, not just timeouts, so we are planning a generic "repeated config cache" solution in the near future.

ggreenway:
Please don't force-push to PRs. It makes it much more difficult to review. https://github.com/envoyproxy/envoy/blob/28db3d51c018a722703ce7b43f99b2713ce9b071/CONTRIBUTING.md?plain=1#L135

ggreenway (approved):

yurykats:
> Please don't force-push to PRs. It makes it much more difficult to review.

Sorry about that! My accepting of your suggested edit in the github UI messed up DCO, and the instructions to correct it told me to force push. 

Thanks for the review!

PR #24661, opened 2022-12-22
util: make OptRef not derive from absl::optional<std::reference<>> to reduce size footprint from 16 bytes to 8.

jmarantz:
Commit Message: I wrote OptRef to add syntactic sugar to users of optional references, which were very awkward to program with. However @pianiststickman recently observed with sadness that it consumes aligned(8+1) bytes, ie 16 bytes, where in previous times we would just used a pointer and gone on with our lives :) This often doesn't matter unless we find OptRefs in structs with high cardinality.

This change drops the inheritance from `optional<reference_wrapper<>>` and just implements the required functionality directly. Some of the functionality inherited from absl::optional was hard to replicate, and was generally getting in the way of correctness, performance, or readability. So in a few cases (7 files), small modifications to call-sites had to be supplied.

Notable among them:
 * It was possible to assign `optional<reference_wrapper<T>>` to an `optional<T>` which would copy the object if it has a value. I changed that to require calling a `copy()` method, as I think the previous value-copy was non-obvious and should be explicit.
 * There is some property of  `optional<reference_wrapper<T>>` where you could take a `const optional<reference_wrapper<T>>` and extract out a non-const `T&`. I think this does not make sense and didn't want to replicate it in this container object. This was exploited in a few places. In some of them it was easy to change the call-sites to use a `const T&` but in one of them I added an explicit `const_cast` at the call-site where this was needed.
 
This change duplicates some of the methods from absl::optional that were previously inherited, and thus explicit tests were added to ensure it the wrapper is covered completely with its own unit test, which I verified manually in the debugger.

Additional Description:
Risk Level: low -- this is a low level class but it is well tested and generally errors should be found by the compiler.
Testing: //test/...
Docs Changes: n/a
Release Notes: n/a


yurykats:
Great idea, thank you!

envoy/common/optref.h
```diff
@@ -13,76 +13,125 @@ namespace Envoy {
...
+   * where this const-stripping expectation occurs is
+   * source/common/http/async_client_impl.cc where the constructor has an
+   * arg const AsyncClient::StreamOptions& options, that is used to populate
+   * a non-const Config reference.
+   *
+   * This must be called with has_value() true.
+   *
+   * @return a reference_wrapper around the value.
+   */
+  std::reference_wrapper<T> value() const { return std::reference_wrapper<T>(*ptr_); }
```
yanavlasov, line 126: I think constness is not right here. Probably needs some tests. I would expect `std::reference_wrapper<const T> value() const`

jmarantz: Agreed. See the comment. I think the usage is wrong and I'm not sure how that compiles with the existing impl based on `optional<reference_wraper>`

jmarantz: I think this is addressed now. Resolving.

source/common/http/async_client_impl.cc
```diff
@@ -80,7 +80,8 @@ AsyncClient::Stream* AsyncClientImpl::start(AsyncClient::StreamCallbacks& callba
 AsyncStreamImpl::AsyncStreamImpl(AsyncClientImpl& parent, AsyncClient::StreamCallbacks& callbacks,
                                  const AsyncClient::StreamOptions& options)
     : parent_(parent), stream_callbacks_(callbacks), stream_id_(parent.config_.random_.random()),
-      router_(options.filter_config_ ? options.filter_config_.value().get() : parent.config_,
+      router_(options.filter_config_ ? const_cast<Router::FilterConfig&>(*options.filter_config_)
```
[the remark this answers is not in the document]
jmarantz: This is not really a change in behavior. The current condition is that options.filter_config_.value().get() gives you a non-const ref from a constant `absl::optional<std::reference_wrapper<....>>`, so the usage model requires an effective const-cast. I can look into remove the false `const` claim at the ctor's interface and see what unravels when I pull that thread :)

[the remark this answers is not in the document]
jmarantz: actually it's not too bad; I just needed to remove 2 layers of lies about claiming `const`.

source/common/http/async_client_impl.cc
```diff
@@ -71,16 +71,16 @@ AsyncClient::Request* AsyncClientImpl::send(RequestMessagePtr&& request,
...
   LinkedList::moveIntoList(std::move(new_stream), active_streams_);
   return active_streams_.front().get();
 }
 
 AsyncStreamImpl::AsyncStreamImpl(AsyncClientImpl& parent, AsyncClient::StreamCallbacks& callbacks,
-                                 const AsyncClient::StreamOptions& options)
+                                 AsyncClient::StreamOptions& options)
     : parent_(parent), stream_callbacks_(callbacks), stream_id_(parent.config_.random_.random()),
-      router_(options.filter_config_ ? options.filter_config_.value().get() : parent.config_,
+      router_(options.filter_config_ ? *options.filter_config_ : parent.config_,
```
jmarantz, line 83: note: I also tried the opposite approach here, of constructing router with a const config. But that doesn't work because the config isn't really just a read-only configuration; it's got other context bundled into it that requires non-const access, so it was easier to just change the expectations up the stack to deal with non-const configs.

This added 10 files to this PR, but going the other direction (constifying all the way down) appeared to be much harder.

yanavlasov: I see. It "worked" because std::reference_wrapper behaved like std::unique_ptr by returning constness of reference type regarding of the constness of the wrapper itself. https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper/get

And the new OptRef returns `const T&` if OptRef itself is const. Also its API was inconsistent. Calling OptRef<>::get() was not the same as calling `value().get();` with respect to constness of types.

I think conceptually returning T for cont OptRef is more correct since constness of OptRef<T> should not imply constness of T. Similar to how std::reference_wrapper works.

jmarantz: OK I gave that a try; that enabled me to revert the chain of non-constness I had added previously also.

I'm not sure how I feel about extracting a non-const T from a const OptRef<T> -- I guess it's OK but I wasn't thinking about it like that.

yanavlasov (approved):

jmarantz:

need non-google approver

jmarantz:
Added @snowp for non-google approver

source/common/http/async_client_impl.cc
snowp: Interesting stuff, clearly const-ness is tricky

snowp (approved):
Thanks!

jmarantz:
Thanks Yan & Snow!

PR #24667, opened 2022-12-22
bazel: update to 6.0.0

jwendell:
This PR also cherry-picks a couple of needed commits from main


jwendell:
Hey folks, can I have an approval here? Bazel 6.0 is being used in `main` for quite a while.

keith (approved):
looks good, the current failure here is bogus, but you might have to push an empty commit to fix it

PR #24673, opened 2022-12-22
Reduce Route memory utilization by avoiding weighed clusters config overhead when no such clusters are configured

yurykats:
Reduce Route memory utilization by avoiding weighted clusters config overhead when no such clusters are configured

Currently each Route instance contains 3 member variables that are representing weighted clusters config, whether the route has such clusters or not. For routes without weighted clusters that's 56 bytes overhead (empty vector, empty string, int64).

This PR packs all weighted cluster related config into a separate data structure and only creates an instance of it when the route has such clusters. This is saving 50 bytes for any route that does not have weighted clusters.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/router/config_impl.h
+++ b/source/common/router/config_impl.h
@@ -963,9 +963,14 @@ public:
   // We keep them in a separate data structure to avoid memory overhead for the routes that do not
   // use weightless clusters.
   struct WeightedClustersConfig {
-    std::vector<WeightedClusterEntrySharedPtr> weighted_clusters_;
-    uint64_t total_cluster_weight_;
-    std::string random_value_header_name_;
+    WeightedClustersConfig(const std::vector<WeightedClusterEntrySharedPtr>& weighted_clusters,
+                           uint64_t total_cluster_weight,
+                           const std::string& random_value_header_name)
+        : weighted_clusters_(weighted_clusters), total_cluster_weight_(total_cluster_weight),
+          random_value_header_name_(random_value_header_name) {}
+    const std::vector<WeightedClusterEntrySharedPtr> weighted_clusters_;
+    const uint64_t total_cluster_weight_;
+    const std::string random_value_header_name_;
   };
 
 protected:
--- a/source/common/router/config_impl.h
+++ b/source/common/router/config_impl.h
@@ -961,7 +961,7 @@ public:
   using WeightedClusterEntrySharedPtr = std::shared_ptr<WeightedClusterEntry>;
   // Container for route config elements that pertain to a weightless clusters.
   // We keep them in a separate data structure to avoid memory overhead for the routes that do not
-  // use weighless clusters.
+  // use weightless clusters.
   struct WeightedClustersConfig {
     std::vector<WeightedClusterEntrySharedPtr> weighted_clusters_;
     uint64_t total_cluster_weight_;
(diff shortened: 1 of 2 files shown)


source/common/router/config_impl.h
```diff
@@ -959,6 +959,14 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
   };
 
   using WeightedClusterEntrySharedPtr = std::shared_ptr<WeightedClusterEntry>;
+  // Container for route config elements that pertain to a weightless clusters.
+  // We keep them in a separate data structure to avoid memory overhead for the routes that do not
+  // use weighless clusters.
+  struct WeightedClustersConfig {
+    std::vector<WeightedClusterEntrySharedPtr> weighted_clusters_;
+    uint64_t total_cluster_weight_;
+    std::string random_value_header_name_;
```
adisuissa, line 968: nit: keep const (possibly by introducing a c'tor)

source/common/router/config_impl.h
```diff
@@ -959,6 +959,14 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
   };
 
   using WeightedClusterEntrySharedPtr = std::shared_ptr<WeightedClusterEntry>;
+  // Container for route config elements that pertain to a weightless clusters.
+  // We keep them in a separate data structure to avoid memory overhead for the routes that do not
+  // use weighless clusters.
+  struct WeightedClustersConfig {
+    std::vector<WeightedClusterEntrySharedPtr> weighted_clusters_;
+    uint64_t total_cluster_weight_;
```
adisuissa, line 967: This may also be converted to const, by doing the computation using a local var.

source/common/router/config_impl.h
adisuissa: On a second look, why not make all const, by doing local var computations and moving them at the end to a new std::unique_ptr<WeightedClustersConfig>?

adisuissa (commented):
Thanks. Small nit, but otherwise LGTM.

source/common/router/config_impl.h

source/common/router/config_impl.h

source/common/router/config_impl.h
```diff
@@ -959,6 +959,19 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
   };
 
   using WeightedClusterEntrySharedPtr = std::shared_ptr<WeightedClusterEntry>;
+  // Container for route config elements that pertain to a weightless clusters.
```
adisuissa, line 962:
```suggestion
  // Container for route config elements that pertain to a weighted clusters.
```

source/common/router/config_impl.h
```diff
@@ -959,6 +959,19 @@ class RouteEntryImplBase : public RouteEntryAndRoute,
...
+  // We keep them in a separate data structure to avoid memory overhead for the routes that do not
+  // use weightless clusters.
```
adisuissa, line 964:
```suggestion
  // use weighted clusters.
```

adisuissa (commented):
Thanks, LGTM.
A couple more comment nits that I've missed on the first round, but otherwise good to go.

source/common/router/config_impl.h

source/common/router/config_impl.h

adisuissa (approved):
LGTM, thanks!

yurykats:
@zuercher please take a look, thanks!

zuercher (approved):

PR #24683, opened 2022-12-24
WiP util: add new class to represent optional strings with significantly less overhead compared to std::string

jmarantz:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/common/utility.h
```diff
@@ -768,13 +769,168 @@ class InlineStorage : public NonCopyable {
   }
 };
 
+// Utility functions and stateful representation to efficiently encode and
```
jmarantz, line 772: Note to reviewers: this code was moved from its prior location in `Stats::SymbolTable::Encoding`, to its own standalone helper class NumericEncoding. The implementations are inlined rather than in a C++ file to avoid a performance degradation measured in test/common/stats/symbol_table_speed_test.cc. Previously, with Stats::SymbolTable::Encoding's implementation visible to the compiler at its call-site, the compiler can optimize better than if it only sees the API in the header.

source/common/stats/symbol_table.cc
```diff
@@ -14,13 +14,6 @@
 namespace Envoy {
 namespace Stats {
 
-// Masks used for variable-length encoding of arbitrary-sized integers into a
```
jmarantz, line 17: The changes in this file are to move the code for SymbolTable::Encoding into its own standalone class in source/common/common/utility.h so the encoding mechanism can be shared by TinyString. There is no functional change to the symbol table implementation; just renaming/moving the helper class.

test/common/common/utility_test.cc
```diff
@@ -1095,6 +1095,60 @@ TEST(InlineStorageTest, InlineString) {
   EXPECT_EQ("Hello, world!", hello->toString());
 }
 
+TEST(TinyStringTest, Empty) {
```
jmarantz, line 1098: If we think this approach makes sense and it is worth it, unit-tests for NumericEncoding need to be added to this PR.

jmarantz (commented):
@yurykats and @yanavlasov  : I'd like your opinion about this approach.

The main benefit of this PR is that it could shrink the footprint of an `absl::optional<std::string>` from 32 bytes (aligned 24 + 1 bytes) to 8 bytes (single pointer), in cases where the optional string does not need to be incrementally built, and it represents an optional string-setting which is often not provided.

test/common/common/utility_test.cc
yurykats, line 1098: I think a sizeof test that demonstrates the win over `absl::optional<std::string>` would be beneficial, similar to the one you have in PR #24661

source/common/common/utility.h
```diff
@@ -808,6 +964,60 @@ class InlineString : public InlineStorage {
...
+  }
+
+  void operator=(TinyString&& src) {
+    if (&src != this) {
+      data_ = std::move(src.data_);
+    }
+  }
+
+  /**
+   * Caller must ensure hasValue() is true.
```
yurykats, line 996: Could this mention what would happen otherwise?
Related: should toStringView ASSERT nullptr?

yurykats:
> @yurykats and @yanavlasov : I'd like your opinion about this approach.
> 
> The main benefit of this PR is that it could shrink the footprint of an `absl::optional<std::string>` from 32 bytes (aligned 24 + 1 bytes) to 8 bytes (single pointer), in cases where the optional string does not need to be incrementally built, and it represents an optional string-setting which is often not provided.

Yes, I would love to have this for config data structures!

test/common/common/utility_test.cc
jmarantz: makes sense.

source/common/common/utility.h
```diff
@@ -808,6 +964,60 @@ class InlineString : public InlineStorage {
   char data_[];
 };
 
+class TinyString {
```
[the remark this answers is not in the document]
jmarantz: this also makes sense.

source/common/common/utility.h
```diff
@@ -808,6 +964,60 @@ class InlineString : public InlineStorage {
...
+      data_ = std::move(src.data_);
+    }
+  }
+
+  /**
+   * Caller must ensure hasValue() is true.
+   *
+   * @return a std::string copy of the TinyString
+   */
+  std::string toString() const { return std::string(toStringView()); }
```
yanavlasov, line 1000: Is this method used (or expected to be used) outside of new tests in this PR? It hides the cost of allocation and copy. Maybe a better approach is not to provide it at all, and make the called use the `toStringView` to construct std::string if they really need it, making the cost of doing so more explicit.

yanavlasov (commented):
/wait

source/common/common/utility.h
jmarantz: that's OK with me also, but to me, it seems obvious that a function with signature `std::string toString() const` must do a copy.

Issue #23007, opened 2022-09-07
OAuth2 filter POST request with expired token


Alexcei88:
I am using [OAuth2 filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/oauth2_filter) to protect my backend. I faced with following unpleasant issue.

1. User opens my site
2. User authorizes and gets JWT token for 5 minutes.
3. User works on site for 5 minutes. 
4. Token is expired and user sends POST http method instead of GET method(any save operation on my site)
5. Because the token is expired, Envoy redirects user to IDP, user gets new JWT token, and after that instead of HTTP POST method my backend is recieved HTTP GET method, but initially my site had sent HTTP POST method.

Is there a best way to resolve such behaviour?

wbpcode:
cc @snowp

Alexcei88:
I think when we recieved callback from IDP, we need to send source request further without redirect to browser.

Alexcei88:
Could you add the tag "help wanted" please

Alexcei88:
Could you add the tag "help wanted" please?

vikaschoudhary16:
Why this is closed given that linked PR is still open?

PR #24684, opened 2022-12-25
Oauth2 filter: refresh token support

Alexcei88:
Commit Message:
Using a refresh token for updating the access token if that is provided by identity provider server
Additional Description:

A refresh token is a special token that is used to obtain additional access tokens. This allows you to have short-lived access tokens without having to collect credentials every time one expires.  If a refresh token is provided by oauth endpoint then after that the access token is expired the filter gets a new access token using by a refresh token without having to ask the user to log in again.
I see these changes to resolve for [this ](https://github.com/envoyproxy/envoy/issues/23007) problem

For backward compatilibility the special flag `use_refresh_token `was added to configuration. By default it is false, so the refresh token is not used. Its old behaviour. To turn on using refresh token it's needed to manually add `use_refresh_token ` to configuration of the filter.

To decrease duplication of code in the filter some of the pieces of code were extracted to separate functions in the filter. Therefore it looks like a lot of changes but it's not. I just refactored code to do it beatiful.

Also I added additional statistic counters for filter about using refresh token.

Risk Level:
Medium 

Testing:

Need to test oauth2 filter

Docs Changes:

Need to add description about new flag in the configuration

Release Notes:

Need to mention about supporting of refresh token in the filter


KBaichoo:
/assign @derekargueta 

Can you take a look as oauth2 filter owner? Thank you.

/wait

For CI fixes

Alexcei88:
##[error]No space left on device

I have already catched this exception two times today. How to restart CI?

Alexcei88:
It looks like CI is fixed

KBaichoo:
/assign @snowp 

As other oauth owner.

I think @derekargueta might not be active.

Alexcei88:
Hi @snowp , thank you for taking a look.
I would say, it is impossible. If we reduce refactoring then we will duplicate functionality. Then there will be more changes than now. 

I understand that it’s hard to see in commits, but I didn’t touch the old functionality at all. I just added a new one, but it is a quite large. New branch of logic was added to filter.

But I am sure for changes, because it already works on my production environment.

One moment I need to note, which I don't like in my implementation there is no checking a refresh token livespan. So sometimes we try to update access token using by expired refresh token. In this case the attempt finish as failed and user will be redirect to log in again. In the future we can add validation of refresh token before attempting update access token by refresh token.

snowp:
Unfortunately it'll be hard to carve out the time to review this as it stands, I'd suggest first doing a PR that performs all the refactoring that this PR does to the existing code, and then once that has been reviewed do another one which contains the addition of the refresh token support. Does that make sense?

Alexcei88:
@snowp, Your suggestion sounds good. 
But I think I can't spend more time on this feature. 
I think I made a hard work to add this cool feature in the filter. 
If I will have time to  separate PR  I make it. But I am no sure because it will required additional time.

Alexcei88:
@snowp , I submitted otheer PR as you requested
https://github.com/envoyproxy/envoy/pull/25160

wbpcode:
@Alexcei88 could you merge main and resolve the conflict? Thanks.

snowp:
There seems to be no commits left in this PR, did something go wrong during the merge?

Alexcei88:
Thank you for your attention. Yes, I cleared all changes in the branch. I will soon commit significant changes in the branch. Wait me please.

Alexcei88:
ready for review

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -114,6 +114,9 @@ message OAuth2Config {
   // Defines how ``client_id`` and ``client_secret`` are sent in OAuth client to OAuth server requests.
   // RFC https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
   AuthType auth_type = 11 [(validate.rules).enum = {defined_only: true}];
+
+  // Attempt to update access token using by refresh token if that is provided by authorization_endpoint.
+  bool use_refresh_token = 12;
```
lizan, line 119: Should this default to true? If so use BoolValue and document that.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -509,6 +543,57 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
...
+    newCookies.append(value.c_str(), value.size());
+  }
+
+  request_headers_->addReferenceKey(Http::Headers::get().Cookie, newCookies.getStringView());
+  if (config_->forwardBearerToken() && !access_token_.empty()) {
+    setBearerToken(*request_headers_, access_token_);
+  }
+
+  // remember cookies for response(need for set cookie)
+  response_headers_to_add_.reset();
```
lizan, line 582: no need `.reset()`.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -509,6 +543,57 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
...
+  }
+
+  Http::HeaderString newCookies;
+  for (const auto& p : cookies) {
+    if (!newCookies.empty()) {
+      newCookies.append("; ", 2);
+    }
+    std::string value = absl::StrCat(p.first, "=", p.second);
+    newCookies.append(value.c_str(), value.size());
+  }
```
lizan, line 574: `absl::StrJoin(cookies, "; ", absl::PairFormatter("="));`

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -509,6 +543,57 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
   config_->stats().oauth_success_.inc();
 }
 
+void OAuth2Filter::finishUpdateAccessTokenFlow() {
+  ASSERT(config_->useRefreshToken());
+  // update cookie header to send it on actual state to upstream
+  const CookieNames& cookie_names = config_->cookieNames();
+
+  absl::flat_hash_map<std::string, std::string> cookies =
```
lizan, line 551: I don't think you need set those cookies in request to response_headers_to_add again, just new cookies should be fine.

Alexcei88: You suggest to not save new cookies to the response_headers_to_add?
I tried it before, but it doesn't works. I didn't deep why it is. So I just decided to save cookies and set cookies through set-cookie headers in the response

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Alexcei88: The default is false. Because I don't consider to change old behaviour after update envoy. If user wants to use refresh token then manually set the flag to true.

source/extensions/filters/http/oauth2/filter.cc
lizan: I'm suggesting not set all `Cookie` in request header to `Set-Cookie` as this might cause a security issue that a new `Set-Cookie` can overwrite the original expiration, or remove some property like `Secure` or `HttpOnly`.

Alexcei88: `Set-Cookie` is applied only for specific cookies, related with OAuth2: id Token, RefreshToken, AccessToken, OAuth Hmac, OAuth Expires. Other cookies which recieved in the request I resend in initial state. I cannot add additional header `Cookie` in the request. As I know, only one header `Cookie` in the request is allowed. Looks like I don't catch your suggestion, I am sorry. 
Could you explain again please what I need to do?

jmarantz:
@Alexcei88 please do not force-push in Envoy pull requests.

See https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md

> Once your PR is under review, please do not rebase it. If you rebase, you will need to force push to github, and github's user interface will force your reviewer to review the PR from scratch rather than simply look at your latest changes.  ...

Alexcei88:
Could anyone have a chance to look?

snowp:
Yeah I will take a look sometime today, sorry for all the delay here

source/extensions/filters/http/oauth2/filter.cc
snowp: My understanding (and maybe @lizan can clarify if I'm misrepresenting what he's saying) is that you don't need to include all the cookies that are present on the request headers in the response headers, only the ones that you are trying to update. That way you are not resetting more cookies than necessary

test/extensions/filters/http/oauth2/filter_test.cc
```diff
@@ -1237,6 +1364,327 @@ TEST_F(OAuth2Test, OAuthBearerTokenFlowFromQueryParameters) {
             filter_->decodeHeaders(request_headers, false));
 }
 
+TEST_F(OAuth2Test, OAuthTestFullFlowWithUseRefreshToken) {
+
```
snowp, line 1368: remove this newline

test/extensions/filters/http/oauth2/filter_test.cc
```diff
@@ -1237,6 +1364,327 @@ TEST_F(OAuth2Test, OAuthBearerTokenFlowFromQueryParameters) {
...
+  // Deliberately fail the HMAC validation check.
+  EXPECT_CALL(*validator_, setParams(_, _));
+  EXPECT_CALL(*validator_, isValid()).WillOnce(Return(false));
+
+  EXPECT_CALL(*oauth_client_, asyncGetAccessToken("123", TEST_CLIENT_ID, "asdf_client_secret_fdsa",
+                                                  "https://traffic.example.com" + TEST_CALLBACK,
+                                                  AuthType::UrlEncodedBody));
+
+  // Invoke the callback logic. As a side effect, state_ will be populated.
```
snowp, line 1425: Since we don't check state_ i don't see much reason mentioning it

test/extensions/filters/http/oauth2/filter_test.cc
```diff
@@ -1237,6 +1364,327 @@ TEST_F(OAuth2Test, OAuthBearerTokenFlowFromQueryParameters) {
             filter_->decodeHeaders(request_headers, false));
 }
 
+TEST_F(OAuth2Test, OAuthTestFullFlowWithUseRefreshToken) {
```
snowp, line 1367: Maybe add a description to this test talking about what this tests - seems like it first does one authorization that doesn't use the refresh token followed by one that does?

test/extensions/filters/http/oauth2/filter_test.cc
```diff
@@ -1237,6 +1364,327 @@ TEST_F(OAuth2Test, OAuthBearerTokenFlowFromQueryParameters) {
...
+}
+
+TEST_F(OAuth2Test, OAuthTestSetCookiesAfterUpdateAccessTokenWithBasicAuth) {
+
+  init(getConfig(true /* forward_bearer_token */, true /* use_refresh_token */,
+                 ::envoy::extensions::filters::http::oauth2::v3::OAuth2Config_AuthType::
+                     OAuth2Config_AuthType_BASIC_AUTH
+                 /* authType */));
+
+  // the third request to the oauth filter with URI parameters.
```
snowp, line 1645: This doesn't seem like the third request, update?

snowp (dismissed):
Thanks, the overall flow seems right to me, but this needs a bit of work otherwise

I think this would benefit from having a test in oauth_integration_test that verifies that we correctly pause filter iteration while waiting for a response for the refresh request. There are also a few branches that need test coverage per the coverage report (for example in https://storage.googleapis.com/envoy-pr/e9fdd37/coverage/source/extensions/filters/http/oauth2/filter.cc.gcov.html )

source/extensions/filters/http/oauth2/oauth_client.cc
```diff
@@ -78,15 +117,27 @@ void OAuth2ClientImpl::onSuccess(const Http::AsyncClient::Request&,
...
-    parent_->sendUnauthorizedResponse();
+    switch (oldState) {
+    case OAuthState::PendingAccessToken:
+      parent_->sendUnauthorizedResponse();
+      break;
+    case OAuthState::PendingAccessTokenByRefreshToken:
+      parent_->onUpdateAccessTokenFailure();
+      break;
+    default:
+      throw EnvoyException("Malformed oauth client state");
```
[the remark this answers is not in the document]
Alexcei88: NetworkErrorDoubleCallStateInvalid test covers this exception

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -509,6 +543,48 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
...
+    if (id_token_ != EMPTY_STRING) {
+      cookies.insert_or_assign(cookie_names.id_token_, id_token_);
+    }
+    if (refresh_token_ != EMPTY_STRING) {
+      cookies.insert_or_assign(cookie_names.refresh_token_, refresh_token_);
+    }
+  }
+
+  std::string newCookies(absl::StrJoin(cookies, "; ", absl::PairFormatter("=")));
+  request_headers_->addReferenceKey(Http::Headers::get().Cookie, newCookies);
```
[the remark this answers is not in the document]
Alexcei88: Yep, we need to update BearerToken, OauthHMAC, OauthExpires  to send them in the actual state to the upstream.

source/extensions/filters/http/oauth2/oauth_client.cc
```diff
@@ -116,14 +167,35 @@ void OAuth2ClientImpl::onSuccess(const Http::AsyncClient::Request&,
...
-  parent_->onGetAccessTokenSuccess(access_token, id_token, refresh_token, expires_in);
+  switch (oldState) {
+  case OAuthState::PendingAccessToken:
+    parent_->onGetAccessTokenSuccess(access_token, id_token, refresh_token, expires_in);
+    break;
+  case OAuthState::PendingAccessTokenByRefreshToken:
+    parent_->onUpdateAccessTokenSuccess(access_token, id_token, refresh_token, expires_in);
+    break;
+  default:
+    throw EnvoyException("Malformed oauth client state");
```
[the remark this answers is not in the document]
Alexcei88: This exception is not covered by any tests

Alexcei88:
@jmarantz thank you for information. But I use a force push only for satisfying DCO requirement if I forget to sign my  commit

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -123,6 +123,9 @@ message OAuth2Config {
...
```
dio, line 128: Probably update this description to something like: 

"Set to true to enable automatic access token refresh using the associated refresh token, provided that the OAuth server supports this feature."

changelogs/current.yaml
```diff
@@ -156,6 +156,61 @@ new_features:
 - area: oauth filter
   change: |
     extended :ref:`cookie_names <envoy_v3_api_field_extensions.filters.http.oauth2.v3.OAuth2Credentials.cookie_names>` to allow overriding (default) cookie names (``IdToken``, ``RefreshToken``) set by the filter.
+- area: tracing
```
dio, line 159: This is probably a bad merge (?).

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -354,6 +369,15 @@ Http::FilterHeadersStatus OAuth2Filter::decodeHeaders(Http::RequestHeaderMap& he
   return Http::FilterHeadersStatus::StopAllIterationAndBuffer;
 }
 
+Http::FilterHeadersStatus OAuth2Filter::encodeHeaders(Http::ResponseHeaderMap& headers, bool) {
+  if (response_headers_to_add_) {
+    Http::HeaderMapImpl::copyFrom(headers, *response_headers_to_add_);
+    response_headers_to_add_.reset();
+  }
```
dio, line 376: Sorry. Could you add a comment on why we need this condition? Thank you.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -488,6 +512,15 @@ void OAuth2Filter::onGetAccessTokenSuccess(const std::string& access_code,
   finishGetAccessTokenFlow();
 }
 
+void OAuth2Filter::onUpdateAccessTokenSuccess(const std::string& access_code,
```
dio, line 515: Probably `onRefreshAccessTokenSuccess` is a better name (?).

source/extensions/filters/http/oauth2/oauth_client.cc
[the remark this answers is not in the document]
dio: Probably we can explain more about when this case will happen in a real case (?). Also when this crashes envoy, will that be because of the response from the server or upstream?

dio (commented):
Thank you @Alexcei88. Flushing some comments. Also probably we need an example of using this filter under the: `examples` directory (this will be in a separate PR though, seems like you have it on one of your branches: `oauth2_refresh_token`), so it is easier for us to validate new features. It will be very useful, I believe.

source/extensions/filters/http/oauth2/filter.cc
Alexcei88: I agree, your name is better

source/extensions/filters/http/oauth2/filter.cc
Alexcei88: Because response_headers_to_add_ is not null only in the case of refreshing token. In other cases it is null

Alexcei88:
> Thank you @Alexcei88. Flushing some comments. Also probably we need an example of using this filter under the: `examples` directory (this will be in a separate PR though, seems like you have it on one of your branches: `oauth2_refresh_token`), so it is easier for us to validate new features. It will be very useful, I believe.

I hope I will have time and mae an examle

source/extensions/filters/http/oauth2/oauth_client.cc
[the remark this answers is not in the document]
Alexcei88: No, the exception is thrown when the programmer will write a bad code. It doesn't depend from answer of oauth server.

changelogs/current.yaml
Alexcei88: I have updated the file

Alexcei88:
> Thanks, the overall flow seems right to me, but this needs a bit of work otherwise
> 
> I think this would benefit from having a test in oauth_integration_test that verifies that we correctly pause filter iteration while waiting for a response for the refresh request. There are also a few branches that need test coverage per the coverage report (for example in https://storage.googleapis.com/envoy-pr/e9fdd37/coverage/source/extensions/filters/http/oauth2/filter.cc.gcov.html )

Sorry, but I don't have time understand how to write integration test. If anyone can help me, I will appreciate helping. I wrote enough unit tests.

yanavlasov:
Hi @Alexcei88 The integration test suite already exists. https://github.com/envoyproxy/envoy/blob/main/test/extensions/filters/http/oauth2/oauth_integration_test.cc It allows end to end testin of the entire system. You can look at the `doAuthenticationFlow` function for orchestrating the authentication flow.

RyanTheOptimist:
/wait
Looks like this is still waiting on Integration Tests

Alexcei88:
> /wait Looks like this is still waiting on Integration Tests

It's ready

wbpcode:
I will take a look today or tomorrow. sorry for the delay.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -190,6 +191,10 @@ void OAuth2CookieValidator::setParams(const Http::RequestHeaderMap& headers,
   secret_.assign(secret.begin(), secret.end());
 }
 
+bool OAuth2CookieValidator::canUpdateTokenByRefreshToken() const {
+  return token_ != EMPTY_STRING && refresh_token_ != EMPTY_STRING;
```
wbpcode, line 195: `(!token_.empty() && !refresh_token_.empty())`

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -507,6 +540,52 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
...
+  request_headers_->addReferenceKey(Http::Headers::get().Cookie, new_cookies);
+  if (config_->forwardBearerToken() && !access_token_.empty()) {
+    setBearerToken(*request_headers_, access_token_);
+  }
+
+  // At this point we add set-cookie headers to the temporary object of the response
+  // The temporary response will populate the response from upstream in the encodeHeaders method
+  // Eventually set-cookie headers will be send to the user agent (browser)
+  response_headers_to_add_ = Http::ResponseHeaderMapImpl::create();
+  addResponseCookies(*response_headers_to_add_, getEncodedToken());
```
wbpcode, line 576: Why we need create a header map here? I think a simple bool is enough here to tell we should update the response headers?

wbpcode (changes requested):
Thanks for your great contribution. This is a big PR and I am not a oauth expert. So only some minor coments are added first. Seems it's good overall.

There are some historical comments that not be addressed completely but be marked as resolved.
So, I am afraid there are some important concerns that should be resolved but accutally not. And it's hard to iterate all these comments.

source/extensions/filters/http/oauth2/filter.cc
Alexcei88: We need to pass a map object to addResponseCookies. So, we create object. addResponseCookies method populate headers to the map

wbpcode: I mean why cannot we just call the addResponseCookies in the `encodeHeaders()` and use the accutally response headers as the parameter. Then we needn't create a unnecessary temporary object here. did I miss something?

wbpcode:
Thanks for your contribution. And please don't use force push, esp for a big PR like this. It make it's very hard to do a gradually review.

Alexcei88:
@wbpcode all is ready

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -507,6 +540,48 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
   config_->stats().oauth_success_.inc();
 }
 
+void OAuth2Filter::finishUpdateAccessTokenFlow() {
```
wbpcode, line 543: nit: use `finishRefreshAccessTokenFlow` because the `onRefreshAccessTokenSuccess`. I think it would be better to unify these name to use `Update` or `Refresh`.

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -507,6 +540,48 @@ void OAuth2Filter::finishGetAccessTokenFlow() {
...
+  }
+
+  wasRefreshTokenFlow_ = true;
+
+  config_->stats().oauth_refreshtoken_success_.inc();
+  config_->stats().oauth_success_.inc();
+  decoder_callbacks_->continueDecoding();
+}
+
+void OAuth2Filter::onUpdateAccessTokenFailure() {
```
wbpcode, line 579: nit: `onRefreshAccessTokenFailure`

source/extensions/filters/http/oauth2/filter.h
```diff
@@ -178,6 +181,7 @@ class FilterConfig {
   const std::vector<Http::HeaderUtility::HeaderData> pass_through_header_matchers_;
   const CookieNames cookie_names_;
   const AuthType auth_type_;
+  const bool use_refresh_token_ : 1;
```
wbpcode, line 184: nit: `const bool use_refresh_token_{};`

wbpcode (commented):
LGTM overall to these code. Thanks for your contributions. And only some nit comments to the naming.

And friendly ping cc @markdroth for API review.

wbpcode:
Seem @markdroth is too busy recently, re-assign this to @adisuissa for api review.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
adisuissa: I see your point of not breaking current behavior, and I agree it should be false by default at the moment.
It may be better to use a BoolValue type, as in the future it may make sense to convert the default value to true (see https://github.com/envoyproxy/envoy/blob/main/api/STYLE.md#api-style-guidelines re "wrapped scalar types").

adisuissa (commented):
Thanks!
Left a minor comment re API.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Alexcei88: I don't want to make this changes, because I don't have more time to this PR. I killed very much time to fix minor changes. Sorry! I think it is easy to setup exactly refresh token support.

Alexcei88:
What are we waiting yet?

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
adisuissa: Can you please elaborate what you mean by "it is easy to exactly set a refresh token support"?

wbpcode:
I think approvement from @envoyproxy/api-shepherds is necessary. cc @adisuissa any concern?

/wait-any

adisuissa:
> I think approvement from @envoyproxy/api-shepherds is necessary. cc @adisuissa any concern?
> 
> /wait-any

Might be hard to track in the github UI, but [this](https://github.com/envoyproxy/envoy/pull/24684#discussion_r1259133344) was reopened.

/wait-any

source/extensions/filters/http/oauth2/filter.cc
Alexcei88: @snowp , where do you see I touch other cookies except only the ones that are related with oauth2? Take a look please `finishRefreshAccessTokenFlow` method.
I only update `cookies oauth_hmac, oauth_expires, bearer_token, id_token, refresh_token` cookies.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Alexcei88: I mean user can set true or false clearly on configuration if user knows whether or not the oauth server supports refresh token.

adisuissa: > The default is false. Because. I don't consider to change automatically old behaviour after update envoy. If user wants to use refresh token then manually set the flag to true. In the future we can change default behaviour

If in the future we want to be able to update the default value of this field, it should be a wrapped type, rather than a primitive.
/wait-any

derekargueta (commented):
some minor doc suggestions

phlax:
@adisuissa @wbpcode i think this is still waiting on signoff

/wait-any

GuillermoGarciaF:
Hi, could you confirm if the functionality will be merged and made available?
It is an important feature for many of us. Thank you!

Alexcei88:
Hi @GuillermoGarciaF , No the code is not merged to main.

I have been supporting the PR for half of year.  It is very expensive for time. 
The review of PR is very slow.  Mainly due to some small changes. 
This functionality is working excelllent in production server for half of year where I work. But we cannot update Envoy without this PR. And I am interseting to finish this PR.

But periodically the new changes from main branch force me to rewrite new tests for oauth2 filter. 
Last I faced this problem again. And I don't have working environment to make fixes, because I am coding mainly on C#.
So I decided to close MR request and I will return when I will have more time.  
Sorry guys, but I will back sure a little bit later.

roeap:
Thanks for your hard work @Alexcei88!

hoping that this eventually makes it to main, as it would be a great feature for our deployments as well.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Alexcei88: I have changed type of veriable to BoolValue as you requested

wbpcode:
cc @Alexcei88 Thanks very much for your time and patience. And sorry for all bad experiences in the review processing.

I think the last thing need to be addressed is the API after my last review.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -123,6 +125,9 @@ message OAuth2Config {
   // Defines how ``client_id`` and ``client_secret`` are sent in OAuth client to OAuth server requests.
   // RFC https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
   AuthType auth_type = 11 [(validate.rules).enum = {defined_only: true}];
+
+  // Set to true to enable automatic access token refresh using the associated refresh token, provided that the OAuth
+  google.protobuf.BoolValue use_refresh_token = 12;
```
wbpcode, line 130: nit: add the end `.` after this sentence. This will be used to generate docs, so the formatter is also important.

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -123,6 +125,9 @@ message OAuth2Config {
   // Defines how ``client_id`` and ``client_secret`` are sent in OAuth client to OAuth server requests.
   // RFC https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
   AuthType auth_type = 11 [(validate.rules).enum = {defined_only: true}];
+
+  // Set to true to enable automatic access token refresh using the associated refresh token, provided that the OAuth.
```
adisuissa, line 129:
```suggestion
  // If set to true, allows automatic access token refresh using the associated refresh token (see 
  // `RFC 6749 section 6<https://datatracker.ietf.org/doc/html/rfc6749#section-6>`_), provided that the OAuth server supports that.
  // Default value is false.
```

adisuissa (commented):
Thanks!
Left a small suggestion on the API description, but otherwise LGTM.
Will let @derekargueta have a final say on the implementation, as code-owner.

adisuissa (commented):
/lgtm api

Alexcei88:
@adisuissa , your last suggestion broke the [build](https://dev.azure.com/cncf/envoy/_build/results?buildId=150665&view=logs&j=9fd58ae0-0c50-56b1-f517-c2b9afda9c9b&t=c8fc762c-55bb-57d4-4cab-8ec527e7a423).

> ERROR: Unknown target name: "rfc 6749 section 6<[https://datatracker.ietf.org/doc/html/rfc6749#section-6>".](https://datatracker.ietf.org/doc/html/rfc6749#section-6%3E%22.)


Could you suggest how to fix it?

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
```diff
@@ -123,6 +125,11 @@ message OAuth2Config {
   // Defines how ``client_id`` and ``client_secret`` are sent in OAuth client to OAuth server requests.
   // RFC https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1
   AuthType auth_type = 11 [(validate.rules).enum = {defined_only: true}];
+
+  // If set to true, allows automatic access token refresh using the associated refresh token (see 
+  // `RFC 6749 section 6<https://datatracker.ietf.org/doc/html/rfc6749#section-6>`_), provided that the OAuth server supports that.
```
zuercher, line 130: I believe this needs a space after `section 6` and before the angle bracket.

See https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages

wbpcode:
friendly ping @derekargueta

wbpcode:
Seems @derekargueta is inactive recently. I am inclined to merge this after the format problem is resolved. cc @adisuissa

api/envoy/extensions/filters/http/oauth2/v3/oauth.proto
Alexcei88: @zuercher , Thank you, I will try

adisuissa:
> Seems @derekargueta is inactive recently. I am inclined to merge this after the format problem is resolved. cc @adisuissa

@mattklein123 is the other code-owner, assigning Matt to give final review on this.
/assign @mattklein123

mattklein123 (approved):
Skimmed at a high level. I don't have the time/knowledge to understand the security implications of this.

wbpcode (approved):

PR #23684, opened 2022-10-26
router: set stream_info_'s request headers of upstream request

tony612:
Commit Message: router: set stream_info_'s request headers of upstream request
Additional Description:
Now custom_tags are applied also for upstream spans due to https://github.com/envoyproxy/envoy/pull/22987/files#diff-36ebe2d9da39e88ceeccb197a9ff0b8a1dc97a5f72c8597c40d27f61a9ead00f.
But requestHeaders of stream_info is not set for UpstreamRequest and custom_tags won't work. So this PR set it to make it work.

Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:

PR #24685, opened 2022-12-26
route: set route_ of stream_info_ in UpstreamRequest

tony612:
Signed-off-by: Bing Han <h.bing612@gmail.com>

Commit Message: route: set route_ of stream_info_ in UpstreamRequest
Additional Description:

Now custom_tags are applied also for upstream spans due to https://github.com/envoyproxy/envoy/pull/22987/files#diff-36ebe2d9da39e88ceeccb197a9ff0b8a1dc97a5f72c8597c40d27f61a9ead00f.
But route_ of stream_info is not set for UpstreamRequest and route metadata in custom_tags won't work. This is similar to https://github.com/envoyproxy/envoy/pull/23684.

And I checked other custom tags, I believe all of them work like cluster metadata except dynamic metadata(request kind). I'm not sure if we need to set `metadata_` using `parent` directly like `stream_info_.metadata_ = parent_.callbacks()->streamInfo().dynamicMetadata();`. Maybe we should leave it blank in initializing and let upstream filters do that.

Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:


source/common/router/upstream_request.cc
```diff
@@ -106,6 +106,8 @@ UpstreamRequest::UpstreamRequest(RouterFilterInterface& parent,
     }
   }
   stream_info_.setUpstreamInfo(std::make_shared<StreamInfo::UpstreamInfoImpl>());
+  // route_ can be used to set route kind of metadata in custom tags.
```
snowp, line 109: I don't think this comment is necessary, everything we're setting on the stream info is used for custom tags

snowp (changes requested):
Thanks for adding this! Do you think you could add a release note + unit test for this?

tony612:
@snowp Done. PTAL. Thanks!

tony612:
Resolved the conflict.

changelogs/current.yaml
```diff
@@ -134,6 +134,9 @@ bug_fixes:
 - area: tcp_proxy
   change: |
     When tunneling TCP over HTTP, mark the upstream connection as done reading when upstream trailers are read. This behavioral change can be temporarily reverted by setting runtime guard ``envoy.reloadable_features.finish_reading_on_decode_trailers`` to false.
+- area: router
+  change: |
+    fixed a bug that custom tags of the route metadata type are not set for upstream spans.
```
snowp, line 139: I would frame this as a new feature, I don't think this ever worked. I also would refer to it "upstream access logs" instead of spans

snowp (changes requested):
Thanks, one suggestion

changelogs/current.yaml
tony612: Do you mean dynamic/cluster metadata of access logs? I guess this PR is not related to access logs? I set the `route_` in `stream_info_`, and I think it only enables us to add custom tags of route metadata to the upstream spans after [this change](https://github.com/envoyproxy/envoy/pull/22987/files#diff-36ebe2d9da39e88ceeccb197a9ff0b8a1dc97a5f72c8597c40d27f61a9ead00f).

Maybe you mean a *new* `ROUTE_METADATA` in access log? I found there's no `ROUTE_METADATA` in access log now. If we need to support `ROUTE_METADATA` in access log, I think we need to change substitution_formatter.cc. Then we may need a new PR?

snowp: The main/original usage of StreamInfo is to provide information through access logs, e.g. via %ROUTE_NAME% and other things that would read from the route handle on StreamInfo, so I was thinking this was a fix for that. Sounds like its also a fix for trace spans so lets mention both?

tony612: Got it. Done.

snowp (approved):
Thanks!

tony612:
Resolved the conflicts.

snowp:
Seems like this will need to merge main to fix the change logs

Maybe @wbpcode can help with a second review?

tony612:
Resolved the conflict. PTAL, @wbpcode @snowp. Thanks!

wbpcode (approved):
LGTM. Thanks.

PR #24693, opened 2022-12-27
[http] fix http1.0 drain cause delay close.

pyrl247:
When we have the following situation:
 A --http1.0 request--> envoy --http1.1 request--> B
 A <--http1.0 response-- envoy <-- http1.1 response chunked-- B

Envoy response to A will not have content-length, the response will be framed by connection close.
When listener are draining, the drain timeout will cause connection delay close,and A will stop receiving response until connection close.
Since http1.0 is short-lived connections,maybe we don't need drain when use http10?
Signed-off-by: GuangYang <pyrl247@gmail.com>

Risk Level: Low
Testing:
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/conn_manager_impl.cc
+++ b/source/common/http/conn_manager_impl.cc
@@ -1460,6 +1460,7 @@ void ConnectionManagerImpl::ActiveStream::encodeHeaders(ResponseHeaderMap& heade
   // See if we want to drain/close the connection. Send the go away frame prior to encoding the
   // header block.
   if (connection_manager_.drain_state_ == DrainState::NotDraining &&
+      connection_manager_.codec_->protocol() == Protocol::Http11 &&
       (connection_manager_.drain_close_.drainClose() || drain_connection_due_to_overload)) {
 
     // This doesn't really do anything for HTTP/1.1 other then give the connection another boost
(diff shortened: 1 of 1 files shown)


alyssawilk:
Thanks for looking into this.  I thought this was already handled though - if the response is HTTP/1.1 framed by connection close, we should be force-closing on end stream here:

https://github.com/envoyproxy/envoy/blob/main/source/common/http/conn_manager_impl.cc#L257

I don't think that we need to change code other than that.  If you think that isn't sufficient could you write an integration test which exhibits the lag which we can use as a regression test?
/wait

pyrl247:
Oh,It seems already solve the problem.My local code was several versions behind main;I didn't notice this change before.Thanks!!

pyrl247:
@alyssawilk Sorry to bother .
 https://github.com/envoyproxy/envoy/blob/main/source/common/http/conn_manager_impl.cc#L257  .
This maybe not handled the situation when drain_state_ == DrainState::Draining.
https://github.com/envoyproxy/envoy/blob/main/source/common/http/conn_manager_impl.cc#L1312

alyssawilk:
could you just add an e2e test which shows the connection not delay closing in the circumstances you describe?  Once you have that and can show the test fails prior to a change I'd be happy to review.
thanks!
/wait

PR #24695, opened 2022-12-27
Make ProtocolOptionsConfigImpl const correct

yanavlasov:
Commit Message:
this makes it easier to reason about immutability of shared config objects
Risk Level: Low
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/extensions/upstreams/http/config.h
+++ b/source/extensions/upstreams/http/config.h
@@ -49,15 +49,15 @@ public:
   const absl::optional<envoy::config::core::v3::UpstreamHttpProtocolOptions>
       upstream_http_protocol_options_;
 
-  bool use_downstream_protocol_{};
-  bool use_http2_{};
-  bool use_http3_{};
-  bool use_alpn_{};
   using FiltersList = Protobuf::RepeatedPtrField<
       envoy::extensions::filters::network::http_connection_manager::v3::HttpFilter>;
-  FiltersList http_filters_;
-  absl::optional<envoy::config::core::v3::AlternateProtocolsCacheOptions>
+  const FiltersList http_filters_;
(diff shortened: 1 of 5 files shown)


source/extensions/upstreams/http/config.h
```diff
@@ -49,15 +49,15 @@ class ProtocolOptionsConfigImpl : public Upstream::ProtocolOptionsConfig {
...
   using FiltersList = Protobuf::RepeatedPtrField<
       envoy::extensions::filters::network::http_connection_manager::v3::HttpFilter>;
-  FiltersList http_filters_;
-  absl::optional<envoy::config::core::v3::AlternateProtocolsCacheOptions>
+  const FiltersList http_filters_;
+  const absl::optional<const envoy::config::core::v3::AlternateProtocolsCacheOptions>
       alternate_protocol_cache_options_;
+  const bool use_downstream_protocol_{};
+  const bool use_http2_{};
+  const bool use_http3_{};
```
snowp, line 59: It feels like this should be a single enum and not two bools, since I don't think (true, true) is a valid state? This might also make the protocol resolution logic since there would be one function producing the result instead of two

snowp (approved):
Change LGTM, this does seem easier to understand. One comment about a possible improvement

source/extensions/upstreams/http/config.h
yanavlasov: Both `http2 == true AND http3 == true` is a valid state for clusters with the `auto` protocol configuration, which chooses protocol based on upstream's ALPN header. Since in this case both protocols are possible based on what the upstream host is sending to Envoy.

snowp: Got it, that makes sense. I guess a better name might be `can_use_http2/3` but doesn't matter that much

PR #24699, opened 2022-12-28
Explicitly set  no_full_scan flag to true for a TLS test case which requires it

yanjunxiang-google:
https://github.com/envoyproxy/envoy/pull/24483 added below test 
TEST_P(SslSocketTest, MultiCertWithFullScanDisabledOnSniMismatch) {

and a runtime flag: 

RUNTIME_GUARD(envoy_reloadable_features_no_full_scan_certs_on_sni_mismatch);

However, the test is specific to the case when the runtime flag is true.  If the flag is false, this test does not work and crashing.

Explicitly set no_full_scan flag to true for this TLS test case to get test cases pass regardless the default flag setting.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/test/extensions/transport_sockets/tls/ssl_socket_test.cc
+++ b/test/extensions/transport_sockets/tls/ssl_socket_test.cc
@@ -1558,6 +1558,12 @@ TEST_P(SslSocketTest, MultiCertPickRSAOnSniMatch) {
 
 // On SNI mismatch, if full scan is disabled, validate that the first cert is used.
 TEST_P(SslSocketTest, MultiCertWithFullScanDisabledOnSniMismatch) {
+  // This test is specific for no full scan case, i.e, below flag is true case.
+  // If the flag is false, the test needs to be skipped.
+  if (!Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch")) {
+    return;
+  }
   const std::string client_ctx_yaml = absl::StrCat(R"EOF(
     sni: "nomatch.example.com"
     common_tls_context:
--- a/test/extensions/transport_sockets/tls/ssl_socket_test.cc
+++ b/test/extensions/transport_sockets/tls/ssl_socket_test.cc
@@ -1559,11 +1559,9 @@ TEST_P(SslSocketTest, MultiCertPickRSAOnSniMatch) {
 // On SNI mismatch, if full scan is disabled, validate that the first cert is used.
 TEST_P(SslSocketTest, MultiCertWithFullScanDisabledOnSniMismatch) {
   // This test is specific for no full scan case, i.e, below flag is true case.
-  // If the flag is false, the test needs to be skipped.
-  if (!Runtime::runtimeFeatureEnabled(
-          "envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch")) {
-    return;
-  }
+  TestScopedRuntime scoped_runtime;
+  scoped_runtime.mergeValues(
+      {{"envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch", "true"}});
   const std::string client_ctx_yaml = absl::StrCat(R"EOF(
     sni: "nomatch.example.com"
     common_tls_context:


LuyaoZhong (dismissed):
LGTM, thanks.

test/extensions/transport_sockets/tls/ssl_socket_test.cc
```diff
@@ -1558,6 +1558,12 @@ TEST_P(SslSocketTest, MultiCertPickRSAOnSniMatch) {
 
 // On SNI mismatch, if full scan is disabled, validate that the first cert is used.
 TEST_P(SslSocketTest, MultiCertWithFullScanDisabledOnSniMismatch) {
+  // This test is specific for no full scan case, i.e, below flag is true case.
+  // If the flag is false, the test needs to be skipped.
+  if (!Runtime::runtimeFeatureEnabled(
+          "envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch")) {
```
soulxu, line 1564: Emm... a question, how it can be false, without explicitly override the runtime flag value, like
```
TestScopedRuntime scoped_runtime;
  scoped_runtime.mergeValues(
      {{"envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch", "false"}});
```

yanjunxiang-google: We simply changed the runtime flag in source/common/runtime/runtime_features.cc to be false by 
FALSE_RUNTIME_GUARD(envoy_reloadable_features_no_full_scan_certs_on_sni_mismatch), then started a blaze build of ssl_socket_test.cc, and this test failed.

test/extensions/transport_sockets/tls/ssl_socket_test.cc
```diff
@@ -1558,6 +1558,12 @@ TEST_P(SslSocketTest, MultiCertPickRSAOnSniMatch) {
 
 // On SNI mismatch, if full scan is disabled, validate that the first cert is used.
 TEST_P(SslSocketTest, MultiCertWithFullScanDisabledOnSniMismatch) {
+  // This test is specific for no full scan case, i.e, below flag is true case.
+  // If the flag is false, the test needs to be skipped.
+  if (!Runtime::runtimeFeatureEnabled(
```
yanavlasov, line 1563: If the test relies on the value of the runtime flag, it should explicitly set this value.

```
  TestScopedRuntime scoped_runtime;
  scoped_runtime.mergeValues(
      {{"envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch", "true"}});
```

yanavlasov (changes requested):
/wait

yanavlasov:
@LuyaoZhong please LGTM if this change is ok

LuyaoZhong:
> @LuyaoZhong please LGTM if this change is ok

The code looks good.
You need to change the PR description to align with your code change, since this change is overriding runtime flag for flag-based test case and does not disable the test case.

test/extensions/transport_sockets/tls/ssl_socket_test.cc
soulxu: emm...if you change `source/common/runtime/runtime_features.cc` directly, that means you have fork version envoy. It doesn't make sense change the upstream for your downstream code. This test also ensures the envoy default behavior. With this change, it loses this guarantee.

soulxu (changes requested):
better to describe why you want to change `source/common/runtime/runtime_features.cc` in downstream code, otherwise it doesn't make sense for changing upstream code for downstream change.

test/extensions/transport_sockets/tls/ssl_socket_test.cc
LuyaoZhong: Agree with @soulxu , if you use FALSE_RUNTIME_GUARD, the default behavior is changed.

yanjunxiang-google: Thanks for the comments. I was under the impression that our test code need to be passing even if we toggle the runtime flag.  If a test is written for a specific runtime flag state, we should either explicitly specify it, or skip the test otherwise.

yanavlasov:
One issue that confused me is whether the `MultiCertWithFullScanDisabledOnSniMismatch` test **requires** the `envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch == true`?

@LuyaoZhong or @soulxu  is this the case? If this test does not depend on the value of this flag then this PR should be reverted.

/wait-any

ggreenway:
> One issue that confused me is whether the `MultiCertWithFullScanDisabledOnSniMismatch` test **requires** the `envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch == true`?
> 
> @LuyaoZhong or @soulxu is this the case? If this test does not depend on the value of this flag then this PR should be reverted.

Yes, this test is specifically testing that case. The next test in the file is testing the opposite and explicitly sets the runtime flag to false. I think it is more-correct to explicitly set the runtime flag to true in this test, instead of relying on the default.

ggreenway (approved):

yanavlasov (approved):

soulxu:
> > One issue that confused me is whether the `MultiCertWithFullScanDisabledOnSniMismatch` test **requires** the `envoy.reloadable_features.no_full_scan_certs_on_sni_mismatch == true`?
> > @LuyaoZhong or @soulxu is this the case? If this test does not depend on the value of this flag then this PR should be reverted.
> 
> Yes, this test is specifically testing that case. The next test in the file is testing the opposite and explicitly sets the runtime flag to false. I think it is more-correct to explicitly set the runtime flag to true in this test, instead of relying on the default.

Yea, it depends on the value of the flag. I thought that test is for testing the default behavior of envoy with default value of the flag. Not sure we guarantee all the tests can be passed if we flip the value of flag. Just free to change a flag, like `http_strip_fragment_from_path_unsafe_if_disabled`, the test can't be passed.

PR #1124, opened 2017-06-15
config: subscription API interface and gRPC implementation.

htuch:
This PR provides a callback based Subscription API in the Envoy::Config namespace to allow modules
to receive resource updates via the v2 xDS APIs. It's a generalization of the existing
RestApiFetcher, which can be implemented in gRPC bidi, REST and via filesystem inotify watching.

In addition to the interface, the bidi gRPC implementation is provided and a demonstration of how it
works with EDS is provided in test/common/grpc/subscription_impl_test.cc.

After this PR it will be necessary to amend your `WORKSPACE` file with:

```

load("@envoy_api//bazel:repositories.bzl", "api_dependencies")
api_dependencies()
```

PR #24701, opened 2022-12-29
config: Make  gRPC xDS retry backoff configurable

### LINKED ISSUE #24236: Make xDS retry backoff configurable
*Title: Add configuration parameters for xDS retry backoff strategy*

**This issue is a revival of the same issue that became stale:** https://github.com/envoyproxy/envoy/issues/9332

_Description:_
The gRPC xDS implementation uses hardcoded retry durations for re-establishing the gRPC channel. This introduces a problem when a fast failover is required from one control-plane instance to another.
In our use case, we have multiple control-plane instances in _active-standby_ mode and require a fast failover when "active" fails and the "standby" picks up.

We plan to address the TODO to make the backoff parameters configurable:
```
// TODO(htuch): Make this configurable.
static constexpr uint32_t RetryInitialDelayMs = 500;
static constexpr uint32_t RetryMaxDelayMs = 30000; // Do not cross more than 30s
backoff_strategy_ =
        std::make_unique<JitteredBackOffStrategy>(RetryInitialDelayMs, RetryMaxDelayMs, random_);
```

_Relevant Links:_
https://github.com/envoyproxy/envoy/blob/v1.12.1/source/common/config/grpc_stream.h#L39

jaysonjdmello:
Signed-off-by: Jayson Dmello <jdmello@confluent.io>

Commit Message:
The gRPC xDS implementation uses hardcoded retry durations for re-establishing the gRPC channel. This change makes the backoff parameters configurable.

Additional Description:
Risk Level: Low
Testing: unit test
Docs Changes:
Release Notes: Added
Platform Specific Features:
[Optional Runtime guard:]
 Fixes #24236 
[Optional Deprecated:]


KBaichoo:
/assign @yanjunxiang-google 

Can you take a look?

source/common/config/utility.h
```diff
@@ -41,6 +41,9 @@ namespace Config {
 
 constexpr absl::string_view Wildcard = "*";
 
+static constexpr uint64_t RetryBaseIntervalMs = 500;
```
yanjunxiang-google, line 44: Are these global constants GRPC retry mechanism specific?  Need some comments here. Also is here the best place to define them?

test/common/config/utility_test.cc
```diff
@@ -293,6 +293,155 @@ TEST(UtilityTest, PrepareDnsRefreshStrategy) {
...
+                                                                                random);
+
+      // max_interval should be less than or equal test_base_interval * 10
+      EXPECT_EQ(false, dynamic_cast<JitteredExponentialBackOffStrategy*>(strategy.get())
+                           ->isOverTimeLimit(test_base_interval_ms * 10));
+      EXPECT_EQ(true, dynamic_cast<JitteredExponentialBackOffStrategy*>(strategy.get())
+                          ->isOverTimeLimit(test_base_interval_ms * 10 + 1));
+    }
+
+    // set max_interval > base_interval
```
yanjunxiang-google, line 423: Do you mean max_interval < base_interval here?

yanjunxiang-google:
LGTM in general. Just a few nit comments.

source/common/config/utility.h
jaysonjdmello: These are not intended to be specific to GRPC retry timers but rather the the default time for @prepareBackoffStrategy method in this file.

test/common/config/utility_test.cc
jaysonjdmello: will fix ,thanks a lot!

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -288,10 +288,15 @@ message GrpcService {
...
+  // If max interval is not provided, it will be set to ten times the base interval.
+  RetryPolicy retry_policy = 5;
+
   // Additional metadata to include in streams initiated to the GrpcService. This can be used for
   // scenarios in which additional ad hoc authorization headers (e.g. ``x-foo-bar: baz-key``) are to
   // be injected. For more information, including details on header value syntax, see the
   // documentation on :ref:`custom request headers
   // <config_http_conn_man_headers_custom_request_headers>`.
-  repeated HeaderValue initial_metadata = 5;
+  repeated HeaderValue initial_metadata = 6;
```
lizan, line 301: Don't renumber proto field, this breaks API backward compatibility.

lizan (changes requested):

KBaichoo:
/assign @adisuissa 

For xDS code review maintainer pass

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: corrected, thanks!

adisuissa (commented):
Thanks for your contribution.
Before going into the details, can you explain the benefit of using this over the [gRPC retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism?

/wait-any

jaysonjdmello:
The gRPC xDS implementation uses hardcoded retry durations for re-establishing the gRPC channel. This change aims to make the retry durations configurable via envoy config. The purpose is to still use the gRPC retries mechanism as-is

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -294,4 +294,9 @@ message GrpcService {
   // documentation on :ref:`custom request headers
   // <config_http_conn_man_headers_custom_request_headers>`.
   repeated HeaderValue initial_metadata = 5;
+
+  // Indicates the retry policy for re-establishing the gRPC channel
+  // This field is optional. If not provided, base interval is set to 500 ms, max interval set to 30 secs.
+  // If max interval is not provided, it will be set to ten times the base interval.
+  RetryPolicy retry_policy = 6;
```
adisuissa, line 301: The issue with adding this field here is that there are other places that use `GrpcService` (i.e., non-xDS) that at the moment do not parse this field.

I wonder if this should be a property of the underlying client, and then one just configures the retry policy for EnvoyGrpc, and uses the [gRPC retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism for GoogleGrpc. @htuch any opinions on this?

jaysonjdmello: Sorry, I am new to the project. 
Are you suggesting this is moved inside message EnvoyGrpc (https://github.com/envoyproxy/envoy/blob/bb7d941e7e245ebd89ca706d428d92f050d5e7b3/api/envoy/config/core/v3/grpc_service.proto#L32)?

adisuissa: I think that EnvoyGrpc and GoogleGrpc will have different mechanism to handle backoff.
Specifically for this PR, I think you need to have the retry config in `EnvoyGrpc`, and the actual code that handles retries in [AsyncClientImpl](https://github.com/envoyproxy/envoy/blob/c394382139010828c125a23eccb47a5172e2eab8/source/common/grpc/async_client_impl.h#L23). An added value of this is that other uses of `AsyncClientImpl` (say filters that use gRPC), will also have that mechanism implemented.

The GoogleGrpc will have a different retry mechanism that is supported by the library.

jaysonjdmello: Thanks a lot for the feedback.

 It looks like adding the backoff mechanism to `AsyncClientImpl` would require extensive design changes. The goal of https://github.com/envoyproxy/envoy/issues/24236 and this PR was to resolve the TODO of making the  backoff parameters configurable. 

I added the `BackoffStrategy` to `ApiConfigSource` for this iteration,  I can raise another issue for refactoring `EnvoyGrpc` retry mechanism if that works. 

Thanks again.

adisuissa: Thanks! Will it be possible to add the config into `EnvoyGrpc`. The motivation is to avoid the need to deprecate the change to the `ApiConfigSource` in the future?

/wait-any

jaysonjdmello: Sounds good, Moved the config into `EnvoyGrpc`.  Thanks

source/common/config/utility.h
```diff
@@ -531,6 +537,45 @@ class Utility {
...
   }
+
+  /**
+   * Prepares backoff strategy from envoy::config::core::v3::ApiConfigSource
+   * @param config envoy::config::core::v3::ApiConfigSource
+   * @param random random generator
+   * @return BackOffStrategyPtr
+   */
+  static BackOffStrategyPtr
+  prepareBackoffStrategy(const envoy::config::core::v3::ApiConfigSource& api_config_source,
```
adisuissa, line 548: The input is very specific to config-source, and it should probably be more generalized. Seems similar to the code [here](https://github.com/envoyproxy/envoy/blob/5991a2053e981038c105a421886af073be9c5941/source/common/config/datasource.cc#L70-L80).

envoy/common/backoff_strategy.h
```diff
@@ -28,6 +28,14 @@ class BackOffStrategy {
...
    */
   virtual void reset(uint64_t base_interval) PURE;
+
+  /**
+   * Checks if a time interval is greater than the maximum time interval configured for a backoff
+   * strategy.
+   * @param interval time interval to be checked.
+   * @return returns true if interval is greater than the maximum time interval
+   */
+  virtual bool isOverTimeLimit(uint64_t interval_ms) PURE;
```
adisuissa, line 38: The maximum time interval doesn't seem to be a shared property among all backoff strategies, but very specific to `JitteredExponentialBackOffStrategy`. I think this PR can be redesigned without introducing this to the BackoffStrategy interface.

adisuissa (changes requested):
Thanks for your contribution.
A high-level comment: there seems to be a mismatch between the expected use of a specific backoff strategy (`JitteredExponentialBackOffStrategy`), and using the generic type in GrpcStream. IIUC this comes from the need of GrpcStream to query the max-value of that derived class. I suggest using the derived class directly, instead of more generic backoff-strategy, as there is no way to change the type of strategy from the config.

/wait

source/common/config/utility.h
```diff
@@ -531,6 +537,45 @@ class Utility {
...
+                         Random::RandomGenerator& random) {
+
```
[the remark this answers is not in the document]
jaysonjdmello: moved implementation to `source/common/config/utility.cc`.

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -43,6 +44,11 @@ message GrpcService {
     string authority = 2
         [(validate.rules).string =
              {min_len: 0 max_bytes: 16384 well_known_regex: HTTP_HEADER_VALUE strict: false}];
+
+    // Specifies parameters that control :ref:`retry backoff strategy <envoy_v3_api_msg_config.core.v3.BackoffStrategy>`.
+    // This parameter is optional, in which case the default base interval for retry is 500 milliseconds.
+    // The Default maximum interval is 30 seconds.
```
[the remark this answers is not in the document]
jaysonjdmello: done

source/common/config/utility.h
jaysonjdmello: done, refactored to a generalized implementation and reused in `source/common/config/datasource.cc`

envoy/common/backoff_strategy.h
jaysonjdmello: done, redesigned the PR to use the derived class directly

jaysonjdmello:
> 

Thanks for your feedback!
I have redesigned the PR to use the derived class directly. Also made other changes as suggested/

source/common/config/utility.h
```diff
@@ -531,7 +531,23 @@ class Utility {
...
   }
-};
 
+  /**
+   * Prepares Jittered Exponential BackOff Strategy from Retry Policy config
+   * @param config retry policy config
+   * @param random random generator
+   * @return JitteredExponentialBackOffStrategyPtr
+   */
+  static JitteredExponentialBackOffStrategyPtr prepareJitteredExponentialBackOffStrategy(
```
adisuissa, line 541: I think it would be better to have a 2 functions: one that receives the random-generator and an optional RetryPolicy (OptRef<const>), and another that is just like the first but with additional default values (the first function calling the second).
This way the function could check whether there is a retry policy, and if not, use the given default values.

adisuissa (changes requested):
Thanks!
Added a few more comments that I hope will improve things.
/wait

source/common/config/utility.h
jaysonjdmello: Sorry, I am trying to confirm my understanding here. Are you referring to something like this?

 ```
static JitteredExponentialBackOffStrategyPtr prepareJitteredExponentialBackOffStrategy(
      OptRef<const envoy::config::core::v3::RetryPolicy> retry_policy,
      Random::RandomGenerator& random);
```
&

```
static JitteredExponentialBackOffStrategyPtr prepareJitteredExponentialBackOffStrategy(
      OptRef<const envoy::config::core::v3::RetryPolicy> retry_policy,
      Random::RandomGenerator& random, uint32_t retry_initial_delay_ms = 1000,
      uint32_t retry_max_delay_ms = 10000);
```

File: source/common/config/subscription_factory_impl.cc [hunk 10]
```diff
@@ -110,23 +125,35 @@ SubscriptionPtr SubscriptionFactoryImpl::subscriptionFromConfigSource(
...
+
+      JitteredExponentialBackOffStrategyPtr backoff_strategy;
+      auto& grpc_services = api_config_source.grpc_services();
+      if (!grpc_services.empty() && grpc_services[0].has_envoy_grpc() &&
+          grpc_services[0].envoy_grpc().has_retry_policy()) {
+        backoff_strategy = Utility::prepareJitteredExponentialBackOffStrategy(
+            grpc_services[0].envoy_grpc().retry_policy(), api_.randomGenerator());
+      } else {
+        backoff_strategy =
+            Utility::prepareDefaultJitteredExponentialBackOffStrategy(api_.randomGenerator());
```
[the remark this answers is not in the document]
jaysonjdmello: agree. I think to reduce the impact of the change, I am proposing to not make changes to `datasource.cc` in this PR and to set the default values as is to :
```
constexpr uint32_t RetryInitialDelayMilliseconds = 500;
 constexpr uint32_t RetryMaxDelayMilliseconds = 30000;
```

source/common/config/subscription_factory_impl.cc
```diff
@@ -110,23 +125,35 @@ SubscriptionPtr SubscriptionFactoryImpl::subscriptionFromConfigSource(
...
+      }
```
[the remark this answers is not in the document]
jaysonjdmello: Changed the function implementation and declaration

source/common/config/subscription_factory_impl.cc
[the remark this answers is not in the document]
jaysonjdmello: > Thanks! Added a few more comments that I hope will improve things. /wait

Thank you! Updated with suggested comments and refactored the `prepareJitteredExponentialBackOffStrategy` to return `nullptr` if no back-off values are provided through the config.

jaysonjdmello:
> Thanks! Added a few more comments that I hope will improve things. /wait

Thank a lot for your review! I  updated with suggested comments and refactored the `prepareJitteredExponentialBackOffStrategy` method

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -43,6 +43,11 @@ message GrpcService {
     string authority = 2
         [(validate.rules).string =
              {min_len: 0 max_bytes: 16384 well_known_regex: HTTP_HEADER_VALUE strict: false}];
+
+    // Indicates the retry policy for re-establishing the gRPC channel
+    // This field is optional. If max interval is not provided, it will be set to ten times the provided base interval.
+    // Currently only supported for xDS gRPC streams.
```
adisuissa, line 49: Please also state the defaults of the xDS gRPC streams.

source/common/config/datasource.cc
```diff
@@ -64,24 +66,12 @@ RemoteAsyncDataProvider::RemoteAsyncDataProvider(
...
-          PROTOBUF_GET_MS_REQUIRED(source.retry_policy().retry_back_off(), base_interval);
-
-      max_interval_ms = PROTOBUF_GET_MS_OR_DEFAULT(source.retry_policy().retry_back_off(),
-                                                   max_interval, base_interval_ms * 10);
-
-      if (max_interval_ms < base_interval_ms) {
-        throw EnvoyException("max_interval must be greater than or equal to the base_interval");
-      }
-    }
+  backoff_strategy_ = Utility::prepareJitteredExponentialBackOffStrategy(source, random);
```
adisuissa, line 69: You can pass `RetryInitialDelayMilliseconds` and `RetryMaxDelayMilliseconds` as optionals to the `prepareJitteredExponentialBackOffStrategy` function, so it will always return a valid strategy, and will allow one to override the defaults.

adisuissa (commented):
Thanks, and apologies for the late reply.
The review was pending for a while, and I didn't press the submit button.

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: done

source/common/config/grpc_stream.h
```diff
@@ -53,8 +54,10 @@ class GrpcStream : public Grpc::AsyncStreamCallbacks<ResponseProto>,
       });
     }
 
-    backoff_strategy_ = std::make_unique<JitteredExponentialBackOffStrategy>(
-        RetryInitialDelayMs, RetryMaxDelayMs, random_);
+    if (backoff_strategy_ == nullptr) {
+      backoff_strategy_ = std::make_unique<JitteredExponentialBackOffStrategy>(
+          RetryInitialDelayMs, RetryMaxDelayMs, random_);
+    }
```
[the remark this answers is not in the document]
jaysonjdmello: done

source/common/common/backoff_strategy.h
```diff
@@ -34,6 +34,14 @@ class JitteredExponentialBackOffStrategy : public BackOffStrategy {
     reset();
   }
 
+  /**
+   * Checks if a time interval is greater than the maximum time interval configured for a backoff
+   * strategy.
+   * @param interval time interval to be checked.
+   * @return returns true if interval is greater than the maximum time interval
+   */
+  bool isOverTimeLimit(uint64_t interval_ms) { return interval_ms > max_interval_; }
```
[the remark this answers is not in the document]
jaysonjdmello: done

source/common/config/datasource.cc
jaysonjdmello: done, thanks!

jaysonjdmello:
> Thanks, and apologies for the late reply. The review was pending for a while, and I didn't press the submit button.

Thank you for reviewing! I have addressed the review suggestions

source/common/config/subscription_factory_impl.cc
```diff
@@ -77,6 +77,14 @@ SubscriptionPtr SubscriptionFactoryImpl::subscriptionFromConfigSource(
       CustomConfigValidatorsPtr custom_config_validators =
           std::make_unique<CustomConfigValidatorsImpl>(validation_visitor_, server_,
                                                        api_config_source.config_validators());
+
+      JitteredExponentialBackOffStrategyPtr backoff_strategy;
+      auto& grpc_services = api_config_source.grpc_services();
+      if (!grpc_services.empty() && grpc_services[0].has_envoy_grpc()) {
+        backoff_strategy = Utility::prepareJitteredExponentialBackOffStrategy(
+            grpc_services[0].envoy_grpc(), api_.randomGenerator(), absl::nullopt, absl::nullopt);
+      }
```
adisuissa, line 86: This code seems to be repeating itself a few times here.
I think it can be moved to an unnamed namespace function in this file, and also ensure that a backoff-strategy is always created (either with the given config, or default values). This should simplify the `GrpcStream` c'tor.

source/common/config/utility.h
```diff
@@ -531,7 +531,47 @@ class Utility {
...
+    if (config.has_retry_policy() && config.retry_policy().has_retry_back_off()) {
+      return prepareJitteredExponentialBackOffStrategy(config.retry_policy().retry_back_off(),
+                                                       random);
+    }
+
+    // 2. Default base interval and Default maximum interval is specified
+    if (default_base_interval_ms != absl::nullopt && default_max_interval_ms != absl::nullopt) {
+      return std::make_unique<JitteredExponentialBackOffStrategy>(
+          default_base_interval_ms.value(), default_max_interval_ms.value(), random);
+    }
```
adisuissa, line 560: Looking at this again, I think this function can be further simplified. My goal is that this function will always return a valid strategy (non-nullptr), so its callers don't need to handle corner cases.
If the idea is that the caller will always pass a `default_base_interval_ms`, then the default max-interval will always be 10. WDYT?

source/common/upstream/cluster_manager_impl.cc
```diff
@@ -384,6 +384,13 @@ ClusterManagerImpl::ClusterManagerImpl(
             validation_context.dynamicValidationVisitor(), server,
             dyn_resources.ads_config().config_validators());
 
+    JitteredExponentialBackOffStrategyPtr backoff_strategy;
+    auto& grpc_services = dyn_resources.ads_config().grpc_services();
+    if (!grpc_services.empty() && grpc_services[0].has_envoy_grpc()) {
+      backoff_strategy = Config::Utility::prepareJitteredExponentialBackOffStrategy(
+          grpc_services[0].envoy_grpc(), random_, absl::nullopt, absl::nullopt);
+    }
```
adisuissa, line 392: Looks like the function that I commented above.
In this case I think it should be a static function in the subscription-factory.

source/common/config/subscription_factory_impl.cc
adisuissa: EDIT: looks like this function repeats itself in another file (see below),so I think it should be static-public.

adisuissa (commented):
Thanks, looks good.
Left a few minor comments.

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -43,6 +43,12 @@ message GrpcService {
     string authority = 2
         [(validate.rules).string =
              {min_len: 0 max_bytes: 16384 well_known_regex: HTTP_HEADER_VALUE strict: false}];
+
+    // Indicates the retry policy for re-establishing the gRPC channel
```
markdroth, line 47: I think this comment is inaccurate.  If I'm reading the code correctly, this actually controls recreating the gRPC *stream*, not the gRPC channel.

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -43,6 +43,12 @@ message GrpcService {
...
+    // This field is optional. If max interval is not provided, it will be set to ten times the provided base interval.
+    // Currently only supported for xDS gRPC streams.
```
markdroth, line 49: Why is this supported only for xDS streams?  Shouldn't this work for any stream?

If this is actually only for xDS streams, then it's not clear to me that it makes sense to put this in `GrpcService`, since that message is used for many things other than xDS streams.

source/common/config/utility.h
jaysonjdmello: the problem with always returning a valid strategy is that the default values in this case are in `grpc_stream.h`. but the config is accessible only in `subscription_factory_impl.cc`. 
We would have to move the defaults to `subscription_factory_impl.cc` but I am not sure if moving them from `grpc_stream.h`. is a good idea.

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: The intent was to support this only for xDS streams in this iteration. We can add support for others in further iterations if that is okay.

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: agreed, will fix

source/common/upstream/cluster_manager_impl.cc
jaysonjdmello: would it be better if this is moved to the `utility` class as a new method `prepareJitteredExponentialBackOffStrategy(const envoy::config::core::v3::ApiConfigSource& api_config_source, Random::RandomGenerator& random)` instead?

source/common/config/utility.h
adisuissa: > the problem with always returning a valid strategy is that the default values in this case are in `grpc_stream.h`. but the config is accessible only in `subscription_factory_impl.cc`. We would have to move the defaults to `subscription_factory_impl.cc` but I am not sure if moving them from `grpc_stream.h`. is a good idea.

Why should they stay in `grpc_stream.h`? I may be missing something, so feel free to explain a bit here.
FWIW, I think the [original PR](https://github.com/envoyproxy/envoy/pull/1124) had the default value in something that is more similar to the subscription factory that we have at the moment.

source/common/upstream/cluster_manager_impl.cc
jaysonjdmello: done, thanks!

source/common/config/utility.h
jaysonjdmello: thanks for pointing out, I have moved the defaults to `envoy/config/subscription_factory.h`

source/common/config/subscription_factory_impl.cc
jaysonjdmello: moved the repeated code to source/common/config/utility.h

jaysonjdmello:
> Thanks, looks good. Left a few minor comments.

Thanks for reviewing! refactored the changes

api/envoy/config/core/v3/grpc_service.proto
```diff
@@ -43,6 +43,12 @@ message GrpcService {
     string authority = 2
         [(validate.rules).string =
              {min_len: 0 max_bytes: 16384 well_known_regex: HTTP_HEADER_VALUE strict: false}];
+
+    // Indicates the retry policy for re-establishing the gRPC stream
+    // This field is optional. If max interval is not provided, it will be set to ten times the provided base interval.
+    // Currently only supported for xDS gRPC streams.
+    // If not set, xDS gRPC streams default base interval:500ms, maximum interval:30s will be applied.
+    RetryPolicy retry_policy = 3;
```
adisuissa, line 51: nit: just realized there are 2 RetryPolicy types, so let's make this explicit: `config.core.v3.RetryPolicy`.

test/common/config/subscription_factory_impl_test.cc
```diff
@@ -314,6 +314,34 @@ TEST_F(SubscriptionFactoryTest, HttpSubscriptionNoRefreshDelay) {
                             "refresh_delay is required for REST API configuration sources");
 }
 
+TEST_F(SubscriptionFactoryTest, GrpcSubscriptionInvalidRetryConfig) {
```
adisuissa, line 317: nit: please add a short comment before each test (here and in other tests) that explains what the test validates.

source/common/config/utility.cc
```diff
@@ -308,5 +308,31 @@ void Utility::translateOpaqueConfig(const ProtobufWkt::Any& typed_config,
...
+    uint32_t max_interval_ms =
+        PROTOBUF_GET_MS_OR_DEFAULT(backoff.value(), max_interval, base_interval_ms * 10);
+
+    if (max_interval_ms < base_interval_ms) {
+      throw EnvoyException("max_interval must be greater than or equal to the base_interval");
+    }
+    return std::make_unique<JitteredExponentialBackOffStrategy>(base_interval_ms, max_interval_ms,
+                                                                random);
+  }
+  // default maximum interval is specified
```
adisuissa, line 327: nit: Add an assertion constraint for the default_base_interval_ms (similar to the PGV constraint for `base_interval` in `BackoffStrategy`.

jaysonjdmello: done, thanks

test/common/config/subscription_factory_impl_test.cc
jaysonjdmello: done

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: done, thanks

jaysonjdmello:
> Thanks! LGTM modulo small nits.

Thank you so much! refactored to address the review comments

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: reverted the change, it looks like the linter does not like it

```
INFO: Analyzed target @com_github_bufbuild_buf//:bin/buf (1 packages loaded, 1 target configured).
INFO: Found 1 target...
WARNING: /build/tmp/_bazel_envoybuild/b570b5ccd0454dc9af9f65ab1833764d/external/com_github_bufbuild_buf/BUILD.bazel:6:10: @com_github_bufbuild_buf//:bin/buf is a source file, nothing will be built for it. If you want to build a target that consumes this file, try --compile_one_dependency
INFO: Elapsed time: 0.938s, Critical Path: 0.04s
INFO: 1 process: 1 internal.
INFO: Running command line: /build/tmp/_bazel_envoybuild/b570b5ccd0454dc9af9f65ab1833764d/external/com_github_bufbuild_buf/bin/buf lint
INFO: Streaming build results to: https://source.cloud.google.com/results/invocations/6a1a3f8a-844c-42b2-a403-3c2bbbc1ca17
Format check failed, try apply following patch to fix:
diff --git a/api/envoy/config/core/v3/grpc_service.proto b/api/envoy/config/core/v3/grpc_service.proto
index 7cfe602..f266c7b 100644
--- a/api/envoy/config/core/v3/grpc_service.proto
+++ b/api/envoy/config/core/v3/grpc_service.proto
@@ -48,7 +48,7 @@ message GrpcService {
     // This field is optional. If max interval is not provided, it will be set to ten times the provided base interval.
     // Currently only supported for xDS gRPC streams.
     // If not set, xDS gRPC streams default base interval:500ms, maximum interval:30s will be applied.
-    config.core.v3.RetryPolicy retry_policy = 3;
+    RetryPolicy retry_policy = 3;
   }
 
   // [#next-free-field: 9]
```
I am guessing because `RetryPolicy` and `GrpcService` are part of the same package `envoy.config.core.v3`? Please correct me here, thanks

adisuissa: Sorry about the back and forth.

adisuissa (approved):
Thanks, LGTM!

api/envoy/config/core/v3/grpc_service.proto
jaysonjdmello: No worries, thank you so much!

envoy/config/subscription_factory.h
```diff
@@ -13,6 +13,10 @@ class SubscriptionFactory {
 public:
   virtual ~SubscriptionFactory() = default;
 
+  // Default Parameters for the backoff strategy.
+  static constexpr uint32_t RetryInitialDelayMs = 500;
+  static constexpr uint32_t RetryMaxDelayMs = 30000; // Do not cross more than 30s
```
htuch, line 18: What is this comment intending to convey? If it's just "this is the max", then not needed (it's in the name). If there's some reason the max can't be more than 30s, it would be good to explain why.

test/common/config/utility_test.cc
```diff
@@ -294,6 +294,235 @@ TEST(UtilityTest, PrepareDnsRefreshStrategy) {
...
+    )EOF";
+      TestUtility::loadFromYaml(config_yaml, config);
+
+      config.mutable_retry_policy()->mutable_retry_back_off()->mutable_base_interval()->set_seconds(
+          test_base_interval_ms);
+      config.mutable_retry_policy()->mutable_retry_back_off()->mutable_max_interval()->set_seconds(
+          test_max_interval_ms);
+
+      EXPECT_ANY_THROW(
+          Utility::prepareJitteredExponentialBackOffStrategy(config, random, 500, absl::nullopt));
```
htuch, line 521: Is there any test that actually test the retry behavior, rather than just that we build the correct configuration for the retry behavior?

envoy/config/subscription_factory.h
jaysonjdmello: The intent of the PR was to not change the legacy values and behavior from https://github.com/envoyproxy/envoy/blob/main/source/common/config/grpc_stream.h#L22

Unfortunately, I am not sure why 30s was chosen to be the maximum. I can remove the comment here if that is okay.

htuch: I think just delete the comment, yes, seems a bit confusing.

test/common/config/utility_test.cc
htuch: Well, the basic jitter backoff tests make sense as unit tests, but what I'm talking about here is some test that shows that we can both configure and observe this configuration take effect in a gRPC retry. Looking at this PR and `grpc_stream_test.cc`, I don't see any new tests or assertions covering this. FWIW you would need to use fake/mock time here to avoid flakes.

alyssawilk:
@htuch is this one waiting on anything?

htuch:
Yep, I think @jaysonjdmello was going to add some tests.
/wait

test/common/config/utility_test.cc
jaysonjdmello: @htuch , added unit tests in `grpc_stream_test.cc` to test gRPC retry with custom retry values.

jaysonjdmello:
@htuch, would highly appreciate if you could provide feedback on the new added unit tests in `grpc_stream_test` , thanks!

source/common/config/utility.cc
```diff
@@ -308,5 +308,41 @@ void Utility::translateOpaqueConfig(const ProtobufWkt::Any& typed_config,
...
+
+    if (max_interval_ms < base_interval_ms) {
+      throw EnvoyException("max_interval must be greater than or equal to the base_interval");
+    }
+    return std::make_unique<JitteredExponentialBackOffStrategy>(base_interval_ms, max_interval_ms,
+                                                                random);
+  }
+
+  // default_base_interval_ms must be greater than zero
+  if ((default_base_interval_ms > 0) == false) {
```
htuch, line 329: Nit: this is a weird way to write this expression, this should be `default_base_interval_ms == 0`.

test/common/config/utility_test.cc
```diff
@@ -294,6 +294,235 @@ TEST(UtilityTest, PrepareDnsRefreshStrategy) {
   }
 }
 
+// test that default values are used correctly when no retry configuration is provided
+TEST(UtilityTest, prepareJitteredExponentialBackOffStrategyStrategyNoConfig) {
```
htuch, line 298: Nit: `PrepareJittered..`

test/common/config/utility_test.cc
```diff
@@ -294,6 +294,235 @@ TEST(UtilityTest, PrepareDnsRefreshStrategy) {
...
+    EXPECT_THROW_WITH_MESSAGE(JitteredExponentialBackOffStrategyPtr strategy =
+                                  Utility::prepareJitteredExponentialBackOffStrategy(
+                                      api_config_source, random, 0, absl::nullopt),
+                              EnvoyException, "default_base_interval_ms must be greater than zero");
+  }
+}
+
+// confirm that user provided values in the retry configuration are correctly used to prepare the
+// backoff strategy
+TEST(UtilityTest, prepareJitteredExponentialBackOffStrategyStrategyConfigFileValues) {
```
htuch, line 388: Nit: ditto

htuch (commented):
LGTM modulo tiniest nits, thanks for the tests!

source/common/config/utility.cc
jaysonjdmello: done, thanks!

test/common/config/utility_test.cc
jaysonjdmello: done, thanks!

test/common/config/utility_test.cc
jaysonjdmello: done, thanks!

jaysonjdmello:
> LGTM modulo tiniest nits, thanks for the tests!

Thank you for reviewing! I have addressed the issues

htuch (approved):
LGTM, thanks!

Issue #23288, opened 2022-09-28
uhv: validate upstream responses


ameily:
*Title*: uhv: validate upstream responses

*Description*:
The UHV component currently only validates requests and not responses from upstream. Add the plumbing necessary to validate upstream responses using UHV.

ameily:
@yanavlasov I believe you are currently working on this

yanavlasov:
Fixed by #24713

PR #24713, opened 2022-12-30
UHV: adding validation of upstream responses

yanavlasov:
Commit Message:
Implement plumbing of UHV into validation of upstream responses.
Re-enable formerly disabled tests.

Additional Description:
To validate upstream responses UHV needs to be plumbed all the way into the client codec, to validate header before the coded transforms them for the H/1 UPGRADE to H/2 CONNECT case.

This PR add plumbing into H/2 codec only. This is to make review easier. Plumbing into H/1 and H/3 codecs will be implemented in follow up PRs.

Part of Issue #23288

Risk Level: Low
Testing: Unit Tests
Docs Changes: N/A
Release Notes: N/A


api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -163,4 +164,22 @@ message HttpProtocolOptions {
...
+  // #. Syntax of HTTP/1 response line
+  // #. HTTP/2 header map validity according to `RFC 7540 section 8.1.2<https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2`_
+  // #. Syntax of HTTP/2 pseudo headers
+  // #. HTTP/3 header map validity according to `RFC 9114 section 4.3 <https://www.rfc-editor.org/rfc/rfc9114.html>`_
+  // #. Syntax of HTTP/3 pseudo headers
+  // #. Syntax of ``Content-Length`` and ``Transfer-Encoding``
+  // #. Validation of HTTP/1 responses with both ``Content-Length`` and ``Transfer-Encoding`` headers
+  //
+  // [#not-implemented-hide:]
+  // [#extension-category: envoy.http.header_validators]
```
adisuissa, line 183: IIUC, the [actual extension](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/http/header_validators/envoy_default/v3/header_validator.proto) should have the `// [#extension: envoy.http.header_validators.<some_name>]` defined there.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -163,4 +164,22 @@ message HttpProtocolOptions {
   // upstream filters will count as a final response if hedging is configured.
   // [#extension-category: envoy.filters.http.upstream]
   repeated filters.network.http_connection_manager.v3.HttpFilter http_filters = 6;
+
+  // Configuration options for Header Validation (UHV).
+  // UHV is an extensible mechanism for checking validity of HTTP responses.
```
adisuissa, line 169: The rest of the comments seem to refer to a specific extension (UHV), rather than a generic interface/extension. In a sense, the comments look more like a boolean definition of whether to use the specific extension or not. Can this be more generic?

adisuissa (commented):
Thanks! Left a minor API comment.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
yanavlasov: Fixed

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
yanavlasov: I've tried to clarify in the comment, what happens when there is nothing specified in this field, since this is going to be the most common scenario.

alyssawilk (commented):
Sorry high level before I dive in.
the PR description posits that we have to do validation in the codec before we do transforms but I don't grok why we have to on the response path and not the request path (which also does transforms)

source/extensions/upstreams/http/config.cc
```diff
@@ -93,6 +94,51 @@ getAlternateProtocolsCacheOptions(
...
+  header_validator_factory =
+      factory->createFromProto(header_validator_config.typed_config(), validation_visitor);
+  if (!header_validator_factory) {
+    throw EnvoyException(fmt::format("Header validator extension could not be created: '{}'",
+                                     header_validator_config.name()));
+  }
+#else
+  if (options.has_typed_header_validation_config()) {
+    throw EnvoyException(
+        fmt::format("This Envoy binary does not support header validator extensions.: '{}'",
```
adisuissa, line 135: nit:
```suggestion
        fmt::format("This Envoy binary does not support header validator extensions: '{}'",
```

adisuissa (commented):
Thanks, overall LGTM, modulo minor API comment.

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -163,4 +164,27 @@ message HttpProtocolOptions {
...
   // [#extension-category: envoy.filters.http.upstream]
   repeated filters.network.http_connection_manager.v3.HttpFilter http_filters = 6;
+
+  // Configuration options for Unified Header Validation (UHV).
+  // UHV is an extensible mechanism for checking validity of HTTP responses.
+  //
+  // Leaving this field unspecified, selects the default header validator, which uses the value of
+  // the ``http_protocol_options.allow_chunked_length`` field from one of the ``upstream_protocol_options`` fields for configuration.
+  // Using the default ``envoy.http.header_validators.envoy_default`` header validator overrides the
+  // ``http_protocol_options.allow_chunked_length`` in the specified ``upstream_protocol_options`` field.
```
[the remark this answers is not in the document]
yanavlasov: I've left the comment that it is used when field is empty, but removed the rest of the text.

source/extensions/upstreams/http/config.cc
yanavlasov: Done

yanavlasov:
> Sorry high level before I dive in. the PR description posits that we have to do validation in the codec before we do transforms but I don't grok why we have to on the response path and not the request path (which also does transforms)

I do not think we transform headers in the server codecs before they hit UHV (not counting adding pseudo headers in H/1 codec). In client codec in the case of H/1 UPGRADE to H/2 CONNECT the client codec changes the 200 H/2 response to 101 H/1 response which makes response to be illegal by H/2 spec. This is the main reason for pushing UHV all the way into the client codec, as keeping outside of the codec does not work for the UPGRADE->CONNECT use case.

Also looking at the code long enough, I'm inclined to move UHV from HCM into the server codec as well. It adds a bit more plumbing but simplifies so many tests.

alyssawilk:
I thought on the request path we UHV in the HCM? that's absolutely after transforms.

alyssawilk:
I thought on the request path we UHV in the HCM? that's after transforms
https://github.com/envoyproxy/envoy/blob/main/source/common/http/utility.cc#L978 is comparable to
https://github.com/envoyproxy/envoy/blob/main/source/common/http/utility.cc#L965

I think we can still consistently do validation in one place, or we have to revisit what we're doing on the request path if we're not comfortable with that.  perhaps sync Monday?

source/common/http/codec_client.h
```diff
@@ -223,7 +224,9 @@ class CodecClient : protected Logger::Loggable<Logger::Id::client>,
                          public ResponseDecoderWrapper,
                          public RequestEncoderWrapper {
     ActiveRequest(CodecClient& parent, ResponseDecoder& inner)
-        : ResponseDecoderWrapper(inner), RequestEncoderWrapper(nullptr), parent_(parent) {
+        : ResponseDecoderWrapper(inner), RequestEncoderWrapper(nullptr), parent_(parent),
+          header_validator_(
+              parent.host_->cluster().makeHeaderValidator(parent.codec_->protocol())) {
```
alyssawilk, line 229: each client will get their own validator?  We probably do this on the downstream side but is this expensive? Should we instead have one shared across the cluster, and call validate() with the codec, or is it largely stateless so cheap?

test/integration/protocol_integration_test.cc
```diff
@@ -3974,21 +4011,79 @@ TEST_P(DownstreamProtocolIntegrationTest, InvalidRequestHeaderName) {
       codec_client_->startRequest(Http::TestRequestHeaderMapImpl{{":method", "POST"},
                                                                  {":path", "/test/long/url"},
                                                                  {":authority", "sni.lyft.com"},
+                                                                 {":scheme", "http"},
                                                                  {"foo}name", "foo_value"}});
   auto response = std::move(encoder_decoder.second);
 
+  ASSERT_TRUE(codec_client_->waitForDisconnect());
```
alyssawilk, line 4018: was this failing on wrong error before?

alyssawilk (commented):
This is _way_ simpler on the source side, thanks so much for suffering through the rewrite
/wait

source/common/http/codec_client.h
yanavlasov: They will need to have each own validator to support upgrade/connect normalization. It is not expensive. Right now UHV has no state at all. Once I move upgrade/connect normalization there, it will have state for that. But it is no more compared to the non-UHV case.

test/integration/protocol_integration_test.cc
yanavlasov: It was not checking for client disconnect for H/2 case, but waiting for reset with the `ConnectionTermination` reason with is equivalent. I have simply harmonized the condition that test test should be waiting on.

alyssawilk (approved):
LGTM for google pass :-)

adisuissa (commented):
/lgtm api

api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto
```diff
@@ -163,4 +164,14 @@ message HttpProtocolOptions {
...
+
+  // Configuration options for Unified Header Validation (UHV).
+  // UHV is an extensible mechanism for checking validity of HTTP responses.
+  //
+  // [#comment:TODO(yanavlasov): Make it a link to the default header validator doc when it becomes visible.]
+  // Leaving this field unspecified, selects the default header validator ``envoy.http.header_validators.envoy_default``.
+  //
+  // [#not-implemented-hide:]
+  // [#extension-category: envoy.http.header_validators]
+  config.core.v3.TypedExtensionConfig typed_header_validation_config = 7;
```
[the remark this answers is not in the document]
yanavlasov: Renamed

adisuissa (commented):
/lgtm api

lizan (approved):

PR #24720, opened 2023-01-01
Fix upstream response decoder to close stream on trailers frame

### LINKED ISSUE #24704: When using TCP tunneling and streaming, connection closing does not cause filter chain draining
*Title*: When using TCP tunneling, client closing a connection does not cause filter chain draining

*Description*:
Say there is the following setup:
1. TCP client
2. Envoy process with the following filters:
a. Custom passthrough filter
b. TCP proxy filter with TCP tunneling (upgrade to HTTP/2) applied
4. HTTP/2 server

After this setup running, the TCP client opens a connection, then streams data for 5 seconds, then waits for 5 seconds and closes the connection using FIN TCP flag. The outcome observed is that the HTTP/2 server is aware of the connection being closed, but in Envoy perspective, the filter chain is not being drained, and all filters are still alive (although upstream and downstream both have closed the connection).

Further information per local debugging:
1. The destructor for the custom passthrough filter was not called
2. [void Filter::onDownstreamEvent(Network::ConnectionEvent event)](https://github.com/envoyproxy/envoy/blob/main/source/common/tcp_proxy/tcp_proxy.cc#:~:text=void%20Filter%3A%3AonDownstreamEvent(Network%3A%3AConnectionEvent%20event)%20%7B) is never called (in common/tcp_proxy/tcp_proxy.cc, line 661)
3. [HttpUpstream::onDownstreamEvent(Network::ConnectionEvent event)](https://github.com/envoyproxy/envoy/blob/main/source/common/tcp_proxy/upstream.cc#:~:text=HttpUpstream%3A%3AonDownstreamEvent(Network%3A%3AConnectionEvent%20event)%20%7B) is never called (in common/tcp_proxy/upstream.cc, line 100)
4. [void HttpUpstream::doneReading()](https://github.com/envoyproxy/envoy/blob/main/source/common/tcp_proxy/upstream.cc#:~:text=void%20HttpUpstream%3A%3AdoneReading()%20%7B) is never called (in common/tcp_proxy/upstream.cc, line 142)
5. [void HttpUpstream::doneWriting()](https://github.com/envoyproxy/envoy/blob/main/source/common/tcp_proxy/upstream.cc#:~:text=void%20HttpUpstream%3A%3AdoneWriting()%20%7B) is being called once after the HTTP/2 server closes the connection (in common/tcp_proxy/upstream.cc, line 149), however read_half_closed_ is false at this time (which is because doneReading was not called)
6. I have enabled 'connection half close', but this still happens. I am not sure this is really related, because it is expected that end_stream from downstream would never be true, regardless of half close state. That's because the TCP client closes the connection with a delayed TCP packet, and not in conjunction with a data payload.
7. Seems like what could have solved it, is if "doneReading" was called. However, the only place this function is being called is in [decodeData](https://github.com/envoyproxy/envoy/blob/main/source/common/tcp_proxy/upstream.h#:~:text=parent_.doneReading()%3B), when end_stream is true. But again, it would never be true.
In all the sections above, the determination if a function was called or not was made by printing logs and verifying that they do not exist in the output.

I'm not an Envoy expert so I might have missed something here, but overall, the observable behavior is not as expected. Seems to me that calling doneReading when FIN event is received would solve this, but I am still not sure.

*Config*:
```
- name: custom_filter
typed_config:
  "@type": type.googleapis.com/envoy.extensions.filters.custom_filter.CustomFilterConfig
- name: tcp_proxy
typed_config:
  stat_prefix: tcp_stats
  cluster: "upstream"
  tunneling_config:
	hostname: host.com:443
	use_post: true
	post_path: <path>
```

ohadvano:

Fixes #24704


ohadvano:
@alyssawilk can you please review?

source/common/tcp_proxy/upstream.h
```diff
@@ -183,17 +183,24 @@ class HttpUpstream : public GenericUpstream, protected Http::StreamCallbacks {
     void decodeData(Buffer::Instance& data, bool end_stream) override {
       parent_.upstream_callbacks_.onUpstreamData(data, end_stream);
       if (end_stream) {
+        done_reading_called_ = true;
```
alyssawilk, line 186: I don't think this boolean is necessary.  We should only get end stream once 
Arguably we should handle done reading in headers as well if there's a headers-only response.

source/common/tcp_proxy/upstream.h
```diff
@@ -183,17 +183,24 @@ class HttpUpstream : public GenericUpstream, protected Http::StreamCallbacks {
...
+        parent_.doneReading();
+      }
+    }
+    void decodeTrailers(Http::ResponseTrailerMapPtr&&) override {
+      if (!done_reading_called_) {
+        done_reading_called_ = true;
         parent_.doneReading();
```
alyssawilk, line 193: this seems like a good fix, but needs unit tests

alyssawilk (commented):
Thanks for tackling this!  It needs to be runtime guarded per CONTRIBUTING.md and should have a release note as well.
/wait

ohadvano:
> Thanks for tackling this! It needs to be runtime guarded per CONTRIBUTING.md and should have a release note as well. /wait

Should runtime guard be added when it's referring to a bug fix? If yes, what should be the default in this case? Calling the added "doneReading()" or not?

source/common/tcp_proxy/upstream.h
ohadvano: Is it not possible that end_stream would be true but decodeTrailers will be called as well? If not then I agree and the boolean is redundant.

alyssawilk:
yeah we runtime guard bug-fixes (default correct behavior true) because sometimes operators depend on the legacy buggy behavior and it gives them an option to disable it while we sort out a long term solution for them.

source/common/tcp_proxy/upstream.h
ohadvano: Also @alyssawilk seems like in decodeHeaders the resetEncoder is being called directly in case of headers-only response (then the end_stream) would be true. Practically doneReading() would call resetEncoder, so this is case is covered. What do you say?

ohadvano:
I pushed a commit not signed off. Tried to fix it but made it worse somehow. Here's a new PR:
https://github.com/envoyproxy/envoy/pull/24769

alyssawilk:
We have instructions on the contributing page for how to fix DCO.  Please follow them through rather than continuing to open up new PRs (https://github.com/envoyproxy/envoy/pull/24769  https://github.com/envoyproxy/envoy/pull/24800)

PR #24723, opened 2023-01-01
common: refactor exponential backoff calculation

franklange:
Hi,
this is my first time contributing to this project.
I noticed that the `std::min()` call in the exponential backoff interval calculation wasn't needed since `backoff <= max_interval` is already guaranteed, so I removed the call.
Additionally, since `max_interval` is const,  I added a const member for the "halfway point" when checking whether we can double again or clamp to max. Trying this in godbolt it seemed to get rid of the div/shift. Hope this is okay, tested it locally and UTs where green.

Best, 
Frank

Risk Level: Low
Testing: Unit Test (backoff_strategy_test)
Docs Changes: N/A
Release Notes: N/A
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


KBaichoo (approved):
makes sense, thank you for your contribution!

PR #24730, opened 2023-01-03
mongo proxy bson: fix OOB read in removeString

robertfemmer:
Fixes a possible OOB read in removeString of the bson parser. The std::string constructor relies on finding a null-byte, but should rely on the bounds-checked length given by the line format.

### DIFF
--- a/source/extensions/filters/network/mongo_proxy/bson_impl.cc
+++ b/source/extensions/filters/network/mongo_proxy/bson_impl.cc
@@ -98,6 +98,9 @@ std::string BufferHelper::removeString(Buffer::Instance& data) {
   }
 
   char* start = reinterpret_cast<char*>(data.linearize(length));
+  // The BSON spec encodes both strings and cstrings with an additional null
+  // byte, however strings may contain embedded null bytes, therefore the
+  // constructor needs to be given the length of the string explicitly.
   std::string ret(start, length > 0 ? length - 1 : 0);
   data.drain(length);
   return ret;
--- a/source/extensions/filters/network/mongo_proxy/bson_impl.cc
+++ b/source/extensions/filters/network/mongo_proxy/bson_impl.cc
@@ -98,8 +98,8 @@ std::string BufferHelper::removeString(Buffer::Instance& data) {
   }
 
   char* start = reinterpret_cast<char*>(data.linearize(length));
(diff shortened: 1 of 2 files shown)


mattklein123 (changes requested):
Thanks for the fix. Please add a test that covers this case.

/wait

robertfemmer:
After thinking about this and writing the test, I have doubts about the way strings are serialized and de-serialized in the `BufferHelper` class:

The `writeString` method uses `data.add(value.data(), value.size() + 1);`. Accessing memory at `value.size() + 1` seems to be UB to me and may crash if the string is perfectly aligned at the end of a page boundary and the next page is not mapped or at least leak one byte of memory, if the string is at very end of the payload. My suggestion would be to serialize the string like so:

```
void BufferHelper::writeString(Buffer::Instance& data, const std::string& value) {
  writeInt32(data, value.size());
  data.add(value.data(), value.size());
}
```

This change could break the current set of tests.
There may be something I'm missing, though.

mattklein123:
> The writeString method uses data.add(value.data(), value.size() + 1);. Accessing memory at value.size() + 1 seems to be UB to me and may crash if the string is perfectly aligned at the end of a page boundary and the next page is not mapped or at least leak one byte of memory, if the string is at very end of the payload.

Presumably this is serializing the string and the null byte at the end of it since it's not length delimited? I think this is fine?

mattklein123:
I just checked in https://bsonspec.org/spec.html and even length delimited strings appear to have a null byte so I think the encoding is correct.

robertfemmer:
> > The writeString method uses data.add(value.data(), value.size() + 1);. Accessing memory at value.size() + 1 seems to be UB to me and may crash if the string is perfectly aligned at the end of a page boundary and the next page is not mapped or at least leak one byte of memory, if the string is at very end of the payload.
> 
> Presumably this is serializing the string and the null byte at the end of it since it's not length delimited? I think this is fine?

The issue for me was that `.data()` does not guarantee a null-byte at the end, whereas `.c_str()` does. But apparently I am mistaken here, C++11 does indeed add a null byte at the end, so there is no difference between `.c_str()` and `.data()`.

mattklein123:
Thanks can you check format?

/wait

mattklein123 (approved):
Thanks!

Issue #24459, opened 2022-12-08
mobile stats: add a config knob for enabling stats


Augustyniak:
*Title*: *A method on the engine builder that allows to enable extra envoy stats*

*Description*:
Envoy Mobile hardcodes the list of Envoy stats that it enables as part of its config.cc file in [here](https://github.com/envoyproxy/envoy/blob/main/mobile/library/common/config/config.cc#L482-L506). Add methods to engine builder(s) that allow the users of Envoy Mobile to enable the stats that they are interested in a more dynamic way.

PR #24733, opened 2023-01-03
config/perf: use prefix matchers instead of regexes

Augustyniak:
Commit Message: Converting regex matches to prefix matches as the former are way more costly. This is supposed to help with Envoy Mobile initialization times. The changes increase the number of stats that are enabled with default configuration of Envoy Mobile from 317 to 461. The profiling did not show any significant performance impact of this change but in theory reducing the number of places where we use regex should help.
Additional Description: related PR https://github.com/envoyproxy/envoy/pull/24521
Risk Level: Low
Testing: Manual
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

Signed-off-by: Rafal Augustyniak <raugustyniak@lyft.com>

### DIFF
--- a/mobile/library/common/config/config.cc
+++ b/mobile/library/common/config/config.cc
@@ -496,28 +496,22 @@ stats_config:
   stats_matcher:
     inclusion_list:
       patterns:
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.upstream_cx_[\w]+'
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.upstream_rq_[\w]+'
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.update_(attempt|success|failure)'
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.http2.keepalive_timeout'
-        - safe_regex:
-            regex: '^dns.apple.*'
-        - safe_regex:
-            regex: '^http.client.*'
-        - safe_regex:
-            regex: '^http.dispatcher.*'
-        - safe_regex:
-            regex: '^http.hcm.decompressor.*'
-        - safe_regex:
-            regex: '^http.hcm.downstream_rq_[\w]+'
-        - safe_regex:
-            regex: '^pbf_filter.*'
-        - safe_regex:
-            regex: '^pulse.*'
+        - prefix: cluster.base.upstream_cx_
+        - prefix: cluster.base_h2.upstream_cx_
+        - prefix: cluster.stats.upstream_cx_
+        - prefix: cluster.base.upstream_rq_
+        - prefix: cluster.base_h2.upstream_rq_
+        - prefix: cluster.stats.upstream_rq_
+        - prefix: cluster.base.http2.keepalive_timeout
+        - prefix: cluster.base_h2.http2.keepalive_timeout
+        - prefix: cluster.stats.http2.keepalive_timeout
+        - prefix: dns.apple.
+        - prefix: http.client.
+        - prefix: http.dispatcher.
+        - prefix: http.hcm.decompressor.
+        - prefix: http.hcm.downstream_rq_
+        - prefix: pbf_filter.
+        - prefix: pulse.
         - safe_regex:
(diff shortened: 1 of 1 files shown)


Augustyniak:
@jmarantz I was able to convert all but one matcher to `prefix`. The last one is tricky - I could try to convert is using `suffix` instead but I wanted to confirm whether using `suffix` has the same performance benefits as using `prefix`?

jmarantz:
It's always better to use suffix-matching than a regex, but it doesn't have the same special-case code currently as `"prefix: foo."` (the trailing dot triggers the optimization). It would not be hard to add an optimization for `"suffix .foo" (leading dot would trigger).

The cases like `cluster.base.upstream_cx_` are better than using regex but will still be expensive, as the elaborated string representation of the stat names will need to be created, and the rejected names remembered in a map.

jmarantz:
This new simplification looks great -- it errs on the side of disabling a wider subset of stats; assuming that's OK.

I think this could motivate someone -- perhaps in E-M , to optimize for the remaining safe-regex by only applying elaborating the string for stats whose StatName starts with the "vhost" token. That will greatly reduce the amount of string elaboration required.

But it would make sense to take some more flame-graphs with this PR in place first; this might be enough.

jmarantz:
Another option is to punt on that vhost exclusion and just let that stat live :)

Augustyniak:
> Another option is to punt on that vhost exclusion and just let that stat live :)

I definitely thought about doing this but in our case - Lyft - we have quite a few of virtual clusters and enabling all `vhost.*` stats leads to a significant increase in the total number of stats that need to be created (we are talking about thousands of stats in here). In general, we should make this inclusion list configurable (https://github.com/envoyproxy/envoy/issues/24459) so that each company can tweak it as needed and decide what trade-offs work best for them.

Augustyniak:
>This new simplification looks great -- it errs on the side of disabling a wider subset of stats; assuming that's OK.

The PR increase the number of stats enabled by default in EM from 317 to 462 but I think that that increase is fine given the fact that we use regexes in only one place now.

jmarantz:
The existence of *any* rejection expression that is not of the form `prefix: foo.` causes *all* stats to be name-elaborated so they can run through the string-matcher API.

How many more stats are added if you drop the vhost regex?

Augustyniak:
>How many more stats are added if you drop the vhost regex?

For the Envoy Mobile configuration that I am interested in the numbers look as follows:
regex vhost matcher: 2336 stats
prefix vhost matcher: 6067 stats

Augustyniak:
@jmarantz if you do not mind, I would like to merge this PR without further changes to `vhost` and collect some prod data to see whether the changes that I made in here helped or not. That being said I am a curious what you think about the `vhost` stats after learning about the increase in the number of stats I observed in https://github.com/envoyproxy/envoy/pull/24733#issuecomment-1370170222.

Merging this PR today as opposed to tmrw will speed up the process of collecting the data by a week since we are doing a beta/release cut later today.

alyssawilk (approved):

jmarantz (approved):

jmarantz:
Checking this in now and iterating makes sense to me. This is a great improvement.

checkAndRememberRejection would drop out of this flame-graph if our rejections were just based on prefix tokens w/o regexes, but it's probably a lot smaller than it used to be due to this change.

mobile/library/common/config/config.cc
```diff
@@ -496,28 +496,14 @@ stats_sinks: *stats_sinks
...
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.upstream_cx_[\w]+'
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.upstream_rq_[\w]+'
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.update_(attempt|success|failure)'
-        - safe_regex:
-            regex: '^cluster\.[\w]+?\.http2.keepalive_timeout'
-        - safe_regex:
-            regex: '^dns.apple.*'
```
Augustyniak, line 508: disable dns.apple - I added it but it has never been used by anybody I think so should be safe to remove it.

PR #24734, opened 2023-01-03
build: removing optional extensions from stripped_main_base

alyssawilk:
This sets up the stripped main library to not pull in all extensions (which in monorepo, pulls in all upstream Envoy extensions).

Risk Level: Low for Envoy, Medium for Envoy Mobile (worst case explicitly link your extensions)
Testing: CI
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/exe/BUILD
+++ b/source/exe/BUILD
@@ -67,7 +67,7 @@ envoy_cc_library(
     srcs = ["stripped_main_base.cc"],
     hdrs = ["stripped_main_base.h"],
     deps = [
-        ":envoy_common_lib",
+        ":envoy_common_with_core_extensions_lib",
         ":platform_impl_lib",
         ":process_wide_lib",
         "//source/common/thread_local:thread_local_lib",


RyanTheOptimist (approved):
LGTM. Presumably we should make sure Lyft is cool with this?

alyssawilk:
indeed, hence JP and no auto-merge
I think the force-register means CI would break if this caused problems for EM, but it could also require tweaks for their in -house build so I'm erring on the side of caution :-P

Augustyniak:
@jpsim  is OOO until next week. Our beta cut is on Tuesday night and since I would prefer not to merge this change just before the cut let's aim at merging it this week or next week (not a blocker, more of a suggestion/ask).

Could we ran some checks to check the size of the output Envoy binary? Do we expect the size of the binary to change as the result of this PR?

Augustyniak:
Ran `./bazelw build //test/performance:test_binary_size --config=sizeopt` locally:

Before: 45712653
After: 45582733

Augustyniak (approved):

alyssawilk:
@Augustyniak when you pull this into Lyft, please make sure something in your library target pulls in all your extensions 
I _think_ this would have to happen because of the forceRegister library so it _should_ be Ok but just want to call it out.

PR #24743, opened 2023-01-03
Add the ability to create struct like objects with mostly empty elements.

pradeepcrao:

Commit Message:
Additional Description:
Risk Level: Low
Testing:
Docs Changes: N/A
Release Notes:  N/A


jmarantz:
What's the use-case for this?

pradeepcrao:
The main use case is to reduce memory usage when many or most elements of a proto are expected to be empty. For eg. not [each](https://github.com/envoyproxy/envoy/blob/main/source/common/router/config_impl.h#L613) of `scheme_redirect`, `host_redirect`, `port_redirect` and `path_redirect` are expected to be present in a [config](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/route/v3/route_components.proto#L1623).

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,207 @@
...
+#include "source/common/common/assert.h"
+#include "source/common/common/utility.h"
+#include "source/common/protobuf/utility.h"
+
+#include "absl/strings/str_cat.h"
+
+namespace Envoy {
+/**
+ * These are helper macros to pack elements of the same type into a struct like
+ * object, with the expectation that some or most of the elements are not set.
```
jmarantz, line 12: I was thinking about this kind of issue for the router config, per discussion with @yurykats .

However what I was thinking was that what we really want is a map, so that if I have 10 optional strings which are mostly never specified, I can make either a `flat_hash_map<id, string>`, or maybe a unique_ptr<flat_hash_map<id, string>>`. That way the cost for all the unspecified optional things is 8 bytes, and the incremental cost for populating them is pretty small also.

It looks to me like you are saving a unique_ptr for every element here and we can do better when none/few of them are populated.

jmarantz: Also we should use InlineStringPtr if we want to reduce storage cost when they *are* specified, and and don't need to mutate the strings.

jmarantz: @yurykats FYI

pradeepcrao: `[flat_hash_map](https://abseil.io/docs/cpp/guides/container#:~:text=The%20container%20uses,for%20our%20estimates.)` is probably not efficient enough.

jmarantz: in terms of memory footprint? I can believe that.

std::map might actually better (or the absl btree_map version)

jmarantz: also `std::unique_ptr<map<>>` should be considered. But if we are really talking about a handful of elements max, then a linked list might be the best approach

struct Entry {
  Entry* next{nullptr};
  KeyEnum name;
  InlineStringPtr value;
};

That will cost 8 bytes for the Entry* in the config structure, and an extra 16 bytes overhead per populated item.

jmarantz: if you do the above we'd want the API you present to return value as an optional of some flavor, rather than having a separate has_xxx() and xxx() interface which would each cost a list-walk.

pradeepcrao: With `PACKED_DATA` as opposed to linked list, you don't pay the cost of `next` for every element (only once for the first). Also, you avoid the padding after `name` as all indexes are at the end and one byte each. Lastly, you get constant access times. We could have `T` be `InlineStringPtr` instead of `std::string` for `PACKED_DATA`.

What are your thoughts about that?

Having said that, I'm not a fan of macros but don't see a better way of reducing memory usage further.

jmarantz: What's the footprint for PACKED_DATA when nothing is populated? That's the most common case, right?

pradeepcrao: It's 8 + n + 2 bytes + padding, where n is the maximum number of elements (If my calculation is correct). So for n = 5 for example, it's 16 bytes.

jmarantz: well the linked list only has 8 bytes regardless of n :)

I don't understand how you get 8+n+2bytes+padding. I see this in the PR:
```
std::unique_ptr<T[]> data(capacity > 0 ? new T[capacity] : nullptr);
```

Doesn't that mean T*n ?

pradeepcrao: The way to use it is to first calculate the number of non-empty elements, and initialise it with that number. If all elements are empty, `capacity` will be 0.

jmarantz: Per IM I am really warming up to this data layout and have asked Pradeep to see if he can build a usable API w/o macros :)

/wait

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,316 @@
...
+  PackedStruct(const PackedStruct&) = delete;
+
+  friend void swap(PackedStruct& first, PackedStruct& second) {
+    using std::swap;
+    swap(first.data_, second.data_);
+    swap(first.indices_, second.indices_);
+  }
+
+  // Move constructor and assignment operator.
+  PackedStruct(PackedStruct&& other) noexcept : PackedStruct(0) { swap(*this, other); }
```
jmarantz, line 96: why swap? Why not just std::move the data_ an indices_ ?

jmarantz (commented):
do you still need all the macros?

source/common/common/packed_struct.h
jmarantz: I'm not convinced it's that bad to use it w/o macros, and it might be clearer and more debuggable. WDYT?

pradeepcrao: Definitely clearer and more debuggable without the macros. Will get rid of them.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,152 @@
...
+
+  void assign(size_t element_idx, T t) {
+    RELEASE_ASSERT(element_idx < static_cast<size_t>(max_size),
+                   absl::StrCat("element_idx (", element_idx, ") should be less than ", max_size));
+    auto const current_size = size();
+    // If we're at capacity and we don't have a slot for element_idx, increase capacity by 1.
+    if (!has(element_idx) && current_size == capacity()) {
+      std::unique_ptr<T[]> tmp(new T[++indices_[max_size]]);
+      for (size_t idx = 0; idx < current_size; ++idx) {
+        swap(tmp.get()[idx], data_.get()[idx]);
```
jmarantz, line 135: why not std::move ?

I think we could require that T support std::move in the API.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,152 @@
...
+ *
+ *    // Code that accesses the elements of the struct needs to check for existence first.
+ *    if (packed_struct.has(MyElementNames::element0) {
+ *      auto& element0 = packed_struct.get(MyElements::element0);
+ *    }
+ * ```
+ */
+template <class T, uint8_t max_size, class ElementName> class PackedStruct {
+public:
+  PackedStruct(size_t a_capacity = 0) : data_(a_capacity > 0 ? new T[a_capacity] : nullptr) {
```
jmarantz, line 54: I'm guessing it would be awkward to set capacity correctly from the caller's perspective.

What would be the impact if we just always sized it the the number of elements? Sure it would be an n^2 initialization process but N would be quite small I think.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,152 @@
...
+  friend void swap(PackedStruct& first, PackedStruct& second) {
+    using std::swap;
+    swap(first.data_, second.data_);
+    swap(first.indices_, second.indices_);
+  }
+
+  // Move constructor and assignment operator.
+  PackedStruct(PackedStruct&& other) noexcept : PackedStruct(0) { swap(*this, other); }
+  PackedStruct& operator=(PackedStruct&& other) {
+    swap(*this, other);
```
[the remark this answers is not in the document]
pradeepcrao: I tried `std::move` and couldn't get my test `StringStructMove` to pass. The moved from object did not seem to change.

source/common/common/packed_struct.h
pradeepcrao: Are you suggesting we always start with `capacity = max_size` and then introduce a method like `finalize` to resize it?

jmarantz: Actually I was suggesting just starting capacity=0 and incrementing as fields are set.

pradeepcrao: That makes more sense. I misunderstood, because that is already the default behaviour. Are you suggesting I get rid of the constructor argument entirely?

jmarantz: I guess there's not really a compelling reason to remove it, but I'd love to understand how it would be used. Maybe you could add another Draft PR that patches this in for a config block, and makes use of it, as a demo?

pradeepcrao: The main use case came from the now deleted macros, where I iterated over the proto message to see which fields had values first :-)

jmarantz: oh this makes sense. does that functionality require macros or could we provide it via a templated API?

pradeepcrao: I don't know how to do it without macros.

source/common/common/packed_struct.h
```diff
@@ -131,9 +129,7 @@ template <class T, uint8_t max_size, class ElementName> class PackedStruct {
     // If we're at capacity and we don't have a slot for element_idx, increase capacity by 1.
     if (!has(element_idx) && current_size == capacity()) {
       std::unique_ptr<T[]> tmp(new T[++indices_[max_size]]);
-      for (size_t idx = 0; idx < current_size; ++idx) {
-        swap(tmp.get()[idx], data_.get()[idx]);
-      }
+      std::move(data_.get(), std::next(data_.get(), current_size), tmp.get());
```
jmarantz, line 132: this is looking more like std::vector::insert :) Any reason not to use that?

source/common/common/packed_struct.h
jmarantz: Sorry about missing the point about protobufs. As that seems like a likely use case I'm fine with macros to support that if you prefer. I'm OK either way.

source/common/common/packed_struct.h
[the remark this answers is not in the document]
jmarantz: std::move guarantees the source object will be in a consistent state but doesn't guarantee it's cleared.

https://en.cppreference.com/w/cpp/utility/move

> Unless otherwise specified, all standard library objects that have been moved from are placed in a "valid but unspecified state", meaning the object's class invariants hold (so functions without preconditions, such as the assignment operator, can be safely used on the object after it was moved from):

The reason I suggest move() here is that swap() may do extra work to guarantee the state of the object we'll never look at again, so move() may be faster.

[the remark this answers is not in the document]
pradeepcrao: Got rid of swap.

source/common/common/packed_struct.h
pradeepcrao: Mainly because sizeof(std::vector) is 24. That's additional 16 bytes.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // Number of non-empty elements in the struct. Note that this can be less than
+  // capacity.
+  size_t size() const {
+    return std::accumulate(indices_.begin(), indices_.end() - 1, 0,
+                           [](uint8_t a, uint8_t b) { return b < max_size ? ++a : a; });
+  }
+
+  size_t capacity() const { return indices_[max_size]; }
```
jmarantz, line 84: do users of this class ever need to check capacity? Maybe this could be private, with friend access for tests?

source/common/common/packed_struct.h
jmarantz: that's quite a good reason :)

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,139 @@
...
+
+  void assign(size_t element_idx, T t) {
+    RELEASE_ASSERT(element_idx < static_cast<size_t>(max_size),
+                   absl::StrCat("element_idx (", element_idx, ") should be less than ", max_size));
+    auto const current_size = size();
+    // If we're at capacity and we don't have a slot for element_idx, increase capacity by 1.
+    if (!has(element_idx) && current_size == capacity()) {
+      std::unique_ptr<T[]> tmp(new T[++indices_[max_size]]);
+      std::move(data_.get(), std::next(data_.get(), current_size), tmp.get());
+      swap(data_, tmp);
```
jmarantz, line 124: would data_ = std::move(tmp) work?

jmarantz (commented):
lgtm at high level with minor comments.

source/common/common/packed_struct.h
pradeepcrao: sgtm.

source/common/common/packed_struct.h
jmarantz: Yeah I was just suggesting to swap the swap for a move, as we don't care about the residual value in tmp.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,141 @@
...
+  // Move constructor and assignment operator.
+  PackedStruct(PackedStruct&& other) noexcept = default;
+  PackedStruct& operator=(PackedStruct&& other) = default;
+
+  friend class PackedStructTest;
+
+private:
+  // Accessors.
+  T& operator[](size_t idx) {
+    RELEASE_ASSERT(idx < static_cast<size_t>(max_size),
```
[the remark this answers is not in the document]
pradeepcrao: This class is meant to be used to store user provided data (i.e. config data). If that invariant is violated in release because of faulty/brittle code, isn't it better to just crash? Not crashing is probably the worse outcome in this case. What do you think?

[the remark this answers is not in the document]
pradeepcrao: I can change it to `ENVOY_BUG` if that is your preference.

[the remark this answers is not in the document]
jmarantz: Hm.

Could this be a static assert? The public API to this is just via the enum.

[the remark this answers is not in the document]
pradeepcrao: That is an excellent suggestion. Will make changes.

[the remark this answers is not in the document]
pradeepcrao: PTAL

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,137 @@
...
+    if (!has<element_name>()) {
+      indices_[element_idx] = current_size;
+    }
+    data_.get()[indices_[element_idx]] = t;
+  }
+
+  // Number of non-empty elements in the struct. Note that this can be less than
+  // capacity.
+  size_t size() const {
+    return std::accumulate(indices_.begin(), indices_.end() - 1, 0,
```
jmarantz, line 104: is std::accumulate what you want here?

std::count_if would be more intuitive to me.

source/common/common/packed_struct.h
pradeepcrao: SGTM. I'll add that when I use this class for memory optimisations.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,138 @@
...
+  // Set element.
+  template <ElementName element_name> void set(T t) {
+    sizeCheck<element_name>();
+
+    static constexpr auto element_idx =
+        static_cast<std::underlying_type_t<ElementName>>(element_name);
+    auto const current_size = size();
+    // If we're at capacity and we don't have a slot for element_idx, increase capacity by 1.
+    if (!has<element_name>() && current_size == capacity()) {
+      std::unique_ptr<T[]> tmp(new T[++indices_[max_size]]);
```
jmarantz, line 92: nit: `auto tmp = std::make_unique<T[]>(++indices_[max_size]);` may be the more modern way to do this :)

You also have an instance on line 55 and it might be better to use a static helper method.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,138 @@
...
+    indices_[max_size] = a_capacity;
+  }
+
+  // Accessors.
+  template <ElementName element_name> T& get() {
+    sizeCheck<element_name>();
+    hasCheck<element_name>();
+    return (data_.get())[indices_[static_cast<std::underlying_type_t<ElementName>>(element_name)]];
+  }
+  template <ElementName element_name> const T& get() const {
```
jmarantz, line 71: to remove duplicated code you could define the const get in terms of the non-const one with a const_cast. up to you.

jmarantz (approved):
lgtm with a couple of minor nits.

jmarantz (approved):
this is great, thanks!

pradeepcrao:
> this is great, thanks!

Thank you for the review!

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,137 @@
...
+ *      auto& element0 = packed_struct.get<MyElements::element0>();
+ *    }
+ * ```
+ */
+template <class T, uint8_t max_size, class ElementName> class PackedStruct {
+public:
+  PackedStruct(size_t a_capacity = 0)
+      : data_(a_capacity > 0 ? std::make_unique<T[]>(a_capacity) : nullptr) {
+    static_assert(std::is_enum_v<ElementName>);
+    static_assert(max_size > 0);
```
mattklein123, line 58: pretty sure this has to be true as it is unsigned?

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,137 @@
...
+  size_t capacity() const { return indices_[max_size]; }
+
+  template <ElementName element_name> static constexpr void sizeCheck() {
+    static_assert(static_cast<size_t>(element_name) < static_cast<size_t>(max_size));
+  }
+  template <ElementName element_name> void hasCheck() const {
+    ENVOY_BUG(has<element_name>(),
+              absl::StrCat("Element corresponding to index ",
+                           static_cast<std::underlying_type_t<ElementName>>(element_name),
+                           " is not assigned"));
```
mattklein123, line 127: If this fails, the calling code is almost certainly going to enter undefined behavior and do something bad. Shouldn't this outright crash or if that is not desirable instead of has() just have get() return a StatusOr and have the caller decide what to do?

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,137 @@
...
+#include "source/common/common/assert.h"
+#include "source/common/common/utility.h"
+#include "source/common/protobuf/utility.h"
+
+#include "absl/strings/str_cat.h"
+
+namespace Envoy {
+
+/**
+ * Helper class to pack elements of the same type into a struct like
```
mattklein123, line 16: Can you write a little summary of the implementation details? It's a bit magical and it would be good to have a summary to avoid having to look through all the code to figure out how it works.

source/common/common/packed_struct.h
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  // Set element.
+  template <ElementName element_name> void set(T t) {
+    sizeCheck<element_name>();
+
+    static constexpr auto element_idx =
+        static_cast<std::underlying_type_t<ElementName>>(element_name);
+    auto const current_size = size();
+    // If we're at capacity and we don't have a slot for element_idx, increase capacity by 1.
```
mattklein123, line 89: I'm having a hard time reading this. Can you add more comments in this function?

mattklein123 (changes requested):
Neat. A few comments, thank you.

/wait

source/common/common/packed_struct.h
pradeepcrao: I had a `RELEASE_ASSERT` before :-). I see how `StatusOr` makes this much more cleaner. Will implement that.

source/common/common/packed_struct.h
pradeepcrao: PTAL, and let me know if the summary makes sense.

source/common/common/packed_struct.h
pradeepcrao: Added comments. PTAL.

pradeepcrao:
> Neat. A few comments, thank you.
> 
> /wait

Thanks for the review Matt!

source/common/common/packed_struct.h
pradeepcrao: I used OptRef instead. PTAL.

mattklein123 (approved):
LGTM, thanks.

mattklein123 (approved):

PR #24745, opened 2023-01-03
mobile: making api listener configurable

alyssawilk:
Adding envoy.reloadable_features.use_api_listener to control if the regular listener manager or the api listener manager is used.  note this does not use the usual reloadable or restart flags mechanism (due to it not being loaded at the time) but instead checks for the string literal in bootstrap YAML.

Risk Level: low
Testing: new integration test
Docs Changes: n/a
Release Notes: n/a


abeyad (approved):

mobile/library/common/engine_common.cc
```diff
@@ -5,20 +5,26 @@
...
 
 std::string hotRestartVersion(bool) { return "disabled"; }
 
 EngineCommon::EngineCommon(int argc, const char* const* argv)
-    : options_(argc, argv, &hotRestartVersion, spdlog::level::info, listener_type),
-      base_(options_, real_time_system_, default_listener_hooks_, prod_component_factory_,
-            std::make_unique<PlatformImpl>(), std::make_unique<Random::RandomGeneratorImpl>(),
-            nullptr) {
+    : options_(argc, argv, &hotRestartVersion, spdlog::level::info, default_listener_type) {
+  if (absl::StrContains(options_.configYaml(), "envoy.reloadable_features.use_api_listener")) {
```
Augustyniak, line 20: 1. Nothing against this check if the plan is to remove it soon after we prove that it's fine to use the new listener in prod. Just to confirm: after we prove that `envoy.listener_manager_impl.api` works fine we will move back to `USE_API_LISTENER` compiler flag to control whether to use old/new listener as we still need the old one for running proxy tests, right? Asking since I think that keeping `StrContains` in here in the long term is not a good idea as it can have some perf impact especially since our config tends to be relatively big (especially in cases where we make certs a part of it).
2. It became a little bit confusing with both `USE_API_LISTENER` and `envoy.reloadable_features.use_api_listener` controlling the same thing. In theory we could get rid of `USE_API_LISTENER` once we add a runtime knobs to the pubic API of the engine builder but I guess we do not want to do that since if we remove it know we will need to re-add it after we "ship" `envoy.reloadable_features.use_api_listener`

Augustyniak: 3. I am unsure of the exact format of `configYaml()` string but just making sure that "envoy.reloadable_features.use_api_listener" substring will be present in there since the config.cc looks as follows (notice multiple lines):
https://github.com/envoyproxy/envoy/blob/d2fe5cc5560149f8b93e18e44674d1bd3426ae1c/mobile/library/common/config/config.cc#L522-L534

alyssawilk: hmmmmm.   I'd been planning on just reverting to the ifndef but realistically I could imagine someone wanting to have an E-M instance with proxy capabilities at some point down the line.  I wonder if we should just have an explicit optional listener manager type in the bootstrap configs and use that.

@adisuissa @envoyproxy/api-shepherds the ListenerManager class which handles the creation and destruction is pluggable so E-M can avoid all the TCP downstream code.  Would anyone object to me adding a field for that config, a placeholder proto for the default listener manager Envoy always uses and an E-M specific proto for the API listener manager so we can configure all of this explicitly rather than trying to work around runtime instantiation issues?  I really would prefer to just stuff this in a runtime string since no one else needs it but runtime is initialized after the listener manager is created.

adisuissa: Can this be defined by a [bootstrap extension](https://github.com/envoyproxy/envoy/blob/1d42ee8d707c2cae05008606f83bc5fdad4fa914/api/envoy/config/bootstrap/v3/bootstrap.proto#L282)?
If so, it won't add an E-M specific field to the bootstrap.

alyssawilk: Yeah I was looking at that, but it's not optimal because it's one of those "if X is not configured we need to default to Y" type things.  It's easy to tell if an optional proto field is empty. It's both somewhat ugly to test if a factory of a given type is registered, and then the listener manager factory is a different type than the bootstrap factory so we need some exceptions there.  
I'll prototype this morning (sans tests) and ping you for thoughts.

lizan (commented):
/lgtm api

mobile/library/common/engine_common.cc
```diff
@@ -5,20 +5,19 @@
...
 std::string hotRestartVersion(bool) { return "disabled"; }
 
 EngineCommon::EngineCommon(int argc, const char* const* argv)
-    : options_(argc, argv, &hotRestartVersion, spdlog::level::info, listener_type),
-      base_(options_, real_time_system_, default_listener_hooks_, prod_component_factory_,
-            std::make_unique<PlatformImpl>(), std::make_unique<Random::RandomGeneratorImpl>(),
-            nullptr) {
+    : options_(argc, argv, &hotRestartVersion, spdlog::level::info) {
+  // TODO(alyssar) when this defaults true, move E-M default config over to boostrap config.
+  if (absl::StrContains(options_.configYaml(), "use_api_listener: true")) {
```
Augustyniak, line 13: OK, so we will be able to get rid of that `StrContains` call altogether once we "ship" the light listener manager. I am fine with having it in the meantime.

Augustyniak (approved):
left one comment - pls correct me if what I stated it in is incorrect. Otherwise looks good to me.

mobile/library/common/engine_common.cc
alyssawilk: indeed.

Augustyniak:
LGTM

mobile/test/common/integration/client_integration_test.cc
```diff
@@ -340,5 +347,25 @@ TEST_P(ClientIntegrationTest, DirectResponse) {
...
+  override_builder_config_ = true;
+
+  config_helper_.addConfigModifier([&](envoy::config::bootstrap::v3::Bootstrap& bootstrap) -> void {
+    envoy::config::listener::v3::ApiListenerManager api;
+    auto* boostrap_extension = bootstrap.add_bootstrap_extensions();
+    boostrap_extension->mutable_typed_config()->PackFrom(api);
+    boostrap_extension->set_name("envoy.listener_manager_impl.api");
+  });
+
+  basicTest();
```
abeyad, line 367: why doesn't TearDown need to be called here?

abeyad (approved):
LGTM, just a small question

mobile/test/common/integration/client_integration_test.cc
```diff
@@ -71,6 +73,11 @@ TEST_P(ClientIntegrationTest, Basic) {
   ASSERT_EQ(cc_.on_complete_received_byte_count, 67);
 }
 
+TEST_P(ClientIntegrationTest, Basic) {
+  basicTest();
+  TearDown();
```
RyanTheOptimist, line 78: Isn't TearDown() called automatically by the test infrastructure? I'm not familiar with it being called explicitly.

alyssawilk: it does. this is a temporay holdover from me trying to add loging at which point we ned to explictly destruct before loging macros go away. that code wasn't finished but I'd mildly prefer to leave this as-is to regresion test teardown can be called multiple times

mobile/library/common/engine_common.cc
adisuissa: Thanks @alyssawilk for working on this, and apologies for the extra work!
I agree that having a restartable_feature would be nice, but as you pointed out, the order of initialization prevents us from using it as is.
When I read your comment initially I thought this is all temporary, and I was thinking if we can reduce the churn on the APIs.

After looking at the current code, I think that your previous plan of adding an optional listener-factory extension to the bootstrap config is better. It will not enable a runtime-flag support, but I'm not sure if this is needed or not.

adisuissa (dismissed):
Thanks, LGTM!

adisuissa (approved):
LGTM, thanks!

PR #24746, opened 2023-01-03
Reduce VirtualHost memory utilization by avoiding Stats scope for virtual clusters when no virtual clusters are present

yurykats:
Reduce VirtualHost memory utilization by avoiding Stats scope for virtual clusters when no virtual clusters are present.

Currently every virtual host always creates a stats scope for virtual clusters, but that scope is only used when there are actual virtual clusters present. If no such clusters exist, the scope is just memory overhead.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/router/config_impl.cc
+++ b/source/common/router/config_impl.cc
@@ -1716,9 +1716,6 @@ VirtualHostImpl::VirtualHostImpl(
     ProtobufMessage::ValidationVisitor& validator,
     const absl::optional<Upstream::ClusterManager::ClusterInfoMaps>& validation_clusters)
     : stat_name_storage_(virtual_host.name(), factory_context.scope().symbolTable()),
-      vcluster_scope_(Stats::Utility::scopeFromStatNames(
-          scope, {stat_name_storage_.statName(),
-                  factory_context.routerContext().virtualClusterStatNames().vcluster_})),
       global_route_config_(global_route_config),
       per_filter_configs_(virtual_host.typed_per_filter_config(), optional_http_filters,
                           factory_context, validator),
@@ -1799,6 +1796,9 @@ VirtualHostImpl::VirtualHostImpl(
   }
 
   if (!virtual_host.virtual_clusters().empty()) {
+    vcluster_scope_ = Stats::Utility::scopeFromStatNames(
+        scope, {stat_name_storage_.statName(),
+                factory_context.routerContext().virtualClusterStatNames().vcluster_});
     virtual_cluster_catch_all_ = std::make_unique<CatchAllVirtualCluster>(
         *vcluster_scope_, factory_context.routerContext().virtualClusterStatNames());
     for (const auto& virtual_cluster : virtual_host.virtual_clusters()) {


jmarantz (approved):
/assign-from envoyproxy/senior-maintainers

source/common/router/config_impl.cc
```diff
@@ -1799,6 +1796,9 @@ VirtualHostImpl::VirtualHostImpl(
   }
 
   if (!virtual_host.virtual_clusters().empty()) {
+    vcluster_scope_ = Stats::Utility::scopeFromStatNames(
```
yanavlasov, line 1799: I think it may be better to move into constructor initializer list, and use the ternary operator? In this way the vcluster_scope_ can be made const (which it should have been in the first place).

yanavlasov (commented):
/wait-any

source/common/router/config_impl.cc
yurykats: That brings a question... Does it even need to be a class member?

jmarantz: I see your point, but looking at this, I think I'd vote against it.

Instead, what we should actually do (IMO) is drop stat_name_storage_ as a member variable altogether, and create that stat-name only in this if-block.

jmarantz: the scope needs to be a class member as that's what dictates the lifetime of the stats.

When the shared ref to the scope is dropped all the stats created in it will be deleted.

yurykats: Having said that, I think having all vcluster-related initializations happen in one place is more readable/discoverable. But if you feel strongly about it being const, I can move it back to the c'tor initializer list.

jmarantz: You can do that, but like I said above -- you'll need to use a static helper method.

That will give you readability and also population in the initializer.

jmarantz:
Fwiw I somewhat prefer the current structure for reasons Yuru gives.

source/common/router/config_impl.cc
yurykats: Oh, so `stat_name_storage_` does not need to be created either? And does not need to live for the lifetime of the vhost?

jmarantz:
Correct. It never needs to live beyond the creation of the score.

source/common/router/config_impl.cc
yurykats: This isn't obvious to me, since `stat_name_storage_` is used in vhost's `statName()` API...

jmarantz: I missed that we exposed that to end-users, and that it's possible to have a non-empty vhost-name but empty clusters.

QQ: how often does that actually happen?

yurykats: It happens most of the time. vhosts typically (always?) have names. virtual clusters are an advanced feature. I don't know how widely it's used.

jmarantz: OK I'm back to thinking what you have now looks good :)

yurykats: Thanks, resolving this thread.

zuercher (approved):

adisuissa:
Merging as this @jmarantz and @zuercher approved.

PR #24749, opened 2023-01-04
Lua: describe the get result when multiple headers have the same key

### LINKED ISSUE #24718: Does Envoy guarantee all headers fetch from the Lua/Wasm API to be lowercase?
AFAIK, from the implementation, it seems Envoy always converts the headers to lowercase in the API.

BTW, when using an API to get a single header value, like https://www.envoyproxy.io/docs/envoy/v1.24.1/configuration/http/http_filters/lua_filter#get, and there are multiple headers in the same name, which header will be guaranteed to be returned, the first or the last?

RRobot-lm:
Signed-off-by: RRobot-lm <RRobot-lm@users.noreply.github.com>

Commit Message: Lua: describe the get result when multiple headers have the same key
Additional Description:
Risk Level: Low
Testing: N/A
Docs Changes: Done
Release Notes:
Platform Specific Features:
Fixes #24718

### DIFF
--- a/docs/root/configuration/http/http_filters/lua_filter.rst
+++ b/docs/root/configuration/http/http_filters/lua_filter.rst
@@ -631,7 +631,8 @@ get()
   headers:get(key)
 
 Gets a header. *key* is a string that supplies the header key. Returns a string that is the header
-value or nil if there is no such header.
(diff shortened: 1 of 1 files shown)


mattklein123 (approved):
Thanks!

PR #24750, opened 2023-01-04
core: add missing extension registration

Augustyniak:
Commit Message: A follow up fixes to https://github.com/envoyproxy/envoy/pull/24495. Without these fixes application crashes after `addDirectResponse` call on `TestEngineBuilder` (iOS only api) as the `handler_` in https://github.com/envoyproxy/envoy/blob/93ea91f42ef749a4560a4596174402a5aad8ee65/source/server/worker_impl.cc#L61 is `nullptr`.
Additional Description: The question is whether without the added registration calls application may crash even without the `addDirectResponse` call. cc @alyssawilk. Asking since Lyft has pushed https://github.com/envoyproxy/envoy/pull/24495 to prod pipeline already and the question is whether we need to cherry-pick the fix from this PR now.
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

Signed-off-by: Rafal Augustyniak <raugustyniak@lyft.com>

### DIFF
--- a/mobile/envoy_build_config/BUILD
+++ b/mobile/envoy_build_config/BUILD
@@ -28,6 +28,7 @@ envoy_cc_library(
         "@envoy//source/extensions/filters/http/router:config",
         "@envoy//source/extensions/filters/network/http_connection_manager:config",
         "@envoy//source/extensions/http/header_formatters/preserve_case:config",
+        "@envoy//source/extensions/listener_managers/listener_manager:connection_handler_lib",
         "@envoy//source/extensions/network/dns_resolver/getaddrinfo:config",
         "@envoy//source/extensions/request_id/uuid:config",
         "@envoy//source/extensions/stat_sinks/metrics_service:config",
--- a/mobile/envoy_build_config/extension_registry.cc
+++ b/mobile/envoy_build_config/extension_registry.cc
@@ -16,6 +16,7 @@
 #include "source/extensions/filters/network/http_connection_manager/config.h"
 #include "source/extensions/http/header_formatters/preserve_case/config.h"
 #include "source/extensions/http/original_ip_detection/xff/config.h"
+#include "source/extensions/listener_managers/listener_manager/connection_handler_impl.h"
 #include "source/extensions/listener_managers/listener_manager/listener_manager_impl.h"
 #include "source/extensions/network/dns_resolver/getaddrinfo/getaddrinfo.h"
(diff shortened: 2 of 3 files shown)


alyssawilk (approved):

Augustyniak:
@alyssawilk the crash I observed was happening when we were trying to add a listener https://github.com/envoyproxy/envoy/blob/2e8413cfe5467f96f56c3a1f857762d34653852b/source/server/worker_impl.cc#L57. That's something that we do only in our tests so it should not impact "real" apps but I wonder whether you can see any other cases for when the EM would crash (without the fix from this PR) if the application does not add/remove listeners?

alyssawilk:
Legacy add direct response doesn't serve a direct response but instead proxies to the loopback listener, which wouldn't play well when missing a connection handler.  That said I thought I fixed the proxying for add direct response in https://github.com/envoyproxy/envoy/pull/24402 so I'm not clear why adding a direct response would trigger this path any more :-/

Augustyniak:
looking at the `EnvoyConfiguration.m`:
https://github.com/envoyproxy/envoy/blob/2e8413cfe5467f96f56c3a1f857762d34653852b/mobile/library/objective-c/EnvoyConfiguration.m#L128-L137

 I think that we are still injecting the fake listener for cases when direct responses are defined. We replace `#{custom_listeners}` substring https://github.com/envoyproxy/envoy/blob/2e8413cfe5467f96f56c3a1f857762d34653852b/mobile/library/common/config/config.cc#L375 with the list of listeners (that includes the fake listener) in here https://github.com/envoyproxy/envoy/blob/2e8413cfe5467f96f56c3a1f857762d34653852b/mobile/library/objective-c/EnvoyConfiguration.m#L141-L142

Augustyniak:
Removing the listener sounds fine 👍 

@alyssawilk Do you think that having https://github.com/envoyproxy/envoy/pull/24495 without having whatever is in this PR fine for cases when EM does not add direct responses? I am asking since we pushed exactly that (https://github.com/envoyproxy/envoy/pull/24495 without the registration calls) to our prod release pipeline yesterday and I am trying to determine whether I need to patch that pushed version with the fix from this PR to avoid crashing our apps? I cannot repro a crash without `addDirectResponse` call but maybe you can think of any other case when such crash would happen?

Issue #18958, opened 2021-11-10
Support retaining clusters when still referenced by route/listeners


howardjohn:
*Title*: Support retaining clusters when still referenced by route/listeners

*Description*:

Currently, our ingress's gets clusters for all services we know about. This is often extremely wasteful, since most of them are not referenced by any routes.

We added an option to filter these out, but started seeing some issues when

* Initial state: route /foo -> cluster foo-v1
* Desired end state: route /foo -> cluster foo-v2, remove foo-v1

In practice, there is a gap between the cluster being removed and the route being updated. This results in a period of downtime on route updates.

Our ideal would be that envoy doesn't delete the cluster while a route still references it.

----

The alternative would be to solve it in the control plane. But that seems pretty hard... I guess we would need to realize foo-v1 is referenced by a route and keep it around until the route is updated.

This may be ~impossible for us (until we move to Delta XDS, anyways), since "keep it around" with sotw means we literally need a copy of it in the control plane and we don't store config after we send it, nor can we generate it again.

And to know its updated we need to send the route and wait for the ack I guess? Which also introduces a ton of complexity to the control plane and requires it to be stateful.

It seems a bit complex, and I am not even sure it can work if the proxy reconnects to a new control plane instance during the middle of this operation, unless our control plane instances sync data which is extremely complicated

[optional *Relevant Links*:]
Issue tracker on our side: https://github.com/istio/istio/issues/29131

Shikugawa:
Our expectation is that prevent the immediate drain of the cluster which is bound to the route when the route is updated via RDS, and wait to update the cluster until the next cluster is available. Is it correct?

howardjohn:
Yeah that would be ideal I think.

mattklein123:
Adding this is going to be very complicated and IMHO this is a control plane problem concern, not an Envoy concern. I'm going to push back pretty hard on adding this unless there is a very good reason.

howardjohn:
I actually generally agree with that, but my concern is that I am not convinced a control plane can actually solve this in any reasonable way.

Consider:

* Initial state: route `/foo -> cluster foo-v1`. Connecting to CP-1 instance of control plane.
* User config changes to point `/foo -> cluster foo-v2`.
* Control plane is smart, so it sends the new cluster `foo-v2` first.
* Control plane crashes, new instance CP-2 starts up and envoy reconnects
[...]

PR #16861, opened 2021-06-07
xds: allow "*" to indicate a wildcard subscription

markdroth:
Signed-off-by: Mark D. Roth <roth@google.com>

Commit Message: xds: allow `*` to indicate a wildcard subscription
Additional Description: Update xDS spec to say that `*` indicates a wildcard subscription and the existing behavior of specifying no resource names to subscribe to is just a syntactic alias for subscribing to `*`.
Risk Level: Low
Testing: N/A
Docs Changes: Included in PR
Release Notes: N/A
Platform Specific Features: N/A

Alternative doc proposal for #16855. CC @krnowak

PR #16855, opened 2021-06-07
Add explicit wildcard mode

krnowak:
Commit Message:
This adds a way to explicitly specifying the wildcard subscription. Having an explicit way of specifying the wildcard subscription allows the client to also explicitly express an interest in some additional resources on the same stream without them being ignored.

Additional Description:
-

Risk Level:
High, changes in wildcard mode.

Testing:
As a part of #15523. Also done internally.

Docs Changes:
Updated the xds-protocol docs about explicit wildcard mode.

Release Notes:
* xds: added the explicit wildcard mode, which allows subscribing to specific resource names on top of a wildcard subscription. This means that the resource name ``*`` is reserved. This may mean that in some cases the initial wildcard subscription request on the stream will not be empty, but will have a non-empty list of resources with a special name among them. See :ref:`wildcard mode description <xds_protocol_resource_hints>` and :ref:`unsubscribing from resources <xds_protocol_unsubscribing_from_resources>` for details.

Platform Specific Features:
N/A

[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #15523, opened 2021-03-16
on-demand, upstream: on-demand cluster discovery

krnowak:
I'm filing an work-in-progress PR for the feature to gather some initial feedback on the code. Right now I'm working on writing tests.

Commit Message:
This PR adds an on-demand cluster discovery feature to Envoy. The process of the discovery is driven through the on-demand filter. The on-demand filter is updated to also use a per-route configuration to enable ODCDS.

Additional Description:
There is a design document for the feature - https://docs.google.com/document/d/1AqhQnrX_7SS6PAkNwoaA2peCtk9htgwSBTRg1gzFPNw/edit#

I haven't yet implemented checking if CDS is using delta, as ODCDS should not be used together with SotW CDS.

Risk Level:
Medium (new feature)

Testing:
Currently writing tests, will also do some manual testing with cluster_header in RouteAction.

Docs Changes:
TODO (should be both API docs and some usage documentation in on-demand filter, I think).

Release Notes:
New feature: on_demand: On demand discovery of the clusters.

Platform Specific Features:
N/A

[Optional Runtime guard:]

[Optional Fixes #Issue]
Partially addresses #2500 (the clusters part).

[Optional Deprecated:]

[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #24763, opened 2023-01-04
Clarify that on-demand CDS does not support SotW xDS

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

Commit Message: Clarify that on-demand CDS does not support SotW xDS
Additional Description:
Risk Level: n/a
Testing: n/a
Docs Changes: It's a doc change.
Release Notes: n/a

I was trying to configure on-demand CDS in an Istio ingress-gateway, but the process failed with the following error:
```
2023-01-04T13:38:10.587354Z     critical        envoy assert    panic: not supported
2023-01-04T13:38:10.587411Z     critical        envoy backtrace Caught Aborted, suspect faulting address 0x5390000001e
2023-01-04T13:38:10.587416Z     critical        envoy backtrace Backtrace (use tools/stack_decode.py to get line numbers):
2023-01-04T13:38:10.587418Z     critical        envoy backtrace Envoy version: 383b3bdbec79392badac7cbf3525f09d1f0f18f8/1.24.1-dev/Clean/RELEASE/BoringSSL
2023-01-04T13:38:10.587694Z     critical        envoy backtrace #0: [0x7fcd2f3e4520]
...
2023-01-04T13:38:11.139240Z     info    ads     ADS: "@" istio-ingressgateway-5d68845c68-hqp4t.istio-system-2 terminated
2023-01-04T13:38:11.139360Z     info    ads     ADS: "@" istio-ingressgateway-5d68845c68-hqp4t.istio-system-1 terminated
2023-01-04T13:38:11.140610Z     error   Envoy exited with error: signal: aborted (core dumped)
```
I had to dive into the code base to figure out why it fails and I found this method:
```
void OdCdsApiImpl::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& resources,
  ...
  // On-demand cluster updates are only supported for delta, not sotw.
  PANIC("not supported");
}
```
so I think it's worth to mention about this limitation in the docs.

Could I also add more details to that panic message or is it intentionally so general? I noticed that most panic messages are very general, so I wonder if this is a convention...?

**Question:**
Btw. is it an intended limitation or could it be implemented? I'm considering on-demand CDS as a workaround for the problem described in #18958, so if there are no obstacles to make it work with SotW ADS, I would be interested in implementing this feature.

### DIFF
--- a/api/envoy/extensions/filters/http/on_demand/v3/on_demand.proto
+++ b/api/envoy/extensions/filters/http/on_demand/v3/on_demand.proto
@@ -19,6 +19,11 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 // [#protodoc-title: On Demand Discovery]
 // On Demand Discovery :ref:`configuration overview <config_http_filters_on_demand>`.
 // [#extension: envoy.filters.http.on_demand]
+//
+// .. warning::
+//
+//    On-demand CDS does not support SotW streams. Currently it only supports incremental xDS.
+//
 
 // Configuration of on-demand CDS.
 message OnDemandCds {
--- a/api/envoy/extensions/filters/http/on_demand/v3/on_demand.proto
+++ b/api/envoy/extensions/filters/http/on_demand/v3/on_demand.proto
@@ -22,7 +22,7 @@ option (udpa.annotations.file_status).package_version_status = ACTIVE;
 //
 // .. warning::
 //
-//    On-demand CDS does not support SotW streams. Currently it only supports incremental xDS.
+//    Envoy currently only supports On-demand CDS when using Incremental-xDS and not State-of-the-World.
 //
 
 // Configuration of on-demand CDS.


markdroth:
A quick drive-by comment: I can't think of any reason in principle that on-demand CDS wouldn't work with SotW.  This is very similar to how gRPC uses CDS at the protocol level (although in gRPC, the fetch of a CDS resource is triggered by the presence of the cluster name in the RouteConfig, not by receiving a request that actually needs to be sent to that cluster).

adisuissa:
Looking at the discussion of the original issue #15523, and quoting from the [doc](https://docs.google.com/document/d/1AqhQnrX_7SS6PAkNwoaA2peCtk9htgwSBTRg1gzFPNw/edit#):

> ... the interaction between “on-demand” discovery and “state-of-the-world” discovery may be troublesome. Basically once a cluster is discovered on-demand, it will also need to appear in the state-of-the-world discovery response, otherwise Envoy will remove the cluster. This means that the management service would need to track which Envoy instance requested for which cluster, so it can send a “tailored” state-of-the-world response.

so this seems intentional.

I wonder if this is a misbehavior of the specific management service that was used.

markdroth:
@adisuissa I think that conversation assumes that Envoy will always use a wildcard subscription for CDS.  In principle, there's no reason that Envoy can't request individual CDS resources in addition to the wildcard subscription.  I would expect it to do exactly that when on-demand CDS is used, as per discussion in #16861 and #16855.

It may be that Envoy does not yet support this mechanism for SotW, but there's no protocol-level reason that it can't do so.

adisuissa:
I agree that Envoy essentially supports CDS as wildcard/collection subscription, and that is why SotW isn't supported (although generally speaking, it shouldn't be a problem).

> My use case is to prevent loading new routes by the listener manager when ClusterLoadAssignments were not received and clusters were not warmed. In such a case, a listener returns 503, because it routes request to a non-existing cluster that in fact exists, but is not ready to serve traffic, because of warming. On-demand CDS should solve this problem, because it pauses processing requests until cluster exist. It's also important to note that in this case, the control plane would not need to tailor SotW response, because it will send the expected state.

IIUC this is because there's no warming between RDS and CDS. Technically speaking, a control-plane can send an RDS update with a route to non-existent cluster (or that will be added in a future CDS response), and that route will also return 503s.
If the control-plane were to wait until the endpoints (EDS) of that cluster were ack'd before sending the listener and/or route, that wouldn't have happened, no?

jewertow:
> IIUC this is because there's no warming between RDS and CDS.

Yes, you understand it correctly.

> If the control-plane were to wait until the endpoints (EDS) of that cluster were ack'd before sending the listener and/or route, that wouldn't have happened, no?

You're right. I was considering this approach, but it has the following downsides when we consider SotW stream:
1. When a route is changed so that it switches from cluster c1 to c2, the control plane would have to send both c1 and c2 clusters to avoid breaking traffic during an update. It means the control plane must somehow remember the previous state, but this implies more memory consumption and scalability concerns. Beside that, there is still a [case](https://github.com/envoyproxy/envoy/issues/18958#issuecomment-966470378) when it would not work.
2. Waiting on EDS ACK before sending would impact performance of pushing resources to Envoy and @howardjohn is concerned about this.

markdroth:
> If the control-plane were to wait until the endpoints (EDS) of that cluster were ack'd before sending the listener and/or route, that wouldn't have happened, no?

In general, I don't think it's a good idea for control planes to do things like that, because this is a layering violation: the transport protocol layer should just treat the resources as opaque blobs and should not need to know anything about their contents or their relationships to each other.  (Note that this is essentially the same argument I made against the control plane being required to re-send the EDS resource when the CDS resource changes in #13009.)

markdroth:
I think the ideal solution to @jewertow's use case would be for Envoy to have the ability to fetch only a specific set of individual cluster names (determined from processing the RouteConfig) rather than making a wildcard CDS subscription.  This is basically what gRPC does today, and I think it would effectively involve warming between RDS and CDS.

But if that is too complicated for Envoy to implement, then I think on-demand CDS would also be a fine approach.  And I do not believe that there is any conceptual reason that this won't work with the xDS transport protocol.

Ultimately, I think this comes down to a question of which approach is easier to implement in Envoy.

jewertow:
Thank you very much for feedback, but after analyzing this feature in more detail, I decided to not working on this for now.
@adisuissa can we merge this warning?

adisuissa (commented):
Thanks, left a minor suggestion.

adisuissa:
FWIW, I think that it is possible that ODCDS will use SotW, it just needs to handle a few corner cases.
It would be simpler if there will be a separation between singleton and collection subscriptions, but this is somewhat orthogonal to this discussion.

jewertow:
> FWIW, I think that it is possible that ODCDS will use SotW, it just needs to handle a few corner cases.

Sure, thanks. I wanted to add this warning to just save someone's time when trying to use this feature with Istio.

adisuissa (approved):
LGTM, thanks!

PR #24764, opened 2023-01-04
quic: add dumb implementation to QuicSslConnectionInfo

danzh2010:
Commit Message: overriding the existing implementation in the base class, so that those boring SSL calls to X509 methods won't cause any problem.

Risk Level: low
Testing: added new tests
Docs Changes: N/A
Release Notes: N/A


test/integration/quic_http_integration_test.cc
```diff
@@ -494,7 +494,21 @@ INSTANTIATE_TEST_SUITE_P(QuicHttpMultiAddressesIntegrationTest,
...
 TEST_P(QuicHttpIntegrationTest, GetRequestAndEmptyResponse) {
+  useAccessLog("%DOWNSTREAM_TLS_VERSION% %DOWNSTREAM_TLS_CIPHER% %DOWNSTREAM_TLS_SESSION_ID%");
   testRouterHeaderOnlyRequestAndResponse();
+  std::string log = waitForAccessLog(access_log_name_);
+  EXPECT_THAT(log, testing::HasSubstr("TLSv1.3 TLS_AES_128_GCM_SHA256 -"));
+}
+
+TEST_P(QuicHttpIntegrationTest, GetPeerAndLocalCertsInfo) {
+  // These are not implemented yet, but configuring them shouldn't cause crash.
+  useAccessLog("%DOWNSTREAM_PEER_CERT% %DOWNSTREAM_PEER_ISSUER% %DOWNSTREAM_PEER_SERIAL% "
```
alyssawilk, line 505: I they're not implemented I think that configuring them should cause a failure of the process to start up.  Configuring mTLS and having it not actually validate mutual TLS seems worse than simply not starting up?
/wait

danzh2010: I think you are concerned about a different issue and what you are asking for is already addressed in https://github.com/envoyproxy/envoy/pull/24011. Retrieval of these certs in Access log or during XFCC header handling are not strictly behind mTLS support. These test didn't config mTLS support at all, but they can still config Envoy to retrieve these certs.

alyssawilk: Ah indeed sorry I did look too quickly.  So the problem here is not configuring mTLS, but confiuring logging fields from mTLS even with mTLS not configured.  And your theory is we should allow configuring Envoy to log those fields and just log empty entries?  That seems fine to me.

alyssawilk (approved):

RyanTheOptimist (approved):

alyssawilk:
@ggreenway for !google pass

ggreenway (approved):

PR #24765, opened 2023-01-05
delta-xds: fix map name in state tracking

adisuissa:
Commit Message: fix map name in state tracking
Additional Description:
Seems that this is a copy-paste error. Fixing var name.

Risk Level: Low
Testing: N/A.
Docs Changes: N/A.
Release Notes: N/A.

### DIFF
--- a/source/common/config/new_delta_subscription_state.cc
+++ b/source/common/config/new_delta_subscription_state.cc
@@ -172,7 +172,7 @@ bool NewDeltaSubscriptionState::isHeartbeatResponse(
   }
 
   if (const auto itr = ambiguous_resource_state_.find(resource.name());
-      itr != wildcard_resource_state_.end()) {
+      itr != ambiguous_resource_state_.end()) {
     // In theory we should move the ambiguous resource to wildcard, because probably we shouldn't be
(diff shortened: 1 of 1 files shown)


source/common/config/new_delta_subscription_state.cc
```diff
@@ -172,7 +172,7 @@ bool NewDeltaSubscriptionState::isHeartbeatResponse(
   }
 
   if (const auto itr = ambiguous_resource_state_.find(resource.name());
-      itr != wildcard_resource_state_.end()) {
+      itr != ambiguous_resource_state_.end()) {
```
htuch, line 175: Makes sense, can you add a test that would have caught this?

adisuissa: I don't think that technically this is possible. In all my tests `wildcard_resource_state_.end()` and `ambiguous_resource_state_.end()` are always equal.
I'm assuming that this is because a the iterator points to some "null" location, that `end()` just returns, but will need to verify this through the abseil code if requested.

htuch: Sure, just wanted to check, thanks.

htuch (approved):

PR #24769, opened 2023-01-05
Fix upstream response decoder to close stream on trailers frame

ohadvano:


ohadvano:
@alyssawilk I am trying to avoid re-writing unit tests code in case the source code would change. Can you please confirm that the source changes are fine? I'll add the unit tests in another commit afterwards.

changelogs/current.yaml
```diff
@@ -125,6 +125,9 @@ bug_fixes:
 - area: health_checker
   change: |
     prevent writing pending data for health checkers by introducing ``ConnectionCloseType::Abort`` to avoid cascading handshake overhead from health checker's requests on timeout. This fix is related to `issue \#23718 <https://github.com/envoyproxy/envoy/issues/23718>`_.
+- area: tcp_proxy
+  change: |
```
alyssawilk, line 129: Let's change that first sentence to

When tunneling HTTP over TCP, mark the upstream connection as done reading when upstream trailers are read.

alyssawilk (commented):
Code looks good modulo lack of tests.

ohadvano:
For some reason, I had issues with the ADO build for this branch, even after many retries. Opened identical PR with same code: https://github.com/envoyproxy/envoy/pull/24800

PR #24770, opened 2023-01-05
Reduce RDS subscription memory overhead by removing unnecessary `absl:optional`

yurykats:
Reduce RDS subscription memory overhead by removing unnecessary `absl:optional`

Wrapping a raw pointer in `absl::optional` does not provide much value, but adds extra 8 bytes of memory overhead. A nullptr conveys the same meaning as "no value".

This PR switches `absl::optional<RouteConfigProvider*>` to just `RouteConfigProvider*` and `has_value` checks with checks against nullptr.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/rds/rds_route_config_provider_impl.cc
+++ b/source/common/rds/rds_route_config_provider_impl.cc
@@ -16,12 +16,12 @@ RdsRouteConfigProviderImpl::RdsRouteConfigProviderImpl(
   });
   // It should be 1:1 mapping due to shared rds config.
   ASSERT(subscription_->routeConfigProvider() == nullptr);
-  subscription_->routeConfigProvider() == this;
+  subscription_->routeConfigProvider() = this;
 }
 
 RdsRouteConfigProviderImpl::~RdsRouteConfigProviderImpl() {
   ASSERT(subscription_->routeConfigProvider() != nullptr);
-  subscription_->routeConfigProvider() == nullptr;
+  subscription_->routeConfigProvider() = nullptr;
 }
 
 const absl::optional<RouteConfigProvider::ConfigInfo>&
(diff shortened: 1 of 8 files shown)


source/common/rds/rds_route_config_provider_impl.cc
```diff
@@ -15,13 +15,13 @@ RdsRouteConfigProviderImpl::RdsRouteConfigProviderImpl(
     return std::make_shared<ThreadLocalConfig>(initial_config);
   });
   // It should be 1:1 mapping due to shared rds config.
-  ASSERT(!subscription_->routeConfigProvider().has_value());
-  subscription_->routeConfigProvider().emplace(this);
+  ASSERT(subscription_->routeConfigProvider() == nullptr);
+  subscription_->routeConfigProvider() == this;
```
pradeepcrao, line 19: Did you mean to use `=` instead of `==` here?

yurykats: Duh, of course, thanks for catching this (before tests did :))

source/common/rds/rds_route_config_subscription.h
```diff
@@ -54,7 +54,7 @@ class RdsRouteConfigSubscription : Envoy::Config::SubscriptionCallbacks,
 
   ~RdsRouteConfigSubscription() override;
 
-  absl::optional<RouteConfigProvider*>& routeConfigProvider();
+  RouteConfigProvider*& routeConfigProvider() { return route_config_provider_; }
```
pradeepcrao, line 57: Is the expectation to use this to change the value of `route_config_provider_`? Seems confusing to me to return a reference if that is not the case.

pradeepcrao (commented):
Thanks so much for doing this Yury!

source/common/rds/rds_route_config_subscription.h
yurykats: Yes, it is used to update the value. The same way the optional value was updated before (via a reference). I did contemplate adding a setter instead, but opted for consistency in this class -- if you look at other member methods they are also references.

pradeepcrao (approved):
LGTM

zuercher (approved):

adisuissa (approved):
LGTM, thanks!

PR #24776, opened 2023-01-05
Make Envoy Mobile's C++ targets public

mjbshaw:
envoy_engine_cc_lib_no_stamp is insufficient by itself. engine_builder_lib is required because it's the target that exports the engine_builder.h header. And anyone using a stamped binary needs access to envoy_engine_cc_lib.

Signed-off-by: Michael Bradshaw <mjbshaw@google.com>

Commit Message: Make Envoy Mobile's C++ targets public
Additional Description: envoy_engine_cc_lib_no_stamp is insufficient by itself. engine_builder_lib is required because it's the target that exports the engine_builder.h header. And anyone using a stamped binary needs access to envoy_engine_cc_lib.
Risk Level: Low
Testing: None, honestly.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/mobile/library/cc/BUILD
+++ b/mobile/library/cc/BUILD
@@ -13,6 +13,7 @@ envoy_cc_library(
         "engine_builder.h",
     ],
     repository = "@envoy",
+    visibility = ["//visibility:public"],
     deps = [
         ":envoy_engine_cc_lib_no_stamp",
         "@envoy//source/common/common:assert_lib",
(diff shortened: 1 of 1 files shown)


RyanTheOptimist (commented):
Thanks for doing this! This change looks good to me. Though I'm curious why envoy_engine_cc_lib_no_stamp is public but not the other two. @jpsim Do you have any backstory here?

The motivation for this PR is visibility in our internal repo which we can solve a different way, but it seems like it would be nice for open source Envoy Mobile to be internally consistent. So these targets should probably all be public (as per this PR) or none of them should be?

RyanTheOptimist:
@Augustyniak as discussed in the team meeting today...

RyanTheOptimist (approved):

Augustyniak (approved):

PR #24779, opened 2023-01-05
Add more coverage and query param support for pattern rewrite

silverstar194:

Commit Message:
Add more coverage and query param support for pattern rewrite

Additional Description:
Path template rewrites should append the query parameters to the end of the path regardless of rewrite. This is consistent with the other rewrite types (regex, prefix, etc).

Risk Level:
Testing:
Unit tests in router filter.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/router/config_impl.cc
```diff
@@ -1036,7 +1036,7 @@ absl::optional<std::string> RouteEntryImplBase::currentUrlPathAfterRewriteWithMa
       return std::string(headers.getPathValue());
     }
 
-    return *std::move(new_path);
+    return path.replace(0, just_path.size(), new_path.value());
```
RyanTheOptimist, line 1039: Is this a bug fix?

RyanTheOptimist (commented):
Nice! Thanks for adding more tests! Does this allow us to increase the limit in per_file_coverage.sh?

source/common/router/config_impl.cc
silverstar194: Yes. Currently the url params are truncated instead of appended.

RyanTheOptimist: Ah, I see. In that case it seems like this is a user-visible change. Is that right? If so, we'll probably need a reloadable feature to control it. (Unless this codepath is alpha?)

silverstar194: It is not in use anywhere. This was caught before rollout.

RyanTheOptimist (approved):
LGTM!

Please tweak the PR title/description to discuss the bugfix first and the extra tests second. Also, please fill out the following fields in the description: 

Risk Level:
Testing:
Docs Changes:
Release Notes:

(and remove any of the optional fields that do not need to be filled out).

Does this allow us to increase the limit in per_file_coverage.sh for this code?

RyanTheOptimist:
@mattklein123 for non-Google review.

mattklein123 (approved):

PR #24780, opened 2023-01-05
mobile grpc: expose cancel method on a stream

Augustyniak:
Commit Message: Expose `cancel` method on a gRPC stream.
Additional Description:
Risk Level: Low, new API
Testing: Unit/integration tests
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
+++ b/mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
@@ -51,14 +51,17 @@ class GRPCStream(
   }
 
   /**
-   * Close this connection.
+   * Close the stream, wait for the peer to finish before actually closing the stream.
    */
   fun close() {
+    // The gRPC protocol requires the client stream to close with a DATA frame.
+    // More information here:
+    // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     underlyingStream.close(ByteBuffer.allocate(0))
   }
 
   /**
-   * Cancel this connection.
+   * Cancel the stream forcefully regardless of whether the peer has more data to send.
    */
   fun cancel() {
     underlyingStream.cancel()
--- a/mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
+++ b/mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
@@ -56,4 +56,11 @@ class GRPCStream(
   fun close() {
     underlyingStream.close(ByteBuffer.allocate(0))
   }
+
+  /**
+   * Cancel this connection.
+   */
+  fun cancel() {
+    underlyingStream.cancel()
+  }
 }
--- a/mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
+++ b/mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
@@ -51,11 +51,11 @@ class GRPCStream(
   }
 
   /**
-   * Close stream by sending the "end stream" signal to the peer and 
+   * Close stream by sending the "end stream" signal to the peer and
    * then waiting for the peer to finish before actually closing the stream.
    */
   fun close() {
-    // TODO(Augustyniak): Remove the method once `cancel` method is proved 
+    // TODO(Augustyniak): Remove the method once `cancel` method is proved
     // to work fine.
 
     // The gRPC protocol requires the client stream to close with a DATA frame.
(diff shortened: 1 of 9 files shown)


mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
```diff
@@ -51,9 +51,19 @@ class GRPCStream(
   }
 
   /**
-   * Close this connection.
+   * Close the stream, wait for the peer to finish before actually closing the stream.
```
alyssawilk, line 54: Are there cases where this is used?  I thought the plan was to replace it entirely

mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
```diff
@@ -51,9 +51,19 @@ class GRPCStream(
...
+    // The gRPC protocol requires the client stream to close with a DATA frame.
+    // More information here:
+    // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     underlyingStream.close(ByteBuffer.allocate(0))
   }
+
+  /**
+   * Cancel the stream forcefully regardless of whether the peer has more data to send.
+   */
+  fun cancel() {
```
alyssawilk, line 66: mildly prefer reset stream but given there's already an API for cancel I'm going to say it's fine as-is :-P

mobile/library/swift/grpc/GRPCStream.swift
```diff
@@ -61,11 +61,16 @@ public final class GRPCStream: NSObject {
...
+  /// Close the stream, wait for the peer to finish before actually closing the stream.
   public func close() {
     // The gRPC protocol requires the client stream to close with a DATA frame.
     // More information here:
     // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     self.underlyingStream.close(data: Data())
   }
+
+  /// Cancel the stream forcefully regardless of whether the peer has more data to send.
+  public func cancel() {
```
alyssawilk, line 73: LGTM.  tests?
I'm fine with a test TODO if it's urgent to land.

alyssawilk (commented):
release note?  Also should we update the other APIs to have cancel as well?

mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
Augustyniak: I do not think that we will need it at Lyft so I am fine with removing it but since I want to run an A/B experiment first I want to keep `close` method for the time being so that my control branch continues to do what it does now.

Going to follow up with the removal if we find out that `cancel` is a clear win for us.

Augustyniak: I can add a TODO.

mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
Augustyniak: speaking from the perspective of mobile dev I decided to use `cancel` verb in here as `URLSession` and `OkHttp` networking libraries have a concept of cancelling a request while they do not have a concept of resetting a request. That does not mean that we _need_ to use `cancel` in here but in general I think that this name will be more user-friendly for its target group.

mobile/library/kotlin/io/envoyproxy/envoymobile/grpc/GRPCStream.kt
RyanTheOptimist: If you decide to keep this method (and I'm happy for it to go away :>) I wonder if we might want to say something likes, "Close these stream by sending the "end stream" signal to the peer and then wait for the peer to finish before the stream will actually close"

But less code is better if we can get away with it!

mobile/library/swift/grpc/GRPCStream.swift
```diff
@@ -61,11 +61,16 @@ public final class GRPCStream: NSObject {
     return self
   }
 
-  /// Close this connection.
+  /// Close the stream, wait for the peer to finish before actually closing the stream.
```
RyanTheOptimist, line 64: nit: I notice that these have "///" instead of "//". Is that a Swift thing or a bug?

Augustyniak: Swift thing/our convention: `///` for docs, `//` for comments

Augustyniak:
> Also should we update the other APIs to have cancel as well?

What other APIs are u referring too? Is if forGRPCStream? I think that I covered all of the existing implementations (Kotlin and Swift) but let me know if this is not the case.

PR #24790, opened 2023-01-06
generic proxy: tracing support for the generic proxy based on the generic tracing

wbpcode:
Commit Message: generic proxy: tracing support for the generic proxy based on the generic tracing
Additional Description:

Add tracing support to the generic proxy.

Risk Level: Low.
Testing: n/a.
Docs Changes: n/a.
Release Notes: wait.


wbpcode:
Before formal working on this, there is a problem need to be clarified first.

Should I create a new `Tracing` proto message in the `api/config/tracing/v3` or somewhere for protocol-independent usage?

Tracing should be protocol independent but current `Tracing` proto message in the `api/extensions/filters/network/http_connnection_manager/v3` is HTTP specific.

cc @envoyproxy/api-shepherds @envoyproxy/api-watchers 

And there is a related issue #22695. cc @mattklein123 do you have some new idea to this problem?

htuch:
This is probably a place that we would refactor if we could I think. In the absence of that option, we could either deprecate the one in HCM and move to a more neutral location as you suggest (lots of impacted users, somewhat cleaner) or have generic proxy take an HCM proto dependency (fragile, logically nonsensical but probably the least impactul). The alternative, of maintaining two parallel tracing top-level config protos with the same fields, is not great, since it means perma-skew in features and is a PITA to maintain. Any other thoughts here?

mattklein123:
> or have generic proxy take an HCM proto dependency (fragile, logically nonsensical but probably the least impactul)

I would probably vote for this just to avoid churn. We have made similar compromises in the past like this.

wbpcode:
> > or have generic proxy take an HCM proto dependency (fragile, logically nonsensical but probably the least impactul)
> 
> I would probably vote for this just to avoid churn. We have made similar compromises in the past like this.

It's acceptable for me.

But it would be better to record this to somewhere to tell future developers it's a compormise rather than a design or error. 🤔

KBaichoo:
/wait 

for formatting

@htuch I think this is waiting on your input

htuch (commented):
/lgtm api

htuch:
@wbpcode there are a lot of TODOs and no tests yet, are you planning on adding more to this PR merge? If so, please LMK when ready for contrib review/merge, thanks.

wbpcode:
> @wbpcode there are a lot of TODOs and no tests yet, are you planning on adding more to this PR? If so, please LMK when ready for contrib review/merge, thanks.

Some pre-work need to be completed first in other PRs.

I will mark this as waiting before I complete everything. Thanks.

/wait

wbpcode:
This is ready for a review. And some core code modification is split out to #25420.

cc @htuch cc @soulxu

htuch (commented):
/lgtm api

htuch:
@soulxu can you take a look as CODEOWNER? Thanks.

contrib/generic_proxy/filters/network/source/proxy.cc
```diff
@@ -16,19 +16,59 @@ namespace Extensions {
...
-    if (filter->isDualFilter()) {
-      continue;
-    }
-    filter->filter_->onDestroy();
+
+  auto decision = tracingDecision(connection_manager_tracing_config_.value(), parent_.runtime_);
+  if (decision.traced) {
+    stream_info_.setTraceReason(decision.reason);
   }
+  active_span_ = tracer->startSpan(*this, *downstream_request_stream_, stream_info_, decision);
```
soulxu, line 51: Does `NullSpan` can be used when there is no config. Then we needn't check everywhere whether the active span is null or not.

wbpcode: In my completely personal opinion, the null `active_span_` means no valid tracing config and do nothing for tracing. We cannot avoid all these checking. And there are only two positions at where `active_span_` will be checked. So I personally think it's ok. 🤔

soulxu: Actually you already use the nullspan at https://github.com/envoyproxy/envoy/pull/24790/files/824b4dae787d253e9eaab10a0dfec0d61dda30c0#diff-eb91cb182a2739c76dc1f18aa5ebfbb5139160a49e2490c2f46cf3de31010a80R237, but yes, it only save two checks, so I'm fine with it :)

soulxu:
no more question, LGTM, thanks

soulxu (approved):

wbpcode:
I think I forgot the changelog. I will add it.

wbpcode:
Hi, @soulxu, could you give a new LGTM to this PR? Thanks. 🙇

wbpcode:
And friendly ping @htuch

htuch (approved):

PR #24800, opened 2023-01-06
Fix tcp_proxy: mark done reading on upstream trailers

### LINKED ISSUE #24704: When using TCP tunneling and streaming, connection closing does not cause filter chain draining

ohadvano:

Fixes #24704


ohadvano:
@alyssawilk can you take another look? Had to close the previous PR and open a new one. I have added tests to the changes.

ohadvano:
@alyssawilk can we merge this one? :)

alyssawilk (approved):

PR #24803, opened 2023-01-06
http: removes nghttp2 from a utility library

birenroy:
Signed-off-by: Biren Roy <birenroy@google.com>

Commit Message: Rewrites Http::Common::Utility to avoid use of nghttp2 types.
Additional Description:
Risk Level: low
Testing: ran Envoy tests locally
Docs Changes:
Release Notes:
Platform Specific Features:


moderation:
/lgtm deps

Can `nghttp2` be removed as a dependency? https://github.com/envoyproxy/envoy/blob/main/bazel/repository_locations.bzl#L454-L467

KBaichoo (approved):
lgtm, thank you.

@moderation  I don't think nghttp2 can be removed yet? But I think we're making great progress towards it.

PR #24806, opened 2023-01-07
Clean up the Envoy Mobile JNI BUILD file

### LINKED ISSUE #24799: libndk_envoy_jni.so depends on both java_jni_support and ndk_jni_support
The functions `jni_log_fmt()`, `jni_log()`, and `attach_vm()`, are defined in both ndk_jni_support.cc and java_jni_support.cc. As a result if both of these files in include in a binary it will result in duplicate symbols. However, when the android toolchain is used then libndk_envoy_jni.so includes both. The dependency tree is as follows:
 
```
  libndk_envoy_jni.so -> base_java_jni_lib -> java_jni_support
  libndk_envoy_jni.so -> jni_interface_lib -> ndk_jni_support
```

This seems to not be a problem because this target is test only and doesn't seem to be build on Android.

RyanTheOptimist:
Clean up the Envoy Mobile JNI BUILD file
- Replace the two mutually exclusive targets, java_jni_support and ndk_jni_support, with a single new target, jni_support_lib, that uses select() to determine the correct source files.
- Remove android_test_jni_interface.cc in favor of the identical android_jni_interface.cc.
- Replace test-only libndk_envoy_jni.so and libjava_jni_lib.so with a single libenvoy_jni_with_test_extensions.so, and move from //library/common/jni to //test/common/jni.
- Remove some unnecessary build targets.
- Remove obsolete linkopts.
- Add some missing explicit dependencies.
- Add comments to various targets.
- Rename targets to avoid bazel warnings.
- Only use alwayslink = 1 where it is necessary.

Results in a net removal of 64 lines across library/common/jni/BUILD and test/common/jni/BUILD 

Fixes #24799

Risk Level: Low
Testing: Existing
Docs Changes: N/A
Release Notes: N/A


mobile/library/common/jni/BUILD
```diff
@@ -1,31 +1,52 @@
-load("//bazel:kotlin_lib.bzl", "envoy_mobile_so_to_jni_lib")
 load("//bazel:android_debug_info.bzl", "android_debug_info")
 load("@envoy//bazel:envoy_build_system.bzl", "envoy_package")
-load("//bazel:envoy_mobile_test_extensions.bzl", "TEST_EXTENSIONS")
 
 licenses(["notice"])  # Apache 2
 
 envoy_package()
 
+# Suport for JNI, using NDK on android and Java otherwise.
```
abeyad, line 8: it might be helpful to explain here, or elsewhere, why we use NDK on Android and "java" (not sure what java means in this context?) otherwise in order to get JNI functionality?  i didn't understand that part

mobile/library/common/jni/BUILD
```diff
@@ -34,38 +55,67 @@ cc_library(
...
     # We need this to ensure that we link this into the .so even though there are no code references.
     alwayslink = True,
 )
 
+# Cert verification related functions which call into AndroidNetworkLibrary.
 cc_library(
-    name = "envoy_jni_lib",
+    name = "android_network_utility_lib",
+    srcs = [
+        "android_network_utility.cc",
```
abeyad, line 72: should we keep the `android_network_utility_lib` target at the bottom where it was before, so it makes the diff between the two easier to follow?  i think there were no diffs but not 100% sure

RyanTheOptimist: I did consider that, and I'm happy to move it back. you're right that there are no changes. I chose to move it here so that the BUILD file could be read in-order, if that makes sense. In other words, I wanted to make sure we didn't depend on target until we'd seen them, otherwise it's pretty hard to keep the right stuff in my mind. But happy to move it back if you prefer?

RyanTheOptimist (commented):
Thanks!

mobile/library/common/jni/BUILD
abeyad: Ahh, that's great to know, thanks!  Probably can just add a very basic statement then like "using NDK on android and Java otherwise, since the JNI APIs differ a bit on Android JVMs" or something like that?

mobile/library/common/jni/BUILD
abeyad: I like your idea of keeping dependencies above their dependents, that's how I like to structure BUILD files too, so lets keep it as is.  Just wanted to make sure I didn't miss anything in the diffs but since you confirmed it, I'm good :)

mobile/library/common/jni/BUILD
```diff
@@ -82,14 +132,9 @@ config_setting(
...
 cc_binary(
     name = "libenvoy_jni.so",
-    linkopts = [
-        "-lm",
-    ] + select({
+    linkopts = select({
         "@envoy//bazel:dbg_build": ["-Wl,--build-id=sha1"],
         "//conditions:default": [],
-    }) + select({
-        "@envoy//bazel:android": ["-llog"],
```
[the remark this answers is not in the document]
abeyad: ACK, thanks!

abeyad (approved):
Thanks Ryan, great detective work to figure all this out!

abeyad (approved):

mobile/library/common/jni/BUILD
jpsim: My Android/Java expertise is limited, but Ryan's comment matches my understanding too. +1

jpsim (approved):
Epic cleanup, thank you sir.

Approved pending some internal Lyft end-to-end validation, which is kicked off, I should have results later today.

mobile/bazel/kotlin_test.bzl
```diff
@@ -32,7 +32,7 @@ def _internal_kt_test(name, srcs, deps = [], data = [], jvm_flags = [], reposito
 
 # A basic macro to make it easier to declare and run kotlin tests which depend on a JNI lib
 # This will create the native .so binary (for linux) and a .jnilib (for OS X) look up
-def envoy_mobile_jni_kt_test(name, srcs, native_deps = [], deps = [], library_path = "library/common/jni", repository = "", exec_properties = {}):
+def envoy_mobile_jni_kt_test(name, srcs, native_deps = [], deps = [], library_path = "library/common/jni:test/common/jni", repository = "", exec_properties = {}):
```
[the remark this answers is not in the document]
RyanTheOptimist: Oh, nice! Done.

mobile/library/common/jni/BUILD
```diff
@@ -1,31 +1,53 @@
-load("//bazel:kotlin_lib.bzl", "envoy_mobile_so_to_jni_lib")
 load("//bazel:android_debug_info.bzl", "android_debug_info")
 load("@envoy//bazel:envoy_build_system.bzl", "envoy_package")
-load("//bazel:envoy_mobile_test_extensions.bzl", "TEST_EXTENSIONS")
 
 licenses(["notice"])  # Apache 2
 
 envoy_package()
 
+# Suport for JNI, using NDK on android and Java otherwise, since the JNI APIs
```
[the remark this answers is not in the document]
RyanTheOptimist: Good point. Done.

mobile/test/java/integration/BUILD
```diff
@@ -15,8 +15,8 @@ envoy_mobile_android_test(
         "sandboxNetwork": "standard",
     },
     native_deps = [
-        "//library/common/jni:libndk_envoy_jni.so",
-        "//library/common/jni:libndk_envoy_jni_jnilib",
+        "//test/common/jni:libenvoy_jni_with_test_extensions.so",
+        "//test/common/jni:libenvoy_jni_with_test_extensions_jnilib",
```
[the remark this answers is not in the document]
RyanTheOptimist: Sadly, that doesn't appear to work :(
```
java.lang.UnsatisfiedLinkError: no envoy_jni_with_test_extensions_jnilib in java.library.path: [library/common/jni, test/common/jni]
```

RyanTheOptimist (commented):
Thanks!

PR #24826, opened 2023-01-09
minor update: update method name of TraceContext

wbpcode:
Commit Message: minor update: update method name of TraceContext
Additional Description:

Update `authority()` of `TraceContext` to `host()` as the `host` is more widely used to represent k8s service name, dns domain, VIP, etc.

Risk Level: n/a.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


yanavlasov:
I'm unsure this is the right naming. `authority` represents a well defined term in the HTTP standard and the header map returns the `authority` per the standard. `host` does not have this well defined meaning. I understand that `host` is used more widely in k8s but this code deals with HTTP, not k8s. As such `authority` is a more appropriate name here.

kyessenov:
I tend to agree with Yan: `authority` is internally consistent with the rest of the Envoy codebase, for better or worse.  There'll always be a nomenclature gap between the user and the machine APIs, that's why we have so many control planes that do the translation.

wbpcode:
@yanavlasov yeah, `authority` is best for HTTP. But the `TraceContext` is designed as a protocol independent abstraction. It may be implemented by thrift, dubbo or some others third party protocol proxies.
So I want to use `host` here to represent the meaning of `collection of instances` (domain, VIP, registered service name, etc) of different protocols. 🤔

Cc @kyessenov

kyessenov:
I'll leave this one to Yan. I think it makes sense to generalize tracing beyond HTTP and follow https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#nethostname.

wbpcode:
Friendly ping @yanavlasov

wbpcode:
friendly ping @yanavlasov

yanavlasov (approved):

yanavlasov:
Ok, makes sense. Thanks.

PR #24827, opened 2023-01-09
http1 codec: add fuzzer for decoder

robertfemmer:

This PR adds a fuzzer specifically targeting the http1 decoder. The current fuzzer, which is part of the http_codec_impl_fuzz_test prepares state, encodes it to wire format and then decodes the wire format again.
Hence it is bound to miss coverage in the decoder. Further it is quite slow due to reloading the
server with a randomly generated configuration for each fuzz case.


KBaichoo:
/assign @bencebeky 

Can you take a first pass since you've been digging into the http1 codec?

test/common/http/http1/http1_codec_impl_fuzz_test.cc
```diff
@@ -0,0 +1,143 @@
...
+  NiceMock<Network::MockConnection> mock_server_connection;
+  NiceMock<MockServerConnectionCallbacks> mock_server_callbacks;
+
+  ServerConnectionPtr server_;
+};
+
+static std::unique_ptr<Http1Harness> harness;
+static void reset_harness() { harness = nullptr; }
+
+// KISS fuzz
```
bencebeky, line 78: What does "KISS" mean?

test/common/http/http1/http1_codec_impl_fuzz_test.cc
```diff
@@ -0,0 +1,143 @@
...
+//  1) RequestDecoder for untrusted downstreams
+//  2) ResponseDecoder for untrusted upstreams.
+// In HTTP1, requests and responses are quite similar. Let the fuzzer generate
+// one set of headers and body to be used with both and some artifacts for the
+// HTTP Request and HTTP Response lines. To improve performance, requests and
+// responses are encoded s.t. they resemble valid HTTP streams. This
+// is a trade-off and subject to opinion. E.g. coverage could be gained by
```
bencebeky, line 85: I fear this fuzzer misses out on important corner cases by sticking to close to valid messages.  I assume the fuzzing framework in Envoy is coverage-guided, is it not?  In that case, starting with a corpus of valid messages but allowing free-form input should not pose much performance penalty, but would cover important corner cases (like invalid HTTP versions, or header lines without colon).

test/common/http/http1/http1_codec_impl_fuzz_test.cc
robertfemmer: Keep it Simple (and) Stupid.

test/common/http/http1/http1_codec_impl_fuzz_test.cc
robertfemmer: The fuzzing framework is coverage guided and as my own comment suggests, I agree with this. I might reduce this two a fuzzer that doesn't leverage protobuf for message structuring at all and try to generate more (valid) messages for the corpus to get a bigger coverage of the bat.

robertfemmer:
I opted for the unconstrained fuzzer and performance seems to be superior to the protobuf mutator. I generated a bunch of (valid) test cases with methods, versions, headers and values taken out of the envoy source files for the mutator to play with.

test/common/http/http1/http1_codec_impl_fuzz_test.cc
```diff
@@ -0,0 +1,97 @@
...
+  if (harness == nullptr) {
+    Http1Settings server_settings = fromHttp1Settings();
+    Http1Settings client_settings = fromHttp1Settings();
+    harness = std::make_unique<Http1Harness>(server_settings, client_settings);
+    atexit(reset_harness);
+  }
+
+  Buffer::OwnedImpl httpmsg;
+  httpmsg.add(buf, len);
+  harness->fuzz_request(httpmsg);
```
bencebeky, line 90: Optional: add a comment stating that request and response parsing are done by two different parser instances that do not interact (I assume this is the case, otherwise feeding the same input as request and response would not be effective).

bencebeky (changes requested):

test/common/http/http1/http1_codec_impl_fuzz_test.cc
robertfemmer: Correct, added a comment for clarification. Seems like azure has some issues causing the CI checks to fail.

bencebeky (approved):

bencebeky:
> I opted for the unconstrained fuzzer and performance seems to be superior to the protobuf mutator. I generated a bunch of (valid) test cases with methods, versions, headers and values taken out of the envoy source files for the mutator to play with.

Thank you, this looks good to me.

yanavlasov (approved):

KBaichoo (approved):

PR #24836, opened 2023-01-09
Add filters for Buf Connect RPC protocol.

### LINKED ISSUE #22679: Add support for Connect RPC protocol
*Title*: *Add support for the Connect RPC protocol, on par with gRPC and gRPC-Web*

*Description*:
Add two features to support the [Connect RPC protocol](https://connect.build/docs/protocol):

1. Export metrics for proxied Connect traffic, as Envoy does today for gRPC.
2. Add a filter to translate from the Connect protocol to gRPC, as Envoy does today to expose gRPC backends to gRPC-Web clients.

The Connect protocol and the associated RPC libraries are similar to gRPC: they use HTTP as a transport, support unary and streaming RPCs, and are encoding agnostic but usually paired with Protobuf. Unlike gRPC, the Connect protocol works with HTTP/1.1,<sup>&dagger;</sup> doesn't use HTTP trailers, and keeps the wire representation of unary RPCs [dead simple](https://connect.build/docs/protocol#unary-examples). Because it demands very little of the underlying HTTP transport, Connect is easy to adopt: clients can be as simple as a cURL one-liner, and popular frameworks like Rails, Flask, and Laravel can serve Connect traffic.

Envoy's excellent metrics are, by themselves, a significant incentive for organizations to adopt gRPC. It'd be wonderful to bring the same consistent observability to Connect traffic.

Of course, many organizations already have large gRPC deployments. A translating filter would let web browsers call these backends using a protocol that's more developer-friendly than gRPC-Web. We designed Connect with translating proxies in mind, so converting to and from gRPC is simple, reasonably efficient, and semantically lossless.

Thanks for considering this enhancement!

*Relevant Links*:
- The [Connect website](https://connect.build), especially the [project overview](https://connect.build/docs/introduction)
- The [Connect protocol specification](https://connect.build/docs/protocol)
- The [Go implementation](https://github.com/bufbuild/connect-go)
- The [TypeScript implementation](https://github.com/bufbuild/connect-web) (for browsers)
- Implementations for Node.js, Kotlin, and Swift are in progress.

<hr>
<sup>&dagger;</sup> Naturally, bidirectional streaming still requires HTTP/2. Unary, server streaming, and client streaming RPCs work over HTTP/1.1 or HTTP/2.

jchadwick-buf:
Signed-off-by: John Chadwick <jchadwick@buf.build>

Implements two Connect filters:
- `connect_stats`: Similar to `grpc_stats`. Counts messages and success/failure as stats.
- `connect_grpc_bridge`: Provides a bridge from Connect clients to gRPC servers.

The [Buf Connect](https://connect.build/) protocol provides similar features to gRPC, but supports simple HTTP/1.1 requests for unary procedure calls and does not require trailers for streaming procedure calls.

The stats filter is similar to `grpc_stats`, as mentioned prior. It could potentially be rolled into `grpc_stats`, although the Connect protocol does differ from gRPC in a fair few ways, so I do anticipate the resulting code would be a little hard to follow. The bridge filter is not exactly analogous to any existing gRPC filters. For streaming Buf Connect requests, it is somewhat similar to the `grpc_web` filter, but for unary Buf Connect requests, it's more similar to `grpc_http1_bridge`.

The bridge filter is mainly useful for customers who are interested in transitioning to using Connect incrementally, whereas the stats filter is useful for anyone who wants to ensure that their stats covers usage of the Connect protocol, either independently or in addition to gRPC.

Closes #22679.


akshayjshah:
@mattklein123 appreciate you taking the time to chat with us today and maybe sponsor this contribution. @jchadwick-buf is going to highlight areas of the diff which could use a maintainer's eye and point out the few areas where we still have a little work to do.

source/extensions/filters/http/connect_grpc_bridge/end_stream_response.cc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+} // namespace
+
+/**
+ * Serializes a Buf Connect Error object to JSON.
+ *
+ * https://connect.build/docs/protocol#error-end-stream
+ */
+void serializeJson(const Error& error, Buffer::Instance& buffer) {
```
jchadwick-buf, line 76: I think it's fair to say that this is not an ideal way to deal with serializing JSON. I am open to rewriting this using a different approach if there is a preferable one that would work for our use case.

source/extensions/filters/http/connect_stats/connect_stats_filter.cc
```diff
@@ -0,0 +1,278 @@
...
+namespace Extensions {
+namespace HttpFilters {
+namespace ConnectStats {
+
+namespace {
+
+// TODO(jchadwick-buf): This is a copy of GrpcServiceMethodToRequestNamesMap from
+// grpc_stats_filter.cc. Either this extension should be somehow merged with gRPC Stats, or at
+// least some code should probably be shared.
+class ConnectServiceMethodToRequestNamesMap {
```
jchadwick-buf, line 28: The duplication here may hint that it is better to simply merge this filter with `grpc_stats`. I'm open to that perspective even if the bridge filter remains independent. Of course, if there's another good way to reduce this duplication, I'm all-ears.

mattklein123: I would vote to merge it, yes. I think there is so much overlap in config, etc. that it doesn't make sense to have a new filter for this part.

source/extensions/filters/http/connect_stats/response_frame_counter.cc
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  try {
+    Json::ObjectSharedPtr json_body = Json::Factory::loadFromString(status_frame_->toString());
+    if (auto error = json_body->getObject("error", true); error && !error->empty()) {
+      status_code_ = error->getString("code", "");
+    } else {
+      status_code_ = "";
+    }
+  } catch (Json::Exception&) {
```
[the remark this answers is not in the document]
mattklein123: It would be better to just make a variant of `loadFromString` that returns a status/result.

mattklein123 (changes requested):
At a high level this makes sense. I left some comments and I'm happy to sponsor. Thanks.

/wait

source/extensions/filters/http/connect_grpc_bridge/end_stream_response.cc
jchadwick-buf: After some wrangling, Proto WKT works for both this use case as well as the JSON decoding we do in stats, so it has been replaced. Now both use Proto WKT. Resolving.

source/extensions/filters/http/connect_stats/connect_stats_filter.cc
jchadwick-buf: Merged. Thanks.

source/extensions/filters/http/connect_stats/response_frame_counter.cc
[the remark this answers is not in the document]
jchadwick-buf: Resolved by replacing with Protobuf WKT usage, which doesn't have this issue. Resolving thread.

source/extensions/filters/http/connect_grpc_bridge/config.cc
```diff
@@ -0,0 +1,32 @@
...
+namespace Extensions {
+namespace HttpFilters {
+namespace ConnectGrpcBridge {
+
+Http::FilterFactoryCb ConnectGrpcFilterConfigFactory::createFilterFactoryFromProtoTyped(
+    const envoy::extensions::filters::http::connect_grpc_bridge::v3::FilterConfig&,
+    const std::string&, Server::Configuration::FactoryContext&) {
+  return [](Http::FilterChainFactoryCallbacks& callbacks) {
+    callbacks.addStreamFilter(std::make_shared<ConnectUnaryToGrpcFilter>());
+    callbacks.addStreamFilter(std::make_shared<ConnectStreamingToGrpcFilter>());
```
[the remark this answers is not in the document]
jchadwick-buf: I decided to put in the effort to merge them. It was definitely a bit tricky but I think the resulting filter is not too hard to read after all, so it was probably worth it. I am considering this thread resolved.

jchadwick-buf:
I don't know why it says I added and then removed reviewers: I didn't, I only hit "Ready for Review." In any case, I think this is ready for review now. I've tested it in many scenarios and I've tried to cover the code with an adequate amount of tests. If desired, I can split this into more PRs, merge/rebase commits, etc. I tried to avoid too much rebasing for the moment so that the progres/changes were not obscured.

api/envoy/extensions/filters/http/grpc_stats/v3/config.proto
```diff
@@ -70,6 +70,10 @@ message FilterConfig {
   // This config can fix incorrect gRPC metrics with dots because the existing stats tag extractor
   // assumes no dots in the gRPC service name. By default this is set as false.
   bool replace_dots_in_grpc_service_name = 5;
+
+  // If true, the filter will instrument `Buf Connect <https://connect.build/>` protocol messages in
+  // addition to gRPC.
+  bool enable_buf_connect_support = 6;
```
adisuissa, line 76: I might be missing something here, but the filter is for GrpcStats, and Buf-Connect is a different protocol, so shouldn't there be a BufConnectStats filter (that has the relevant config)?

jchadwick-buf: What we wound up doing was merging Buf Connect support into the gRPC stats protocol due to the overlap between the protocols. (Connect is semantically compatible with gRPC, so it's somewhat akin to say, gRPC-web.) I'm open to trying to make this make more sense. Context: https://github.com/envoyproxy/envoy/pull/24836#discussion_r1065012094

mattklein123: Given the similarity of the protocols I think it's reasonable to have it be part of this filter. We can potentially make this more clear in the docs. I will take a look.

mattklein123:
Sorry for the review delay. I will get back to this within the next week.

api/envoy/extensions/filters/http/grpc_stats/v3/config.proto
mattklein123: Is there any reason to actually have this be configurable? It's turned on by content-type, right? Can we just turn it on automatically and then just document that it is supported somewhere in the top level filter documentation?

docs/root/configuration/http/http_filters/connect_grpc_bridge_filter.rst
```diff
@@ -0,0 +1,11 @@
+.. _config_http_filters_connect_grpc_bridge:
```
mattklein123, line 1: Can you also update relevant docs here: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/grpc

And ref link back to this filter where appropriate.

test/extensions/filters/http/connect_grpc_bridge/connect_grpc_bridge_integration_test.cc
```diff
@@ -0,0 +1,73 @@
+#include "test/integration/http_integration.h"
+
+#include "gtest/gtest.h"
+
+namespace Envoy {
+namespace {
+
+class ConnectIntegrationTest : public testing::TestWithParam<Network::Address::IpVersion>,
```
mattklein123, line 8: Can you have an integration test for the actual end to end flow for both unary and streaming?

In general can you also check the code coverage report and make sure you are reasonably covering all of the code?

mattklein123 (changes requested):
Thanks at a high level this LGTM. Given the alpha status I think it's fine to ship and iterate with some real customers.

Note that if we want to call this eventually robust to untrusted data we should have some limited fuzzing of all of the relevant parsing code. This can be done later.

Thank you!

/wait

api/envoy/extensions/filters/http/grpc_stats/v3/config.proto
jchadwick-buf: Done.

docs/root/configuration/http/http_filters/connect_grpc_bridge_filter.rst
jchadwick-buf: I added the bridge to the list, is that what you mean? Or should it have a section just for connect?

test/extensions/filters/http/connect_grpc_bridge/connect_grpc_bridge_integration_test.cc
jchadwick-buf: Done.

I did try to ensure the important cases are covered, and we have >96% on everything. We can aim for 100% if you would prefer though. I think the main tricky case is trying to find a way to get WKT JSON encoding to fail.

mattklein123 (approved):
Great, thanks!

jchadwick-buf:
Thanks! Github says workflows need approval still. Did I need to do something there?

PR #24837, opened 2023-01-09
ci: run all mobile CI jobs on `main`

jpsim:
Commit Message: ci: run all mobile CI jobs on `main`
Additional Description: This will give us more visibility when changes that break Envoy Mobile land in Envoy. Slack discussion: https://envoyproxy.slack.com/archives/C02QMNG92A3/p1673288665957089
Risk Level: Moderate on CI stability, low on the Envoy libraries themselves
Testing: CI, these jobs have already been running for nearly two months on PRs with mobile/deps changes
Docs Changes: None
Release Notes: None
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/tools/should_run_ci.sh
+++ b/mobile/tools/should_run_ci.sh
@@ -22,13 +22,12 @@ function failure() {
   echo "run_ci_job=false" >> "$GITHUB_OUTPUT"
 }
 
-# TODO(jpsim): Consider enabling this if the load on EngFlow is ok
-# if [[ $branch_name == "main" ]]; then
-#   # Run all mobile CI jobs on `main`
-#   echo "Running $job because current branch is main"
-#   echo "run_ci_job=true" >> "$GITHUB_OUTPUT"
-#   exit 0
-# fi
+if [[ $branch_name == "main" ]]; then
+  # Run all mobile CI jobs on `main`
+  echo "Running $job because current branch is main"
+  echo "run_ci_job=true" >> "$GITHUB_OUTPUT"
+  exit 0
+fi
 
 base_commit="$(git merge-base origin/main HEAD)"
 changed_files="$(git diff "$base_commit" --name-only)"


Augustyniak (approved):
Looks good to me, let's make sure (per what you said already) that EngFlow is OK with this.

alyssawilk (approved):
I think if we're going to make the envoy CI jobs always run and be required we should do some deflaking.  I still have to retest in almost all of my Mobile CI PRs so I have very low confidence of a failed run meaning a problem for E-M.
I'm fine submitting this PR, and asking the maintainers and community to send flakes towards E-M dev, and we can force-require CIs as they're considered stable enough.  SG?

Augustyniak:
>I'm fine submitting this PR, and asking the maintainers and community to send flakes towards E-M dev, and we can force-require CIs as they're considered stable enough. SG?

Sounds good. From my personal experience the flakiness is more because of issues with the state of CI machines (fetching necessary resources on EngFlow side, spinning up CI machines in general) rather than issues with the implementation of individual tests but it's all just anecdotal and I definitely saw both of these categories of issues. I think that it's fair to say that we should invest into making them more solid.

Augustyniak:
speaking of which... 2 Android CI jobs in this PR seems to have failed due to timing out:
Screenshot 2023-01-09 at 2 54 39 PM

jpsim:
> speaking of which... 2 Android CI jobs in this PR seems to have failed due to timing out:

From what I can tell in this case, they just took a really long time to build, maybe the CI machines were over-scheduled.

We could bump the timeouts from 50 minutes to 90 minutes, but I'm actually hoping that by running these CI jobs more frequently, CI builds will have increased cache hits and might actually get faster.

> I'm fine submitting this PR, and asking the maintainers and community to send flakes towards E-M dev, and we can force-require CIs as they're considered stable enough. SG?

This PR doesn't make the EM CI jobs run for every PR, but rather for every commit on `main` (i.e. after a PR is merged).

Scanning the commits on `main` today, I see that most commits have a failed CI job, which in most cases is due to a failed Azure Pipelines job, not a failed GitHub Actions job.

image

So I don't think that enabling EM CI jobs on `main` will really change much for PR authors, but it will be a useful tool for EM maintainers to see if there are any `main` commits that broke EM.

PR #24840, opened 2023-01-09
Enhance local close details

### LINKED ISSUE #24283: net: downstream_local_disconnect documentation is vague on cause
*Description*:
> We have seen customers that have long-running http requests close unexpectedly, with a RST sent to the upstream. This is often associated with the error-code "downstream_local_disconnect" in the access log. The doc for this error is "“The client connection was locally closed for an unspecified reason.” 

> That description is not very actionable to end-users. This bug tracks efforts to add more logging detail around this error. I think this involves finding the code-paths where Network::ConnectionEvent::LocalClose is populated, and walking back the call-stack to understand why. A superficial glance tells us things like `HttpUpstream::onResetStream` was called or `doneReading()` was called, or HttpUpstream was destructed with reset_encoder_ non-null. However we probably need another few levels of causality there to relate it to something understandable from outside the details of Envoy.

> Per suggestion from @alyssawilk I added
```
    ASSERT(event == Network::ConnectionEvent::RemoteClose, absl::StrCat("event=", event));
```
to ConnectionManagerImpl::onEvent and ran tests. Lots of them fail; I think most or all were integration tests. This needs deeper investigation.

@mattklein123 @KBaichoo FYI

KBaichoo:
Commit Message: Enhance visibility in local close.
Additional Description:
Risk Level: medium (envoy bug asserting local close reason provided; if there are edge cases tests miss might trigger)
Testing: ran tests
Docs Changes: included
Release Notes: included
 Fixes #24283


source/common/http/conn_manager_impl.cc
```diff
@@ -188,8 +188,9 @@ void ConnectionManagerImpl::checkForDeferredClose(bool skip_delay_close) {
     close = Network::ConnectionCloseType::FlushWrite;
   }
   if (drain_state_ == DrainState::Closing && streams_.empty() && !codec_->wantsToWrite()) {
-    doConnectionClose(close, absl::nullopt,
-                      StreamInfo::ResponseCodeDetails::get().DownstreamLocalDisconnect);
+    // We are closing a draining connection with no active streams and the codec has
+    // nothing to write. As such the details are unnecessary.
+    doConnectionClose(close, absl::nullopt, "deferred_close_on_drained_connection");
```
yanavlasov, line 193: I think it makes sense to crate a singleton with constants for these reasons, since it will b easier to see all possible reason strings.

yanavlasov (changes requested):
/wait

source/common/http/conn_manager_impl.cc
KBaichoo: This list will be non-complete as close can dependent on dependencies i.e. QUICHE  which is why I originally didn't opt for this.  Happy to do as a follow up if you feel strongly about this.

yanavlasov (approved):

Issue #24841, opened 2023-01-09
"Start simulator" phase occasionally times out after 50 minutes


RyanTheOptimist:
We've noticed that CI sometimes times out waiting for the simulator to start. For example:

https://github.com/envoyproxy/envoy/actions/runs/3876678545/jobs/6612426608

It's possible there is a race-condition in the logic we're using to detect if the emulator is really started and so we miss the notification.

[Context from https://envoyproxy.slack.com/archives/C02F93EEJCE/p1671723837976249]

PR #24842, opened 2023-01-09
Wait at most 5 minutes for the emulator to start

RyanTheOptimist:
Wait at most 5 minutes for the emulator to start.  A more correct solution would be to find a better way to determine if the emulator is actually booted, but hopefully this fails faster, at least.

Bug: #24841

Risk Level: Low - CI only
Testing: CI
Docs Changes: N/A
Release Notes: N/A


RyanTheOptimist:
@Augustyniak Thanks for the suggestion!

jpsim:
Sorry I didn't get to reviewing this in time. Here's an alternative which only considers the boot time in the timeout (not the SDK installation and emulator creation time) and "fails open": https://github.com/envoyproxy/envoy/pull/24850

Issue #24007, opened 2022-11-15
stats: remove temporary cast operator added to Scope to help land #23851


jmarantz:
To address https://github.com/envoyproxy/envoy/issues/23862 , #23851 removes an inheritance, which affects a fairly large number of call-sites.

To make #23851 reviewable, a horrible implicit cast `operator Scope&` was added to Stats::Store. This keeps keeps a lot of noisy changes out of that PR, but leaves behind tech-debt. This bug tracks that tech-debt, which should be resolvable in a huge PR with no semantic changes, after #23851 lands.

PR #24843, opened 2023-01-10
stats: Stats remove operator scope hack phase 2 (see #24567)

jmarantz:
Commit Message: Second of 3 PRs that will remove dependence on Store::operator Scope&() -- the final PR will remove that operator.

Additional Description: This is a partial fix for https://github.com/envoyproxy/envoy/issues/24007 , a piece of technical debt that was taken in https://github.com/envoyproxy/envoy/pull/23851 and partially resolved in https://github.com/envoyproxy/envoy/pull/24567 in order to solve a structural problem where Stats::Store was a subclass of Stats::Scope, which didn't make OO sense (a Store is not a Scope), but was convenient (you could pass a Store into a function that was expecting a Scope), and the Store worked as a root Scope. That needed to be cleaned up to clarify the ownership model where Store is typically instantiated directly in a production factory or test, and Scope is always held by shared_ptr.

Risk Level: low -- this is a very large PR but it's just trivial one-line changes, getting the root scope explicitly from the store in a 100+ different places. There will be another smaller PR after this one. There are no functional changes or refactors in this PR.

With this PR, //test/... passes with the `operator Scope&` in `Stats::Store` commented out, but there are still dependencies on it remaining in //contrib/..., which will be dealt with in phase 3.
Testing: //test/...
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/envoy/stats/store.h
+++ b/envoy/stats/store.h
@@ -160,7 +160,7 @@ public:
   // accessing the rootScope() to call these methods. The first in the series is
   // https://github.com/envoyproxy/envoy/pull/24567 which just takes care of
   // test/common/...
-  // operator Scope&() { return *rootScope(); }
+  operator Scope&() { return *rootScope(); }
 
   // Delegate some methods to the root scope; these are exposed to make it more
   // convenient to use stats_macros.h. We may consider dropping them if desired,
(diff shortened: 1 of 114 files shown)


ggreenway (approved):

jmarantz:
Thanks Greg!

PR #24844, opened 2023-01-10
mobile: disable static extension registration

jpsim:
When building with certain release optimizations, aggressive deadcode stripping causes these statics to be removed, causing extensions to fail to be found at runtime when they're attempted to be used, often causing a crash or even worse some functionality being silently disabled.

By disabling these statics completely for Envoy Mobile, extensions will fail to resolve unless registered by explicitly calling the `forceRegister` functions at runtime.

This will make Envoy Mobile's behavior more deterministic and less prone to breaking based on the level of dead code stripping.

Commit Message: mobile: disable static extension registration
Additional Description:
Risk Level: Moderate for Envoy, high for Envoy Mobile. Mitigated by using a `--define` flag.
Testing: The iOS async/await example app & CI.
Docs Changes: None.
Release Notes: Added.
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim:
I plan on splitting this PR into a few separate ones to make it easier to review.

wbpcode:
cc @alyssawilk

jpsim:
I split up this PR into two more focused ones that should be no-ops to make it easier to review:

* https://github.com/envoyproxy/envoy/pull/24864
* https://github.com/envoyproxy/envoy/pull/24865

Once that's done, then the changes to actually disable static extension factory registration are relatively small and focused:

image

image

So I'll be asking reviewers to first review #24864 and #24865, then I'll merge `main` into this PR and ask for the final piece to be reviewed.

alyssawilk:
Hm, so I think I like this idea overall but I want to call out it has a pretty high risk of causing crashes on the data plane for anyone who uses it so I'm inclined to think it should be CI-only and come with warnings that it should not be used in production builds.  There's extensions like the HTTP matchers which could easily be turned on by config, might not be currently culled by dead stripping, and would then crash on a query which triggerd those matchers.  WDYT?

jpsim:
> I'm inclined to think it should be CI-only and come with warnings that it should not be used in production builds

Yes that's definitely an option I was also considering, although I was more thinking that we could only enable this in debug builds, which is mostly what CI uses.

mobile/test/swift/integration/StatFlushIntegrationTest.swift
```diff
@@ -1,8 +1,14 @@
 import Envoy
 import Foundation
+import TestExtensions
 import XCTest
 
 final class StatFlushIntegrationTest: XCTestCase {
+  override static func setUp() {
+    super.setUp()
+    register_test_extensions()
```
jpsim, line 9: If this is too noisy, I can easily move this to a base test class that all Swift integration tests inherit from instead of `XCTestCase`.

jpsim:
@Augustyniak @alyssawilk this is ready for review now

mobile/.bazelrc
```diff
@@ -67,6 +68,10 @@ build:dbg-android --config=android
 build:dbg-ios --config=dbg-common
 build:dbg-ios --config=ios
 
+# Default flags for Android tests
+# TODO(jpsim): Explicitly register test extensions for Android tests
+build:android-test --define=static_extension_registration=enabled
```
Augustyniak, line 73: Most of our configs use the naming convention where the name of the platform is the suffix: release-android, release-common, dbg-android etc. Maybe we should call this new config `test-android` or `ci-test-android`? 

 While on this, could we just use add `static_extension_registration=enabled` to `dbg-android` and use `dbg-android` in tests? Why did we decided to add a separate config in here?

Augustyniak: Do we have an issue that tracks an effort of disabling `static_extension_registration` in Android tests?

Augustyniak (approved):

PR #24845, opened 2023-01-10
ext_authz: allow headers propagation when allow_debugging_failures is set

agrawroh:
## Background

Currently, the `ext_authz` filter [[Link]](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto) doesn't propagate the headers sent from the upstream service back to the client in case of errors. These errors could be useful for debugging purposes. (See #14001)

## Changes

This PR adds a new boolean field called `allow_debugging_failures` in `ext_authz` filter which could be used to control whether or not to propagate the response headers received from the upstream service in case of 5XX errors. It's defaulted to `false` to maintain the current behavior.

**Commit Message:** Allow upstream headers propagation when `allow_debugging_failures` is set for ext_authz filter.
**Additional Description:** Adds a boolean field in ext_authz filter to control whether or not the response headers received from upstream should be propagated back to the client in case of errors.
**Risk Level:** Low
**Testing:** Unit Tests
**Docs Changes:** Added description of this new field in the docs.
**Release Notes:** Added


source/extensions/filters/http/ext_authz/ext_authz.cc
```diff
@@ -418,9 +418,40 @@ void Filter::onComplete(Filters::Common::ExtAuthz::ResponsePtr&& response) {
...
+              // will override existing headers.
+              for (const auto& header : headers) {
+                response_headers.remove(header.first);
+              }
+              // Then set all the requested headers, allowing the same header to be set multiple
+              // times, e.g. `Set-Cookie`.
+              for (const auto& header : headers) {
+                ENVOY_STREAM_LOG(trace, " '{}':'{}'", callbacks, header.first.get(), header.second);
+                response_headers.addCopy(header.first, header.second);
+              }
```
wbpcode, line 447: may be `setCopy` is enough?

wbpcode:
cc @mattklein123 for API change and quick review to this design.

api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
```diff
@@ -63,6 +63,16 @@ message ExtAuthz {
...
 
+  //  This changes the filter's behaviour on errors:
+  //
+  //  1. When set to true, the filter would retain the status code received from the upstream and
+  //  would also propagate the response body and headers back to the client (if available) for
+  //  further debugging.
+  //
+  //  2. When set to false, the filter would return a 403 or `status_on_error` (if set) and would
+  //  not propagate any response body or headers back to the client. This is the default behavior.
+  bool allow_debugging_failures = 18;
```
mattklein123, line 74: I think I would name this `propagate_response_on_failure` ?

Also, is this feature HTTP specific as it seems to pivot on 5xx? Should this actually be part of the HttpService object vs. global?

mattklein123 (changes requested):
High level question/comment to get started. Thank you!

/wait

api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto
ggreenway: What about adding a field analogous to `allowed_headers` for `allowed_response_headers`?

wbpcode: > Also, is this feature HTTP specific as it seems to pivot on 5xx? Should this actually be part of the HttpService object vs. global?

Sounds reasonable.

agrawroh: Thanks, @mattklein123. That sounds reasonable to me as well.

@ggreenway Sorry, I'm not able to understand the intention behind adding `allowed_response_header`. Is the idea to limit what could be propagated back to the client when there are failures? We expose [two lists](https://github.com/envoyproxy/envoy/blob/e802d9ee9ac2c355255b37afcf5c85e3e1fc82ee/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L305) `allowed_client_headers` and `allowed_upstream_headers` today which serves similar purpose.

source/extensions/filters/http/ext_authz/ext_authz.cc
agrawroh: Changed to `setCopy` as suggested. Thanks!

changelogs/current.yaml
```diff
@@ -306,6 +306,10 @@ new_features:
     added support to allowlist headers included in the check request to gRPC authorization server (previously only available for HTTP authorization server).
     Pre-existing field :ref:`allowed_headers <envoy_v3_api_field_extensions.filters.http.ext_authz.v3.AuthorizationRequest.allowed_headers>` is deprecated in favour
     of the new field :ref:`allowed_headers <envoy_v3_api_field_extensions.filters.http.ext_authz.v3.ExtAuthz.allowed_headers>`.
+- area: ext_authz
+  change: |
+    added a new field :ref:`propagate_response_on_failure <envoy_v3_api_field_extensions.filters.http.ext_authz.v3.ExtAuthz.HttpService.propagate_response_on_failure>` to control whether or
```
wbpcode, line 311: May be the label should be `envoy_v3_api_field_extensions.filters.http.ext_authz.v3.HttpService.propagate_response_on_failure` to fix the docs building error?

agrawroh: Thank You

source/extensions/filters/http/ext_authz/ext_authz.cc
```diff
@@ -418,9 +418,52 @@ void Filter::onComplete(Filters::Common::ExtAuthz::ResponsePtr&& response) {
           *decoder_callbacks_, enumToInt(config_->statusOnError()));
       decoder_callbacks_->streamInfo().setResponseFlag(
           StreamInfo::ResponseFlag::UnauthorizedExternalService);
-      decoder_callbacks_->sendLocalReply(
-          config_->statusOnError(), EMPTY_STRING, nullptr, absl::nullopt,
-          Filters::Common::ExtAuthz::ResponseCodeDetails::get().AuthzError);
+      if (config_->propagateResponseOnFailure()) {
```
mattklein123, line 421: It seems odd to me that this logic is split across multiple files and that there is additional logic here for things like stats. Why do we additionally increment stats here? Also, if the header processing is done in the other file can't this file just stay a unified path?

mattklein123 (changes requested):
Looks like this is on the right track. A few comments, thanks.

/wait

PR #24850, opened 2023-01-10
ci: wait up to 5 minutes for Android emulator to start

jpsim:
And if 5 minutes elapses, assume the emulator booted successfully and we missed the `sys.boot_completed` event.

Remove the `timeout-minutes` setting on the GitHub Actions job definitions so we only consider booting the emulator in the timeout and not installing the SDK or creating the emulator.

Bug: #24841

Risk Level: Low - CI only
Testing: CI
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -59,7 +59,6 @@ jobs:
     - name: 'Start simulator'
       if: steps.should_run.outputs.run_ci_job == 'true'
       run: cd mobile && ./ci/mac_start_emulator.sh
-      timeout-minutes: 5
     # Return to using:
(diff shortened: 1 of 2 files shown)


Augustyniak (approved):
will be curious to see how it plays out. If the emulator does not start within 5 minutes and we continue with the flow we will need to wait until we build the example app to learn that the app cannot be installed on an emulator as it's not running. 

So whether we want to fail open or fail close will depend on what we will happen in the case of the builds that end up timing out after 300s.

jpsim:
> So whether we want to fail open or fail close will depend on what we will happen in the case of the builds that end up timing out after 300s.

This change makes it so we wait 300s only for the emulator to boot. Nothing is building during this time. And sure, we can adjust this timeout as needed.

PR #24856, opened 2023-01-10
ci: Workaround disk space issues in release job

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/cleanup.sh
+++ b/.azure-pipelines/cleanup.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -e
+
+echo "Disk space before package removal"
+df -h
+
+# Temporary script to remove tools from Azure pipelines agent to create more disk space room.
+
+PURGE_PACKAGES=(
+    "adoptopenjdk-*"
+    "ant"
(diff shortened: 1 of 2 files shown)


RyanTheOptimist (approved):
Out of curiosity, what's the long-term solution here?

phlax:
> Out of curiosity, what's the long-term solution here?

not sure - i dont think we have control over the build image that is used (@lizan ?) - so we either need to be more conservative in our build or workaround i think

previously some deps were updated (iirc wasm-related) or similar that reduced the build size - but it happens for different reasons and with different problems (ie output binary size increases <> build env increase)

phlax:
https://dev.azure.com/cncf/envoy/_build/results?buildId=124933&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec&l=592

PR #24860, opened 2023-01-10
Add RTDS and ADS apis to mobile engine builder

caschoener:
Updating engine_builder to support rtds and ads configuration.  Currently ads always uses google_grpc, so updated the test flags to compile that in.


mobile/test/common/integration/base_client_integration_test.cc
```diff
@@ -194,9 +194,8 @@ void BaseClientIntegrationTest::createEnvoy() {
     }
   }
 
-  finalizeConfigWithPorts(config_helper_, ports, use_lds_);
-
   if (override_builder_config_) {
```
caschoener, line 197: `override_builder_config_` is always `false` now, so this is dead code.  But I remember Alyssa mentioned it might be good to leave it in in case we needed to use `override_builder_config_` again?

mobile/test/common/integration/xds_integration_test.cc
```diff
@@ -27,20 +72,17 @@ XdsIntegrationTest::XdsIntegrationTest() : BaseClientIntegrationTest(ipVersion()
...
-    bootstrap.mutable_static_resources()->add_clusters()->MergeFrom(
-        createSingleEndpointClusterConfig(std::string(XDS_CLUSTER)));
+    auto config_str = builder_.generateConfigStr();
+    TestUtility::loadFromYaml(absl::StrCat(config_header, config_str), bootstrap);
   });
+  // Set up the basic bootstrap config for xDS.
+  // We use a fake port at first, and assign a real one during initialization
+  builder_.setCustomClusters(
+      xdsClustersConfig(Network::Test::getLoopbackAddressString(ipVersion()), 0, 0));
+  builder_.disableStatsConfig(true);
```
caschoener, line 85: Reposting an unresolved comment from Alyssa in the other repo:

"do you know what stat counters we need for xDS? maybe we should we just add them into the default matchers now we're adding builder xDS support?"

Might need tips on how to do this

mobile/test/common/integration/rtds_integration_test.cc
```diff
@@ -72,8 +70,8 @@ TEST_P(RtdsIntegrationTest, RtdsReload) {
   EXPECT_EQ(cc_.on_complete_calls, 1);
   EXPECT_EQ(cc_.on_cancel_calls, 0);
   EXPECT_EQ(cc_.on_error_calls, 0);
-  EXPECT_EQ(cc_.on_header_consumed_bytes_from_response, 13);
-  EXPECT_EQ(cc_.on_complete_received_byte_count, 41);
+  EXPECT_EQ(cc_.on_header_consumed_bytes_from_response, 27);
+  EXPECT_EQ(cc_.on_complete_received_byte_count, 67);
```
caschoener, line 74: these changed because we switched to http 1 from https 2

mobile/test/common/integration/sds_integration_test.cc
```diff
@@ -19,9 +19,25 @@ const envoy::service::secret::v3::SdsDummy _sds_dummy;
 class SdsIntegrationTest : public XdsIntegrationTest {
 public:
   SdsIntegrationTest() {
+    override_builder_config_ = true;
```
caschoener, line 22: moving some of the shared code from xds to here because the rtds test has switched to using the builder API

caschoener:
Think I will wait to add APIs for the other languages until someone says this looks good btw

mobile/library/common/config/config.cc
```diff
@@ -498,6 +498,8 @@ stats_sinks: *stats_sinks
       patterns:
         - prefix: cluster.base.
         - prefix: cluster.base_h2.
+        - prefix: sds.
+        - prefix: runtime.
```
abeyad, line 502: Nice!

mobile/library/common/config/config.cc
```diff
@@ -538,5 +540,44 @@ stats_sinks: *stats_sinks
...
+        name: some_rtds_layer
+        rtds_config:
+          initial_fetch_timeout:
+            seconds: 1
+          resource_api_version: V3
+          api_config_source:
+            api_type: {}
+            transport_api_version: V3
+            grpc_services:
+              envoy_grpc:
```
abeyad, line 562: While Envoy Mobile has google_grpc disabled by default, we will need to enable it for the Traffic Director testing amongst other use cases (some users may want to enable google_grpc for the same reasons).  So we probably also want to make this configurable through the APIs, or provide a separate API, e.g. addGoogleGrpcXdsCluster vs. addEnvoyGrpcXdsCluster.

mobile/library/cc/engine_builder.cc
```diff
@@ -189,6 +189,18 @@ EngineBuilder& EngineBuilder::enableH2ExtendKeepaliveTimeout(bool h2_extend_keep
   return *this;
 }
 
+EngineBuilder& EngineBuilder::addXdsCluster(const std::string& address, int port) {
+  custom_clusters_ = fmt::format(rtds_layer_insert, address, port);
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::addRtdsLayer(const std::string& api_type,
+                                           const std::string xds_cluster) {
```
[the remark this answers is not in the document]
caschoener: Isn't everything const already? Guessing maybe you meant the int port?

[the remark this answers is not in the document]
RyanTheOptimist: I think "const std::string xds_cluster" => "const std::string& xds_cluster".

mobile/library/common/config/config.cc
```diff
@@ -538,5 +540,44 @@ stats_sinks: *stats_sinks
 R"(
         overload:
           global_downstream_max_connections: 0xffffffff # uint32 max
+#{custom_layers}
 )";
 // clang-format on
+
+const char* xds_cluster_insert = R"(
+    - name: some_static_layer
```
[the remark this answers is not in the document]
caschoener: Sorry. I mixed up the variable names for the rtds and xds insert lol.  I think this comment still applies though, seems like we don't want this random some_static_layer in production? Maybe the best thing to do would be to change the test to only use dynamic layers:
1. send an RTDS request to set test_feature to True, then verify it's true
2. send RTDS request to set test_feature to False, then verify

mobile/library/cc/engine_builder.h
```diff
@@ -61,8 +61,13 @@ class EngineBuilder {
   EngineBuilder& enableH2ExtendKeepaliveTimeout(bool h2_extend_keepalive_timeout_on);
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
-
+  // Creates an XDS cluster with a single static endpoint, where the endpoint is intended to be of
+  // a fake upstream on the loopback address.
+  EngineBuilder& addXdsCluster(const std::string& address, int port);
```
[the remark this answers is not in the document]
caschoener: Does adding a configurable name solve this? It does seem to me that some use cases might want to share xds clusters, or create new one per xds service.  Not sure if that's the case though.

mobile/library/common/config/config.cc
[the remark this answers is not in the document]
abeyad: is the variable only used in RTDS tests?  if so, i think it makes sense to move the flag to the dynamic layer.  in any case, we should consider adding a static runtime layer API to the engine builder.

mobile/library/common/config/templates.h
```diff
@@ -102,3 +102,13 @@ extern const char* default_cert_validation_context_template;
...
+
+/**
+ * Config template for an RTDS layer
+ */
+extern const char* rtds_layer_insert;
+
+/**
+ * Config template for an XDS cluster
+ */
+extern const char* xds_cluster_insert;
```
abeyad, line 114: i think we probably want to change this name to `cluster_insert` and the builder API to `addCluster`

abeyad (commented):
I think this is looking more along the lines of what I was thinking!  Let's see what Alyssa thinks.  Thanks for the iterations on it.

mobile/library/common/config/config.cc
alyssawilk: I'm inclined to get Lyft's opinion on this but as currently we only have one use case for xDS which uses google's gRPC, I'd be inclined to as we add API docs, add an API doc saying that addWhateverWeCallThis requires building with --build_flags_for_google_grpc.  when/if folks want to do xDS with non-google-gRPC we could add a boolean to the function indicating which client should be used.

alyssawilk (commented):
Sweet!  Some thoughts from my side since Ali's out

mobile/library/cc/engine_builder.h
```diff
@@ -61,7 +61,10 @@ class EngineBuilder {
   EngineBuilder& enableH2ExtendKeepaliveTimeout(bool h2_extend_keepalive_timeout_on);
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
-
+  // Creates an XDS cluster with a single static endpoint
+  EngineBuilder& addCluster(const std::string& name, const std::string& address, int port);
```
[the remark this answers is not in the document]
alyssawilk: oh, or update the comment if the insert is generic and not xds specific :-)

mobile/library/common/config/config.cc
Augustyniak: How many stats does this effectively enable - as in, how many more are tracked by Envoy after that change (we can just look at the number of active stats after we open the example app I guess)? (also, you may need to rebase as `patterns:` were updated in main)

mobile/library/common/config/config.cc
Augustyniak: The solution proposed by Alyssa sounds good to me. It's fine to keep xds coupled with google grpc client as long as it's well documented behavior.

mobile/library/cc/engine_builder.h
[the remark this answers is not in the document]
abeyad: +1, the API is adding a generic cluster, it just so happens we'll use it in the `addRtdsLayer` API as well as other ones added in the future.  So I think `addCluster` makes the most sense.  Comment needs to be updated as Alyssa mentioned.

mobile/library/common/config/config.cc
abeyad: That sounds fine to me as well, but then, we'd have to update the integration tests to use google_grpc instead

caschoener: Working on switching to google_grpc now

caschoener:
cool here's a new version.  Enabled  google_grpc for non-production use-cases in the bazelrc.  Switched to google_grpc in the test, so there's no longer an xds cluster. Switched from a test-specific static layer to just read/writing one of the existing params. 

Changes can be seen here https://github.com/envoyproxy/envoy/pull/24860/files/436d1bebe152c80326a6c0ee4d3c46d236e338ea..8a2a68dfa9bba5b86d809ffc86f0dd0af5a495f7

caschoener:
Latest commit adds the kotlin API.  I realized it's simplest to store all the params in the builder and then do the formatting later so let's go with that.

mobile/library/cc/engine_builder.cc
```diff
@@ -303,6 +313,13 @@ std::string EngineBuilder::generateConfigStr() const {
     insertCustomFilter(filter_config, config_template);
   }
 
+  if (!rtds_layer_name_.empty()) {
+    std::string custom_layers_ = fmt::format(rtds_layer_insert, rtds_layer_name_, rtds_layer_name_,
```
abeyad, line 317: trailing underscores on variable names are only meant for class members, but see note above about making custom_layers_ the class variable instead of a local one

mobile/library/common/config/config.cc
```diff
@@ -538,5 +540,41 @@ stats_sinks: *stats_sinks
...
+          resource_api_version: V3
+          api_config_source:
+            api_type: {}
+            transport_api_version: V3
+            grpc_services:
+              google_grpc:
+                target_uri: "{}:{}"
+            set_node_on_first_message_only: true)";
+            
+const char* cluster_insert = R"(
```
abeyad, line 562: don't think we're using this anymore; remove

mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -571,6 +571,27 @@ open class EngineBuilder(
     return this
   }
 
+  /**
+   * Add a google_grpc RTDS layer to config
+   *
+   * @param layer_name the layer name.
```
abeyad, line 577: don't think we need the extra line of whitespace between each "@param"

mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -571,6 +571,27 @@ open class EngineBuilder(
...
+   *
+   * @return this builder.
+   */
+  fun addRtdsLayer(layer_name: String, api_type: String, address: String, port: Int): EngineBuilder {
+    this.rtds_layer_name_ = layer_name
+    this.rtds_api_type_ = api_type
+    this.rtds_address_ = address
+    this.rtds_port_ = port
+    return this
+  }
```
abeyad, line 593: what is this doing exactly? the Kotlin functions would need to use a bridge to call the C++ APIs, no?  for starters, i think it needs to be passed into https://sourcegraph.com/github.com/envoyproxy/envoy@ec9099786796da6f834a6d562d0c3939c342a5e1/-/blob/mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt?L583, which would then go through some other layers till it's initializing the Engine creation.

Also, I don't think you declared the rtds_* variables in the Kotlin file, so this probably won't compile.

mobile/library/cc/engine_builder.cc
```diff
@@ -189,6 +189,16 @@ EngineBuilder& EngineBuilder::enableH2ExtendKeepaliveTimeout(bool h2_extend_keep
   return *this;
 }
 
+EngineBuilder& EngineBuilder::addRtdsLayer(const std::string& layer_name,
+                                           const std::string& api_type, const std::string& address,
+                                           const int port) {
+  rtds_layer_name_ = layer_name;
```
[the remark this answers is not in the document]
caschoener: It seems like this would complicate the API for the other languages because I'd have to do a fmt::format equivalent in each language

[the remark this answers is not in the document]
caschoener: I suppose the more I look at it it's going to be very tedious either way so IDK

caschoener:
After talking it over I think adding all the languages in one diff will be a lot, so let's just ship this one as-is.

mobile/library/common/config/config.cc
abeyad: you didn't remove this?

mobile/library/common/config/config.cc
```diff
@@ -538,5 +540,41 @@ stats_sinks: *stats_sinks
...
+
+const char* rtds_layer_insert = R"(
+    - name: {}
+      rtds_layer:
+        name: {}
+        rtds_config:
+          initial_fetch_timeout:
+            seconds: 1
+          resource_api_version: V3
+          api_config_source:
```
abeyad, line 554: it just occurred to me that for Traffic Director, it only supports ADS.  So we will also want to support ads instead of api_config_source: https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/core/v3/config_source.proto#L225.  It would require setting the bootstrap's ads_config field: https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/bootstrap/v3/bootstrap.proto#L101

I think we can start with only supporting ADS, and add non-ADS support later.  This will just require moving a bunch of this config stuff to `bootstrap.dynamic_resources.ads_config` and have `rtds_config` just set the `ads` field.

mobile/library/cc/engine_builder.cc
```diff
@@ -189,6 +189,17 @@ EngineBuilder& EngineBuilder::enableH2ExtendKeepaliveTimeout(bool h2_extend_keep
   return *this;
 }
 
+EngineBuilder& EngineBuilder::addRtdsLayer(const std::string& layer_name) {
+  custom_layers_ = fmt::format(rtds_layer_insert, layer_name, layer_name);
+  return *this;
+}
+EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& api_type,
```
abeyad, line 196: `api_type` will always be `GRPC`, at least for now, so we can further simply it

mobile/library/common/config/templates.h
```diff
@@ -102,3 +102,18 @@ extern const char* default_cert_validation_context_template;
...
+
+/**
+ * Config template for an RTDS layer
+ */
+extern const char* rtds_layer_insert;
+
+/**
+ * Config template for a new cluster
+ */
+extern const char* cluster_insert;
```
abeyad, line 114: need to remove this too

mobile/library/cc/engine_builder.cc
```diff
@@ -303,6 +314,16 @@ std::string EngineBuilder::generateConfigStr() const {
     insertCustomFilter(filter_config, config_template);
   }
 
+  if (!custom_layers_.empty()) {
+    absl::StrReplaceAll({{"#{custom_layers}", absl::StrCat("#{custom_layers}\n", custom_layers_)}},
+                        &config_template);
+  }
+
+  if (!custom_ads_.empty()) {
+    absl::StrReplaceAll({{"#{custom_ads}", absl::StrCat("#{custom_ads}\n", custom_ads_)}},
```
abeyad, line 323: before setting custom_layers_, we should set custom_ads_, and then before setting custom_layers_, check that custom_ads_ is set, and if not, throw a std::runtime_error

mobile/library/cc/engine_builder.cc
caschoener: Ali changed his mind

abeyad (dismissed):

mobile/library/cc/engine_builder.cc
```diff
@@ -184,6 +184,22 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
   return *this;
 }
 
+EngineBuilder& EngineBuilder::enableH2ExtendKeepaliveTimeout(bool h2_extend_keepalive_timeout_on) {
```
abeyad, line 187: this shouldn't be added as part of this PR?

mobile/library/common/config/config.cc
```diff
@@ -491,6 +492,25 @@ R"(
     typed_extension_protocol_options: *h3_protocol_options
 stats_flush_interval: *stats_flush_interval
 stats_sinks: *stats_sinks
+stats_config:
+  stats_matcher:
+    inclusion_list:
+      patterns:
+        - prefix: cluster.base.
```
abeyad, line 499: seems like a lot more patterns were added, probably from a merge conflict?

mobile/library/cc/engine_builder.cc
```diff
@@ -302,6 +326,7 @@ std::string EngineBuilder::generateConfigStr() const {
   if (config_str.find("{{") != std::string::npos) {
     throw std::runtime_error("could not resolve all template keys in config");
   }
+  std::cout << config_str << std::endl;
```
abeyad, line 329: please remove

abeyad (approved):

abeyad:
Added @jpsim for cross-company review

mobile/test/common/integration/rtds_integration_test.cc
```diff
@@ -85,17 +68,16 @@ TEST_P(RtdsIntegrationTest, RtdsReload) {
...
       Config::TypeUrl::get().Runtime, {some_rtds_layer}, {some_rtds_layer}, {}, "1");
   // Wait until the RTDS updates from the DiscoveryResponse have been applied.
   ASSERT_TRUE(waitForCounterGe(load_success_counter, load_success_value + 1));
 
   // Verify that the Runtime config values are from the RTDS response.
-  EXPECT_FALSE(Runtime::runtimeFeatureEnabled("envoy.reloadable_features.test_feature_false"));
-  EXPECT_FALSE(Runtime::runtimeFeatureEnabled("envoy.reloadable_features.test_feature_true"));
+  EXPECT_TRUE(Runtime::runtimeFeatureEnabled(
+      "envoy.reloadable_features.skip_dns_lookup_for_proxied_requests"));
 }
```
abeyad, line 82: Can we also add a test to ensure we throw an error when addRtdsLayer is called without calling setAggregatedDiscoveryService?

mobile/library/cc/engine_builder.h
```diff
@@ -60,6 +60,10 @@ class EngineBuilder {
   EngineBuilder& enableDrainPostDnsRefresh(bool drain_post_dns_refresh_on);
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
+  // Adds an RTDS layer to default config
+  EngineBuilder& addRtdsLayer(const std::string& layer_name);
+  EngineBuilder& setAggregatedDiscoveryService(const std::string& api_type,
```
abeyad, line 65: Can we add a comment to this method?

mobile/library/cc/engine_builder.h
```diff
@@ -60,6 +60,10 @@ class EngineBuilder {
   EngineBuilder& enableDrainPostDnsRefresh(bool drain_post_dns_refresh_on);
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
+  // Adds an RTDS layer to default config
```
abeyad, line 63: Can we also add that setAggregatedDiscoveryService must be called before addRtdsLayer is called?

abeyad (dismissed):

abeyad (approved):

jpsim:
Just an FYI according to the [mobile_perf CI job](https://github.com/envoyproxy/envoy/actions/runs/4008291380) this increases the binary size by ~1.4%. Not a blocker for me though, but raising it because reducing binary size have been an active effort in the last few months.

jpsim (dismissed):
Do you have a ticket to port this over to the Swift/Kotlin engine builders?

RyanTheOptimist:
> Just an FYI according to the [mobile_perf CI job](https://github.com/envoyproxy/envoy/actions/runs/4008291380) this increases the binary size by ~1.4%. Not a blocker for me though, but raising it because reducing binary size have been an active effort in the last few months.

Hm. There's not a lot of new code in this PR, so I can't imagine that's causing a 1% binary size increase. Is this because of the new grpc dependency? I see that the bazelrc is change to only disable grpc in release mode. What config does the mobile_perf job use?

caschoener:
> Hm. There's not a lot of new code in this PR, so I can't imagine that's causing a 1% binary size increase. Is this because of the new grpc dependency? I see that the bazelrc is change to only disable grpc in release mode. What config does the mobile_perf job use?

Oh yeah, sorry. I resolved this in chats with some people.  Your guess is correct, mobile_perf was using the test build flags instead of the release ones.  I changed it to use release config in https://github.com/envoyproxy/envoy/pull/25189

caschoener:
I changed the sha56 to fix the error 
```
Error in download_and_extract: java.io.IOException: Error downloading [https://github.com/google/libprotobuf-mutator/archive/v1.1.tar.gz] to /build/tmp/_bazel_envoybuild/400406edc57d332f0b9b805d2b8e33a1/external/com_github_google_libprotobuf_mutator/temp10529668633102100589/v1.1.tar.gz: Checksum was 9b862857a725a07287dc2dd182f32b76b38b8129b38aad3e26ebceccd9d6c620 but wanted fd299fd72c5cf664259d9bd43a72cb74dc6a8b9604d107fe2d2e90885aeb7c16
```
Downloaded the file and confirmed it is the new sha56. Maybe we weren't actually downloading this file before, because google_grpc was previously compiled out in tests?

bazel/repository_locations.bzl
```diff
@@ -309,7 +309,7 @@ REPOSITORY_LOCATIONS_SPEC = dict(
         project_desc = "Library to randomly mutate protobuffers",
         project_url = "https://github.com/google/libprotobuf-mutator",
         version = "1.1",
-        sha256 = "fd299fd72c5cf664259d9bd43a72cb74dc6a8b9604d107fe2d2e90885aeb7c16",
+        sha256 = "9b862857a725a07287dc2dd182f32b76b38b8129b38aad3e26ebceccd9d6c620",
```
phlax, line 312: @caschoener this should not change if the version has not changed

caschoener: I'll change it back.  Was trying to fix a strange checksum error, thought maybe the sha56 was incorrect

phlax: if you require updated deps for your PR - perhaps try and update these independently - that is generally preferable from review/revert pov

caschoener: After talking with JP apparently github tarball checksums aren't stable so this might be a good change if it makes tests pass

phlax: (otherwise - you definitely do not want to change any hashes without changing the corresponding version)

phlax: > After talking with JP apparently github tarball checksums aren't stable so this might be a good change if it makes tests pass

they are stable otherwise we would not be able to hash them

caschoener: You may want to check the envoy mobile collaboration slack channel for discussion on this point

mobile/library/cc/engine_builder.cc
```diff
@@ -242,6 +244,23 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
...
+  return *this;
+}
+EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& api_type,
+                                                            const std::string& address,
+                                                            const int port) {
+  bootstrapIncompatible();
+#ifndef ENVOY_GOOGLE_GRPC
+  raise std::runtime_error("google_grpc must be enabled in bazel to use ADS")
+#endif
+      custom_ads_ = fmt::format(ads_insert, api_type, address, port);
```
jpsim, line 260: Nit: Indentation?

mobile/library/common/config/config.cc
```diff
@@ -538,5 +539,27 @@ stats_sinks: *stats_sinks
...
+#{custom_layers}
 )";
+
+const char* rtds_layer_insert = R"(
+    - name: {}
+      rtds_layer:
+        name: {}
+        rtds_config:
+          initial_fetch_timeout:
+            seconds: 1
```
jpsim, line 551: A 1 second timeout seems very aggressive considering real world mobile networking conditions.

I'm not familiar with the implications of this timeout though.

abeyad: agree, this should probably be 5s to start off with.  we can experiment with it as necessary.

caschoener: I'll make it configurable but I think it needs to be 1 second in tests because the tests depend on the initial fetch timing out

jpsim: Why not expose a builder function to configure this then?

caschoener: Just did! Sorry I wasn't clear haha

jpsim: Great, sorry I missed that.

abeyad (approved):

caschoener:
Alyssa suggests that I update to make compatible with the new C++ bootstrap builder

abeyad:
> Per Alyssa's comments, we'll want to change this to use the proto builder APIs

Ah I see you made the same comment above Chris

mobile/library/cc/engine_builder.cc
```diff
@@ -242,6 +244,23 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
...
+EngineBuilder& EngineBuilder::addRtdsLayer(const std::string& layer_name, int timeout_seconds) {
+  rtds_layer_name_ = layer_name;
+  rtds_timeout_seconds_ = timeout_seconds;
+  return *this;
+}
+EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& api_type,
+                                                            const std::string& address,
+                                                            const int port) {
+#ifndef ENVOY_GOOGLE_GRPC
+  raise std::runtime_error("google_grpc must be enabled in bazel to use ADS");
```
abeyad, line 256: what does `raise` do in C++?  shouldn't this be `throw`?  not familiar with `raise` in c++.

mobile/library/common/config/config.cc
```diff
@@ -538,5 +539,27 @@ stats_sinks: *stats_sinks
 R"(
         overload:
           global_downstream_max_connections: 0xffffffff # uint32 max
+#{custom_layers}
 )";
+
+const char* rtds_layer_insert = R"(
```
abeyad, line 545: i don't believe we need any of this anymore, right?

mobile/library/common/config/config.cc
```diff
@@ -538,5 +539,27 @@ stats_sinks: *stats_sinks
...
+    - name: {}
+      rtds_layer:
+        name: {}
+        rtds_config:
+          initial_fetch_timeout:
+            seconds: {}
+          resource_api_version: V3
+          ads: {{}})";
+
+const char* ads_insert = R"(
```
abeyad, line 555: same, i don't think we need this anymore, right?

mobile/library/cc/engine_builder.cc
```diff
@@ -366,6 +385,21 @@ std::string EngineBuilder::generateConfigStr() const {
     insertCustomFilter(filter_config, config_template);
   }
 
+  if (!rtds_layer_name_.empty() && ads_api_type_.empty()) {
+    throw std::runtime_error("ADS must be configured when using RTDS");
+  }
+  if (!rtds_layer_name_.empty()) {
```
abeyad, line 391: don't think we need L391 to L402 anymore, right?

mobile/library/cc/engine_builder.cc
caschoener: Ah good catch

mobile/library/common/config/config.cc
caschoener: Same reply, both builders are currently in use. We are using the yaml approach but verifying it matches against the protobuf.

abeyad (approved):

abeyad:
@caschoener You don't want to ever rebase, it loses the commit review history for those who are re-reviewing subsequent changes.  The only time rebase is allowed is if you've forgotten to sign your commit with `-s`, as rewriting that commit is the only way to circumvent that problem.  Otherwise, it should always be a merge instead of a rebase.

abeyad (approved):

jpsim (approved):

PR #24862, opened 2023-01-10
ci: fix flaky swift test

Augustyniak:
Commit Message: That test is flaky. It continues to fail due to a wrong order in which expectations are fulfilled - there appears to be some kind of a race condition in there. Making the problematic test weaker by making it not check the order in which expectations are fulfilled.
Additional Description:
Risk Level: Low
Testing: 
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim (approved):
I think this is reasonable and necessary. 👍

RyanTheOptimist:
Thanks for doing this!

PR #24864, opened 2023-01-11
mobile: register all Envoy Mobile extension factories explicitly

jpsim:
In preparation for disabling static extension factory registration in https://github.com/envoyproxy/envoy/pull/24844, start by explicitly force-registering the extension factories Envoy Mobile is currently using.

To do this, some of the extension factories that were previously only registered also need to be declared with `DECLARE_FACTORY` so its `forceRegister...` function can be made accessible externally.

This is effectively a no-op since all these extension factories should already be registered statically.

Commit Message: mobile: register all Envoy Mobile extension factories explicitly
Risk Level: Low for Envoy, moderate for Envoy Mobile
Testing: Existing unit & integration tests
Docs Changes: None
Release Notes: None
Platform Specific Features: Mobile-only


mobile/envoy_build_config/extension_registry.cc
```diff
@@ -38,70 +52,113 @@
...
 
 #ifdef ENVOY_ENABLE_QUIC
+  Quic::forceRegisterEnvoyDeterministicConnectionIdGeneratorConfigFactory();
+  Quic::forceRegisterEnvoyQuicCryptoServerStreamFactoryImpl();
+  Quic::forceRegisterEnvoyQuicProofSourceFactoryImpl();
+  Quic::forceRegisterQuicClientTransportSocketConfigFactory();
+  Quic::forceRegisterQuicHttpServerConnectionFactoryImpl();
   Quic::forceRegisterQuicServerTransportSocketConfigFactory();
 #endif
-
```
Augustyniak, line 94: nice clean up

mobile/test/common/BUILD
```diff
@@ -10,6 +10,7 @@ envoy_cc_test(
     repository = "@envoy",
     deps = [
         "//library/common:engine_common_lib",
+        "@envoy_build_config//:extension_registry",
```
Augustyniak, line 13: not OSS-related per se: we should be able to depend on it in our private envoy mobile wrapper repo and just call `registerFactories` from there too

source/common/http/match_delegate/config.cc
```diff
@@ -14,7 +14,7 @@ namespace Common {
 namespace Http {
 namespace MatchDelegate {
 
-namespace {
+namespace Factory {
```
Augustyniak, line 17: not sure whether this name makes sense as the class within it has a Factory suffix anyway but I do not have a strong opinion in here and I assume that this change was needed for some reason.

source/extensions/listener_managers/listener_manager/filter_chain_manager_impl.h
```diff
@@ -352,5 +353,10 @@ class FilterChainManagerImpl : public Network::FilterChainManager,
   // Index filter chains by name, used by the matcher actions.
   FilterChainsByName filter_chains_by_name_;
 };
+
+namespace FilterChain {
```
Augustyniak, line 357: is this to avoid some naming conflicts?

Augustyniak (commented):
Left some comments. Did _not_ verify whether the list of added registration calls is the minimal required set of registrations that's needed by Envoy Mobile.

source/common/http/match_delegate/config.cc
jpsim: I added this namespace name because I don't think it's possible to refer to anonymous namespaces externally, which is needed to call the `forceRegister...` function. Open to better names.

source/extensions/listener_managers/listener_manager/filter_chain_manager_impl.h
jpsim: I added this namespace name because I don't think it's possible to refer to anonymous namespaces externally, which is needed to call the `forceRegister...` function. Open to better names.

mobile/envoy_build_config/extension_registry.cc
```diff
@@ -38,70 +52,113 @@
...
+  Extensions::HttpFilters::BufferFilter::forceRegisterBufferFilterFactory();
+  Extensions::HttpFilters::Decompressor::forceRegisterDecompressorFilterFactory();
+  Extensions::HttpFilters::DynamicForwardProxy::forceRegisterDynamicForwardProxyFilterFactory();
+  Extensions::HttpFilters::LocalError::forceRegisterLocalErrorFilterFactory();
+  Extensions::HttpFilters::NetworkConfiguration::forceRegisterNetworkConfigurationFilterFactory();
+  Extensions::HttpFilters::PlatformBridge::forceRegisterPlatformBridgeFilterFactory();
+  Extensions::HttpFilters::RouteCacheReset::forceRegisterRouteCacheResetFilterFactory();
+  Extensions::HttpFilters::RouterFilter::forceRegisterRouterFilterConfig();
+  Extensions::HttpFilters::SocketTag::forceRegisterSocketTagFilterFactory();
+  Extensions::HttpFilters::TestAccessor::forceRegisterTestAccessorFilterFactory();
```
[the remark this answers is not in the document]
jpsim: Yes, it's only used for Swift tests right now though: https://github.com/envoyproxy/envoy/blob/main/mobile/test/swift/integration/test_extensions.cc

I'm not sure if it's problematic to register these extensions for non-tests, do you?

source/common/http/match_delegate/config.cc
Augustyniak: I thought that for unnamed namespaces you could just omit them when referring to members declared as part of them. As in `Http::MatchDelegate::WHATEVER_YOU_WANT_TO_ACCESS` as opposed to `Http::MatchDelegate::Factory::WHATEVER_YOU_WANT_TO_ACCESS`

jpsim: Here's the compilation failure I get when trying to access things this way:

```
external/envoy/source/common/http/match_delegate/config.h:109:1: error: function 'Envoy::Common::Http::MatchDelegate::(anonymous namespace)::forceRegisterSkipActionFactory' has internal linkage but is not defined [-Werror,-Wundefined-internal]
DECLARE_FACTORY(SkipActionFactory);
^
external/envoy/envoy/registry/registry.h:660:61: note: expanded from macro 'DECLARE_FACTORY'
#define DECLARE_FACTORY(FACTORY) ABSL_ATTRIBUTE_UNUSED void forceRegister##FACTORY()
                                                            ^
<scratch space>:25:1: note: expanded from here
forceRegisterSkipActionFactory
^
external/envoy_build_config/extension_registry.cc:68:32: note: used here
  Common::Http::MatchDelegate::forceRegisterSkipActionFactory();
                               ^
1 error generated.
```

mobile/envoy_build_config/extension_registry.cc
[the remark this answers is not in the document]
jpsim: Ah yes, this does increase binary size by ~1.9% according to the `mobile_perf` CI jobs. Let me see if I can only register these for tests in this PR to avoid any binary size increase.

source/common/http/match_delegate/config.cc
Augustyniak: Got it, thank you for the explanation & error message.

jpsim:
I’ve removed the test extensions from the prod build so there’s no longer any binary size impact from this PR.

3EB80E7F-3045-4C6E-A959-5EC67A92B021

abeyad (approved):
Awesome, and it seems like there was no binary size change either!

Augustyniak:
> I’ve removed the test extensions from the prod build so there’s no longer any binary size impact from this PR.

Great. Could you just confirm what's the state of things now? As in, how were we able to move extensions out of the production builds - were we able to accomplish that by indirectly limiting the scope of the changes to iOS?

jpsim:
> were we able to accomplish that by indirectly limiting the scope of the changes to iOS?

The change to disable automatic registration of extension factories is coming in https://github.com/envoyproxy/envoy/pull/24844

I'll see where I can enable that in that PR.

PR #24865, opened 2023-01-11
extensions: split extension registration macro

jpsim:
In preparation for disabling static extension factory registration in https://github.com/envoyproxy/envoy/pull/24844, split the `REGISTER_FACTORY` macro into two, one that doesn't take a deprecated name and one that does, called `LEGACY_REGISTER_FACTORY`.

This is done because the previous approach of appending a `{deprecated_name}` suffix at the usage site is incompatible with the approach taken by https://github.com/envoyproxy/envoy/pull/24844 to not statically register the extension factories.

Commit Message: extensions: split extension registration macro
Risk Level: Low
Testing: Existing unit tests
Docs Changes: None
Release Notes: Inline comments

### DIFF
--- a/envoy/registry/registry.h
+++ b/envoy/registry/registry.h
@@ -497,7 +497,7 @@ private:
  * unit. For an example of a typical use case, @see NamedNetworkFilterConfigFactory.
  *
  * Example registration: REGISTER_FACTORY(SpecificFactory, BaseFactory);
- *                       REGISTER_FACTORY(SpecificFactory, BaseFactory){"deprecated_name"};
+ *                       REGISTER_FACTORY_D(SpecificFactory, BaseFactory, "deprecated_name");
  */
 template <class T, class Base> class RegisterFactory {
 public:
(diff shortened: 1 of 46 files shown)


envoy/registry/registry.h
```diff
@@ -618,6 +618,16 @@ template <class T, class Base> class RegisterInternalFactory {
   static Envoy::Registry::RegisterFactory</* NOLINT(fuchsia-statically-constructed-objects) */     \
                                           FACTORY, BASE>                                           \
       FACTORY##_registered
+/**
+ * Macro used for static registration with deprecated name.
+ */
+#define REGISTER_FACTORY_D(FACTORY, BASE, DEPRECATED_NAME)                                         \
```
alyssawilk, line 624: I'd lean towards a readable macro name, like LEGACY_REGISTER_FACTORY.
Alternately and maybe as a follow-up, can we get rid of these yet?  cc @mattklein123 
/wait

alyssawilk (approved):

PR #16793, opened 2021-06-03
general tracing context interface

wbpcode:
Signed-off-by: wbpcode <comems@msn.com>

Sub-PR of #16049. Check #16049 get more background information.

This PR designed a new generic abstraction `TraceContext` to replace `Http::RequestHeaderMap` to provide tracing context to the tracer driver. `Http::RequestHeaderMapImpl` already inherits from TraceContext and implements the relevant interfaces.

Next, we just need simply replace `Http::RequestHeaderMap` with `TraceContext` in all tracer drivers implementations. After that, the main body of the entire general tracing system is completed. I'm not sure, whether the replacement step requires a separate PR. Because it does not involve any new logic, but there will be a lot of file changes. 

Risk Level: Low.
Testing: Added.
Docs Changes: N/A
Release Notes:  N/A

PR #16244, opened 2021-04-30
remove trace drivers' dependency on HttpTracerImpl

wbpcode:
Signed-off-by: wbpcode <comems@msn.com>

This PR is part of #16049 to support general tracing. Please check #16049 get more details.

Commit Message: remove trace drivers' dependency on HttpTracerImpl
Additional Description:

Now all tracers (zipkin, skywalking, etc.) will depend on HttpTracerImpl, making it difficult for Tracers to be reused by other protocols (Dubbo, Thrift, etc.). The purpose of this PR is to change this dependency.

Risk Level: Low (No new logic).
Testing: Add.
Docs Changes: N/A
Release Notes: N/A

Issue #15784, opened 2021-03-31
general tracing support


wbpcode:
*Title*: *general tracing support*

*Description*:

Envoy provides powerful tracing capabilities. However, it is currently only available for the HTTP protocol. Because Tracer uses Http::RequestHeaderMap to pass the tracing context of the request.
While other protocols can implement an Http::RequestHeaderMap to directly reuse the existing Tracer, this requires a significant amount of work.
But if a simpler abstraction can be provided, then other protocols can more easily integrate the tracer into it. 
For example, simply implementing the following method:

```
class TraceableStream {
    virtual void setContext(const std::string_view key, const std::string_view value) PURE;
    virtual std::optional<std::string_view> getContext(const std::string_view key) const PURE;
};

```

And then create tracing span from it:

```
  /**
   * Start driver specific span.
   */
  virtual SpanPtr startSpan(const Config& config, TraceableStream& stream,
                            const std::string& operation_name, SystemTime start_time,
                            const Tracing::Decision tracing_decision) PURE;
```

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

mattklein123:
Yeah this is something I would love to clean up. Would definitely appreciate cleanup work here to make this code all generic.

wbpcode:
I can try to figure it out. Because I just want to enhance the tracing of the dubbo protocol. General tracing support is a very important pre-work.  🤔

wbpcode:
This is ok now.

PR #24866, opened 2023-01-11
tracing: make the tracer & tracer manager protocol independent

wbpcode:
Commit Message: tracing: make the tracer & tracer manager protocol independent
Additional Description:

By previous PRs (see #15784, #16244, #16793, #17212), We had made the tracing driver protocol independent. The non-HTTP proxy could re-use the all the tracing drivers, such as skywalking, zipkin, etc.

However, the `Tracer` and `TracerManager` is still HTTP-specific and the non-HTTP proxy still need to do lots of work to add complete tracing support. And this PR try to make it's easier to add tracing for non-HTTP proxy by:
- 1. make the `Tracer` and `TracerManager` protocol independent.
- 2. split related utility functions to protocol independent part and protocol specific part.

Risk Level: mid.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
It seems that no mainatainer is in charge of tracing. So, I can only assign this to you first. @mattklein123 🤣

mattklein123 (approved):
I looked at this from a refactor perspective, not a code change perspective. From that standpoint it LGTM. I assume there are no logic changes as part of this PR? Thank you.

PR #24871, opened 2023-01-11
ci: simplify the clean up of kotlin test

Augustyniak:
Commit Message: The change has a low chance of helping us with a flakiness of this test. https://github.com/envoyproxy/envoy/actions/runs/3894061911/jobs/6647516688
Additional Description:
Risk Level: Low, tests change only.
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/test/kotlin/integration/EngineApiTest.kt
+++ b/mobile/test/kotlin/integration/EngineApiTest.kt
@@ -8,26 +8,17 @@ import io.envoyproxy.envoymobile.engine.JniLibrary
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import org.assertj.core.api.Assertions.assertThat
-import org.junit.After
 import org.junit.Test
 
 class EngineApiTest {
-  private var engine: Engine? = null
-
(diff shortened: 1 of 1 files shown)


mobile/test/kotlin/integration/EngineApiTest.kt
```diff
@@ -8,36 +8,29 @@ import io.envoyproxy.envoymobile.engine.JniLibrary
...
-  fun teardown() {
-    engine?.terminate()
-    engine = null
-  }
-
   @Test
   fun `verify engine APIs`() {
     val countDownLatch = CountDownLatch(1)
-    engine = EngineBuilder()
+    val engine = EngineBuilder()
```
Augustyniak, line 21: I am not even sure whether this tests does anything useful since we do not call addGrpcDomain method on the engine so we are effectively not even enabling stats sink... will probably just disable it if it keeps being flaky.

jpsim (approved):

PR #24880, opened 2023-01-11
Envoy RingHash optimization using a Sharding algorithm

RamaMalladiAWS:
Commit Message: Envoy RingHash optimization using a Sharding algorithm
Additional Description: RingHash can be optimized by using a bucket/shard algorithm. This PR is an implementation of it.
Risk Level: Low
Testing: Validated benchmark tests/ runs.
Docs Changes: To use this optimization, define "SHARD_ALGORITHM" and "SHARD_SHIFT" in the build scripts/ options.
Release Notes: Mention of RingHash optimization using a Sharding algorithm.
Platform Specific Features: None.
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


KBaichoo:
/assign @zuercher 

Since you're familiar with this area.

/wait 

For CI format fail

suniltheta:
Without dwelling deep into the implementation, here are some high level comments

You might want to run formatter https://github.com/envoyproxy/envoy/blob/main/support/README.md#fixing-format-problems 
to fix these spelling and formatting issues https://dev.azure.com/cncf/envoy/_build/results?buildId=125136&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39 so that the CI is clean and ready for review

Consider using a runtime guard https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#runtime-guarding instead of #ifdef, incase you think the changes are still experimental, which I think is Envoy way instead of using build scripts/options.

Feel free to mention this under appropriate section on https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml to make changes under release notes.

Did you happen to make any unit test changes https://github.com/envoyproxy/envoy/blob/main/test/common/upstream/ring_hash_lb_test.cc so that we get good code coverage?

zuercher:
I haven’t looked closely, but I expect a runtime guard will be required so that this change can be tested and quickly reverted.

Does this change alter the output of the algorithm?

RamaMalladiAWS:
> Without dwelling deep into the implementation, here are some high level comments
> 
> You might want to run formatter https://github.com/envoyproxy/envoy/blob/main/support/README.md#fixing-format-problems to fix these spelling and formatting issues https://dev.azure.com/cncf/envoy/_build/results?buildId=125136&view=logs&j=55bc2ab9-36c1-5ab8-58a6-8c258728eeff&t=33b57bf9-ef13-58c8-bc2b-f804aeb16d39 so that the CI is clean and ready for review
> 
> Consider using a runtime guard https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#runtime-guarding instead of #ifdef, incase you think the changes are still experimental, which I think is Envoy way instead of using build scripts/options.
> 
> Feel free to mention this under appropriate section on https://github.com/envoyproxy/envoy/blob/main/changelogs/current.yaml to make changes under release notes.
> 
> Did you happen to make any unit test changes https://github.com/envoyproxy/envoy/blob/main/test/common/upstream/ring_hash_lb_test.cc so that we get good code coverage?
> 
> Finally, you can avoid force pushing the new commits into this PR, since when the PR is finally merged all the commits will be squashed. This makes the job of both reviewer and committer easier to just look at the diff since any change request.

Thank you @suniltheta. I got the formatting fixes and runtime-guarding code changes done. I am checking the unittests and changes any needed.

RamaMalladiAWS:
> I haven’t looked closely, but I expect a runtime guard will be required so that this change can be tested and quickly reverted.
> 
> Does this change alter the output of the algorithm?

Hi @zuercher, I am checking the unittests and will update shortly.

mattklein123:
Just to echo some previous comments. A few things:

1) This change should not be behind an if-def. If it's a permanent change for the better, it should be behind a runtime guard. If it changes behavior, it should be behind a configuration flag. It will also need documentation as it's not clear to me from a quick scan what this does.
2) Needs unit/integration/etc. testing.

Thank you!

RamaMalladiAWS:
I uploaded commit bbaa3efc86a0de88652acc0933606d7b2420b2d2 with changes for runtime guarding and ring_hash_lb unittest.

RamaMalladiAWS:
> Just to echo some previous comments. A few things:
> 
> 1. This change should not be behind an if-def. If it's a permanent change for the better, it should be behind a runtime guard. If it changes behavior, it should be behind a configuration flag. It will also need documentation as it's not clear to me from a quick scan what this does.
> 2. Needs unit/integration/etc. testing.
> 
> Thank you!

I added runtime guarding for the code. I can make it FALSE by default (it isn't so now).
Regarding documentation of what this code does, the idea of the shard ring hash algorithm is to mimic how a general sharding algorithm works! Divide the hosts list into buckets (using right shift of hashes) and lookup on those buckets instead of full host list. So, we do fewer lookups and hence faster execution of the ring hash code!

RamaMalladiAWS:
> > I haven’t looked closely, but I expect a runtime guard will be required so that this change can be tested and quickly reverted.
> > Does this change alter the output of the algorithm?
> 
> Hi @zuercher, I am checking the unittests and will update shortly.

@zuercher, I did check and found that the port counts didn't exactly match with expected and I had to modify the unittest to make it NEAR instead of EQ. So, there is a change in hash to host allocation when using this shard algorithm/ implementation.

source/common/upstream/ring_hash_lb.cc
```diff
@@ -59,8 +59,35 @@ HostConstSharedPtr RingHashLoadBalancer::Ring::chooseHost(uint64_t h, uint32_t a
...
-  int64_t lowp = 0;
-  int64_t highp = ring_.size();
+  int64_t lowp, highp;
+
+  // Algorithm is to shard the indices and lookup the host for a given hash from a shard
+  // (instead of a lookup of all hosts). Hence fewer lookups/ accesses and faster execution.
+  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.shard_ringhash")) {
+
+    // Given a hash 'h', find the shard index by shifting it to right (by rightShift).
+    uint64_t shard_index = (h >> (rightShift - 1)) >> 1;
```
suniltheta, line 69: Just for my understanding, why is this not just `h >> rightShift`?

test/common/upstream/ring_hash_lb_test.cc
```diff
@@ -623,9 +623,9 @@ TEST_P(RingHashLoadBalancerTest, HostWeightedLargeRing) {
     ++counts[port - 90];
   }
 
-  EXPECT_EQ(987, counts[0]);  // :90 | ~1000 expected hits
-  EXPECT_EQ(1932, counts[1]); // :91 | ~2000 expected hits
-  EXPECT_EQ(3081, counts[2]); // :92 | ~3000 expected hits
+  EXPECT_EQ(987, counts[0]);        // :90 | ~1000 expected hits
+  EXPECT_NEAR(1932, counts[1], 10); // :91 | ~2000 expected hits
```
suniltheta, line 627: you can add code coverage with both runtime flag enabled & disabled. So you can keep this as `EXPECT_EQ` with runtime flag being disabled and add a new test case with runtime flag enabled.

source/common/upstream/ring_hash_lb.h
```diff
@@ -65,6 +66,9 @@ class RingHashLoadBalancer : public ThreadAwareLoadBalancerBase,
 
     std::vector<RingEntry> ring_;
 
+    std::vector<uint64_t> ring_shard_;
+    uint64_t rightShift = 9; // SHARD_SHIFT parameter
```
suniltheta, line 70: 1. might need snake_case here to satisfy the format check.
2. Can we use a better variable name here?

source/common/upstream/ring_hash_lb.cc
```diff
@@ -193,6 +220,54 @@ RingHashLoadBalancer::Ring::Ring(const NormalizedHostWeightVector& normalized_ho
...
+      if (current_shard != prev_shard) {
+        prev_shard = current_shard;
+        ring_shard_.push_back(ring_index);
+      }
+      ring_index++;
+    }
+    // For the last shard, we need end and hence storing ring_size value.
+    ring_shard_.push_back(ring_size);
+
+    /******* END: code for sharding algorithm *******/
```
suniltheta, line 268: Regarding documentation, see if it is useful to mention about the optimization in https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers.html?highlight=ring%20hash#ring-hash 

I would prefer if the runtime flag is made true by default. So that others can benefit from this change. Else we wouldn't normally expect folks to explicitly enable this flag. 

Include a note in changeslog that this change optimize the execution. if this optimization is not required then this change can be reverted using the runtime flag.

test/common/upstream/ring_hash_lb_test.cc
RamaMalladiAWS: > you can add code coverage with both runtime flag enabled & disabled. So you can keep this as `EXPECT_EQ` with runtime flag being disabled and add a new test case with runtime flag enabled.

I did this change suggested. Thanks @suniltheta

RamaMalladiAWS: > If this changes the ring layout this can't be behind a runtime flag. I think it should be a configuration setting? That is where the documentation would end up.

Hi @mattklein123, I couldn't get code changes done for configuration settings. For now, I am uploading a version with runtime guarding enabled. Can you help me with the code changes for configuration setting? Thanks.

source/common/upstream/ring_hash_lb.cc
RamaMalladiAWS: I didn't change the documentation yet. I updated the changelog: `changelogs/current.yaml`.
I changed `rightShift` to `rshift_to_shard`.

source/common/upstream/ring_hash_lb.h
RamaMalladiAWS: I tried to make the change @suniltheta. Please let me know if ok or suggest further change. Thanks

RamaMalladiAWS:
With this code change/ optimization of RingHash, the perf upside on an AWS Graviton3 `c7g.16xlarge` instance for `load_balancer_benchmark` was:

```c
Test                                                             Baseline    Shard Algorithm     Base/Shard ratio
benchmarkRingHashLoadBalancerChooseHost/100/65536/100000         20.2 ms     16.4 ms             1.23
benchmarkRingHashLoadBalancerChooseHost/200/65536/100000         21.1 ms     15.7 ms             1.34
benchmarkRingHashLoadBalancerChooseHost/500/65536/100000         24.6 ms     20.3 ms             1.21
benchmarkRingHashLoadBalancerChooseHost/100/256000/100000        29.6 ms     13.7 ms             2.16
benchmarkRingHashLoadBalancerChooseHost/200/256000/100000        31.3 ms     14.2 ms             2.20
benchmarkRingHashLoadBalancerChooseHost/500/256000/100000        34.3 ms     24.9 ms             1.38
```

Command-line:
```c
$ bazel run -c opt //test/common/upstream:load_balancer_benchmark
$ gcc -v
gcc version 11.3.0 (Ubuntu 11.3.0-1ubuntu1~22.04)
```

test/common/upstream/ring_hash_lb_test.cc
mattklein123: Sorry I don't have time to help with this code. I would recommend looking through recent changes that add configuration for some examples.

RamaMalladiAWS: I will check @mattklein123. Thanks

suniltheta:
Do you have similar benchmark results for x86 instance?

I can help point you towards making configuration changes as suggested by Matt.

RamaMalladiAWS:
> Do you have similar benchmark results for x86 instance?
> 
> I can help point you towards making configuration changes as suggested by Matt.

Here are the results on `c6i.16xlarge` instance for baseline and with the shard RingHash implementation.

```c
Test                                                             Baseline      Shard Algorithm    Base/Shard ratio
benchmarkRingHashLoadBalancerChooseHost/100/65536/100000         21.6           16.9              1.28
benchmarkRingHashLoadBalancerChooseHost/200/65536/100000         21.7           16                1.36
benchmarkRingHashLoadBalancerChooseHost/500/65536/100000         23             18.8              1.22
benchmarkRingHashLoadBalancerChooseHost/100/256000/100000        27             16.6              1.63
benchmarkRingHashLoadBalancerChooseHost/200/256000/100000        27.5           17                1.62
benchmarkRingHashLoadBalancerChooseHost/500/256000/100000        29.8           22.6              1.32
```

Yes, it would help if you can point me to the configuration changes needed for this implementation. Thank you @suniltheta.

suniltheta:
I was looking at LoadBalancing (LB) config to understand where you can make these changes.

Traditionally the v3 API config was present as LbConfig inside Cluster config. Example [config.cluster.v3.Cluster.RoundRobinLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-roundrobinlbconfig) which inturn can use [config.cluster.v3.Cluster.CommonLbConfig.ConsistentHashingLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-commonlbconfig-consistenthashinglbconfig) (see in [code](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/cluster/v3/cluster.proto#L596)). You can probably place the config to `enable sharding` and to set `rshift_to_shard` (if useful to be configured by user instead of hardcoded 9).

Recently there is a push to make these LB configs as [Load balancing policies](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing_policies) extensions. In that case you can add inside [consistent_hashing_lb_config](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/common/v3/common.proto#envoy-v3-api-msg-extensions-load-balancing-policies-common-v3-consistenthashinglbconfig) (see in [code](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/load_balancing_policies/common/v3/common.proto#L90)) for [ring_hash.proto](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/ring_hash/v3/ring_hash.proto). LB policies will eventually make the old technique deprecated but until that time I think it is better to have this capability to configure sharding in both the places.

Another question, is that you can decide if you want to add this config directly inside [extensions.load_balancing_policies.common.v3.ConsistentHashingLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/common/v3/common.proto#extensions-load-balancing-policies-common-v3-consistenthashinglbconfig) or inside [Ring Hash Load Balancing Policy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/ring_hash/v3/ring_hash.proto). The ConsistentHashingLbConfig is common for both Ring Hash & Maglev. Would it be possible to extend the sharding technique for Maglev as well? If not can we need to check with API Shepards on what they think of placing the sharding config inside ConsistentHashingLbConfig. Can we mention that sharding is only supported for ring hash and not for Maglev, or do we need to add this config in different place.

RamaMalladiAWS:
Hi @zuercher, I am able to work on this actively now. Can you please remove the stale tag? Thanks

PR #24882, opened 2023-01-11
build: test fix

alyssawilk:
Fixing a bad merge where an Envoy test is missing a required extension

### DIFF
--- a/mobile/test/common/BUILD
+++ b/mobile/test/common/BUILD
@@ -10,6 +10,7 @@ envoy_cc_test(
     repository = "@envoy",
     deps = [
         "//library/common:engine_common_lib",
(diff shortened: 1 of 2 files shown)


mobile/test/common/BUILD
```diff
@@ -10,6 +10,7 @@ envoy_cc_test(
     repository = "@envoy",
     deps = [
         "//library/common:engine_common_lib",
+         "@envoy//source/extensions/listener_managers/listener_manager:listener_manager_lib",
```
jpsim, line 13: Shouldn't this dep go in `//library/common:engine_common_lib` since it should always be compiled in, not just for tests?

Augustyniak (approved):

PR #24884, opened 2023-01-11
fix: for android without AlternateGetifaddrs support

alyssawilk:
Risk Level: low
Testing: 
Docs Changes:
Release Notes: n/a

### DIFF
--- a/mobile/library/common/network/android.cc
+++ b/mobile/library/common/network/android.cc
@@ -31,9 +31,10 @@ namespace {
 void setAlternateGetifaddrs() {
 #if defined(INCLUDE_IFADDRS)
   auto& os_syscalls = Api::OsSysCallsSingleton::get();
-  ENVOY_BUG(!os_syscalls.supportsGetifaddrs(),
-            "setAlternateGetifaddrs should only be called when supportsGetifaddrs is false");
-
+  if (!os_syscalls.supportsGetifaddrs()) {
+    ENVOY_LOG_MISC(warn, "supportsGetifaddrs unsupported");
+    return;
+  }
   Api::AlternateGetifaddrs android_getifaddrs =
       [](Api::InterfaceAddressVector& interfaces) -> Api::SysCallIntResult {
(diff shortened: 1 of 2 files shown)


mobile/library/common/network/android.cc
```diff
@@ -31,9 +31,10 @@ namespace {
 void setAlternateGetifaddrs() {
 #if defined(INCLUDE_IFADDRS)
```
RyanTheOptimist, line 32: I find it surprising that we have both an ifdef and a runtime guard here.

That being said, I might be inclined to do:

```
void setAlternateGetifaddrs() {
  if (!os_syscalls.supportsGetifaddrs()) {
    ENVOY_LOG_MISC(warn, "supportsGetifaddrs unsupported");
    return;
  }

#if defined(INCLUDE_IFADDRS)
  Api::AlternateGetifaddrs android_getifaddrs = ...
#endif
```
But I don't think I quite understand INCLUDE_IFADDRS. Should/could it just be ANDROID_API instead?

alyssawilk: I mean the function is always called, and we only expect it to work on android, which is why I only warned in the ifdef where we assume it'll work

RyanTheOptimist: Bah! Sorry! Sorry! Sorry. Cut/paste error. I was attempting to describe how the code would look if we wanted the old invariant to remain. I should have said BUG instead of LOG:

```
void setAlternateGetifaddrs() {
  ENVOY_BUG(!os_syscalls.supportsGetifaddrs(),
            "setAlternateGetifaddrs should only be called when supportsGetifaddrs is false");
            
#if defined(INCLUDE_IFADDRS)
  Api::AlternateGetifaddrs android_getifaddrs = ...
#endif
```

But if we decide the method can be called unconditionally, then I agree with your new code completely

alyssawilk: sure, I'll go that route

PR #24889, opened 2023-01-11
Add // IWYU: export to log_level.h

RyanTheOptimist:
Add // IWYU: export to log_level.h in order to fix a clang-tidy error which shows up internally.

### DIFF
--- a/mobile/library/cc/log_level.h
+++ b/mobile/library/cc/log_level.h
@@ -7,7 +7,7 @@
 namespace Envoy {
 namespace Platform {
 
-using LogLevel = Envoy::Logger::Logger::Levels;
+using LogLevel = Envoy::Logger::Logger::Levels;  // IWYU: export
 
 std::string logLevelToString(LogLevel method);
 LogLevel logLevelFromString(const std::string& str);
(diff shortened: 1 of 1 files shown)


abeyad (approved):

jpsim:
Is this a false positive in clang-tidy? Is the header actually used?

RyanTheOptimist:
> Is this a false positive in clang-tidy? Is the header actually used?

Ah. So clang-tidy complains in some code we have internally which includes log_level.h and then attempts to use LogLevel which is defined on line 10. I think it's a legit warning. Or alternatively, we could get rid of that "using" and instead just use the enum from source/common/common/base_logger.h directly (and have callers include that file).

WDYT?

mjbshaw:
Yeah, clang-tidy sees that `mobile/library/cc/log_level.h` exports the type `LogLevel`, so if that's all I use then there are no complaints. But the enum constants are not defined in that header, so if I do something like `LogLevel::off` then clang-tidy will complain that I'm not including `#include "source/common/common/base_logger.h"`.

jpsim:
> WDYT?

Fine with me, I was just trying to understand a bit better. Thanks for explaining!

PR #24890, opened 2023-01-11
Add a destructor to Envoy::Platform::Engine which calls terminate()

RyanTheOptimist:
Add a destructor to Envoy::Platform::Engine which calls terminate()


jpsim (approved):
👍 as long as there aren't any threading concerns here.

RyanTheOptimist:
> 👍 as long as there aren't any threading concerns here.

Thanks. Looking at how we use this, there should be no threading concerns.

Issue #22055, opened 2022-07-07
load balancer: earlyExitNonLocalityRouting support for the thread aware load balancer


wbpcode:
*Title*: *load balancer: earlyExitNonLocalityRouting support for the thread aware load balancer*

*Description*:

If the `locality_weighted_lb_config` in cluster is set, but the `locality` in the `LocalityLbEndpoints` is not set, then the thread aware load balancer may create empty hash ring.

It would be better to support some pre-check for the thread aware load balancer, like the zone aware load balancer do.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

wbpcode:
In fact, this like a bug. 🤣

wbpcode:
cc @snowp

PR #24899, opened 2023-01-12
fix the possbile UH of ringhash/maglev when locality_weighted_lb_config is set or LoadBalancingPolicyConfig is used

wbpcode:
Commit Message: fix the possbile UH of ringhash/maglev when locality_weighted_lb_config is set or LoadBalancingPolicyConfig is used
Additional Description:

See #22055.

Risk Level: Low.
Testing: n/a.
Docs Changes: n/a.
Release Notes: n/a.


wbpcode:
The specific switch would be better.

PR #21428, opened 2022-05-25
tls: fix daysUntilFirstCertExpires() return nil when expired

daixiang0:
Signed-off-by: Loong Dai <loong.dai@intel.com>

Add days_until_first_cert_expiring stat integration test to ensure no nil error.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #24900, opened 2023-01-12
Add support for multiple sds configs in DownstramTlsContext

### LINKED ISSUE #24824: SDS secret config not usable for SNI based cert selection with more than two certificates
*Title*: SDS secret config not usable for SNI based cert selection with more than two certificates

*Description*:
Related to #24483 which allows a `DownstreamTlsContext` to support multiple TLS certificates via the `tls_certificates` config, with the listener selecting a certificate on handshake based on SNI.

We're trying to use this feature with `tls_certificate_sds_secret_configs`, which according to the docs should support the same number and types of certificates as `tls_certificates`.

When configuring with more than 2 SDS secret configs this way, Envoy rejects the config with a validation error (because the proto has a [`max_items=2` rule](https://github.com/envoyproxy/envoy/blob/ffa7295273834e2c777b1464896397bb05c0f68a/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L258))

there doesn't seem to be another way to provide more than 2 certificates via SDS to a DownstreamTLSListener? The limit was added in https://github.com/envoyproxy/envoy/pull/16605 to only support an ECDSA and a RSA cert via SDS, is that still a limitation for certs provided via SDS or is the fix as simple as removing the validation rule?

*Repro steps*:
- Running `envoyproxy/envoy-dev:317efa53c6add02fa9533fc8f93f10f590077dca` (post the feature above being merged)
- Configure a Listener with a DownstreamTLSContext and more than 2 SDSSecretConfigs
- Rejected Update below (including invalid config)
```
[2023-01-09 11:39:28.992][1][warning][config] [source/common/config/grpc_subscription_impl.cc:128] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) https_listener: Proto constraint validation failed (DownstreamTlsContextValidationError.CommonTlsContext: embedded message failed validation | caused by CommonTlsContextValidationError.TlsCertificateSdsSecretConfigs: value must contain no more than 2 item(s)): common_tls_context {
  tls_params {
    tls_minimum_protocol_version: TLSv1_2
    cipher_suites: "[ECDHE-ECDSA-AES128-GCM-SHA256|ECDHE-ECDSA-CHACHA20-POLY1305]"
    cipher_suites: "[ECDHE-RSA-AES128-GCM-SHA256|ECDHE-RSA-CHACHA20-POLY1305]"
    cipher_suites: "ECDHE-ECDSA-AES256-GCM-SHA384"
    cipher_suites: "ECDHE-RSA-AES256-GCM-SHA384"
    cipher_suites: "ECDHE-ECDSA-AES256-SHA"
    cipher_suites: "ECDHE-RSA-AES256-SHA"
  }
  alpn_protocols: "h2"
  alpn_protocols: "http/1.1"
  tls_certificate_sds_secret_configs {
    name: "AAAAAAAA-KEY_TYPE_RSA_2048-12"
    sds_config {
      ads {
      }
      resource_api_version: V3
    }
  }
  tls_certificate_sds_secret_configs {
    name: "BBBBBBBBB-KEY_TYPE_RSA_2048-6"
    sds_config {
      ads {
      }
      resource_api_version: V3
    }
  }
  tls_certificate_sds_secret_configs {
    name: "CCCCCCCC-b7f5"
    sds_config {
      ads {
      }
      resource_api_version: V3
    }
  }
}
```

LuyaoZhong:
Remove max_items=2 validation rules from sds configs proto. Add test cases to verify that multiple sds configs is allowed and works with SNI-based cert selection.

Fixes #24824


test/extensions/transport_sockets/tls/context_impl_test.cc
```diff
@@ -1805,16 +1805,28 @@ TEST_F(ServerContextConfigImplTest, TlsCertificatesAndSdsConfig) {
       "SDS and non-SDS TLS certificates may not be mixed in server contexts");
 }
 
-TEST_F(ServerContextConfigImplTest, MultiSdsConfig) {
+TEST_F(ServerContextConfigImplTest, SdsConfigNoName) {
```
LuyaoZhong, line 1808: The old test does not make sense. The validation failure is caused by not setting name instread of dual configs.

test/config/integration/certs/servercert_info.h
```diff
@@ -1,5 +1,6 @@
 // NOLINT(namespace-envoy)
-constexpr char TEST_SERVER_CERT_HASH[] = "DC:E2:2B:65:90:43:9A:36:1C:8E:6D:CA:42:8A:8C:37:C7:A1:77:"
-                                         "00:5B:C1:3E:33:8A:B9:2D:04:2C:B1:3F:0A";
-constexpr char TEST_SERVER_CERT_NOT_BEFORE[] = "Apr 7 16:46:35 2022 GMT";
-constexpr char TEST_SERVER_CERT_NOT_AFTER[] = "Apr 6 16:46:35 2024 GMT";
+constexpr char TEST_SERVER_CERT_256_HASH[] =
```
LuyaoZhong, line 2: regenerating this header file causes [CI failure](https://dev.azure.com/cncf/envoy/_build/results?buildId=125097&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec) since it overrides the old content, TEST_SERVER_CERT_NOT_AFTER is not generated but other test still replies on it.

I checked that servercert_info.h was first introduced by https://github.com/envoyproxy/envoy/pull/21428.
@daixiang0 How did you generate this file, I use generate_info_header in certs.sh, but the output is different.

ggreenway (approved):
LGTM. Thanks!

test/config/integration/certs/servercert_info.h
ggreenway: From a comment (https://github.com/envoyproxy/envoy/pull/21428/files#r884045881) it sounds like it was supposed to be generated, but it seems like that isn't the case for some reason. If you'd like you can manually add NOT_BEFORE and NOT_AFTER for the new correct values, and file an issue to fix so that they're generated automatically like they should be. Or if you'd like to fix it, that would be great.

LuyaoZhong: I see, it is probably that the command in certs.sh is modified downstream to generate header info but not get committed and upstreamed.

daixiang0: > regenerating this header file causes [CI failure](https://dev.azure.com/cncf/envoy/_build/results?buildId=125097&view=logs&j=8c169225-0ae8-53bd-947f-07cb81846cb5&t=6b0ace90-28b2-51d9-2951-b1fd35e67dec) since it overrides the old content, TEST_SERVER_CERT_NOT_AFTER is not generated but other test still replies on it.
> 
> I checked that servercert_info.h was first introduced by #21428. @daixiang0 How did you generate this file, I use generate_info_header in certs.sh, but the output is different.

When run the script, it will update all contents, so you can add value and update all.

LuyaoZhong: @daixiang0 yes, but we should add changes to certs.sh otherwise when other people run certs.sh the old content is not generated.

daixiang0:
https://dev.azure.com/cncf/envoy/_build/results?buildId=125213&view=logs&j=372032e6-8bf2-5d86-e328-540457529814&t=85caf536-e2ae-5b30-3a82-f9aadd42f391&l=183

Network issue, wait for a moment then retest.

lizan (commented):
/lgtm api

Issue #24729, opened 2023-01-03
idle time active health check


zhangbo1882:
Now envoy can be configured with active health check for the upstream hosts periodically. If there are huge number of upstream hosts, the health check traffic will also become huge. 

Actually, if there is continuous traffic for an upstream host, there is no need to do health check for that host with a fixed frequency. 
Is it possible to do health check only when there is no correct business traffic response to that host?
For example, if there is no traffic for 1 second, then do active health check. If there is always traffic to the host (HTTP response 2xx), the host should be considered as healthy without active health check.

htuch:
I'm not sure this solves the scalability problem of active HC in general, since it's really dependent on traffic patterns and a specific notion of upstream health. This is threading the line between outlier detection and active HC while removing much of the distinguishing advantages of active HC. It is a plausible optimization with these caveats. @mattklein123 ?

mattklein123:
Yeah I have thought about an optimization like this, but it really depends on the use case and it would certainly need to be configurable. Without passive health checking configured it wouldn't work at all obviously. I will mark this help wanted.

zhangbo1882:
I propose a PR https://github.com/envoyproxy/envoy/pull/24903. 
Can someone have an initial review to see whether the functionality works as expected ?
I will add test case code.

spacewander:
Is this issue solved by #24903?

mattklein123:
> Is this issue solved by https://github.com/envoyproxy/envoy/pull/24903?

Yes I think so, thanks.

PR #24903, opened 2023-01-12
add idle active hc for TCP/HTTP/GRPC

zhangbo1882:
https://github.com/envoyproxy/envoy/issues/24729

### DIFF
--- a/test/integration/health_check_integration_test.cc
+++ b/test/integration/health_check_integration_test.cc
@@ -646,6 +646,7 @@ TEST_P(TcpHealthCheckIntegrationTest, SingleEndpointHealthyTcp) {
 
 TEST_P(TcpHealthCheckIntegrationTest, DisableHCForActiveTraffic) {
   const uint32_t cluster_idx = 0;
+  auto& cluster_data = clusters_[cluster_idx];
   initialize();
   initTcpHealthCheck(cluster_idx, true);
 
@@ -673,6 +674,9 @@ TEST_P(TcpHealthCheckIntegrationTest, DisableHCForActiveTraffic) {
   timeSystem().advanceTimeWait(std::chrono::seconds(10));
 
   test_server_->waitForCounterEq("cluster.cluster_1.health_check.attempt", 3);
+
+  ASSERT_TRUE(cluster_data.host_fake_raw_connection_->waitForData(
+      FakeRawConnection::waitForInexactMatch("Ping")));
   result = clusters_[cluster_idx].host_fake_raw_connection_->write("Pong");
   RELEASE_ASSERT(result, result.message());
 


zhangbo1882:
work in progress

api/envoy/config/core/v3/health_check.proto
```diff
@@ -414,4 +414,6 @@ message HealthCheck {
   // the cluster's :ref:`transport socket <envoy_v3_api_field_config.cluster.v3.Cluster.transport_socket>`
   // will be used for health check socket configuration.
   google.protobuf.Struct transport_socket_match_criteria = 23;
+
+  bool enable_idle_hc = 25;
```
mattklein123, line 418: Please add docs so we can agree on the semantics.

source/common/router/router.cc
```diff
@@ -359,6 +359,9 @@ void Filter::chargeUpstreamCode(uint64_t response_status_code,
     if (upstream_host && Http::CodeUtility::is5xx(response_status_code)) {
       upstream_host->stats().rq_error_.inc();
     }
+    if (upstream_host && Http::CodeUtility::is2xx(response_status_code)) {
+      upstream_host->setLastTrafficTime2xx(callbacks_->dispatcher().timeSource().monotonicTime());
+    }
```
mattklein123, line 364: You probably will want to handle gRPC here. 2xx isn't enough for that.

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -419,6 +420,26 @@ HealthCheckerImplBase::ActiveHealthCheckSession::clearPendingFlag(HealthTransiti
...
+    MonotonicTime traffic_pass_time;
+    switch(parent_.healthCheckerType()) {
+      case envoy::data::core::v3::HealthCheckerType::TCP:
+        traffic_pass_time = host_->lastTrafficPassTime();
+        break;
+      case envoy::data::core::v3::HealthCheckerType::HTTP:
+        traffic_pass_time = host_->lastTrafficPassTime2xx();
+        break;
+      default:
+        break;
```
mattklein123, line 434: What about gRPC, etc.? We should handle the cases for all health checkers if possible. Redis, etc.

mattklein123 (changes requested):
Thanks for working on this. Left some initial comments, thanks.

/wait

api/envoy/config/core/v3/health_check.proto
```diff
@@ -414,4 +414,9 @@ message HealthCheck {
   // the cluster's :ref:`transport socket <envoy_v3_api_field_config.cluster.v3.Cluster.transport_socket>`
   // will be used for health check socket configuration.
   google.protobuf.Struct transport_socket_match_criteria = 23;
+
+  // When the health status of a host is healthy, if this field is enabled, when envoy wants to send active health check packet to target host, it will first check whether
+  // there is already business traffic hit the target host during the interval. If yes, no need to send health check packet.
+  // The default value is false.
+  bool enable_idle_hc = 25;
```
mattklein123, line 421: I think I would call this: `disable_health_check_if_active_traffic` or something more specific.

api/envoy/config/core/v3/health_check.proto
```diff
@@ -414,4 +414,9 @@ message HealthCheck {
   // the cluster's :ref:`transport socket <envoy_v3_api_field_config.cluster.v3.Cluster.transport_socket>`
   // will be used for health check socket configuration.
   google.protobuf.Struct transport_socket_match_criteria = 23;
+
+  // When the health status of a host is healthy, if this field is enabled, when envoy wants to send active health check packet to target host, it will first check whether
+  // there is already business traffic hit the target host during the interval. If yes, no need to send health check packet.
+  // The default value is false.
```
mattklein123, line 420: I will help you copy edit this later, but in the meantime this needs more clarity around how HTTP is handled, gRPC, etc. Also, I think this should also work for other health checkers like Redis?

Additionally, this should have some documentation added in the health checking arch overview page.

envoy/upstream/host_description.h
```diff
@@ -202,6 +202,12 @@ class HostDescription {
    *         healthy state via an active healthchecking.
    */
   virtual absl::optional<MonotonicTime> lastHcPassTime() const PURE;
+  virtual MonotonicTime lastTrafficPassTime() const PURE;
+  virtual MonotonicTime lastTrafficPassTime2xx() const PURE;
+  virtual MonotonicTime lastTrafficPassTimeGrpc() const PURE;
+  virtual void setLastTrafficTime(MonotonicTime last_traffic_pass_time) const PURE;
+  virtual void setLastTrafficTime2xx(MonotonicTime last_traffic_pass_time) const PURE;
+  virtual void setLastTrafficTimeGrpc(MonotonicTime last_traffic_pass_time) const PURE;
```
mattklein123, line 210:
```suggestion
  virtual void setLastTrafficTimeGrpcSuccess(MonotonicTime last_traffic_pass_time) const PURE;
```

?

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -419,6 +419,30 @@ HealthCheckerImplBase::ActiveHealthCheckSession::clearPendingFlag(HealthTransiti
 }
 
 void HealthCheckerImplBase::ActiveHealthCheckSession::onIntervalBase() {
+  /*Only if host is healthy, check whether need idle hc*/
```
mattklein123, line 422: nit: use `//` comments here and below.

envoy/upstream/host_description.h
```diff
@@ -202,6 +202,12 @@ class HostDescription {
    *         healthy state via an active healthchecking.
    */
   virtual absl::optional<MonotonicTime> lastHcPassTime() const PURE;
+  virtual MonotonicTime lastTrafficPassTime() const PURE;
+  virtual MonotonicTime lastTrafficPassTime2xx() const PURE;
+  virtual MonotonicTime lastTrafficPassTimeGrpc() const PURE;
+  virtual void setLastTrafficTime(MonotonicTime last_traffic_pass_time) const PURE;
```
mattklein123, line 208:
```suggestion
  virtual void setLastTrafficTimeTcpSuccess(MonotonicTime last_traffic_pass_time) const PURE;
```

? 

Methods above should be altered to match.

mattklein123 (changes requested):
Thanks generally looks like it's on the right track. Left some more comments and you can start working on tests.

/wait

api/envoy/config/core/v3/health_check.proto
```diff
@@ -414,4 +414,12 @@ message HealthCheck {
   // the cluster's :ref:`transport socket <envoy_v3_api_field_config.cluster.v3.Cluster.transport_socket>`
   // will be used for health check socket configuration.
   google.protobuf.Struct transport_socket_match_criteria = 23;
+
+  // When the health status of a host is healthy, if this field is enabled, when envoy wants to send active health check packet to target host, it will first check whether
+  // there is already active business traffic hit the target host during the interval. If yes, no need to send health check packet.
+  // For http, if the business traffic response is 2xx, it will be considered as active traffic
+  // For tcp,  if there is any business traffic, it will be considered as active traffic
+  // For grpc, if grpc_status is non-zero, it will be considered as active traffic
+  // The default value is false.
```
mattklein123, line 423:
```suggestion
  // When the health status of a host is healthy, if this field is enabled, when envoy wants to send an active health check packet to a target host, it will first check whether
  // there was *successful* non-health check traffic that targeted the host during the previous interval. If there was successful non-health check traffic, Envoy will not send a health check packet.
  // For HTTP, if the traffic response is 2xx, it will be considered as successful traffic.
  // For TCP,  if any TCP connections successfully connect, it will be considered as successful traffic.
  // For gRPC, if *grpc-status* is zero, it will be considered as successful traffic
  // The default value is false.
```

You indicated gRPC status as non-zero when I think you meant zero, right?

envoy/upstream/host_description.h
```diff
@@ -202,6 +202,12 @@ class HostDescription {
    *         healthy state via an active healthchecking.
    */
   virtual absl::optional<MonotonicTime> lastHcPassTime() const PURE;
+  virtual MonotonicTime lastTrafficPassTime() const PURE;
+  virtual MonotonicTime lastTrafficPassTime2xx() const PURE;
+  virtual MonotonicTime lastTrafficPassTimeGrpc() const PURE;
+  virtual void setLastTrafficTimeTcpSuccess(MonotonicTime last_traffic_pass_time) const PURE;
+  virtual void setLastTrafficTimeHttp2xx(MonotonicTime last_traffic_pass_time) const PURE;
+  virtual void setLastTrafficTimeGrpcSuccess(MonotonicTime last_traffic_pass_time) const PURE;
```
mattklein123, line 210: nit: can you refactor this to have a singlet getter and setter that takes an enum type? I think this would be more clear and better for future proofing.

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -419,6 +419,31 @@ HealthCheckerImplBase::ActiveHealthCheckSession::clearPendingFlag(HealthTransiti
...
+      break;
+    case envoy::data::core::v3::HealthCheckerType::HTTP:
+      traffic_pass_time = host_->lastTrafficPassTime2xx();
+      break;
+    case envoy::data::core::v3::HealthCheckerType::GRPC:
+      traffic_pass_time = host_->lastTrafficPassTimeGrpc();
+      break;
+    default:
+      break;
+    }
```
mattklein123, line 438: What is this going to do for cases like Redis? I think you either have to handle this for all health checkers or error out the config if it is set on one that is not supported?

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -419,6 +419,31 @@ HealthCheckerImplBase::ActiveHealthCheckSession::clearPendingFlag(HealthTransiti
...
+    auto duration = time_source_.monotonicTime() - traffic_pass_time;
+    if (duration.count() < std::chrono::milliseconds(parent_.interval_ * 1000000).count()) {
```
mattklein123, line 440: I'm not sure why we are multiplying by 1,000,000 here. Can you use duration casts and direct duration comparison to make this more clear?

source/common/upstream/health_checker_base_impl.cc
```diff
@@ -419,6 +419,31 @@ HealthCheckerImplBase::ActiveHealthCheckSession::clearPendingFlag(HealthTransiti
...
+      /* During the time, there is already business traffic for the host, no need to send HC packet
+       */
```
mattklein123, line 442: nit: // style comments.

mattklein123 (changes requested):
Added some comments. Also, I don't think you actually have a test that tests the behavior? Can you work on that? It would also be nice to get an integration test as well. Thanks.

/wait

api/envoy/config/core/v3/health_check.proto
zhangbo1882: **Filter::handleNon5xxResponseHeaders(absl::optional<Grpc::Status::GrpcStatus> grpc_status,
                                         UpstreamRequest& upstream_request, bool end_stream,
                                         uint64_t grpc_to_http_status) {
  // We need to defer gRPC success until after we have processed grpc-status in
  // the trailers.
  if (grpc_request_) {
    if (end_stream) {
      if (grpc_status && !Http::CodeUtility::is5xx(grpc_to_http_status)) {
        upstream_request.upstreamHost()->stats().rq_success_.inc();
      } else {
        upstream_request.upstreamHost()->stats().rq_error_.inc();
      }**
      From this function, I think when grpc-status is non-zero, it will be consider as successful traffic.

mattklein123: I don't think that's what that function does exactly, or it's not implemented correctly. A non-zero grpc-status is an error. Can you double check?

source/common/upstream/health_checker_base_impl.cc
zhangbo1882: I will error out the config for custom health check type.

source/common/upstream/health_checker_base_impl.cc
zhangbo1882: fixed

source/common/upstream/health_checker_base_impl.cc
zhangbo1882: fixed

envoy/upstream/host_description.h
zhangbo1882: fixed.

api/envoy/config/core/v3/health_check.proto
zhangbo1882: From current code, if grpc-status is zero, the request error counter will increase. It seems that it is not correct. 
Just as you said non-zero grpc-status was an error (https://grpc.github.io/grpc/core/md_doc_statuscodes.html). I am really confused with the current logic.

zhangbo1882: In my opinion, the grpc_status checking should be removed.

void Filter::handleNon5xxResponseHeaders(absl::optional<Grpc::Status::GrpcStatus> grpc_status,
                                         UpstreamRequest& upstream_request, bool end_stream,
                                         uint64_t grpc_to_http_status) {
  // We need to defer gRPC success until after we have processed grpc-status in
  // the trailers.
  if (grpc_request_) {
    if (end_stream) {
      if ( !Http::CodeUtility::is5xx(grpc_to_http_status)) {
        upstream_request.upstreamHost()->stats().rq_success_.inc();
        upstream_request.upstreamHost()->setLastSuccessfulTrafficTime(
            envoy::data::core::v3::HealthCheckerType::GRPC,
            callbacks_->dispatcher().timeSource().monotonicTime());
      } else {
        upstream_request.upstreamHost()->stats().rq_error_.inc();
      }
    } else {
      upstream_request.grpcRqSuccessDeferred(true);
    }
  } else {
    upstream_request.upstreamHost()->stats().rq_success_.inc();
  }

mattklein123:
Can you please take a crack at adding an integration test to https://github.com/envoyproxy/envoy/blob/main/test/integration/health_check_integration_test.cc? Thanks.

/wait

zhangbo1882:
@mattklein123 , I have added the integration tests. Could you help review?
Thanks

mattklein123 (changes requested):
Thanks generally looks good. In at least one of the integration tests can you test not sending traffic during the interval and then sending an HC request? Thanks.

/wait

zhangbo1882:
> Thanks generally looks good. In at least one of the integration tests can you test not sending traffic during the interval and then sending an HC request? Thanks

@mattklein123 , I have updated the test case.

test/integration/health_check_integration_test.cc
```diff
@@ -669,6 +669,15 @@ TEST_P(TcpHealthCheckIntegrationTest, DisableHCForActiveTraffic) {
 
   EXPECT_EQ(1, test_server_->counter("cluster.cluster_1.health_check.success")->value());
   EXPECT_EQ(0, test_server_->counter("cluster.cluster_1.health_check.failure")->value());
+
+  timeSystem().advanceTimeWait(std::chrono::seconds(10));
+
+  test_server_->waitForCounterEq("cluster.cluster_1.health_check.attempt", 3);
+  result = clusters_[cluster_idx].host_fake_raw_connection_->write("Pong");
+  RELEASE_ASSERT(result, result.message());
```
mattklein123, line 677: This doesn't actually check that you get the ping though, right? Can we actually check that?

/wait

zhangbo1882: fixed

mattklein123 (approved):
Thanks!

PR #24905, opened 2023-01-12
upstream: lb policy configuration memory optimization

DiazAlan:
Commit Message: Changes how LB policies are stored to save memory
Additional Description:
Changed how some of these configs are stored to be unique_ptr instead of optionals. This will help save memory in cases where some of these policies are not set or used.

This will save a percentage of memory used by ClusterInfo and Subset_Lb

sizeof() test results:
Without these changes:
ClusterInfoImpl: 2896
Subset_LB: 536

With these changes:
ClusterInfoImpl: 2704
Subset_Lb: 384

Risk Level: Low
Testing: Updated existing load balancing tests to ensure success with new types
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/upstream/subset_lb.cc
+++ b/source/common/upstream/subset_lb.cc
@@ -36,22 +36,22 @@ SubsetLoadBalancer::SubsetLoadBalancer(
       lb_ring_hash_config_(
           lb_ring_hash_config.has_value()
               ? std::make_unique<const envoy::config::cluster::v3::Cluster::RingHashLbConfig>(
-                    lb_ring_hash_config.value())
+                    lb_ring_hash_config.ref())
               : nullptr),
       lb_maglev_config_(
           lb_maglev_config.has_value()
               ? std::make_unique<const envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
-                    lb_maglev_config.value())
+                    lb_maglev_config.ref())
               : nullptr),
       round_robin_config_(
           round_robin_config.has_value()
               ? std::make_unique<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
-                    round_robin_config.value())
+                    round_robin_config.ref())
               : nullptr),
       least_request_config_(
           least_request_config.has_value()
               ? std::make_unique<const envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>(
-                    least_request_config.value())
+                    least_request_config.ref())
               : nullptr),
       common_config_(common_config), stats_(stats), scope_(scope), runtime_(runtime),
       random_(random), fallback_policy_(subsets.fallbackPolicy()),
--- a/source/common/upstream/subset_lb.cc
+++ b/source/common/upstream/subset_lb.cc
@@ -2,6 +2,7 @@
 
 #include <memory>
 
+#include "envoy/common/optref.h"
 #include "envoy/config/cluster/v3/cluster.pb.h"
 #include "envoy/config/core/v3/base.pb.h"
 #include "envoy/runtime/runtime.h"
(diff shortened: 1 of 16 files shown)


source/common/upstream/subset_lb.cc
```diff
@@ -25,19 +26,35 @@ SubsetLoadBalancer::SubsetLoadBalancer(
...
     const envoy::config::cluster::v3::Cluster::CommonLbConfig& common_config,
     TimeSource& time_source)
-    : lb_type_(lb_type), lb_ring_hash_config_(lb_ring_hash_config),
-      lb_maglev_config_(lb_maglev_config), round_robin_config_(round_robin_config),
-      least_request_config_(least_request_config), common_config_(common_config), stats_(stats),
-      scope_(scope), runtime_(runtime), random_(random), fallback_policy_(subsets.fallbackPolicy()),
+    : lb_type_(lb_type),
+      lb_ring_hash_config_(
+          lb_ring_hash_config.has_value()
+              ? std::make_unique<const envoy::config::cluster::v3::Cluster::RingHashLbConfig>(
```
pradeepcrao, line 38: nit: Can you use `ref()` here instead of `value()` to be more explicit that the copy constructor is invoked.

adisuissa:
cc @wbpcode - I think you've been working in this area recently

source/common/upstream/upstream_impl.h
```diff
@@ -764,25 +765,45 @@ class ClusterInfoImpl : public ClusterInfo,
...
   }
-  const absl::optional<envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>&
+  OptRef<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>
   lbRoundRobinConfig() const override {
-    return lb_round_robin_config_;
+    if (lb_round_robin_config_ == nullptr) {
+      return absl::nullopt;
+    } else {
+      return *lb_round_robin_config_;
+    }
```
adisuissa, line 774: nit: reduce code complexity by reducing nesting (here and below)
This can be rewritten by:
```
if (...) {
  return ;
}
return ;
```

source/common/upstream/upstream_impl.h
```diff
@@ -931,12 +952,16 @@ class ClusterInfoImpl : public ClusterInfo,
   const std::string maintenance_mode_runtime_key_;
   std::shared_ptr<UpstreamLocalAddressSelector> upstream_local_address_selector_;
   LoadBalancerType lb_type_;
-  absl::optional<envoy::config::cluster::v3::Cluster::RoundRobinLbConfig> lb_round_robin_config_;
-  absl::optional<envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>
+  const std::unique_ptr<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>
+      lb_round_robin_config_;
```
adisuissa, line 956: Are they all optional or is there a requirement that there will be at least one LbPoilcy?
If all are optional, consider having all of them in a single struct to reduce a bit more memory.
This can be done in a future PR.

source/common/upstream/subset_lb.h
```diff
@@ -358,11 +356,47 @@ class SubsetLoadBalancer : public LoadBalancer, Logger::Loggable<Logger::Id::ups
   const ProtobufWkt::Value* getMetadataFallbackList(LoadBalancerContext* context) const;
   LoadBalancerContextWrapper removeMetadataFallbackList(LoadBalancerContext* context);
 
+  OptRef<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig> lbRoundRobinConfig() const {
+    if (round_robin_config_ != nullptr) {
```
adisuissa, line 360: Same comment about reducing nesting (see below).

source/common/upstream/subset_lb.h
```diff
@@ -358,11 +356,47 @@ class SubsetLoadBalancer : public LoadBalancer, Logger::Loggable<Logger::Id::ups
...
+      return absl::nullopt;
+    }
+  }
+
   const LoadBalancerType lb_type_;
-  const absl::optional<envoy::config::cluster::v3::Cluster::RingHashLbConfig> lb_ring_hash_config_;
-  const absl::optional<envoy::config::cluster::v3::Cluster::MaglevLbConfig> lb_maglev_config_;
-  const absl::optional<envoy::config::cluster::v3::Cluster::RoundRobinLbConfig> round_robin_config_;
-  const absl::optional<envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>
+  const std::unique_ptr<const envoy::config::cluster::v3::Cluster::RingHashLbConfig>
```
adisuissa, line 393: Same comment as below regarding putting them all in one struct that has a pointer to it. (maybe future PR).

adisuissa (commented):
Thanks for reducing the amount of memory a cluster uses!
Overall LGTM, other than a few nits.

source/common/upstream/upstream_impl.h
yurykats: Maybe sometimes there aren't any. LoadBalancerType has 8 values, but there are only 5 configs here. A policy like "Random" is probably config-less.

yurykats (approved):

source/common/upstream/upstream_impl.h
pradeepcrao: Does only 1 of them usually being set make them a good candidate for `PackedStruct` ? #24743

adisuissa: > Does only 1 of them usually being set make them a good candidate for `PackedStruct` ? #24743

Doesn't `PackedStruct` require them to be of the same type?

yurykats: I think this is a discussed for a follow PR!

pradeepcrao: They would have to be stored as uniqe_ptrs to base class. (We can discuss in said follow-up PR)

source/common/upstream/subset_lb.h
DiazAlan: Done

source/common/upstream/upstream_impl.h
DiazAlan: Done

source/common/upstream/subset_lb.cc
DiazAlan: Done

adisuissa (approved):
LGTM, thanks!

source/common/upstream/upstream_impl.h
wbpcode: Usually, 0 or 1 of them will be set.

wbpcode: `variant` may be a good choice. But not sure if it's valueable to use it just for 20-30 bytes memory saving.

wbpcode:
LGTM overall, but **No merge this before 1.25.0 is released**.

ggreenway (approved):
LGTM. As per previous comment, waiting until post 1.25-release to merge.

source/common/upstream/upstream_impl.h
yurykats: 20-30 bytes quickly accumulate when you have tens of thousands of clusters. Any savings is valuable!

PR #24906, opened 2023-01-12
mobile: skipping command line parsing for E-M start up

alyssawilk:
This is a slightly cleaner code path setting options directly, rather than turning into command line options and back, and may provide a minor start up boost.

Risk Level: n/a for Envoy, medium for E-M
Testing: options test verifies options remain consistent 
Docs Changes: n/a
Release Notes: n/a


alyssawilk:
Ok unit tests fixed up - PTAL!

source/server/options_impl.h
```diff
@@ -54,6 +54,9 @@ class OptionsImpl : public Server::Options, protected Logger::Loggable<Logger::I
   OptionsImpl(const std::string& service_cluster, const std::string& service_node,
               const std::string& service_zone, spdlog::level::level_enum log_level);
 
+  // Constructor for mobile
+  OptionsImpl() {}
```
RyanTheOptimist, line 58: nit: = default is, I think, preferred though I honestly don't remember why.

RyanTheOptimist (dismissed):
Wow, this is great! I'm surprised that it's this simple.

RyanTheOptimist (dismissed):

jpsim:
Could this assertion error in the Swift tests have anything to do with this change?

> [external/envoy/source/extensions/listener_managers/listener_manager/listener_impl.cc:119] assert failure: sockets_.size() == num_sockets.

jpsim (approved):
👍 if we can confirm that the Swift tests assertion failure is unrelated.

jpsim:
Hmmm, the kotlin tests have some concerning failures too: https://github.com/envoyproxy/envoy/actions/runs/3959781241/jobs/6783722362

RyanTheOptimist (approved):

alyssawilk:
Thanks for the call-out on swift/kotlin @jpsim it was an indeed a bug fixed and regression-tested with ConcurrencyZeroIsOne
This is now passing all E-M builds so I'm going to enable auto-merge.

jpsim (approved):

PR #24907, opened 2023-01-12
remove include_ifaddrs from Envoy Mobile build config

RyanTheOptimist:
Remove include_ifaddrs from Envoy Mobile build config
The code which uses INCUDE_IFADDRS is android specific so we can just key off of Android instead of adding a layer of indirection which must be true on Android and false elsewhere.

### DIFF
--- a/mobile/library/common/network/android.cc
+++ b/mobile/library/common/network/android.cc
@@ -29,7 +29,7 @@ namespace {
 #endif
 
 void setAlternateGetifaddrs() {
-#if defined(__ANDROID_API__)
+#if defined(__ANDROID_API__) && __ANDROID_API__ < 24
   auto& os_syscalls = Api::OsSysCallsSingleton::get();
   ENVOY_BUG(!os_syscalls.supportsGetifaddrs(),
             "setAlternateGetifaddrs should only be called when supportsGetifaddrs is false");
--- a/mobile/library/common/network/android.cc
+++ b/mobile/library/common/network/android.cc
@@ -19,7 +19,7 @@ namespace Network {
 namespace Android {
 namespace Utility {
 
-#if defined(INCLUDE_IFADDRS)
+#if defined(__ANDROID_API__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wold-style-cast"
 namespace {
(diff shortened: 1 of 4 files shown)


mobile/library/common/network/android.cc
```diff
@@ -29,7 +29,7 @@ namespace {
 #endif
 
 void setAlternateGetifaddrs() {
-#if defined(INCLUDE_IFADDRS)
+#if defined(__ANDROID_API__)
   auto& os_syscalls = Api::OsSysCallsSingleton::get();
   ENVOY_BUG(!os_syscalls.supportsGetifaddrs(),
```
alyssawilk, line 34: Do we also want the min version or to change the bug call here?

RyanTheOptimist: Hm. Which min version?

alyssawilk: 24?  
https://github.com/envoyproxy/envoy/blob/main/source/common/api/posix/os_sys_calls_impl.cc#L352
/wait

Augustyniak: I just want to point out that this is a compilation (and not a runtime check) and that we effectively always use a custom (or 'alternate' as in the name of the method) implementation of `getifaddrs` as Android is compiled against `ndk_api_level` equal to 21. https://github.com/envoyproxy/envoy/blob/c394382139010828c125a23eccb47a5172e2eab8/mobile/WORKSPACE#L96

And yea, I think that @alyssawilk has a good point when it comes to the min version check - we probably do not want to use the custom implementation if the system one is available.

RyanTheOptimist: Ah, I see. Ok, I've moved the min version check here.

That being said, I wonder if it would actually make more sense to move this Envoy Mobile alternate implementation out of Envoy Mobile and into Envoy itself in OsSysCallsImpl::supportsGetifaddrs(). This would resolve the TODO there:

 ```
 // TODO: eliminate this branching by upstreaming an alternative Android implementation
// e.g.: https://github.com/envoyproxy/envoy-mobile/blob/main/third_party/android/ifaddrs-android.h
```

Would that be a good idea?

RyanTheOptimist: I took a look at this and I think this is the "right" solution. Here's PR which does the needful. What do y'all think?

https://github.com/envoyproxy/envoy/pull/24983

alyssawilk (approved):

mobile/library/common/network/android.cc
Augustyniak: I approved the other PR - let us know what you what to do with this one given that there is an overlap between it and https://github.com/envoyproxy/envoy/pull/24983

PR #24915, opened 2023-01-12
[Envoy-mobile]add enums in Cronvoy to match Cronet interface

RenjieTang:

Commit Message: add enums in Cronvoy to match Cronet interface
Additional Description: In Cronet, those java enums are generated from the native code as such: https://source.chromium.org/chromium/chromium/src/+/main:net/BUILD.gn;l=1724. In Cronvoy, we don't have such native code. Thus we need to manually create the enum classes so that Cronvoy customers get the full API.
Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


RyanTheOptimist (approved):
AH! That's the mystery solved. Thanks for doing this!

PR #24919, opened 2023-01-12
mobile: retry starting emulator up to 3 times

jpsim:
Commit Message:
Additional Description:
Risk Level: Low, CI only
Testing: CI
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -60,7 +60,7 @@ jobs:
       if: steps.should_run.outputs.run_ci_job == 'true'
       name: 'Start emulator'
       with:
-        timeout_minutes: 10
+        timeout_minutes: 6
         max_attempts: 3
         command: cd mobile && ./ci/mac_start_emulator.sh
     # Return to using:
@@ -105,7 +105,7 @@ jobs:
       if: steps.should_run.outputs.run_ci_job == 'true'
       name: 'Start simulator'
       with:
-        timeout_minutes: 10
+        timeout_minutes: 6
         max_attempts: 3
         command: cd mobile && ./ci/mac_start_emulator.sh
     # Return to using:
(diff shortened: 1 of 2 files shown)


.github/workflows/android_build.yml
```diff
@@ -56,9 +56,13 @@ jobs:
     - run: cd mobile && ./ci/mac_ci_setup.sh --android
       if: steps.should_run.outputs.run_ci_job == 'true'
       name: 'Install dependencies'
-    - name: 'Start simulator'
+    - uses: nick-fields/retry@943e742917ac94714d2f408a0e8320f2d1fcafcd
       if: steps.should_run.outputs.run_ci_job == 'true'
-      run: cd mobile && ./ci/mac_start_emulator.sh
+      name: 'Start emulator'
+      with:
+        timeout_minutes: 10
```
Augustyniak, line 63: could we lower it? Not sure whether 5 is not too low but I think that we should be fine with something that's lower than 10. Maybe somewhere between 5 and 7?

jpsim: Took 4:41s here: https://github.com/envoyproxy/envoy/actions/runs/3905297829/jobs/6673475684

Let's try with 6.

.github/workflows/android_build.yml
```diff
@@ -97,9 +101,13 @@ jobs:
     - name: 'Install dependencies'
       if: steps.should_run.outputs.run_ci_job == 'true'
       run: cd mobile && ./ci/mac_ci_setup.sh --android
-    - name: 'Start simulator'
+    - uses: nick-fields/retry@943e742917ac94714d2f408a0e8320f2d1fcafcd
       if: steps.should_run.outputs.run_ci_job == 'true'
-      run: cd mobile && ./ci/mac_start_emulator.sh
+      name: 'Start simulator'
```
RyanTheOptimist, line 106: If it's not too much trouble, mind changing this from "simulator" to "emulator" like you did on line 61? (Here and the two others below)

RyanTheOptimist (commented):
Oh, this is very cool! Fingers crossed.

.github/workflows/android_build.yml
jpsim: Oops, thanks for catching the ones I missed.

Augustyniak (approved):
Fingers crossed for this helping us.

PR #24920, opened 2023-01-12
filter state: share once with upstream

kyessenov:
Commit Message: Add an option to avoid transitive sharing of the filter state objects with the upstream filter state.
Additional Description: Consider a chain of two internal listeners: decap (socket listener), HTTP listener (internal), encap (internal). A shared filter state from the decap listener is propagated to both the HTTP and the encap listener. That's not always desirable if the filter state is processed in the HTTP listener because it forces additional pools for the connections between the HTTP and encap listener.  This option allows copying once from the decap to the HTTP listener without affecting the pooling between the HTTP and encap listeners. 
Risk Level: low
Testing: unit
Docs Changes: none
Release Notes: none

cc  @yanjunxiang-google


envoy/stream_info/filter_state.h
```diff
@@ -75,6 +75,11 @@ class FilterState {
     // upstream connections. Note that this affects connection pooling,
     // preventing any re-use of the upstream connections in the worst case.
     SharedWithUpstreamConnection,
+
+    // Same as SharedWithUpstreamConnection, except that the filter state is
+    // not transitively shared. The filter state is imported into the upstream
+    // connection filter state as exclusive to the upstream connection.
+    SharedWithUpstreamConnectionOnce,
```
yanjunxiang-google, line 82: So, this means with loop back listeners as 1 (decap) -> 2 -> 3 (encap),  the filterstate sharing is only available to 2?

yanjunxiang-google: Another question is that if the config only has one loop back listeners like 1(decap) -> 2(encap), then this change is no-op, right?

source/common/network/connection_impl.cc
```diff
@@ -927,9 +927,13 @@ ClientConnectionImpl::ClientConnectionImpl(
   if (transport_options) {
     for (const auto& object : transport_options->downstreamSharedFilterStateObjects()) {
       // This does not throw as all objects are distinctly named and the stream info is empty.
-      stream_info_.filterState()->setData(object.name_, object.data_, object.state_type_,
-                                          StreamInfo::FilterState::LifeSpan::Connection,
-                                          object.stream_sharing_);
+      stream_info_.filterState()->setData(
```
yanjunxiang-google, line 930: Could you please extend the current internal_upstream_integration_test.cc 

https://github.com/envoyproxy/envoy/blob/57f4a09ea7dc1ddc56bd9463a53ed9a09d336805/test/extensions/transport_sockets/internal_upstream/internal_upstream_integration_test.cc#L118

 to add integration tests to test :SharedWithUpstreamConnectionOnce. Like maybe having another filter like this but use SharedWithUpstreamConnectionOnce?

https://github.com/envoyproxy/envoy/blob/57f4a09ea7dc1ddc56bd9463a53ed9a09d336805/test/integration/filters/header_to_filter_state.cc#L32

yanjunxiang-google:
Thanks for working on this! LGTM in general except a few nit comments.

envoy/stream_info/filter_state.h
kyessenov: Yes to both questions.

source/common/network/connection_impl.cc
kyessenov: Good idea, updated the integration test to cover both cases.

yanjunxiang-google: Thanks for adding the test to cover "SharedWithUpstreamConnectionOnce" behaves the same as "SharedWithUpstreamConnection" in the case of 1->2.  

I am looking for a test to cover the filter state is not available in 3) in the case of chained loopback configuration, like 1->2->3 with "SharedWithUpstreamConnectionOnce".   Does such test already exist in your tests but I missed?

kyessenov: Oh, that doesn't exist. Will need to add a setup with multiple internal listeners.

yanjunxiang-google:
LGTM

source/common/network/connection_impl.cc
kyessenov: Thanks for holding me to higher standards. Found a bug in the implementation as I wrote the negative integration test.

yanjunxiang-google (approved):

kyessenov:
Gentle ping @snowp.

snowp (approved):
Thanks this LGTM

PR #24921, opened 2023-01-12
Avoid a race condition in CancelProofEnvoyStreamTest

RyanTheOptimist:
Avoid a race condition in CancelProofEnvoyStreamTest by creating the CountDownLatch before starting the thread which uses it to countdown.


jpsim (approved):

Augustyniak (dismissed):
Thank you!

RyanTheOptimist:
> Thank you!

You're welcome. Thank YOU for doing all the hard work to identify why this was timing out!

jpsim (approved):

PR #24923, opened 2023-01-12
tls: Control enforce_rsa_key_usage

tyxia:
Background : Boring SSL team is going to set `enforce_rsa_key_usage` to true very soon. If it is true, the handshake will fail if the keyUsage extension is present and incompatible with the TLS usage. However, the backend services/VMs might not be ready for this change and it had caused outage. I think this is also applicable to OSS Envoy customer since their certificate may not be ready as well. 

Change:
- Add the config field to control `enforce_rsa_key_usage`. It is false by default now but can be changed to true (which is aligned with Boring SSL's request) later once the customers are ready.
- Set it when ClientContext's SSL object is created. This ssl object will be used later in ssl handshake.
- It is added in `upstreamTlsContext` proto and set in `ClientContext` because this change in Boring SSL only affects Envoy->Backend (Upstream TLS) but not Client-> Envoy (Downstream TLS)
- Add stats to track/report the invalid use case by leveraging SSL_was_key_usage_invalid API introduced [here](https://github.com/google/boringssl/commit/a614d46d40509ea2f0c10d005972a08909c32b8c)
- Improve the error handling/report for `SSL_ERROR_SYSCALL`


api/envoy/extensions/transport_sockets/tls/v3/tls.proto
```diff
@@ -53,6 +54,9 @@ message UpstreamTlsContext {
   //
   // Defaults to 1, setting this to 0 disables session resumption.
   google.protobuf.UInt32Value max_session_keys = 4;
+
+  // If true, the handshake will fail if the keyUsage extension is present and incompatible with the TLS usage.
```
mattklein123, line 58: Is this using a WKT because we eventually want to change the default to true? Or will the default always be false? Either way I would specify the default is false.

Additionally, it would be optimal if we could warn somehow if using a certificate where this would fail and the setting is false? Or potentially warn if the certs are OK and this could be set to true?

mattklein123 (changes requested):
A high level comment. Also this needs test, thanks.

/wait

api/envoy/extensions/transport_sockets/tls/v3/tls.proto
tyxia: Exactly. As i mentioned in the PR description, the default will be changed to true eventually. But yes, I should note it in the comment. Done.

Regarding warning, there is a flag `was_key_usage_invalid` that will be set to TRUE [here](https://github.com/google/boringssl/blob/80a243e07ef77156af66efa7d22ac35aba44c1b3/ssl/handshake_client.cc#L1399) when certificate would fail but the error was skipped because the setting is false. So, we can set the enforcement to true when we don't see this flag to ever flipped to true

tyxia:
> > * It is false by default now but can be changed to true (which is aligned with Boring SSL's request) later once the customers are ready.
> 
> If you want to switch the default, you'll need a way to warn users that their cert isn't compatible, and leave that warning in for a couple releases. Is there a way to query BoringSSL to ask whether a given cert would have triggered an error, but still allow using the cert (assuming the value of the new config param is false)?
> 
> /wait-any

 Flag `was_key_usage_invalid` that will be set to TRUE [here](https://github.com/google/boringssl/blob/80a243e07ef77156af66efa7d22ac35aba44c1b3/ssl/handshake_client.cc#L1399) when the certificate would fail(due to mismatch between key and tls usage) but the error is skipped because this enforcement is disabled.  I think this can be used to warn the users about their incompatible certificate.

tyxia:
@mattklein123 @ggreenway  Thanks for the review! sorry about the slow response as there were some other priorities on our end. I think I have addressed all the comments. PTAL.

Also, I try to improve the error handling/reporting for `SSL_ERROR_SYSCALL` as underlying transport does not participate in the error queue(i.e., the failure reason is not recorded in drainErrorQueue).
But there is a windows-specific failure that failure reason of `server_connection` is still empty, which I guess is caused by TLS/SSL behavior difference on windows platform.  I skipped this specific check on windows because `client_connect (Envoy->backend)` is really the connection in interest and where the `rsa_key_enforcement` in this PR makes the difference.

changelogs/current.yaml
```diff
@@ -63,6 +63,9 @@ new_features:
 - area: upstream
   change: |
     added :ref:`maglev extension <envoy_v3_api_msg_extensions.load_balancing_policies.maglev.v3.Maglev>` to suppport the :ref:`load balancer policy <envoy_v3_api_field_config.cluster.v3.Cluster.load_balancing_policy>`.
+- area: tls
+  change: |
+    added support to configure the ``enforce_rsa_key_usage`` of SSL_CONFIG. This can be used to override its configuration in BoringSSL.
```
mattklein123, line 68: Please ref link to the new configuration option. Also duplicate the text that I suggested for the warning on configs that would fail.

envoy/network/transport_socket.h
```diff
@@ -191,6 +191,13 @@ class TransportSocket {
...
   virtual void configureInitialCongestionWindow(uint64_t bandwidth_bits_per_sec,
                                                 std::chrono::microseconds rtt) PURE;
+
+  /**
+   * @return optional bool whether the SSL_ERROR_SYSCALL has occurred on the transport socket.
+   * By default, it returns absl::nullopt. The class overrides this method will return the value
+   * based on its own implementation.
+   */
+  virtual absl::optional<bool> sslSyscallErrorOccurred() { return absl::nullopt; }
 };
```
mattklein123, line 201: This seems like a pretty substantial abstraction leak in this interface. Why is this needed?

mattklein123 (changes requested):
A few comments. We still need the warning for when a cert config would fail if the option were set to true. Thank you.

/wait

source/common/network/connection_impl.cc
```diff
@@ -715,6 +715,12 @@ void ConnectionImpl::onWriteReady() {
     // It is possible (though unlikely) for the connection to have already been closed during the
     // write callback. This can happen if we manage to complete the SSL handshake in the write
     // callback, raise a connected event, and close the connection.
+    ENVOY_CONN_LOG(debug, "remote close", *this);
```
ggreenway, line 718: Please either remove this or change it to `trace` level

source/common/network/connection_impl.cc
```diff
@@ -715,6 +715,12 @@ void ConnectionImpl::onWriteReady() {
...
+    if (transport_socket_->sslSyscallErrorOccurred().has_value() &&
+        transport_socket_->sslSyscallErrorOccurred().value()) {
+      setFailureReason(
```
ggreenway, line 721: Instead of adding `sslSyscallErrorOccurred()` to the interface, could the specific transport socket call this when an error occurs? Adding an ssl-specific bit to the transport socket interface is not ideal.

test/extensions/transport_sockets/tls/ssl_socket_test.cc
```diff
@@ -6902,6 +6912,50 @@ TEST_P(SslSocketTest, AsyncCustomCertValidatorFails) {
...
+  // Both server connection (Client-> Envoy) and client connection (Envoy->Backend) are expected to
+  // be successful.
+  TestUtilOptionsV2 test_options(listener, client_tls_context, true, version_);
+  testUtilV2(test_options);
+
+  // Enable the rsa_key_usage enforcement.
+  client_tls_context.mutable_enforce_rsa_key_usage()->set_value(true);
+  // The transport failure reason is empty for server connection on windows platform, which will
+  // cause the test failure.
+#if defined(WIN32)
```
ggreenway, line 6944: Why is this different on windows?

ggreenway (changes requested):
You'll need a log message that specifically details what is wrong with the certificate, and which config setting it is related to, and at least some vague instructions of how to generate a certificate that will comply with the new validation.

Also, @mattklein123 should the failure case increment the metric for deprecated feature use? I think we're basically creating this new config default setting as a deprecated value (since the default is intended to change in the future).

/wait

mattklein123:
> Also, @mattklein123 should the failure case increment the metric for deprecated feature use? I think we're basically creating this new config default setting as a deprecated value (since the default is intended to change in the future).

Yup. Sorry I should have mentioned this, but it goes along with the warning which we both commented on.

tyxia:
> > Also, @mattklein123 should the failure case increment the metric for deprecated feature use? I think we're basically creating this new config default setting as a deprecated value (since the default is intended to change in the future).
> 
> Yup. Sorry I should have mentioned this, but it goes along with the warning which we both commented on.

Regarding the warning, SSL_was_key_usage_invalid() api is not currently available in Envoy because it was [added](https://github.com/google/boringssl/commit/a614d46d40509ea2f0c10d005972a08909c32b8c) in Dec 2022 but currently Envoy is using boring ssl from [July 2022](https://github.com/envoyproxy/envoy/blob/0b89c597a5443eafa5a7e26cc5a0840b89a905d2/bazel/repository_locations.bzl#L117). 

I have tested the stats metric/log in Google internal Envoy though and it works as expected. Thus, I left a TODO there. 
Also, I am happy to help update the boring ssl version if you think we are ready and if you can share me any notes that I need to pay attention.

mattklein123:
> I have tested the stats metric/log in Google internal Envoy though and it works as expected. Thus, I left a TODO there.
Also, I am happy to help update the boring ssl version if you think we are ready and if you can share me any notes that I need to pay attention.

I don't think we should merge this without the warning, otherwise I think it's fairly confusing to the users. I think we should bump boringssl if we need to to get this change? But can we do that in a different change?

ggreenway:
@tyxia it would be great if you're willing to update boringssl to a version that has the needed function; preferably do that as a separate PR. We can merge that, then you can add the needed check to this PR. Are you ok with doing that?

tyxia:
@ggreenway and @mattklein123  Sure thing, I will bump the boringssl version first. Separate PR is also my preferred approach.

ggreenway:
@tyxia are these boringSSL APIs available in our FIPS version of boringssl?

tyxia:
> @tyxia are these boringSSL APIs available in our FIPS version of boringssl?

I don't think FIPS has. FIPS was update in [8/8/2019](https://github.com/envoyproxy/envoy/blob/0b89c597a5443eafa5a7e26cc5a0840b89a905d2/bazel/repository_locations.bzl#L132) and this API was merged in [12/8/2022](https://github.com/google/boringssl/commit/a614d46d40509ea2f0c10d005972a08909c32b8c)

Also, I don't think linux/beta has it yet as boringSSl version in [DEPS](https://chromium.googlesource.com/chromium/src/+/refs/tags/110.0.5481.52/DEPS) points to commit from [12/6/2022](https://github.com/google/boringssl/commit/7b00d84b025dff0c392c2df5ee8aa6d3c63ad539) while this API is from 12/8/2022

So, @mattklein123 and @ggreenway  I think I will have to wait for future release to bump boringSSL version then update this PR, if certificate warning is must to have.

mattklein123:
I think without the warning it's pretty difficult to perform any kind of default switch here. I suppose we can add the warning later but we absolutely can't change the default until we have the warning and the deprecated stat for multiple releases. So if you want to add TODOs around that that's OK with me.

tyxia:
@ggreenway and @mattklein123 . `Hold off this PR` OR `Check it in` either works for me.

The key here I think is that: we need to check in this PR **before** bumping boringSsl to the version that enforce the rsa_key_usage.  In other words, we need to disable `rsa_key_usage` via this PR.

Regarding the warning, the API for it in boringSSL (I mentioned before) will be available **earlier** than that enforcement becomes effective. So I will make sure the warning in place before bumping to boringSSL has that enforcement.

ggreenway:
/assign @envoyproxy/api-shepherds 

Can an API shepherd comment on how we'll handle the deprecation of the old value of the setting? This will be a change in a default of a BoolValue, after some deprecation period. Is it good enough to have a new metric that is incremented in the offending case, or does it need to specifically increment `deprecated_feature_use`?

adisuissa:
> Can an API shepherd comment on how we'll handle the deprecation of the old value of the setting? This will be a change in a default of a BoolValue, after some deprecation period. Is it good enough to have a new metric that is incremented in the offending case, or does it need to specifically increment `deprecated_feature_use`?

AFAIK the deprecated_feature_use was introduced to count uses of a deprecated field, not a specific offending value of a field. IIUC the proposed counter tracks a specific use-case that may happen in runtime, and that should continue to be tracked regardless of whether the default value of the feature is true or false. FWIW, it may be better to rename the stat a bit, as `was_key_usage_invalid` seems to be more general than invalid RSA key usage.

Re changing the default value, I suggest having that as a runtime-guarded change. The default-value of wrapped types is not considered to be stable across versions, and these fields should be explicitly configured by the servers.

tyxia:
@ggreenway , Based on @adisuissa 's input on `deprecated_feature_use `(which is used for a deprecated field), I think it is not applicable here. The flip of default value should only happens later once the all (most) Envoy clients are compliant to the this rsa_key usage. We can use runtime guard, or set it to true in the code to change that. We probably don't want to deprecate the field as the customer can still has control over this if they don't want to enforce it.

Regarding the name of stats, it comes from API `SSL_was_key_usage_invalid` that is designed to report this. I can update the stat name to `was_rsa_key_usage_invalid` if it is strongly preferred. 

Please take another look and let me what do you think. Thanks!

ggreenway:
> Re changing the default value, I suggest having that as a runtime-guarded change. The default-value of wrapped types is not considered to be stable across versions, and these fields should be explicitly configured by the servers.

This is somewhat of a corner case, but in the past we've always considered changing a default boolean value like this to be a breaking change. I agree that when we change it, it should get a runtime guard; that's what we've done in the past for similar changes.

api/envoy/extensions/transport_sockets/tls/v3/tls.proto
```diff
@@ -53,6 +54,11 @@ message UpstreamTlsContext {
   //
   // Defaults to 1, setting this to 0 disables session resumption.
   google.protobuf.UInt32Value max_session_keys = 4;
+
+  // Currently, the default value is false, but it is expected to be changed to true by default in a future release.
```
ggreenway, line 58: Please document what this setting does, and what conditions result in invalid rsa key usage.

docs/root/_include/ssl_stats.rst
```diff
@@ -18,3 +18,4 @@
    curves.<curve>, Counter, Total successful TLS connections that used ECDHE curve <curve>
    sigalgs.<sigalg>, Counter, Total successful TLS connections that used signature algorithm <sigalg>
    versions.<version>, Counter, Total successful TLS connections that used protocol version <version>
+   was_key_usage_invalid, Counter, Total successful TLS connections that used protocol version has invalid RSA key usage
```
ggreenway, line 21:
```suggestion
   was_key_usage_invalid, Counter, Total successful TLS connections that used an invalid RSA key usage
```

Can you link to some docs (either other envoy docs, or boringssl docs) that describe what this means?

source/common/network/connection_impl.cc
```diff
@@ -715,6 +715,10 @@ void ConnectionImpl::onWriteReady() {
     // It is possible (though unlikely) for the connection to have already been closed during the
     // write callback. This can happen if we manage to complete the SSL handshake in the write
     // callback, raise a connected event, and close the connection.
+    if (transport_socket_->ssl() != nullptr &&
+        transport_socket_->ssl()->sslSyscallErrorOccurred()) {
+      setFailureReason(absl::StrCat("The connection has been closed during the write callback."));
```
ggreenway, line 720: `The connection has been closed during the write callback` isn't a reason that the end user can easily understand. Can you make this more descriptive regarding things that may cause this to happen?

source/extensions/transport_sockets/tls/ssl_handshaker.cc
```diff
@@ -98,6 +98,12 @@ Network::PostIoAction SslHandshakerImpl::doHandshake() {
     case SSL_ERROR_WANT_CERTIFICATE_VERIFY:
       state_ = Ssl::SocketState::HandshakeInProgress;
       return PostIoAction::KeepOpen;
+    case SSL_ERROR_SYSCALL:
```
ggreenway, line 101: Is this change needed for error handling to correctly increment `was_key_usage_invalid_`? Or is it an unrelated change?

test/extensions/transport_sockets/tls/test_data/bad_rsa_key_usage_cert.pem
```diff
@@ -0,0 +1,24 @@
+-----BEGIN CERTIFICATE-----
```
ggreenway, line 1: Generation of this cert should be added to `certs.sh`.

ggreenway (changes requested):
/wait

source/common/network/connection_impl.cc
tyxia: Here is the generic prefix for failure reason, more descriptive info comes from underlying SSL lib via `transport_socket_->failureReason` when the final error message is constructed [here](https://github.com/envoyproxy/envoy/blob/020de5c9838729acc007b4748ecae63331182c5b/source/common/network/connection_impl.cc#L560) . An example in this PR is that `KEY_USAGE_BIT_INCORRECT` is appended in the final error message.  

If you still feel it is not descriptive enough, I am wonder do you have suggested description?

tyxia (commented):
Thanks all for the review and inputs! PTAL

source/common/network/connection_impl.cc
ggreenway: This error gets surfaced in access logs (and other places), but this string is specific to the implementation and isn't easily understood by users.

test/extensions/transport_sockets/tls/test_data/bad_rsa_key_usage_cert.cfg
```diff
@@ -0,0 +1,38 @@
...
+organizationalUnitName = Lyft Engineering
+organizationalUnitName_default = Lyft Engineering
+commonName = Test Server
+commonName_default = Test Server
+commonName_max  = 64
+
+[v3_req]
+basicConstraints = CA:FALSE
+keyUsage = nonRepudiation, digitalSignature, keyEncipherment
+# keyUsage = nonRepudiation, keyEncipherment
```
ggreenway, line 23: remove commented out line

test/extensions/transport_sockets/tls/test_data/bad_rsa_key_usage_cert.cfg
```diff
@@ -0,0 +1,38 @@
...
+extendedKeyUsage = clientAuth, serverAuth
+subjectKeyIdentifier = hash
+subjectAltName = @alt_names
+
+[v3_ca]
+# delete critical,
```
ggreenway, line 29: Is this a TODO for the next line?

test/extensions/transport_sockets/tls/test_data/bad_rsa_key_usage_cert.cfg
```diff
@@ -0,0 +1,38 @@
+[req]
```
ggreenway, line 1: Please add a comment in this file indicating exactly which field causes this to be an incorrect key usage.

ggreenway (changes requested):
This looks great overall; just a few small things

/wait

ggreenway (dismissed):

source/extensions/transport_sockets/tls/ssl_handshaker.cc
```diff
@@ -98,6 +98,12 @@ Network::PostIoAction SslHandshakerImpl::doHandshake() {
...
+      // When SSL_ERROR_SYSCALL occurred, the underlying transport does not participate in the error
+      // queue, which means the failure reason is not recorded in drainErrorQueue. Thus, setting
+      // flag to here to report the error.
+      syscall_error_occurred_ = true;
+      ABSL_FALLTHROUGH_INTENDED;
```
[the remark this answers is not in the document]
tyxia: Good idea! Done.

test/extensions/transport_sockets/tls/test_data/bad_rsa_key_usage_cert.cfg
tyxia: It was a comment leftover. Deleted

adisuissa (commented):
/lgtm api

docs/root/_include/ssl_stats.rst
```diff
@@ -18,3 +18,4 @@
    curves.<curve>, Counter, Total successful TLS connections that used ECDHE curve <curve>
    sigalgs.<sigalg>, Counter, Total successful TLS connections that used signature algorithm <sigalg>
    versions.<version>, Counter, Total successful TLS connections that used protocol version <version>
+   was_key_usage_invalid, Counter, Total successful TLS connections that used an `invalid keyUsage extension <https://github.com/google/boringssl/blob/6f13380d27835e70ec7caf807da7a1f239b10da6/ssl/internal.h#L3117>`_. (This is not avaiable in BoringSSL FIPS yet due to `issue <https://github.com/envoyproxy/envoy/issues/28246>`_)
```
adisuissa, line 21: nit: s/issue <...>/issue #28246<...>/

adisuissa:
@mattklein123 seems that this PR has a change requested from you.
Can you take one more look to see if it is good for merge?

ggreenway (approved):

mattklein123 (approved):

PR #24934, opened 2023-01-13
http2: add fuzzer for decoding connection

robertfemmer:
This PR adds a fuzzer targeting the HTTP/2 codec(s), namely nghttp2 and oghttp2. As opposed to the http_codec_impl_fuzz_test, we directly target the decoder. The fuzz cases are made up of structured frames, which are translated to (mostly) valid HTTP/2 frames and so-called "junk" frames, which may contain arbitrary data. The corpus
consists of frames containing header keys and values from the envoy source as well as junk frames that represent mostly valid HTTP/2 frames.


test/common/http/http2/http2_connection.proto
```diff
@@ -0,0 +1,95 @@
...
+
+message Http2DataFrame {
+  bytes data = 1;
+  // Use only 255 bytes maximum
+  bytes padding = 2;
+}
+
+message Http2HeaderFrame {
+  uint32 stream_dependency = 1;
+  uint32 weigth = 2;
```
birenroy, line 14: spelling: weight?

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,314 @@
...
+    } else {
+      // Empty frame
+      return Http2Frame("");
+    }
+
+    Http2Frame result(streamid, type, flags);
+    result.appendData(payload);
+    result.adjustPayloadSize();
+    return result;
+  } else if (pb_frame.has_junk()) {
```
birenroy, line 195: Optionally, handle the "junk" case before the "h2frame" case, since it is short and easy to understand.

test/common/http/http2/http2_connection.proto
```diff
@@ -0,0 +1,95 @@
...
+  uint32 error_code = 2;
+  bytes debug_data = 3;
+}
+
+message Http2WindowUpdateFrame {
+  uint32 window_size_increment = 1;
+}
+
+message Http2ContinuationFrame {
+  test.fuzz.Headers headers = 1;
```
birenroy, line 59: Not sure how I feel about this, as the payload of a CONTINUATION frame is likely to be a subrange of a deflated header block, rather than a complete header block unto itself. I understand that we'd like to fuzz this frame type, so I don't have a better suggestion.

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,314 @@
...
+    }
+  }
+
+  void setPayloadSize(uint32_t size) {
+    data_[0] = (size >> 16) & 0xff;
+    data_[1] = (size >> 8) & 0xff;
+    data_[2] = size & 0xff;
+  }
+
+  void adjustPayloadSize() { setPayloadSize(data_.size() - HeaderSize); }
```
birenroy, line 48: Optionally: fixPayloadSize(), or rectifyPayloadSize().

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,314 @@
...
+    if (data_.size() >= HeaderSize) {
+      std::memcpy(&data_[5], &id, sizeof(id));
+    }
+  }
+
+  void appendData(absl::string_view data) { data_.insert(data_.end(), data.begin(), data.end()); }
+
+  const absl::string_view payload() const {
+    const char* d = reinterpret_cast<const char*>(data_.data());
+    return absl::string_view(d, data_.size());
```
birenroy, line 71: Shouldn't the return value omit the frame header?

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,314 @@
...
+      auto f = h2frame.data();
+      uint8_t padding_len = static_cast<uint8_t>(f.padding().size() & 0xff);
+      payload.push_back(static_cast<char>(padding_len));
+      payload.append(f.data());
+      payload.append(f.padding().c_str(), padding_len);
+    } else if (h2frame.has_headers()) {
+      type = 1;
+      auto f = h2frame.headers();
+      uint8_t padding_len = static_cast<uint8_t>(f.padding().size() & 0xff);
+      payload.push_back(static_cast<char>(padding_len));
```
birenroy, line 130: The padding length byte is supposed to only be added if the PADDED flag is set. Is it intentional that we always add the padding length byte, regardless of flag state?

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,314 @@
...
+    uint8_t flags = static_cast<uint8_t>(h2frame.flags() & 0xff);
+    uint8_t type;
+    std::string payload;
+    if (h2frame.has_data()) {
+      type = 0;
+      auto f = h2frame.data();
+      uint8_t padding_len = static_cast<uint8_t>(f.padding().size() & 0xff);
+      payload.push_back(static_cast<char>(padding_len));
+      payload.append(f.data());
+      payload.append(f.padding().c_str(), padding_len);
```
birenroy, line 125: Why treat f.data() and f.padding() differently? Are we guaranteed that the fuzzer won't stick null characters into f.data()?

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,314 @@
...
+
+class Http2Harness {
+public:
+  Http2Harness() : server_settings(http2Settings()), client_settings(http2Settings()) {
+    ON_CALL(mock_server_callbacks, newStream(_, _))
+        .WillByDefault(Invoke(
+            [&](ResponseEncoder&, bool) -> RequestDecoder& { return orphan_request_decoder; }));
+  }
+
+  void fuzz_response(std::vector<Http2Frame>& frames) {
```
birenroy, line 224: Will this be used by (a) future test case(s)?

birenroy (commented):
Looks good overall, but I have a few questions/comments.

test/common/http/http2/http2_connection.proto
robertfemmer: Fixed the spelling, also the field was not in use, so I fixed that too.

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: I used the same name for the same method as in http2_frame.cc in the same directory, so I'll leave it like this.

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: I renamed the method to 'serialize' to better reflect that the method returns the wire format of the frame to feed to the decoder. The frame header needs to be included in the wire format.

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: I removed the dead code for now.

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: For the DATA case, I folded the padding into the data and let the mutator explore the parsing code.

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: For the HEADERS and PUSH_PROMISE the code now adds padding depending on the flag. Unconstrained fuzzing is still happening due to the junk frames.

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,307 @@
...
+    } else if (h2frame.has_priority()) {
+      type = 2;
+      auto f = h2frame.priority();
+      uint32_t stream_dependency = f.stream_dependency();
+      payload.append(reinterpret_cast<char*>(&stream_dependency), sizeof(stream_dependency));
+      payload.push_back(static_cast<char>(f.weight() & 0xff));
+    } else if (h2frame.has_rst()) {
+      type = 3;
+      auto f = h2frame.rst();
+      uint32_t error_code = f.error_code();
```
birenroy, line 159: Do we need calls to htonl() when serializing integers? (Here and below.)

birenroy (dismissed):
Just one final question. Looks good otherwise.

yanavlasov (dismissed):

test/common/http/http2/http2_connection_fuzz_test.cc
```diff
@@ -0,0 +1,307 @@
...
+
+namespace Envoy {
+namespace Http {
+namespace Http2 {
+namespace {
+
+static const uint32_t MAX_HEADERS = 32;
+static const size_t MAX_HD_TABLE_BUF_SIZE = 4096;
+
+class Http2Frame {
```
yanavlasov, line 25: My only suggestion would to be use this class if possible: https://github.com/envoyproxy/envoy/blob/beecb4c563edabbd61b867891b54ab4888be7a5d/test/common/http/http2/http2_frame.h#L31

Maybe move the functionality that you need into the existing class.

yanavlasov (commented):
/wait

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: Not necessarily, the fuzzer can generate integers in either byte order. However, it might make a crash case more (human-)readable.

test/common/http/http2/http2_connection_fuzz_test.cc
robertfemmer: Done.

yanavlasov (approved):

PR #24939, opened 2023-01-13
Allow C++ engine builder to use platform cert verification on iOS

RyanTheOptimist:
Allow C++ engine builder to use platform cert verification on iOS.
Remove PANIC from the engine builder since we do actually support platform cert verification on iOS and update test expectations accordingly.


jpsim:
Should this run some other kind of assertion on iOS? Or is it ok to just not assert anything there?

RyanTheOptimist:
> Should this run some other kind of assertion on iOS? Or is it ok to just not assert anything there?

Oh, hah! Actually this code shouldn't panic at all since we DO support cert verification on iOS.

Augustyniak (approved):

Issue #23596, opened 2022-10-20
envoy_reloadable_features_http2_new_codec_wrapper deprecation


alyssawilk:
Your change #20600 (Re-enables the HTTP/2 wrapper API by default.) introduced a runtime guarded feature. It has been 6 months since the new code has been exercised by default, so it's time to remove the old code path. This issue tracks source code cleanup so we don't forget.

PR #24943, opened 2023-01-13
http2: replaces nghttp2 functions with equivalents from QUICHE.

birenroy:
Signed-off-by: Biren Roy <birenroy@google.com>

Commit Message: Replaces nghttp2 functions with equivalents from QUICHE.
Additional Description: Cleanup related to https://github.com/envoyproxy/envoy/issues/23596
Risk Level: Low
Testing: ran unit and integration tests locally
Docs Changes:
Release Notes:
Platform Specific Features:


diannahu (approved):

source/common/http/header_utility.cc
```diff
@@ -190,17 +190,16 @@ bool HeaderUtility::schemeIsValid(const absl::string_view scheme) {
 }
 
 bool HeaderUtility::headerValueIsValid(const absl::string_view header_value) {
-  return nghttp2_check_header_value(reinterpret_cast<const uint8_t*>(header_value.data()),
-                                    header_value.size()) != 0;
+  return http2::adapter::HeaderValidator::IsValidHeaderValue(header_value,
```
RyanTheOptimist, line 193: Hm, do we need a runtime guard for this? I would imagine it's possible that this is a behavior change in the event that QUICHE and nghttp2 do something different?

RyanTheOptimist (commented):
Should we link to the GitHub issues in the PR description?

source/common/http/header_utility.cc
RyanTheOptimist: What do you think about a runtime guard for this change?

birenroy: With ObsTextOption::kAllow, the set of characters allowed by the oghttp2 method is exactly the same as the nghttp2 function.

birenroy:
@RyanTheOptimist I think this is ready for another look. I verified that the behavior is unchanged with the new method.

source/common/http/header_utility.cc
RyanTheOptimist: Hm. Can we write a unit test then where we loop through all the possible characters and verify we get the same output? I just want to make 100% sure that this change doesn't result in some crazy corner case behavior change which isn't protected by a runtime guard? Or alternatively, if it obvious from inspection can you point me at the two methods? (Sorry to be pedantic here!)

RyanTheOptimist: Thanks for the offline discussion and the pointer to the test.

RyanTheOptimist:
@zuercher for cross-company review.

source/common/http/header_utility.cc
```diff
@@ -190,17 +190,16 @@ bool HeaderUtility::schemeIsValid(const absl::string_view scheme) {
...
 }
 
 bool HeaderUtility::headerNameContainsUnderscore(const absl::string_view header_name) {
   return header_name.find('_') != absl::string_view::npos;
 }
 
 bool HeaderUtility::authorityIsValid(const absl::string_view header_value) {
-  return nghttp2_check_authority(reinterpret_cast<const uint8_t*>(header_value.data()),
-                                 header_value.size()) != 0;
+  return http2::adapter::HeaderValidator::IsValidAuthority(header_value);
```
zuercher, line 202: We've confirmed this is identical to nghttp as well?

birenroy: Looks like the QUICHE method rejects "@", which nghttp2 allows. I think this is slightly more adherent to [RFC 3986 section 3.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2).

zuercher: Because it's stricter, I think that means we should have a runtime guard for this behavior so that someone can revert it.

RyanTheOptimist: Agreed. @birenroy can you add a guard for this?

birenroy: Done.

wbpcode:
Please merge main and resolve the ci first, thanks.

wbpcode:
cc @zuercher could you give a review again when you have free time? Thanks.

source/common/http/BUILD
```diff
@@ -510,7 +510,7 @@ envoy_cc_library(
     srcs = ["header_utility.cc"],
     hdrs = ["header_utility.h"],
     external_deps = [
-        "nghttp2",
+        "quiche_http2_adapter",
```
zuercher, line 513: I'm by no means a bazel expert, but should this continue to reference nghttp2 until the runtime flag is removed?

zuercher (commented):
One question about deps, but otherwise lgtm.

birenroy:
> One question about deps, but otherwise lgtm.

Done.

birenroy:
@yanavlasov @KBaichoo I think this PR is ready for merge.

KBaichoo (approved):

PR #24947, opened 2023-01-14
mobile linker: add 'ldl' linker option

Augustyniak:
Commit Message: Without this option the app crashes when attempted to being launched on arm64 emulator. The `ldl` flag was removed in https://github.com/envoyproxy/envoy/pull/24806.
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/library/common/jni/BUILD
+++ b/mobile/library/common/jni/BUILD
@@ -107,7 +107,9 @@ cc_library(
 # Library which contains all the JNI related targets.
 cc_library(
     name = "envoy_jni_lib",
-    linkopts = select({
+    linkopts = [
+        "-ldl",
+    ] + select({
         "@envoy//bazel:dbg_build": ["-Wl,--build-id=sha1"],
         "//conditions:default": [],
     }),
@@ -133,7 +135,10 @@ config_setting(
 # Main dynamic library for the Envoy Mobile aar
 cc_binary(
     name = "libenvoy_jni.so",
-    linkopts = select({
+    linkopts = [
+        "-lm",
+    ] +
+     select({
         "@envoy//bazel:dbg_build": ["-Wl,--build-id=sha1"],
         "//conditions:default": [],
     }),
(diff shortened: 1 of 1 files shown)


mobile/library/common/jni/BUILD
```diff
@@ -107,7 +107,9 @@ cc_library(
 # Library which contains all the JNI related targets.
 cc_library(
     name = "envoy_jni_lib",
-    linkopts = select({
+    linkopts = [
```
Augustyniak, line 110: fixing

mobile/library/common/jni/BUILD
```diff
@@ -133,7 +133,9 @@ config_setting(
 # Main dynamic library for the Envoy Mobile aar
 cc_binary(
     name = "libenvoy_jni.so",
-    linkopts = select({
+    linkopts = [
+        "-ldl",
```
RyanTheOptimist, line 137: Let's add a comment here which explains why we need thisand mentions that it's not covered by CI so that we don't accidentally delete it again :). Maybe something like:

  Without this option the app crashes when attempted to being launched on arm64 emulator which is not covered by CI.

RyanTheOptimist (approved):

mobile/library/common/jni/BUILD
keith: Might as well figure out how to cover it on CI. It's a launch crash so I'm sure that coverage will help other things too

RyanTheOptimist: Agreed! Do you happen to know how plausible it would be to get arm64 coverage in CI?

keith: I assume so but I don't recall if we tried

Augustyniak: It _should_ be relatively easy - I assume we would need to specify an arm emulator in https://github.com/envoyproxy/envoy/blob/4fb3eeb1a12b9005133639e96425fef16816c2a4/mobile/ci/mac_start_emulator.sh#L5.

Augustyniak: [rafa](https://github.com/envoyproxy/envoy/issues/24974)

PR #24953, opened 2023-01-15
do not initiate parent shutdown sequence when hot restart is disabled

ramaraochavali:
Commit Message: do not initiate parent shutdown sequence when hot restart is disabled
Additional Description:
Risk Level: Low
Testing: N/A
Docs Changes: N/A
Release Notes: N/A


mattklein123:
Is there a reason fro this change? Is it just to avoid confusing logs? Thank you.

/wait-any

ramaraochavali:
Yes. Just to avoid confusing log lines.

mattklein123:
The log is in `void DrainManagerImpl::startParentShutdownSequence()` right? Perhaps just put an early return there to keep the logic closer to where you don't want the code to execute? Thanks.

/wait

Issue #24893, opened 2023-01-11
Envoy Mobile's RequestHeadersBuilder should be able to parse a URI


mjbshaw:
[`RequestHeadersBuilder::RequestHeadersBuilder()`](https://github.com/envoyproxy/envoy/blob/98ca3abcd0974c02e268e87059da835701adb781/mobile/library/cc/request_headers_builder.h#L19-L20) takes a `scheme`, `authority`, and `path` (nitpick: this should probably be named `path_and_query` or something so it's clearer to users this should include the query).

In several system APIs I've used, all I have access to is a URI. I can parse it into `scheme`, `authority`, and `path+query` pretty easily with the regex `^(?:([^:/?#]+):)?(?://([^/?#]*))?(.*)` (it's actually simple enough that this is reasonable to do without a full regex engine).

But it would be nice if `RequestHeadersBuilder` had a second constructor that just took the full URI and parsed it for me so I don't have to roll my own parsing code.

I mostly care about C++ here though I suppose this could apply to other languages; I haven't checked their current APIs.

PR #24955, opened 2023-01-15
Add a RequestHeadersBuilder constructor which takes a URL

RyanTheOptimist:
Add a RequestHeadersBuilder constructor which takes a URL

Fixes: #24893

Risk Level: Low
Testing: New unit test
Docs Changes: None
Release Notes: Updated version_history.rst

### DIFF
--- a/mobile/docs/root/intro/version_history.rst
+++ b/mobile/docs/root/intro/version_history.rst
@@ -20,6 +20,7 @@ Bugfixes:
 
 Features:
 
+- api: Add a constructor which takes a URL to C++ RequestEngineBuilder.
 - api: add option to support platform provided certificates validation interfaces on iOS and Android. (:issue `#2144 <2144>`)
 - api: Add a ``setPerTryIdleTimeoutSeconds()`` method to C++ EngineBuilder.
 - swift/kotlin: add an option to enable DNS cache by calling ``enableDNSCache(_:)`` method.
--- a/mobile/library/cc/BUILD
+++ b/mobile/library/cc/BUILD
@@ -83,6 +83,7 @@ envoy_cc_library(
         "//library/common/api:c_types",
         "//library/common/data:utility_lib",
         "//library/common/extensions/key_value/platform:config",
+	"@envoy//source/common/http:utility_lib",
     ],
 )
 
--- a/mobile/library/cc/request_headers_builder.cc
+++ b/mobile/library/cc/request_headers_builder.cc
@@ -5,28 +5,32 @@ namespace Platform {
 
 RequestHeadersBuilder::RequestHeadersBuilder(RequestMethod request_method, std::string scheme,
                                              std::string authority, std::string path) {
-  this->internalSet(":method", {requestMethodToString(request_method)});
-  this->internalSet(":scheme", {std::move(scheme)});
-  this->internalSet(":authority", {std::move(authority)});
-  this->internalSet(":path", {std::move(path)});
+  internalSet(":method", {requestMethodToString(request_method)});
+  internalSet(":scheme", {std::move(scheme)});
+  internalSet(":authority", {std::move(authority)});
+  internalSet(":path", {std::move(path)});
+}
+
+RequestHeadersBuilder::RequestHeadersBuilder(RequestMethod request_method, Envoy::Http::Utility::Url url)
+  : RequestHeadersBuilder(request_method, std::string(url.scheme()), std::string(url.hostAndPort()), std::string(url.pathAndQueryParams())) {
 }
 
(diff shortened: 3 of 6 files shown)


RyanTheOptimist:
@mjbshaw Would this work for you? In https://github.com/envoyproxy/envoy/issues/24893 you asked for the Builder to do the parsing for you, and I did that initially, but then the Builder had to handle the case where the URL failed to parse and it wasn't immediately obvious how to report that error. Instead this PR requires a URL to be passed in, but as you can see from the test, we have a one-liner for parsing URLs.

Augustyniak:
any plans to implement the same functionality in Swift and Kotlin `RequestHeadersBuilder`s? If not a part of this PR could we create a GH issue for this work?

mjbshaw:
> In https://github.com/envoyproxy/envoy/issues/24893 you asked for the Builder to do the parsing for you, and I did that initially, but then the Builder had to handle the case where the URL failed to parse and it wasn't immediately obvious how to report that error.

I'm not entirely convinced the builder would need to report an error if it was given a garbage URL. The builder currently doesn't give you any feedback if you give it a garbage scheme, authority, or path right now.

The regex `^(?:([^:/?#]+):)?(?://([^/?#]*))?(.*)$` matches _any_ string, including invalid URLs. If we use this pattern to parse the URL then we can parse any input and there will never be an error to report. True, if we feed it garbage the result won't be meaningful, but that's already the case with the current API and its separate scheme, authority, and path.

I guess my main complaint with this current approach is that it tries to move the validation requirement to the user, but ultimately Envoy still has to perform the same validation again later when it's internally handling the scheme, authority, and path. We now have two points of failure where technically we only really need one.

> Instead this PR requires a URL to be passed in, but as you can see from the test, we have a one-liner for parsing URLs.

That works, assuming `source/common/http/utility.h` has the same build visibility as `mobile/library/cc/request_headers_builder.h`. That said, it seems like a departure from the current API. The Envoy Mobile API is entirely self-contained (at least everything I've explored). It even re-exports `LogLevel` in its own API.

I'm not sure what Envoy Mobile's API encapsulation goals are but it might be worth double checking that this aligns with that.

RyanTheOptimist:
> any plans to implement the same functionality in Swift and Kotlin `RequestHeadersBuilder`s? If not a part of this PR could we create a GH issue for this work?

Sure, I'd be happy to do this for those builders as well. Once we sort out the C++ API, I'll either add to this PR or file an issue. Good idea!

RyanTheOptimist:
> > In #24893 you asked for the Builder to do the parsing for you, and I did that initially, but then the Builder had to handle the case where the URL failed to parse and it wasn't immediately obvious how to report that error.
> 
> I'm not entirely convinced the builder would need to report an error if it was given a garbage URL. The builder currently doesn't give you any feedback if you give it a garbage scheme, authority, or path right now.
> 
> The regex `^(?:([^:/?#]+):)?(?://([^/?#]*))?(.*)$` matches _any_ string, including invalid URLs. If we use this pattern to parse the URL then we can parse any input and there will never be an error to report. True, if we feed it garbage the result won't be meaningful, but that's already the case with the current API and its separate scheme, authority, and path.
> 
> I guess my main complaint with this current approach is that it tries to move the validation requirement to the user, but ultimately Envoy still has to perform the same validation again later when it's internally handling the scheme, authority, and path. We now have two points of failure where technically we only really need one.
> 
> > Instead this PR requires a URL to be passed in, but as you can see from the test, we have a one-liner for parsing URLs.
> 
> That works, assuming `source/common/http/utility.h` has the same build visibility as `mobile/library/cc/request_headers_builder.h`. That said, it seems like a departure from the current API. The Envoy Mobile API is entirely self-contained (at least everything I've explored). It even re-exports `LogLevel` in its own API.
> 
> I'm not sure what Envoy Mobile's API encapsulation goals are but it might be worth double checking that this aligns with that.

I see. I had initially hidden the construction of the URL inside the Builder's constructor and swallowed the URL parsing error. Since that's the API you prefer, I've switched it back. I had assumed that a caller would rather learn early that a URL was invalid than wait for a request to fail with the subsequent error, but this approach is fine with me.

PTAL

mjbshaw:
LGTM, thanks!

jpsim (approved):
+1 on the current bits, filing tickets for the Swift/Kotlin APIs could be good.

On the Swift side, I think the idiomatic way to do this would be to have `RequestHeadersBuilders` have an initializer that takes a [URL](https://developer.apple.com/documentation/foundation/url), maybe additionally another initializer that takes a [URLComponents](https://developer.apple.com/documentation/foundation/urlcomponents).

Issue #24219, opened 2022-11-27
SMTP Proxy network filter in envoy


VishalDamgude:
*Title*: *Add SMTP Proxy support in Envoy *

*Description*:
>Currently Envoy does not have SMTP protocol support. There is a starttls transport socket available in envoy which will be used to support STARTTLS command of SMTO protocol. Our use case is the TLS termination on envoy. So the traffic to upstream will be plain text.But this filter can be made generic to support TLS in downstream and upstream connection i.e. relay STARTTLS to upstream.

[optional *Relevant Links*:]
>https://www.rfc-editor.org/rfc/rfc3207.html

kyessenov:
CC @cpakulski (STARTTLS)

VishalDamgude:
This issue needs to be marked as in progress.I don't have permissions to do so.

jsbucy:
dup of #19765 ?

PR #24972, opened 2023-01-17
SMTP Network Filter in Envoy contrib

VishalDamgude:
Commit Message: Smtp Network Filter in contrib

Additional Description: 
This smtp network filter has been developed as per RFC 5321 (https://www.rfc-editor.org/rfc/rfc5321) and supports STARTTLS extension. It will parse SMTP commands and responses as per the SMTP protocol and provide protocol specific metrics.
The downstream starttls is supported by default. When smtp client sends STARTTLS, the filter will start secure transport on downstream connection. The upstream encryption is configurable. If enabled and if TLS handshake fails during upstream TLS negotiation, the filter responds to the client with error and closes the connection. If upstream TLS handshake succeeds, both the downstream and upstream connections are encrypted for secure smtp mail transfer.

Risk Level: Low
Testing: Unit, Manual testing
Docs Changes:
Release Notes:
[Optional Runtime guard:]
[Optional Fixes #Issue] https://github.com/envoyproxy/envoy/issues/24219
[Optional Deprecated:]


cpakulski:
@VishalDamgude It seems that DCO signature is missing. Please refer to https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#dco-sign-your-work. I suggest that you install hooks, so it is automatically added each time you do a commit.

VishalDamgude:
Hi @cpakulski @wbpcode ,
Will create another PR. Closing this one. It seems the branch is all messed up while adding DCO signature.

cpakulski:
I was about to suggest closing and opening a new one ... Thanks.

PR #24981, opened 2023-01-17
[file_system_http_cache] Add more features to API, and remove some not-implemented tags

ravenblackx:
Commit Message: [file_system_http_cache] Add more features to API, and remove some not-implemented tags
Additional Description:
  Add API for:
  * `evict_fraction`: evicting a fraction of the exceeded limits extra to reduce churn.
  * `max_eviction_frequency`: do an eviction pass before it's necessary, to ensure the stats aren't too out of date.
  * `min_eviction_frequency`: don't do an eviction pass even if limits are reached, to reduce churn.
  * `create_cache_path`: option to attempt to create the cache path if it does not exist.

  Remove not-implemented-hide for:
  * `max_cache_size_bytes`
  * `max_cache_entry_count`
  
  Clarify name for:
  * `max_cache_entry_size_bytes` -> `max_individual_cache_entry_size_bytes`: to avoid confusion with `max_cache_size_bytes`

Risk Level: None, extension is WIP and no change to existing behavior.
Testing: n/a
Docs Changes: Adds some generated docs.
Release Notes: n/a


api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -77,4 +74,46 @@ message FileSystemHttpCacheConfig {
...
+  // the on-disk state. This can be important if multiple instances are accessing the same
+  // cache in parallel.
+  //
+  // If an eviction pass has not happened within this duration, the eviction thread will
+  // be awoken and perform an eviction.
+  //
+  // If unset, there will be no eviction passes except those triggered by cache limits.
+  //
+  // [#not-implemented-hide:]
+  google.protobuf.Duration max_eviction_frequency = 8;
```
htuch, line 103: Technically it's a period, the reciprocal is frequency.

htuch:
@jmarantz @toddmgreer can you folks provide a domain specific review?

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
ravenblackx: Good call. Updated. (Though "every 10 minutes" is a frequency, there's no way to phrase that as a duration and variable name.)

RyanTheOptimist:
@jmarantz friendly maintainer on-call ping.

jmarantz (approved):
This looks great, thanks!

api/envoy/extensions/http/cache/file_system_http_cache/v3/file_system_http_cache.proto
```diff
@@ -77,4 +74,54 @@ message FileSystemHttpCacheConfig {
...
+
+  // The longest amount of time to wait before running a cache eviction pass. An eviction
+  // pass may not necessarily remove any files, but it will update the cache state to match
+  // the on-disk state. This can be important if multiple instances are accessing the same
+  // cache in parallel.
+  //
+  // If an eviction pass has not happened within this duration, the eviction thread will
+  // be awoken and perform an eviction pass.
+  //
+  // If unset, there will be no eviction passes except those triggered by cache limits.
```
htuch, line 100: I'm a little unclear after reading this what help there is from an eviction pass not triggered by cache limits. There's something above about syncing on-disk state, but I think this could be clearer.

ravenblackx: Added a top-level link to the design doc (which isn't actually there yet, but the [PR that adds it](https://github.com/envoyproxy/envoy/pull/24642) and this one are racing for who can be the slowest!)
Also elaborated with an example.

jmarantz:
I think I am good with this and this one is just waiting for @htuch API review and he can just merge.

htuch (approved):
Thanks1

PR #24983, opened 2023-01-17
Move the old-android implementation of getifaddrs from Envoy Mobile to Envoy code

RyanTheOptimist:
Move the old-android implementation of getifaddrs from Envoy Mobile to Envoy code
Before this PR, Envoy had an implementation of getifaddrs which only worked on relatively recent version of android (24, 25, 26, etc). It also had an API for an embedder to supply an alternate implementation which would be used instead of the core implementation. Meanwhile, Envoy Mobile had and implementation of getifaddrs which worked on older version of android (23, 22, 21, etc). This code moves that Envoy Mobile "alternate" implementation into core Envoy leaving Envoy with support for *any* version of Android. Also, since Envoy Mobile was the only user of Envoy's "alternate getifaddrs implementation" codepath, this PR also removes the now unused setAlternateGetifaddrs() method.

This is an alternative to #24907

Risk Level: Low
Testing: Unit
Docs Changes: N/A
Release Notes: N/A
Platform Specific Features: Provides a getifaddrs implementation for old versions of Android.


Augustyniak (approved):

source/common/api/posix/os_sys_calls_impl.cc
```diff
@@ -8,6 +8,15 @@
 #include "source/common/api/os_sys_calls_impl.h"
 #include "source/common/network/address_impl.h"
 
+#if defined(__ANDROID_API__) && __ANDROID_API__ < 24
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wold-style-cast"
+namespace android {
+#include "source/common/api/posix/android/ifaddrs-android.h"
+}  // namespace tmp
```
Augustyniak, line 16: should it say `namespace android`?

RyanTheOptimist: Oh, good point. Done.

RyanTheOptimist:
@alyssawilk now that this is FINALLY passing CI, PTAL

test/common/api/os_sys_calls_test.cc
```diff
@@ -72,7 +72,7 @@ TEST(OsSyscallsTest, SetAlternateGetifaddrs) {
   auto& os_syscalls = Api::OsSysCallsSingleton::get();
   const bool pre_alternate_support = os_syscalls.supportsGetifaddrs();
   Api::InterfaceAddressVector interfaces{};
-#if defined(WIN32) || (defined(__ANDROID_API__) && __ANDROID_API__ < 24)
```
alyssawilk, line 75: issetalternategetifsddrs used any more?  If not, can we remove?

alyssawilk (commented):
Amazing clean up!
LGTM modulo one question and a request to add a bit more explanation to the PR description for anyone else confused the way I was :-P

test/common/api/os_sys_calls_test.cc
RyanTheOptimist: Removed!

RyanTheOptimist: No, I don't think so. This doesn't change any API or behavior in Envoy Mobile itself (just an implementation detail about how this behavior is implemented). And while this does change code in envoy/api, this is not part of Envoy's "API" which is the protos.

RyanTheOptimist:
> Amazing clean up! LGTM modulo one question and a request to add a bit more explanation to the PR description for anyone else confused the way I was :-P

Description updated. PTAL!

alyssawilk (approved):
move LGTM.  trusting the moved code as it's well vetted on E-M

test/common/api/os_sys_calls_test.cc
jpsim: Thanks for clarifying, yes I was referring to this breaking Envoy's API, but it's good to know that Envoy's programmatic interface isn't considered a stable API, only the protos.

PR #19693, opened 2022-01-26
WiP stats: Reimplement existing /stats API in a way that can be chunked.

jmarantz:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

Issue #16139, opened 2021-04-23
Prometheus stats handler used too much memory.


cocotyty:
**If you are reporting *any* crash or *any* potential security issue, *do not*
open an issue in this repo. Please report the issue via emailing
envoy-security@googlegroups.com where the issue will be triaged appropriately.**

*Title*: *One line description*
Prometheus stats handler used too much memory.

*Description*:
![image](https://user-images.githubusercontent.com/2241753/115835141-8ae81c00-a448-11eb-8a6e-04f861162471.png)
![image](https://user-images.githubusercontent.com/2241753/115834975-5f653180-a448-11eb-8a66-f1411bb3d9a0.png)

>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

>**Note**: The [Envoy_collect tool](https://github.com/envoyproxy/envoy/blob/main/tools/envoy_collect/README.md)
gathers a tarball with debug logs, config and the following admin
endpoints: /stats, /clusters and /server_info. Please note if there are
privacy concerns, sanitize the data prior to sharing the tarball/pasting.

*Admin and Stats Output*:
>Include the admin output for the following endpoints: /stats,
/clusters, /routes, /server_info. For more information, refer to the
[admin endpoint documentation.](https://www.envoyproxy.io/docs/envoy/latest/operations/admin)

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Config*:
>Include the config used to configure Envoy.

*Logs*:
>Include the access logs and the Envoy logs.

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Call Stack*:
> If the Envoy binary is crashing, a call stack is **required**.
Please refer to the [Bazel Stack trace documentation](https://github.com/envoyproxy/envoy/tree/main/bazel#stack-trace-symbol-resolution).

cocotyty:
https://github.com/envoyproxy/envoy/blob/8a2143613d43d17d1eb35a24b4a4a4c432215606/source/server/admin/prometheus_stats.cc#L211 

Maybe use http chunks to response can fix this problem.

cocotyty:
@alyssawilk @lizan

alyssawilk:
cc @jmarantz

jmarantz:
[...]

PR #24998, opened 2023-01-18
Prom stats perf improvements

rulex123:
Commit Message: patch to add support for streaming/chunking of prometheus stats - basically adapts the algorithm introduced in https://github.com/envoyproxy/envoy/pull/19693 for prometheus stats (while keeping common parts of the algorithm in a shared class to avoid code duplication). 
With this patch we should be able to improve the issue described here https://github.com/envoyproxy/envoy/issues/16139

Risk Level: only underlying implementation changes, but otherwise no change of behaviour/features. So low risk I would say.
Testing: current tests should remain unchanged (though they might be moved around due to refactoring of code); new tests are also added.

Docs Changes: n/a
Release Notes: n/a
[Optional Runtime guard:]
[Optional Fixes #Issue]: https://github.com/envoyproxy/envoy/issues/16139

Draft PR (already went through some review) -> https://github.com/envoyproxy/envoy/pull/24716

Benchmark data from running `stats_handler_speed_test.cc` on a dev box:

BEFORE

```
---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
BM_AllCountersText                      744 ms          744 ms            1
BM_UsedCountersText                    76.5 ms         76.4 ms            9
BM_FilteredCountersText                2866 ms         2866 ms            1
BM_Re2FilteredCountersText              371 ms          371 ms            2
BM_AllCountersJson                      917 ms          917 ms            1
BM_UsedCountersJson                    79.0 ms         79.0 ms            9
BM_FilteredCountersJson                3087 ms         3086 ms            1
BM_Re2FilteredCountersJson              369 ms          369 ms            2
BM_AllCountersPrometheus               8125 ms         8124 ms            1
BM_UsedCountersPrometheus               334 ms          334 ms            2
BM_FilteredCountersPrometheus          3680 ms         3680 ms            1
BM_Re2FilteredCountersPrometheus       1022 ms         1022 ms            1
```

AFTER

```
---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
BM_AllCountersText                      742 ms          742 ms            1
BM_UsedCountersText                    77.6 ms         77.6 ms            9
BM_FilteredCountersText                2862 ms         2861 ms            1
BM_Re2FilteredCountersText              509 ms          508 ms            2
BM_AllCountersJson                     1559 ms         1558 ms            1
BM_UsedCountersJson                    81.0 ms         81.0 ms            9
BM_FilteredCountersJson                2857 ms         2856 ms            1
BM_Re2FilteredCountersJson              373 ms          373 ms            2
BM_AllCountersPrometheus               1842 ms         1842 ms            1
BM_UsedCountersPrometheus              92.4 ms         92.4 ms            8
BM_FilteredCountersPrometheus          2863 ms         2862 ms            1
BM_Re2FilteredCountersPrometheus        372 ms          372 ms            2
```


rulex123:
@jmarantz ready for review when you are :-) Thanks!

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+namespace Envoy {
+namespace Server {
+
+GroupedStatsRequest::GroupedStatsRequest(Stats::Store& stats, const StatsParams& params,
+                                         Stats::CustomStatNamespaces& custom_namespaces,
+                                         UrlHandlerFn url_handler_fn)
+    : StatsRequest(stats, params, url_handler_fn), custom_namespaces_(custom_namespaces),
+      global_symbol_table_(stats.constSymbolTable()) {
+
+  // the "type" query param is ignored for prometheus stats, so always start from
```
jmarantz, line 19: Hm.

I suggested renaming this GroupedStatsRequest and I'm fine with sticking with it, but right now there's prometheus-specific stuff in here -- type 'type' policy, the prometheus-specific text-readouts policy, maybe other things.

Maybe just leave a TODO to clean this up if we decide to build a grouped view in HTML or JSON.

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+    } else if (params_.re2_filter_ != nullptr &&
+               !re2::RE2::PartialMatch(stat->name(), *params_.re2_filter_)) {
+      return true;
+    }
+
+    // capture stat by either adding to a pre-existing variant or by creating a new variant
+    std::string tag_extracted_name = global_symbol_table_.toString(stat->tagExtractedStatName());
+    auto [iterator, inserted] = stat_map_.try_emplace(
+        tag_extracted_name, std::vector<Stats::RefcountPtr<StatType>>({stat}));
+    if (!inserted) {
```
jmarantz, line 54: the way you wrote this I think we might get a superfluous allocation when !inserted.

WDYT of inserting an empty vector and appending the stat unconditionally?

It might be slightly faster, both because of 1 less call to the memory allocator and one less branch.

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+  auto prefixed_tag_extracted_name = prefixedTagExtractedName<SharedStatType>(name);
+  if (prefixed_tag_extracted_name.has_value()) {
+    PrometheusStatsRender* const prometheus_render =
+        dynamic_cast<PrometheusStatsRender*>(render_.get());
+    // increment stats count
+    ++phase_stat_count_;
+
+    // sort group
+    std::vector<SharedStatType> group = absl::get<std::vector<SharedStatType>>(variant);
+    std::sort(group.begin(), group.end(),
```
jmarantz, line 89: perf nit: use SymbolTable::sortByStatName, which is slightly faster because it will hold the symbol table lock over the entire sorting operation, rather than taking the and releasing its lock for each comparison.

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+}
+
+template <class SharedStatType>
+void GroupedStatsRequest::renderStat(const std::string& name, Buffer::Instance& response,
+                                     const StatOrScopes& variant) {
+  auto prefixed_tag_extracted_name = prefixedTagExtractedName<SharedStatType>(name);
+  if (prefixed_tag_extracted_name.has_value()) {
+    PrometheusStatsRender* const prometheus_render =
+        dynamic_cast<PrometheusStatsRender*>(render_.get());
+    // increment stats count
```
jmarantz, line 84: I think you can remove the comment -- I think the code speaks for itself :)

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+    std::vector<SharedStatType> group = absl::get<std::vector<SharedStatType>>(variant);
+    std::sort(group.begin(), group.end(),
+              [this](const Stats::RefcountPtr<Stats::Metric>& stat1,
+                     const Stats::RefcountPtr<Stats::Metric>& stat2) -> bool {
+                return global_symbol_table_.lessThan(stat1->statName(), stat2->statName());
+              });
+
+    // render group
+    StatOrScopesIndex index = static_cast<StatOrScopesIndex>(variant.index());
+    std::string type = (index == StatOrScopesIndex::Counter) ? "counter" : "gauge";
```
jmarantz, line 97: ASSERT that it's a gauge if not a counter? It looks like you can get here with a text-readout; is that right?

This may be correct but it could use some comments.

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+                                           const StatOrScopes& variant) {
+  auto histogram = absl::get<std::vector<Stats::HistogramSharedPtr>>(variant);
+  auto prefixed_tag_extracted_name = prefixedTagExtractedName<Stats::HistogramSharedPtr>(name);
+
+  if (prefixed_tag_extracted_name.has_value()) {
+    // increment stats count
+    phase_stat_count_++;
+
+    // sort group
+    std::sort(histogram.begin(), histogram.end(),
```
jmarantz, line 131: SymbolTable::sortByStatNames

source/server/admin/grouped_stats_request.cc
```diff
@@ -0,0 +1,156 @@
...
+              [this](const Stats::RefcountPtr<Stats::Metric>& stat1,
+                     const Stats::RefcountPtr<Stats::Metric>& stat2) -> bool {
+                return global_symbol_table_.lessThan(stat1->statName(), stat2->statName());
+              });
+
+    // render group
+    response.add(fmt::format("# TYPE {0} {1}\n", prefixed_tag_extracted_name.value(), "histogram"));
```
jmarantz, line 138: I guess this is where a Render interface might be useful. Let's just call this out as prometheus-specific and it could be refactored incrementally later.

Also, you *do* have a file of prom rendering functions -- should we at least move this magic there?

source/server/admin/grouped_stats_request.h
```diff
@@ -0,0 +1,56 @@
...
+#include "source/server/admin/stats_request.h"
+#include "source/server/admin/utils.h"
+
+#include "absl/container/btree_map.h"
+#include "absl/types/variant.h"
+
+namespace Envoy {
+namespace Server {
+
+class GroupedStatsRequest
```
jmarantz, line 18: a class comment would be good, something like:

// Renders stats grouped by tag-extracted name. This is currently used for Prometheus (and has some
// Prometheus-specific logic in it) but this could be generalized for other formats.

source/server/admin/prometheus_stats_formatter.cc
```diff
@@ -0,0 +1,84 @@
...
+#include "source/common/stats/histogram_impl.h"
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_replace.h"
+
+namespace Envoy {
+namespace Server {
+
+namespace {
+
```
jmarantz, line 15: I assume this code is unchanged from what was in prometheus_stats.cc.

Actually would it make more sense to merge this with the prometheus render code below?

jmarantz (commented):
flushing some comments; sorry this has taken longer than expected to review.

jmarantz (commented):
flushing a couple more comments

source/server/admin/prometheus_stats_formatter.cc
rulex123: Yep, this is pre-existing code: sure, I will move it to the `PrometheusStatsRender` class.

source/server/admin/grouped_stats_request.cc
rulex123: Yep, you can reach this code with a text-readout, which are considered of type `gauge` when rendered (see https://github.com/envoyproxy/envoy/blob/9d9940b49f642fcf483ac8ff387ecb521dcf94fc/source/server/admin/prometheus_stats.cc#L278-L280).

I will add a comment to explain that.

source/server/admin/grouped_stats_request.cc
rulex123: Hm I am not sure that will work though. Inserting an empty vector would "reset" any other stats with the same tag extracted name, but different tags

jmarantz: From https://en.cppreference.com/w/cpp/container/map/insert

> insert(value_type): Returns a pair consisting of an iterator to the inserted element (or to the element that prevented the insertion) and a bool value set to true if and only if the insertion took place.

So it won't overwrite.

rulex123: aha, got it! will update the code

source/server/admin/grouped_stats_request.cc
rulex123: I have included a TODO in the header file.

source/server/admin/grouped_stats_request.cc
rulex123: I tried swapping this with SymbolTable::sortByStatName and I am getting a couple of failures in grouped_stats_requests_test.cc with some lock-related failure (excerpt below)

```
[mutex.cc : 1391] RAW: Acquiring absl::Mutex 0x7fd91a812b28 while holding  0x7fd91a812a18; a cycle in the historical lock ordering graph has been observed
```
I am still digging, though I don't think further reviewing has to block on this.

source/server/admin/stats_render.h
```diff
@@ -95,5 +97,54 @@ class StatsJsonRender : public StatsRender {
   std::string value_buffer_; // Used for Json::sanitize for text-readout values.
 };
 
+// Implements the Render interface for textual representation of Prometheus stats
+// (see: https://prometheus.io/docs/concepts/data_model)
+class PrometheusStatsRender : public StatsRender {
```
rulex123, line 102: Perhaps it would make more sense to have `GroupedStatsRender` and `UngroupedStatsRender`, but IMO we could leave that to a future refactoring.

source/server/admin/grouped_stats_request.cc
rulex123: Fixed now!

source/server/admin/stats_render.h
jmarantz: I think what you have now is good. I feel like "Prometheus" vs (say) "Html" or "Text" is one dimension, and "Grouped" vs "Ungrouped" is orthogonal.

test/server/admin/grouped_stats_request_test.cc
```diff
@@ -66,159 +67,211 @@ class PrometheusStatsFormatterTest : public testing::Test {
...
+  Stats::Counter& c1 = store_.rootScope()->counterFromStatNameWithTags(
+      makeStatName("cluster.upstream.cx.total"), c1Tags);
+  c1.add(10);
+
+  EXPECT_EQ(1, iterateChunks(*makeRequest(false)));
 }
 
-TEST_F(PrometheusStatsFormatterTest, SanitizeMetricName) {
-  Stats::CustomStatNamespacesImpl custom_namespaces;
-  std::string raw = "An.artist.plays-violin@019street";
```
jmarantz, line 128: this is great. I just want to leave a note for @ggreenway for senior maintainer review that these previous tests were moved to the prometheus section of render_test.cc, not removed.

jmarantz (approved):
Awesome work!

jmarantz:
@ggreenway can you take a senior maintainer pass on this?

ggreenway:
> @ggreenway can you take a senior maintainer pass on this?

Yep, I'll review

ggreenway:
Sorry for the delay; the code movement isn't tracked well by github so it's somewhat difficult to review. I'll post a review today or tomorrow.

source/server/admin/stats_render.cc
```diff
@@ -262,5 +299,124 @@ void StatsJsonRender::collectBuckets(const std::string& name,
...
+void PrometheusStatsRender::generate(Buffer::Instance&, const std::string&, uint64_t) {}
+
+void PrometheusStatsRender::generate(Buffer::Instance&, const std::string&,
+                                     const Stats::ParentHistogram&) {}
+
+std::string PrometheusStatsRender::formattedTags(const std::vector<Stats::Tag>& tags) {
+  std::vector<std::string> buf;
+  buf.reserve(tags.size());
+  for (const Stats::Tag& tag : tags) {
+    buf.push_back(fmt::format("{}=\"{}\"", sanitizeName(tag.name_), sanitizeValue(tag.value_)));
```
ggreenway, line 390: Consider using absl::StrCat for consistency with formatting above.

source/server/admin/stats_render.cc
```diff
@@ -262,5 +299,124 @@ void StatsJsonRender::collectBuckets(const std::string& name,
...
+                                stats.sampleCount(), "\n"));
+      response.add(absl::StrCat(prefixed_tag_extracted_name, "_sum{", tags, "} ",
+                                fmt::format("{0:.32g}", stats.sampleSum()), "\n"));
+      response.add(absl::StrCat(prefixed_tag_extracted_name, "_count{", tags, "} ",
+                                stats.sampleCount(), "\n"));
+    }
+  }
+}
+
+void PrometheusStatsRender::finalize(Buffer::Instance&) {}
```
ggreenway, line 377: Why are these 4 empty? Isn't this the StatsRender interface? At least add a comment about what's going on here.

source/server/admin/stats_request.h
```diff
@@ -26,22 +30,36 @@ class StatsRequest : public Admin::Request {
...
     TextReadouts,
     CountersAndGauges,
+    Counters,
+    Gauges,
     Histograms,
   };
 
+  struct Phase {
+    PhaseName phase;
+    std::string phase_label;
```
ggreenway, line 45: Should the `phase_label` be looked up from a constant table based on the `PhaseName`? Or can the label vary for the `PhaseName`?

source/server/admin/ungrouped_stats_request.cc
```diff
@@ -0,0 +1,108 @@
...
+
+    // check if filtered
+    if (params_.filter_ != nullptr) {
+      if (!std::regex_search(name, *params_.filter_)) {
+        return true;
+      }
+    } else if (params_.re2_filter_ != nullptr &&
+               !re2::RE2::PartialMatch(name, *params_.re2_filter_)) {
+      return true;
+    }
```
ggreenway, line 51: This is the same as the prometheus filtering. Can it be pulled out to a utility/helper?

test/server/admin/grouped_stats_request_test.cc
```diff
@@ -1,62 +1,63 @@
-#include <regex>
-#include <string>
-#include <vector>
+#include <memory>
```
ggreenway, line 1: Refactoring/rewriting all the tests makes it EXTREMELY difficult to understand if there were any behavior changes to what prometheus stats were output. Can you try leave the prometheus tests as untouched as possible in this PR, and we can consider revising them later?

Also, I think it makes sense to leave a whole set of prometheus-format-specific tests.

ggreenway (changes requested):
/wait

source/server/admin/stats_handler.cc
```diff
@@ -74,75 +75,37 @@ Http::Code StatsHandler::handlerStatsRecentLookupsEnable(Http::ResponseHeaderMap
...
-                                         Buffer::Instance& response) {
-  StatsParams params;
-  Http::Code code = params.parse(path_and_query, response);
-  if (code != Http::Code::OK) {
-    return code;
-  }
-
-  if (server_.statsConfig().flushOnAdmin()) {
-    server_.flushStats();
+  if (params.format_ == StatsFormat::Prometheus || path == "/stats/prometheus") {
```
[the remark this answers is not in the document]
rulex123: Agreed, I will update the code accordingly.

source/server/admin/stats_render.cc
```diff
@@ -1,14 +1,51 @@
...
 constexpr absl::string_view JsonQuoteCloseBrace = "\"}";
+
+const Envoy::Regex::CompiledGoogleReMatcher& promRegex() {
+  CONSTRUCT_ON_FIRST_USE(Envoy::Regex::CompiledGoogleReMatcher, "[^a-zA-Z0-9_]", false);
+}
+
+/**
+ * Take a string and sanitize it according to Prometheus conventions.
+ */
+std::string sanitizeName(const absl::string_view name) {
```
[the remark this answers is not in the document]
rulex123: I will move this and other prom-related methods out of the anonymous namespace and to the `PrometheusStatsRender` class, so I think we leave this name unchanged.

source/server/admin/stats_render.cc
```diff
@@ -1,14 +1,51 @@
...
+  // prometheus. Refer to https://prometheus.io/docs/concepts/data_model/.
+  // The initial [a-zA-Z_] constraint is always satisfied by the namespace prefix.
+  return promRegex().replaceAll(name, "_");
+}
+
+/**
+ * Take tag values and sanitize it for text serialization, according to
+ * Prometheus conventions.
+ */
+std::string sanitizeValue(const absl::string_view value) {
```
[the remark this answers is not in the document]
rulex123: Same as for `sanitizeName`.

source/server/admin/stats_render.cc
```diff
@@ -262,5 +299,124 @@ void StatsJsonRender::collectBuckets(const std::string& name,
   *histogram_array_->add_values() = ValueUtil::structValue(histogram_obj);
 }
 
+// Writes output for a Prometheus stat of type Gauge.
+void PrometheusStatsRender::generate(Buffer::Instance& response,
+                                     const std::string& prefixed_tag_extracted_name,
+                                     const std::vector<Stats::GaugeSharedPtr>& gauge) {
+  response.add(fmt::format("# TYPE {0} {1}\n", prefixed_tag_extracted_name, "gauge"));
```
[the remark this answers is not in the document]
rulex123: Agreed, I will update the code to factor out any common logic.

source/server/admin/stats_render.cc
rulex123: Yeah, I agree this is a bit odd. I will refactor the `StatsRender` interface so that we don't end up with these empty methods in the prometheus render.

source/server/admin/ungrouped_stats_request.cc
rulex123: Very similar but not exactly the same: the prometheus logic avoids a `stat->name()` unless there is a filter. The original code had a comment explaining this, I will add it back.

test/server/admin/grouped_stats_request_test.cc
rulex123: Fair point about the difficulty of reviewing due to moving all the tests. I will move the pre-existing prom related ones back to original files, hopefully that will ease code review.
 
> Also, I think it makes sense to leave a whole set of prometheus-format-specific tests.
Yeah, all of the pre-existing prometheus tests are left pretty much untouched, I only added new tests in this patch. Or maybe I am misunderstanding what you mean with this comment?

source/server/admin/stats_render.cc
jmarantz: I think it's OK for 'finalize' to be empty for prom -- this is empty for 'text' mode as well. It was needed for JSON format because the structure of the way the histograms are written out makes it impossible to stream the histograms. They are to collected and then emitted all once from the finalize routine.

source/server/admin/stats_handler.cc
```diff
@@ -71,78 +72,39 @@ Http::Code StatsHandler::handlerStatsRecentLookupsEnable(Http::ResponseHeaderMap
...
-                                         Buffer::Instance& response) {
-  StatsParams params;
-  Http::Code code = params.parse(path_and_query, response);
-  if (code != Http::Code::OK) {
-    return code;
-  }
-
-  if (server_.statsConfig().flushOnAdmin()) {
-    server_.flushStats();
+  if (params.format_ == StatsFormat::Prometheus || is_prometheus_path) {
```
jmarantz, line 87: this seems a little awkward; can we change the call-site `/stats/prometheus` to set the format,  and drop the extra bool?

test/server/admin/prometheus_stats_test.cc
```diff
@@ -47,16 +46,16 @@ class PrometheusStatsFormatterTest : public testing::Test {
   void addCounter(const std::string& name, Stats::StatNameTagVector cluster_tags) {
     Stats::StatNameManagedStorage name_storage(baseName(name, cluster_tags), *symbol_table_);
     Stats::StatNameManagedStorage tag_extracted_name_storage(name, *symbol_table_);
-    counters_.push_back(alloc_.makeCounter(name_storage.statName(),
-                                           tag_extracted_name_storage.statName(), cluster_tags));
+    test_scope_ptr_->counters_.push_back(alloc_.makeCounter(
```
jmarantz, line 49: I don't think we need to access the private member vars of TestScope -- we can just use the Scope API to add these, right?

test/server/admin/prometheus_stats_test.cc
```diff
@@ -98,67 +99,93 @@ class PrometheusStatsFormatterTest : public testing::Test {
 
   void clearStorage() {
     pool_.clear();
-    counters_.clear();
-    gauges_.clear();
-    histograms_.clear();
-    textReadouts_.clear();
+    test_scope_ptr_->counters_.clear();
```
jmarantz, line 102: I think you can just do test_scope_ptr_.reset() instead of all this.

jmarantz (commented):
I need to a look a bit more at the other changes in this test, but flushing comments for now.

Ideally we should have almost no changes to the test body, modulo namespace changes.

test/server/admin/prometheus_stats_test.cc
rulex123: Hm `numSymbols()` of the symbol table doesn't seem to clear though when using `test_scope_ptr_ .reset()`?

jmarantz: Hm. That seems surprising to me. Could you try calling symbol_table_->debugPrint() to see what's still there?

jmarantz: never mind I know what the problem is. For that implementation, the scopes don't really own the stats; the Store does.

You'd have to clear out the MockStore, which is annoying since it's not a shared_ptr. What you have is fine. Sorry for the confusion. You could leave a comment saying why we have to reach into the Store to clean all the stats in it, and maybe we could clean it up later.

test/server/admin/prometheus_stats_test.cc
rulex123: Possibly, but then we would have to change the `PrometheusStatsFormatterTest` class even more? 

My idea was to keep the test setup as close as possible the current version, that's why I ended up adding a new `TestScope` class and accessing its private vars. Do we need to change that?

test/server/admin/stats_request_test_base.h
```diff
@@ -0,0 +1,79 @@
...
+  Stats::StatNamePool pool_;
+  Stats::AllocatorImpl alloc_;
+  testing::NiceMock<Stats::MockSink> sink_;
+  testing::NiceMock<Event::MockDispatcher> main_thread_dispatcher_;
+  testing::NiceMock<ThreadLocal::MockInstance> tls_;
+  Stats::ThreadLocalStoreImpl store_;
+  Buffer::OwnedImpl response_;
+};
+
+class TestScope : public Stats::IsolatedScopeImpl {
```
jmarantz, line 42: I missed this before. Why do we need this version of TestScope?

Can't we just use MockStore::createScope()?

Actually come to think of it, can we just use IsolatedStoreImpl directly, or maybe Stats::TestUtil::TestStore and TestScope, without introducing these new wrappers?

test/server/admin/prometheus_stats_test.cc
jmarantz: I think it would not be more of a change that what you have already. You can just call test_scope_ptr_->counterFromStatName().

I noted this in the header file you added, but I'm not clear why you needed to subclass IsolatedStore and IsolatedScope, rather than just using them directly.

source/server/admin/stats_request.h
```diff
@@ -87,31 +108,42 @@ class StatsRequest : public Admin::Request {
   // current phase.
   void populateStatsForCurrentPhase(const ScopeVec& scope_vec);
 
-  // Populates all the metrics of the templatized type from scope_vec. Here we
-  // exploit that Scope::iterate is a generic templatized function to avoid code
-  // duplication.
-  template <class StatType> void populateStatsFromScopes(const ScopeVec& scope_vec);
+  virtual Stats::IterateFn<Stats::TextReadout> saveMatchingStatForTextReadout() PURE;
```
jmarantz, line 111: I think these can be in the protected section.

source/server/admin/ungrouped_stats_request.cc
```diff
@@ -0,0 +1,102 @@
...
+
+Stats::IterateFn<Stats::Histogram> UngroupedStatsRequest::saveMatchingStatForHistogram() {
+  return saveMatchingStat<Stats::Histogram>();
+}
+
+template <class SharedStatType>
+void UngroupedStatsRequest::renderStat(const std::string& name, Buffer::Instance& response,
+                                       const StatOrScopes& variant) {
+  auto stat = absl::get<SharedStatType>(variant);
+  dynamic_cast<StatsRender*>(render_.get())->generate(response, name, stat->value());
```
jmarantz, line 72: it would be good to brainstorm about how to get rid of the dynamic_cast. I hadn't noticed these before.

They are a mild anti-pattern IMO, and also they are slow.

They don't seem too hard to get rid of -- can't we have the `render_` object in UngroupedStatsRequest be a StatsRender?

source/server/admin/stats_request.h
```diff
@@ -87,31 +108,42 @@ class StatsRequest : public Admin::Request {
...
+  virtual void processCounter(const std::string& name, Buffer::Instance& response,
+                              const StatOrScopes& variant) PURE;
+  virtual void processHistogram(const std::string& name, Buffer::Instance& response,
+                                const StatOrScopes& variant) PURE;
 
   // Sets the chunk size.
   void setChunkSize(uint64_t chunk_size) { chunk_size_ = chunk_size; }
 
-private:
+protected:
```
jmarantz, line 128: Hmm. WDYT of leaving this private and moving some of these to the subclass (e.g. render_ in particular), or accessing them from subclasses via protected API?

jmarantz (commented):
Great progress.

I feel like some of the recent changes require some further thought.

I am regretting suggesting that we fold Prom into the StatsRender class hierarchy; I think this might have added complexity that may only benefit us in the future if we do a grouped version of the HTML renderer. This was my bad.

I'm now thinking we should go back to keeping the Prom version of StatsRequest separate and just do its rendering in the most natural way.

source/server/admin/stats_request.h
rulex123: I revisited all of the access modifiers in `stats_request.h`, I realise I had unnecessarily "relaxed" some of the pre-existing access.

test/server/admin/stats_request_test_base.h
rulex123: I think the issue with using existing test infrastructure for scope|store boils down to histograms: the `PrometheusStatsFormatterTest` tests heavily rely on mocking behaviour for histograms and I don't see an easy way to plug that into `IsolatedScopeImpl`.

I also looked a bit at `MockScope|Store` but I don't think they can be used as is for those tests: I might be missing something here, so LMK in case I am wrong :)

source/server/admin/stats_request.cc
```diff
@@ -77,34 +51,30 @@ bool StatsRequest::nextChunk(Buffer::Instance& response) {
   while (response.length() - starting_response_length < chunk_size_) {
     while (stat_map_.empty()) {
       if (phase_stat_count_ == 0) {
-        render_->noStats(response, phase_string_);
+        getRender().noStats(response, phase_labels_[phases_.at(phase_index_)]);
```
jmarantz, line 54: perf nit: hoist the render object into a temp outside the outer `while` avoid calling the same virtual call 3x per iteration.

source/server/admin/ungrouped_stats_request.h
```diff
@@ -0,0 +1,53 @@
...
+#include "source/server/admin/stats_request.h"
+#include "source/server/admin/utils.h"
+
+#include "absl/container/btree_map.h"
+#include "absl/types/variant.h"
+
+namespace Envoy {
+namespace Server {
+
+class UngroupedStatsRequest
```
jmarantz, line 16: Add comment explaining that this is for treating each stat on its own, as opposed to the Grouped version which groups by tag-name.

test/server/admin/stats_request_test_base.h
jmarantz: Let's leave this like you have it but it's a bit of technical debt to have to subclass IsolatedScopeImpl for an individual set of tests. There are so many Scope impls we should be able to re-use one.

Can you add a TODO for me to look at this and try to think of how to clean it after this is merged?

jmarantz (commented):
Mostly nits remaining. Thank you.

Greg -- there are a lot of changes now to prometheus_stats_test.cc to to calling model,, but it's not too hard to scan it and see it's testing a superset of what was previously tested. WDYT?

jmarantz (approved):
Thank you!

jmarantz:
@ggreenway over to you.

htuch:
@ggreenway friendly ping.

adisuissa:
@ggreenway can you PTAL?

source/server/admin/stats_render.cc
```diff
@@ -262,5 +272,163 @@ void StatsJsonRender::collectBuckets(const std::string& name,
...
+  const std::string formTags = formattedTags(tags);
+  return absl::StrCat(prefixed_tag_extracted_name, "{", formTags, "} 0\n");
+}
+
+std::string
+PrometheusStatsRender::generateHistogramOutput(const Stats::HistogramSharedPtr& metric,
+                                               const std::string& prefixed_tag_extracted_name) {
+  auto parent_histogram = dynamic_cast<Stats::ParentHistogram*>(metric.get());
+  if (parent_histogram != nullptr) {
+    // TODO(rulex123): support for the 3 histogram "bucket modes" offered in other
```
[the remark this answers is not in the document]
jmarantz: This makes sense.

FYI in https://github.com/envoyproxy/envoy/pull/26472 I have added a new bucket mode tentatively called 'detailed' which will have the raw bucket data. That *may* also be useful for Prom as you can see a lot more interesting data with this IMO, but it might complicate aggregation in Prom across multiple Envoys.

Don't worry Greg -- that PR is still a ways away from being atomized into probably around 4 for review, but I think the first PR will be the addition of a new bucket-mode to show the original buckets computed by circlhist.

Nevertheless I agree removing this TODO is fine for now.

jmarantz (approved):

ggreenway (approved):

PR #23958, opened 2022-11-11
Datadog tracing extension: remove OpenTracing

dgoffredo:
Commit Message: Datadog tracing extension uses new tracing library
Additional Description: See below.
Risk Level: Medium
Testing: Existing unit tests are sufficient.
Docs Changes: User-facing documentation for the Datadog tracing integration does not need to change, because these changes do not change the configuration.
Release Notes:  N/A or unsure
Platform Specific Features: N/A
[Optional Runtime guard:]
[Optional Fixes #Issue] [#21083](https://github.com/envoyproxy/envoy/issues/21083)
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

Overview
--------
This is rewrite of the Datadog tracing extension.

Now that there is a non-OpenTracing Datadog C++ Tracing library
([dd-trace-cpp][1]), the first order of business is to remove the
OpenTracing dependency on Envoy's Datadog tracing extension by rewriting
the extension in terms of the new library.

This pull request addresses a longstanding [issue][2], and if accepted will
allow Datadog tracing in Envoy to catch up with Datadog tracing in other
projects. Due to the OpenTracing dependency of the existing extension, Datadog
tracing has been frozen on an [old version][3] for a long time.

Changing to a non-OpenTracing, recent tracing library allows us to address
multiple problems with the current extension:

- improper handling of `error.*` span tags,
- lack of support for Datadog-specific sampling controls such as the
  `DD_TRACE_SAMPLE_RATE`, `DD_TRACE_RATE_LIMIT`, and `DD_SPAN_SAMPLING_RULES`
  environment variables,
- malformed traces resulting from multiple request passes through a proxy within
  the same trace.

In general, we have not been able to update the Datadog tracing extension since
January 2021. With these changes we will be able to make updates from now on.

Design
------
The components of this extension are described in the included `README.md` file.
There's even a diagram.

There is further description of the inner workings of the underlying tracing
library in its own [readme][4] and more extensively in its [maintainer
documentation][5].

[dd-trace-cpp][1] exports mostly concrete types. However, there are two
context-dependent interfaces that are implemented in this pull request.
See `event_scheduler.{h,cc}` and `agent_http_client.{h,cc}`.

Testing
-------
I made minimal modifications to the existing unit tests, and added several new ones.

There is also a `demo/` directory in the source tree that I used for manual
ad hoc integration testing during development. If that directory does not
[...]

PR #25003, opened 2023-01-18
datadog: time_util portion of new tracing library

dgoffredo:


source/extensions/tracers/datadog/time_util.h
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace Tracers {
+namespace Datadog {
+
+// Return the `datadog::tracing::TimePoint` that most closely matches the
+// specified `SystemTime`. Use the optionally specified
+// `datadog::tracing::Clock` to read the current time. If a clock is not
+// specified, then the default clock is used.
```
jmarantz, line 15: can you format these using doxygen syntax per other envoy files?

source/extensions/tracers/datadog/time_util.h
```diff
@@ -0,0 +1,22 @@
...
+datadog::tracing::TimePoint estimateTime(SystemTime);
+datadog::tracing::TimePoint estimateTime(SystemTime, const datadog::tracing::Clock&);
```
jmarantz, line 17: what is the SystemTime used for in this variant of estimateTime?

source/extensions/tracers/datadog/time_util.cc
```diff
@@ -0,0 +1,27 @@
...
+namespace Extensions {
+namespace Tracers {
+namespace Datadog {
+
+datadog::tracing::TimePoint estimateTime(SystemTime wall) {
+  return estimateTime(wall, datadog::tracing::default_clock);
+}
+
+datadog::tracing::TimePoint estimateTime(SystemTime wall, const datadog::tracing::Clock& clock) {
+  auto point = clock();
```
jmarantz, line 15: can you give a precise type here? It's not obvious from context what type `point` is.

source/extensions/tracers/datadog/time_util.cc
```diff
@@ -0,0 +1,27 @@
...
+  auto elapsed = point.wall - wall;
```
jmarantz, line 16: I often feel like working with abstract integer-like types that might be negative that it is more obviously correct to compare the two integer-like vars together, rather than subtracting them and comparing that variable  to 0, which might have unexpected types depending on signed-ness of the type

source/extensions/tracers/datadog/time_util.h
```diff
@@ -0,0 +1,22 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace Tracers {
+namespace Datadog {
+
+// Return the `datadog::tracing::TimePoint` that most closely matches the
+// specified `SystemTime`. Use the optionally specified
+// `datadog::tracing::Clock` to read the current time. If a clock is not
+// specified, then the default clock is used.
+datadog::tracing::TimePoint estimateTime(SystemTime);
```
jmarantz, line 16: is this an *estimate* really? or are we just converting a SystemTime to a DataDog time?

Would it be clearer at call-sites to call this convertSystemTimeToDataDog() ?

jmarantz (commented):
lgtm mostly with some minor comments. still need to read the test.

source/extensions/tracers/datadog/time_util.h
dgoffredo: Sure, I'll make that change here and elsewhere.

source/extensions/tracers/datadog/time_util.cc
dgoffredo: Will do. It's `datadog::tracing::TimePoint`, the return value of `datadog::tracing::Clock`.

source/extensions/tracers/datadog/time_util.h
dgoffredo: Yes, it's an estimate. The background is explained [here][1] as part of the original pull request (that readme will be added again in a subsequent pull request).

Envoy has a time type that, like Go's, contains both a system time point and a steady ("monotonic") time point. However, only the system time is exposed to the tracing subsystem. Not sure why.

This is problematic for the new library (dd-trace-cpp), because it uses the steady time to calculate the duration of a span (`end.tick - begin.tick`). So, we need to get a steady clock time from a given system clock time. The scheme is to measure the current system/steady time, compare the system part with the given system time, and then adjust the measured steady time accordingly. This is correct if the system clock has not been adjusted since the given system time was measured. It's incorrect otherwise, hence an estimate.

What do you think about all this? Is there a better way?

[1]: https://github.com/DataDog/envoy/blob/david.goffredo/dd-trace-cpp/source/extensions/tracers/datadog/README.md#time_utilhcc

source/extensions/tracers/datadog/time_util.h
dgoffredo: The same as in the previous variant. The contract applies to both overloads.

source/extensions/tracers/datadog/time_util.cc
```diff
@@ -0,0 +1,26 @@
...
+namespace Datadog {
+
+datadog::tracing::TimePoint estimateTime(SystemTime wall) {
+  return estimateTime(wall, datadog::tracing::default_clock);
+}
+
+datadog::tracing::TimePoint estimateTime(SystemTime wall, const datadog::tracing::Clock& clock) {
+  datadog::tracing::TimePoint point = clock();
+  if (point.wall > wall) {
+    point.tick -= point.wall - wall;
```
jmarantz, line 17: should we also verify that the delta we are subtracting is less than point.tick?

source/extensions/tracers/datadog/time_util.cc
```diff
@@ -0,0 +1,26 @@
+#include "source/extensions/tracers/datadog/time_util.h"
+
+#include <chrono>
+
+namespace Envoy {
+namespace Extensions {
+namespace Tracers {
+namespace Datadog {
+
+datadog::tracing::TimePoint estimateTime(SystemTime wall) {
```
jmarantz, line 10: why are we not plumbing the MonotonicTime (or maybe just the TimeSource into here?

jmarantz (commented):
looks pretty good. It'd be good to understand if we can just use MonotonicTime if that's what we need.

source/extensions/tracers/datadog/time_util.cc
dgoffredo: `point.tick` is a time point, while `point.wall - wall` is a duration, so no.

source/extensions/tracers/datadog/time_util.cc
dgoffredo: The goal is to implement the interfaces defined in <https://github.com/envoyproxy/envoy/blob/main/envoy/tracing/trace_driver.h>.

There are two time points in those interfaces: [Span::spawnChild][1] and [Driver::startSpan][2]. Both take `SystemTime` only, without a monotonic counterpart.

If I were redesigning the interface, I would have `spanChild` and `startSpan` take both a `SystemTime` and a `MonotonicTime`, and additionally have `finishSpan` take a `MonotonicTime` (or both, but at least the `MonotonicTime`).

I thought it would be inappropriate to propose to change the interface and everything that implements it.

What do you think?

[1]: https://github.com/envoyproxy/envoy/blob/ae071db29b5e7ceaed4a4b76e6d7b7c138f18f2d/envoy/tracing/trace_driver.h#L58-L65
[2]: https://github.com/envoyproxy/envoy/blob/ae071db29b5e7ceaed4a4b76e6d7b7c138f18f2d/envoy/tracing/trace_driver.h#L110-L112

source/extensions/tracers/datadog/time_util.cc
jmarantz: got it, thanks!

source/extensions/tracers/datadog/time_util.cc
jmarantz: I think it's fine to push forward with this but we should also try to correct the API to get the best results. It's not that hard to change these APIs; there's probably not that much fallout, particularly if you just plumb through some additional fields and don't remove the SystemTime.

source/extensions/tracers/datadog/time_util.h
jmarantz: sgtm thanks. The above explanation might be good to throw into the code here as a comment.

jmarantz (approved):

yanavlasov (approved):

dgoffredo:
> Code coverage for source/extensions/tracers is lower than limit of 95.4 (95.3)

dgoffredo:
https://storage.googleapis.com/envoy-pr/3629d09/coverage/source/extensions/tracers/datadog/time_util.cc.gcov.html

I could write a test for that overload. Will do this tomorrow.

yanavlasov (approved):

dgoffredo:
Thanks @yanavlasov @jmarantz

PR #25004, opened 2023-01-18
jwt_authn: change log level to debug

liwenhao0810:
Signed-off-by: Marco Li <liwenhao@google.com>

Commit Message: Change log level to debug to avoid log spam on mis-configuration.
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/filters/http/common/jwks_fetcher.cc
+++ b/source/extensions/filters/http/common/jwks_fetcher.cc
@@ -47,7 +47,7 @@ public:
     // Check if cluster is configured, fail the request if not.
     const auto thread_local_cluster = cm_.getThreadLocalCluster(remote_jwks_.http_uri().cluster());
     if (thread_local_cluster == nullptr) {
-      ENVOY_LOG(error, "{}: fetch pubkey [uri = {}] failed: [cluster = {}] is not configured",
+      ENVOY_LOG(debug, "{}: fetch pubkey [uri = {}] failed: [cluster = {}] is not configured",
                 __func__, remote_jwks_.http_uri().uri(), remote_jwks_.http_uri().cluster());
       complete_ = true;
       receiver_->onJwksError(JwksFetcher::JwksReceiver::Failure::Network);


kyessenov:
This looks fine to me, but @qiwzhang has better context, so passing on for the review.

source/extensions/filters/http/common/jwks_fetcher.cc
```diff
@@ -47,7 +47,7 @@ class JwksFetcherImpl : public JwksFetcher,
     // Check if cluster is configured, fail the request if not.
     const auto thread_local_cluster = cm_.getThreadLocalCluster(remote_jwks_.http_uri().cluster());
     if (thread_local_cluster == nullptr) {
-      ENVOY_LOG(error, "{}: fetch pubkey [uri = {}] failed: [cluster = {}] is not configured",
+      ENVOY_LOG(debug, "{}: fetch pubkey [uri = {}] failed: [cluster = {}] is not configured",
```
qiwzhang, line 50: Not,  I feel that this change will make it very difficult to find the root cause if "cluster" name is wrong.  Usually, envoy log is not at debug level.  

I feel that having error level making it easier to find the root cause.

kyessenov: For root causing, we should attach the error detail on the request itself. The issue is that this makes logs explode when the error happens, and is inconsistent with other RDS/LDS cases with the missing cluster.

kyessenov: Normally, Envoy doesn't log per request (that's reserved for debug/tracing levels). In this case, it logs an error for every request, and that explodes with high traffic.

Inconsistency is allowed in xDS, routes can reference missing clusters, etc. Envoy doesn't log an error then, but adds a detail per request, see `cluster_not_found` response code detail.

yanavlasov: Typically such error is attaches as the reason string when rejecting the request (assuming that in this case request is rejected). In this way it will be logged in the access log.

PR #25005, opened 2023-01-18
ZK filter: Fix "buffer underflow" error

Winbobob:
Commit Message: ZK filter: Fix "buffer underflow" error
Additional Description: When using ZK filter, we noticed some decoder errors. After looking into the issue, we found that some create requests had negative data len without real data and then the ZK filter failed to parse/skip the real data. Negative data len is because the data fetched from buffer is `0xffffffff` in our case, and the type of `len` is `int32_t`. Since the data is beyond the max value of int32, it becomes negative.
Risk Level: low
Testing: unit tests
Docs Changes: N/A
Release Notes: N/A

Signed-off-by: Zhewei Hu \<zhu@pinterest.com\>


Winbobob:
cc @rgs1, if you have a chance to review, thanks!

rgs1 (approved):
LGTM, thanks for the fix!

mattklein123 (approved):

PR #25020, opened 2023-01-18
datadog: dict_util portion of new tracing library

dgoffredo:

`dict_util` contains implementations of the `datadog::tracing::DictReader` and `datadog::tracing::DictWriter` interfaces. `dd-trace-cpp` uses these interfaces anywhere it needs to read from or write to mappings of string, such as when extracting trace context, injecting trace context, reading HTTP response headers, or writing HTTP request headers.


source/extensions/tracers/datadog/dict_util.cc
```diff
@@ -0,0 +1,78 @@
...
+    return result[0]->value().getStringView();
+  }
+
+  // There's more than one matching header entry.
+  // Per RFC 2616, this is the same as if there were only one entry whose
+  // value is the comma-separated concatenation of the multiple values.
+  // I don't expect the Agent to repeat response headers, and we don't even
+  // examine the Agent's response headers, but here's a solution anyway.
+  std::size_t i = 0;
+  datadog::tracing::assign(buffer_, result[i]->value().getStringView());
```
jmarantz, line 41: perf nit: make a `vector<string_view>` here rather than accumulating a string, and then use `absl::StrJoin(vec, ", ");`

I think that will also wind up being slightly less code, but (in the odd chance there's a lot of headers for the same name) it will be fewer allocations, since absl::StrJoin will right-size the string buffer once.

Also do we need hold

source/extensions/tracers/datadog/dict_util.h
```diff
@@ -0,0 +1,59 @@
...
+  Http::RequestHeaderMap& headers_;
+
+public:
+  explicit RequestHeaderWriter(Http::RequestHeaderMap& headers);
+
+  void set(datadog::tracing::StringView key, datadog::tracing::StringView value) override;
+};
+
+class ResponseHeaderReader : public datadog::tracing::DictReader {
+  const Http::ResponseHeaderMap& headers_;
```
jmarantz, line 30: style-guide nit: put the private members at the bottom, in a `private:` section -- what you have is totally fine it's just for consistency with the codebase and style guide.

source/extensions/tracers/datadog/dict_util.h
```diff
@@ -0,0 +1,59 @@
...
+
+  datadog::tracing::Optional<datadog::tracing::StringView>
+  lookup(datadog::tracing::StringView key) const override;
+
+  void visit(const std::function<void(datadog::tracing::StringView key,
+                                      datadog::tracing::StringView value)>& visitor) const override;
+};
+
+class TraceContextReader : public datadog::tracing::DictReader {
+  const Tracing::TraceContext& context_;
```
jmarantz, line 44: ditto

jmarantz (commented):
Looks great; just some trivial nits/suggestions.

source/extensions/tracers/datadog/dict_util.cc
dgoffredo: The `vector<string_view>` would need to allocate to accommodate the `result.size()` string views.

The benefit of this member `std::string` is that it reuses its storage across multiple calls to `lookup`, so allocations will creep up to some maximum size and then stay there until the `ResponseHeaderReader` is destroyed.

> Also do we need hold

If you mean the `/hold` command, then no. I was off Friday and am back today.

dgoffredo: This also means that the return value is invalidated on subsequent calls to `lookup`, which would be a dangerous API in general but is safe with usage here.

RyanTheOptimist:
Looks like this needs a main merge

source/extensions/tracers/datadog/dict_util.cc
dgoffredo: I need to fix a merge conflict in this PR. While I'm at it, I'll implement your suggestion.

dgoffredo: Ah, I can't believe I missed this. We can't use `absl::StrJoin`, because the return type is `Optional<StringView>`, i.e. the return value must refer to some non-temporary storage. _That's_ why I used the `std::string` data member; not as an optimization, but as a necessity.

So, I would like to keep this code as is, unless you have an alternative.

jmarantz: I don't feel that strongly about it, but I think it will work, be fewer lines of code, and faster :). You can assign the StrJoin to your string member var.

dgoffredo: Ah, right, can just assign to the member. This code will get executed zero times, so I'll use your suggestion to make it less to read.

jmarantz: You could also microbenchmark if you like :). The "faster" argument is that you can reserve a right-sized `vector<string_view>` from result.size(), fill in the views from all the header values, and then absl::StrJoin will right-size its string buffer before returning it. It will be RVO-optimized to be held in your buffer so you can return it as a string-view.

dgoffredo: Nah, it's all moot due to this being effectively dead code, and I bet $100 that we could write this as poorly as possible and we wouldn't be able to measure the difference in Envoy.

Here's the scenario where reusing `buffer_`'s storage would be advantageous: when `lookup` is called many times, and each time it has to construct a comma-separated result. In that case, we would at some point have enough storage inside `buffer_` to create comma-separated results without doing any allocations in all subsequent calls to `lookup`.

With the `StrJoin` solution, we must allocate the `vector` each time. I suppose we could mitigate that by making it a data member, too, ok. But `StrJoin` will always allocate and new `string`, and then when we assign it to `buffer_`, `string::operator=(string&&)` will replace `buffer_`'s storage, which is fine because we're already paying for the allocation in `StrJoin`. There is at least one allocation per call to `lookup` (assuming we're dealing with more than ~20 bytes).

With the original code, the total number of allocations is an amortized constant.

And it doesn't matter :)

jmarantz: Ah good point about the temp vector and subsequent calls, and I agree this doesn't matter. I just usually look at the pattern "create a string with intervening commas" and think "why not use StrJoin to simplify it"

dgoffredo: Yeah writing the logic for special-casing the first element gets tiring. Better to have a "join" utility.

Have a good weekend.

jmarantz (approved):

ggreenway (approved):

dgoffredo:
:tada: Thanks @jmarantz @ggreenway

PR #25023, opened 2023-01-18
TCP Tunneling: Optionally propagate response trailers to downstream info

IssaAbuKalbein:
Signed-off-by: IssaAbuKalbein <isaqalbein@gmail.com>

Commit Message: Add support to save response trailers from CONNECT tunnels in tcp_proxy
Additional Description: -
Risk Level: low (opt-in)
Testing: unit, integration
Docs Changes: none
Release Notes: yes


IssaAbuKalbein:
@kyessenov, @yanavlasov and @alyssawilk Can you review this PR instead of the closed one, please. Sorry for the inconvenience!

IssaAbuKalbein:
@yanavlasov What about your "co-variant return type" suggestions from closed PR?

source/common/tcp_proxy/tcp_proxy.h
```diff
@@ -113,21 +113,43 @@ using RouteConstSharedPtr = std::shared_ptr<const Route>;
...
  * Response headers for the tunneling connections.
  */
-class TunnelResponseHeaders : public StreamInfo::FilterState::Object {
+class TunnelResponseHeaders : public TunnelResponseHeadersOrTrailers {
 public:
   TunnelResponseHeaders(Http::ResponseHeaderMapPtr&& response_headers)
       : response_headers_(std::move(response_headers)) {}
-  const Http::ResponseHeaderMap& value() const { return *response_headers_; }
-  ProtobufTypes::MessagePtr serializeAsProto() const override;
+  const Http::ResponseHeaderMap& value() const override { return *response_headers_; }
```
yanavlasov, line 132:
```suggestion
  const Http::HeaderMap& value() const override { return *response_headers_; }
```

source/common/tcp_proxy/tcp_proxy.h
```diff
@@ -113,21 +113,43 @@ using RouteConstSharedPtr = std::shared_ptr<const Route>;
...
 };
 
+/**
+ * Response trailers for the tunneling connections.
+ */
+class TunnelResponseTrailers : public TunnelResponseHeadersOrTrailers {
+public:
+  TunnelResponseTrailers(Http::ResponseTrailerMapPtr&& response_trailers)
+      : response_trailers_(std::move(response_trailers)) {}
+  const Http::ResponseTrailerMap& value() const override { return *response_trailers_; }
```
yanavlasov, line 146:
```suggestion
  const Http::HeaderMap& value() const override { return *response_trailers_; }
```

yanavlasov (changes requested):
/wait

source/common/tcp_proxy/tcp_proxy.h
IssaAbuKalbein: Can you, please, explain why to do this change?
Also what you suggest is a breaking change for those are using "value()" with the return type "ResponseHeaderMap&".

IssaAbuKalbein: @yanavlasov ??

yanavlasov: While it is legal in C++ to override a virtual function with a different return type as long as they are covariant, it is not really needed in this case. You can safely return the reference to the base type of the header map.

IssaAbuKalbein: This previous code:
https://github.com/envoyproxy/envoy/blob/f6cb005211c389df0dc17d71b6819912e083b5cd/source/common/tcp_proxy/tcp_proxy.h#L123

IssaAbuKalbein: I've changed the method's signature in order to use it in "serializeAsProto" method:
https://github.com/envoyproxy/envoy/blob/b6bd4eacc63082bd2e7aa7e2991e0bac2eaf912c/source/common/tcp_proxy/tcp_proxy.cc#L567
And to keep the previous use of the method in the ResponseHeaderMap class by returning  a covariant type.
 
However, if you are not ok with that, I can do the following: 

- Remove the declaration of "value()" from the base class (TunnelResponseHeadersOrTrailers) and keep it as a regular method in both derived classes for those who need to get a return value of ResponseHeaderMap/ResponseTrailerMap (In my case, I need the derived class specific type).

- Introduce another method in the base class (let's say "virtual const Http::HeaderMap& headerMap() const PURE;") for the use in "serializeAsProto" method.

WDYT?

yanavlasov: The value() method was unused (I can not find any existing uses for it). You can change its return value as you need.

IssaAbuKalbein: It's not used here in the repo, but it is could be used by not built-in filters (I have a filter that uses this method with the current return type).

yanavlasov: I see. If you are using the `value()` method in your proprietary code, when yes, you need to create a new virtual method.

Also note, that the `value()` method may be deleted as unused code in the future, since maintainers have no way to know that a proprietary Envoy's extension is using it. To avoid this, you need to create tests that use this method and add comment stating that it is also used in a proprietary code to avoid its deletion.

IssaAbuKalbein: Got it, thanks!

yanavlasov (approved):

PR #25025, opened 2023-01-19
datadog: logger portion of new tracing library

dgoffredo:

`Logger` implements the `datadog::tracing::logger` interface in terms of a `spdlog::logger`. This allows `dd-trace-cpp` to produce logs using Envoy's logging.


RyanTheOptimist (commented):
/lgtm deps

dgoffredo:
https://storage.googleapis.com/envoy-pr/be8597a/coverage/source/extensions/tracers/datadog/logger.cc.gcov.html

I'll add a test that modifies the configured logging verbosity so that these `return;` branches are covered.

jmarantz:
/wait

needs main merge also

source/extensions/tracers/datadog/logger.h
```diff
@@ -0,0 +1,37 @@
...
+#include "source/common/common/logger.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace Tracers {
+namespace Datadog {
+
+/**
+ * Logging adapter for use by dd-trace-cpp. This class implements dd-trace-cpp's
+ * datadog::tracing::Logger interface in terms of spdlog::logger, which is what
```
jmarantz, line 14: non-actionable comment -- I'm just saying this into the wind...

It seems unfortunate that spdlog is a leaky abstraction from the Logger interface provided by Envoy's common library. I had thought that was an implementation detail, but I guess not.

This is some daunting tech-debt it would be nice to clean up some decade in the future.

test/extensions/tracers/datadog/logger_test.cc
```diff
@@ -0,0 +1,109 @@
...
+  logger.log_error([](std::ostream& log) { log << "Beware the ides of March."; });
+  EXPECT_EQ("Beware the ides of March.", sink->payload_);
+  EXPECT_EQ(absl::nullopt, sink->pattern_);
+  EXPECT_EQ(nullptr, sink->formatter_);
+  EXPECT_FALSE(sink->flush_);
+
+  sink->reset();
+  // callback-style startup banner
+  logger.log_startup([](std::ostream& log) {
+    log << "It's my stapler, the Swingline. It's been mine for a very long time.";
```
jmarantz, line 63: quote brings a smile to my face!

(also non-actionable)

test/extensions/tracers/datadog/logger_test.cc
```diff
@@ -0,0 +1,109 @@
...
+  });
+  EXPECT_EQ("It's my stapler, the Swingline. It's been mine for a very long time.", sink->payload_);
+  EXPECT_EQ(absl::nullopt, sink->pattern_);
+  EXPECT_EQ(nullptr, sink->formatter_);
+  EXPECT_FALSE(sink->flush_);
+
+  sink->reset();
+  // Error-style error
+  logger.log_error(datadog::tracing::Error{datadog::tracing::Error::OTHER,
+                                           "I'm sorry, Dave, I'm afraid I can't do that."});
```
jmarantz, line 73: in light of recent events this one brings me nightmares.

test/extensions/tracers/datadog/logger_test.cc
```diff
@@ -0,0 +1,109 @@
...
+  EXPECT_EQ(absl::nullopt, sink->pattern_);
+  EXPECT_EQ(nullptr, sink->formatter_);
+  EXPECT_FALSE(sink->flush_);
+
+  sink->reset();
+  // Errors are logged at "error" level, and so if the logger's level threshold
+  // is more severe than that, errors will not be logged.
+  // TODO: callback style
+  sink->reset();
+  // TODO: Error style
```
jmarantz, line 95: are you actually going to follow up with these? Usually we would do `TODO(github-id)` or `TODO(#issue-number)`.

I guess this is my only actionable comment.

jmarantz (commented):
lgtm with one actionable nit.

@wbpcode can you do a pass?

wbpcode (commented):
This is great and high quality contribution. Thanks. Only some minor comment.

And I think all the TODO should have a more formal format and more detailed context. Thanks again.

source/extensions/tracers/datadog/logger.h
dgoffredo: It is worth reconsidering the design I came up with. Was basically hacking around to see what worked.

The Envoy way is to use logging macros that conceal the underlying implementation. This has the advantage of allowing source file and line number to be included in the logged message.

We wouldn't be able to export those macros to a separately compiled library like dd-trace-cpp, and besides dd-trace-cpp has a policy of "no logging macros" for internal reasons :)

spdlog is a leaky abstraction, and regardless of what we use, dd-trace-cpp error messages will say they come from this component instead of from the underlying dd-trace-cpp component.

I'm open to ideas about how we can make dd-trace-cpp logging within Envoy more useful to users.

test/extensions/tracers/datadog/logger_test.cc
dgoffredo: I googled around for "____ screenplay" to find these quotes. It's amazing how much a movie's dialog differs from the screenplay.

test/extensions/tracers/datadog/logger_test.cc
dgoffredo: Oops! Will fill in these stubs, my bad.

dgoffredo:
Thanks for the reviews. I'll address your comments and merge in main within the next couple of days.

I'm also open to ideas on how we could do logging differently here, either to avoid the leaking abstraction and/or to expose dd-trace-cpp's source files and line numbers.

jmarantz:
I don't think we need to solve that problem in this PR -- this is not the first case where spdlog was referenced from envoy code outside its logging library; I just hadn't noticed it before.

jmarantz (approved):

jmarantz:
@wbpcode feel free to merge when ready (assuming CI passes)

PR #25027, opened 2023-01-19
util: Account for hash collisions in ObjectSharedPool

### LINKED ISSUE #24617: util: ObjectSharedPool will fail on hash collision
When @zyfjeff wrote this @htuch commented:

@[htuch](https://github.com/htuch) htuch [on Mar 6, 2020](https://github.com/envoyproxy/envoy/pull/10042#discussion_r389131531)
Is hashing sufficient here (in the absence of a crypto strong hash)? Should we have hash followed by an equality comparison?

Member
Author
@[zyfjeff](https://github.com/zyfjeff) zyfjeff [on Mar 10, 2020](https://github.com/envoyproxy/envoy/pull/10042#discussion_r390675462)
Yes, there is a need to consider hash conflicts. let me open another PR to perfect here.

However there is still only a hash in this container; no compare.

pradeepcrao:
Fixes #24617 

Commit Message: util: Account for hash collisions in ObjectSharedPool
Additional Description:
Risk Level: Low
Testing: new test added
Docs Changes: N/A
Release Notes: N/A


source/common/config/metadata.cc
```diff
@@ -107,13 +107,12 @@ bool Metadata::metadataLabelMatch(const LabelSet& label_set,
...
-            return std::make_shared<
-                SharedPool::ObjectSharedPool<const envoy::config::core::v3::Metadata, MessageUtil>>(
-                dispatcher);
-          });
+  return manager.getTyped<SharedPool::ObjectSharedPool<const envoy::config::core::v3::Metadata,
+                                                       MessageUtil, MessageUtil>>(
+      SINGLETON_MANAGER_REGISTERED_NAME(const_metadata_shared_pool), [&dispatcher] {
+        return std::make_shared<SharedPool::ObjectSharedPool<
+            const envoy::config::core::v3::Metadata, MessageUtil, MessageUtil>>(dispatcher);
+      });
```
jmarantz, line 115: did this code actually changed? I think this is just reformatted. Revert?

source/common/config/metadata.h
```diff
@@ -18,8 +18,9 @@
 namespace Envoy {
 namespace Config {
 
-using ConstMetadataSharedPoolSharedPtr = std::shared_ptr<
-    SharedPool::ObjectSharedPool<const envoy::config::core::v3::Metadata, MessageUtil>>;
+using ConstMetadataSharedPoolSharedPtr =
+    std::shared_ptr<SharedPool::ObjectSharedPool<const envoy::config::core::v3::Metadata,
+                                                 MessageUtil, MessageUtil>>;
```
jmarantz, line 23: formatting change only?

source/common/shared_pool/shared_pool.h
```diff
@@ -61,10 +68,10 @@ class ObjectSharedPool : public Singleton::Instance,
   std::shared_ptr<T> getObject(const T& obj) {
     ASSERT(std::this_thread::get_id() == thread_id_);
     auto hashed_value = HashFunc{}(obj);
-    auto object_it = object_pool_.find(hashed_value);
-    if (object_it != object_pool_.end()) {
-      auto lock_object = object_it->second.lock();
-      if (lock_object) {
+    auto object_range = object_pool_.equal_range(hashed_value);
```
jmarantz, line 71: rather than emulating a hash-table here by getting a 'bucket' and then iterating through it, can we just use an absl::flat_hash_map in the first place?

test/common/shared_pool/shared_pool_test.cc
```diff
@@ -174,5 +174,47 @@ TEST_F(SharedPoolTest, RaceCondtionForGetObjectWithObjectDeleter) {
...
+    static_assert(MyHash{}(15) == 1);
+    static_assert(MyHash{}(12) == 1);
+
+    // Instantiate objects that hash to the same value.
+    auto o = pool->getObject(4);
+    auto o1 = pool->getObject(3);
+
+    // Verify that there are separate entries in the pool for objects with the
+    // same hash value.
+    ASSERT_EQ(2, pool->poolSize());
```
jmarantz, line 197: these should all be EXPECT_EQ I think. I only use ASSERT_EQ if a failure should end the test, like ASSERT_EQ that an array size is expected prior to accessing array elements, which would crash if the size was not as expected.

source/common/config/metadata.h
pradeepcrao: It's hard to read, but there's an extra template argument (see the repeated `MessageUtil`) for `EqualFunc`

source/common/config/metadata.cc
pradeepcrao: It's hard to read, but there's an extra template argument (see the repeated `MessageUtil`) for `EqualFunc`

source/common/shared_pool/shared_pool.h
pradeepcrao: Could you elaborate on what the `absl::flat_hash_map` is for? What is the key and what is the value for this? 

This is my current understanding of the issue:
The object_pool is currently an `absl::flat_hash_map`, but the keys in that map are non-unique for the case when `HashFunc` produces duplicate values. This can be solved by replacing the `absl::flat_hash_map` with a  multimap that allows for non-unique keys.

test/common/shared_pool/shared_pool_test.cc
pradeepcrao: Thanks, makes sense.

source/common/config/metadata.cc
jmarantz: got it

source/common/config/metadata.h
jmarantz: got it

source/common/shared_pool/shared_pool.h
jmarantz: I see why this is a little tricky, and why you went with a mutlimap.

How about using something similar to the Stats::AllocatorImpl pattern? Use absl::flat_hash_set<T*, HashSharedPtr<T>, CompareSharedPtr<T>>. Clean the set on the shared_ptr's delete callback.

pradeepcrao: The nice thing about `ObjectSharedPool` as it's currently implemented is that it doesn't use locks, as the assumption is that everything is allocated  on the same (main) thread. It does however have a mechanism to schedule deletions of the container values on the original (main) thread. 

I will attempt to keep this behaviour with a `flat_hash_set`.

jmarantz: Yeah definitely. I think that should be do-able. If there's something blocking you I'm OK with the approach you have now, but fundamentally I think this can just be a set :)

jmarantz:
@yanjunxiang-google

test/common/shared_pool/shared_pool_test.cc
```diff
@@ -90,27 +90,27 @@ TEST_F(SharedPoolTest, NonThreadSafeForGetObjectDeathTest) {
...
-  {
-    // different threads
-    createObjectSharedPool(pool);
-    Thread::ThreadFactory& thread_factory = Thread::threadFactoryForTest();
-    auto thread =
-        thread_factory.createThread([&pool]() { pool->deleteObject(std::hash<int>{}(4)); });
-    thread->join();
-  }
-}
+// TEST_F(SharedPoolTest, ThreadSafeForDeleteObject) {
```
pradeepcrao, line 93: Investigate test failure.

test/common/shared_pool/shared_pool_test.cc
```diff
@@ -142,7 +142,7 @@ TEST_F(SharedPoolTest, GetObjectAndDeleteObjectRaceForSameHashValue) {
   // Because the storage is actually a new weak_ptr and the reference count is not zero, it is not
   // deleted
   dispatcher_->post([&pool, this]() {
-    EXPECT_EQ(1, pool->poolSize());
+    EXPECT_EQ(0, pool->poolSize());
```
pradeepcrao, line 145: Investigate test failure.

pradeepcrao: This fails because `absl::flat_hash_set<T, Hash, Compare>` will skip a call to `set.emplace(x)` if it finds a `T` which returns true for `Compare`.

source/common/shared_pool/shared_pool.h
```diff
@@ -4,6 +4,7 @@
 #include <memory>
 #include <thread>
 #include <type_traits>
+#include <unordered_map>
```
jmarantz, line 7: still needed?

source/common/shared_pool/shared_pool.h
```diff
@@ -30,60 +31,63 @@ namespace SharedPool {
...
+  void deleteObject(T* ptr) {
     if (std::this_thread::get_id() == thread_id_) {
-      // There may be new inserts with the same hash value before deleting the old element,
-      // so there is no need to delete it at this time.
-      if (object_pool_.find(hash_key) != object_pool_.end() &&
-          object_pool_[hash_key].use_count() == 0) {
-        object_pool_.erase(hash_key);
+      if (auto iter = object_pool_.find(ptr); iter != object_pool_.end()) {
+        // It is possible that the entry in object_pool_ corresponds to a
+        // different weak_ptr, due to a race condition in a shared_ptr being
```
jmarantz, line 48: This comment confused me but I think I see the point. It isn't that it's a *different* weak_ptr; it's that we have kind of deferred the destruction of a shared_ptr whose ref-count goes to zero with the callback, and in the meantime another reference to the shared_ptr may be created, making its use-count transition from 0 to 1 while waiting for the post.

source/common/shared_pool/shared_pool.h
```diff
@@ -30,60 +31,63 @@ namespace SharedPool {
...
+      delete ptr;
     } else {
       // Most of the time, the object's destructor occurs in the main thread, but with some
       // exceptions, it is destructed in the worker thread. In order to keep the object_pool_ thread
       // safe, the deleteObject needs to be delivered to the main thread.
       auto this_shared_ptr = this->shared_from_this();
       // Used for testing to simulate some race condition scenarios
       sync_.syncPoint(DeleteObjectOnMainThread);
-      dispatcher_.post([hash_key, this_shared_ptr] { this_shared_ptr->deleteObject(hash_key); });
+      dispatcher_.post([ptr, this_shared_ptr] { this_shared_ptr->deleteObject(ptr); });
```
jmarantz, line 66: WDYT of having breaking out a separate function deleteObjectMainThread where we put lines 46-58, and having the post call that directly, rather than re-checking the thread ID?

It might make the intent clearer.

source/common/shared_pool/shared_pool.h
```diff
@@ -30,60 +31,63 @@ namespace SharedPool {
...
 
-  void deleteObject(const size_t hash_key) {
+  void deleteObject(T* ptr) {
     if (std::this_thread::get_id() == thread_id_) {
-      // There may be new inserts with the same hash value before deleting the old element,
-      // so there is no need to delete it at this time.
-      if (object_pool_.find(hash_key) != object_pool_.end() &&
-          object_pool_[hash_key].use_count() == 0) {
-        object_pool_.erase(hash_key);
+      if (auto iter = object_pool_.find(ptr); iter != object_pool_.end()) {
```
jmarantz, line 46: my high level thought is that this whole class would be a lot simpler if we just protected the map with a mutex, which would be uncontended most of the time. Was there ever any data showing perf bottlenecks due to the mutex?

This might be aa question for the original writer @zyfjeff and reviewer(s) @htuch (IIRC).

jmarantz (commented):
this looks great; thanks for dealing with the complexity of the custom mixed-mode hasher/comparator; i can see how this might not have been the first approach when writing this :)

test/common/shared_pool/shared_pool_test.cc
jmarantz: so are you going to change it?

jmarantz (commented):
/wait

test/common/shared_pool/shared_pool_test.cc
pradeepcrao: Yes, sorry.

source/common/shared_pool/shared_pool.h
pradeepcrao: Removed this, thanks for catching this!

source/common/shared_pool/shared_pool.h
pradeepcrao: Sounds good. Also made the methods private.

jmarantz (approved):

zuercher (approved):

PR #25031, opened 2023-01-19
mobile ci: cancel ongoing mobile CI jobs when new SHA pushed to PR

### LINKED ISSUE #24982: mobile ci: cancel ongoing mobile ci jobs when a new SHA is pushed to a PR
*Title*: *cancel ongoing mobile ci jobs when a new SHA is pushed to a PR*

*Description*:
Canceling ongoing CI jobs in the case of a new SHA being pushed to a given PR should help us the limit the usage of CI that's used to run mobile checks (quite a limited resource). It would be a nice chance no matter whether we decide to eventually run mobile CI checks (instead of returning early from them) or not. Github actions - which are currently used to coordinate mobile CI checks - have an API that should allow us to implement what we want here.

glokta1:
Signed-off-by: glokta1 <rafeyahmad@protonmail.com>

Commit Message: [mobile ci: cancel ongoing mobile CI jobs when new SHA pushed to PR](https://github.com/glokta1/envoy/commit/ff707a607086cb32cd82ecde3d3d8d03bd0a5a9a)

Additional Description:  Added the `concurrency` key to all workflows which had a `pull_request` event attached. Some workflows are also additionally triggered on push to main, so added a fallback variable `github.run_id` to avoid syntax errors.

I was a little confused on whether to cancel the workflows that are run by cronjobs like  `codeql-daily.yml` but still include a `pull_request` event in some step. I have set it to cancel but let me know if this needs to be changed.

Fixes #24982

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ${{ github.head_ref-github.workflow || github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -6,6 +6,10 @@ on:
     - main
   pull_request:
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   androidbuild:
     name: android_build
--- a/.github/workflows/android_tests.yml
+++ b/.github/workflows/android_tests.yml
@@ -6,6 +6,10 @@ on:
     - main
   pull_request:
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   kotlintestsmac:
     # revert to //test/kotlin/... once fixed
--- a/.github/workflows/android_tests.yml
+++ b/.github/workflows/android_tests.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ${{ github.head_ref-github.workflow || github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/asan.yml
+++ b/.github/workflows/asan.yml
@@ -6,6 +6,10 @@ on:
     - main
   pull_request:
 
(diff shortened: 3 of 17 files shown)


yanavlasov:
Looks like format check is unhappy with TAB characters.

glokta1:
@yanavlasov fixed!

jpsim (approved):
Thanks so much for tackling this.

Based on my understanding of [GitHub Actions contexts](https://docs.github.com/en/actions/learn-github-actions/contexts) this seems to be a reasonable implementation.

jpsim (changes requested):
Actually I was wrong. I was wondering why only some of the GH Actions CI jobs actually ran on this PR, and I'm now realizing that every job currently shares the same concurrency group per PR.

I think we need to combine the `github.head_ref` with `github.action`. Maybe like this?

```yaml
concurrency:
  group: ${{ github.head_ref-github.action || github.run_id }}
  cancel-in-progress: true
```

glokta1:
@jpsim Oh yeah, you're absolutely right. I completely missed that! I would suggest appending `github.workflow` instead of `github.action` to the group expression, since we're defining the concurrency key at the workflow level. 

From the docs:
`github.action`
> The name of the action currently running, or the [id](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsid) of a step.

`github.workflow`
> The name of the workflow. If the workflow file doesn't specify a name, the value of this property is the full path of the workflow file in the repository.

From my understanding, If we use `github.action`, the new workflow file will not cancel the old one if it's on a different step.

glokta1 (commented):
@jpsim Oh yeah, you're absolutely right. I completely missed that! I would suggest appending `github.workflow` instead of `github.action` to the group expression, since we're defining the concurrency key at the workflow level. 

From the docs:
`github.action`
> The name of the action currently running, or the [id](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsid) of a step.

`github.workflow`
> The name of the workflow. If the workflow file doesn't specify a name, the value of this property is the full path of the workflow file in the repository.

From my understanding, If we use `github.action`, the new workflow file will not cancel its corresponding old one if it's on a different step/job.

jpsim:
`github.workflow` sounds good.

jpsim (approved):
LGTM

Might need https://github.com/envoyproxy/envoy/pull/25148 for the ios_tests jobs to pass though.

PR #25044, opened 2023-01-19
util: adds a script to run fast local format fixer

jmarantz:
Commit Message: Envoy's various format-checking and fixing scripts take a while to run on the whole repo, but usually they only need to run on the files you've changed.

This script does that, and runs the scripts w/o docker.
Additional Description:
Risk Level: low -- this is a new script not called by anything else
Testing: I've been using and evolving a variant of this for a couple of years, but this one (without my personal env var settings) has only been tested once, really, with "-main".
Docs Changes: n/a
Release Notes: n/a


tools/local_fix_format.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/sh
+#
+# Runs the Enovy format fixers on the current changes. By default, this runs on
```
soulxu, line 3: nit, s/Enovy/Envoy/

tools/local_fix_format.sh
```diff
@@ -0,0 +1,54 @@
...
+# changes that are not yet committed. You can also specify:
+#
+#   -all: runs on the entire repository (very slow)
+#   -main: runs on the changes amde from the main branch
```
soulxu, line 7: amde is made?

tools/local_fix_format.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# CLANG_FORMAT     -- points to the clang formatter binary, perhaps found in
+#                     $CLANG_TOOLCHAIN/bin/clang-format
+# BUILDIFIER_BIN   -- buildifier location, perhaps found in $HOME/go/bin/buildifier
+# BUILDOZER_BIN    -- buildozer location, perhaps found in $HOME/go/bin/buildozer
+
+# If DISPLAY is set, then tkdiff pops up for some BUILD changes.
+unset DISPLAY
+
+# Trigger an error if the required env vars are not set.
```
soulxu, line 25: the script `./tools/code_format/check_format.py` did that also, it also prints a message for reminding set the required env. so we can ignore this error here.

tools/local_fix_format.sh
```diff
@@ -0,0 +1,54 @@
...
+  if [[ $# > 0 && "$1" == "-main" ]]; then
+    shift
+    echo "Checking all files that have changed since the main branch."
+    args=$(git diff main | grep ^diff | awk '{print $3}' | cut -c 3-)
+  elif [[ $# == 0 ]]; then
+    args=$(git status|egrep '(modified:|added:)'|awk '{print $2}')
+    args+=$(git status|egrep 'new file:'|awk '{print $3}')
+  else
+    args="$@"
+  fi
```
lizan, line 43: Use `./tools/git/modified_since_last_github_commit.sh`? proto_format does same so it provides consistency between toolings.

tools/local_fix_format.sh
```diff
@@ -0,0 +1,54 @@
...
+
+  if [[ "$args" == "" ]]; then
+    echo No files selected. Bailing out.
+    exit 0
+  fi
+fi
+
+echo ./tools/code_format/check_format.py fix $args '|& grep -v "no longer checks API"'
```
lizan, line 51: set -x instead of echo and run?

tools/local_fix_format.sh
jmarantz: done

tools/local_fix_format.sh
jmarantz: done

tools/local_fix_format.sh
jmarantz: done -- though somewhat reluctantly because I think the message previously printed was better.

tools/local_fix_format.sh
jmarantz: I was not able to get that to produce the same results. Its implementation is certainly completely different, but for me it generates nothing when I run it on the client where I'm editing this.

tools/local_fix_format.sh
jmarantz: sure OK. I think the messaging was cleaner before due to the grep filter but it's not that big a deal.

tools/local_fix_format.sh
phlax: it greps for file extensions - perhaps that is why

tools/local_fix_format.sh
```diff
@@ -0,0 +1,49 @@
...
+
+  if [[ "$args" == "" ]]; then
+    echo No files selected. Bailing out.
+    exit 0
+  fi
+fi
+
+set -x
+./tools/code_format/check_format.py fix $args |& grep -v "no longer checks API"
+./tools/spelling/check_spelling_pedantic.py fix $args
```
phlax, line 49: this is missing most of the checks that `format_pre.sh` has

phlax (commented):
as discussed offline i think this is generally a good idea - but i think it needs to (configurably) do the same as `format_pre.sh` to be of general utility

likewise i think that that the aim needs to be to unify these code paths so we dont have to maintain even more fragmented code check tooling

i guess the key questions to my mind are:

- check in/exclusion
- file discovery - git/any files/changed files
- fail fast or continue
- error/warning handling
- param/flag passing

ideally unifying these code paths would make the above configurable for ci, local dev, and git hooks

tools/local_fix_format.sh
```diff
@@ -0,0 +1,49 @@
...
+    args="$@"
+  fi
+
+  if [[ "$args" == "" ]]; then
+    echo No files selected. Bailing out.
+    exit 0
+  fi
+fi
+
+set -x
```
phlax, line 47: i would strongly recommend against `-x` as a default - its noisy and imhe it rarely provides the debug info that you actually want

tools/local_fix_format.sh
```diff
@@ -0,0 +1,49 @@
...
+./tools/code_format/check_format.py fix $args |& grep -v "no longer checks API"
```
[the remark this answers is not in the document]
phlax: #25063

tools/local_fix_format.sh
jmarantz: I modified a .cc file and it did not catch it. Is there some trick to calling it?

phlax: not sure - i havent really used it directly - just hacked the code that does

```console
$ git grep since_last_github
tools/code_format/format_python_tools.sh:"$(dirname "$0")"/../git/modified_since_last_github_commit.sh ./ py || \
tools/proto_format/proto_format.sh:./tools/git/modified_since_last_github_commit.sh ./api/envoy proto || \
```

tools/local_fix_format.sh
jmarantz: Yes cause this wants to be fast so I don't hesitate to run it before every commit.

I was looking at other stuff in format_pre.sh and some of them look at the whole tree, run bazel, etc.  I think I'll add some flags to tweak behavior so this script can still be run the way I use it now.

jmarantz:
I'm going to iterate a bit more before asking for more comments.

tools/local_fix_format.sh
jmarantz: I definitely like to see what it's doing but I can make this a flag.

tools/local_fix_format.sh
jmarantz: well I think I'll leave this unresolved as what I have works for me, and (for whatever reason) the other script does not.

tools/local_fix_format.sh
jmarantz: Let's leave this as an exercise for future excellence :)

jmarantz:
Marked this 'ready' again.

phlax (dismissed):
happy to land iterate if this tool is useful now

mattklein123 (approved):
Awesome thank you, sounds like a great place to start.

jmarantz:
Needs another approval due the irony of me writing this bash script in a way that passes the format checks :)

phlax (approved):
lgtm, thanks @jmarantz

phlax:
one follow up thought i had - re my bullet list above https://github.com/envoyproxy/envoy/pull/25044#pullrequestreview-1263480765

these were basically the design criteria behind `envoy.code.check` - we had a few checks they all did similar things differently (and v slowly) so i put them together with a load of optimization

there are some issues with this - eg its not making very good use of bazel, and the parameter/flag handling desperately needs some love (eg if you limit path or file criteria it should automatically exclude any checks that do full repo scans, but many others)

otoh its v fast, and ~easy to integrate new linters/checkers etc

perhaps the way forward is to include the clang-tidy checks in `envoy.code.check` and improve the flag handling as required so we can effectively get what this pr provides without too much complexity

PR #25045, opened 2023-01-19
unified-delta-xds: fix map name in state tracking

adisuissa:
Commit Message: unified-delta-xds: fix map name in state tracking
Additional Description:
Similar to #24765 this fixes the var name when fetching the end-iterator.
Risk Level: Low.
Testing: N/A.
Docs Changes: N/A.
Release Notes: N/A.


yanavlasov (commented):
It looks like test coverage is lacking in this area. Do you think a test could be possible to construct that covers this if statement?

/wait-any

adisuissa:
The [coverage](https://storage.googleapis.com/envoy-pr/d5a7869/coverage/source/common/config/xds_mux/delta_subscription_state.cc.gcov.html) complains about a different line that the one this PR changes. It seems to be under a runtime-flag, and maybe that's why it is not tested.
I'll try to cover it in another PR.

yanavlasov (approved):

PR #25047, opened 2023-01-19
Fix divide by zero fuzz issue

DiazAlan:
Signed-off-by: AlanDiaz <diazalan@google.com>

Commit Message: Fix for divide by zero bug
Additional Description: If the number of active requests meets the max value of uint64, the + 1 will overflow the value to zero and cause a divide by zero error. This is unlikely to happen in normal cases but stops a fuzz test from failing.
Risk Level: Low
Testing: N/A
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/upstream/load_balancer_impl.h
+++ b/source/common/upstream/load_balancer_impl.h
@@ -686,48 +686,7 @@ protected:
 
 private:
   void refreshHostSource(const HostsSource&) override {}
-  double hostWeight(const Host& host) override {
-    // This method is called to calculate the dynamic weight as following when all load balancing
-    // weights are not equal:
-    //
-    // `weight = load_balancing_weight / (active_requests + 1)^active_request_bias`
-    //
-    // `active_request_bias` can be configured via runtime and its value is cached in
-    // `active_request_bias_` to avoid having to do a runtime lookup each time a host weight is
-    // calculated.
-    //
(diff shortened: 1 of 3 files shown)


source/common/upstream/load_balancer_impl.h
```diff
@@ -707,11 +707,19 @@ class LeastRequestLoadBalancer : public EdfLoadBalancerBase {
 
     double host_weight = static_cast<double>(host.weight());
 
+    // If the value of active requests is the max value, adding +1 will overflow
+    // it and cause a divide by zero. This won't happen in normal cases but stops
+    // failing fuzz tests
+    auto active_request_value =
```
adisuissa, line 713: nit: const, and use an explicit type for easier understanding.

adisuissa (commented):
LGTM modulo small nit.
Please also add the fuzz test to the corpus directory.
/wait

source/common/upstream/load_balancer_impl.h
DiazAlan: Done

snowp (changes requested):
This makes sense to me but seems like we want the corpus file added?

DiazAlan:
> This makes sense to me but seems like we want the corpus file added?

I just added the test file to the branch

adisuissa (approved):
LGTM, thanks!

snowp (approved):
Thanks!

PR #25049, opened 2023-01-19
Make Envoy Mobile enable HTTP/3 by default

RyanTheOptimist:
Change the builder APIs to enable HTTP/3 by default.

Risk Level: Low
Testing: Unit
Docs Changes: N/A
Release Notes: Updated release notes


mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -225,16 +225,16 @@ open class EngineBuilder(
...
    * to be utilized.
    *
-   * @param enableHttp3 whether to enable HTTP/3.
+   * @param disableHttp3 whether to disable HTTP/3.
    *
    * @return This builder.
    */
-  fun enableHttp3(enableHttp3: Boolean): EngineBuilder {
-    this.enableHttp3 = enableHttp3
+  fun disableHttp3(disableHttp3: Boolean): EngineBuilder {
```
Augustyniak, line 236: can we keep the naming of the method as it used to be (enable...)? just easier to reason about the API if all methods follow the same `enable...` naming convention - we can document the fact that default is `TRUE`. It's fine to change the default to be `true` (@jpsim I think that if we are OK with that change of a default we may want to ex[licitly disable that in our apps)

RyanTheOptimist: Sure, if that's the route we prefer! 

That being said, I do slightly wonder if it would make more sense for these enable/disable methods to actually take 0 arguments, so that the semantics are clear. But if we prefer to always do enable, that's cool with me! @jpsim does that work for you?

abeyad: That would require two methods (one enable, one disable) for every feature.  But ultimately I'm indifferent, so either option we go with is fine with me.

RyanTheOptimist: I've switched this PR over to `enableHttp3(false)` from `disableHttp3(true)`, so hopefully that does what Rafal suggested.

@abeyad I was think we would only have 1 method... the one that is not a no-op. So, in this case, only `disableHttp3()`. But I don't feel passionately about this at all!

abeyad: Sorry, I totally missed your point about 0 args.  In any case, yeah what you have LGTM

abeyad (approved):

RyanTheOptimist:
@jpsim Does this look good to you?

abeyad (approved):

jpsim (approved):
Looks good. For what it's worth, we'll keep this disabled in Lyft's apps since we don't talk to anything that speaks HTTP/3 yet, and we'll want to explicitly flip it on to experiment when we get to that, but this seems like the right default overall.

RyanTheOptimist:
> Looks good. For what it's worth, we'll keep this disabled in Lyft's apps since we don't talk to anything that speaks HTTP/3 yet, and we'll want to explicitly flip it on to experiment when we get to that, but this seems like the right default overall.

Thanks. Sounds good!

Issue #24994, opened 2023-01-18
Clean up Filesystem abstraction


ravenblackx:
Sequence of proposed operations.
- [ ] Add support for `stat()` to `Filesystem` (including an abstraction wrapper around `struct stat` / `WIN32FINDDATA`)
- [ ] Migrate `Directory()` into `Filesystem` - there is existing Directory() support, but it only works with the real file system, not through the abstraction (meaning no mock, no fake)
- [ ] Migrate `async_file` implementation to using `Filesystem` under the hood, rather than `OsSysCalls`
- [ ] Migrate `file_system_http_cache` implementation to using `Filesystem` rather than a hodgepodge mix of Filesystem, OsSysCalls and Directory()
- [ ] (maybe) Remove from OsSysCalls the functions that are better abstracted through Filesystem.

jmarantz:
FWIW, mod_pagespeed has a file-system API that implements the above requirements, including a mem-file-system which facilitates testing:

https://github.com/apache/incubator-pagespeed-mod/blob/master/pagespeed/kernel/base/file_system.h

https://github.com/apache/incubator-pagespeed-mod/blob/master/test/pagespeed/kernel/base/mem_file_system.h

This is presented for reference in case it's helpful.

PR #25050, opened 2023-01-19
Separate TestRandomGenerator from utility_lib

ravenblackx:
Commit Message: Separate TestRandomGenerator from utility_lib
Additional Description: This is a precursor cleanup to #24994. When adding stat() to Filesystem operations, the fake file system will need to keep track of file timestamps. As such, it needs access to a TimeSource to generate those timestamps, i.e. `simulated_time_system.h`. However, utility.h depends on the Memfile filesystem, and simulated_time_system depended on utility.h, so this creates a circular dependency. simulated_time_system only actually wants random numbers and thread_factory from utility.h, so separating TestRandomGenerator into its own library allows for a reduction in the dependency chain size and a solution to that otherwise-upcoming circular dependency.
Risk Level: None, this is test-only and no-op.
Testing: Test-only and no-op.
Docs Changes: n/a
Release Notes: n/a


yanavlasov:
Build failure seems legit

/wait

pradeepcrao:
LGTM

pradeepcrao (approved):

RyanTheOptimist (approved):

Issue #25052, opened 2023-01-19
wasm: uri fetch triggers a crash


kyessenov:
When configured with `uri` fetch, Wasm HTTP filter sometimes triggers a crash:
```
(redacted)
2023-01-11T23:38:04.086902Z	warning	envoy wasm	createWasm: failed to load (in progress)
2023-01-11T23:38:04.086942Z	warning	envoy wasm	createWasm: failed to load (cached) 
2023-01-11T23:38:07.882552Z	critical	envoy backtrace	Caught Segmentation fault, suspect faulting address 0x8
2023-01-11T23:38:07.882594Z	critical	envoy backtrace	Backtrace (use tools/stack_decode.py to get line numbers):
2023-01-11T23:38:07.882597Z	critical	envoy backtrace	Envoy version: 4d99fb094f34379565ba63c5a0bb19d86c8d7447/1.23.4-dev/Modified/RELEASE/BoringSSL-FIPS
2023-01-11T23:38:07.882996Z	critical	envoy backtrace	#0: [0x7f90ee624520]
2023-01-11T23:38:07.893044Z	critical	envoy backtrace	#1: std::__1::__function::__func<>::operator()() [0x12bbe8b]
2023-01-11T23:38:07.903881Z	critical	envoy backtrace	#2: Envoy::Http::FilterManager::applyFilterFactoryCb() [0x27884d6]
2023-01-11T23:38:07.913225Z	critical	envoy backtrace	#3: Envoy::Extensions::NetworkFilters::HttpConnectionManager::HttpConnectionManagerConfig::createFilterChainForFactories() [0x27558bf]
```

kyessenov:
This is important now for some of our efforts due to some security concerns with inline xDS delivery.

What needs to happen is:
* fix and confirm warming semantics works for delayed application of Wasm config;
* evaluate container fetch support;
* monitoring for the health of Wasm filter (due to fundamental limitations of NACKs https://github.com/envoyproxy/envoy/issues/35675

PR #25054, opened 2023-01-19
wasm: warn on remote URI use

kyessenov:

Commit Message: Signal a warning that remote URI code fetch may crash. https://github.com/envoyproxy/envoy/issues/25052. Using field annotation doesn't work because the field (DataSource) is shared.
Risk Level: low
Testing: none
Docs Changes: none
Release Notes: none


yanavlasov (approved):

yanavlasov:
@kyessenov The wasm-cc example keeps failing with what looks like outdated key. Maybe merge main and see if it clears the issue? Hard to imagine this PR is causing this breakage.

PR #25056, opened 2023-01-20
ZK filter: add RelativeZxid support for SetWatches requests

Winbobob:
Signed-off-by: Zhewei Hu <zhu@pinterest.com>

Commit Message: ZK filter: add RelativeZxid support for SetWatches requests
Additional Description: When parsing the SetWatches requests, we noticed some decoder errors. After looking into the requests bytes, we found that the [RelativeZxid](https://github.com/apache/zookeeper/blob/master/zookeeper-jute/src/main/resources/zookeeper.jute#L78) was not skipped during the decoding. This diff try to skip RelativeZxid then the filter could parse later watches data correctly without raising the "read beyond the max length" or "packet is too big" errors.
Risk Level: low
Testing: unit tests
Docs Changes: N/A
Release Notes: N/A


Winbobob:
cc @rgs1, please review if you have a chance, thank you!

rgs1 (approved):
Thanks for the fix!

Btw, we need to add support for persistent recursive watches at some point:

https://github.com/apache/zookeeper/blob/master/zookeeper-jute/src/main/resources/zookeeper.jute#L83

Winbobob:
Thank you very much Raul for the quick response! Sure thing, will do it in my next diff.

mattklein123 (approved):

PR #25060, opened 2023-01-20
access log: Log ports as numbers, not as strings (JSON)

mkauf:
Affected command operators:
- %DOWNSTREAM_LOCAL_PORT%
- %DOWNSTREAM_REMOTE_PORT%
- %DOWNSTREAM_DIRECT_REMOTE_PORT%
- %UPSTREAM_LOCAL_PORT%
- %UPSTREAM_REMOTE_PORT%


source/common/formatter/substitution_formatter.cc
```diff
@@ -760,6 +760,14 @@ class StreamInfoAddressFieldExtractor : public StreamInfoFormatter::FieldExtract
       return unspecifiedValue();
     }
 
+    if (extraction_type_ == StreamInfoFormatter::StreamInfoAddressFieldExtractionType::JustPort) {
+      auto port = StreamInfo::Utility::extractDownstreamAddressJustPort(*address);
```
zuercher, line 764: I think we probably need to add a runtime flag for this to give operators time to handle the change. Otherwise it's going to break someone's log processing.

botengyao: +1, there could be some log process assuming it is string, but now it will be numbers.
You can add runtime guard [here](https://github.com/envoyproxy/envoy/blob/b41adf18d0419d4c34cc153bf53b3d27295068a3/source/common/runtime/runtime_features.cc):

mkauf: Thank you for the suggestion, I have added a runtime guard.

zuercher (commented):
There's a merge conflict to be resolved, but otherwise I think it looks good.

zuercher:
@wbpcode can you have a look as well?

source/common/formatter/substitution_formatter.cc
```diff
@@ -760,6 +760,15 @@ class StreamInfoAddressFieldExtractor : public StreamInfoFormatter::FieldExtract
       return unspecifiedValue();
     }
 
+    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.format_ports_as_numbers")) {
+      if (extraction_type_ == StreamInfoFormatter::StreamInfoAddressFieldExtractionType::JustPort) {
+        auto port = StreamInfo::Utility::extractDownstreamAddressJustPort(*address);
```
wbpcode, line 765: nit: `const auto`

wbpcode (commented):
It's great and LGTM. Thanks. Only a nit comment that you can accept or just ignore it.

source/common/formatter/substitution_formatter.cc
mkauf: changed, thanks!

mkauf:
Thank you, I have updated the pull request.

zuercher (approved):

PR #25079, opened 2023-01-20
Tcp: add a tcp async client

botengyao:

Add a Tcp async client and factory to thread local cluster.

Commit Message:
Additional Description:
Risk Level: Low
Testing: integration & unit test
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


botengyao:
/assign @yanavlasov 

Thanks for the review!

For the test filter, I am not using the filter owned buffer, but directly use the underlying connection buffer to achieve the stream.

envoy/tcp/async_tcp_client.h
```diff
@@ -0,0 +1,78 @@
...
+/*
+ * AsyncTcpClientCallbacks for peer connection callbacks and data.
+ */
+class AsyncTcpClientCallbacks : public Network::ConnectionCallbacks {
+public:
+  ~AsyncTcpClientCallbacks() override = default;
+
+  /*
+   * Invoked when data is delivered from the peer connection.
+   * @param data supplies data from the upstream
```
yanavlasov, line 18:
```suggestion
   * @param data supplies data from the peer
```

envoy/tcp/async_tcp_client.h
```diff
@@ -0,0 +1,78 @@
...
+   * Connect to a remote host. Errors or connection events are reported via the
+   * event callback registered via addAsyncTcpClientCallbacks().
+   */
+  virtual bool connect() PURE;
+
+  /**
+   * Close the client, the underlying connection will also be defer deleted when
+   * a Close event is received.
+   */
+  virtual void close() PURE;
```
yanavlasov, line 41: The semantics of the call need to be clarified a little. Presently it abortively closes the connection discarding any unsent data.

yanavlasov (changes requested):
A couple of small nits. LGTM otherwise.

/wait

envoy/tcp/async_tcp_client.h
botengyao: Done

envoy/tcp/async_tcp_client.h
botengyao: Thanks!

botengyao:
@yanavlasov PTAL, thanks!

yanavlasov (dismissed):

yanavlasov:
I have pinged maintainers for non Google review

source/common/tcp/async_tcp_client_impl.cc
```diff
@@ -0,0 +1,78 @@
...
+  return true;
+}
+
+void AsyncTcpClientImpl::close() {
+  if (connection_) {
+    connection_->close(Network::ConnectionCloseType::NoFlush);
+  }
+}
+
+void AsyncTcpClientImpl::addAsyncTcpClientCallbacks(AsyncTcpClientCallbacks& callbacks) {
```
lizan, line 45: What this method does is setAsyncTcpClientCallbacks but not add.

source/common/upstream/cluster_manager_impl.h
```diff
@@ -523,6 +525,8 @@ class ClusterManagerImpl : public ClusterManager,
                                               LoadBalancerContext* context) override;
       Host::CreateConnectionData tcpConn(LoadBalancerContext* context) override;
       Http::AsyncClient& httpAsyncClient() override;
+      Tcp::AsyncTcpClientPtr tcpAsyncClient(LoadBalancerContext* context,
```
lizan, line 528: The AsyncTcpClient has different semantics from the HTTP async client so it might be a bit confusing. While the HTTP async client allows you to start multiple stream/request to the cluster, the AsyncTcpClient is bound to 1 connection and not reusable.

botengyao: Right, this is a factory method of the AsyncTcpClient, do you have some other suggestions?.

source/common/tcp/async_tcp_client_impl.cc
botengyao: Thanks! Makes sense.

source/common/upstream/cluster_manager_impl.h
lizan: That depends on how this is going to be used, you might want follow the same pattern of the HTTP async client that starting multiple streams (connections in TCP).

botengyao: I see, 1 connection works for our use case right now, and I am thinking the client factory can be configurable e.g., we may add multiple streams support either in the current object or a new implementation of in future. WDYT? Thanks!

lizan: SGTM, what I'm saying is that the name convention here confuses a bit compared to the http client above. The benefit of the http client is that you don't need to store/pass around a ClusterManager. In your use case, what if you want to retry that 1 connection?

botengyao: Thanks! We don't retry right now, I added an options to the factory, and then user can decide how the client is created & used.

lizan (approved):

PR #25080, opened 2023-01-20
Upstream config memory optimization

DiazAlan:
Commit Message: Reduce memory usage if cluster does not contain an upstream config
Additional Description: In cases where a cluster does not have a set upstream config, this will save memory using unique pointers
Risk Level: Low
Testing: Changed mocks as needed
Docs Changes: NA
Release Notes: NA

### DIFF
--- a/envoy/upstream/upstream.h
+++ b/envoy/upstream/upstream.h
@@ -1007,7 +1007,7 @@ public:
    * @return const absl::optional<envoy::config::core::v3::TypedExtensionConfig>& the configuration
    *         for the upstream, if a custom upstream is configured.
    */
-  virtual const absl::optional<envoy::config::core::v3::TypedExtensionConfig>&
+  virtual OptRef<envoy::config::core::v3::TypedExtensionConfig>
   upstreamConfig() const PURE;
 
   /**
--- a/envoy/upstream/upstream.h
+++ b/envoy/upstream/upstream.h
@@ -1007,8 +1007,7 @@ public:
    * @return const absl::optional<envoy::config::core::v3::TypedExtensionConfig>& the configuration
    *         for the upstream, if a custom upstream is configured.
    */
-  virtual OptRef<envoy::config::core::v3::TypedExtensionConfig>
-  upstreamConfig() const PURE;
+  virtual OptRef<envoy::config::core::v3::TypedExtensionConfig> upstreamConfig() const PURE;
 
   /**
    * @return Whether the cluster is currently in maintenance mode and should not be routed to.
--- a/envoy/upstream/upstream.h
+++ b/envoy/upstream/upstream.h
@@ -1007,7 +1007,7 @@ public:
    * @return const absl::optional<envoy::config::core::v3::TypedExtensionConfig>& the configuration
    *         for the upstream, if a custom upstream is configured.
    */
-  virtual OptRef<envoy::config::core::v3::TypedExtensionConfig> upstreamConfig() const PURE;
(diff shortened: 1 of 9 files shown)


DiazAlan:
@yurykats @adisuissa

source/common/upstream/upstream_impl.h
```diff
@@ -800,9 +800,11 @@ class ClusterInfoImpl : public ClusterInfo,
     }
     return *lb_original_dst_config_;
   }
-  const absl::optional<envoy::config::core::v3::TypedExtensionConfig>&
-  upstreamConfig() const override {
-    return upstream_config_;
+  OptRef<envoy::config::core::v3::TypedExtensionConfig> upstreamConfig() const override {
```
jmarantz, line 803: nit: I feel like we should be returning an `OptRef<const envoy::config::core::v3::TypedExtensionConfig>` given that this is a const method.

Would that work at call-sites?

The version checked in now actually returned a const optional ref, so I think this PR should also return a config that cannot be modified by the caller.

jmarantz (commented):
thanks for doing this; I have a drive-by comment.

envoy/upstream/upstream.h
```diff
@@ -1007,8 +1007,7 @@ class ClusterInfo : public Http::FilterChainFactory {
    * @return const absl::optional<envoy::config::core::v3::TypedExtensionConfig>& the configuration
    *         for the upstream, if a custom upstream is configured.
    */
-  virtual const absl::optional<envoy::config::core::v3::TypedExtensionConfig>&
-  upstreamConfig() const PURE;
+  virtual OptRef<envoy::config::core::v3::TypedExtensionConfig> upstreamConfig() const PURE;
```
adisuissa, line 1010: nit: return const TypedExtensionConfig

source/common/upstream/upstream_impl.h
adisuissa: +1

source/common/upstream/upstream_impl.h
```diff
@@ -957,7 +959,7 @@ class ClusterInfoImpl : public ClusterInfo,
       lb_maglev_config_;
   const std::unique_ptr<const envoy::config::cluster::v3::Cluster::OriginalDstLbConfig>
       lb_original_dst_config_;
-  absl::optional<envoy::config::core::v3::TypedExtensionConfig> upstream_config_;
+  std::unique_ptr<envoy::config::core::v3::TypedExtensionConfig> upstream_config_;
```
adisuissa, line 962: nit: const

adisuissa (commented):
Thanks! LGTM modulo using const.
+1 on @jmarantz comment, should prefer const if possible.

/wait

envoy/upstream/upstream.h
DiazAlan: Done

source/common/upstream/upstream_impl.h
adisuissa: Can this also be `const TypedExtensionConfig`?

adisuissa (commented):
Thanks!
One more nit that needs to be addressed.

source/common/upstream/upstream_impl.h
adisuissa: After discussing this IRL it seems that this is non-const because of the tests, and I think that keeping it non-const is ok for now.

adisuissa (approved):
LGTM, thanks!
cc @jmarantz if there are additional comments.

yanavlasov (approved):

PR #25104, opened 2023-01-23
Fix nullptr constructor in grpc_access_log_utils

ravenblackx:
Commit Message: Fix nullptr constructor in grpc_access_log_utils
Additional Description: This is effectively a no-op change for google-impl of absl::string_view, which constructs an empty string_view from nullptr and then compares, to see if the comparator string_view is empty. For std::string_view implementation, however, constructing from nullptr is a problem.
Risk Level: None; intended behavior is preserved.
Testing: Passes existing tests, also tested 'live' in the environment where the old version has a problem.
Docs Changes: n/a
Release Notes: n/a


ravenblackx:
Superceded by #25108 doing the same thing.

PR #25109, opened 2023-01-23
removing non-functional flag lazy_map_min_size

alyssawilk:
Follow-up clean up from https://github.com/envoyproxy/envoy/pull/23613/files

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: none


RyanTheOptimist (approved):
Wow, nice catch!

PR #25112, opened 2023-01-23
subset-lb: fixing a mismatch between checked and returned lb config

adisuissa:
This was approved for a public repo fix because the PR causing the issue was merged 4 days ago.

Commit Message: subset-lb: fixing a mismatch between checked and returned lb config
Additional Description:
PR #24905 optimized the memory used for clusters. However it introduced a subtle bug when Least-Request LB is used, which results in returning the default config for that policy.

Risk Level: Low
Testing: This is a private method that is not easy to validate.
Docs Changes: N/A.
Release Notes: N/A.
Fixes commit #24905
[Optional Deprecated:]

### DIFF
--- a/source/common/upstream/subset_lb.h
+++ b/source/common/upstream/subset_lb.h
@@ -365,7 +365,7 @@ private:
 
   OptRef<const envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>
   lbLeastRequestConfig() const {
-    if (round_robin_config_ != nullptr) {
+    if (least_request_config_ != nullptr) {
       return *least_request_config_;
     }
     return absl::nullopt;
--- a/source/common/upstream/subset_lb.h
+++ b/source/common/upstream/subset_lb.h
@@ -432,7 +432,7 @@ private:
 
   TimeSource& time_source_;
 
-  friend class SubsetLoadBalancerDescribeMetadataTester;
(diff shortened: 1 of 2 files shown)


adisuissa:
Two open questions:
1. Tests for this are hard to add, because the method is private and no easy access to the generated object.
2. Should I add a release note?

adisuissa:
cc @KBaichoo

KBaichoo:
> Tests for this are hard to add, because the method is private and no easy access to the generated object.

https://github.com/envoyproxy/envoy/pull/25112/files#diff-5789d0f7635142c87cc64c724c475ba691543b41b5859b7cbac953dc3cf9ddfbR435 Seems to be a test class friend, could that also be leveraged in this case to check this?

> Should I add a release note?

I think probably posting to envoy-announce would be better as an FYI to folks running near head.

adisuissa:
> https://github.com/envoyproxy/envoy/pull/25112/files#diff-5789d0f7635142c87cc64c724c475ba691543b41b5859b7cbac953dc3cf9ddfbR435 Seems to be a test class friend, could that also be leveraged in this case to check this?

Added test (and renamed the friend test-class).

> I think probably posting to envoy-announce would be better as an FYI to folks running near head.

We can send this after the PR is merged (also sending sha range).

KBaichoo (approved):
Thank you for catching this!

PR #25113, opened 2023-01-23
Fix TestScope so it correctly supports [metric]FromStatNameWithTags

ravenblackx:
Commit Message: Fix TestScope so it correctly supports [metric]FromStatNameWithTags
Additional Description: A metric created with e.g. counterFromStatNameWithTags, with the tags value populated, would fail verifyConsistency as the internal stat identifier is not == the stat name without tags.
This change fixes `TestScope::verifyConsistency` to take a tags parameter, and `TestScope::(counter|gauge|histogram)FromStatNameWithTags` to consistently use its tags parameter.
Risk Level: None, test-only change.
Testing: Tested against a [future] situation where the behavior was incorrectly causing a test to fail.
Docs Changes: n/a
Release Notes: n/a


jmarantz (approved):

mattklein123 (approved):

PR #25115, opened 2023-01-24
xds: Adds support for xdstp in State-of-the-World

abeyad:
Previously, xdstp support was limited to the Delta protocol. This commit adds support for xdstp in State-of-the-World (SotW) xDS, for both the unified gRPC mux and the legacy gRPC mux.


test/integration/ads_integration_test.cc
```diff
@@ -2374,4 +2377,106 @@ TEST_P(XdsTpAdsIntegrationTest, EdsAlternatingLedsUsage) {
...
+
+INSTANTIATE_TEST_SUITE_P(
+    UnifiedSotw, UnifiedSotwXdsTpAdsIntegrationTest,
+    testing::Combine(testing::ValuesIn(TestEnvironment::getIpVersionsForTest()),
+                     // There should be no variation across clients.
+                     testing::Values(Grpc::ClientType::EnvoyGrpc),
+                     testing::Values(Grpc::SotwOrDelta::UnifiedSotw),
+                     testing::Values(OldDssOrNewDss::Old, OldDssOrNewDss::New)));
+
+TEST_P(UnifiedSotwXdsTpAdsIntegrationTest, Basic) {
```
adisuissa, line 2393: Is this a copy-paste of the Basic test?
If so, and assuming that more tests will be added for both of SotW and Delta, I suggest merging the two (test-parameter) and skipping the LEDS tests in SotW mode.

source/common/config/subscription_factory_impl.cc
```diff
@@ -146,6 +146,14 @@ SubscriptionPtr SubscriptionFactoryImpl::collectionSubscriptionFromUrl(
             callbacks, resource_decoder, stats, dispatcher_,
             Utility::configSourceInitialFetchTimeout(config), false, options);
       }
+      case envoy::config::core::v3::ApiConfigSource::AGGREGATED_GRPC: {
+        if (!Runtime::runtimeFeatureEnabled("envoy.reloadable_features.unified_mux")) {
+          throw EnvoyException("AGGREGATED_GRPC only supported for unified_mux");
```
adisuissa, line 151: I'm probably missing something here, but why is unified mux needed here?

test/integration/ads_integration_test.cc
```diff
@@ -1749,23 +1749,26 @@ TEST_P(AdsIntegrationTest, ContextParameterUpdate) {
 class XdsTpAdsIntegrationTest : public AdsIntegrationTest {
 public:
   void initialize() override {
-    config_helper_.addConfigModifier([](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
+    config_helper_.addConfigModifier([this](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
```
adisuissa, line 1752: nit: instead of passing `this` consider passing `sotwOrDelta()`.

adisuissa (commented):
Thanks, left a few comments.
/wait

source/common/config/subscription_factory_impl.cc
adisuissa: Not sure I fully grok the strange behavior. If the cluster was updated, it should apply it IMHO.

If it doesn't require many changes then supporting non-unified might be beneficial, as one can easily change the flag and see that everything works.

test/integration/ads_integration_test.cc
```diff
@@ -1749,149 +1749,172 @@ TEST_P(AdsIntegrationTest, ContextParameterUpdate) {
...
+    test_server_->waitForCounterEq("listener_manager.listener_removed", 1);
+    makeSingleRequest();
+  }
 }
 
 // Basic CDS/EDS/LEDS update that warms and makes active a single cluster.
 TEST_P(XdsTpAdsIntegrationTest, BasicWithLeds) {
   initialize();
+
+  if (isSotw()) {
```
adisuissa, line 1913: question: can this be done even before `initialize()` is called?

test/integration/ads_integration_test.cc
```diff
@@ -1749,149 +1749,172 @@ TEST_P(AdsIntegrationTest, ContextParameterUpdate) {
...
-  // Remove bar listener from the foo namespace.
-  sendDiscoveryResponse<envoy::config::listener::v3::Listener>(
-      Config::TypeUrl::get().Listener, {}, {},
-      {"xdstp://test/envoy.config.listener.v3.Listener/foo-listener/"
-       "bar?xds.node.cluster=cluster_name&xds.node.id=node_name"},
-      "3");
-  EXPECT_TRUE(compareDiscoveryRequest(Config::TypeUrl::get().Listener, "4", {}, {}, {}));
-  test_server_->waitForCounterEq("listener_manager.listener_removed", 1);
-  makeSingleRequest();
+  // Updates and deletions only apply to the Delta protocol, not SotW.
```
adisuissa, line 1884: Not sure I understand this... for LDS (wildcard) removal can be done by sending the rest of the listeners.

test/integration/ads_integration_test.cc
```diff
@@ -1749,149 +1749,172 @@ TEST_P(AdsIntegrationTest, ContextParameterUpdate) {
...
+                                      {route_name_0}, {}, /*expect_node=*/isSotw()));
+  const auto route_config = buildRouteConfig(route_name_0, cluster_name);
   sendDiscoveryResponse<envoy::config::route::v3::RouteConfiguration>(
-      Config::TypeUrl::get().RouteConfiguration, {}, {buildRouteConfig(route_name_0, cluster_name)},
-      {}, "1");
+      Config::TypeUrl::get().RouteConfiguration, {route_config}, {route_config}, {}, "1");
 
-  EXPECT_TRUE(compareDiscoveryRequest(Config::TypeUrl::get().Listener, "1", {}, {}, {}));
-  EXPECT_TRUE(compareDiscoveryRequest(Config::TypeUrl::get().RouteConfiguration, "1", {}, {}, {}));
+  EXPECT_TRUE(compareDiscoveryRequest(Config::TypeUrl::get().Listener, "1", {}, {}, {},
```
adisuissa, line 1853: I'm not sure why node is being sent. I think you can set `ads_config->set_set_node_on_first_message_only(true);` in the initialization phase.

test/integration/ads_integration_test.cc
```diff
@@ -1749,149 +1749,172 @@ TEST_P(AdsIntegrationTest, ContextParameterUpdate) {
...
+      buildListener("xdstp://test/envoy.config.listener.v3.Listener/foo-listener/"
+                    "baz?xds.node.cluster=cluster_name&xds.node.id=node_name",
+                    route_name_1);
   sendDiscoveryResponse<envoy::config::listener::v3::Listener>(
-      Config::TypeUrl::get().Listener, {},
-      {buildListener("xdstp://test/envoy.config.listener.v3.Listener/foo-listener/"
-                     "baz?xds.node.cluster=cluster_name&xds.node.id=node_name",
-                     route_name_1)},
-      {}, "2");
+      Config::TypeUrl::get().Listener, {second_listener}, {second_listener}, {}, "2");
```
adisuissa, line 1867: To keep consistency between SotW and Delta, I believe the 2nd argument needs to pass all the relevant listeners.

adisuissa (commented):
Thanks! Left a few more comments.
/wait

abeyad (commented):
I implemented xdstp sotw on the legacy mux as well.  PTAL

source/common/config/grpc_mux_impl.cc
```diff
@@ -355,10 +386,24 @@ void GrpcMuxImpl::processDiscoveryResources(const std::vector<DecodedResourcePtr
       continue;
     }
     std::vector<DecodedResourceRef> found_resources;
+    absl::flat_hash_set<std::string> added_resources;
     for (const auto& watched_resource_name : watch->resources_) {
       auto it = resource_ref_map.find(watched_resource_name);
       if (it != resource_ref_map.end()) {
         found_resources.emplace_back(it->second);
+        added_resources.insert(watched_resource_name);
```
adisuissa, line 394: I'm not sure I fully understand why `added_resources` is needed.
It seems that both wildcard and non-wildcard watchers will need to have the resource as part of the `found_resources`, no?

nit: in the TODO I think it would be better to have a dedicated map of wildcard watches, so the search can be inverted (go over the resources, and find the wildcard-watches that is relevant). WDYT?

adisuissa (commented):
Thanks, overall LGTM mostly nits.

source/common/config/grpc_mux_impl.cc
abeyad: The reason for added_resources is because if we have a watch on a wildcard, we loop through the resource_ref_map, so we could end up adding the same entry from the resource_ref_map more than once.  E.g. if we are watching `xdstp://test/foo` and `xdstp://test/*`, and the resource_ref_map contains `xdstp://test/foo`.  However, instead of having the set keep track of whether we've already inserted it, I just did an erase from the resource_ref_map once we've inserted into found_resources.

RE the TODO, sorry I think I wasn't clear, what I meant is basically what you're suggesting.  I tried clearing up the language.b

adisuissa: I guess my comment wasn't clear. In the case of watching `xdstp://test/foo` and `xdstp://test/*` why aren't both watchers notified with the resource? I guess it could also raise another question: in the current implementation why does singleton-watcher take precedence over collection-watcher?

abeyad: Discussed offline, removed the deduping check as we should not have singleton and collection resources in the same watch; and I commented the assumption.

source/common/config/grpc_mux_impl.cc
```diff
@@ -359,6 +390,19 @@ void GrpcMuxImpl::processDiscoveryResources(const std::vector<DecodedResourcePtr
       auto it = resource_ref_map.find(watched_resource_name);
       if (it != resource_ref_map.end()) {
         found_resources.emplace_back(it->second);
+      } else if (isXdstpWildcard(watched_resource_name)) {
+        // See if the resources match the xdstp wildcard entry.
+        // Note: the assumption is that we don't have singleton and collection subscriptions in the
+        // same watch. For that reason, we do not need to check for duplicates in adding to
+        // `found_resources`.
```
adisuissa, line 397: nit: even if we there are a singleton and collection watchers, I think it is ok, because of the `else if` above.
Maybe remove or rephrase it as follows:
```suggestion
        // Note: although it is unlikely that Envoy will need to support a resource that is mapped to both a singleton and collection watch,
        // this code still supports this use case.
```

adisuissa (commented):
Thanks! Left one more small nit.

adisuissa (dismissed):
LGTM, thanks!

abeyad:
Thanks @adisuissa !  Adding @lizan for cross-company review.

source/common/config/grpc_mux_impl.cc
```diff
@@ -31,6 +32,26 @@ class AllMuxesState {
   absl::flat_hash_set<GrpcMuxImpl*> muxes_;
 };
 using AllMuxes = ThreadSafeSingleton<AllMuxesState>;
+
+// Returns true if `resource_name` contains the wildcard xdstp resource, for example:
+// xdstp://test/envoy.config.cluster.v3.Cluster/foo-cluster/*
+// xdstp://test/envoy.config.cluster.v3.Cluster/foo-cluster/*?node.name=my_node
+bool isXdstpWildcard(const std::string& resource_name) {
```
lizan, line 39: Nit: XdsTp? Seems other place (hasXdsTpScheme) is using this instead of Xdstp.

source/common/config/grpc_mux_impl.cc
```diff
@@ -31,6 +32,26 @@ class AllMuxesState {
...
+         (absl::EndsWith(resource_name, "/*") || absl::StrContains(resource_name, "/*?"));
+}
+
+// Converts the xdstp resource name to its wildcard equivalent.
+// Must only be called on xdstp resource names.
+std::string convertToWildcard(const std::string& resource_name) {
+  ASSERT(XdsResourceIdentifier::hasXdsTpScheme(resource_name));
+  xds::core::v3::ResourceName xdstp_resource = XdsResourceIdentifier::decodeUrn(resource_name);
+  const auto pos = xdstp_resource.id().find_last_of('/');
+  xdstp_resource.set_id(pos == std::string::npos ? "*" : xdstp_resource.id().substr(0, pos) + "/*");
```
lizan, line 50: Nit: absl::StrCat instead of +

lizan (commented):
LGTM modulo two bits. Defer to @adisuissa for final merge.

source/common/config/grpc_mux_impl.cc
abeyad: done

source/common/config/grpc_mux_impl.cc
abeyad: done

abeyad:
Thanks for the review, @lizan !  @adisuissa can you kindly re-approve since I pushed up a commit to address Lizan's feedback, thanks!

adisuissa (approved):
LGTM thanks.
@lizan approved as well, so this can be merged.

PR #25121, opened 2023-01-24
datadog: span portion of new tracing library

dgoffredo:

This revision introduces a new library dependency for Datadog tracing, and adds a component that will be used in the future when more of the new library is integrated into the Datadog tracing extension.

This revision also contains all of the changes from https://github.com/envoyproxy/envoy/pull/25003, on which it depends. When that PR is merged, the diff of this PR will shrink.


dgoffredo:
> ERROR: From ./test/extensions/tracers/datadog/span_test.cc
> ERROR: ./test/extensions/tracers/datadog/span_test.cc:114: Don't use mktime; use absl::Time equivalent instead
> ERROR: check format failed. diff has been applied'

I guess I'll use `absl::FromCivil` and `absl::ToChronoTime` instead.

test/extensions/tracers/datadog/span_test.cc
```diff
@@ -100,6 +100,22 @@ struct TestSetup {
   }
 };
 
+std::chrono::system_clock::time_point now() {
```
jmarantz, line 103: Use TimeSource, or maybe SimulatedTimeSystem for this.

https://github.com/envoyproxy/envoy/blob/main/test/test_common/simulated_time_system.h

test/extensions/tracers/datadog/span_test.cc
```diff
@@ -0,0 +1,294 @@
...
+    rule.sample_rate = 0;
+    config.trace_sampler.rules.push_back(std::move(rule));
+    return config;
+  }
+};
+
+std::chrono::system_clock::time_point now() {
+  // 14:45 May 6th, 2010 in New York
+  absl::CivilSecond datetime{2010, 5, 3, 14, 45};
+  return absl::ToChronoTime(absl::FromCivil(datetime, absl::FixedTimeZone(-4 * 60 * 60)));
```
jmarantz, line 108: hm.

The back-story here is we didn't used have all the absl time functions in Envoy at all, but we wanted to limit sleeps and other use of real time in tests, so we have a bunch of lint checks. The intent of the lint checks is to use timers and/or simulated time or mock time rather than sleeps and similar functions.

You avoid the lint checks with this, but I feel like this subverts the intent.

Why do we need all this? Can we use simulated time instead?

dgoffredo: `Span::spawnChild` and `Span::log` are part of the Envoy tracing interface that we must implement. Both have a parameter of type `SystemTime`. Any caller of one of those member functions must provide a `SystemTime` value.

For the purposes of these tests, the value of the `SystemTime` does not matter, so long as I know what it is. Initially I used the current time. Then I constructed a fixed time using `mktime`. Now I've constructed a fixed time using `absl::ToChronoTime`.

How could simulated time be used here?

jmarantz: SimulatedTimeSystem has a method systemTime() which returns a SystemTime.

It gets initialized, by default, from the current real-time when the test is run, but it does not advance with real-time after that; it only advances due to methods called on the SimulatedTimeSystem to advance time.

So for your purposes you can just instantiate one and call systemTime() on it.

test/extensions/tracers/zipkin/zipkin_core_types_test.cc:118 is one usage that might be similar enough.

dgoffredo: Thanks, will do.

source/extensions/tracers/datadog/span.cc
```diff
@@ -0,0 +1,118 @@
...
+
+  TraceContextWriter writer{trace_context};
+  span_->inject(writer);
+}
+
+Tracing::SpanPtr Span::spawnChild(const Tracing::Config&, const std::string& name,
+                                  SystemTime start_time) {
+  if (!span_) {
+    // I don't expect this to happen. This means that `spawnChild` was called
+    // after `finishSpan`.
```
jmarantz, line 76: should we use an ENVOY_BUG here (logs a message in production build, asserts in debug/default builds)?

source/extensions/tracers/datadog/span.h
```diff
@@ -0,0 +1,67 @@
...
+ * optional<datadog::tracing::Span> is not empty and member functions are
+ * forwarded to it, and another state when the optional<datadog::tracing::Span>
+ * is empty and member functions have no effect.
+ */
+class Span : public Tracing::Span {
+public:
+  explicit Span(datadog::tracing::Span&& span);
+
+  const datadog::tracing::Optional<datadog::tracing::Span>& impl() const;
+
```
jmarantz, line 37: nit: per Envoy style you don't need all the blank lines between the override definitions.

Typically we also put a comment above the override block saying where the definition is (which might be relevant for multiple inheritance or deep hierarchies).

source/extensions/tracers/datadog/time_util.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+/**
+ * This file contains functions related to time points and durations.
+ *
+ * Envoy has a time type that contains both a system time point and a steady
```
jmarantz, line 6: nit: Envoy has a TimeSource abstraction that provides the system time point and a steady ....


The time types themselves just have one of those.

test/extensions/tracers/datadog/span_test.cc
```diff
@@ -0,0 +1,290 @@
...
+  nlohmann::json config_json() const override {
+    return nlohmann::json::object({{"type", "Envoy::Extensions::Tracers::Datadog::MockCollector"}});
+  }
+
+  ~MockCollector() override = default;
+
+  std::vector<std::vector<std::unique_ptr<datadog::tracing::SpanData>>> chunks;
+};
+
+struct TestSetup {
```
jmarantz, line 73: any reason not to make this a text fixture (make it a class derived from testing::Test, and use TEST_F macro below).

Also, independent of that, Envoy style is to suffix struct member vars with _ (overriding Google style).

jmarantz (commented):
This looks great; really nicely tested. Just a few nits.

I think @moderation should look at this also for span semantics.

source/extensions/tracers/datadog/span.cc
dgoffredo: That might be worth doing.

I'd reserve `assert` and friends for cases where the detected behavior is in violation of the component's contract. Because this `Span` has the escape hatch of going into "null-mode," it's not necessary. By that reasoning, perhaps the code comment is not necessary either.

You mentioned the idea of @moderation looking over the behavior of this `Span`. Maybe they can decide whether "spawn-after-finish" ought to be considered forbidden in these tracing implementations.

source/extensions/tracers/datadog/time_util.h
dgoffredo: Yeah, I noticed that last week! Glad you caught it, I'll change the wording to be accurate.

test/extensions/tracers/datadog/span_test.cc
dgoffredo: Will change the member names.

There was a voice in my head saying "you don't need `setUp` and `tearDown` methods -- a fixure obscure the simplicity in this case." Looking at it again, maybe this departure from common practice distracts more than simplifies.

jmarantz: You would not define SetUp or TearDown -- there are empty default impls for those in testing::Test

wbpcode:
Please merge main and resolve the conflict first. Thanks.

source/extensions/tracers/datadog/span.cc
```diff
@@ -0,0 +1,118 @@
...
+  return std::make_unique<Span>(span_->create_child(config));
+}
+
+void Span::setSampled(bool sampled) {
+  if (!span_) {
+    return;
+  }
+
+  auto priority = int(sampled ? datadog::tracing::SamplingPriority::USER_KEEP
+                              : datadog::tracing::SamplingPriority::USER_DROP);
```
wbpcode, line 95: prefer `static_cast` here

source/extensions/tracers/datadog/time_util.h
```diff
@@ -0,0 +1,45 @@
...
+ * (end.tick - begin.tick). So, we need to get a steady clock time from a given system
+ * clock time. The scheme is to measure the current system/steady time,
+ * compare the system part with the given system time, and then adjust the
+ * measured steady time accordingly. This is correct if the system clock has not
+ * been adjusted since the given system time was measured. It's incorrect
+ * otherwise, hence only an estimate. This conversion is performed by the
+ * estimateTime function.
+ */
+
+#include <datadog/clock.h>
```
wbpcode, line 22: nit: `#include "datadog/clock.h"`

source/extensions/tracers/datadog/span.cc
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  span_->set_tag(name, value);
+}
+
+void Span::log(SystemTime, const std::string&) {
+  // Datadog spans don't have in-bound "events" or "logs".
+}
+
+void Span::finishSpan() { span_.reset(); }
```
wbpcode, line 60: Then datadog span will report the data to collector here automatically?

wbpcode (commented):
LGTM overall. It's great. Only some minor comments.

jmarantz (dismissed):
this all looks good from my perspective modulo @wbpcode comments.

I think it might need a main merge to kick off CI also.

source/extensions/tracers/datadog/span.cc
dgoffredo: If finishing this span causes the associated [TraceSegment][1] to be finished, then the spans for that segment will be sent to the collector.

[1]: https://github.com/DataDog/dd-trace-cpp/blob/main/src/datadog/trace_segment.h

source/extensions/tracers/datadog/time_util.h
dgoffredo: I'll change this here and in subsequent PRs, and will have to make the corresponding change in code that has already been merged.

dgoffredo:
https://github.com/envoyproxy/envoy/pull/24826 :frowning_face:

dgoffredo:
```
test/extensions/tracers/datadog/span_test.cc:292:47: error: expression with side effects will be evaluated despite being used as an operand to 'typeid' [-Werror,-Wpotentially-evaluated-expression]
  EXPECT_EQ(typeid(Tracing::NullSpan), typeid(*child));
                                              ^
1 error generated.
```
That's a new one.

I suppose instead of this:
```C++
  const Tracing::SpanPtr child =
      span.spawnChild(Tracing::MockConfig{}, "child", time_.timeSystem().systemTime());
  EXPECT_NE(nullptr, child);
  EXPECT_EQ(typeid(Tracing::NullSpan), typeid(*child));
```
it wants this:
```
  const Tracing::SpanPtr child =
      span.spawnChild(Tracing::MockConfig{}, "child", time_.timeSystem().systemTime());
  EXPECT_NE(nullptr, child);
  Tracing::Span& child_span = *child;
  EXPECT_EQ(typeid(Tracing::NullSpan), typeid(child_span));
```
I'll make that change and see if it helps.

jmarantz:
@moderation  I've lgtm'd can you look this over and merge if you are good with it?

moderation:
@jmarantz I don't have merge permissions but happy to see this move ahead as it will allow us to remove some old dependencies. Maybe @phlax can merge?

source/extensions/tracers/datadog/span.cc
```diff
@@ -0,0 +1,117 @@
...
+
+void Span::setBaggage(absl::string_view, absl::string_view) {
+  // not implemented
+}
+
+std::string Span::getTraceIdAsHex() const {
+  if (!span_) {
+    return std::string{};
+  }
+  return absl::StrCat(absl::Hex(span_->id()));
```
zirain, line 111: @dgoffredo Is this right?

PR #25123, opened 2023-01-24
CDS mobile builder API and integration test

caschoener:
Adds a builder API to enable CDS, using ADS to configure the address.  Then a simple unit test which adds a cluster with CDS and verifies it's been added


mobile/library/cc/engine_builder.cc
```diff
@@ -938,6 +953,12 @@ std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap> EngineBuilder::generate
     ads_config->set_api_type(api_type_enum);
     ads_config->add_grpc_services()->mutable_google_grpc()->set_target_uri(target_uri);
   }
+  if (enable_cds_) {
+    auto* cds_config = bootstrap->mutable_dynamic_resources()->mutable_cds_config();
```
abeyad, line 957: We'll also need to support xdstp wildcard, once https://github.com/envoyproxy/envoy/pull/25115 lands (which should be today), which means we'd have to set https://github.com/envoyproxy/envoy/blob/main/test/integration/ads_integration_test.cc#L1689.  If you're ready to submit this PR before the other PR lands, if you can please add a TODO to add xdstp support and come back to it.  Thanks!

mobile/library/cc/engine_builder.h
```diff
@@ -65,7 +65,9 @@ class EngineBuilder {
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
   // Adds an RTDS layer to default config. Requires that ADS be configured
-  EngineBuilder& addRtdsLayer(const std::string& layer_name, int timeout_seconds = 5);
+  EngineBuilder& addRtdsLayer(const std::string& layer_name, const int timeout_seconds = 5);
+  // Adds a CDS layer to default config. Requires that ADS be configured
```
abeyad, line 69: super nit: add a period at the end of "configured", same for the RTDS comment above

mobile/test/common/integration/base_client_integration_test.cc
```diff
@@ -249,6 +249,38 @@ testing::AssertionResult BaseClientIntegrationTest::waitForCounterGe(const std::
   return testing::AssertionSuccess();
 }
 
+uint64_t BaseClientIntegrationTest::getGaugeValue(const std::string& name) {
```
abeyad, line 252: we don't need these changes, BaseClientIntegrationTest already has these functions

mobile/test/common/integration/cds_integration_test.cc
```diff
@@ -0,0 +1,55 @@
...
+    default_request_headers_.setScheme("http");
+  }
+
+  void createEnvoy() override {
+    sotw_or_delta_ = sotwOrDelta();
+    const std::string api_type = sotw_or_delta_ == Grpc::SotwOrDelta::Sotw ||
+                                         sotw_or_delta_ == Grpc::SotwOrDelta::UnifiedSotw
+                                     ? "GRPC"
+                                     : "DELTA_GRPC";
+    builder_.addCdsLayer(1);
```
abeyad, line 26: add `/*timeout_seconds=*/` to the `1` so that the reader of the test can clearly know what that param value means

mobile/test/common/integration/cds_integration_test.cc
```diff
@@ -0,0 +1,55 @@
...
+INSTANTIATE_TEST_SUITE_P(IpVersionsClientTypeDelta, CdsIntegrationTest,
+                         DELTA_SOTW_GRPC_CLIENT_INTEGRATION_PARAMS);
+
+TEST_P(CdsIntegrationTest, Basic) {
+  initialize();
+  envoy::config::cluster::v3::Cluster cluster1 = ConfigHelper::buildStaticCluster(
+      ClusterName, fake_upstreams_[0]->localAddress()->ip()->port(),
+      Network::Test::getLoopbackAddressString(ipVersion()), "ROUND_ROBIN");
+  initializeXdsStream();
+  auto cluster_count = getGaugeValue("cluster_manager.active_clusters");
```
abeyad, line 44: use int, instead of auto

mobile/library/cc/engine_builder.cc
```diff
@@ -914,7 +928,8 @@ std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap> EngineBuilder::generate
     sink->mutable_typed_config()->PackFrom(metrics_config);
   }
 
-  if (!rtds_layer_name_.empty() && ads_api_type_.empty()) {
+  bootstrap->mutable_dynamic_resources();
```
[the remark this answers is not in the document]
caschoener: This adds an empty "dynamic resources" object to make it match the yaml, which has one.

mobile/test/common/integration/base_client_integration_test.cc
caschoener: BaseClientIntegrationTest has similar functions getCounterValue and waitForCounterGe, but not the gauge equivalents.

abeyad: ah you're right, missed that, thanks for adding it!

mobile/library/cc/engine_builder.cc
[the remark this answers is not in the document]
abeyad: correct me if i'm wrong, but it seems to me like "dynamic_resources" only exists if CDS is enabled: https://github.com/envoyproxy/envoy/pull/25123/files#diff-46ea49cf2b9b992e3ec1e61b738a06149e6cd7bc00826bf7c034b3142b3db5efR438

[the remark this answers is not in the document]
caschoener: If you look at config.cc above, the tag "dynamic_resources:" is always in.  Just makes the design simpler since we now have two things which could get added to dynamic resources

mobile/library/cc/engine_builder.cc
```diff
@@ -407,8 +414,8 @@ std::string EngineBuilder::generateConfigStr() const {
     insertCustomFilter(filter_config, config_template);
   }
 
-  if (!rtds_layer_name_.empty() && ads_api_type_.empty()) {
-    throw std::runtime_error("ADS must be configured when using RTDS");
+  if ((!rtds_layer_name_.empty() || enable_cds_) && ads_api_type_.empty()) {
+    throw std::runtime_error("ADS must be configured when using XDS");
```
abeyad, line 418: nit: s/XDS/xDS

abeyad: Question: did we decide if we want to make this an EnvoyException instead of runtime_error?

mobile/library/cc/engine_builder.h
```diff
@@ -64,8 +64,10 @@ class EngineBuilder {
   EngineBuilder& enableDrainPostDnsRefresh(bool drain_post_dns_refresh_on);
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
-  // Adds an RTDS layer to default config. Requires that ADS be configured
-  EngineBuilder& addRtdsLayer(const std::string& layer_name, int timeout_seconds = 5);
+  // Adds an RTDS layer to default config. Requires that ADS be configured.
+  EngineBuilder& addRtdsLayer(const std::string& layer_name, const int timeout_seconds = 5);
```
abeyad, line 68: since we are using 5 as the default timeout, we can make this a constant in the header file and use it in the default param values

abeyad (commented):
LGTM minus a few nits I just left!  Also, if you can please update the title of the PR and description to indicate that you've added a CDS layer API that relies on ADS and you've added an integration test for CDS (since none existed before).

mobile/library/cc/engine_builder.cc
caschoener: I don't remember but it seems safe enough to follow the convention of this file.

abeyad (approved):

abeyad:
added JP for cross company review

mobile/library/cc/engine_builder.cc
```diff
@@ -856,6 +875,7 @@ std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap> EngineBuilder::generate
   list->add_patterns()->set_exact("cluster.stats.http2.keepalive_timeout");
   list->add_patterns()->set_prefix("http.hcm.downstream_rq_");
   list->add_patterns()->set_prefix("http.hcm.decompressor.");
+  list->add_patterns()->set_prefix("cluster_manager.");
```
abeyad, line 878: one thing Alyssa pointed out, `cluster_manager.` prefix may have a lot of stats.  We probably want to limit this to the ones we care about.  Might be good to take stock of the metrics and which ones we'll care about for mobile CDS and then only allow-list those patterns.  Also, we probably only want to add this prefix *if* enable_cds_ is true.

abeyad (changes requested):
Left a comment on the stats prefix

caschoener:
> Left a comment on the stats prefix

Done!

mobile/test/common/integration/BUILD
```diff
@@ -41,6 +41,25 @@ envoy_cc_test(
     ],
 )
 
+envoy_cc_test(
+    name = "cds_integration_test",
+    srcs = ["cds_integration_test.cc"],
+    exec_properties = {
+        # TODO(willengflow): Remove this once the sandboxNetwork=off works for ipv4 localhost addresses.
```
jpsim, line 48: @WillEngFlow does this still apply?

jpsim (approved):
No objections from me, as long as @abeyad also reviews

mobile/test/common/integration/BUILD
caschoener: This comment is repeated many times throughout the codebase fwiw.  Whenever this particular exec_property is used

abeyad (approved):

PR #25124, opened 2023-01-24
test: update hyphen matching in regex in version_out_test.sh

TristanWedderburn:
Commit Message: update hyphen matching in regex in version_out_test.sh
Additional Description: We're looking to update the output string of the `--version` command to include more info about the build version in our envoy fork using additional hyphens. In the version_out_test.sh file, we noticed that the regex for these tests could be updated to support that and the current version output by changing the hyphen to be the last character in the `[...]` expression in the third group to avoid it being considered a range.
Risk Level: Low
Testing: successful run of version_out_test.sh using `bazel test //test/exe:version_out_test`
Docs Changes: N/A
Release Notes: N/A


phlax (approved):
lgtm, thanks @TristanWedderburn

PR #25129, opened 2023-01-24
grpc_json_transcoder: to support skip transcoding based on route name.

shuoyang2016:
Additional Description:
grpc_json_transcoder: to support skip transcoding based on route name. We have a requirement to configure istio gateway to skip transcoding for certain routes. However, since istio doesn't support per_filter_config in the route, we cannot use that or other metadata to decorate the route for this feature. Thus we propose to add this naming convention hack until per_filter_config is supported in istio.

Risk Level: Low
Testing: unit test
Docs Changes: N/A
Release Notes: Add an option in grpc_json_transcoder to support skip transcoding based on route naming conventions.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
+++ b/api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
@@ -288,4 +288,15 @@ message GrpcJsonTranscoder {
   //
   // If unset, the current stream buffer size is used.
   google.protobuf.UInt32Value max_response_body_size = 16 [(validate.rules).uint32 = {gt: 0}];
+
+  // If true, the transcoder will honor route naming conventions for the below
+  // suffix applied to the route.
+  //
+  // "_skip_transcoding": the transcoder will skip transcoding for the route.
+  //
+  // This is a workaround when per_filter_config isn't available for the route
+  // config. For example, istio control plane.
+  // TODO: remove this feature after istio supports per filter config in their
+  // first-class route APIs.
+  bool use_route_naming_conventions = 17;
 }
--- a/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
+++ b/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc
@@ -62,6 +62,8 @@ using RcDetails = ConstSingleton<RcDetailsValues>;
 
 namespace {
 
+constexpr absl::string_view SkipTranscodingSuffix = "_skip_transcoding";
+
 const Http::LowerCaseString& trailerHeader() {
   CONSTRUCT_ON_FIRST_USE(Http::LowerCaseString, "trailer");
 }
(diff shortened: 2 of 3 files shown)


shuoyang2016:
@nareddyt @dafudeng

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -288,4 +288,15 @@ message GrpcJsonTranscoder {
...
+  // If true, the transcoder will honor route naming conventions for the below
+  // suffix applied to the route.
+  //
+  // "_skip_transcoding": the transcoder will skip transcoding for the route.
+  //
+  // This is a workaround when per_filter_config isn't available for the route
+  // config. For example, istio control plane.
+  // TODO: remove this feature after istio supports per filter config in their
+  // first-class route APIs.
+  bool use_route_naming_conventions = 17;
```
kyessenov, line 301: Can we come up with something better? I don't think Istio treats xDS names are stable (and likely change with infra changes).

Does EnvoyFilter not have better filtering condition?

shuoyang2016: We thought about using EnvoyFilter to patch the route component. But was told not to do so because it's also not stable. Do you have some ideas?

kyessenov: It's not stable, but so are the rest of Istio APIs to be replaced by Gateway long term. If your usecase (ESF?) can be covered by https://istio.io/latest/blog/2022/getting-started-gtwapi/, I think we need to start making extensions be configured that way. At the very least, Istio could add some annotations on xDS produced by Gateway API per route.

cc @costinm

shuoyang2016: >Istio could add some annotations on xDS produced by Gateway API per route

That would be great. Thanks.

We haven't on boarded Gateway API yet, is there any option to unblock us for the short-term?

@ygong

howardjohn: Definitely -1 to having this in Envoy. I had previously mentioned this may be an acceptable short term solution when I thought you were talking about writing your own custom filter with vendor-specific logic that could be coupled to Istio's naming; making Envoy/XDS core do this seems inappropriate.

shuoyang2016: Thank you both. Yes, unfortunately we have to customize this filter. Looks like there's no way for istio to configure per route behavior of this filter?

nareddyt: Seems like our best option forward is to use `EnvoyFilter`, what are the concerns about stability for it? I feel it is not a huge concern if we have the proper integration testing in place?

zuercher:
Is this PR still active? The discussion implies perhaps this is being handled in istio.

Issue #24682, opened 2022-12-23
Local Rate Limit With Descriptors: Are tokens from default bucket consumed? 


diegomazine:
*Title*: *Local Rate Limit With Descriptors: Are tokens from default bucket consumed?*

*Description*:

The docs about [Using rate limit descriptors for local rate limiting](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#config-http-filters-local-rate-limit) states that If there is no matching descriptor entries, the default token bucket is used:

> The local descriptor’s token bucket settings are then used to decide if the request should be rate limited or not depending on whether the local descriptor’s entries match the route’s rate limit actions descriptor entries. If there is no matching descriptor entries, the default token bucket is used.

It gives me the idea that no token would be consumed from the default token bucket if there is a match between the request and the descriptors. However looking at [this comment](https://github.com/envoyproxy/envoy/pull/20869#discussion_r857278484) and evaluating the Local Rate Limiter implementation source code, it looks like [tokens from global bucket are consumed for all requests](https://github.com/envoyproxy/envoy/blob/5ff90255dc64852a0d14d910ffa30eb070b510bc/source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc#L173), even when descriptors match.

To confirm it, I have done some tests, as shown below:

- Global token bucket containing less tokens than the descriptor token bucket:

Envoy configuration:
<img width="943" alt="Envoy Configuration" src="https://user-images.githubusercontent.com/53839801/209372381-3b3b4e14-329b-49e8-bf68-fbc0044ed2ef.png">

Requests:
<img width="956" alt="Requests" src="https://user-images.githubusercontent.com/53839801/209376798-685437bb-35fc-4425-a2a3-d117853ca4dd.png">

In this scenario, although we have configured a rate limit of 5 requests each 30 seconds to client application `app1`, it was able to execute only one, since the global bucket was configured with 1 token each 30 seconds.

- Global token bucket containing more tokens than the descriptor token bucket:

Envoy configuration:
<img width="955" alt="image" src="https://user-images.githubusercontent.com/53839801/209378302-be57ef76-bc5f-4278-b3fd-2c81283dc9e5.png">

Requests:
<img width="981" alt="image" src="https://user-images.githubusercontent.com/53839801/209378549-8ff9f805-7079-43d8-8cf3-660e57a4372e.png">

[...]

PR #25139, opened 2023-01-24
local_ratelimit: burn tokens from local descriptors bucket first

diegomazine:
**Commit Message:** Originally, since tokens from the default bucket were burned before local descriptors being validated, requests that matched a specific descriptor might consume all tokens from the global bucket (even when rate limited by the overridden configuration) and prevent requests that matched other descriptors from being accepted. The local rate limit filter was updated to burn tokens from local descriptor's bucket before burning tokens from the default bucket, thus requests rate limited due to an descriptor overridden configuration will not consume tokens from the default bucket. 

**Additional Description:** [Local Rate Limit With Descriptors: Are tokens from default bucket consumed?](https://github.com/envoyproxy/envoy/issues/24682)

**Risk Level:** Low

**Testing:** Local rate limit unit tests were updated and all existing tests were executed.

**Docs Changes:** N/A

**Release Notes:** Done.


source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
```diff
@@ -90,9 +90,9 @@ LocalRateLimiterImpl::~LocalRateLimiterImpl() {
 }
 
 void LocalRateLimiterImpl::onFillTimer() {
+  refill_counter_++;
   onFillTimerHelper(tokens_, token_bucket_);
   onFillTimerDescriptorHelper();
-  refill_counter_++;
```
wbpcode, line 95: Is this change necessary?

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
```diff
@@ -167,29 +167,21 @@ OptRef<const LocalRateLimiterImpl::LocalDescriptorImpl> LocalRateLimiterImpl::de
...
+            return false;
           }
           break;
         }
       }
     }
   }
-  return allow;
+  // Since global token is not sorted, so we suggest it should be larger than other descriptors.
+  return requestAllowedHelper(tokens_);
```
wbpcode, line 184: Typically, there is no perfect approach for this problem. Seems this a little better than the old behavior.

But I am not sure if a runtime guard is necessary for this behavior change?  cc @mattklein123 

cc @StarryVae for you are working on local ratelimit related things.

test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc
```diff
@@ -386,27 +397,23 @@ TEST_F(LocalRateLimiterDescriptorImplTest, TokenBucketMultipleTokensPerFillDescr
...
-  dispatcher_.globalTimeSystem().advanceTimeAndRun(std::chrono::milliseconds(50), dispatcher_,
-                                                   Envoy::Event::Dispatcher::RunType::NonBlock);
-  EXPECT_CALL(*fill_timer_, enableTimer(std::chrono::milliseconds(50), nullptr));
-  fill_timer_->invokeCallback();
+  for (int i = 0; i < 2; i++) {
+    dispatcher_.globalTimeSystem().advanceTimeAndRun(std::chrono::milliseconds(50), dispatcher_,
+                                                     Envoy::Event::Dispatcher::RunType::NonBlock);
+    EXPECT_CALL(*fill_timer_, enableTimer(std::chrono::milliseconds(50), nullptr));
+    fill_timer_->invokeCallback();
+  }
```
wbpcode, line 416: Sorry, but I did't get why you make these change?

wbpcode (commented):
Sorry for the delay. Thanks for your great contribution. Some comments are added.

I am a little worry about that this behavior change will impact some exist users. Although we never promise the order of token comsuption.

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
diegomazine: Actually, I have changed it due to the logic utilized to check whether the bucket refil should occur. Bellow is a snippet of the code

```c++
void LocalRateLimiterImpl::onFillTimerDescriptorHelper() {
  for (const auto& descriptor : descriptors_) {
    // Descriptors are refilled every Nth timer hit where N is the ratio of the
    // descriptor refill interval over the global refill interval. For example,
    // if the descriptor refill interval is 150ms and the global refill
    // interval is 50ms, this descriptor is refilled every 3rd call.
    if (refill_counter_ % descriptor.multiplier_ == 0) {
      onFillTimerHelper(*descriptor.token_state_, descriptor.token_bucket_);
    }
  }
}
```

Note that the refill will occur whenever the remainder of the division of `refill_counter_` by `descriptor.multiplier_` is zero. However since `refill_counter` is initialized with zero, all descriptor buckets will be refilled the first time the method executes. For example, consider the rate limit configuration below:

```yaml
- match:
    prefix: "/"
  route:
    cluster: app
    priority: "DEFAULT"
    rate_limits:
    - actions:
      - request_headers:
          header_name: x-client-application
          descriptor_key: client_application
          skip_if_absent: true
  typed_per_filter_config:
    envoy.filters.http.local_ratelimit:
      "@type": type.googleapis.com/envoy.extensions.filters.http.local_ratelimit.v3.LocalRateLimit
      stat_prefix: local_rate_limiter
      token_bucket:
        max_tokens: 10
        tokens_per_fill: 10
        fill_interval: 10s
      filter_enabled:
        runtime_key: local_rate_limit_enabled
        default_value:
          numerator: 100
          denominator: HUNDRED
      filter_enforced:
        runtime_key: local_rate_limit_enforced
        default_value:
          numerator: 100
          denominator: HUNDRED
      descriptors:
      - entries:
        - key: client_application
          value: app1
        token_bucket:
          max_tokens: 2
          tokens_per_fill: 2
          fill_interval: 30s
      - entries:
        - key: client_application
          value: app2
        token_bucket:
          max_tokens: 4
          tokens_per_fill: 4
          fill_interval: 30s
```

In this case, the default bucket will be refilled every 10 seconds and the descriptors bucket every 30s.

The method `LocalRateLimiterImpl::onFillTimer` will be called after 10 seconds, and te default bucket will be refilled as it should. Since `refill_counter_` is zero all descriptor buckets will also be refilled although they should only be after 30 seconds.

Executing `refill_counter_++;` before the method `onFillTimerDescriptorHelper` will ensure that descriptor buckets will be refilled every Nth timer hit.

test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc
diegomazine: The existing tests utilized two different ways to implement it: 

1. [Duplicating the calls](https://github.com/envoyproxy/envoy/blob/9d9940b49f642fcf483ac8ff387ecb521dcf94fc/test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc#L388):
```c++
  // 0 -> 2 tokens
  dispatcher_.globalTimeSystem().advanceTimeAndRun(std::chrono::milliseconds(50), dispatcher_,
                                                   Envoy::Event::Dispatcher::RunType::NonBlock);
  EXPECT_CALL(*fill_timer_, enableTimer(std::chrono::milliseconds(50), nullptr));
  fill_timer_->invokeCallback();
  dispatcher_.globalTimeSystem().advanceTimeAndRun(std::chrono::milliseconds(50), dispatcher_,
                                                   Envoy::Event::Dispatcher::RunType::NonBlock);
  EXPECT_CALL(*fill_timer_, enableTimer(std::chrono::milliseconds(50), nullptr));
  fill_timer_->invokeCallback();
```

2. [Using for loops](https://github.com/envoyproxy/envoy/blob/9d9940b49f642fcf483ac8ff387ecb521dcf94fc/test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc#L545):
```c++
  // 0 -> 2 tokens for descriptor_
  for (int i = 0; i < 60; i++) {
    dispatcher_.globalTimeSystem().advanceTimeAndRun(std::chrono::milliseconds(50), dispatcher_,
                                                     Envoy::Event::Dispatcher::RunType::NonBlock);
    EXPECT_CALL(*fill_timer_, enableTimer(std::chrono::milliseconds(50), nullptr));
    fill_timer_->invokeCallback();
  }
```

Since I had to update the tests to validate the changes that were implemented, I took the opportunity to standardize how it is done.

In my opinion the second format is clearer and easier to understand than the first one.

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
wbpcode: get it. Thanks for your detailed explanation.

test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc
```diff
@@ -516,10 +530,6 @@ TEST_F(LocalRateLimiterDescriptorImplTest, TokenBucketDifferentDescriptorStatus)
   // Limited by global token
```
wbpcode, line 530: I think this comment is also wrong?

wbpcode (commented):
LGTM overall. One nit comment is added. Thanks. 😄 

cc @mattklein123 for final review or merge.

test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc
diegomazine: You are right.

Actually, besides the comment, I think the [two validations right after it](https://github.com/envoyproxy/envoy/blob/fa34d0cb70af5ce38383b12f2026c389261b7001/test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc#L531) will not be necessary since the same checks are performed [a little further down](https://github.com/envoyproxy/envoy/blob/fa34d0cb70af5ce38383b12f2026c389261b7001/test/extensions/filters/common/local_ratelimit/local_ratelimit_test.cc#L539).

```c++
// Limited by global token
EXPECT_FALSE(rate_limiter_->requestAllowed(descriptor_));
EXPECT_EQ(rate_limiter_->maxTokens(descriptor_), 2);

// 0 -> 0 tokens for descriptor_ and descriptor2_
EXPECT_FALSE(rate_limiter_->requestAllowed(descriptor2_));
...
EXPECT_FALSE(rate_limiter_->requestAllowed(descriptor_));
EXPECT_EQ(rate_limiter_->maxTokens(descriptor_), 2);
```

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
mattklein123: Can you add some comments in this function if the order matters this much? It will help future readers understand and not move it.

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
mattklein123: I'm on the fence about a guard. It's probably OK without. I don't understand the comment though about sorting and size. Can you expand the comment?

mattklein123 (changes requested):
Thanks at a high level this change makes sense to me. Just some comment comments, thank you.

/wait

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
diegomazine: For sure.

source/extensions/filters/common/local_ratelimit/local_ratelimit_impl.cc
diegomazine: Sure. I've added an extra comment to give more context about how descriptor tokens are consumed.

```c++
bool LocalRateLimiterImpl::requestAllowed(                                                
    absl::Span<const RateLimit::LocalDescriptor> request_descriptors) const {             
  // Matched descriptors will be sorted by tokens per second and tokens consumed in order.
  // In most cases, if one of them is limited the remaining descriptors will not consume  
  // their tokens.                                                                        
  if (!descriptors_.empty() && !request_descriptors.empty()) {                            
    for (const auto& descriptor : sorted_descriptors_) {                                  
      for (const auto& request_descriptor : request_descriptors) {                        
        if (descriptor == request_descriptor) {                                           
          // Descriptor token is not enough.                                              
          if (!requestAllowedHelper(*descriptor.token_state_)) {                          
            return false;                                                                 
          }                                                                               
          break;                                                                          
        }                                                                                 
      }                                                                                   
    }                                                                                     
  }                                                                                       
  // Since global tokens are not sorted, it should be larger than other descriptors.      
  return requestAllowedHelper(tokens_);                                                   
}                                                                                         
```

mattklein123 (approved):
Thanks!

PR #25149, opened 2023-01-25
Maglev: Compact maglev table

KBaichoo:
Commit Message: Implement a compact maglev table, and pick between the implementations based on the cost.
Additional Description:
Estimated savings are as follows using default maglev table size:
Screenshot 2023-01-26 10 20 23 AM

Running existing benchmark: 
```
bazel test -c opt //test/common/upstream:load_balancer_benchmark_test
```
With Original Maglev:
```
-----------------------------------------------------------------------------
[2023-01-26 16:21:17.582][19][critical] [test/benchmark/main.cc:103] Expensive benchmarks are being skipped; see test/README.md for more information
2023-01-26T16:21:17+00:00
Running /b/f/w/bazel-out/k8-opt/bin/test/common/upstream/load_balancer_benchmark_test.runfiles/envoy/test/common/upstream/load_balancer_benchmark
Run on (2 X 2800.28 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x1)
  L1 Instruction 32 KiB (x1)
  L2 Unified 1024 KiB (x1)
  L3 Unified 33792 KiB (x1)
Load Average: 0.15, 0.12, 0.04
----------------------------------------------------------------------------------------------------------------------
Benchmark                                                            Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------------------------------

benchmarkMaglevLoadBalancerBuildTable/100                         4.86 ms         4.87 ms            1 memory=1057.22k memory_per_host=10.572k
benchmarkMaglevLoadBalancerBuildTable/200                         4.97 ms         4.97 ms            1 memory=1057.22k memory_per_host=5.286k
benchmarkMaglevLoadBalancerBuildTable/500                         5.42 ms         5.42 ms            1 memory=1057.22k memory_per_host=2.114k

benchmarkMaglevLoadBalancerChooseHost/100/100000                  9.79 ms         9.79 ms            1 mean_hits=1000 relative_stddev_hits=0.029227 stddev_hits=29.227
benchmarkMaglevLoadBalancerChooseHost/200/100000                  10.6 ms         10.6 ms            1 mean_hits=500 relative_stddev_hits=0.0415841 stddev_hits=20.7921
benchmarkMaglevLoadBalancerChooseHost/500/100000                  12.1 ms         12.1 ms            1 mean_hits=200 relative_stddev_hits=0.0687299 stddev_hits=13.746

benchmarkMaglevLoadBalancerHostLoss/500/1/10000                   16.9 ms         16.8 ms            1 host_loss_over_N_optimal=0.2 percent_different=1
benchmarkMaglevLoadBalancerHostLoss/500/2/10000                   16.8 ms         16.7 ms            1 host_loss_over_N_optimal=0.4 percent_different=1.65
benchmarkMaglevLoadBalancerHostLoss/500/3/10000                   16.4 ms         16.3 ms            1 host_loss_over_N_optimal=0.6 percent_different=1.9
benchmarkMaglevLoadBalancerWeighted/500/5/1/1/10000               16.6 ms         16.6 ms            1 optimal_percent_different=0 percent_different=0

benchmarkMaglevLoadBalancerWeighted/500/5/1/127/1000              15.9 ms         15.8 ms            1 optimal_percent_different=81.9863 percent_different=80.8
benchmarkMaglevLoadBalancerWeighted/500/5/127/1/10000             19.7 ms         19.6 ms            1 optimal_percent_different=81.9863 percent_different=81.72
benchmarkMaglevLoadBalancerWeighted/500/50/1/127/1000             14.8 ms         14.8 ms            1 optimal_percent_different=49.2188 percent_different=50.8
benchmarkMaglevLoadBalancerWeighted/500/50/127/1/10000            16.9 ms         16.9 ms            1 optimal_percent_different=49.2188 percent_different=50.16
benchmarkMaglevLoadBalancerWeighted/500/95/1/127/1000             14.5 ms         14.4 ms            1 optimal_percent_different=4.95857 percent_different=6.5
benchmarkMaglevLoadBalancerWeighted/500/95/127/1/10000            16.5 ms         16.5 ms            1 optimal_percent_different=4.95857 percent_different=7.25
benchmarkMaglevLoadBalancerWeighted/500/95/25/75/1000             15.2 ms         15.2 ms            1 optimal_percent_different=0.139958 percent_different=0.8
benchmarkMaglevLoadBalancerWeighted/500/95/75/25/10000            16.9 ms         16.9 ms            1 optimal_percent_different=0.139958 percent_different=0.74
```

With Compact Maglev:
```
CPU Caches:
  L1 Data 32 KiB (x1)
  L1 Instruction 32 KiB (x1)
  L2 Unified 1024 KiB (x1)
  L3 Unified 33792 KiB (x1)
Load Average: 0.17, 0.10, 0.04
----------------------------------------------------------------------------------------------------------------------
Benchmark                                                            Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------------------------------------------------
benchmarkMaglevLoadBalancerBuildTable/100                         4.14 ms         4.13 ms            1 memory=67.816k memory_per_host=678
benchmarkMaglevLoadBalancerBuildTable/200                         4.20 ms         4.19 ms            1 memory=77.416k memory_per_host=387
benchmarkMaglevLoadBalancerBuildTable/500                         4.69 ms         4.60 ms            1 memory=90.6k memory_per_host=181

benchmarkMaglevLoadBalancerChooseHost/100/100000                  8.75 ms         8.75 ms            1 mean_hits=1000 relative_stddev_hits=0.029227 stddev_hits=29.227
benchmarkMaglevLoadBalancerChooseHost/200/100000                  9.41 ms         9.41 ms            1 mean_hits=500 relative_stddev_hits=0.0415841 stddev_hits=20.7921
benchmarkMaglevLoadBalancerChooseHost/500/100000                  10.6 ms         10.6 ms            1 mean_hits=200 relative_stddev_hits=0.0687299 stddev_hits=13.746

benchmarkMaglevLoadBalancerHostLoss/500/1/10000                   14.1 ms         14.0 ms            1 host_loss_over_N_optimal=0.2 percent_different=1
benchmarkMaglevLoadBalancerHostLoss/500/2/10000                   13.9 ms         13.8 ms            1 host_loss_over_N_optimal=0.4 percent_different=1.65
benchmarkMaglevLoadBalancerHostLoss/500/3/10000                   13.3 ms         13.3 ms            1 host_loss_over_N_optimal=0.6 percent_different=1.9

benchmarkMaglevLoadBalancerWeighted/500/5/1/1/10000               13.8 ms         13.8 ms            1 optimal_percent_different=0 percent_different=0
benchmarkMaglevLoadBalancerWeighted/500/5/1/127/1000              14.3 ms         14.2 ms            1 optimal_percent_different=81.9863 percent_different=80.8
benchmarkMaglevLoadBalancerWeighted/500/5/127/1/10000             15.7 ms         15.6 ms            1 optimal_percent_different=81.9863 percent_different=81.72
benchmarkMaglevLoadBalancerWeighted/500/50/1/127/1000             12.3 ms         12.2 ms            1 optimal_percent_different=49.2188 percent_different=50.8
benchmarkMaglevLoadBalancerWeighted/500/50/127/1/10000            13.4 ms         13.4 ms            1 optimal_percent_different=49.2188 percent_different=50.16
benchmarkMaglevLoadBalancerWeighted/500/95/1/127/1000             11.6 ms         11.5 ms            1 optimal_percent_different=4.95857 percent_different=6.5
benchmarkMaglevLoadBalancerWeighted/500/95/127/1/10000            13.0 ms         13.0 ms            1 optimal_percent_different=4.95857 percent_different=7.25
benchmarkMaglevLoadBalancerWeighted/500/95/25/75/1000             11.5 ms         11.5 ms            1 optimal_percent_different=0.139958 percent_different=0.8
benchmarkMaglevLoadBalancerWeighted/500/95/75/25/10000            13.2 ms         13.2 ms            1 optimal_percent_different=0.139958 percent_different=0.74

```
In particular I would like to point out the change in memory usage in `benchmarkMaglevLoadBalancerBuildTable`  between the two. I suspect the reason the compact maglev is performing better than the original maglev is due to L1 data cache sizes  -- in practice in a deployment (e.g. load testing QPS the Maglev), I expect both should be similar, but given that Maglev choosing a host is not a cpu bottleneck (and we can easily turn it off).

Risk Level: low-medium, release flag would make it easy to default to original behavior. 
Testing: Added tests
Docs Changes:
Release Notes: Included
Platform Specific Features: `BitArray` does not support windows. 
Runtime guard: Yes


source/common/common/bit_array.h
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  static inline uint64_t LoadUnsignedWord(const void* source) {
+    uint64_t destination;
+    safeMemcpyUnsafeSrc(&destination, source);
+    return destination;
+  }
+
+  // Backing storage for the underlying array of bits.
+  std::unique_ptr<uint8_t[]> array_start_;
```
pradeepcrao, line 114: Why not use `std::vector`, since it'll only be an additional 8 bytes?

pradeepcrao (commented):
Thanks so much for doing this! Very nifty idea :-)

source/common/common/bit_array.h
```diff
@@ -0,0 +1,125 @@
...
+ * inline.
+ *
+ * Limitations to this class are:
+ *  1) It is not thread-safe
+ *  2) It assumes little-endianness and a 64-bit architecture.
+ *  3) It works with a maximum width of 32-bits
+ */
+class BitArray {
+public:
+  BitArray(int width, size_t num_items)
```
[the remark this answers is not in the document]
KBaichoo: They're orthogonal (will comment the ctor for this).

For ex: we might have 10000 1 bit sized elements (e.g. this is a bitmap) or we might have 10000 7 bit sized elements.

source/common/common/bit_array.h
```diff
@@ -0,0 +1,125 @@
...
+#include <cstdint>
+#include <cstring>
+
+#include "source/common/common/assert.h"
+#include "source/common/common/safe_memcpy.h"
+
+namespace Envoy {
+
+/**
+ * BitArray is an array of fixed width bits.
```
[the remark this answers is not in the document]
KBaichoo: done.

source/common/common/bit_array.h
KBaichoo: Makes sure I alloc the exact amount of memory without having to do any shrink to fit or similar dance. 

The vec for 10k might stl vec have size != capacity and alloc additional space.

source/common/common/bit_array.h
```diff
@@ -0,0 +1,125 @@
...
+    safeMemcpyUnsafeSrc(&destination, source);
+    return destination;
+  }
+
+  // Backing storage for the underlying array of bits.
+  std::unique_ptr<uint8_t[]> array_start_;
+  // Pointer to the end of the array. In cases where we allocate a word size of
+  // bytes it's possible that the logical "end" of the e.g. based on num_items
+  // is before this address.
+  const uint8_t* end_;
```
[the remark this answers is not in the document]
KBaichoo: done.

test/common/common/bit_array_test.cc
```diff
@@ -0,0 +1,79 @@
...
+
+TEST(BitArray, ShouldAssertIfInvalidBitWidth) {
+  {
+    const int bit_width = 32;
+    BitArray c(bit_width, 10);
+  }
+
+  {
+    const int bit_width = 33;
+    EXPECT_DEATH({ BitArray c(bit_width, 10); }, "");
```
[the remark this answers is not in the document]
KBaichoo: It will change depending on the *SAN. Just leaving empty to keep it simple.

KBaichoo (commented):
Thanks for the review @pradeepcrao !

source/common/common/bit_array.h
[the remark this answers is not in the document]
pradeepcrao: That makes sense. I was being daft.

pradeepcrao (commented):
The asan failure looks legit.

What do you think about adding a performance test to see how this compares to std::vector?

test/common/common/bit_array_test.cc
```diff
@@ -38,7 +38,7 @@ class BitArrayTest : public testing::TestWithParam<BitArrayParams> {
     vec.reserve(num_items_);
 
     Random::RandomGeneratorImpl gen;
-    uint64_t max_given_bit_width = 1 << bit_width_;
+    uint64_t max_given_bit_width = static_cast<uint64_t>(1) << width;
```
pradeepcrao, line 41: Should this be `bit_width_` ? I don't see where `width` is defined.

source/common/upstream/maglev_lb.cc
```diff
@@ -54,26 +93,111 @@ MaglevTable::MaglevTable(const NormalizedHostWeightVector& normalized_host_weigh
...
+      entry.next_++;
+      entry.count_++;
+      table_index++;
     }
   }
 }
 
-HostConstSharedPtr MaglevTable::chooseHost(uint64_t hash, uint32_t attempt) const {
+MaglevTableSharedPtr
+MaglevTable::createMaglevTable(const NormalizedHostWeightVector& normalized_host_weights,
```
pradeepcrao, line 154: It seems a little wonky to me that a base class method should know about subclasses. WDYT about making this a standalone method?

source/common/upstream/maglev_lb.h
```diff
@@ -62,9 +69,81 @@ class MaglevTable : public ThreadAwareLoadBalancerBase::HashingLoadBalancer,
...
+
+private:
+  void constructImplementationInternals(std::vector<TableBuildEntry>& table_build_entries,
+                                        double max_normalized_weight) override;
+  void logMaglevTable(bool use_hostname_for_hashing) const override;
+
+  // Leverage a BitArray to more compactly fit represent the MaglevTable.
+  // The BitArray will index into the host_table_ which will provide the given
+  // host to load balance to.
+  BitArray table_;
```
pradeepcrao, line 145: There's seems a fair amount of code duplicated between the two derived classes. 

What do you think about having an abstraction around just `table_` and `host_table_` for `CompactMaglevTable` and `table_` for `OriginalMaglevTable`?

source/common/common/bit_array.h
```diff
@@ -0,0 +1,137 @@
...
+
+      const uint64_t mask_to_write = ((static_cast<uint64_t>(1) << bit_width_) - 1)
+                                     << index_of_0th_bit;
+      const uint64_t mask_to_preserve = ~mask_to_write;
+      const uint64_t value_to_store =
+          ((LoadUnsignedWord(byte0) & mask_to_preserve) | (shifted_value & mask_to_write));
+      StoreUnsignedWord(byte0, value_to_store);
+    } else {
+      // We're at the end, so we'll be starting from the last word
+      uint8_t* last_word = const_cast<uint8_t*>(end_) - WordSize;
```
pradeepcrao, line 93: What do you think about padding the allocated space by `WordSize` so we don't need the else code block? Either way is fine with me, but doing so will make the code easier to follow.

Also, could you please add more comments? The code here is very hard for me to follow/reason about.

source/common/upstream/maglev_lb.h
```diff
@@ -27,26 +29,26 @@ struct MaglevLoadBalancerStats {
...
-              MaglevLoadBalancerStats& stats);
-
-  // ThreadAwareLoadBalancerBase::HashingLoadBalancer
-  HostConstSharedPtr chooseHost(uint64_t hash, uint32_t attempt) const override;
+  MaglevTable(uint64_t table_size, MaglevLoadBalancerStats& stats);
+  virtual ~MaglevTable() = default;
 
   // Recommended table size in section 5.3 of the paper.
   static const uint64_t DefaultTableSize = 65537;
+  static const uint64_t MaxNumberOfHostsForCompactMaglev = (static_cast<uint64_t>(1) << 32) - 1;
```
pradeepcrao, line 49: nit: s/const/constexpr

source/common/upstream/maglev_lb.cc
```diff
@@ -4,11 +4,76 @@
...
+bool shouldUseCompactTable(size_t num_hosts, uint64_t table_size) {
+#if defined(WIN32)
+  // BitArray does not support Windows.
+  return false;
+#else
+  if (num_hosts > MaglevTable::MaxNumberOfHostsForCompactMaglev) {
+    return false;
+  }
+
+  const uint64_t original_maglev_cost = 8 * table_size;
```
pradeepcrao, line 17: Is this `sizeof(std::shared_ptr<T>) * table_size` ?

source/common/upstream/maglev_lb.cc
```diff
@@ -54,26 +139,88 @@ MaglevTable::MaglevTable(const NormalizedHostWeightVector& normalized_host_weigh
...
-    for (uint64_t i = 0; i < table_.size(); i++) {
-      const absl::string_view key_to_hash = hashKey(table_[i], use_hostname_for_hashing);
-      ENVOY_LOG(trace, "maglev: i={} address={} host={}", i, table_[i]->address()->asString(),
-                key_to_hash);
+  // Vector to track whether or not a given fixed width bit is set in the
+  // BitArray used as the maglev table.
+  std::vector<bool> occupied(table_size_, false);
+
+  // Iterate through the table build entries as many times as it takes to fill up the table.
+  uint64_t table_index = 0;
```
pradeepcrao, line 168: Since most of the code here is repeated for both original and compact maglev tables, what do you think about my previous comment about wrapping just `table_` and `host_table_` in a class for compact representation, and `table_` from the original maglev table for the "original" representation? 

Will need a solution for tracking "occupied" slots though. You could just keep the `occupied` object in the combined implementation of the algorithm.

pradeepcrao: @jmarantz What do you think? (since you've been tagged as a reviewer).

source/common/upstream/maglev_lb.h
pradeepcrao: I was unclear with my wording. I was suggesting that instead of duplicating the algorithmic details in both derived classes, limit the functionality of the derived classes to just setting/getting the host for an index.

source/common/upstream/maglev_lb.cc
pradeepcrao: Based on my discussion with @KBaichoo offline, I'm okay with having a TODO for now to evaluate this again at a later date.

source/common/common/bit_array.h
KBaichoo: done.

source/common/upstream/maglev_lb.cc
KBaichoo: sg, added a TODO

KBaichoo (commented):
Thanks for the review @pradeepcrao , I think the bit array looks much simpler now with the padding.

source/common/upstream/maglev_lb.cc
```diff
@@ -14,11 +14,12 @@ bool shouldUseCompactTable(size_t num_hosts, uint64_t table_size) {
     return false;
   }
 
-  const uint64_t original_maglev_cost = 8 * table_size;
+  const size_t shared_ptr_size = sizeof(HostConstSharedPtr);
```
pradeepcrao, line 17: nit: `constexpr`

pradeepcrao (commented):
LGTM.

Thanks so much for working on this @KBaichoo !

pradeepcrao (dismissed):

jmarantz (commented):
flushing comments

source/common/common/bit_array.h
```diff
@@ -0,0 +1,148 @@
...
+   * @param num_items the number of elements the bit array must hold.
+   */
+  BitArray(int width, size_t num_items)
+      : array_start_(std::make_unique<uint8_t[]>(bytesNeeded(width, num_items))), bit_width_(width),
+        // This will fit in a uint32_t as with the maximum shift of 32, we'd
+        // subtract one to fit in 32 bits.
+        mask_(static_cast<uint32_t>((static_cast<uint64_t>(1) << width) - 1)),
+        num_items_(num_items) {
+    RELEASE_ASSERT(width <= MaxBitWidth, "Using BitArray with invalid parameters.");
+    static_assert(
```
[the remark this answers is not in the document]
jmarantz: Damn. For some reason I had thought it was. However stackoverflow has found a few options, which I messed around with:
```
#include <cstdio>
#include <cstdint>

#if _WIN32 || _WIN64
    #if _WIN64
        #define POINTER_SIZE 8
    #else
        #define POINTER_SIZE 4
    #endif
#elif __GNUC__
    #if __x86_64__ || __ppc64__
        #define POINTER_SIZE 8
    #else
        #define POINTER_SIZE 4
    #endif
#elif UINTPTR_MAX > UINT_MAX
    #define POINTER_SIZE 8
#else
    #define POINTER_SIZE 4
#endif


#define IMAX_BITS(m) ((m)/((m)%255+1) / 255%255*8 + 7-86/((m)%255+12))
#define UINTPTR_MAX_BITWIDTH IMAX_BITS(UINTPTR_MAX)

int main() {
  fprintf(stderr, "uintpr_max_pointer size=%d\n", POINTER_SIZE);
  fprintf(stderr, "uintpr_max_bitwidth=%ld\n", UINTPTR_MAX_BITWIDTH);
  fprintf(stderr, "uintpr_ptrmax=%lx\n", UINTPTR_MAX);
}
```
yields
```
uintpr_max_pointer size=8
uintpr_max_bitwidth=64
uintpr_ptrmax=ffffffffffffffff
```

hooking on little-endian seems wrong.

https://stackoverflow.com/questions/4079243/how-can-i-use-sizeof-in-a-preprocessor-macro

There were other hacks on the web also. But ultimately it's not too hard to do static-assert but I think it's cleaner to do correct by construction, which you can do by basically depending on compiler-supplied preprocessor macros, which seems OK. I'll let you sort out which mechanism seems most robust and maintainable.

source/common/common/bit_array.h
```diff
@@ -0,0 +1,152 @@
...
+    }
+    // We locate the first byte that contains part of the element located at
+    // the given index.
+    const size_t bit0_offset = index * bit_width_;
+    const size_t byte0_offset = bit0_offset >> 3;
+    uint8_t* byte0 = array_start_.get() + byte0_offset;
+
+    // Find the starting bit within byte0 of this element.
+    // It will be in the range of 0-7.
+    const size_t index_of_0th_bit = bit0_offset & 0x7;
```
jmarantz, line 95: nit: you could probably get away (performance wise) with factoring out the common parts of this with get() into common helper function(s). It should all be inlined by the compiler anyway.

jmarantz (commented):
this is great stuff -- a few more comments. 

LGTM otherwise.

jmarantz:
Also I think a senior maintainer should review after this that's pretty familiar with the LBs. probably @snowp or @mattklein123

source/common/upstream/maglev_lb.h
```diff
@@ -62,9 +64,81 @@ class MaglevTable : public ThreadAwareLoadBalancerBase::HashingLoadBalancer,
...
+
+private:
+  void constructImplementationInternals(std::vector<TableBuildEntry>& table_build_entries,
+                                        double max_normalized_weight) override;
+  void logMaglevTable(bool use_hostname_for_hashing) const override;
+
+  // Leverage a BitArray to more compactly fit represent the MaglevTable.
+  // The BitArray will index into the host_table_ which will provide the given
+  // host to load balance to.
+  BitArray table_;
```
jmarantz, line 140: I have another suggestion which might further improve performance and reduce memory. Maybe for a follow-up.

At a high level, this could be a variable-size data structure. See `InlineStorage` in source/common/common/utility.h. It might break the abstraction between BitArray and CompactMaglevTable so it would be a little messy.

But at the end you'd have one less indirection and 8 fewer bytes: the unique_ptr in BitArray would not be needed as the data would be inline.

KBaichoo: Cool to learn about `InlineStorage` as an available technique. I think here the odds of it helping might be slim as the size of the BitArray object itself would balloon in size and wouldn't fit in cache lines. Which would likely lead to an additional memory access. We can revisit if this becomes an issue.

source/common/common/bit_array.h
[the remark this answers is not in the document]
KBaichoo: You're right re: endianess, removed. 

Only requirement is being 8 byte pointers, I think a lot of the codebase I think assumes 64-bit, so having a static_assert which is a C++ standard seems to be a more maintainable way to state this assumption than toolchain specific elements.

source/common/common/bit_array.h
KBaichoo: Leaning towards keeping the code where it is so it can flow better with the comments, context.

KBaichoo (commented):
Thank you for the very through review @jmarantz , learned some new stuff 🚀

source/common/common/bit_array.h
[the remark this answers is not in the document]
jmarantz: You're probably right; static_assert seems fine now that you've removed ifdefing altogether. Matt will let us know if e-m needs 32-bit. I'm guessing that it does not.

source/common/common/bit_array.h
jmarantz: ok, sgtm

source/common/common/bit_array.h
```diff
@@ -0,0 +1,152 @@
...
+   * Constructs a BitArray.
+   * @param width the fixed width of the array elements. For example, if we need
+   *  to represent 10 distinct values we need at least 4 bits as the width.
+   * @param num_items the number of elements the bit array must hold.
+   */
+  BitArray(int width, size_t num_items)
+      : array_start_(std::make_unique<uint8_t[]>(bytesNeeded(width, num_items))), bit_width_(width),
+        // This will fit in a uint32_t as with the maximum shift of 32, we'd
+        // subtract one to fit in 32 bits.
+        mask_(static_cast<uint32_t>((static_cast<uint64_t>(1) << width) - 1)),
```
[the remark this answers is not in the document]
KBaichoo: The factor I think that helped with the microbenchmark is that the compact table representation e.g. `array_start_` can have much more data fit in L1 data cache (32kb) than the original table representation.

source/common/common/bit_array.h
```diff
@@ -0,0 +1,147 @@
...
+ * word_size of bytes beyond what is necessary to hold the items in the BitArray.
+ *
+ * Methods are deliberately implemented in the header to hint to the compiler to
+ * inline.
+ *
+ * Limitations to this class are:
+ *  1) It is thread compatible.
+ *  2) It assumes a 64-bit architecture.
+ *  3) It works with a maximum width of 32-bits
+ */
```
mattklein123, line 27: This is probably a dumb question but isn't this https://en.cppreference.com/w/cpp/utility/bitset? If it's not can you add comments on how it's different?

mattklein123 (commented):
One quick question to start out with. Cool stuff!

source/common/common/bit_array.h
KBaichoo: From comment above: " BitArray is an fixed sized array of fixed width bits."

Note the "fixed width bits", entries can range from being 1-32 bits each. For bit set elements are just 1 bit width. From looking at get: https://en.cppreference.com/w/cpp/utility/bitset/operator_at we can see it just rets a bool

mattklein123: OK, got it. Pedantically I found this a bit confusing. Maybe something like "BitArray is a fixed sized array of fixed with bit sets" or something like that, or just a note to clarify how it's different from bitset. As an aside, I assume that in some cases `std::array<std::bitset>` or similar would not compress as well, but would that end up a simpler implementation? Anyway I would just have some note about bitset to make it more clear for quick readers.

KBaichoo: sg, added a note to clarify how it is different from `std::bitset`.

Yea the compression is __significantly worse__ in all cases except when using bit array for 32-bit elements.
```
std::bitset<1> foo;
std::cout << "Size of foo: " << sizeof(foo) << std::endl;
```

Yields 4, which would be 32x  more than BitArray representation. Moreover we'd need to instantiate templates for `bitset<1>, bitset<2>, ..., bitset<32>`. which seems like an easy way to bloat the binary.

mattklein123 (approved):
Nice. LGTM at a high level!

jpsim:
This looks neat @KBaichoo, although I think this caused a build failure for Envoy Mobile for some architectures:

```
source/common/common/bit_array.h:48:5: error: static_assert failed due to requirement 'sizeof(unsigned char *) == 8' "Pointer underlying size is not 8 bytes, are we running on a 64-bit architecture?"
    static_assert(
    ^
```

We build for the following architectures: `x86`, `x86_64`, `armeabi-v7a`, `arm64-v8a`

I suspect it's the `x86` that's failing based on the error message.

jmarantz:
Ah we had asserted that E-M did not every run on 32-bit. If that's not the case I think we can roll this back and then push forward with this PR, but with https://github.com/envoyproxy/envoy/pull/25149/commits/1df41955766a197df2eb41544d04f0131dfede2e backed out, and replaced with code that ifdefs based on bit-width

I added comments suggesting how we might do that, as it's not possible to us sizeof in an ifdef.

@KBaichoo WDYT?

jmarantz:
Comment with suggestions about how to ifdef for bit-size: https://github.com/envoyproxy/envoy/pull/25149/files/1df41955766a197df2eb41544d04f0131dfede2e#r1103639479

KBaichoo:
That sounds a reasonable approach @jmarantz . Unless there's a better way that EM approaches these e.g. I suspect there might be other parts in the codebase that originally assumed 64-bit?

@jpsim can you provide a bazel command so I could try to test this in the next iteration?

jpsim:
This should trigger it:

```
cd mobile
./bazelw build --fat_apk_cpu=x86_64 //:android_dist
```

jmarantz:
I would've thought x86_64 would have passed that static assert. Did you mean --fat_apk_cpu=x86 ?

jpsim:
Yes, sorry I typo'd. I meant `--fat_apk_cpu=x86`.

PR #25160, opened 2023-01-25
Oauth2 filter refactoring

Alexcei88:
Commit Message:

Refactoring oauth2 filter without modification of functionality. It's needed for preparing the filter to add refresh token supporting.

Additional Description:

Need for other PR https://github.com/envoyproxy/envoy/pull/24684

Risk Level:
Low

Testing:

Docs Changes:

Need to test oauth2 filter

Release Notes:

None. It's intermediate PR


source/extensions/filters/http/oauth2/filter.h
```diff
@@ -108,11 +108,14 @@ struct CookieNames {
...
         oauth_hmac_(oauth_hmac.empty() ? "OauthHMAC" : oauth_hmac),
-        oauth_expires_(oauth_expires.empty() ? "OauthExpires" : oauth_expires) {}
+        oauth_expires_(oauth_expires.empty() ? "OauthExpires" : oauth_expires),
+        id_token_("IdToken"), refresh_token_("RefreshToken") {}
 
   const std::string bearer_token_;
   const std::string oauth_hmac_;
   const std::string oauth_expires_;
+  const std::string id_token_;
+  const std::string refresh_token_;
```
snowp, line 118: Since id/refresh token are constants we end up paying for string copies unnecessarily every time we allocate this object.  How about something like `static constexpr absl::string_view IdToken = "IdToken"` on this class instead?

source/extensions/filters/http/oauth2/filter.cc
```diff
@@ -470,19 +459,38 @@ void OAuth2Filter::finishFlow() {
...
+  Http::ResponseHeaderMapPtr response_headers = fillResponseHeader(getEncodedToken());
+  response_headers->setStatus(std::to_string(enumToInt(Http::Code::Found)));
+  response_headers->setLocation(state_);
 
+  decoder_callbacks_->encodeHeaders(std::move(response_headers), true, REDIRECT_LOGGED_IN);
+  config_->stats().oauth_success_.inc();
+}
+
+Http::ResponseHeaderMapPtr
+OAuth2Filter::fillResponseHeader(const std::string& encoded_token) const {
```
snowp, line 487: Maybe it would be a bit clearer if this function was called `addResponseCookies` and it took a `&ResponseHeaderMap` that it populates with the cookies? That way we can keep the previous 
```
 Http::ResponseHeaderMapPtr response_headers{Http::createHeaderMap<Http::ResponseHeaderMapImpl>(
      {{Http::Headers::get().Status, std::to_string(enumToInt(Http::Code::Found))}})};
```
init of the header map and use this helper function to add all the cookies?

snowp (changes requested):
Thanks for splitting things up! A couple comments

lizan (approved):

wbpcode:
friendly ping @snowp 

I think maybe it would be better to give a final review by you?

snowp (approved):
Thanks!

PR #25168, opened 2023-01-25
mobile: running tsan common tests on Envoy changes

alyssawilk:
I'd mildly prefer .github/workflows/core.yml but that runs on macos-12 and AFIK those machines are constrained.


alyssawilk:
cc @RyanTheOptimist @jpsim

alyssawilk:
alternately we could move core.yml over to running on linux, and enable that for all builds

PR #25170, opened 2023-01-25
datadog: event scheduling portion of new tracing library

dgoffredo:

This revision introduces a new library dependency for Datadog tracing, and adds a component that will be used in the future when more of the new library is integrated into the Datadog tracing extension.

`EventScheduler` allows the new tracing library to schedule a recurring timer that fires every two seconds; this is when the tracing library will send its batched traces to the Datadog Agent.


moderation:
The dependency changes are in multiple PRs - https://github.com/envoyproxy/envoy/pull/25121/files#diff-a2ca502474bb8ec5f092176f7ee6d2f11fc395ebb00e1717d6fbad9b16ee8fd8

Is this intentional?

dgoffredo:
Yes, it's intentional. In retrospect, it might have been better to first add only the dependency changes, wait for those changes to be merged into `main`, and then propose the lowest level of new code.

I wanted instead to propose the first round of pull requests ASAP, and so duplicated the shared added dependency in all of those lowest level PRs.

I could create a dedicated dependency PR and rebase the others onto it, if you like.

wbpcode:
ping myself, sorry for the delay. 😞

wbpcode (commented):
Thanks for you contribution and some comments are added.

source/extensions/tracers/datadog/event_scheduler.cc
```diff
@@ -0,0 +1,62 @@
...
+EventScheduler::Cancel
+EventScheduler::schedule_recurring_event(std::chrono::steady_clock::duration interval,
+                                         std::function<void()> callback) {
+  // Yes, a shared pointer to a pointer.
+  auto self = std::make_shared<Event::Timer*>();
+
+  Event::TimerPtr timer = dispatcher_.createTimer([self, interval, callback = std::move(callback)] {
+    (**self).enableTimer(std::chrono::duration_cast<std::chrono::milliseconds>(interval));
+    callback();
+  });
```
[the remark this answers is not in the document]
dgoffredo: Good idea, will do.

source/extensions/tracers/datadog/event_scheduler.cc
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    timer->disableTimer();
+    auto found = std::find_if(timers_.begin(), timers_.end(),
+                              [&](const auto& element) { return element.get() == timer; });
+    assert(found != timers_.end());
+    timers_.erase(found);
+    timer = nullptr;
+  };
+}
```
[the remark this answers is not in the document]
dgoffredo: The contract of the function says that use of the cancellation function after this object is destroyed results in undefined behavior. Mean-spirited, perhaps, but I don't think that it can happen in any case.

What kind of guard might be appropriate?

My thinking is that the `assert` does the job of crashing during testing should we mistakenly break the contract in the future. I should probably replace it with one of Envoy's assert macros instead.

source/extensions/tracers/datadog/event_scheduler.h
```diff
@@ -0,0 +1,50 @@
...
+   * @return a zero-parameter function that cancels the recurring event
+   */
+  Cancel schedule_recurring_event(std::chrono::steady_clock::duration interval,
+                                  std::function<void()> callback) override;
+
+  nlohmann::json config_json() const override;
+
+private:
+  Event::Dispatcher& dispatcher_;
+  std::vector<Event::TimerPtr> timers_;
```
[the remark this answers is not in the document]
dgoffredo: This class will be storing at most one timer in real usage, but the interface we're implementing supports arbitrarily many. I figured a linear search was a good compromise.

source/extensions/tracers/datadog/event_scheduler.cc
```diff
@@ -0,0 +1,62 @@
+#include "source/extensions/tracers/datadog/event_scheduler.h"
+
+#include <cassert>
+#include <datadog/json.hpp>
```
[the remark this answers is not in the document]
dgoffredo: will do

dgoffredo:
Thanks for the review. I'll push a commit and merge from main within the next couple of days.

source/extensions/tracers/datadog/event_scheduler.h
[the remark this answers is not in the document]
wbpcode: ok.

source/extensions/tracers/datadog/event_scheduler.cc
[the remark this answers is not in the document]
wbpcode: Any possible runtime error or crash is dangerous for Envoy (as gateway/lb or sidecar) and we will try to ensure it will not happend.
But I have no much background knowledge about these scheduler, so may be I am wrong.

A guard could be a simple shared_pointer of bool. We can catch this pointer in every async callbck. And when the object is destroyed, we can set the value of bool to false. And the async callback run accutually logic only when the value of bool is still be true.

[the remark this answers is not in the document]
dgoffredo: That would work.

Then we'd have to test it, though, and we'd be testing a code path that "shall not" actually happen. I think that a `RELEASE_ASSERT` suffices here. Let me know if you disagree.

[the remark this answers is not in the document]
wbpcode: I am ok if you ensure that "that "shall not" actually happen". 😸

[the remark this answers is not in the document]
dgoffredo: I prefer to leave it as undefined behavior, though I appreciate your instinct of avoiding footguns.

dgoffredo:
> Code coverage for source/extensions/tracers/datadog is lower than limit of 96.6 (96.5)

Hmmm. Looks like calling `config_json` in a test might help: https://storage.googleapis.com/envoy-pr/0f305a0/coverage/source/extensions/tracers/datadog/event_scheduler.cc.gcov.html

Also interesting in the logger, which was recently merged, I'd except these kinds of tests: https://github.com/envoyproxy/envoy/pull/25025/files#diff-ab9f820e99ec92b52dd7334d825976ab5b5e2240dc9693e58fd4c99fab4c7dddR96-R103
to cover these kinds of branches: https://storage.googleapis.com/envoy-pr/9763eb9/coverage/source/extensions/tracers/datadog/logger.cc.gcov.html
but they don't.

I'll have to look closer.

wbpcode (dismissed):
LGTM. Thanks for you great contribution.

cc @jmarantz for final review or merge.

dgoffredo:
Ping @jmarantz

source/extensions/tracers/datadog/event_scheduler.cc
```diff
@@ -0,0 +1,83 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace Tracers {
+namespace Datadog {
+
+EventScheduler::EventScheduler(Event::Dispatcher& dispatcher) : dispatcher_(dispatcher) {}
+
+EventScheduler::~EventScheduler() {
+  for (const auto& timer : timers_) {
+    timer->disableTimer();
```
jmarantz, line 19: nit: this is not needed. Timers will be disabled on destruction.

source/extensions/tracers/datadog/event_scheduler.cc
```diff
@@ -0,0 +1,83 @@
...
+
+  timer_raw->enableTimer(std::chrono::duration_cast<std::chrono::milliseconds>(interval));
+
+  return [this, timer = timer_raw]() mutable {
+    if (!timer) {
+      return; // idempotent
+    }
+
+    timer->disableTimer();
+    auto found = std::find_if(timers_.begin(), timers_.end(),
```
jmarantz, line 65: does this need to be fast? Should we use an associative structure?

source/extensions/tracers/datadog/event_scheduler.h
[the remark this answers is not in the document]
jmarantz: probably a comment with that assumption would help future readers.

test/extensions/tracers/datadog/event_scheduler_test.cc
```diff
@@ -0,0 +1,111 @@
...
+TEST(DatadogEventSchedulerTest, ScheduleRecurringEventCallsCreatesATimer) {
+  testing::NiceMock<ThreadLocal::MockInstance> thread_local_storage_;
+
+  EventScheduler scheduler{thread_local_storage_.dispatcher_};
+  testing::MockFunction<void()> callback;
+  // The interval is arbitrary in these tests; we just have to be able to
+  // compare it to what was passed to the mocks.
+  // The only requirement is that it be divisible by milliseconds, because
+  // that's what `Timer::enableTimer` accepts.
+  const auto interval = std::chrono::milliseconds(2000);
```
jmarantz, line 27: could you write this as `const std::chrono::milliseconds interval(2000);` ?

test/extensions/tracers/datadog/event_scheduler_test.cc
```diff
@@ -0,0 +1,111 @@
...
+
+  EXPECT_CALL(thread_local_storage_.dispatcher_, createTimer_(testing::_));
+
+  (void)scheduler.schedule_recurring_event(interval, callback.AsStdFunction());
```
jmarantz, line 31: nit: in Envoy (for whatever reason) we do not use this syntax; we just drop the void. IMO it does convey useful info but that's not the official style.

test/extensions/tracers/datadog/event_scheduler_test.cc
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// This could be tested above, but introducing an `Event::MockTimer` disrupts
+// our ability to track calls to `MockDispatcher::createTimer_`. So, two
+// separate tests.
+TEST(DatadogEventSchedulerTest, ScheduleRecurringEventEnablesATimer) {
+  testing::NiceMock<ThreadLocal::MockInstance> thread_local_storage_;
+  auto* const timer_ = new testing::NiceMock<Event::MockTimer>(&thread_local_storage_.dispatcher_);
```
jmarantz, line 39: nit: don't use an underscore suffix for a local. Comment on memory management: who deletes timer?

test/extensions/tracers/datadog/event_scheduler_test.cc
```diff
@@ -0,0 +1,111 @@
...
+  const auto interval = std::chrono::milliseconds(2000);
+
+  EXPECT_CALL(*timer_, enableTimer(interval, testing::_));
+
+  (void)scheduler.schedule_recurring_event(interval, callback.AsStdFunction());
+}
+
+TEST(DatadogEventSchedulerTest, TriggeredTimerInvokesCallbackAndReschedulesItself) {
+  testing::NiceMock<ThreadLocal::MockInstance> thread_local_storage_;
+  auto* const timer_ = new testing::NiceMock<Event::MockTimer>(&thread_local_storage_.dispatcher_);
```
jmarantz, line 52: dito

jmarantz (commented):
Sorry for the delay this fell off my radar.

I found a few minor suggestions, but I'm not going to be available for the next few days. Whatever you decide is fine.

@wbpcode  feel free to make the final call and merge for this.

source/extensions/tracers/datadog/event_scheduler.cc
dgoffredo: I went digging to see whether that's true.

The interface's destructor doesn't mention it: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/envoy/event/timer.h#L30

This implementation's destructor is default, and so doesn't call `disableTimer` (more on this later): https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/timer_impl.h#L53

This implementation's destructor _does_ call `disableTimer`: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/scaled_range_timer_manager_impl.cc#L39

However, it looks like the scaled timer is distinct from "normal" timers in the dispatcher's implementation:
- https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/dispatcher_impl.cc#L67-L87
- https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/dispatcher_impl.cc#L213-L243

The code to follow seems to be: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/dispatcher_impl.cc#L237-L242

which is determined by the scheduler initialized here: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/dispatcher_impl.cc#L75

I'm guessing that scheduler ends up being the `RealScheduler`: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/real_time_system.cc#L16

In that case, it's the "base scheduler" that actually creates the timer. That's determined here: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/dispatcher_impl.h#L158

so the timer is created here: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/libevent_scheduler.cc#L38-L40

and that gives us one of these: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/timer_impl.h#L53

and its destructor does not call `disableTimer`.

However, looking at that implemention of `disableTimer`: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/timer_impl.cc#L30C17-L33

It deletes the underlying event. `TimerImpl` also derives from `ImplBase`: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/timer_impl.h#L53

`ImplBase` has a destructor that performs the same operation as `disableTimer`: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/source/common/event/event_impl_base.cc#L10

Aha!

Ok... my conclusion is: "If we assume `Event::Dispatcher` produces `Event::Timer` objects that are instances of `Event::TimerImpl`, then we may also assume that calling `disableTimer` is unnecessary on destruction, because an equivalent operation will be performed by one of the destructors of `Event::TimerImpl`."

Strictly speaking, the interface itself, `Event::Timer`, does not guarantee this behavior, and that is why I `disableTimer` in a loop in this destructor.

Do you still think it is best to remove this loop? I now agree that it's redundant in practice.

source/extensions/tracers/datadog/event_scheduler.cc
dgoffredo: Vector is fine for the reason discussed in comments on the header file. I'll add a code comment.

test/extensions/tracers/datadog/event_scheduler_test.cc
dgoffredo: This pattern is copied from here: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/test/extensions/tracers/datadog/datadog_tracer_impl_test.cc#L54

Which was copied from here: https://github.com/envoyproxy/envoy/blob/40d15a9c391c91b05353abdbef78407ff49972fa/test/extensions/tracers/zipkin/zipkin_tracer_impl_test.cc#L55

It occurs to me now that the raw pointer is probably a substitute for "optional" in the original code.

> who deletes timer?

Nobody, it's leaked, as in the copied-from code.

I'll see if it works the same with just a `testing::NiceMock<Event::MockTimer>` on the stack (no pointer). That'd be better.

dgoffredo: Ah, it's not for lack of optional. This is indeed the intended usage of `MockTimer`: https://github.com/envoyproxy/envoy/blob/eee9b7cf37545147e135a81624589de7465179e9/test/mocks/event/mocks.h#L183-L186

source/extensions/tracers/datadog/event_scheduler.cc
dgoffredo: I removed the loop, but still feel that it had its place.

Adding a comment to the `Timer` interface's destructor might be worth considering.

source/extensions/tracers/datadog/event_scheduler.cc
dgoffredo: I changed it to `absl::flat_hash_set`. Initially I avoided it because I thought that heterogeneous lookups wouldn't be supported without some boilerplate, but it turns out they work out of the box.

A flat hash set has very little overhead compared to a small vector, so we're good.

jmarantz (approved):

dgoffredo:
Thanks!

Another week has gone by and I still haven't produced the final pair of pull requests. Those are coming soon...

PR #25171, opened 2023-01-25
Use route name suffix to specify header only decoding.

shuoyang2016:
Context: we have a requirement to support http backends. But we still need to wave the information in the header (path & query params) to proto message in order to develop the features on top of proto message, for example, resource extraction for IAM authz; field scrubbing for audit logging.

Therefore we came up with this idea to perform "header-only" decoding so we can offer the above features without touching bodies or any response data.

We have a subsequent filter which will read the decoded header data in the filter state and will process accordingly.

This is an early draft to seek feedback. Let us know how you think about it. Thank you.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
+++ b/api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
@@ -288,4 +288,17 @@ message GrpcJsonTranscoder {
   //
   // If unset, the current stream buffer size is used.
   google.protobuf.UInt32Value max_response_body_size = 16 [(validate.rules).uint32 = {gt: 0}];
+
+  // If true, the transcoder will honor route naming conventions for the below
+  // suffix applied to the route.
+  //
+  // "_skip_transcoding": the transcoder will skip transcoding for the route.
+  // "_header_only_decoding": the transcoder will only decode at headers
+  //   and will ignore body, trailer and all encoding paths.
+  //
+  // This is a workaround when per_filter_config isn't available for the route
+  // config. For example, istio control plane.
+  // TODO: remove this feature after istio supports per filter config in their
+  // first-class route APIs.
+  bool use_route_naming_conventions = 17;
 }
--- a/source/extensions/filters/http/grpc_json_transcoder/BUILD
+++ b/source/extensions/filters/http/grpc_json_transcoder/BUILD
@@ -23,6 +23,7 @@ envoy_cc_library(
         "api_httpbody_protos",
     ],
     deps = [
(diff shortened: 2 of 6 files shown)


api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
```diff
@@ -288,4 +288,17 @@ message GrpcJsonTranscoder {
...
+  // If true, the transcoder will honor route naming conventions for the below
+  // suffix applied to the route.
+  //
+  // "_skip_transcoding": the transcoder will skip transcoding for the route.
+  // "_header_only_decoding": the transcoder will only decode at headers
+  //   and will ignore body, trailer and all encoding paths.
+  //
+  // This is a workaround when per_filter_config isn't available for the route
+  // config. For example, istio control plane.
+  // TODO: remove this feature after istio supports per filter config in their
```
adisuissa, line 301: If the idea is the field should be removed, then it should probably not be a field, as APIs in Envoy are essentially forever.

CC'ing @qiwzhang @lizan as extension owners who may have better ideas on how to achieve this.

adisuissa (commented):
Thanks.
Pinging code-owners to get a second opinion.

api/envoy/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto
lizan: Agreed this shouldn't be added if we expect them to be removed.

> We have a subsequent filter which will read the decoded header in the
filter state and will process accordingly.

If you have a subsequent filter, you can do header_only there by removing body and encoding paths.

shuoyang2016: > APIs in Envoy are essentially forever

I see. This seems a big rigid limitation. I've worked on this less. Is it possible to relax it a bit for example naming it experiment or comment with warning?

> If you have a subsequent filter, you can do header_only there by removing body and encoding paths.
Yeah. We are essentially having another filter embedding this with a few workarounds. That's why I am exploring options to bend the API which could avoid much more hacks in our code.

kyessenov: ExtensionWithMatcher might be the way forward IMHO. Have you tried that option? That works well with EnvoyFilter since it's independent of routes.

PR #25177, opened 2023-01-26
go filter extension: check the filter state while invoking C api from Go

doujiang24:
Commit Message:

To make the APIs robust, users could call the Go exposed API safely.

Additional Description:

Now, just panic while invoking C API in an unexpected stage,  will add go recover in the Goroutine entry function, in another PR.

Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/contrib/golang/filters/http/source/cgo.cc
+++ b/contrib/golang/filters/http/source/cgo.cc
@@ -22,105 +22,105 @@ absl::string_view copyGoString(void* str) {
 
 extern "C" {
 
-void envoyGoFilterHandlerWrapper(void* r, std::function<void(std::shared_ptr<Filter>&)> f) {
+int envoyGoFilterHandlerWrapper(void* r, std::function<int(std::shared_ptr<Filter>&)> f) {
   auto req = reinterpret_cast<httpRequestInternal*>(r);
   auto weakFilter = req->weakFilter();
   if (auto filter = weakFilter.lock()) {
-    f(filter);
+    return f(filter);
   }
+  return CAPIFilterIsGone;
 }
 
-void envoyGoFilterHttpContinue(void* r, int status) {
-  envoyGoFilterHandlerWrapper(r, [status](std::shared_ptr<Filter>& filter) {
-    filter->continueStatus(static_cast<GolangStatus>(status));
+int envoyGoFilterHttpContinue(void* r, int status) {
+  return envoyGoFilterHandlerWrapper(r, [status](std::shared_ptr<Filter>& filter) -> int {
+    return filter->continueStatus(static_cast<GolangStatus>(status));
   });
 }
 
-void envoyGoFilterHttpSendLocalReply(void* r, int response_code, void* body_text, void* headers,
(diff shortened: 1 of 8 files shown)


contrib/golang/filters/http/source/go/pkg/api/api.h
```diff
@@ -6,6 +6,13 @@
 extern "C" {
 #endif
 
+// This describes the return value of a C Api from Go.
```
wangfakang, line 9: Maybe it would be better to use enum type, and then use `CAPIStatus` instead of int type.
```
enum CAPIStatus {
……
}
```

contrib/golang/filters/http/source/cgo.cc
```diff
@@ -22,105 +22,105 @@ absl::string_view copyGoString(void* str) {
 
 extern "C" {
 
-void envoyGoFilterHandlerWrapper(void* r, std::function<void(std::shared_ptr<Filter>&)> f) {
+int envoyGoFilterHandlerWrapper(void* r, std::function<int(std::shared_ptr<Filter>&)> f) {
```
wangfakang, line 25: use `CAPIStatus` instead of int type. same elsewhere

contrib/golang/filters/http/source/processor_state.cc
```diff
@@ -37,6 +37,15 @@ void BufferList::clearAll() {
   queue_.clear();
 };
 
+bool BufferList::checkExisting(Buffer::Instance* data) {
+  for (auto it = queue_.begin(); it != queue_.end(); it = queue_.erase(it)) {
```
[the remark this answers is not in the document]
doujiang24: Oh, nice catch, it's a copy&paste bug, already fixed.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -574,95 +604,158 @@ void copyHeaderMapToGo(Http::HeaderMap& m, GoString* go_strs, char* go_buf) {
...
     ENVOY_LOG(debug, "golang filter has been destroyed");
-    return;
+    return CAPIStatus::CAPIFilterIsDestroy;
+  }
+  auto& state = getProcessorState();
+  if (!state.isProcessingInGo()) {
+    ENVOY_LOG(debug, "golang filter is not processing Go");
+    return CAPIStatus::CAPINotInGo;
+  }
+  if (headers_ == nullptr) {
```
wangfakang, line 636:
```suggestion
  if (headers_ == nullptr) {
     ENVOY_LOG(debug, "cgo api called at the invalid phase: {}", __func__);
```

wangfakang: nit, same elsewhere.

wangfakang (commented):
Nice patch, thanks. except for some nits, others LGTM.

wangfakang (approved):

ggreenway (approved):

Issue #24976, opened 2023-01-17
mobile: avoid YAML/JSON parsing step during Envoy initialization


Augustyniak:
**Stop configuring Envoy Engine with the use of a huge string config.**

As it is now Envoy engine builder on all platforms configures Envoy by creating a huge string config which is then passed to the engine where it's parsed by YAML and/or JSON parsers. This string is based off the content of config.cc file and it's manipulated by engine_builder 

This has a few negative consequences:

- YAML/JSON parsing steps are expensive - they can cause a big performance impact and there is often no way around them. 
- We often end up having to modify the config using string manipulation operations which cause further performance impact i.e. add tens/hundreds of virtual cluster configs requires a lot of string concatenation operations.
- Working with a huge string is error prone, no type (or any other really) checks that can help you to catch issues before you learn that the app just crashes on app launch

We noticed a big performance impact of string manipulation/parsing operations when adding around 700-900 virtual cluster configs to Envoy builder and starting the engine:

<img width="1268" alt="Screenshot 2023-01-03 at 2 48 13 PM copy" src="https://user-images.githubusercontent.com/4410346/212926140-8a1640af-d777-4b47-8869-cbbd9ed10477.png">

The flame chart was captured on Android platform. Eyeballing the attached flame chart that was used to capture the initialization of Envoy Engine we can see that parsing JSON/YAML accounted for around 3-40% of the whole Envoy engine initialization time.

Augustyniak:
just FYI @alyssawilk as you mentioned that you were looking into improving things for c++ builder.

Augustyniak:
@alyssawilk and @jpsim made a huge progress on reducing the amount of strings/JSON-YAML parsing that happens when Envoy engine is initialized. That being said, virtual clusters are still configured using strings (which end up being parsed by JSON/YAML parser). Keeping the issue open for now.

Augustyniak:
@alyssawilk you mentioned that you could help us to add support for adding support for initializing virtual clusters with the use of protobuf fields. In case you offer still stands this is the list of properties that we use when working with virtual clusters (otherwise, it's a good opportunity for me to document what we use and will eventually need to add protobuf support for :)): 

1. [config.route.v3.VirtualHost](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost) proto where we use `virtual_clusters` field.
[...]

PR #25188, opened 2023-01-26
mobile: cleaning up runWithTemplate APIs

alyssawilk:
Changing the java and kotlin APIs to by default, createYaml(), and optionally run with custom registrations and yaml.
This allows Lyft (and OSS CI) to retain the ability to run custom yaml, but changes the default to doing builder-only engines, in preparation for moving Java to use the C++ engine in bootstrap mode.

Risk Level: Medium (though some test tweaks may be necessary for Lyft)
Testing: CI
Docs Changes: n/a (runWithTemplate didn't show up in docs)
Release Notes: n/a (ditto)
https://github.com/envoyproxy/envoy/issues/24976

### DIFF
--- a/mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -11,6 +11,7 @@ import javax.annotation.Nullable;
 import io.envoyproxy.envoymobile.engine.types.EnvoyHTTPFilterFactory;
 import io.envoyproxy.envoymobile.engine.types.EnvoyStringAccessor;
 import io.envoyproxy.envoymobile.engine.types.EnvoyKeyValueStore;
+import io.envoyproxy.envoymobile.engine.JniLibrary;
 
 /* Typed configuration that may be used for starting Envoy. */
 public class EnvoyConfiguration {
@@ -133,6 +134,7 @@ public class EnvoyConfiguration {
       Map<String, EnvoyStringAccessor> stringAccessors,
       Map<String, EnvoyKeyValueStore> keyValueStores, List<String> statSinks,
       Boolean enableSkipDNSLookupForProxiedRequests, boolean enablePlatformCertificatesValidation) {
+    JniLibrary.load();
     this.adminInterfaceEnabled = adminInterfaceEnabled;
     this.grpcStatsDomain = grpcStatsDomain;
     this.connectTimeoutSeconds = connectTimeoutSeconds;
@@ -169,30 +171,20 @@ public class EnvoyConfiguration {
     this.enablePlatformCertificatesValidation = enablePlatformCertificatesValidation;
     this.enableSkipDNSLookupForProxiedRequests = enableSkipDNSLookupForProxiedRequests;
   }
-
   /**
-   * Resolves the provided configuration template using properties on this
-   * configuration.
+   * Creates configuration YAML based on the configuration of the class
    *
-   * @param configTemplate the template configuration to resolve.
-   * @param platformFilterTemplate helper template to build platform http filters.
-   * @param nativeFilterTemplate helper template to build native http filters.
-   * @param altProtocolCacheFilterInsert helper insert to include the alt protocol cache filter.
-   * @param gzipFilterInsert helper to include to enable gzip compression.
-   * @param brotliFilterInsert helper to include to enable brotli compression.
-   * @param socketTagFilterInsert helper to include to enable socket tagging.
-   * @param persistentDNSCacheConfigInsert helper to include to enable DNS cache.
-   * @param certValidationTemplate helper template to enable cert validation.
-   * @return String, the resolved template.
-   * @throws ConfigurationException, when the template provided is not fully
+   * @return String, the resolved yaml.
+   * @throws ConfigurationException, when the yaml provided is not fully
    *                                 resolved.
    */
-  String resolveTemplate(final String configTemplate, final String platformFilterTemplate,
-                         final String nativeFilterTemplate,
-                         final String altProtocolCacheFilterInsert, final String gzipFilterInsert,
-                         final String brotliFilterInsert, final String socketTagFilterInsert,
-                         final String persistentDNSCacheConfigInsert,
-                         final String certValidationTemplate) {
+  String createYaml() {
+    final String configTemplate = JniLibrary.configTemplate();
+    final String certValidationTemplate =
+        JniLibrary.certValidationTemplate(enablePlatformCertificatesValidation);
+    final String platformFilterTemplate = JniLibrary.platformFilterTemplate();
+    final String nativeFilterTemplate = JniLibrary.nativeFilterTemplate();
+
     final StringBuilder customFiltersBuilder = new StringBuilder();
 
     for (EnvoyHTTPFilterFactory filterFactory : httpPlatformFilterFactories) {
@@ -208,17 +200,21 @@ public class EnvoyConfiguration {
     }
 
     if (enableHttp3) {
+      final String altProtocolCacheFilterInsert = JniLibrary.altProtocolCacheFilterInsert();
       customFiltersBuilder.append(altProtocolCacheFilterInsert);
     }
 
     if (enableGzip) {
+      final String gzipFilterInsert = JniLibrary.gzipConfigInsert();
       customFiltersBuilder.append(gzipFilterInsert);
     }
 
     if (enableBrotli) {
+      final String brotliFilterInsert = JniLibrary.brotliConfigInsert();
       customFiltersBuilder.append(brotliFilterInsert);
     }
     if (enableSocketTagging) {
+      final String socketTagFilterInsert = JniLibrary.socketTagConfigInsert();
       customFiltersBuilder.append(socketTagFilterInsert);
     }
 
@@ -257,6 +253,7 @@ public class EnvoyConfiguration {
         .append("\n");
 
     if (enableDNSCache) {
+      final String persistentDNSCacheConfigInsert = JniLibrary.persistentDNSCacheConfigInsert();
       configBuilder.append(
           String.format("- &persistent_dns_cache_config %s\n", persistentDNSCacheConfigInsert));
     }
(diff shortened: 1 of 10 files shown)


mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngine.java
```diff
@@ -22,22 +22,31 @@ public interface EnvoyEngine {
...
+   *
+   * @param envoyConfiguration The EnvoyConfiguration used to start Envoy.
+   */
+  void performRegistration(EnvoyConfiguration envoyConfiguration);
+
   /**
    * Run the Envoy engine with the provided yaml string and log level.
    *
-   * The envoyConfiguration is used to resolve the configurationYAML.
+   * This does not perform registration, and performRegistration may need to be called first.
```
RyanTheOptimist, line 37: nit: consider changing performRegistration to `performRegistration()`

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -169,30 +171,20 @@ public EnvoyConfiguration(
...
    *
-   * @param configTemplate the template configuration to resolve.
-   * @param platformFilterTemplate helper template to build platform http filters.
-   * @param nativeFilterTemplate helper template to build native http filters.
-   * @param altProtocolCacheFilterInsert helper insert to include the alt protocol cache filter.
-   * @param gzipFilterInsert helper to include to enable gzip compression.
-   * @param brotliFilterInsert helper to include to enable brotli compression.
-   * @param socketTagFilterInsert helper to include to enable socket tagging.
-   * @param persistentDNSCacheConfigInsert helper to include to enable DNS cache.
-   * @param certValidationTemplate helper template to enable cert validation.
```
RyanTheOptimist, line 185: Yay! These arguments always felt a bit unnecessary as their values live in config.cc.

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -208,17 +200,21 @@ String resolveTemplate(final String configTemplate, final String platformFilterT
     }
 
     if (enableHttp3) {
+      final String altProtocolCacheFilterInsert = JniLibrary.altProtocolCacheFilterInsert();
```
RyanTheOptimist, line 203: Out of curiosity, did you consider eliminating the local variable and just doing: `customFiltersBuilder.append(JniLibrary.altProtocolCacheFilterInsert());`? I think that might be a bit shorter and simpler.

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -65,18 +65,14 @@ public String dumpStats() {
   }
 
   /**
-   * Run the Envoy engine with the provided yaml string and log level.
-   *
-   * The envoyConfiguration is used to resolve the configurationYAML.
+   * Performs various JNI registration prior to engine running.
    *
    * @param configurationYAML The configuration yaml with which to start Envoy.
```
RyanTheOptimist, line 70: nit: Can this @param and the logLevel @param now be removed.

mobile/library/kotlin/io/envoyproxy/envoymobile/mocks/MockEnvoyEngine.kt
```diff
@@ -13,9 +13,10 @@ import io.envoyproxy.envoymobile.engine.types.EnvoyStringAccessor
 internal class MockEnvoyEngine : EnvoyEngine {
   override fun runWithConfig(envoyConfiguration: EnvoyConfiguration?, logLevel: String?): Int = 0
 
-  override fun runWithTemplate(
+  override fun performRegistration(envoyConfiguration: EnvoyConfiguration) = Unit
```
RyanTheOptimist, line 16: I don't speak Kotlin, but I notice that this method has `= Unit` and the other two have `= 0`. Do they mean the same thing?

mobile/test/java/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
```diff
@@ -160,13 +108,13 @@ class EnvoyConfigurationTest {
     assertThat(resolvedTemplate).contains("&h2_connection_keepalive_timeout 333s")
 
     // H3
-    assertThat(resolvedTemplate).contains(APCF_INSERT);
+    assertThat(resolvedTemplate).contains("http3_protocol_options:");
```
RyanTheOptimist, line 111: I'm fine with this implementation, but we could also do `contains(JniLibrary.altProtocolCacheFilterInsert())` if we wanted to. I'm not sure if that's better (because it's checking for the full string and is resilient in the face of changes to that template string) or if it's worse (because it would hide bugs if we typo'd that template string).

mobile/test/java/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
```diff
@@ -194,64 +142,9 @@ class EnvoyConfigurationTest {
     assertThat(resolvedTemplate).contains("test_config")
 
     // Cert Validation
-    assertThat(resolvedTemplate).contains("custom_validator_config")
+    assertThat(resolvedTemplate).contains("trusted_ca:")
```
RyanTheOptimist, line 145: Out of curiosity, why does this change?

mobile/test/java/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
```diff
@@ -194,64 +142,9 @@ class EnvoyConfigurationTest {
...
 
     // Proxying
     assertThat(resolvedTemplate).contains("&skip_dns_lookup_for_proxied_requests false")
   }
-
-  @Test
-  fun `configuration resolves with alternate values`() {
```
RyanTheOptimist, line 204: I would think we would still want coverage for the generated YAML with alternate builder values. Is that coverage provided elsewhere?

jpsim:
> Love it! Do we need a similar PR for Swift as well?

In the community call, Alyssa said she'd tackle the Kotlin side and I offered to take on the Swift side. I can work on that shortly.

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -93,16 +89,20 @@ public int runWithTemplate(String configurationYAML, EnvoyConfiguration envoyCon
...
-                envoyConfiguration.enablePlatformCertificatesValidation)),
-        logLevel);
+  /**
+   * Run the Envoy engine with the provided yaml string and log level.
+   *
+   * This does not perform registration, and performRegistration may need to be called first.
+   *
+   * @param configurationYAML The configuration yaml with which to start Envoy.
+   * @param logLevel          The log level to use when starting Envoy.
+   * @return A status indicating if the action was successful.
```
jpsim, line 101: Not new to this PR, but it's ambiguous what status values indicate success. Is zero success like unix exit codes or is one success like boolean 0/false and 1/true?

mobile/library/kotlin/io/envoyproxy/envoymobile/mocks/MockEnvoyEngine.kt
jpsim: It's because `runWithConfig` returns an integer and `runWithConfig` returns "nothing" (aka `void` / `Void` / `Unit`).

mobile/test/java/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
jpsim: +1

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
RyanTheOptimist: I wonder if we should switch these ints over to the Java enum EnvoyStatus?

jpsim: Probably a good idea

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngine.java
alyssawilk: done

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
alyssawilk: yes thanks

mobile/test/java/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
alyssawilk: I'd prefer to leave as is because the end goal is to get rid of the template inserts entirely

mobile/test/java/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
alyssawilk: sorry was focused on the template vlaues not boolean

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
alyssawilk: true but again the end goal is to remove all of this and I'm trying to minimize diffs in between.  Given that, OK to leave as is?

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
jpsim: Yes, out of scope here.

jpsim (approved):

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
RyanTheOptimist: 100%

RyanTheOptimist (approved):

PR #25189, opened 2023-01-26
Change mobile binary size comparison workflow to use release build

caschoener:
Was debugging a binary size increase in https://github.com/envoyproxy/envoy/actions/runs/4008291380/workflow and noticed that the comparison is based on --config=remote-ci-linux-clang.  Talked with JP and it seems more appropriate to compare a release build instead.

Test plan: Look at CI

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -26,7 +26,7 @@ jobs:
       run: |
         cd mobile && ./bazelw build \
             --config=sizeopt \
-            --config=remote-ci-linux-clang \
+            --config=release-common \
             --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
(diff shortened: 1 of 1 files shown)


caschoener:
image
The number changed from 858 MB to 28.  This seems like a more accurate representation of our binary size though.  What do you think JP?

jpsim (approved):
I think this is appropriate. Thanks for the fixes!

jpsim:
> The number changed from 858 MB to 28.

This isn't terribly surprising. On iOS, where we statically link and deadstrip some code at link time, we see a ~5MB size increase from using Envoy Mobile to the compressed single-architecture iOS app.

caschoener:
image
Sweet

PR #25207, opened 2023-01-27
Unrevert streaming shadows

paul-r-gall:
Commit Message: Revert the reversion of shadow streaming.
Additional Description: Shadow streaming was implemented in #24365, and rolled back in #25184 due to a bad interaction with retry buffering. This fixes the retry issue.
Risk Level: See #24365 
Testing: Regression test for the issue added.


paul-r-gall:
@snowp can you review this?

jpsim (approved):
+1 I've confirmed this fixes the Envoy Mobile regressions we encountered in Lyft's iOS apps

paul-r-gall:
I believe that the failure is due to an actual bug in the way QUIC downstreams interact with filters that add bodies to header-only requests.

paul-r-gall:
Turns out I just didn't understand how QUIC works :)

alyssawilk (approved):
Looks like this largely boils down to https://github.com/envoyproxy/envoy/pull/25207/commits/86d5410f527a49a9aa672846aa4a4e39067be985 reverting an ordering change I'd missed.  Good find and thanks for the help reproducing, @jpsim !

snowp (approved):
Thanks!

PR #25210, opened 2023-01-27
Release assert `inotify_init1` output

davinci26:
With the order previously if `inotify_init1` failed then Envoy would crash in `createFileEvent` with `SOCKET_VALID(fd)` and the inotify specific assert would never hit.

I recently tried to debug this failure and I had to use `strace` to figure out how things are failing and that release assert could have saved me quite a bit of time.

Signed-off-by: Sotiris Nanopoulos <sotiris.nanopoulos@reddit.com>


davinci26:
PTAL @envoyproxy/first-pass-reviewers

ggreenway (approved):
LGTM. Putting assert before use of the value makes sense. Thanks!

Issue #25178, opened 2023-01-26
Go HTTP Filter: fails with cgo argument has Go pointer to Go pointer


mhofstetter:
*Title*: *Go HTTP Filter: fails with cgo argument has Go pointer to Go pointer*

*Description*:
CGO calls from a [Go HTTP Filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/golang_filter) via C-API implementation are failing with `cgo argument has Go pointer to Go pointer` if string or slice arguments are directy passed to C via `unsafe.Pointer`. Due to the [cgo pointer check](https://github.com/golang/go/blob/master/src/runtime/cgocall.go#L393) which panics if Go pointers are detected - e.g. strings which have been created at runtime and aren't string literals. 

An example is the [C-API implementation](https://github.com/envoyproxy/envoy/blob/main/contrib/golang/filters/http/source/go/pkg/http/capi_impl.go#L59) of `SendLocalReply` which fails when using it with a `body` which is created at runtime like in the [Go HTTP filter example](https://github.com/envoyproxy/envoy/blob/main/contrib/golang/filters/http/test/test_data/echo/filter.go#L16-L17). The same error would occur for the string-slice `strs` if the headers map would contain an entry. 

I would expect, that all string & slice arguments in the C-API implementation are safe for the pointer check, by using `reflect.StringHeader` & `reflect.SliceHeader` (and passing `data` property as `unsafe.Pointer`) like it's done in other cases.

* `HttpSendLocalReply` -> `body_text`, `headers`, `details`
* `HttpGetHeader` -> `key`
* `HttpSetHeader` -> `key`, `value`
* `HttpRemoveHeader` -> `key`
* ...

doujiang24:
@mhofstetter I don't think it's a bug, go HTTP filter is desired to work with the `GODEBUG=cgocheck=0` ENV variable, which disables the cgo pointer check.

cgo pointer check is used for normal cgo usage, those people don't know the cgo implementation deeply.

The go HTTP filter relies on cgo heavily, and we disable cgo pointer check for better performance and easier usage.
In exchange, we have to make sure cgo pointer safety manually.

I do think we could make sure those cgo pointers safety.

mhofstetter:
@doujiang24 thanks for the clarification.

[...]

PR #25222, opened 2023-01-28
docs: Golang filter requires disabling cgocheck.

doujiang24:
Partial fix for https://github.com/envoyproxy/envoy/issues/25178

### DIFF
--- a/docs/root/configuration/http/http_filters/golang_filter.rst
+++ b/docs/root/configuration/http/http_filters/golang_filter.rst
@@ -13,7 +13,7 @@ See the `Envoy's Golang extension proposal documentation
 for more details on the filter's implementation.
 
 .. attention::
-  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable,
+  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable, when running envoy with Go plugin enabled,
   which disables the cgo pointer check.
 
 Developing a Go plugin
--- a/docs/root/configuration/http/http_filters/golang_filter.rst
+++ b/docs/root/configuration/http/http_filters/golang_filter.rst
@@ -12,6 +12,9 @@ See the `Envoy's Golang extension proposal documentation
 <https://docs.google.com/document/d/1noApyS0IfmOGmEOHdWk2-BOp0V37zgXMM4MdByr1lQk/edit?usp=sharing>`_
 for more details on the filter's implementation.
 
+.. attention::
+  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable,
+  which disables the cgo pointer check.
 
 Developing a Go plugin
 ----------------------
--- a/docs/root/configuration/http/http_filters/golang_filter.rst
+++ b/docs/root/configuration/http/http_filters/golang_filter.rst
@@ -13,8 +13,9 @@ See the `Envoy's Golang extension proposal documentation
 for more details on the filter's implementation.
 
 .. attention::
-  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable, when running envoy with Go plugin enabled,
-  which disables the cgo pointer check.
+  The Envoy Golang filter is designed to be run with the `GODEBUG=cgocheck=0` environment variable set.
+
+  This disables the cgo pointer check.
 
 Developing a Go plugin
 ----------------------
--- a/docs/root/configuration/http/http_filters/golang_filter.rst
+++ b/docs/root/configuration/http/http_filters/golang_filter.rst
@@ -12,10 +12,12 @@ See the `Envoy's Golang extension proposal documentation
 <https://docs.google.com/document/d/1noApyS0IfmOGmEOHdWk2-BOp0V37zgXMM4MdByr1lQk/edit?usp=sharing>`_
 for more details on the filter's implementation.
 
-.. attention::
+.. warning::
   The Envoy Golang filter is designed to be run with the `GODEBUG=cgocheck=0` environment variable set.
 
   This disables the cgo pointer check.
+
+  Failure to set this environment variable will cause Envoy to crash!
 
 Developing a Go plugin
 ----------------------


docs/root/configuration/http/http_filters/golang_filter.rst
```diff
@@ -13,7 +13,7 @@ See the `Envoy's Golang extension proposal documentation
 for more details on the filter's implementation.
 
 .. attention::
-  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable,
+  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable, when running envoy with Go plugin enabled,
```
phlax, line 16:
```suggestion
  The Envoy Golang filter is designed to work with the `GODEBUG=cgocheck=0` environment variable
```

Im not totally clear if this relates to building, running or both

phlax: i guess my question then is where should this be set - i can see it in the tests but not anywhere else

does this mean that an end-user _must_ do:

```console
$ export GODEBUG=cgocheck=0
$ ./bin/envoy -c path-to-config-with-go-filter.yaml
```

doujiang24: > does this mean that an end-user _must_ do:
> 
> ```
> $ export GODEBUG=cgocheck=0
> $ ./bin/envoy -c path-to-config-with-go-filter.yaml
> ```

yep, that's expected.

docs/root/configuration/http/http_filters/golang_filter.rst
```diff
@@ -13,8 +13,8 @@ See the `Envoy's Golang extension proposal documentation
 for more details on the filter's implementation.
 
 .. attention::
-  It is designed to work with the `GODEBUG=cgocheck=0` ENV variable, when running envoy with Go plugin enabled,
-  which disables the cgo pointer check.
+  The Envoy Golang filter is designed to be run with the `GODEBUG=cgocheck=0` environment variable set,
+  when running Envoy with Golang filter enabled, which disables the cgo pointer check.
```
phlax, line 17:
```suggestion
  The Envoy Golang filter is designed to be run with the `GODEBUG=cgocheck=0` environment variable set.
  
  This disables the cgo pointer check.
```

the sentence doesnt quite read correctly

docs/root/configuration/http/http_filters/golang_filter.rst
phlax: hmm - that seems less than ideal

can we not do this in code?

what happens if the end-user forgets/omits to do this?

doujiang24: > can we not do this in code?

Yep, do it in code should be better. It's already on our TODO list.

> what happens if the end-user forgets/omits to do this?

Go will panic, envoy will crash then.

docs/root/configuration/http/http_filters/golang_filter.rst
doujiang24: Thanks, fixed it with a force push, due to miss DSO in the previous commit.

docs/root/configuration/http/http_filters/golang_filter.rst
```diff
@@ -12,6 +12,10 @@ See the `Envoy's Golang extension proposal documentation
 <https://docs.google.com/document/d/1noApyS0IfmOGmEOHdWk2-BOp0V37zgXMM4MdByr1lQk/edit?usp=sharing>`_
 for more details on the filter's implementation.
 
+.. attention::
```
phlax, line 15:
```suggestion
.. warning::
```

docs/root/configuration/http/http_filters/golang_filter.rst
```diff
@@ -12,6 +12,10 @@ See the `Envoy's Golang extension proposal documentation
...
+  The Envoy Golang filter is designed to be run with the `GODEBUG=cgocheck=0` environment variable set.
+
+  This disables the cgo pointer check.
```
phlax, line 18:
```suggestion
  This disables the cgo pointer check.
  
  Failure to set this environment variable will cause Envoy to crash!
```

docs/root/configuration/http/http_filters/golang_filter.rst
phlax: > Yep, do it in code should be better. It's already on our TODO list.

hmm, well - given that it can crash we defo need to do something - happy to land this until it is fixed

im wondering tho - would it not be easier to fix than document the workaround ?

phlax: i guess the other ~related question (which is less of a priority) is what if a user _does_ want to debug with `cgocheck=1` (excuse my limited go knowledge if this is a stupid question)

doujiang24: > im wondering tho - would it not be easier to fix than document the workaround ?

There may be other items in the `GODEBUG` env variable, i.e. `GODEBUG=gctrace=1,invalidptr=1,cgocheck=0,madvdontneed=1`, and we only overwriting the `cgocheck` item.
So, it is a bit complicated, but, yep, it's not a big deal.

I personally prefer doc first, and we still need a doc for hardcoding this in code when we do it in code.
It's also ok for me, if you think doing it in code first could be better.

> what if a user does want to debug with `cgocheck=1`

Oh, I think `cgocheck=1` is not for debugging.
In short, `cgocheck` checks we could not pass a struct with Go pointer inside, to C side, so that C could not touch the Go pointer - to make Go memory safe, could not be modified by C unexpectedly.
It's a higher level of protection for people who do not understand the cgo details.

Go filter extension is a cgo heavily use case, so we disable it, use low level of protection instead - as discussed here: https://github.com/envoyproxy/envoy/issues/25178#issuecomment-1404995919

Note: users who use the Go APIs from the Go filter extension, do not need to care about the `cgocheck`, Go APIs are always memory safety.

phlax: thanks for the explanation re GODEBUG - i guess updating requires parsing what is there already

> I personally prefer doc first, and we still need a doc for hardcoding this in code when we do it in code.

i am definitely happy getting docs with code - but this is not a simple documentation issue

The current default (ie GODEBUG _not_ set) causes Envoy to crash - this is one of the worst type of bugs - addressing that is a huge priority

Documenting broken software is generally a bad idea if fixing it is more/less trivial

>  and we still need a doc for hardcoding this in code when we do it in code.

sure, but hopefully it wont have to warn anyone about anything crashing

phlax (approved):
lgtm, thanks @doujiang24

PR #25227, opened 2023-01-28
fix: go plugin config parse

wangfakang:
Commit Message: fix the wrong ASSERT.
Additional Description: fix: go plugin config parse
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


wangfakang:
dependency internal Server Error.

/retest

contrib/golang/filters/http/test/golang_integration_test.cc
```diff
@@ -76,7 +75,7 @@ TEST_P(GolangIntegrationTest, Echo) {
   EXPECT_EQ("403", response->headers().getStatusValue());
 
   // check body for echo
-  auto body = StringUtil::toUpper(absl::StrFormat("forbidden from go, path: %s\r\n", path));
+  auto body = StringUtil::toUpper(absl::StrFormat("%s, path: %s\r\n", echo_body, path));
   EXPECT_EQ(body, StringUtil::toUpper(response->body()));
```
doujiang24, line 79: `StringUtil::toUpper` is not necessary.

doujiang24 (commented):
one comment, otherwise, LGTM.

contrib/golang/filters/http/test/golang_integration_test.cc
wangfakang: Yes, it could delete now. Previously, headers were case insensitive.

phlax:
a very high-level comment - if you make the response configurable - ie it doesnt echo the request - then im wondering if it should still be called echo

doujiang24 (approved):

wangfakang:
> a very high-level comment - if you make the response configurable - ie it doesnt echo the request - then im wondering if it should still be called echo

Thank you for your comments.
In the current configuration, the config's `body` is a part of the response body. So there will be a echo body anyway.

ggreenway (approved):

PR #25234, opened 2023-01-29
Host override in SNI dynamic forward proxy

### LINKED ISSUE #25093: Dynamically decide upstream host in network filters
In my setup I have two network filters in my filter chain:

- Custom network filter that in onNewConnection() has some logic to resolve a variable, say "dst"
- TcpProxy filter

I have an upstream that exposes many endpoints, for each option of the variable "dst". For example:

- x1.host.com
- x2.host.com

Is there currently an option in Envoy to dynamically select between x1/x2 based on the resolution of "dst"? Meaning that if "dst" is resolved to "x1" then the tcp proxy will be routed to x1.host.com, similarly if "dst" is resolved to "x2" then the routing will be to x2.host.com

Thanks for helping!

ohadvano:
Adds an ability to dynamically set host value for the SNI DFP filter, by setting a filter state value for a per-connection stream.

Please read following issue for an example use case:
Resolves #25093


ohadvano:
@lizan I can see that you created the PR to add SNI dynamic forward proxy. Can you please review this as well?

mattklein123:
Thanks for the contribution. I don't think we want to copy/paste the SNI filter into this new filter. Either we should just add this functionality to the existing SNI dynamic forward proxy filter (albeit it would then be poorly named) or factor out most of the code into a shared base class.

I would opt for the former just for simplicity even if the naming is a bit ugly, as long we we document it well. WDYT?

/wait

ohadvano:
@mattklein123, I stated in the PR description that currently it is missing: "Extracting common code from this filter and SNI dynamic forward proxy to a base class".

I am fully aligned on not copy-pasting that code. I created the PR before doing this so I can get feedback about intention of adding this kind of filter. You actually hit the point I was wondering about - whether we can spare this filter if we modify the SNI dynamic forward proxy filter to ignore the SNI field and take the host value from the filter state. 

In my opinion - if we add an option to override the host origin in the SNI filter, then it completely changes the meaning of that filter causing a filter that seemingly related to SNI, to be not related to SNI at all, which makes me wonder about user experience and confusion. I would have preferred this way if some of the functionality of that filter would still be related to SNI after applying that flag. That's why even though most of the functionality is similar, I would prefer separating it to another filter.
Furthermore, an adjustment to the dynamic forward proxy cluster is still needed. 
WDYT?

mattklein123:
cc @kyessenov for thoughts on whether we should do a new filter or add this to the existing SNI DFP filter.

I personally would probably just add it to the SNI DFP filter and possibly rename (at least in the docs) to something like "layer 4 DFP filter" and have it operate on both SNI and filter state.

I don't feel super strongly about it though. Let's see what @kyessenov thinks.

kyessenov:
Yeah, I agree with @mattklein123 . If we rename it in the code and the docs, then the only remaining bit with SNI is the proto package which cannot be changed. That's not so terrible, and having less code is better.

ohadvano:
@kyessenov is it legitimate to change the SNI filter name? That will cause a breaking change for many consumers. Also what did you mean by saying that the proto package cannot be changed? How so you suggest to handle this if it can't be changed?

kyessenov:
@ohadvano No we cannot change anything in protobufs. But we can change the internal names (C++ directories, etc). Users only see the protobuf name (in xDS)  and the extension name (in extension build manifest), so those two have to stay the same.

ohadvano:
@kyessenov so does it mean that practically we must keep calling that filter SNI dynamic forward proxy in the protos and the extension names and we can only modify the docs to specify that the behavior of the filter can be pivoted to be used as a "filter state" dynamic forward proxy instead of SNI?

kyessenov:
Yes, but this should be an addition to the extension to override SNI, essentially.  There are many cases where a custom filter state changes how the extension behave (route, original_dst), so I don't think this is special. The docs should clearly spell out how the decision is made (e.g. filter state takes priority), and the name should be more general (e.g. L4 DFP) to make it easier to discover.

ohadvano:
I see, so to conclude and make sure I got this right, the name of the filter in code, the name of the extension and proto definitions stay with SNI terminology, the docs terminology however changes to L4 dynamix forward proxy and states that it is used with the SNI dynamic forward proxy filter, while mentioning that the behavior can be overrided using additional boolean config to that filter. Have I got it right @kyessenov?
Would this flag also need to be applied to the dynamic forward proxy cluster? Or would it by default look for that filter state value and grant it precendece if it exists?

lizan:
+1 to add this to existing SNI DFP filter, with potential name change.

api/envoy/extensions/filters/network/filter_state_dynamic_forward_proxy/v3/filter_state_dynamic_forward_proxy.proto
```diff
@@ -0,0 +1,30 @@
...
+  // The DNS cache configuration that the filter will attach to. Note this
+  // configuration must match that of associated :ref:`dynamic forward proxy
+  // cluster configuration
+  // <envoy_v3_api_field_extensions.clusters.dynamic_forward_proxy.v3.ClusterConfig.dns_cache_config>`.
+  common.dynamic_forward_proxy.v3.DnsCacheConfig dns_cache_config = 1
+      [(validate.rules).message = {required: true}];
+
+  oneof port_specifier {
+    // The port number to connect to the upstream.
+    uint32 port_value = 2 [(validate.rules).uint32 = {lte: 65535 gt: 0}];
```
lizan, line 28: we have `port_value` because SNI doesn't contain port. You might want an option to retrieve the port from filter state as well.

ohadvano:
@mattklein123, @kyessenov, @lizan  I have updates the PR according to our conversation. Could you please review again?

source/common/upstream/dynamic_host_filter_state.h
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Envoy {
+namespace Upstream {
+
+class DynamicHostFilterState : public StreamInfo::FilterState::Object {
+public:
+  DynamicHostFilterState(absl::string_view host) : host_(host) {}
+  const std::string& value() const { return host_; }
+  static const std::string& key() {
+    CONSTRUCT_ON_FIRST_USE(std::string, "envoy.clusters.dynamic_host_filter_state");
```
lizan, line 13: `envoy.upstream.dynamic_host`?

ohadvano: Done

source/common/upstream/dynamic_host_filter_state.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "envoy/stream_info/filter_state.h"
+
+namespace Envoy {
+namespace Upstream {
+
+class DynamicHostFilterState : public StreamInfo::FilterState::Object {
```
lizan, line 8: On a second thought, shall we just use StringAccessor instead of a customized filter state here as it just stores a string? That give the setter more flexibility (e.g. via wasm).

ohadvano: Can you point to an example of this StringAccessor? I'm not familiar with what you are referring. In any way, I followed the same pattern as other filters are using to add custom behavior, such as the TCP proxy that uses filter state to dynamically select the upstream cluster. I believe we should follow the same pattern

mattklein123:
Generally this makes sense to me. I will pass the review over to @lizan and @kyessenov as they have already commented.

source/common/upstream/dynamic_host_filter_state.h
lizan: StringAccessor is [here](https://github.com/envoyproxy/envoy/blob/main/envoy/router/string_accessor.h). It is still the same pattern but not define a customized object to store the data.

lizan: You should still use the key name `envoy.upstream.dynamic_host` but the type of filter object should be a StringAccessor and I don't see why it needs to be a custom type. A string accessor is easier to get/set by other filters and for [logging and header substitution](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-filter-state)

ohadvano: Done

lizan (approved):

kzorfy:
@ohadvano Curious to know whether you considered enabling this capability of dynamically selecting SNI value from dynamic metadata in addition to filter state

mandarjog:
@ohadvano @kyessenov @mattklein123  anyone using this in production ? The stated maturity level is still "alpha/not ready for production". We have a really good use case for this filter. Potentially 50k clients connecting to 1K SNIs. 
cc @lizan 
Any more info is appreciated.

kyessenov:
@rshriram and @soulxu are listed as owners for sni_dynamic_forward_proxy. I haven't used this extension in production myself.

mandarjog:
@soulxu do you use sni_dynamic_forward_proxy in production ? @lizan do you know anyone ?

PR #25235, opened 2023-01-29
lua: new option of http call to skip sending xff

### LINKED ISSUE #25190: Don't set x-forwarded-for header in Lua request handler httpCall
*Title*: Don't set x-forwarded-for header in Lua request handler httpCall

*Description*:

I'm trying to use the [Lua httpCall function](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#httpcall) to contact the GCP Metadata server and retrieve a service account token, which I can then use to interact with some GCP services mid-request.
The problem is that the `httpCall` function adds an X-Forwarded-For header to any requests it makes, which the [GCP Metadata server explicitly rejects](https://cloud.google.com/compute/docs/metadata/overview#limitations).

Is there any way I can prevent the X-Forwarded-For header from being injected into the `htppCall`, or at least remove it before the requests is sent to the GCP Metadata server?

wbpcode:
Commit Message: lua: new option of http call to skip sending xff
Additional Description:

To close #25190

Risk Level: low.
Testing: unit.
Docs Changes: added.
Release Notes: added.

### DIFF
--- a/changelogs/current.yaml
+++ b/changelogs/current.yaml
@@ -66,5 +66,8 @@ new_features:
 - area: router
   change: |
     support route info in upstream access log.
+- area: lua
+  change: |
+    added an new option to the options of lua ``httpCall``. This allows to skip adding ``x-forwarded-for`` by setting ``{["send_xff"] = false}`` as the ``options``.
 
 deprecated:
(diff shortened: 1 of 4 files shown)


mattklein123 (approved):
LGTM, thanks!

PR #25241, opened 2023-01-30
dev: add ENVOY_DOCKER_SKIP_PULL to skip docker pull

spacewander:
Signed-off-by: spacewander <spacewanderlzx@gmail.com>

Commit Message: dev: add ENVOY_DOCKER_SKIP_PULL to skip docker pull
Additional Description: it  saves lots of time when developing with run_envoy_docker.sh wrapper
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -70,10 +70,10 @@ Besides `http_proxy` and `https_proxy`, maybe you need to set `go_proxy` to repl
 IMAGE_NAME=envoyproxy/envoy-build-ubuntu go_proxy=https://goproxy.cn,direct http_proxy=http://proxy.foo.com:8080 https_proxy=http://proxy.bar.com:8080 ./ci/run_envoy_docker.sh <build_script_args>
 ```
 
-The script will always pull the docker image. If you want to skip this action, you can specify `ENVOY_DOCKER_SKIP_PULL=true`.
+To force the Envoy build image to be refreshed by Docker you can set `ENVOY_DOCKER_PULL=true`.
 
 ```bash
-ENVOY_DOCKER_SKIP_PULL=true ./ci/run_envoy_docker.sh <build_script_args>
+ENVOY_DOCKER_PULL=true ./ci/run_envoy_docker.sh <build_script_args>
 ```
 
 ## On Linux
--- a/ci/README.md
+++ b/ci/README.md
@@ -70,6 +70,12 @@ Besides `http_proxy` and `https_proxy`, maybe you need to set `go_proxy` to repl
 IMAGE_NAME=envoyproxy/envoy-build-ubuntu go_proxy=https://goproxy.cn,direct http_proxy=http://proxy.foo.com:8080 https_proxy=http://proxy.bar.com:8080 ./ci/run_envoy_docker.sh <build_script_args>
 ```
 
+The script will always pull the docker image. If you want to skip this action, you can specify `ENVOY_DOCKER_SKIP_PULL=true`.
+
+```bash
+ENVOY_DOCKER_SKIP_PULL=true ./ci/run_envoy_docker.sh <build_script_args>
+```
+
 ## On Linux
 
 An example basic invocation to build a developer version of the Envoy static binary (using the Bazel `fastbuild` type) is:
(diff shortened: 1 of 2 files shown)


ci/run_envoy_docker.sh
```diff
@@ -80,7 +80,9 @@ if ! is_windows; then
     VOLUMES+=(-v "${SHARED_TMP_DIR}":"${SHARED_TMP_DIR}")
 fi
 
-time docker pull "${ENVOY_BUILD_IMAGE}"
+if [[ -z "${ENVOY_DOCKER_SKIP_PULL}" ]]; then
+    time docker pull "${ENVOY_BUILD_IMAGE}"
```
phlax, line 84: we could probably get rid of the `time` here also now i think

phlax (commented):
happy to land this - separating the pull part was mostly done so we could see how much CI time was being used to pull the image

im wondering if we should make it the default to not pull separately

ci/run_envoy_docker.sh
phlax: altho - if we make it the non-default then perhaps it would be useful to keep

spacewander:
> happy to land this - separating the pull part was mostly done so we could see how much CI time was being used to pull the image
> 
> im wondering if we should make it the default to not pull separately

If we make it the default, we need to update CI to always pull the latest one, otherwise a stale image may be used.

phlax:
> ... we need to update CI to always pull the latest one ...

i think its ok - we dont need to worry about that so much as the build image uses a "unique" hash for its tag (of course it could theoretically change but wont unless something has gone wrong)

spacewander:
Should I update the PR to make the docker pull non-default now, or should we wait for more opinions from the other people?

ci/README.md
```diff
@@ -70,6 +70,12 @@ Besides `http_proxy` and `https_proxy`, maybe you need to set `go_proxy` to repl
 IMAGE_NAME=envoyproxy/envoy-build-ubuntu go_proxy=https://goproxy.cn,direct http_proxy=http://proxy.foo.com:8080 https_proxy=http://proxy.bar.com:8080 ./ci/run_envoy_docker.sh <build_script_args>
 ```
 
+The script will always pull the docker image. If you want to skip this action, you can specify `ENVOY_DOCKER_SKIP_PULL=true`.
+
+```bash
+ENVOY_DOCKER_SKIP_PULL=true ./ci/run_envoy_docker.sh <build_script_args>
```
phlax, line 76:
```suggestion
ENVOY_DOCKER_PULL=true ./ci/run_envoy_docker.sh <build_script_args>
```

ci/run_envoy_docker.sh
```diff
@@ -80,7 +80,9 @@ if ! is_windows; then
     VOLUMES+=(-v "${SHARED_TMP_DIR}":"${SHARED_TMP_DIR}")
 fi
 
-time docker pull "${ENVOY_BUILD_IMAGE}"
+if [[ -z "${ENVOY_DOCKER_SKIP_PULL}" ]]; then
```
phlax, line 83:
```suggestion
if [[ -n "${ENVOY_DOCKER_PULL}" ]]; then
```

phlax (commented):
> should we wait for more opinions from the other people?

if you update, im happy to land

ci/README.md
```diff
@@ -70,6 +70,12 @@ Besides `http_proxy` and `https_proxy`, maybe you need to set `go_proxy` to repl
 IMAGE_NAME=envoyproxy/envoy-build-ubuntu go_proxy=https://goproxy.cn,direct http_proxy=http://proxy.foo.com:8080 https_proxy=http://proxy.bar.com:8080 ./ci/run_envoy_docker.sh <build_script_args>
 ```
 
+The script will always pull the docker image. If you want to skip this action, you can specify `ENVOY_DOCKER_SKIP_PULL=true`.
```
phlax, line 73:
```suggestion
To force the Envoy build image to be refreshed by Docker you can set `ENVOY_DOCKER_PULL=true`.
```

phlax (approved):
great! thanks for working on this @spacewander , lgtm

im thinking about following up with something to quieten the docker pull that happens when it runs in CI

PR #25249, opened 2023-01-30
[balsa] Increase header size limit in a DownstreamProtocolIntegrationTest.

bencebeky:

Increase header size limit in DownstreamProtocolIntegrationTest.ManyTrailerHeaders.

This is because BalsaParser includes every byte of headers (and trailers) in the size that it compares to the limit.  This includes ": " separator between key and header, and "\r\n" after each header line.  In contrast, http-parser does not have built-in size limit enforcement, and limit is incremented in ConnectionImpl using HeaderMapImpl::byteSize(), which only considers key and value lengths.

Commit Message: [balsa] Increase header size limit in a DownstreamProtocolIntegrationTest.
Additional Description:
Docs Changes: n/a
Release Notes: n/a


bencebeky:
Dianna and Biren: PTAL.  Thanks!

test/integration/protocol_integration_test.cc
```diff
@@ -2230,12 +2230,7 @@ TEST_P(DownstreamProtocolIntegrationTest, LargeRequestTrailersRejected) {
 // This test uses an Http::HeaderMapImpl instead of an Http::TestHeaderMapImpl to avoid
 // time-consuming byte size verification that will cause this test to timeout.
 TEST_P(DownstreamProtocolIntegrationTest, ManyTrailerHeaders) {
-  if (GetParam().http1_implementation == Http1ParserImpl::BalsaParser) {
-    // TODO(#21245): Re-enable this test for BalsaParser.
-    return;
-  }
-
-  setMaxRequestHeadersKb(96);
+  setMaxRequestHeadersKb(200);
```
diannahu, line 2233: I'm supportive of this change (and thank you for the explanation in the description!). Maybe alright to document this slight behavior diff somewhere? I'd be surprised if it caused any practical difference in the wild, but having some list of "known difference" quirks could be helpful just in case.

diannahu (approved):
Nice!

test/integration/protocol_integration_test.cc
bencebeky: Great idea!  I created a doc and shared it privately.

KBaichoo (approved):
LGTM, thanks for adding a doc capturing these differences. 

I think as we get closer to this landing we should consider adjusting defaults to give more headroom so that we don't start to reject many requests that we'd currently accept. 

We could even have these as distinct values since they're representing different things under balsa.

Issue #25264, opened 2023-01-31
Test flake: QuicHttpIntegrationTests/QuicHttpIntegrationTest.DeferredLogging/IPv6


ggreenway:
```
[ RUN      ] QuicHttpIntegrationTests/QuicHttpIntegrationTest.DeferredLogging/IPv6
test/integration/quic_http_integration_test.cc:1206: Failure
Expected: (std::stoi(metrics.at(1))) > (0), actual: 0 vs 0
```

This was added a few days ago in #23648.

cc @pksohn @RyanTheOptimist

ggreenway:
Failing run: https://dev.azure.com/cncf/envoy/_build/results?buildId=126681&view=logs&j=767be981-567e-57d8-68c3-2140ede0a0bd&t=1fb0b72f-6293-562d-a4d1-bb649223eaca

pksohn:
I'll open a PR to fix this soon.

ggreenway:
Thanks!

PR #25267, opened 2023-01-31
Fix flaky QUIC test for deferred logging

pksohn:
Commit Message: Fix flaky QUIC test for deferred logging
Risk Level: N/A, test-only

Fixes https://github.com/envoyproxy/envoy/issues/25264

### DIFF
--- a/test/integration/quic_http_integration_test.cc
+++ b/test/integration/quic_http_integration_test.cc
@@ -1194,7 +1194,7 @@ TEST_P(QuicHttpIntegrationTest, DeferredLogging) {
   codec_client_ = makeHttpConnection(makeClientConnection(lookupPort("http")));
   sendRequestAndWaitForResponse(default_request_headers_, /*request_size=*/0,
                                 default_response_headers_,
-                                /*response_size=*/0,
+                                /*response_size=*/1024 * 1024,
                                 /*upstream_index=*/0, TestUtility::DefaultTimeout);
   codec_client_->close();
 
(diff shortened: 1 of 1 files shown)


ggreenway (dismissed):

test/integration/quic_http_integration_test.cc
```diff
@@ -1203,7 +1203,7 @@ TEST_P(QuicHttpIntegrationTest, DeferredLogging) {
   std::vector<std::string> metrics = absl::StrSplit(log, ",");
   ASSERT_EQ(metrics.size(), 21);
   EXPECT_EQ(/* PROTOCOL */ metrics.at(0), "HTTP/3");
-  EXPECT_GT(/* ROUNDTRIP_DURATION */ std::stoi(metrics.at(1)), 0);
```
danzh2010, line 1206: If this returns 0, we can't tell if ROUNDTRIP_DURATION is actually 0ms or not working at all.
Can you add some delay at the fake_upstream_ end?

pksohn: Thanks; used a large response size which should ensure RT > 0 instead.

pksohn:
@ggreenway do you mind re-approving?

phlax:
not sure if this affects the 1.25 branch

/backport

test/integration/quic_http_integration_test.cc
ggreenway: I'm not sure sending 1MB will guarantee a non-zero RTT. A rate faster than 1GB/s would be under 1ms, and on localhost on a fast server that seems possible. Can you do what @danzh2010 suggested and force some delay?

danzh2010 (approved):
Thanks for the fix!

pksohn:
@ggreenway mind taking another look? Thanks!

ggreenway (approved):
LGTM. Thanks for fixing this!

PR #25278, opened 2023-02-01
OAuth: Use URL-encoding when adding URL as query parameters.

### LINKED ISSUE #23167: OAuth2 filter fails a request with Non-ASCII characters

yanavlasov:
Commit Message:
Decode these query parameters using reverse algorithm but leaving intact character sequences that must be encoded in URLs.

Additional Description:
This change addresses issue raised in #23167, which prevents OAuth filter from working with URLs that contain %-encoded UTF-8 characters in URL path or query.

During development I have also noticed that OAuth uses an odd character set for encoding URLs, which does not correspond to any standard I know. I have replaced it as well with well defined URL-encoding.

The behavioral changes are reflected in the additional tests.

Risk Level: Medium, Behavioral change in OAuth filter
Testing: Unit Tests
Docs Changes: Yes
Release Notes: Yes
Optional Runtime guard: envoy.reloadable_features.oauth_use_url_encoding
Optional Fixes #23167


source/common/http/utility.cc
```diff
@@ -1073,6 +1073,107 @@ std::string Utility::PercentEncoding::decode(absl::string_view encoded) {
...
+    0b00000000000000000000000000000000,
+    0b00000000000000000000000000000000,
+    0b00000000000000000000000000000000,
+    0b00000000000000000000000000000000,
+};
+
+bool testChar(const uint32_t table[8], char c) {
+  uint8_t uc = static_cast<uint8_t>(c);
+  return (table[uc >> 5] & (0x80000000 >> (uc & 0x1f))) != 0;
+}
```
snowp, line 1115: Is this an existing algorithm for checking whether a char is one of a well known set of chars? Not immediately obvious to me how it this works. First part takes the most significant few bits and use them to index into one of the bitmasks, and then the least significant bits are compared against the bitmask?

snowp (changes requested):
Thanks! Makes sense to me

source/common/http/utility.cc
yanavlasov: Yes, we use it in other places, like UHV. It is CPU cache friendly bit array lookup.

The character validity check could be implemented as a table of size 256 with 0 or 1 in each element indicating character validity. This algorithm packs all 1s and 0s into bits of 32 bit words. That expression with shifts is equivalent to division by 32 and getting the remainder of division by 32 to figure which 32 bit word and which bit in that word to check. It is more CPU cache friendly because table size it 64 bytes instead of 256.

docs/root/configuration/http/http_filters/oauth2_filter.rst
```diff
@@ -37,6 +37,20 @@ is set to true the filter will send over a
 cookie named ``BearerToken`` to the upstream. Additionally, the ``Authorization`` header will be populated
 with the same value.
 
+The OAuth filer encodes URLs in query parameters using the 
+`URL encoding alogirthm. <https://www.w3.org/TR/html5/forms.html#application/x-www-form-urlencoded-encoding-algorithm>`_
```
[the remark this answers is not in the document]
yanavlasov: done

source/common/http/utility.cc
snowp: Makes sense, thanks for explaining!

snowp (approved):

PR #25298, opened 2023-02-01
Remove option java_generic_services from proto files

### LINKED ISSUE #25172: Remove java_generic_services option from service proto files
*Title*: *Remove java_generic_services option from service proto files*

*Description*:
Since version 2.4.0 (2018) this option has been deprecated and such generated classes should not be used anymore (https://developers.google.com/protocol-buffers/docs/reference/java-generated?hl=en#service).
When using plugin that produce class with the same name, protoc complains about collision and failed.

Such plugins are provided by Quarkus gRPC Mutiny generator and it is not possible to generate Mutiny service from https://github.com/envoyproxy/java-control-plane/tree/main/api

scrocquesel:
Commit Message: Remove option java_generic_services from proto files
Additional Description: Generic services are deprecated since protoc version 2.4.0 (2010). Protoc plugins that generates code may require that generic services are disabled, so that they can generate their own classes of the same name.
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
Optional Fixes #25172

### DIFF
--- a/api/contrib/envoy/extensions/filters/network/sip_proxy/tra/v3alpha/tra.proto
+++ b/api/contrib/envoy/extensions/filters/network/sip_proxy/tra/v3alpha/tra.proto
@@ -14,7 +14,6 @@ option java_package = "io.envoyproxy.envoy.extensions.filters.network.sip_proxy.
 option java_outer_classname = "TraProto";
 option java_multiple_files = true;
 option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/sip_proxy/tra/v3alpha";
-option java_generic_services = true;
 option (udpa.annotations.file_status).package_version_status = ACTIVE;
 
 // [#protodoc-title: TRA]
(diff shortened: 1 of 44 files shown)


htuch:
Will need to update https://github.com/envoyproxy/envoy/blob/main/tools/protoprint/protoprint.py. @phlax since you responded to the original issue, do you want to take this PR?

phlax:
@scrocquesel thanks for raising this

i think these 2 lines https://github.com/envoyproxy/envoy/blob/eb4933645c3f5ddd47442174cce48cf59ba9e956/tools/protoprint/protoprint.py#L244-L245 need to be removed

it might also need to be removed from `tools/testdata/protoxform/envoy/v2/discovery_service.proto.active_or_frozen.gold`

phlax:
(as a side point - its generally better not to force push unless there is a reason - ie use merge over rebase etc - it doesnt matter so much at this point - but its generally frowned upon)

scrocquesel:
> 

@phlax Yes, I have only search the directive in proto files.

phlax (approved):
lgtm, thanks @scrocquesel 

cc @htuch @adisuissa for final api stamp

htuch (approved):

PR #25299, opened 2023-02-01
[balsa] Re-enable header integration tests for BalsaParser.

bencebeky:
[balsa] Re-enable IntegrationTest.[BadHeader,BadTrailer] for BalsaParser.

Explicitly enable trailers in IntegrationTest.BadTrailer, otherwise BalsaParser does not validate trailers.

Commit Message: [balsa] Re-enable header integration tests for BalsaParser.
Additional Description:
Testing: //test/integration:integration_test
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/http/http1/balsa_parser.cc
+++ b/source/common/http/http1/balsa_parser.cc
@@ -126,6 +126,13 @@ BalsaParser::BalsaParser(MessageType type, ParserCallbacks* connection, size_t m
     : message_type_(type), connection_(connection) {
   ASSERT(connection_ != nullptr);
 
+  quiche::HttpValidationPolicy http_validation_policy;
+  http_validation_policy.disallow_header_continuation_lines = false;
+  http_validation_policy.require_header_colon = true;
+  http_validation_policy.disallow_multiple_content_length = false;
+  http_validation_policy.disallow_transfer_encoding_with_content_length = false;
+  framer_.set_http_validation_policy(http_validation_policy);
+
   framer_.set_balsa_headers(&headers_);
   if (enable_trailers) {
     framer_.set_balsa_trailer(&trailers_);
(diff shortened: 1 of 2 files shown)


bencebeky:
Dianna and Biren: PTAL. Thanks!

test/integration/integration_test.cc
```diff
@@ -1162,7 +1153,8 @@ TEST_P(IntegrationTest, Http09Enabled) {
 
 TEST_P(IntegrationTest, Http09WithKeepalive) {
   if (http1_implementation_ == Http1ParserImpl::BalsaParser) {
-    // TODO(#21245): Re-enable this test for BalsaParser.
+    // HTTP/0.9 does not allow for headers.
+    // BalsaParser correctly ignores data after "\r\n".
```
diannahu, line 1157: :D:

source/common/http/http1/balsa_parser.cc
```diff
@@ -126,6 +126,13 @@ BalsaParser::BalsaParser(MessageType type, ParserCallbacks* connection, size_t m
     : message_type_(type), connection_(connection) {
   ASSERT(connection_ != nullptr);
 
+  quiche::HttpValidationPolicy http_validation_policy;
```
diannahu, line 129: Great use of the struct, very clear!

diannahu (approved):
Nice! Comments are no action, just emotional reactions.

bencebeky:
Kevin: PTAL.  For context, this change was made possible by https://github.com/google/quiche/commit/1fdac9bdbf85d27537e8ce47ca4f1eaa034f1ff5#diff-43919d5c357729334210d167fa9d89748b58c4e280723695f7529d1325d0edb2.

source/common/http/http1/balsa_parser.cc
KBaichoo: 🤞 that we always remember to set a default value in struct otherwise could run into balsa update wherein Envoy side isn't updated here and if no default => uninit value used :( .

KBaichoo (approved):
LGTM, thank you!

source/common/http/http1/balsa_parser.cc
bencebeky: Yup, good point.  It also helps to be explicit here about which checks are enabled and which ones are disabled.

PR #25297, opened 2023-02-01
mobile: Java using C++ builder under junit and testing equivalence

alyssawilk:
Functional changes:
   -- switching the c++ stats API back to simple strings
   -- change the Java DNS and Cluster APIs to ingress individual items 
   -- walking the stack in production code to determine if the code is under junit
   -- in test code, doing both current YAML string generation, and using the new C++ Java JNI function to generate the same YAML via the C++ builder, and verify equivalence.

The goal is to land this, have Lyft make sure it passes their CI, and then remove the Java YAML code generation and replace it with pure bootstrap code.

Risk Level: 
Testing: that's largely all this is
Docs Changes: inline
Release Notes: inline 

https://github.com/envoyproxy/envoy/issues/24976

PR #25315, opened 2023-02-02
mobile: renaming EnvoyNativeFilterConfig to EnvoyNativeConfig

alyssawilk:
in https://github.com/envoyproxy/envoy/pull/25297 we move non-filter config to using the struct with [name, typed_config] so doing a pure rename PR to remove "Filter" from the name

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

Part of https://github.com/envoyproxy/envoy/pull/25314


RyanTheOptimist (commented):
This seems to do a lot more than simply renaming. Is that expected? Should we tweak the PR description?

alyssawilk:
sorry this was based off the C++ builder but I hadn't merged main so the diffs weren't correct. Fixed!

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -54,7 +54,7 @@ public enum TrustChainVerification {
   public final String appId;
   public final TrustChainVerification trustChainVerification;
   public final String virtualClusters;
-  public final List<EnvoyNativeFilterConfig> nativeFilterChain;
+  public final List<EnvoyNativeConfig> nativeFilterChain;
```
jpsim, line 57: Should the variable be renamed to match?

```suggestion
  public final List<EnvoyNativeConfig> nativeConfigs;
```

jpsim (commented):
Seems ok to do, but if we're renaming types we should rename the variables/properties/parameters to keep those in sync.

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
alyssawilk: Good question  this stays as is because it's filter chain configuration.  In the API change one we also switch stats from "big yaml blob" to "name and typed config" though now that I think about it I wonder if I can rework so I'm going to hold on this one
/wait

PR #25322, opened 2023-02-02
access log: add downstream transport failure reason

botengyao:

access log: add downstream transport failure reason

Commit Message:
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/extensions/listener_managers/listener_manager/active_stream_listener_base.cc
```diff
@@ -112,6 +112,8 @@ void ActiveTcpConnection::onEvent(Network::ConnectionEvent event) {
   // Any event leads to destruction of the connection.
   if (event == Network::ConnectionEvent::LocalClose ||
       event == Network::ConnectionEvent::RemoteClose) {
+    connection_->connectionInfoSetter().setDownstreamTransportFailureReason(
```
tyxia, line 115: Here is possible that the connection is closed successfully without `transportFailure`, right? Do we want to treat that differently? i.e., not reporting error.

We can check` if (!connection_->transportFailureReason.empty())` first

test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
```diff
@@ -735,6 +742,44 @@ TEST_P(SslCertficateIntegrationTest, ServerEcdsaClientRsaOnly) {
...
+  server_rsa_cert_ = true;
+  server_ecdsa_cert_ = true;
+  ConnectionCreationFunction creator = [&]() -> Network::ClientConnectionPtr {
+    return makeSslClientConnection(rsaOnlyClientOptions());
+  };
+  testRouterRequestAndResponseWithBody(1024, 512, false, false, &creator);
+  checkStats();
+  codec_client_->close();
+  auto log_result = waitForAccessLog(listener_access_log_name_);
+  EXPECT_THAT(log_result, StartsWith("DOWNSTREAM_TRANSPORT_FAILURE_REASON=- FILTER_CHAIN_NAME=-"));
```
tyxia, line 780: Yea, this is why I think we could check `!empty` first. IMO, reporting failureReason (even empty) is confusing for successful connection case.  WDYT?

What will be log_result here if you don't set `setDownstreamTransportFailureReason` above? I think it is ok we just don't have access log for successful connection case as this log is called failureReason anyways.

tyxia (commented):
Thanks for working on this!

envoy/network/socket.h
```diff
@@ -112,6 +112,12 @@ class ConnectionInfoProvider {
    * @return roundTripTime of the connection
    */
   virtual const absl::optional<std::chrono::milliseconds>& roundTripTime() const PURE;
+
```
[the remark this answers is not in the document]
botengyao: Thanks! The reason is mainly that all the info in ConnectionInfo is related to downstream connection, we can even get the ssl_connection() it that class. downstream_transport_failure can be at that scope I think. I am okay to move, but let's see others' option.

botengyao (commented):
Thanks for the review!

envoy/network/socket.h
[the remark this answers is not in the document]
botengyao: Oh, I think I sticked to this `stream_info.downstreamAddressProvider().downstreamTransportFailureReason()` method. if ConnectionInfoProvider/Setter can be shared both upstream and downstream, it makes sense to move out from it. Thanks!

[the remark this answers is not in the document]
tyxia: I know there are some downstream related code there. However, by nature, it is part of connection and it can be used by both upstream and downstream. right? See enableSettingInterfaceName() in that class which is used by upstream. 

Off top of my head, as I mentioned above, you can have a new class DownstreamInfo and create it in the `ServerConnectionImpl::ServerConnectionImpl` to strictly differentiate them, but I think it is overkill just for adding failure reason.

source/extensions/listener_managers/listener_manager/active_stream_listener_base.cc
tyxia: Yea , I missed that you already check !empty in substitution_formatter.cc. Check here is not necessary

envoy/network/socket.h
```diff
@@ -112,6 +112,12 @@ class ConnectionInfoProvider {
...
+  /**
+   * @return absl::string_view the upstream transport failure reason,
+   *         e.g. certificate validation failed.
+   */
+  virtual absl::string_view downstreamTransportFailureReason() const PURE;
```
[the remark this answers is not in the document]
tyxia: No, there is no addictional copy if you `std::string, initializing std::string` as I mentioned above(last paragraph)
Also, see [go/totw/101](https://goto.google.com/totw/101)

Since you already stored `downstream_transport_failure_reason_` as string, you can just return std::string. ( neither const string& nor absl::string_view)

So, returning `absl::string_view` doesn't save anything.

envoy/network/socket.h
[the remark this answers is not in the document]
botengyao: Agreed, Thanks!

envoy/network/socket.h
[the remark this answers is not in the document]
botengyao: Oh I see, thanks so much! It may make sense to replace others.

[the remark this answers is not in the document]
tyxia: Also, I do see lots of place returning `absl::string_view` in Envoy code base. I think it is fine to keep as it is.
I just want to mention this for your reference.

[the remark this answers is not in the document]
botengyao: Good to know this C++17 stuff 😄, thanks!

api/envoy/data/accesslog/v2/accesslog.proto
```diff
@@ -171,6 +171,12 @@ message AccessLogCommon {
   // state serialized to any message other than `google.protobuf.Any` it will be packed into
   // `google.protobuf.Any`.
   map<string, google.protobuf.Any> filter_state_objects = 21;
+
+  // If downstream connection in listenner failed due to transport socket (e.g. TLS handshake), provides the
+  // failure reason from the transport socket. The format of this field depends on the configured
+  // upstream transport socket. Common TLS failures are in
```
htuch, line 177: Why does the downstream transport failure depend on upstream socket selection?

api/envoy/data/accesslog/v2/accesslog.proto
```diff
@@ -70,7 +70,7 @@ message ConnectionProperties {
 }
 
 // Defines fields that are shared by all Envoy access logs.
-// [#next-free-field: 22]
+// [#next-free-field: 23]
```
htuch, line 73: This is v2 API and frozen. Do you want to be updating v3?

docs/root/configuration/observability/access_log/usage.rst
```diff
@@ -516,6 +516,17 @@ The following command operators are supported:
   TCP/UDP
     Not implemented ("-")
 
+.. _config_access_log_format_downstream_transport_failure_reason:
+
+%DOWNSTREAM_TRANSPORT_FAILURE_REASON%
+  HTTP/TCP
+    If downstream connection failed due to transport socket (e.g. TLS handshake), provides the failure
```
htuch, line 523: Do we generate access logs for failed TLS handshakes?

botengyao: Yes, one possible error for TLS termination is: `DOWNSTREAM_TRANSPORT_FAILURE_REASON=TLS_error:_268435709:SSL_routines:"
                           "OPENSSL_internal:NO_COMMON_SIGNATURE_ALGORITHMS`

api/envoy/data/accesslog/v2/accesslog.proto
botengyao: Thanks! Yeah, V3 should be used.

api/envoy/data/accesslog/v2/accesslog.proto
botengyao: Ops, thanks for catching this. It depends on listener format setting I think.

tyxia (commented):
Overall LGTM, just some comments. Thanks!

htuch (commented):
/lgtm api

docs/root/configuration/observability/access_log/usage.rst
yanavlasov: Note, that the DOWNSTREAM_TRANSPORT_FAILURE_REASON is valid only in the listener access log. The HTTP or TCP access logs would always observe empty DOWNSTREAM_TRANSPORT_FAILURE_REASON.

I think it needs to be explained in the doc to reduce possible confusion.

yanavlasov (changes requested):
/wait

docs/root/configuration/observability/access_log/usage.rst
botengyao: Thanks! done!

source/common/formatter/substitution_formatter.cc
```diff
@@ -1466,6 +1466,22 @@ const StreamInfoFormatter::FieldExtractorLookupTbl& StreamInfoFormatter::getKnow
                                     return connection_info.urlEncodedPemEncodedPeerCertificate();
                                   });
                             }}},
+                          {"DOWNSTREAM_TRANSPORT_FAILURE_REASON",
+                           {CommandSyntaxChecker::COMMAND_ONLY,
+                            [](const std::string&, const absl::optional<size_t>&) {
+                              return std::make_unique<StreamInfoStringFieldExtractor>(
+                                  [](const StreamInfo::StreamInfo& stream_info) {
+                                    absl::optional<std::string> result;
```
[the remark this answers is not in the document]
botengyao: Thanks!

test/extensions/transport_sockets/tls/integration/ssl_integration_test.cc
```diff
@@ -97,6 +99,11 @@ SslIntegrationTestBase::makeSslClientConnection(const ClientSslTransportOptions&
       client_transport_socket_factory_ptr->createTransportSocket({}, nullptr), nullptr, nullptr);
 }
 
+void SslIntegrationTestBase::setupDownstreamTransportAccessLog() {
```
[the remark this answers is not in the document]
botengyao: Done.

source/common/stream_info/stream_info_impl.h
```diff
@@ -375,6 +376,14 @@ struct StreamInfoImpl : public StreamInfo {
   void setIsShadow(bool is_shadow) { is_shadow_ = is_shadow; }
   bool isShadow() const override { return is_shadow_; }
 
+  void setDownstreamTransportFailureReason(absl::string_view failure_reason) override {
+    downstream_transport_failure_reason_ = std::string(failure_reason);
+  }
+
+  absl::string_view downstreamTransportFailureReason() const override {
+    return absl::string_view(downstream_transport_failure_reason_);
```
tyxia, line 384: Explicit `absl::string_view()`  conversion is redundant. The implicit conversion from string to string_view should be triggered

test/mocks/stream_info/mocks.cc
```diff
@@ -184,6 +184,13 @@ MockStreamInfo::MockStreamInfo()
       .WillByDefault(Invoke([this](const BytesMeterSharedPtr& downstream_bytes_meter) {
         downstream_bytes_meter_ = downstream_bytes_meter;
       }));
+  ON_CALL(*this, setDownstreamTransportFailureReason(_))
+      .WillByDefault(Invoke([this](absl::string_view failure_reason) {
+        downstream_transport_failure_reason_ = std::string(failure_reason);
+      }));
+  ON_CALL(*this, downstreamTransportFailureReason()).WillByDefault(Invoke([this]() {
+    return absl::string_view(downstream_transport_failure_reason_);
```
tyxia, line 192: Ditto. 
The key here is returning live value of `downstream_transport_failure_reason_`, string_view conversion here is redundant.

Alternatively you can use  `ReturnPointee` to return the live value, but it is up to you.
```
 ON_CALL(*this, downstreamTransportFailureReason())
      .WillByDefault(ReturnPointee(&downstream_transport_failure_reason_));
```

tyxia (approved):
LGTM, modulo two nits.

Thanks for your patience!

source/common/stream_info/stream_info_impl.h
botengyao: My initial thinking is same,  but it will fail clang-asan check without this.

botengyao: Thanks, changed back. The mock make me think somewhere wired.

test/mocks/stream_info/mocks.cc
botengyao: Thanks, done.

botengyao:
Hi All, PTAL, Thanks!

tyxia (dismissed):
LGTM, Thanks

api/envoy/data/accesslog/v3/accesslog.proto
```diff
@@ -215,6 +215,11 @@ message AccessLogCommon {
   // :ref:`stream_id <envoy_v3_api_field_data.accesslog.v3.AccessLogCommon.stream_id>` to
   // correlate all these intermediate log entries and final log entry.
   bool intermediate_log_entry = 27;
+
+  // If downstream connection in listener failed due to transport socket (e.g. TLS handshake), provides the
+  // failure reason from the transport socket. The format of this field depends on the configured
+  // listener socket. Common TLS failures are in :ref:`TLS trouble shooting <arch_overview_ssl_trouble_shooting>`.
```
yanavlasov, line 221:
```suggestion
  // downstream transport socket. Common TLS failures are in :ref:`TLS trouble shooting <arch_overview_ssl_trouble_shooting>`.
```

yanavlasov (changes requested):
LGTM with a small documentation correction.

/wait

api/envoy/data/accesslog/v3/accesslog.proto
botengyao: Done, thanks!

yanavlasov (approved):

htuch (commented):
/lgtm api

botengyao:
Thanks all for the review! Hi @envoyproxy/senior-maintainers, PTAL and can we get this merged? Thanks

htuch (approved):

XinyiZhangAws:
Since Envoy does have attribute upstream.transport_failure_reason, do we have a plan to add downstream_transport_failure_reason as [connection attributes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes.html#connection-attributes) as well?

PR #25259, opened 2023-01-31
mobile: finishing C++ proto builder and making it default

alyssawilk:
Always doing the whitespace regex takes the mobile builder from
Envoy::BM_LoadFromString        239 ms          239 ms            3
Envoy::BM_LoadFromProto       0.871 ms        0.870 ms          805
to
Envoy::BM_LoadFromString        239 ms          239 ms            3
Envoy::BM_LoadFromProto        1.76 ms         1.75 ms          394

so it's impactful (double!) but 
1) we can fix it once all the builders are moved
2) we mostly defer to platform verifiers in production so don't use certs so don't do the substitution
3) it's still orders of magnitude faster than YAML

Risk Level: Medium
Testing: improved unit test coverage
Docs Changes: n/a
Release Notes:
Part of https://github.com/envoyproxy/envoy/issues/24976

PR #25324, opened 2023-02-02
mobile: moving stats back to string vector

alyssawilk:
Undoing the API change from https://github.com/envoyproxy/envoy/pull/25259/files and adding builder support for the original API.

Risk Level: low
Testing: yes
Docs Changes: no
Release Notes: no
Part of https://github.com/envoyproxy/envoy/issues/24976

### DIFF
--- a/mobile/library/cc/engine_builder.cc
+++ b/mobile/library/cc/engine_builder.cc
@@ -84,8 +84,8 @@ EngineBuilder& EngineBuilder::setOnEngineRunning(std::function<void()> closure) 
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addStatsSinks(const std::vector<std::string>& stats_sinks) {
-  stats_sinks_ = stats_sinks;
+EngineBuilder& EngineBuilder::addStatsSinks(std::vector<std::string> stats_sinks) {
+  stats_sinks_ = std::move(stats_sinks);
   return *this;
 }
 
--- a/mobile/library/cc/engine_builder.cc
+++ b/mobile/library/cc/engine_builder.cc
@@ -85,7 +85,7 @@ EngineBuilder& EngineBuilder::setOnEngineRunning(std::function<void()> closure) 
 }
 
 EngineBuilder& EngineBuilder::addStatsSinks(const std::vector<std::string>& stats_sinks) {
-  stats_sinks_ = std::move(stats_sinks);
+  stats_sinks_ = stats_sinks;
   return *this;
 }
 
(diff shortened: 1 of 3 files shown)


mobile/library/cc/engine_builder.cc
```diff
@@ -84,8 +84,8 @@ EngineBuilder& EngineBuilder::setOnEngineRunning(std::function<void()> closure)
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addStatsSink(std::string name, std::string typed_config) {
-  stats_sinks_.push_back(std::make_pair(name, typed_config));
+EngineBuilder& EngineBuilder::addStatsSinks(const std::vector<std::string>& stats_sinks) {
```
RyanTheOptimist, line 87: nit: This argument should be by value, not by const ref since it's being moved down below. This allows the caller to move the vector into this method and avoid a copy.

mobile/library/cc/engine_builder.cc
```diff
@@ -903,6 +894,11 @@ std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap> EngineBuilder::generate
 
   bootstrap->mutable_typed_dns_resolver_config()->CopyFrom(
       *dns_cache_config->mutable_typed_dns_resolver_config());
+
+  for (auto& sink_yaml : stats_sinks_) {
```
RyanTheOptimist, line 898: nit: auto& => const std::string&

mobile/library/cc/engine_builder.h
```diff
@@ -75,7 +75,7 @@ class EngineBuilder {
   EngineBuilder& addDnsPreresolveHostnames(const std::vector<std::string>& hostnames);
   EngineBuilder& addNativeFilter(std::string name, std::string typed_config);
   EngineBuilder& enableAdminInterface(bool admin_interface_on);
-  EngineBuilder& addStatsSink(std::string name, std::string typed_config);
+  EngineBuilder& addStatsSinks(const std::vector<std::string>& stat_sinks);
```
RyanTheOptimist, line 78: (if we change the signature in the .cc file, we need to change the signature here, of course)

RyanTheOptimist (approved):
Nice! I think this API is definitely cleaner.
Thanks for doing this. LGTM mod a nit.

alyssawilk:
Changing back to what we had in https://github.com/envoyproxy/envoy/pull/25259/files which was const string& but member_ = argument

jpsim (approved):
This is definitely cleaner, though a user of this API would still be left on their own to figure out what kinds of strings are expected in this vector. Ideally there'd be a link or description of the expected format around the API, but not blocking on that since it wasn't there before either.

RyanTheOptimist:
> Changing back to what we had in https://github.com/envoyproxy/envoy/pull/25259/files which was const string& but member_ = argument

This is strictly less efficient. Is there a reason to prefer this approach? (FWIW, we switch the request headers builder over to "pass by value and move" for efficiency.)

RyanTheOptimist (approved):

PR #25325, opened 2023-02-02
mobile: allow overriding dns cache key value save interval

jpsim:
In Lyft's usage, we have up to 14 calls per second to save DNS results to the key value store. Buffering these to save at most once per second should be sufficient.

Commit Message: mobile: set dns cache key value save interval to 1s
Additional Description: In Lyft's usage, we have up to 14 calls per second to save DNS results to the key value store. Buffering these to save at most once per second should be sufficient.
Risk Level: Low
Testing: In Lyft's apps
Docs Changes: Added
Release Notes: Added

### DIFF
--- a/mobile/library/cc/engine_builder.h
+++ b/mobile/library/cc/engine_builder.h
@@ -69,7 +69,7 @@ public:
   // Adds an ADS layer.
   EngineBuilder& setAggregatedDiscoveryService(const std::string& api_type,
                                                const std::string& address, const int port);
-  EngineBuilder& enableDnsCache(bool dns_cache_on);
+  EngineBuilder& enableDnsCache(bool dns_cache_on, int save_interval_seconds = 1);
   EngineBuilder& setForceAlwaysUsev6(bool value);
   EngineBuilder& setSkipDnsLookupForProxiedRequests(bool value);
   EngineBuilder& addDnsPreresolveHostnames(const std::vector<std::string>& hostnames);
@@ -140,6 +140,7 @@ private:
   std::string ads_address_ = "";
   int ads_port_;
   bool dns_cache_on_ = false;
+  bool dns_cache_save_interval_seconds_ = 1;
 
   absl::flat_hash_map<std::string, KeyValueStoreSharedPtr> key_value_stores_{};
 
--- a/mobile/library/cc/engine_builder.h
+++ b/mobile/library/cc/engine_builder.h
@@ -140,7 +140,7 @@ private:
   std::string ads_address_ = "";
   int ads_port_;
   bool dns_cache_on_ = false;
-  bool dns_cache_save_interval_seconds_ = 1;
+  int dns_cache_save_interval_seconds_ = 1;
 
   absl::flat_hash_map<std::string, KeyValueStoreSharedPtr> key_value_stores_{};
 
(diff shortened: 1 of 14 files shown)


RyanTheOptimist (dismissed):
LGTM, though Chrome's analogous timer is 60s[1] so we could probably use a larger value if we wanted to limit IO.

1. https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_server_properties.cc;l=34;drc=2a7fa9b6065bba9cbe273ddc4f5e5a4a51bb450d;bpv=1;bpt=1

jpsim:
Oops, marked as ready to review too quickly, tests are failing. I'll update that.

I'll expose a builder parameter to set the interval since different apps might want different values. For us, 1s is fine and 60s would be too long, since storing the results in a sqlite db should be very efficient with our kinds of data, and we have app sessions shorter than 60s.

alyssawilk:
@RyanTheOptimist are we using the DNS cache yet for YT? If not I'd be inclined to suggest we switch the time interval now, and TODO the builder change for 2 weeks just because there's a lot of builder changes in flight
if it'd affect us now, I'll deal with the merge conflicts :-P

jpsim:
I've wired this through all the builder paths to keep things in sync as much as possible and minimize thrash with the proto builder changes.

RyanTheOptimist (approved):
Looks great! Should we have a test for the C++ builder too?

jpsim:
> Looks great! Should we have a test for the C++ builder too?

Yes, I'll add that.

mobile/library/cc/engine_builder.h
```diff
@@ -140,7 +140,7 @@ class EngineBuilder {
   std::string ads_address_ = "";
   int ads_port_;
   bool dns_cache_on_ = false;
-  bool dns_cache_save_interval_seconds_ = 1;
+  int dns_cache_save_interval_seconds_ = 1;
```
RyanTheOptimist, line 143: Ooo! Nice catch.

mobile/test/cc/unit/envoy_config_test.cc
```diff
@@ -38,7 +38,7 @@ TEST(TestConfig, ConfigIsApplied) {
       .addStatsFlushSeconds(654)
       .setAppVersion("1.2.3")
       .setAppId("1234-1234-1234")
-      .enableDnsCache(true)
+      .enableDnsCache(true, /* save_interval_seconds */ 1)
```
RyanTheOptimist, line 41: Since 1 is the default, it would probably be a good idea to select a non-default value.

RyanTheOptimist (dismissed):
Thanks for doing this!

mobile/library/cc/engine_builder.h
jpsim: 😬

mobile/test/cc/unit/envoy_config_test.cc
jpsim: Yes, although I think something is broken here, as I appear to be getting a yaml comparison error when anything other than the default value is specified here:

```
[2023-02-03 15:28:27.464][5642470][critical][assert] [library/cc/engine_builder.cc:961] assert failure: generatedStringMatchesGeneratedBoostrap(generateConfigStr(), *bootstrap). Details: Native C++ checks failed
```

jpsim: Ordering was wrong.

RyanTheOptimist (approved):
Woo hoo!

PR #25327, opened 2023-02-02
deps: Bump project/publishing deps

### LINKED ISSUE #25001: CI auto-release from `main` uses current `main` even if `main` has progressed
During release of 1.25.0 the branch (ie `main`) was reopened before CI had cut the release leading to the release using a newer commit

There is a related issue in that when you manually create the new patch branch - ie `release/v1.25` with no additional commits that also tries to create a tag

We can probably resolve both of these issues together somehow

phlax:
Fix #25001 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
the new flags provided by the pytooling deps have been tested here https://github.com/gpt-4/envoy/releases

this resolves the issue of making every release a "latest" release, and in the case of cutting a new minor from `main` it will mark it as latest

what it doesnt do is know to mark a new patch release against the lastest stable branch as latest - which it should, but i couldnt think of an easy way to do it - so for now that step must be manual

phlax:
cc @mattklein123 this is destined for backports

htuch (approved):

PR #25328, opened 2023-02-02
Simplify function calls in base_client_integration_test.cc

caschoener:
Just cleaning up some tasty spaghetti. Previously we had this hidden requirement that any tests inheriting from base_client_integration_test needed to override tearDown to call cleanUp() or you'd get a segfault.


mobile/test/common/integration/client_integration_test.cc
```diff
@@ -23,9 +23,7 @@ class ClientIntegrationTest : public BaseClientIntegrationTest,
     // TODO(abeyad): Add paramaterized tests for HTTP1, HTTP2, and HTTP3.
     setUpstreamProtocol(Http::CodecType::HTTP1);
   }
-
-  void TearDown() override {
-    cleanup();
+  void TearDown() {
```
caschoener, line 26: Sadly this needs to stick around because this class inherits from 2 classes which both define tearDown. Otherwise we get an error.

abeyad:
looks like this one has legitimate test failures

caschoener:
Talked to ryan.  These tests all inherit from BaseClientIntegrationTest which declares TearDown, and gtest which declares a virtual TearDown. So the only way to resolve it is to define TearDown in the subclass and call the one we want to run.

Still probably a worthwhile change, and I'll add a comment explaining it.

mobile/test/common/integration/base_client_integration_test.h
```diff
@@ -41,15 +41,16 @@ class BaseClientIntegrationTest : public BaseIntegrationTest {
   BaseClientIntegrationTest(Network::Address::IpVersion ip_version,
                             const std::string& bootstrap_config = defaultConfig());
   virtual ~BaseClientIntegrationTest() = default;
+  // Note: This class does not inherit from testing::Test and so this TearDown() method
+  // does not override testing::Test::TearDown(). As a result, it will not be called
+  // automatically by gtest during shutdown and must be called manually.
+  void TearDown();
```
abeyad, line 47: this needs to be `virtual` since client_integration_test overrides it

RyanTheOptimist: Confusingly, ClientIntegrationTest does not override it! Instead it overrides testing::Test::Teardown().

jpsim (dismissed):

abeyad (approved):

caschoener:
Sorry for force push accidentally included a commit from another branch

abeyad (approved):
Thanks for the nice cleanup!

PR #25329, opened 2023-02-03
Let all Android builds use Android special getifaddrs()

### LINKED ISSUE #25326: Envoy-mobile clients fail to speak IPV6 on Android clients with ndk >=24
Currently the posix version of getifaddrs() has 2 paths:
For Android with ndk version < 24, it calls getifaddrs that's specific to [Android](https://github.com/envoyproxy/envoy/blob/main/source/common/api/posix/os_sys_calls_impl.cc#L361).
For all other situations, it uses Envoy's default getifaddrs() [impl](https://github.com/envoyproxy/envoy/blob/main/source/common/api/posix/os_sys_calls_impl.cc#L391).

It is discovered that the default getifaddrs() doesn't quite work with Android with ndk level >=24. It's not able to send traffic in IPV6 but works fine in IPV4.
This is not an issue now since all Envoy-mobile applications are built with <24 ndk. But it's an ticking bomb that will go off with future toolchain upgrade.

It's verified that the <24 code path also works with >=24 build. Thus the easiest solution is to remove the __ANDROID_API__ < 24 guard and jus run the Android specific getifaddrs() for all Android ndk versions.

RenjieTang:

Commit Message: Let all Android builds use Android special getifaddrs()
Additional Description: 
Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
Fixes #25326

### DIFF
--- a/source/common/api/posix/os_sys_calls_impl.cc
+++ b/source/common/api/posix/os_sys_calls_impl.cc
@@ -8,7 +8,7 @@
 #include "source/common/api/os_sys_calls_impl.h"
 #include "source/common/network/address_impl.h"
 
-#if defined(__ANDROID_API__) && __ANDROID_API__ < 24
+#if defined(__ANDROID_API__)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wold-style-cast"
 namespace android {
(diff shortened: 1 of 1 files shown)


wbpcode:
cc @alyssawilk seems this is related to mobile building?

RenjieTang:
I've created https://github.com/envoyproxy/envoy/pull/25414 in favor of this PR.
Once that one lands, I will close this one.

PR #25333, opened 2023-02-03
ZK filter: add more trace logs

Winbobob:
Signed-off-by: Zhewei Hu <zhu@pinterest.com>

Commit Message: ZK filter: add more trace logs
Additional Description: These additional trace logs will be helpful to debug decoder errors.
Risk Level: low
Testing: unit tests
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/extensions/filters/network/zookeeper_proxy/decoder.cc
+++ b/source/extensions/filters/network/zookeeper_proxy/decoder.cc
@@ -48,6 +48,7 @@ void DecoderImpl::decodeOnData(Buffer::Instance& data, uint64_t& offset) {
 
   // Check message length.
   const int32_t len = helper_.peekInt32(data, offset);
+  ENVOY_LOG(trace, "zookeeper_proxy: decoding request with len {} at offset {}", len, offset);
   ensureMinLength(len, INT_LENGTH + XID_LENGTH);
   ensureMaxLength(len);
 
@@ -64,6 +65,7 @@ void DecoderImpl::decodeOnData(Buffer::Instance& data, uint64_t& offset) {
   //       However, some client implementations might expose setWatches
   //       as a regular data request, so we support that as well.
   const int32_t xid = helper_.peekInt32(data, offset);
+  ENVOY_LOG(trace, "zookeeper_proxy: decoding request with xid {} at offset {}", xid, offset);
   switch (static_cast<XidCodes>(xid)) {
   case XidCodes::ConnectXid:
     parseConnect(data, offset, len);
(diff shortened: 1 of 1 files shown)


Winbobob:
cc @JuniorHsu, please review, thanks!

source/extensions/filters/network/zookeeper_proxy/decoder.cc
```diff
@@ -64,6 +65,7 @@ void DecoderImpl::decodeOnData(Buffer::Instance& data, uint64_t& offset) {
   //       However, some client implementations might expose setWatches
   //       as a regular data request, so we support that as well.
   const int32_t xid = helper_.peekInt32(data, offset);
+  ENVOY_LOG(trace, "zookeeper_proxy: decoding request with xid {} at offset {}", xid, offset);
```
JuniorHsu, line 68: Let merge line 46, 51 into this log to avoid too many log

source/extensions/filters/network/zookeeper_proxy/decoder.cc
```diff
@@ -204,7 +210,9 @@ void DecoderImpl::decodeOnWrite(Buffer::Instance& data, uint64_t& offset) {
 
   // Control responses that aren't connect, with XIDs <= 0.
   const auto zxid = helper_.peekInt64(data, offset);
+  ENVOY_LOG(trace, "zookeeper_proxy: decoding response with zxid {} at offset {}", zxid, offset);
   const auto error = helper_.peekInt32(data, offset);
+  ENVOY_LOG(trace, "zookeeper_proxy: decoding response with error {} at offset {}", error, offset);
```
JuniorHsu, line 215: Same here. We don't need 5 lines of logs.

JuniorHsu (changes requested):

source/extensions/filters/network/zookeeper_proxy/decoder.cc
Winbobob: I am thinking if we merge all logs in the end and some errors happen above, we may lost all the information.

JuniorHsu: IMO we can merge 172, 178 into 182. and merge 213 and 215. 
Any other concerns?

Winbobob: So if we merge L172, L177 into L182 and either `ensureMinLength` or `ensureMaxLength` raises an exception, we only know the packet is either too big or too small but not know the exact length extracted from the buffer.

JuniorHsu: Thanks for your clarification!
Log will show the exception body IIRC, but I agree that we can keep log before those exception.
On the other hand, we should avoid exception in the data path if possible, but that could be a follow-up.

Let's merge the logs as much as possible but still keep the info exposure at the same degree.

source/extensions/filters/network/zookeeper_proxy/decoder.cc
Winbobob: I did not merge L51 into L68 since there are length checks at L52-53.

JuniorHsu (approved):
Thanks!

snowp (approved):
Thanks!

PR #25334, opened 2023-02-03
docs: Use PR commit sha

phlax:
This uses the commit sha from the PR HEAD in the staged docs version string, which should make it easier to associate with a pushed PR version

The staged docs are still published to a URL using the merged commit sha - so individual builds (from the same pr commit) are still distinct

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -195,6 +195,7 @@ stages:
         ENVOY_RBE: "true"
         BAZEL_BUILD_EXTRA_OPTIONS: "--config=remote-ci --jobs=$(RbeJobs)"
         AZP_BRANCH: $(Build.SourceBranch)
+        AZP_COMMIT_SHA: $(system.pullRequest.sourceCommitId)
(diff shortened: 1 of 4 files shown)


phlax:
as an example check the current docs - the version string uses 91b2ae9f3a90afb3c74bdd8f74725aecd4cc86d4 altho the url will still redirect to the merged (main) sha

phlax:
im wondering whether we should just use the commit sha for the url also - the only way i can imagine 2 builds off the same commit (with different merge) is if a test that has passed is manually rerun - its a pretty contrived case and even then would kinda just work

lizan (approved):
I'm fine either way merge sha or commit sha for URL, up to you @phlax

phlax:
k - ill land as is - i think there is some benefit to seeing somewhere that its not actually the commit, but a merge

PR #25337, opened 2023-02-03
add reset reason to stream log

ramaraochavali:
Commit Message: add reset reason to stream log
Additional Description: add reset reason to stream log
Risk Level: N/A
Testing: N/A
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/source/common/http/conn_manager_impl.cc
+++ b/source/common/http/conn_manager_impl.cc
@@ -1668,7 +1668,8 @@ void ConnectionManagerImpl::ActiveStream::onResetStream(StreamResetReason reset_
   //       If we need to differentiate we need to do it inside the codec. Can start with this.
   const absl::string_view encoder_details = response_encoder_->getStream().responseDetails();
   ENVOY_STREAM_LOG(debug, "stream reset: reset reason: {}, response details: {}", *this,
-                   Http::Utility::resetReasonToString(reset_reason), encoder_details);
+                   Http::Utility::resetReasonToString(reset_reason),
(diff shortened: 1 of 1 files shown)


source/common/http/conn_manager_impl.cc
```diff
@@ -1666,12 +1666,13 @@ void ConnectionManagerImpl::ActiveStream::onResetStream(StreamResetReason reset_
   //       3) The codec RX a reset
   //       4) The overload manager reset the stream
   //       If we need to differentiate we need to do it inside the codec. Can start with this.
-  ENVOY_STREAM_LOG(debug, "stream reset", *this);
+  const absl::string_view encoder_details = response_encoder_->getStream().responseDetails();
```
cpakulski, line 1669: `responseDetails` may return an empty string. I think you should check for that and either put something like "N/A" or skip `response_details` all together.

ramaraochavali: We do similar thing here https://github.com/envoyproxy/envoy/blob/b41adf18d0419d4c34cc153bf53b3d27295068a3/source/common/router/router.cc#L1279 - So if it is empty, it will be logged as empty I guess

wbpcode: agree it would be better to log a "-" or "N/A" by `encoder_details.empty() ? absl::string_view{"-"} : encoder_details`

wbpcode (approved):
LGTM, thanks.

PR #25351, opened 2023-02-03
nit: fix active listener base comment

botengyao:

Commit Message:
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/extensions/listener_managers/listener_manager/active_stream_listener_base.cc
+++ b/source/extensions/listener_managers/listener_manager/active_stream_listener_base.cc
@@ -121,7 +121,7 @@ void OwnedActiveStreamListenerBase::removeConnection(ActiveTcpConnection& connec
   ActiveConnections& active_connections = connection.active_connections_;
   ActiveConnectionPtr removed = connection.removeFromList(active_connections.connections_);
   dispatcher().deferredDelete(std::move(removed));
-  // Delete map entry only iff connections becomes empty.
+  // Delete map entry only if connections become empty.
   if (active_connections.connections_.empty()) {
     auto iter = connections_by_context_.find(&active_connections.filter_chain_);
     ASSERT(iter != connections_by_context_.end());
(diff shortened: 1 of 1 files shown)


source/extensions/listener_managers/listener_manager/active_stream_listener_base.cc
```diff
@@ -121,7 +121,7 @@ void OwnedActiveStreamListenerBase::removeConnection(ActiveTcpConnection& connec
   ActiveConnections& active_connections = connection.active_connections_;
   ActiveConnectionPtr removed = connection.removeFromList(active_connections.connections_);
   dispatcher().deferredDelete(std::move(removed));
-  // Delete map entry only iff connections becomes empty.
+  // Delete map entry only if connections_ becomes empty.
```
phlax, line 124: i think this might be correct - iff can mean "if and only if" - not sure if that was the original authors intention

botengyao: Thanks! `if and only if` or `iff` makes sense from the below logic. My first glance makes me think it could be a typo. Will close this PR and not going to be nitpicky ;-)

phlax: maybe we should change it to that

it may have been somewhere else (dont rem) but its not the first time someone has raised similar

botengyao: Let make it more clear, to use `Delete map entry if and only if connections_ become empty` explicitly. I don't see the big difference with `if and only if` and `only if` for this logic. Maybe for math it is different. WDYT?

phlax: yep sgtm - its confusing as its just now with `only iff` - so sounds much better

re logic - not sure - my guess tho is that `only iff` is/was supposed to mean `if and only if`

phlax (approved):
lgtm, thanks @botengyao

PR #25353, opened 2023-02-03
[file_system_http_cache] Rename stat with tags

ravenblackx:
Commit Message: [file_system_http_cache] Rename stat with tags
Additional Description: Discovered in testing that a stat named `a.b.c` is not the same as a stat that prints out as `a.b.c`, if `b.c` is derived from tags. This changes the stats for `file_system_http_cache` from having the cache path in the stat-name to having the cache path as a tag value.
Risk Level: Very low, extension is WIP.
Testing: Coverage is there; stat-names could be tested better, but right now the fake metrics don't support it correctly. It would make more sense to test stat name generation while creating and migrating to a stats-common method to make tagged stats anyway.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/stats/isolated_store_impl.cc
+++ b/source/common/stats/isolated_store_impl.cc
@@ -19,17 +19,28 @@ IsolatedStoreImpl::IsolatedStoreImpl(std::unique_ptr<SymbolTable>&& symbol_table
 }
 
 IsolatedStoreImpl::IsolatedStoreImpl(SymbolTable& symbol_table)
-    : alloc_(symbol_table), counters_([this](StatName name) -> CounterSharedPtr {
-        return alloc_.makeCounter(name, name, StatNameTagVector{});
+    : alloc_(symbol_table),
+      counters_([this](const TagUtility::TagStatNameJoiner& joiner,
+                       StatNameTagVectorOptConstRef tags) -> CounterSharedPtr {
+        return alloc_.makeCounter(joiner.nameWithTags(), joiner.tagExtractedName(),
+                                  tags ? tags->get() : StatNameTagVector{});
       }),
-      gauges_([this](StatName name, Gauge::ImportMode import_mode) -> GaugeSharedPtr {
-        return alloc_.makeGauge(name, name, StatNameTagVector{}, import_mode);
+      gauges_([this](const TagUtility::TagStatNameJoiner& joiner, StatNameTagVectorOptConstRef tags,
+                     Gauge::ImportMode import_mode) -> GaugeSharedPtr {
+        return alloc_.makeGauge(joiner.nameWithTags(), joiner.tagExtractedName(),
+                                tags ? tags->get() : StatNameTagVector{}, import_mode);
       }),
-      histograms_([this](StatName name, Histogram::Unit unit) -> HistogramSharedPtr {
-        return HistogramSharedPtr(new HistogramImpl(name, unit, *this, name, StatNameTagVector{}));
+      histograms_([this](const TagUtility::TagStatNameJoiner& joiner,
+                         StatNameTagVectorOptConstRef tags,
+                         Histogram::Unit unit) -> HistogramSharedPtr {
+        return {new HistogramImpl(joiner.nameWithTags(), unit, *this, joiner.tagExtractedName(),
+                                  tags ? tags->get() : StatNameTagVector{})};
       }),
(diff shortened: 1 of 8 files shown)


ravenblackx:
Stacks on #25386 so that I could write a useful test; for review purposes `isolated_store_*` files can be ignored.

source/common/stats/isolated_store_impl.cc
```diff
@@ -19,17 +19,28 @@ IsolatedStoreImpl::IsolatedStoreImpl(std::unique_ptr<SymbolTable>&& symbol_table
...
+        return {new HistogramImpl(joiner.nameWithTags(), unit, *this, joiner.tagExtractedName(),
+                                  tags ? tags->get() : StatNameTagVector{})};
       }),
-      text_readouts_([this](StatName name, TextReadout::Type) -> TextReadoutSharedPtr {
-        return alloc_.makeTextReadout(name, name, StatNameTagVector{});
+      text_readouts_([this](const TagUtility::TagStatNameJoiner& joiner,
+                            StatNameTagVectorOptConstRef tags,
+                            TextReadout::Type) -> TextReadoutSharedPtr {
+        return alloc_.makeTextReadout(joiner.nameWithTags(), joiner.tagExtractedName(),
+                                      tags ? tags->get() : StatNameTagVector{});
```
jmarantz, line 43: Just to make this less verbose could you factor the conditional expression into a static local helper method with signature
`StatNameTagOrVector tagVectorFromOpt(StatNameTagVectorOptConstRef tags);`

source/extensions/http/cache/file_system_http_cache/stats.h
```diff
@@ -19,19 +19,44 @@ namespace FileSystemHttpCache {
...
+  , NAME##_(Envoy::Stats::Utility::gaugeFromStatNames(                                             \
+        scope, {prefix_, stat_names.NAME##_}, Envoy::Stats::Gauge::ImportMode::MODE, tags_))
+#define STATNAME_HELPER_(NAME)
+
+MAKE_STAT_NAMES_STRUCT(CacheStatNames, ALL_CACHE_STATS);
+
 struct CacheStats {
-  ALL_CACHE_STATS(GENERATE_COUNTER_STRUCT, GENERATE_GAUGE_STRUCT)
+  CacheStats(const CacheStatNames& stat_names, Envoy::Stats::Scope& scope,
+             Stats::StatName&& cache_path)
```
jmarantz, line 45: rvalue-ref not necessary for StatName -- it's just a single pointer. Suggest just passing by value.

Think of StatName like a string_view -- its storage needs to be owned somewhere else, but it's cheap to pass them around.

test/common/stats/isolated_store_impl_test.cc
```diff
@@ -189,6 +190,41 @@ TEST_F(StatsIsolatedStoreImplTest, AllWithSymbolTable) {
   EXPECT_EQ(2UL, store_->textReadouts().size());
 }
 
+TEST_F(StatsIsolatedStoreImplTest, CounterWithTag) {
+  StatNameTagVector tags{{makeStatName("tag1"), makeStatName("tag1Value")}};
```
jmarantz, line 194: can you make these tests all include two stats with the same tag-extracted-name but different tags, and make sure they all can be represented?

jmarantz (commented):
Many apologies for how long this has taken to review. Just a few minor issues (assuming that the new tests I suggest work)

/wait

jmarantz:
This is great btw -- these are actually not just for tests; these are used in production also, so fixing the tag behavior is important.

test/common/stats/isolated_store_impl_test.cc
ravenblackx: Done (also in the predecessor change that doesn't touch cache, so if you want to merge them separately we can do that).

source/extensions/http/cache/file_system_http_cache/stats.h
ravenblackx: Done. Thanks for the analogy, that makes it clearer.

source/common/stats/isolated_store_impl.cc
ravenblackx: Done (also in the predecessor PR).

jmarantz (approved):
Thanks!

As this changes core we should get a senior maintainer sign-off. I think this is the superset PR and it's not too big to review, so let's drop the other one.

PR #25354, opened 2023-02-03
Propose a boolean field in ClientStatusRequest to indicate whether xds server should exclude detailed config(xds_config filed in GenericXdsConfig) in response. 

jaychenatr:
Commit Message: 
Propose a boolean field in ClientStatusRequest to indicate whether xds server should exclude detailed config(xds_config filed in GenericXdsConfig) but keep the rest of fields in response.
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/api/envoy/service/status/v3/csds.proto
+++ b/api/envoy/service/status/v3/csds.proto
@@ -86,6 +86,11 @@ message ClientStatusRequest {
 
   // The node making the csds request.
   config.core.v3.Node node = 2;
+
+  // The flag to indicate whether xDS server needs to propagate the detailed
+  // per_xds_config.
+  // [#not-implemented-hide:]
+  bool exclude_detailed_config = 3;
 }
 
 // Detailed config (per xDS) with status.


markdroth:
Please edit the PR subject to describe what this PR is doing.

I don't understand the use-case here.  What is the point of the CSDS response if it does not contain the resource status information?

adisuissa:
Can you provide more background information why is that needed?

jaychenatr:
The main use case is for istioctl to send CSDS to get config status, and we don't need to see the full config but config status would be good enough. 
Mainly for istioctl integration with GSM.

jaychenatr:
> Please edit the PR subject to describe what this PR is doing.
> 
> I don't understand the use-case here. What is the point of the CSDS response if it does not contain the resource status information?

updated the description. Please take a look

markdroth:
Can you please change the PR title, not just the description?  "Signed-off-by: jaychenatr <jiechen@google.com>" is not a useful PR title.

markdroth:
I still don't understand the purpose of this field.  There are basically just two things in the CSDS response, the individual resource information and the node information.  Without populating the individual resource information, the only thing left in the response is the node information.  In what case is it useful for the CSDS client to get the node information?  Is the idea here just to get a list of all clients that are talking to the server?

jaychenatr:
> I still don't understand the purpose of this field. There are basically just two things in the CSDS response, the individual resource information and the node information. Without populating the individual resource information, the only thing left in the response is the node information. In what case is it useful for the CSDS client to get the node information? Is the idea here just to get a list of all clients that are talking to the server?

The proposal is to only exclude **xds_config** filed in GenericXdsConfig, not entire GenericXdsConfig, control plane will still need populate rest of fields in GenericXdsConfig. 
it is used to serve _istioctl proxy-status_, which does not care about the config details but the status/version of the configs. We are expecting _istioctl proxy-status_ will use CSDS to get config status in the near future.

markdroth:
Okay, I got it -- so this isn't going to avoid populating the entire `generic_xds_configs` field, it's only going to avoid populating the `xds_config` field *inside* of the `GenericXdsConfig` message.  That makes sense.

api/envoy/service/status/v3/csds.proto
```diff
@@ -86,6 +86,11 @@ message ClientStatusRequest {
 
   // The node making the csds request.
   config.core.v3.Node node = 2;
+
+  // The flag to indicate whether xDS server needs to propagate the detailed
+  // per_xds_config.
+  // [#not-implemented-hide:]
+  bool exclude_detailed_config = 3;
```
markdroth, line 93: Suggest calling this field `exclude_resource_contents`.

api/envoy/service/status/v3/csds.proto
```diff
@@ -86,6 +86,11 @@ message ClientStatusRequest {
 
   // The node making the csds request.
   config.core.v3.Node node = 2;
+
+  // The flag to indicate whether xDS server needs to propagate the detailed
```
markdroth, line 90: Please change this comment to say something like this:

"""
If true, the server will not include the resource contents in the response (i.e., the generic_xds_configs.xds_config field will not be populated).
"""

The description may need to get more complicated if we also need to support the deprecated `PerXdsConfig` message, because it affects a number of different fields there.

Issue #12651, opened 2020-08-14
quic: server preferred address


ggreenway:
The QUIC spec supports the server specifying a different address for the client to communicate with it (https://github.com/quicwg/base-drafts/blob/master/draft-ietf-quic-transport.md#servers-preferred-address-preferred-address):

> QUIC allows servers to accept connections on one IP address and attempt to transfer these connections to a more preferred address shortly after the handshake. This is particularly useful when clients initially connect to an address shared by multiple servers but would prefer to use a unicast address to ensure connection stability. This section describes the protocol for migrating a connection to a preferred server address.

This can be used when Envoy is behind an L4 load balancer, so ensure that connections continue to flow to the same node even if the client migrates to a different address. The initial handshake packets would flow through the L4 load balancer, but after that, the client would communicate directly to Envoy on it's preferred address, which would be an address dedicated to that instance (not a shared load balancer address).

Steps needed to implement this in Envoy:

* QUIC listeners would need to have a socket bound to each preferred address, in addition to the address specified in the listener. Packets sent to any of the addresses need to go through the same filter chain.
* Egress packets need to go through the original address to start, and can only start flowing through the preferred address after a non-probing packet is received on the new address. From the spec: 
> "The server MUST send non-probing packets from its original address until it receives a non-probing packet from the client at its preferred address and until the server has validated the new path."

ggreenway:
I think this requires #11184.

moderation:
Updated link to the relevant sections in the IETF standard is https://datatracker.ietf.org/doc/html/rfc9000#section-9.6

ggreenway:
This is on my TODO list, but has been too low on the priority list for me to make a lot of progress. If anyone else is looking at this, ping me so we can sync up.

ggreenway:
This is implemented already, with the restriction that the listener must be able to accept packets on both the initial and SPA address.

PR #25356, opened 2023-02-04
quic: support half-way server's preferred address

danzh2010:
Signed-off-by: Dan Zhang <danzh@google.com>

Commit Message: add an extension point to specify server's preferred addresses (SPA) [RFC9000](https://datatracker.ietf.org/doc/html/rfc9000#name-servers-preferred-address) at startup. If configured, QUICHE will advertise preferred v4 and v6 addresses to QUICHE-based clients. Even if the client switches to SPA, the Envoy will continue send responses back via the original server address.

Additional Description: this implementation allows clients to switch to SPA, but Envoy will not probe from the preferred address to the client nor change its own egress source address. This is sufficient for latency improvement on the client to server path. And this behavior is subject to change while QUICHE implementation of the server's preferred address becomes complete.

This change adds an extension implementation to allow Envoy to  listen on `0.0.0.0` or `::` while advertising a SPA of a different network interface with the same port. But handling listening on multiple sockets is NOT part of this change. 

Risk Level: low, new behavior behind config knob
Testing: new integration test
Docs Changes: N/A
Release Notes: N/A
Part of #12651


lizan (commented):
/lgtm api

ggreenway:
Excited to see progress on this!

Quick initial question: do we need an extension point for SPA, or could we just have a configuration of ipv4 address and ipv6 address? Any use cases in mind for it being extensible?

ggreenway:
> > Quick initial question: do we need an extension point for SPA, or could we just have a configuration of ipv4 address and ipv6 address? Any use cases in mind for it being extensible?
> 
> Yes, in case of direct server return we want the ability to grab the Envoy machine address actually serving the request even when the config pipeline doesn't know about the address of each machine.

Ok, sounds reasonable. For both simple cases and for tests, can you create an implementation that configures them statically?

danzh2010:
> 



> > > Quick initial question: do we need an extension point for SPA, or could we just have a configuration of ipv4 address and ipv6 address? Any use cases in mind for it being extensible?
> > 
> > 
> > Yes, in case of direct server return we want the ability to grab the Envoy machine address actually serving the request even when the config pipeline doesn't know about the address of each machine.
> 
> Ok, sounds reasonable. For both simple cases and for tests, can you create an implementation that configures them statically?

Added a test extension to read v4 and v6 preferred addresses directly from config.

api/envoy/config/listener/v3/quic_config.proto
```diff
@@ -68,4 +68,10 @@ message QuicProtocolOptions {
   // If not specified the :ref:`default one configured by <envoy_v3_api_msg_extensions.quic.connection_id_generator.v3.DeterministicConnectionIdGeneratorConfig>` will be used.
   // [#extension-category: envoy.quic.connection_id_generator]
   core.v3.TypedExtensionConfig connection_id_generator_config = 8;
+
+  // Config how to get the server's preferred address to advertise and allow client migrating to it.
```
RyanTheOptimist, line 72: nit: "Configure the server's preferred address to advertise so that client can migrate to it"

api/envoy/extensions/quic/server_preferred_address/v3/empty_server_preferred_address_config.proto
```diff
@@ -0,0 +1,18 @@
...
+option java_outer_classname = "EmptyServerPreferredAddressConfigProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/quic/server_preferred_address/v3;server_preferred_addressv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: QUIC server preferred address config]
+// [#extension: envoy.quic.server_preferred_address.empty]
+
+// Configuration for a server's preferred address config implementation which doesn't provide any preferred addresses.
+message EmptyServerPreferredAddressConfig {
```
RyanTheOptimist, line 17: It seems like this matches the default behavior. is that right?

source/common/quic/envoy_quic_server_preferred_address_config_factory.h
```diff
@@ -0,0 +1,41 @@
...
+#include "envoy/config/typed_config.h"
+#include "envoy/network/address.h"
+#include "envoy/protobuf/message_validator.h"
+
+#include "quiche/quic/platform/api/quic_socket_address.h"
+
+namespace Envoy {
+namespace Quic {
+
+class EnvoyQuicServerPreferredAddressConfig {
```
RyanTheOptimist, line 12: nit: Brief comment, please

source/common/quic/envoy_quic_server_preferred_address_config_factory.h
```diff
@@ -0,0 +1,41 @@
...
+public:
+  virtual ~EnvoyQuicServerPreferredAddressConfig() = default;
+
+  /**
+   * @param local_address the configured default listening address.
+   * Returns a pair of the server preferred addresses in form of {IPv4, IPv6}.
```
RyanTheOptimist, line 18: What happens if the server only supports v4 or only v6? Is one of the addresses empty?

source/common/quic/envoy_quic_server_session.h
```diff
@@ -17,6 +17,14 @@
 namespace Envoy {
 namespace Quic {
 
+#define QUIC_CONNECTION_STATS(COUNTER)                                                             \
```
RyanTheOptimist, line 20: This seems fine to me but I wonder if we should loop in @jmarantz for a double-check?

source/common/quic/envoy_quic_server_session.cc
```diff
@@ -205,6 +206,9 @@ void EnvoyQuicServerSession::ProcessUdpPacket(const quic::QuicSocketAddress& sel
     maybeHandleCloseDuringInitialize();
   }
   quic::QuicServerSessionBase::ProcessUdpPacket(self_address, peer_address, packet);
+  if (self_address != connection()->self_address()) {
```
RyanTheOptimist, line 209: Should we instead check that it's equal to the preferred address instead of that it's not self_address? (Or is that hard to do?)

test/integration/testing_server_preferred_address_config.h
```diff
@@ -0,0 +1,55 @@
...
+      : ip_v4_(ipv4), ip_v6_(ipv6) {}
+
+  std::pair<quic::QuicSocketAddress, quic::QuicSocketAddress> getServerPreferredAddresses(
+      const Network::Address::InstanceConstSharedPtr& local_address) override {
+    int32_t port = local_address->ip()->port();
+    return {quic::QuicSocketAddress(ip_v4_, port), quic::QuicSocketAddress(ip_v6_, port)};
+  }
+
+private:
+  quic::QuicIpAddress ip_v4_;
```
RyanTheOptimist, line 21: nit: const?

RyanTheOptimist (commented):
Exciting!

source/common/quic/envoy_quic_server_session.cc
RyanTheOptimist: I don't feel strongly about this. Happy to see it land as-is.

RyanTheOptimist (dismissed):

RyanTheOptimist:
@jmarantz for the additional stats

test/integration/testing_server_preferred_address_config.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package test.integration;
+
+// [#protodoc-title: QUIC server preferred address config for test]
+// [#extension: envoy.quic.server_preferred_address.test]
+
+// Configuration for TestingServerPreferredAddressConfig.
+// [#next-free-field: 3]
+message TestingServerPreferredAddressConfig {
```
ggreenway, line 10: I think something like this should be added to non-testing for the basic use case where the address is known when config is generated. This could be moved, and then a separate for-testing implementation won't be needed.

ggreenway (changes requested):
/wait

source/common/quic/envoy_quic_server_preferred_address_config_factory.h
```diff
@@ -0,0 +1,43 @@
...
+class EnvoyQuicServerPreferredAddressConfig {
+public:
+  virtual ~EnvoyQuicServerPreferredAddressConfig() = default;
+
+  /**
+   * @param local_address the configured default listening address.
+   * Returns a pair of the server preferred addresses in form of {IPv4, IPv6}. An uninitialized
+   * address value means no preferred address for that address family.
+   */
+  virtual std::pair<quic::QuicSocketAddress, quic::QuicSocketAddress>
```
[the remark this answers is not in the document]
danzh2010: IPv4  and IPv6 interfaces doesn't include port. Some extension implementation might want to specify a different port from the original port. Though our UDP listener might not support listening on multiple sockets yet, but I would think we want port to be part of this interface.

api/envoy/extensions/quic/server_preferred_address/v3/empty_server_preferred_address_config.proto
danzh2010: That's my initial approach, but a new extension category without an extension implementation failed one of the rst document generation step and I had to add an extension referencing the new category in source/extensions/extensions_metadata.yaml.

source/common/quic/envoy_quic_server_preferred_address_config_factory.h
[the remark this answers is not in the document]
ggreenway: Oh right; that makes sense. I agree that alternate port should be supported. I'm fine leaving it as-is then.

api/envoy/extensions/quic/server_preferred_address/v3/empty_server_preferred_address_config.proto
danzh2010: That being said, if I convert the testing extension into non-testing one, the doc generation won't complain any more. And I should be able to leave the extension unset if not configured. Let me try this first.

source/common/quic/envoy_quic_server_preferred_address_config_factory.h
```diff
@@ -0,0 +1,43 @@
...
+
+// An interface to provide the server's preferred addresses at runtime.
+class EnvoyQuicServerPreferredAddressConfig {
+public:
+  virtual ~EnvoyQuicServerPreferredAddressConfig() = default;
+
+  /**
+   * @param local_address the configured default listening address.
+   * Returns a pair of the server preferred addresses in form of {IPv4, IPv6}. An uninitialized
+   * address value means no preferred address for that address family.
```
ggreenway, line 20: Can you add to the comment when this is called? eg at config-load time vs once-per-connection?

danzh2010: done

danzh2010:
@ggreenway PTAL!

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
```diff
@@ -0,0 +1,23 @@
...
+option java_outer_classname = "BasicServerPreferredAddressConfigProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/quic/server_preferred_address/v3;server_preferred_addressv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: QUIC server preferred address config]
+// [#extension: envoy.quic.server_preferred_address.basic]
+
+// Configuration for BasicServerPreferredAddressConfig.
+message BasicServerPreferredAddressConfig {
```
ggreenway, line 17:
```suggestion
message FixedServerPreferredAddressConfig {
```

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
```diff
@@ -0,0 +1,23 @@
...
+  // String representation of ipv(4|6) address, i.e. "127.0.0.2".
+  // If not specified, none will be configured.
+  string ipv4_address = 1;
```
ggreenway, line 20: Since the port of part of the SPA, should this be `message SocketAddress` instead, so that the port can also be configured? Unset/zero could still mean use the listener's port.

source/extensions/quic/server_preferred_address/basic_server_preferred_address_config.cc
```diff
@@ -0,0 +1,39 @@
...
+
+Quic::EnvoyQuicServerPreferredAddressConfigPtr
+BasicServerPreferredAddressConfigFactory::createServerPreferredAddressConfig(
+    const Protobuf::Message& message, ProtobufMessage::ValidationVisitor& validation_visitor) {
+  auto& config =
+      MessageUtil::downcastAndValidate<const envoy::extensions::quic::server_preferred_address::v3::
+                                           BasicServerPreferredAddressConfig&>(message,
+                                                                               validation_visitor);
+  quic::QuicIpAddress ip_v4, ip_v6;
+  const std::string err("bad server preferred address");
```
ggreenway, line 21: Please specify whether it's the ipv4 or ipv6 address in the exception

ggreenway (changes requested):
/wait

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
danzh2010: SocketAddress also has protocol and port fields which are not very meaningful to this extension. And it also just uses a string field to represent ip.

danzh2010: Tag @envoyproxy/api-shepherds on this.

ggreenway: @envoyproxy/api-shepherds we were discussing this in slack, and wanted to get some more opinions. Right now, due to a single UDP socket (on 0.0.0.0) the port for the main listener and SPA must be the same.

In the future, we may support multiple "linked" sockets for a quic listener for SPA, and then could support different ports for the sockets. What is the best way to allow changing this config in the future to support that? Options we talked about:

1) Create a new extension type for that when it's supported. Leave this config exactly as it currently is.
2) Wrap this config in `oneof` so we can later add `SocketAddress`
3) Wrap this config in `oneof` so we can later add a message with ip string and port int.

htuch: From a strict forward compatibility perspective:
* Use a `repeated` now if you anticipate wanting multiple later (is this how linked sockets would work), since it's a breaking JSON wire change to switch to repeated later (but not proto3 wire).
* `oneof` can always be added later I believe, it doesn't have any wire implication.

Do you want any of the resolver extension support in `SocketAddress`? FWIW we've found this useful in a number of internal use cases at Google, where the Envoy maps from some name format to the underlying IP. This includes also named port resolution.

source/common/quic/envoy_quic_server_session.h
jmarantz: Mostly: can we throw these names into QuicStatNames so we don't need to take a global symbol table lock whenever an EnvoyQuicServerSession is created.

I'm not sure how often that happens, but it sounds like it happens on the data-plane so it'd be better not to take what's effectively a singleton  lock at that point.

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
danzh2010: I don't think we need repeated field here, QUIC only allows configuring 2 SPA, one v4 and one v6.

As to resolver extension support in `SocketAddress`, it hard to tell if we need it or not. The only internal use case we have today will have its own implementation of SPA extension to grab the addresses at runtime instead of statically config them.

ggreenway: > * `oneof` can always be added later I believe, it doesn't have any wire implication.

From API_VERSIONING.md, this is disallowed:
```
  * Wrapping an existing field with `oneof`. This has no protobuf or JSON/YAML wire implications,
    but is disruptive to various consuming stubs in languages such as Go, creating unnecessary
    churn.
```

danzh2010: How about note this proto with `option (xds.annotations.v3.file_status).work_in_progress = true;` for now? Since we don't have a clear picture of how it is going to be used and what the implementation should be.

source/common/quic/envoy_quic_server_session.h
danzh2010: I'm confused about how to add stats. Is macros like `MY_COOL_STATS(COUNTER, GAUGE, HISTOGRAM)` still the common way of doing that? QuicStatNames currently has stats for QUIC errors which are big enums and we want to lazy-initialize them. But the stats here only has a few counters.

If the concern is initializing this stats once per connection, is moving it up to be a listener stats more reasonable? Since other than the prefix is "quic.connection", there is no per-connection information in the stats name.

danzh2010: I moved the creation of QuicConnectionStats into EnvoyQuicDispatcher constructor. PTAL!

source/extensions/quic/server_preferred_address/basic_server_preferred_address_config.cc
danzh2010: done

source/common/quic/envoy_quic_server_session.h
jmarantz: Thanks; looks good! I'm assuming here that (a) EnvoyQuicServerSession is only created on startup or config update and (b) the new stats are useful enough you want to instantiate them up front, rather than on demand.

If so, feel free to resolve; lgtm. I didn't look at the rest of the PR.

danzh2010: > I'm assuming here that (a) EnvoyQuicServerSession is only created on startup or config update 

EnvoyQuicServerSession is not. But I think you meant EnvoyQuicDispatcher which instantiates those stats now. And yes, it is only created on startup or LDS update.

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
danzh2010: How shall I move forward on the config API?

phlax:
@danzh1989 is this ready for further review ?

danzh2010:
> @danzh1989 is this ready for further review ?

We are still trying to settle the config API.  But as reviewers are likely away this week, I don't know how much the discussion could move forward. If you have other comments or suggestions, please do not be block on the on-going discussion.

phlax:
np - i was just clearing the stalled PR queue

source/common/quic/envoy_quic_server_session.cc
danzh2010: Given enough time to iterates on this PR, my QUICHE change to expose the SPA from QuicConnection gets into Envoy now. I changed this comparison to be equal to the preferred address.

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
danzh2010: done.

api/envoy/extensions/quic/server_preferred_address/v3/basic_server_preferred_address_config.proto
danzh2010: In fact, I marked the whole extension point of server_preferred_address_config as WIP. We haven't used this feature in production yet and not sure if it will work as expected on the internet. So it's better not to commit to stable API.

htuch: It's fine if you want to mark it WiP for now, but please add a comment to the effect of "TODO: discuss with API shepherds before removing WiP status".

htuch: I'd also just "wrap with oneof" for now, @ggreenway is right about the rules.

danzh2010: done

htuch (commented):
/lgtm api

danzh2010:
Thanks for the review! Mind merging it?

PR #25361, opened 2023-02-05
ci: Only run mobile jobs in the Envoy repo

phlax:

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -12,6 +12,7 @@ concurrency:
 
 jobs:
   androidbuild:
+    if: github.repository == 'envoyproxy/envoy'
     name: android_build
     runs-on: macos-12
     timeout-minutes: 90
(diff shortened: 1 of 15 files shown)


phlax:
cc @jpsim this is fairly urgent as mirror repos are trying to run all of the mobile tests and failing

the solution is far from ideal - having to exclude at the job level - after trawling through the docs and design decisions here i have come to the firm conclusion that github/ms' intention is to trigger as much unwanted CI as they can get away with

jpsim (approved):
+1 to unblock, however we should consider doing the same for the other GitHub Actions CI jobs like CodeQL.

Also, do we have something similar in place for the Azure Pipelines CI?

phlax:
i think we do this already for ~all of the existing gh jobs - in the case of azp we want them to run in current mirror

phlax:
> ...  in the case of azp we want them to run in current mirror

so actually this is slightly more complex - azp has to be setup for a repo for it to trigger at all - so i think the file is harmless if that is not set up, and ~needed if it is

fwiw, triggers in the file are mostly ignored in favour of web config even where it is set up

PR #25374, opened 2023-02-06
examples: Make skywalking/elasticsearch less greedy

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/examples/skywalking/docker-compose.yaml
+++ b/examples/skywalking/docker-compose.yaml
@@ -78,12 +78,11 @@ services:
       discovery.type: single-node
       ingest.geoip.downloader.enabled: "false"
 
-      # This allows ES to run in a low-resource environment, dont do this at home!
-      cluster.routing.allocation.disk.watermark.low: "99%"
-      cluster.routing.allocation.disk.watermark.high: "99%"
-      cluster.routing.allocation.disk.watermark.flood_stage: "99%"
-      cluster.routing.allocation.disk.watermark.flood_stage.frozen: "99%"
-      cluster.routing.allocation.disk.watermark.flood_stage.frozen.max_headroom: "1GB"
+      cluster.routing.allocation.disk.watermark.low: "${ES_WATERMARK_LOW:-85%}"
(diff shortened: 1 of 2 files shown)


phlax:
cc @wangfakang

examples/skywalking/docker-compose.yaml
```diff
@@ -77,6 +77,14 @@ services:
     environment:
       discovery.type: single-node
       ingest.geoip.downloader.enabled: "false"
+
+      # This allows ES to run in a low-resource environment, dont do this at home!
```
adisuissa, line 81: Is this needed just for our CI execution? If so, it may make sense to add a guard here to set these vars when run under our-CI.
This is just to avoid people copy-pasting it to their docker-compose scripts or copying the env vars from that docker.

adisuissa (commented):
LGTM, just minor suggestion.
If it doesn't make sense, LMK, and I'll approve it.

examples/skywalking/docker-compose.yaml
phlax: hmm, yep - i have those messages there for that reason - but dont really want to add this - its not for production

having said that i think its safe here - and probably a good idea - the sandbox was refusing to start because it only had ~5GB spare which is more than enough to run this

perhaps we can keep and improve the "dont do this at home" message

phlax: trying to think if we can use an env var in ci and set a default here - ill test more tomorrow

phlax:
testing this in #25399 - is seems like this is actually required - just testing further to see exactly what is required and use env vars

examples/skywalking/docker-compose.yaml
phlax: updated

adisuissa (approved):
LGTM, thanks!

PR #25378, opened 2023-02-06
changelogs: Merge normally

phlax:
currently changelogs do a union merge, and this has led to a lot of bad merges, and duplicate changelogs

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,3 @@
-/changelogs/*.yaml merge=union
 /generated_api_shadow/envoy/** linguist-generated=true
 /generated_api_shadow/bazel/** linguist-generated=true
 *.svg binary


wbpcode:
Are there some simple screenshots to show the optimization he final conent? Then we can merge it.

phlax:
> Are there some simple screenshots to show the optimization he final conent? Then we can merge it.

not sure i follow really

the issue is that currently given a conflict it will take everything from both sides and add them together

this was more targetted at the way the changelogs worked previously, but either way it creates a continual headache because the merges are often incorrect

switching it, as in here, forces the developer to resolve the conflict manually, as they do for all other conflicts - so in a sense there is a small loss of convenience in the cases were it would have worked correctly

otoh, it breaks so often, and mostly when you are trying to release and it matters the most, in an undectable way without checking the end result carefully, that i really think it makes sense to just switch this

phlax:
another example of where this is problematic is cherry-picking across a branch - it will automatically take _all_ of the changes from the other branch - which is never what you want - you want to pick the commit as just the actual change that it makes

wbpcode:
hah, I get it. Thanks.

wbpcode (approved):
Thanks.

Issue #25355, opened 2023-02-03
mobile: compile out unavailable builder functions


jpsim:
*Title*: *Unavailable engine builder functions should be compiled out*

*Description*:

Currently it's possible to compile out HTTP/3 and admin interface support from Envoy Mobile, however the engine builder APIs are still available, which could possibly lead to crashes if those features are attempted to be used.

Instead, we should compile out those builder functions, making it harder to misuse.

*Relevant Links*: Discussion in https://github.com/envoyproxy/envoy/pull/25352#issuecomment-1416502976

wbpcode:
cc @alyssawilk

PR #25381, opened 2023-02-06
mobile: compile out unavailable engine builder functions

jpsim:
Builder functions won't be available or even compiled if the corresponding flags aren't defined:

* `ENVOY_ENABLE_QUIC`
* `ENVOY_ADMIN_FUNCTIONALITY`
* `ENVOY_MOBILE_REQUEST_COMPRESSION`

Commit Message: mobile: compile out unavailable engine builder functions
Additional Description:
Risk Level: Low
Testing: Used `nm` to confirm that the symbols were being included with the build flags on, and excluded with the build flags off.
Docs Changes: None
Release Notes: None
Fixes: https://github.com/envoyproxy/envoy/issues/25355


alyssawilk (commented):
LGTM but I think this also needs merge fixes with the Java builder PR.

jpsim:
> LGTM but I think this also needs merge fixes with the Java builder PR.

I don't think so but will merge `main` into this branch to be sure.

alyssawilk:
Yeah I think we just need the same #define guards around the calls in jni_interface.cc

jpsim:
> Yeah I think we just need the same #define guards around the calls in jni_interface.cc

Ah yes, of course. Adding those in now.

.github/workflows/android_build.yml
```diff
@@ -219,6 +219,7 @@ jobs:
             $([ -z $GITHUB_TOKEN ] || echo "--config=remote-ci-macos") \
             --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
             --fat_apk_cpu=x86_64 \
+            --define=admin_functionality=enabled \
```
alyssawilk, line 222: Sorry all the code looks good but I'm confused why this is needed given we didn't change the default anywhere?

jpsim: You're right that this isn't needed, this was me trying to fix this error in the Kotlin experimental app:

```
test/kotlin/apps/experimental/MainActivity.kt:58:8: error: unresolved reference: enableAdminInterface
      .enableAdminInterface()
       ^
```

I think the way I'm trying to compile out the Kotlin builder functions isn't working correctly.

jpsim: I figured it out. Reverted this change.

mobile/library/common/jni/BUILD
```diff
@@ -53,10 +53,7 @@ cc_library(
     srcs = [
         "jni_interface.cc",
     ],
-    copts = envoy_select_envoy_mobile_request_compression(
-        ["-DENVOY_MOBILE_REQUEST_COMPRESSION"],
-        "@envoy",
-    ),
+    copts = envoy_mobile_copts("@envoy"),
```
alyssawilk, line 56: can we just move this over to envoy_cc_library?  If not maybe comment why?

mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilderAdminUtil.kt
```diff
@@ -0,0 +1,21 @@
+package io.envoyproxy.envoymobile
+
+/**
+ * Utility to enable admin interface.
+ */
+object EngineBuilderAdminUtil {
```
alyssawilk, line 6: optional, what do you think of just removing these interfaces for kotlin? AFIK we planned on removing overall but are blocked for swift because no stats, but we could remove from Java/C++ now and avoid this extra class.

alyssawilk: that said, these classes working vs not for the compiled in options are pretty darned amazing.  Nicely done!

bazel/envoy_mobile_copts.bzl
```diff
@@ -0,0 +1,8 @@
+# DO NOT LOAD THIS FILE. Load envoy_build_system.bzl instead.
+load(":envoy_select.bzl", "envoy_select_admin_functionality", "envoy_select_enable_http3", "envoy_select_envoy_mobile_request_compression")
+
+# Compute the copts for Envoy mobile.
```
alyssawilk, line 4: we have a lot of copts for mobile which aren't represented here.  Can you comment which ones matter? is it just ones for APIs which are compiled in and out or should we actually replicate more of the select/defines that Envoy has?

alyssawilk (commented):
Holy awesome - this is going to prevent so many mistakes :-)

can you also merge main to make sure this plays well with the no admin default builds - just unsure if we have to enable admin in any more CI.

mobile/library/common/jni/BUILD
jpsim: I'll try moving it over. I _think_ it should be safe.

mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilderAdminUtil.kt
jpsim: I know Rafal cares about keeping admin functionality available with build flags, so I'd rather not rip that out while he's out 😅

bazel/envoy_mobile_copts.bzl
jpsim: Oh sorry, we need to plumb these through to `objc_library` and `swift_library` so since we need this in 3 places, I think it's best to keep centralized.

mobile/library/kotlin/io/envoyproxy/envoymobile/EngineBuilderAdminUtil.kt
alyssawilk: ah totally fine - I thought he only cared about it for iOS.

alyssawilk (dismissed):

alyssawilk:
oh.... well  I thought it looked good but CI disagrees :-/

jpsim:
> oh.... well I thought it looked good but CI disagrees :-/

Added a TODO to investigate migrating these over to envoy_cc_library.

alyssawilk (dismissed):

alyssawilk (approved):

PR #25386, opened 2023-02-06
Make isolated_stats_cache support tags correctly

ravenblackx:
Commit Message: Make isolated_stats_cache support tags correctly
Additional Description: Attempting to test stat names and tags when a metric is created with populated tags results in failure, as isolated_stats_cache converts `a.b` with tags `{c:d}` into `a.b.c.d` in both stat_name_ and tags_extracted_stat_name_, along with an empty tags member. This fix should make it behave the same as other stat stores.
Risk Level: Unknown. It appears `IsolatedStoreImpl` is used in some core server features, so care may need to be taken.
Testing: Passes existing tests; added tests for the behavior with tags.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/source/common/stats/isolated_store_impl.cc
+++ b/source/common/stats/isolated_store_impl.cc
@@ -19,17 +19,28 @@ IsolatedStoreImpl::IsolatedStoreImpl(std::unique_ptr<SymbolTable>&& symbol_table
 }
 
 IsolatedStoreImpl::IsolatedStoreImpl(SymbolTable& symbol_table)
-    : alloc_(symbol_table), counters_([this](StatName name) -> CounterSharedPtr {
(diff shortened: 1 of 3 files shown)


jmarantz (commented):
This PR is a subset of the other one I already commented on, right? Can we just work on that one since I already started to comment on it?

/wait

ravenblackx:
If you want I can just close this one and we can use the other one. I was trying to keep "changes to envoy core" and "changes to the cache" as separate PRs (and keep review-chunks small!)

Or now that you're aware that it's separated that way, I've addressed the comments from the other one (in both tracks), so you could split the review. Up to you. :)

jmarantz:
closing this as it's a subset. LMK if this was the wrong thing to do

PR #25389, opened 2023-02-06
mobile: fixing a bug for junit config

alyssawilk:
Prior builder PR checked the filter chain didn't change, but in doing so changed the filter chain.  Thankfully only under junit but still buggy.

### DIFF
--- a/mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -322,12 +322,12 @@ public class EnvoyConfiguration {
 
     for (StackTraceElement element : Thread.currentThread().getStackTrace()) {
       if (element.getClassName().startsWith("org.junit.")) {
-        List<EnvoyNativeFilterConfig> reverseFilterChain = nativeFilterChain;
+        List<EnvoyNativeFilterConfig> reverseFilterChain = new ArrayList<>(nativeFilterChain);
         Collections.reverse(reverseFilterChain);
         Boolean enforceTrustChainVerification =
             trustChainVerification == EnvoyConfiguration.TrustChainVerification.VERIFY_TRUST_CHAIN;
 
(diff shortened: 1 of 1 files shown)


jpsim (approved):
Looks correct, good catch

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -322,12 +322,12 @@ String createYaml() {
...
     for (StackTraceElement element : Thread.currentThread().getStackTrace()) {
       if (element.getClassName().startsWith("org.junit.")) {
-        List<EnvoyNativeFilterConfig> reverseFilterChain = nativeFilterChain;
+        List<EnvoyNativeFilterConfig> reverseFilterChain = new ArrayList<>(nativeFilterChain);
         Collections.reverse(reverseFilterChain);
         Boolean enforceTrustChainVerification =
             trustChainVerification == EnvoyConfiguration.TrustChainVerification.VERIFY_TRUST_CHAIN;
 
-        byte[][] filter_chain = JniBridgeUtility.toJniBytes(nativeFilterChain);
+        byte[][] filter_chain = JniBridgeUtility.toJniBytes(reverseFilterChain);
```
RyanTheOptimist, line 330: drive-by: This looks like a prod-bug not a test-bug. However, the PR description mentions "junit" which (I think) is a test thing. Should we tweak the PR description? (Should we have a test for this)

alyssawilk: this is inside 

for (StackTraceElement element : Thread.currentThread().getStackTrace()) {
      if (element.getClassName().startsWith("org.junit.")) {

we're about to have a test as it broke something in the "use the bootstrap from Java" PR but I can land a separate one as well as a follow-up

RyanTheOptimist: Oooooh! Sorry, good point!

PR #25392, opened 2023-02-06
java: switching to using the C++ builder

alyssawilk:
Deferring to JP on how we want to land this (off by default, one fell swoop) but defaulting on to CI-validate (especially as unlike C++ I don't know how to benchmark Java)

Switching the Java Envoy engine to create configuration via the C++ builder's bootstrap fast-path rather than in-Java YAML concatanation.

Risk Level: meidum
Testing: yep
Docs Changes: n/a
Release Notes: inline.
Part of  https://github.com/envoyproxy/envoy/issues/24976


RyanTheOptimist:
Looks like this needs a main merge.

mobile/library/common/jni/jni_interface.cc
```diff
@@ -120,10 +121,24 @@ extern "C" JNIEXPORT jlong JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibr
 }
 
 extern "C" JNIEXPORT jint JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibrary_runEngine(
-    JNIEnv* env, jclass, jlong engine, jstring config, jstring log_level) {
+    JNIEnv* env, jclass, jlong engine, jstring config, jlong native_bootstrap, jstring log_level) {
```
RyanTheOptimist, line 124: I know we don't seem to have comments on the JNICALL functions in this file, but I wonder if we should start. In particular, it would be good to comment on what "native_bootstrap" is (it's an optional pointer to a Bootstrap proto). It would also be good to explain what happens if it is null?

Also, I think @jpsim got rid of the "native_" prefix from fields in his recent PR. Do we have a convention for when that prefix is appropriate?

mobile/library/common/jni/jni_interface.cc
```diff
@@ -120,10 +121,24 @@ extern "C" JNIEXPORT jlong JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibr
...
+
+  jint result;
+  if (!bootstrap) {
+    result = run_engine(engine, string_config, string_log_level, "");
+  } else {
+    auto options = std::make_unique<Envoy::OptionsImpl>();
+    options->setConfigProto(std::move(bootstrap));
+    options->setLogLevel(options->parseAndValidateLogLevel(string_log_level));
+    options->setConcurrency(1);
+    auto cast_engine = reinterpret_cast<Envoy::Engine*>(engine);
```
RyanTheOptimist, line 138: How 'bout just:
```
reinterpret_cast<Envoy::Engine*>(engine)->run(std::move(options));
````
(To avoid the unnecessary local variable)

File: mobile/library/common/jni/jni_interface.cc [hunk 3]
```diff
@@ -1196,25 +1211,25 @@ javaObjectArrayToStringPairVector(JNIEnv* env, jobjectArray entries) {
...
-    jlong per_try_idle_timeout_seconds, jstring app_version, jstring app_id,
-    jboolean trust_chain_verification, jobjectArray virtual_clusters, jobjectArray filter_chain,
-    jobjectArray stat_sinks, jboolean enable_platform_certificates_validation,
-    jboolean enable_skip_dns_lookup_for_proxied_requests) {
-  Envoy::Platform::EngineBuilder builder;
-
+void configureBuilder(JNIEnv* env, jstring grpc_stats_domain, jboolean admin_interface_enabled,
+                      jlong connect_timeout_seconds, jlong dns_refresh_seconds,
+                      jlong dns_failure_refresh_seconds_base, jlong dns_failure_refresh_seconds_max,
+                      jlong dns_query_timeout_seconds, jlong dns_min_refresh_seconds,
```
RyanTheOptimist, line 1217: Boy, github sure seems confused about what changed here, for no good reason. /me shakes fist at github.

mobile/library/common/jni/jni_interface.cc
```diff
@@ -1196,25 +1211,25 @@ javaObjectArrayToStringPairVector(JNIEnv* env, jobjectArray entries) {
...
-    jboolean enable_interface_binding, jlong h2_connection_keepalive_idle_interval_milliseconds,
-    jlong h2_connection_keepalive_timeout_seconds, jlong max_connections_per_host,
-    jlong stats_flush_seconds, jlong stream_idle_timeout_seconds,
-    jlong per_try_idle_timeout_seconds, jstring app_version, jstring app_id,
-    jboolean trust_chain_verification, jobjectArray virtual_clusters, jobjectArray filter_chain,
-    jobjectArray stat_sinks, jboolean enable_platform_certificates_validation,
-    jboolean enable_skip_dns_lookup_for_proxied_requests) {
-  Envoy::Platform::EngineBuilder builder;
-
+void configureBuilder(JNIEnv* env, jstring grpc_stats_domain, jboolean admin_interface_enabled,
```
RyanTheOptimist, line 1214: Should this move to an anon namespace now that it's a helper?

mobile/library/common/jni/jni_interface.cc
```diff
@@ -1271,12 +1284,80 @@ extern "C" JNIEXPORT jstring JNICALL Java_io_envoyproxy_envoymobile_engine_JniLi
...
+                   enable_skip_dns_lookup_for_proxied_requests, builder);
+
+  const char* native_yaml = env->GetStringUTFChars(yaml, nullptr);
   builder.generateBootstrapAndCompareForTests(native_yaml);
   env->ReleaseStringUTFChars(yaml, native_yaml);
+}
 
-  jstring result = env->NewStringUTF(config_template);
-  return result;
+extern "C" JNIEXPORT jlong JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibrary_createBootstrap(
```
RyanTheOptimist, line 1327: I think it would also be good to comment on the return value here.

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -213,6 +213,7 @@ String createYaml() {
     final StringBuilder customFiltersBuilder = new StringBuilder();
 
     for (EnvoyNativeFilterConfig filter : nativeFilterChain) {
+      System.err.println("Adding native filter " + filter.name);
```
RyanTheOptimist, line 216: Remove?

RyanTheOptimist (commented):
Looking great! A few drive-by comments.

mobile/library/common/jni/jni_interface.cc
alyssawilk: documented in the JNILibrary butI forgot to update.
Also commented here and referenced that.

mobile/library/common/jni/jni_interface.cc
alyssawilk: I'd prefer to leave as-is as I'm hopeful we'll migrate swift, remove the yaml function and merge into one function again and it does and will keep diffs cleaner

RyanTheOptimist: Oh, sure. Good point.

RyanTheOptimist (approved):

jpsim:
I'd like to get what's currently on `main` up and running in Lyft's Android apps and passing all checks before we merge this.

I'm actively debugging that, but I don't yet have an ETA. Hopefully this week.

mobile/library/common/jni/jni_interface.cc
jpsim: I don't have a strong preference on the naming here, as long as we stay consistent with the Java interface to make it easier to produce the correct matching JNI.

jpsim (commented):
Defaulting to using `createBootstrap` makes sense to me, although I'd really like this to be configurable with the engine builder.

alyssawilk:
sorry not quite clear what the ask is. 
Do you want EnvoyConfiguration to take a boolean the EnvoyEngineImpl checks for which mode?

jpsim:
> Do you want EnvoyConfiguration to take a boolean the EnvoyEngineImpl checks for which mode?

Yes that's right. Something like `builder.useExperimentalConfig(true)` or similar.

jpsim (approved):

mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -195,7 +196,18 @@ public EnvoyConfiguration(
     this.statSinks = statSinks;
     this.enablePlatformCertificatesValidation = enablePlatformCertificatesValidation;
     this.enableSkipDNSLookupForProxiedRequests = enableSkipDNSLookupForProxiedRequests;
+    this.useLegacyBuilder = false;
   }
+
+  /**
+   * Sets the mode the Engine builder will use for config genreation
```
jpsim, line 203:
```suggestion
   * Sets the mode the Engine builder will use for config generation
```

jpsim (approved):

Issue #24485, opened 2022-12-12
Clang-tidy can fail when run against existing code


phlax:
Mostly when we run `clang-tidy` we use `clang-tidy-diff` which only runs against changed code.

If its run against all code (or triggered on certain code) it can fail with issues finding linked in libs

phlax:
cc @RyanTheOptimist 

related #24297

phlax:
i have tracked down the issue/s with clang-tidy - there are several

firstly we effectively dont run clang-tidy in CI - the algo uses a diff against the one-but-last commit which means that only the last commit in a PR is ever tested - adding an empty commit would always "fix" a breaking test

the next issue that we have is that trying to `run_clang_tidy` on all files is broken - `run-clang-tidy.py` has been re/moved to `clang-tools-extra` and is no longer available on the expected filepath

hacking around that problem, and the next issue is clang-tidy is taking up to ~1GB per file when linting. This seems crazy to me so im not sure if this is misconfiguration or just a very badly written tool. Either way with envoys current default `num_cpus` config it is very easy to OOM when trying to run this

with my hack and running on only 2 cores, it ~works, after a fashion and many hours, and with a bit more hacking, but the diff is useless anyway as it is inconsistent - eg it regularly changes the name of a var/function/etc but then doesnt update all instances of its use, sometimes in the same file 

so we have a pretty big tech debt to pay down here

PR #25393, opened 2023-02-06
HCM: Extend Envoy stream lifetime 

### LINKED ISSUE #24975: HTTP/2 request not logging DOWNSTREAM_WIRE_BYTES_SENT properly.
*Title*: HTTP/2 request not logging *DOWNSTREAM_WIRE_BYTES_SENT* properly when the downloaded file size is more than the initial window size set by HTTP/2 client.

*Description*:
It is observed that envoy logs request info to access logs while it's still transferring data to downstream HTTP/2 stream and logs incorrect wire bytes. This happens when HTTP/2 client sets initial window size less than the response size.

For example, when the requested file size is 169K, and HTTP/2 client (e.g., nghttp) sets initial window size to 65K then it is observed that access logs will have *BYTES_SENT=169K* and *DOWNSTREAM_WIRE_BYTES_SENT=~65K*. If we request same file by setting the initial window size (>169K), we do see correct wire bytes logged in access logs.

KBaichoo:
Commit Message: Extend Envoy stream lifetime to correctly capture underlying codec stream.
Additional Description: 
Risk Level: Medium
Testing: Integration test
Docs Changes:
Release Notes: Included
Runtime guard: Included, `envoy_reloadable_features_expand_agnostic_stream_lifetime`
Fixes #24975


envoy/http/codec.h
```diff
@@ -324,6 +324,25 @@ class StreamCallbacks {
   virtual void onBelowWriteBufferLowWatermark() PURE;
 };
 
+/**
+ * An adapter for an HTTP Stream that is tightly coupled with both the
+ * lifetime of the stream and low-level events that can occur on the stream.
+ */
+class StreamAdapter {
```
yanavlasov, line 331: Naming nit. It is unclear why this is an 'adapter'. I think it can just be `CodecEvents` or `StreamEvents`. If you decide to rename it, please also rename the setter method.

yanavlasov (commented):
LGTM implementation. Small naming nit, up to you if you want to take it.

/wait-any

yanavlasov (approved):
LGTM. I have also asked a non Google review on Slack.

yanavlasov:
This changes the lifetime of the filter manager stream info object, which is now tied to completed serialization of all bytes belonging to the stream. So the output to the access log is delayed until all bytes are written into the transport socket, or a timeout or connection error has occurred. This can potentially alter what access log is reporting. For example, presently access log can print 200 OK even though not all bytes were sent yet, and connection later has timed out. After this change the access log will print the connection termination error in such case. This is an unlikely case, but at the margin is a user visible behavior change.

Not sure if we need to reflect it in the change log, since this is something unlikely to be observed.

changelogs/current.yaml
```diff
@@ -17,6 +17,9 @@ minor_behavior_changes:
 - area: local_ratelimit
   change: |
     Tokens from local descriptor's token buckets are burned before tokens from the default token bucket.
+- area: http stream
+  change: |
+    Extended the lifetime of the protocol agnoistic stream object to correct discrepancies between what is access logged and what occurred with the protocol specific stream. This behavior change can be reverted by setting ``envoy_reloadable_features_expand_agnostic_stream_lifetime`` to ``false``.
```
ggreenway, line 22:
```suggestion
    Extended the lifetime of the protocol agnostic stream object to correct discrepancies between what is access logged and what occurred with the protocol specific stream. This behavior change can be reverted by setting ``envoy_reloadable_features_expand_agnostic_stream_lifetime`` to ``false``.
```

ggreenway (dismissed):
Nice fix! Aside from mispelled word, LGTM.

/wait

changelogs/current.yaml
KBaichoo: done

KBaichoo:
Thanks for the quick review @ggreenway !

phlax:
@yanavlasov @ggreenway looks like this is awaiting final signoff

/wait-any

phlax:
@KBaichoo just flagging that this has triggered a clang-tidy storm (which would fail once #25661 lands)

errors are here https://dev.azure.com/cncf/envoy/_build/results?buildId=128797&view=logs&j=b7634614-24f3-5416-e791-4f3affaaed6c&t=21e6aa7d-f369-5abd-5e4e-e888cac18e9c&l=1097

KBaichoo:
Thanks for flagging @phlax , has clang tidy been busted? What's the reason it's not catching before submission ?

phlax:
> has clang tidy been busted? 

afaict it has always been kinda busted

> What's the reason it's not catching before submission ?

 the current algo checks against 1 commit - so if your Pr has more than 1, only the last is checked - in postsubmit that ~works - so currently there are frequent failures there #24485 

the ~fix is here i think #25661 if you fancy reviewing - the sooner we get that on the sooner we can stem this, with the caveat that contributors are going to have to start running clang-tidy-diff against their code or have CI fail after an hour or so

phlax:
... and we either need to just fix everything now, or contributors will face fixing legacy issues in files they touch

KBaichoo:
> the current algo checks against 1 commit - so if your Pr has more than 1, only the last is checked - in postsubmit that ~works - so currently there are frequent failures there

Ah, that explains a lot. 

> going to have to start running clang-tidy-diff against their code or have CI fail after an hour or so

Is the latter (CI fail after an hour) new?

phlax:
if/when we land #25661 - which i think we should do quickly - but just mho - then PRs will have all _changed_ files checked by `clang-tidy-diff`

_changed_ files may have a hidden store of `clang-tidy` issues even tho the contributor makes a single-line hypothetical change

phlax:
`changed` = files changed since the last common ancestor to Envoy `main` (or ideally the target branch in the case of release branches - not quite sure how that will work with PR but i made the target branch configurable at least)

phlax:
... and only `main` will check this in PRs for now

PR #25396, opened 2023-02-07
mobile: re-enable RBE for `mobile_perf` CI jobs

jpsim:
This was accidentally disabled in https://github.com/envoyproxy/envoy/pull/25189, causing the CI job to be extremely slow.


abeyad (approved):
Great catch!

PR #25397, opened 2023-02-07
formatter: use orError for metadata formatter

### LINKED ISSUE #11259: API to Expose Histogram Raw Bins
In order to provide richer telemetry, it would be useful to provide the raw histogram bins found in the internal Circonus histograms to the outside world.

I have added data members to ``histogram_impl`` to record these bins so they can be made available via ``gRPC``.  However, it seems the ``gRPC`` messages from ``MetricsServiceSink`` are lifted directly from ``prometheus``.  This schema does not feature a message type that is appropriate for exposing these bins.

Would you have any guidance or preference on how this data should be exposed to the outside world?

botengyao:

This is PR is going to avoid crash for potential upgrade of Protobuf. 

The pre-release version of Protobuf includes this change:

`Expect fail when serialize inf and nan for Value.number_value in json format. fixes #11259` 
https://github.com/protocolbuffers/protobuf/commit/ca1cb1ba80ef18f5dccfb5b6ee7fa623ba6caab5

https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value

If we bump the Protobuf version, there will be an exception & crash using `NaN/inf/-inf` in `protobuf.struct` `number_value` when performing the serialization, while there is no check for this behavior right now.

The formatter is used in the header_parser. This is going to use the `orError` method for our metadata value formatter, rather than `orDie` to avoid future crash. 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/source/common/formatter/substitution_formatter.cc
+++ b/source/common/formatter/substitution_formatter.cc
@@ -1870,7 +1870,7 @@ MetadataFormatter::formatMetadata(const envoy::config::core::v3::Metadata& metad
   if (value.kind_case() == ProtobufWkt::Value::kStringValue) {
     str = value.string_value();
   } else {
-    str = MessageUtil::getJsonStringFromMessageOrDie(value, false, true);
+    str = MessageUtil::getJsonStringFromMessageOrError(value, false, true);
   }
   truncate(str, max_length_);
   return str;
--- a/test/common/router/header_parser_corpus/upstream_metadata_2
+++ b/test/common/router/header_parser_corpus/upstream_metadata_2
@@ -1,0 +1,22 @@
+headers_to_add {
+  header {
+    key: "d"
+    value: "%START_TIME()%%DYNAMIC_METADATA(|)%"
+  }
+}
+stream_info {
+  dynamic_metadata {
+    filter_metadata {
+      key: "|"
+      value {
+        fields {
+          key: "test"
+          value {
+            number_value: inf
+          }
+        }
+      }
+    }
+  }
+  start_time: 3
+}
--- a/test/common/router/header_parser_corpus/upstream_metadata_2
+++ b/test/common/router/header_parser_corpus/upstream_metadata_2
@@ -10,7 +10,7 @@ stream_info {
       key: "|"
       value {
         fields {
-          key: "test"
+          key: "test_key"
           value {
             number_value: inf
           }


botengyao:
/assign @adisuissa 
Thanks!

adisuissa (commented):
Thanks for fixing this (future) bug!

I've looked at this again, and I think there are a few things that we should do following this change.
1. These [function comments](https://github.com/envoyproxy/envoy/blob/main/source/common/protobuf/utility.h#L479-L520) should be updated to reflect the case where a double value is NaN/Inf/-Inf.
2. We should go through each of the other uses of the `getJsonStringFromMessage*` functions and see if they need to be addressed.
3. Looking at this specific use-case, it may make sense to check that if the value is a number that cannot be processed (nan/inf/-inf), then returning it as a string should be the better option.

cc'ing @yanavlasov for opinions on point (3) above.

botengyao:
Thanks @adisuissa, will do for 1 & 2, I will give my judgement.

I think it makes sense for point 3. This is kind of Protobuf now thinks it is illegal, and we will handle it on behalf of the dependency.

botengyao:
There could be a loop here; like the struct can contain a struct and then a struct... which contains `number_value:inf` for non_typed_metadata. I think the current code block only checks the typed_filter_metadata. e.g. 

```
if (value.kind_case() == ProtobufWkt::Value::kStringValue) {
    str = value.string_value();
  } else if (value.kind_case() == ProtobufWkt::Value::kNumberValue) { // new adding
    str = absl::StrCat(value.number_value());
  } else {
    str = MessageUtil::getJsonStringFromMessageOrError(value, false, true);
  }
```

It still depends the recursive Protobuf::util::MessageToJsonString for untyped one. We probably don't want to make the logic too complicated here.

```
message Struct {
  // Unordered map of dynamically typed values.
  map<string, Value> fields = 1;
}
```
Metadata:
```
  map<string, google.protobuf.Struct> filter_metadata = 1;
  map<string, google.protobuf.Any> typed_filter_metadata = 2;
```

adisuissa:
> There could be some recursive here; like the struct can contain a struct and then a struct... which contains `number_value:inf` for un_typed_metadata. I think the current code block only checks the typed_filter_metadata. e.g.
> 

Not sure I fully grok your question.
If I read the code correctly, it seems to only the `metadata.filter_metadata()` (untyped one) is accessed. I think that this method is currently using the `OrDie()` variant, because the `Struct` you mentioned cannot have an `Any` value, so the underlying call will not cause a release-assert failure. Let me know if I'm missing something here.

I think I see the problem with the recursive call, which will impact a nested structure, e.g.:
```
Struct {
  "key1": { string_value: "my_string" }
  "key2": { number_value: Inf }
  "key3": { Struct_value: {
                      "nested_key": { number_value: NaN }
                   }
                }
}
```
will cause a failure when formatting "key2" or "key3".

The question that I think should guide us is what should the `formatMetadata()` method return in this case. It returns a string, so the question what should be the string value.
If the returned string value could be the string "Inf" for "key2", and json object with `"nested_key": "inf"` for "key3", then it would work for this case.

botengyao:
cc @mattklein123, @htuch as well.

adisuissa:
Ping @yanavlasov for thoughts on [this comment](https://github.com/envoyproxy/envoy/pull/25397#pullrequestreview-1289264013).

PR #25414, opened 2023-02-07
Remove scope id from ipv6 address in getLocalAddress

### LINKED ISSUE #25326: Envoy-mobile clients fail to speak IPV6 on Android clients with ndk >=24

RenjieTang:

Commit Message: Remove scope id from ipv6 address in getLocalAddress
Additional Description: This fixes the issue where Envoy-mobile fails to send out any packet on ipv6.
Risk Level: Low
Testing: unit tests
Docs Changes: n/a
Release Notes: n/a
Fixes #25326

### DIFF
--- a/source/common/network/utility.cc
+++ b/source/common/network/utility.cc
@@ -236,6 +236,9 @@ Address::InstanceConstSharedPtr Utility::getLocalAddress(const Address::IpVersio
       if (!isLoopbackAddress(*interface_address.interface_addr_) &&
           interface_address.interface_addr_->ip()->version() == version) {
         ret = interface_address.interface_addr_;
+        if (ret->ip()->version() == Address::IpVersion::v6) {
+          ret = ret->ip()->ipv6()->addressWithoutScopeId();
+        }
         break;
       }
     }
--- a/envoy/network/address.h
+++ b/envoy/network/address.h
@@ -64,6 +64,12 @@ public:
    * if the Ipv6 address isn't Ipv4 mapped.
    */
   virtual InstanceConstSharedPtr v4CompatibleAddress() const PURE;
+
+  /**
+   * @return Ipv6 address that has no scope/zone identifier. Return `nullptr`
+   * if the conversion failed.
+   */
+  virtual InstanceConstSharedPtr addressWithoutScopeId() const PURE;
 };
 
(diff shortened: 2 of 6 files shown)


source/common/network/utility.cc
```diff
@@ -236,6 +236,9 @@ Address::InstanceConstSharedPtr Utility::getLocalAddress(const Address::IpVersio
       if (!isLoopbackAddress(*interface_address.interface_addr_) &&
           interface_address.interface_addr_->ip()->version() == version) {
         ret = interface_address.interface_addr_;
+        if (ret->ip()->version() == Address::IpVersion::v6) {
```
RyanTheOptimist, line 239: Is it possible that ret is null, or ret->ip() is null? Maybe:
```
if (ret && ret->ip() && ret->ip()->version() == Address::IpVersion::v6) {
````

RyanTheOptimist (commented):
LGTM, modulo one nit.

source/common/network/utility.cc
RenjieTang: I don't think the null guard is needed here because in the if statement above, there's already a `interface_address.interface_addr_->ip()->version() == version` call.

RyanTheOptimist: Oh good lord! You're totally right.

RyanTheOptimist (approved):

PR #25417, opened 2023-02-08
datadog: HTTP client portion of new tracing library

dgoffredo:

This revision adds a component that will be used in the future when more of the new library is integrated into the Datadog tracing extension.

`AgentHTTPClient` allows the new tracing library to send spans to the Datadog Agent by making HTTP POST requests to a configured cluster. `AgentHTTPClient` implements the [datadog::tracing::HTTPClient][1] interface and the [Envoy::Http::AsyncClient::Callbacks][2] interface.

[1]: https://github.com/DataDog/dd-trace-cpp/blob/main/src/datadog/http_client.h
[2]: https://github.com/envoyproxy/envoy/blob/84ca17add2171c04876aaf345762f03b4d6c86b8/envoy/http/async_client.h#L57


source/extensions/tracers/datadog/agent_http_client.cc
```diff
@@ -0,0 +1,143 @@
...
+
+AgentHTTPClient::AgentHTTPClient(Upstream::ClusterManager& cluster_manager,
+                                 const std::string& cluster, const std::string& reference_host,
+                                 TracerStats& stats)
+    : collector_cluster_(cluster_manager, cluster), cluster_(cluster),
+      reference_host_(reference_host), stats_(stats) {}
+
+AgentHTTPClient::~AgentHTTPClient() {
+  for (const auto& [request_ptr, _] : handlers_) {
+    assert(request_ptr);
```
wbpcode, line 33: Please use `ASSERT` or `RELEASE_ASSERT`, thanks.

source/extensions/tracers/datadog/agent_http_client.h
```diff
@@ -0,0 +1,106 @@
...
+
+  void onSuccess(const Http::AsyncClient::Request&, Http::ResponseMessagePtr&&) override;
+  void onFailure(const Http::AsyncClient::Request&, Http::AsyncClient::FailureReason) override;
+  void onBeforeFinalizeUpstreamSpan(Tracing::Span&, const Http::ResponseHeaderMap*) override;
+
+private:
+  absl::flat_hash_map<Http::AsyncClient::Request*, Handlers> handlers_;
+  Upstream::ClusterUpdateTracker collector_cluster_;
+  std::string cluster_;
+  std::string reference_host_;
```
wbpcode, line 99: nit: `const std::string` prefered.

test/extensions/tracers/datadog/agent_http_client_test.cc
```diff
@@ -0,0 +1,451 @@
...
+
+    I can go anywhere
+    Friends to know
+    And ways to grow
+    A reading rainbow
+
+    I can be anything
+    Take a look
+    It's in a book
+    A reading rainbow)body";
```
wbpcode, line 157: hah :)

test/extensions/tracers/datadog/agent_http_client_test.cc
```diff
@@ -0,0 +1,451 @@
...
+  // If the underlying call to `httpAsyncClient().send(...)` returns an error,
+  // then the enclosing call to `AgentHTTPClient::post(...)` returns an error.
+
+  EXPECT_CALL(cluster_manager_.instance_.thread_local_cluster_.async_client_, send_(_, _, _))
+      .WillOnce(
+          Invoke([this](Http::RequestMessagePtr&, Http::AsyncClient::Callbacks& callbacks_arg,
+                        const Http::AsyncClient::RequestOptions&) -> Http::AsyncClient::Request* {
+            callbacks_ = &callbacks_arg;
+            return nullptr; // indicates error
+          }));
```
wbpcode, line 415: call the `onFailture` here to make the mock closer to the real situation.
If request send fail then `onFailure` will be called synchronously.

wbpcode (commented):
👍 Thanks and some minor comments are added. 😸

test/extensions/tracers/datadog/agent_http_client_test.cc
dgoffredo: On the contrary, if the synchronous return value of `post` is an error, then neither callback is invoked.

An alternative design would be that `post` always returns void, and success/failure is conveyed using the callbacks only. I chose the current contract, though.

dgoffredo:
Thanks for the review. I'll commit some changes within the next couple of days.

test/extensions/tracers/datadog/agent_http_client_test.cc
wbpcode: You have made a check in the `onFailture` callback. If the sending fail, the request will no be registered and the `onFailure` call do nothing. So I think current behavior match the requirements of `post`.

I just think we should add a `onFailure` call here to make the failure calling mock more real. Even the `onFailure` do nothing when the calling fails.

dgoffredo: I see, will do.

wbpcode: sorry, but I don't see the `onFailure` call. Do I miss something? /wait-any

dgoffredo: Now I understand.

I didn't read your comments closely enough. You are talking about the behavior of `Http::AsyncClient::send`, and you want this mock to emulate it more closely.

This got me reading <https://github.com/envoyproxy/envoy/blob/main/source/common/http/async_client_impl.cc>, and I don't think that `onFailure` is called in the `send`-returned-`nullptr` case.

I'll play with it a bit and report back.

dgoffredo: I'm having trouble to get that code path to execute in a real build of Envoy (e.g. based off of https://github.com/envoyproxy/envoy/pull/23958).

If the "Agent" is not available, the async client delivers a 503 response. If the "Agent" is configured to close every socket that it accepts, the async client delivers a 503 response.

This is using the `demo/` directory (with some modifications) from https://github.com/envoyproxy/envoy/pull/23958.

Can you show statically that `onFailure` is called when `post` returns `nullptr`?

dgoffredo:
> Code coverage for source/extensions/tracers/datadog is lower than limit of 96.6 (96.5)

hmmmm

test/extensions/tracers/datadog/agent_http_client_test.cc
wbpcode: This will happen when our async request failed imediately. For example, if the target cluster has no health upstream or the target cluster is in the maintain mode.
It's a little hard to find it when you are using static cluster.

dgoffredo: I'll add that, and a comment explaining the potential rarity of the behavior.

dgoffredo: One of the parameters of `onFailure` is a reference to a request. Here, though, we don't have a request. Are you sure `onFailure` is ever invoked in this context?

dgoffredo: I suppose I could use a mock instance just to have something to pass to `onFailure`. Perhaps in the real code, a request is created but later destroyed due to one of the conditions you mentioned. I'll do that.

wbpcode: cc @dgoffredo Sorry, after a quick check to the source code, maybe I am wrong. Seem `onSuccess` will be called with a 503 response when nullptr is returned by the send. I will check it again and give a final conclusion. 🤣

wbpcode: cc @dgoffredo It's clear now. If the upstream cluster is invalid (no endpoint, maintain mode, etc.) the local response will complete directly and `onSuccess` will be called with a 503 response. The request will be destroyed and `send` will return a nullptr.

You can validate it by setting the endpoints of your static cluster to an empty array `[]`.


```
template <typename T> T* AsyncClientImpl::internalStartRequest(T* async_request) {
  async_request->initialize();
  std::unique_ptr<AsyncStreamImpl> new_request{async_request};

  // The request may get immediately failed. If so, we will return nullptr.
  if (!new_request->remote_closed_) {
    LinkedList::moveIntoList(std::move(new_request), active_streams_);
    return async_request;
  } else {
    new_request->cleanup();
    return nullptr;
  }
}
```

source/extensions/tracers/datadog/agent_http_client.cc
```diff
@@ -0,0 +1,144 @@
...
+    ENVOY_LOG(debug, "traces successfully submitted to datadog agent");
+    stats_.reports_sent_.inc();
+  }
+
+  auto found = handlers_.find(const_cast<Http::AsyncClient::Request*>(&request));
+  if (found == handlers_.end()) {
+    ENVOY_LOG(debug, "request at address 0x{} is not in the map of {} Handlers objects",
+              absl::StrFormat("%p", &request), handlers_.size());
+    return;
+  }
```
wbpcode, line 104: Should we do a handler check first here to ensure the request is sent rather than a local failure?

test/extensions/tracers/datadog/agent_http_client_test.cc
dgoffredo: Perfect, thanks for the investigation. I'll try it with the empty cluster array to get a feel for it.

source/extensions/tracers/datadog/agent_http_client.cc
dgoffredo: What do you mean by "local failure" and "handler check"?

In light of our discussion in the unit test, maybe you're referring to the 503 "success" scenarios where a request is not actually ever sent. If so, then I'm not sure what is best. Since we have an HTTP response status, I'm tempted to consider it a "success" for the purposes of the `Handlers` callbacks. On the other hand, being able to distinguish between "sent a request and received a 503" and "never sent a request and indicated this with a 503" might be helpful for future debugging. Ideally the two cases could be distinguished, but I'm not sure they can be.

My preference is to leave the 503 ambiguity and consider it a true response.

What do you think?

test/extensions/tracers/datadog/agent_http_client_test.cc
dgoffredo: I tried it with `[]` for the Datadog Agent cluster's endpoints. It's as you say: `send(...)` returns `nullptr` and then `onSuccess` is invoked with status 503. See the all-caps logging I added for debugging:
```
[2023-02-17T19:25:56.874Z] "GET / HTTP/1.1" 200 - 0 505 7 5 "192.168.1.106" "curl/7.81.0" "b5595a06-9a5c-9988-b286-2221df9cb374" "localhost:1337" "[::1]:8080" "-" "4611974407231517207" "4611974407231517207"
[2023-02-17 14:25:57.803][1158443][critical][tracing] [source/extensions/tracers/datadog/agent_http_client.cc:93] ONSUCCESS() INVOKED WITH STATUS 503!
SEND() IS ABOUT TO CALL CLEANUP() AND THEN RETURN NULLPTR OOOOOOOOOOOO
[2023-02-17 14:25:57.803][1158443][critical][tracing] [source/extensions/tracers/datadog/agent_http_client.cc:70] SEND() RETURNED NULLPTR OH THE HORROR!
[2023-02-17 14:25:57.803][1158443][error][tracing] [source/extensions/tracers/datadog/logger.cc:37] Datadog [error 44]: Failed to create request.
```
For the purposes of this unit test, then, I'll call the success handler with 503, and add a comment that this behavior is not required by `AgentHTTPClient` (this way, if the behavior later changes and breaks this test, a maintainer knows that the test can be changed safely).

source/extensions/tracers/datadog/agent_http_client.cc
wbpcode: Sounds okay to me. :)

dgoffredo:
Ping @wbpcode

wbpcode (approved):
Sorry for the delay. LGTM.

PR #25432, opened 2023-02-08
Revert "CDS mobile builder API and integration test (#25123)"

abeyad:
This reverts commit 4c507479c5dbc30e59c39604e52a459bc911b9aa.

The commit triggered CI failures:
https://github.com/envoyproxy/envoy/actions/runs/4125892811
https://github.com/envoyproxy/envoy/actions/runs/4125892787

Signed-off-by: Ali Beyad <abeyad@google.com>

PR #25438, opened 2023-02-08
Fix race condition in mobile cds_integration_test

caschoener:
This test had a race condition, because clusters may initially be in the warming state and not active in which case the comparison would fail.  Waiting for cluster count to increase guarantees success


caschoener:
This was reverted in https://github.com/envoyproxy/envoy/pull/25432. I can't repro the failures locally, so making another PR to trigger CI

abeyad (approved):

jpsim (approved):

Issue #25116, opened 2023-01-24
local_ratelimit listener filter


JuniorHsu:
*Title*: *local_ratelimit listener filter*

*Description*:
We saw CPU usage could be high if there's too many downstream connections which requesting handshake in a short period. Especially while a host is already busy, there downstream hosts could disconnect by timeout and connect again.

This is the perfect scenario of local rate limit. However, local rate_limit network filter is too late since the filter is involved after handshake.

Therefore, I'd like to move on a local_ratelimit listener filter.
Ideally we want to block too many handshake in a short period but I haven't seen through the implementation.
A possible way is to count all `onAccept` no matter whether it is TLS.
Next step could limit on the `requestedApplicationProtocols` or `detectedTransportProtocol` from `tls_inspector`

cc local ratelimit codeowners @mattklein123 @wbpcode

mattklein123:
Yeah this makes sense to me.

PR #25440, opened 2023-02-08
listener: Local rate limit listener filter

JuniorHsu:

Commit Message: added local rate limit listener filter to enable rate limit before TLS handshake and filter chain matching
Additional Description:
Risk Level: low
Testing: unit and integration test
Docs Changes: *local_rate_limiting.rst
Fixes #Issue: #25116


docs/root/intro/arch_overview/other_features/local_rate_limiting.rst
```diff
@@ -4,7 +4,8 @@ Local rate limiting
 ===================
 
 Envoy supports local (non-distributed) rate limiting of L4 connections via the
-:ref:`local rate limit filter <config_network_filters_local_rate_limit>`.
+:ref:`local rate limit listener filter <config_listener_filters_local_rate_limit>` and :ref:`local rate limit network filter <config_network_filters_local_rate_limit>`.
+The difference is `local rate limit listener filter <config_listener_filters_local_rate_limit>` processes the sockets before TLS handshake and filter matching.
```
PeterL328, line 8: Wonder if we should write `and filter chain matching` to be more precise

JuniorHsu: filter chain is better. thanks!

source/extensions/filters/listener/local_ratelimit/local_ratelimit.cc
```diff
@@ -0,0 +1,70 @@
...
+      Protobuf::RepeatedPtrField<
+          envoy::extensions::common::ratelimit::v3::LocalRateLimitDescriptor>());
+}
+
+bool Config::canCreateConnection() { return rate_limiter_->requestAllowed({}); }
+
+Config::~Config() {
+  // Reset the rate_limiter_ first so that the weak_ptr will be expired() if this was the last
+  // reference.
+  rate_limiter_.reset();
```
yanavlasov, line 40: I'm unsure I understand the need to explicitly reset the shared pointer in the destructor, as it will be reset during normal clean-up.

yanavlasov (changes requested):
/wait

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
```diff
@@ -0,0 +1,81 @@
...
+  ALL_LOCAL_RATE_LIMIT_STATS(GENERATE_COUNTER_STRUCT)
+};
+
+using LocalRateLimiterImplSharedPtr =
+    std::shared_ptr<Filters::Common::LocalRateLimit::LocalRateLimiterImpl>;
+
+/**
+ * Configuration shared across all connections. Must be thread safe.
+ */
+class Config : Logger::Loggable<Logger::Id::filter> {
```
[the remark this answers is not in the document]
JuniorHsu: It's actually filter config. I saw the convention is on the same file with filter. Should I rename to `FilterConfig` to avoid confusion?

source/extensions/filters/listener/local_ratelimit/local_ratelimit.cc
JuniorHsu: right we probably wont need to reset. I stole this from network local ratelimit but I thought reset is needed since the rate limiter is shared in network filter.

source/extensions/filters/listener/local_ratelimit/local_ratelimit.cc
```diff
@@ -0,0 +1,70 @@
...
+namespace ListenerFilters {
+namespace LocalRateLimit {
+
+Config::Config(
+    const envoy::extensions::filters::listener::local_ratelimit::v3::LocalRateLimit& proto_config,
+    Event::Dispatcher& dispatcher, Stats::Scope& scope, Runtime::Loader& runtime)
+    : enabled_(proto_config.runtime_enabled(), runtime),
+      stats_(generateStats(proto_config.stat_prefix(), scope)) {
+
+  rate_limiter_ = std::make_shared<Filters::Common::LocalRateLimit::LocalRateLimiterImpl>(
```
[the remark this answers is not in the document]
JuniorHsu: done. thanks!

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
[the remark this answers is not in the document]
yanavlasov: Normally configuration and filter implementations are split into dedicated files. I would expect the config class to be in the config.h/cc files together with the factory.

yanavlasov (commented):
/wait-any

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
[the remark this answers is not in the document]
JuniorHsu: Factory initializes the filter, and filter needs the config information. 
That is, if we put factory and config into `config.h/cc`, we have circular dependency between config module and filter module, which fails the bazel check. 

I'm unsure the we have precedence putting config with factory. 
We have one case in http land `jwt_authn`, which put config into its own file.
However it needs to have another own file to handle factory to avoid circular dependency.

I could do three pairs of .h/.cpp if we really want to. 

```
 git grep "class FilterConfig" -- *.h
source/extensions/filters/http/alternate_protocols_cache/filter.h:17:class FilterConfig {
source/extensions/filters/http/aws_request_signing/aws_request_signing_filter.h:37:class FilterConfig {
source/extensions/filters/http/aws_request_signing/aws_request_signing_filter.h:67:class FilterConfigImpl : public FilterConfig {
source/extensions/filters/http/bandwidth_limit/bandwidth_limit.h:61:class FilterConfig : public ::Envoy::Router::RouteSpecificFilterConfig {
source/extensions/filters/http/custom_response/config.h:30:class FilterConfig : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/ext_authz/ext_authz.h:54:class FilterConfig {
source/extensions/filters/http/ext_authz/ext_authz.h:242:class FilterConfigPerRoute : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/ext_proc/ext_proc.h:68:class FilterConfig {
source/extensions/filters/http/ext_proc/ext_proc.h:108:class FilterConfigPerRoute : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/grpc_http1_reverse_bridge/filter.h:69:class FilterConfigPerRoute : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/jwt_authn/filter_config.h:41:class FilterConfig {
source/extensions/filters/http/jwt_authn/filter_config.h:62:class FilterConfigImpl : public Logger::Loggable<Logger::Id::jwt>,
source/extensions/filters/http/local_ratelimit/local_ratelimit.h:70:class FilterConfig : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/lua/lua_filter.h:375:class FilterConfig : Logger::Loggable<Logger::Id::lua> {
source/extensions/filters/http/lua/lua_filter.h:414:class FilterConfigPerRoute : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/oauth2/filter.h:128:class FilterConfig {
source/extensions/filters/http/ratelimit/ratelimit.h:41:class FilterConfig {
source/extensions/filters/http/ratelimit/ratelimit.h:118:class FilterConfigPerRoute : public Router::RouteSpecificFilterConfig {
source/extensions/filters/http/tap/tap_filter.h:37:class FilterConfig {
source/extensions/filters/http/tap/tap_filter.h:57:class FilterConfigImpl : public FilterConfig, public Extensions::Common::Tap::ExtensionConfigBase {
source/extensions/filters/http/wasm/wasm_filter.h:24:class FilterConfig : Logger::Loggable<Logger::Id::wasm> {
source/extensions/filters/network/wasm/wasm_filter.h:24:class FilterConfig : Logger::Loggable<Logger::Id::wasm> {
```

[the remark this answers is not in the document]
tyxia: I think the question here is : Do you need this separate class`class config`? Also, looks like the `config` class doesn't use synchronization primitive or use TLS (i.e., RCU pattern). So, I am not sure thread safe is guaranteed here when this class  is not read-only. (canCreateConnection() is  fine as its underlying `requestAllowedHelper` use atomic operation)

I think it could just work if we remove config class and its members/methods are moved into to `class Filter`? The threading safe is also naturally achieved.  WDYT?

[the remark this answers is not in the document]
tyxia: On second thought, it is good to have this `config` class as it is, since its members are read only and rate limiter is not shared.

Maybe rename it to 'FilterConfig' to be consistent with other places in code base.

tyxia (commented):
Thanks for working on this! Left some comments around `config/FilterConfig` class

phlax:
@JuniorHsu seems this needs a main merge and is awaiting some feedback

/wait

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
```diff
@@ -0,0 +1,81 @@
...
+  ~Config() = default;
+
+  bool canCreateConnection();
+  bool enabled() { return enabled_.enabled(); }
+  const LocalRateLimitStats& stats() const { return stats_; }
+
+private:
+  static LocalRateLimitStats generateStats(const std::string& prefix, Stats::Scope& scope);
+
+  Runtime::FeatureFlag enabled_;
```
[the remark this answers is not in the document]
JuniorHsu: done. thanks

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
```diff
@@ -0,0 +1,81 @@
...
+  LocalRateLimitStats stats_;
+  LocalRateLimiterImplSharedPtr rate_limiter_;
```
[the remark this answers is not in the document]
JuniorHsu: make sense. thanks! i'll fix network/local_ratelimit in a separated PR

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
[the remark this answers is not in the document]
JuniorHsu: done. thanks!

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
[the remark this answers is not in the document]
tyxia: I am not sure if we need to fix `network/local_ratelimit`. 

My understanding of `shared_ptr` there is 1) if `share_key` has value(i.e. sharing is enabled), they leverage shared_ptr 2) if `share_key` is empty(i.e., sharing is not enable), they just `make_shared` every time it is created, so each one has separate control with ref_count=1, which basically use it like a unique_ptr. 
And I think the reason `shared_ptr` is used there is because of coding convenience/consistence as you just need same type of object for above two cases.

Thank you for addressing my comments and adding tests! I will take a look later.

yanavlasov:
LGTM from me. Waiting for final review from @tyxia

JuniorHsu:
Hello @markdroth,
Given we're closed to finish the implementation review, could you have a review on the api part?
Generally speaking, this api is a simplified version of [network local rate limit](https://github.com/envoyproxy/envoy/blob/main/api%2Fenvoy%2Fextensions%2Ffilters%2Fnetwork%2Flocal_ratelimit%2Fv3%2Flocal_rate_limit.proto)

Thanks!

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Local Rate Limit listener filter.
+ */
+class Filter : public Network::ListenerFilter, Logger::Loggable<Logger::Id::filter> {
+public:
+  Filter(const FilterConfigSharedPtr& config) : config_(config) {}
+
+  // Network::ListenerFilter
+  Network::FilterStatus onAccept(Network::ListenerFilterCallbacks& cb) override;
+  Network::FilterStatus onData(Network::ListenerFilterBuffer&) override {
```
tyxia, line 69: I am a bit confused, why we  override `onData` and `maxReadBytes` here as this filter inherits from `Network::ListenerFilter` which only has one `onAccept()`

test/extensions/filters/listener/local_ratelimit/local_ratelimit_test.cc
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    Network::FilterStatus onAccept() { return filter_.onAccept(cb_); }
+    Network::FilterStatus onEmptyData() {
+      // This is for test coverage purpose only.
+      // onData shouldn't be called since maxReadBytes is zero.
+      Network::ListenerFilterBufferImpl buffer(
+          io_handle_, dispatcher_, [](bool) {}, [](Network::ListenerFilterBuffer&) {}, 1);
+      return filter_.onData(buffer);
+    }
```
tyxia, line 45: If you remove those unnecessary overrides from implementation, the function (like this one) and related tests  seems can also be removed?

test/extensions/filters/listener/local_ratelimit/local_ratelimit_integration_test.cc
```diff
@@ -0,0 +1,87 @@
...
+                   ->value());
+}
+
+TEST_P(LocalRateLimitTcpIntegrationTest, RateLimited) {
+  setup(R"EOF(
+stat_prefix: local_rate_limit_stats
+token_bucket:
+  max_tokens: 1
+  # Set fill_interval to effectively infinite so we only get max_tokens to start and never re-fill.
+  fill_interval: 100000s
```
tyxia, line 69: Just to confirm (for my own learning purpose): If you add a sleep(100000s) in this test, the token will be re-filled and then it will not be rate limited?

tyxia (commented):
Just left a few comments(mostly nits), we are getting close to there. Thanks!

source/extensions/filters/listener/local_ratelimit/local_ratelimit.cc
```diff
@@ -0,0 +1,63 @@
+#include "source/extensions/filters/listener/local_ratelimit/local_ratelimit.h"
+
+#include <algorithm>
+#include <string>
+#include <vector>
+
+#include "envoy/common/exception.h"
+#include "envoy/common/platform.h"
```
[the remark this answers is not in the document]
JuniorHsu: probably not i could give it a try to remove some but i don't have a good tool

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
```diff
@@ -0,0 +1,81 @@
...
+  bool canCreateConnection();
+  bool enabled() { return enabled_.enabled(); }
+  const LocalRateLimitStats& stats() const { return stats_; }
+
+private:
+  static LocalRateLimitStats generateStats(const std::string& prefix, Stats::Scope& scope);
+
+  const Runtime::FeatureFlag enabled_;
+  const LocalRateLimitStats stats_;
+  LocalRateLimiterImplPtr rate_limiter_;
```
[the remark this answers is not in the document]
JuniorHsu: oh i misread that. bare object should be fine for listener filter

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
JuniorHsu: since we don't have a pass through listener filter to inherit iirc. i could have one but that's should be a separated PR.

JuniorHsu: i.e., all method in ListenerFilter are pure abstract and needs to be override

test/extensions/filters/listener/local_ratelimit/local_ratelimit_test.cc
JuniorHsu: same here.

test/extensions/filters/listener/local_ratelimit/local_ratelimit_integration_test.cc
JuniorHsu: should be yeah. by the way we do enable in our prod and it refills which is great. and we already prove the filter can save handshake issue on scale (see the issue on description.

source/extensions/filters/listener/local_ratelimit/local_ratelimit.h
tyxia: Sorry, I am trying github vscode codespace out.  Somehow, it only shows `virtual FilterStatus onAccept(ListenerFilterCallbacks& cb) PURE;` (same to all other inherited classes).

I checked the latests code. My comment should be ignored.

tyxia: ah, I create the codespace based off my fork repo that is [6590 commits behind](https://github.com/tyxia/envoy/compare/main...envoyproxy:envoy:main)... 
Sorry, please ignore this out-of-date comment

tyxia (approved):
LGTM. Thanks for good work!

tyxia:
@mattklein123  Matt, as the code owner, maybe you can take a look and approve the API and PR if it looks good to you?

mattklein123 (approved):
API LGTM, thanks. LGTM to ship and iterate! Thank you.

mattklein123:
Will defer to @yanavlasov for final/review merge.

yanavlasov (approved):

PR #25441, opened 2023-02-08
custom response: Add ResponseConfig to redirect policy

pradeepcrao:

Commit Message:
Additional Description:
Risk Level: Low
Testing: Added
Docs Changes: (part of proto changes)
Release Notes: Added.


api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +27,78 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
 // upstream.
 // [#next-free-field: 7]
```
adisuissa, line 28: nit: the numbering here needs to be updated.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +27,78 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
+    // 2. if request uri is "/old-path-2?bar=1", users will be redirected to "/new-path-2"
+    // 3. if request uri is "/old-path-3?bar=1", users will be redirected to "/new-path-3?foo=1"
+    //
+    // Note that at least one of host_redirect and path_redirect has to be
+    // specified.
+    string path_redirect = 5
+        [(validate.rules).string = {well_known_regex: HTTP_HEADER_VALUE strict: false}];
+
+    // Indicates that during redirection, the query portion of the URL will
+    // be removed. Default value is false.
```
adisuissa, line 77: The description only matches `strip_query`. I suggest moving the comments next to each of the fields.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +27,78 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
+    //
+    // Example:
+    //
+    // .. code-block:: yaml
+    //
+    //    uri: https://www.mydomain.com/path/to/404.txt
+    //
+    string uri = 1 [(validate.rules).string = {min_len: 1}];
+
+    RedirectAction redirect_action = 2;
```
adisuissa, line 100: Please add comments to this field.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +27,78 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
+    // specified.
+    string path_redirect = 5
+        [(validate.rules).string = {well_known_regex: HTTP_HEADER_VALUE strict: false}];
+
+    // Indicates that during redirection, the query portion of the URL will
+    // be removed. Default value is false.
+    oneof query_specifier {
+      bool strip_query = 6;
+
+      string query = 7;
```
adisuissa, line 81: should this be a repeated [QueryParameter](https://github.com/envoyproxy/envoy/blob/79acebdd40751bacae3b9d9648910efa4f362488/api/envoy/config/core/v3/base.proto#L305)?
It is not clear if this sets new query-params, overrides some, or replaces them all.

I would also suggest looking at: https://github.com/envoyproxy/envoy/blob/79acebdd40751bacae3b9d9648910efa4f362488/api/envoy/service/auth/v3/external_auth.proto#L107-L114 as it seems it is doing something similar.

adisuissa (commented):
Thanks! Left some API comments.
/wait

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +27,78 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
+    // - match: { path: "/old-path-1" }
+    //   redirect: { path_redirect: "/new-path-1" }
+    // - match: { path: "/old-path-2" }
+    //   redirect: { path_redirect: "/new-path-2", strip-query: "true" }
+    // - match: { path: "/old-path-3" }
+    //   redirect: { path_redirect: "/new-path-3?foo=1", strip_query: "true" }
+    //
+    // 1. if request uri is "/old-path-1?bar=1", users will be redirected to "/new-path-1?bar=1"
+    // 2. if request uri is "/old-path-2?bar=1", users will be redirected to "/new-path-2"
+    // 3. if request uri is "/old-path-3?bar=1", users will be redirected to "/new-path-3?foo=1"
```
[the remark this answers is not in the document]
pradeepcrao: Removed `query`. (I can't recollect now why I did it this way.)

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
pradeepcrao: Removed `query`. I'm trying to keep this as close as possible to `config.route.v3.RedirectAction` to allow for maximum code reuse.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
pradeepcrao: Added.

pradeepcrao:
> Thanks! Left some API comments. /wait

Thanks so much for the review @adisuissa !

source/common/http/utility.cc
```diff
@@ -34,6 +34,50 @@
...
+    // In the rare case that X-Forwarded-Proto and scheme disagree (say http URL over an HTTPS
+    // connection), do port stripping based on X-Forwarded-Proto so http://foo.com:80 won't
+    // have the port stripped when served over TLS.
+    absl::string_view request_protocol = headers.getForwardedProtoValue();
+    bool remove_port = !new_port.empty();
+
+    if (new_scheme != request_protocol) {
+      remove_port |= (request_protocol == Http::Headers::get().SchemeValues.Https.c_str()) &&
+                     request_port == ":443";
+      remove_port |= (request_protocol == Http::Headers::get().SchemeValues.Http.c_str()) &&
```
adisuissa, line 68: nit: can `c_str()` be removed?

source/common/http/utility.cc
```diff
@@ -34,6 +34,50 @@
 #include "quiche/http2/adapter/http2_protocol.h"
 
 namespace Envoy {
+namespace {
+
+absl::string_view processRequestHost(const Http::RequestHeaderMap& headers,
```
adisuissa, line 39: Please add a comment for the function.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +26,75 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
-  //
-  // Example:
-  //
-  //  .. code-block:: yaml
-  //
-  //    path: /path/to/503_response.txt
-  //
-  string path = 2 [(validate.rules).string = {min_len: 1}];
+  // [#next-free-field: 7]
+  message RedirectAction {
```
adisuissa, line 30: Taking a step back, what are the main differences between this and [RedirectAction](https://github.com/envoyproxy/envoy/blob/0eaebeb11bdaba33fbd722c0839b074a5c1a1919/api/envoy/config/route/v3/route_components.proto#L1624)?
Keeping parity (and reusing code) may be the right thing to do here, and would like to understand why this shouldn't be done.

adisuissa (commented):
Thanks, left a high-level questio.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
```diff
@@ -26,28 +26,75 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
...
-  // The path for the custom response.
-  //
-  // Example:
-  //
-  //  .. code-block:: yaml
-  //
-  //    path: /path/to/503_response.txt
-  //
-  string path = 2 [(validate.rules).string = {min_len: 1}];
+  // [#next-free-field: 7]
```
yanavlasov, line 29: Can we reuse config.route.v3.RedirectAction here, or is it different enough that a new config is needed?

yanavlasov (commented):
/wait

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
pradeepcrao: The main difference is that certain fields such as `response_code`, `prefix_rewrite` and `regex_rewrite` don't make sense for custom response. `response_code` does not make sense because it's limited to redirect responses, and custom response already has a separate field for it. The other two don't make sense because we don't have access to the original request before `RedirectAction` was applied to it by the route, as well as the original route itself, to infer how to apply these fields for custom response.

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
pradeepcrao: Please see my response to Yan's comment regarding the same. Please let me know if that sounds reasonable to you.

source/extensions/http/custom_response/redirect_policy/redirect_policy.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
+#include <limits>
```
yanavlasov, line 3: Are these new headers needed?

yanavlasov (changes requested):
LGTM with a small nit. Thanks for refactoring and reusing common redirect code.

/wait

source/extensions/http/custom_response/redirect_policy/redirect_policy.h
pradeepcrao: Sorry, this was was needed before, but now anymore. Removed.

source/common/http/utility.h
```diff
@@ -668,6 +669,29 @@ bool isSafeRequest(const Http::RequestHeaderMap& request_headers);
...
+/**
+ * Container for route config elements that pertain to a redirect.
+ */
+struct RedirectConfig {
+  const std::string scheme_redirect_;
+  const std::string host_redirect_;
+  const std::string port_redirect_;
+  const std::string path_redirect_;
+  const std::string prefix_rewrite_redirect_;
+  std::string regex_rewrite_redirect_substitution_;
```
yanavlasov, line 681: I think you can make these fields `const` as well. You can use ASSERT after the structure was created to check for mutually exclusive fields.

yanavlasov (dismissed):

api/envoy/extensions/http/custom_response/redirect_policy/v3/redirect_policy.proto
pradeepcrao: Took a stab at using `RedirectAction` from route_components.proto PTAL

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -126,18 +159,26 @@ ::Envoy::Http::FilterHeadersStatus RedirectPolicy::encodeHeaders(
         }
       });
 
+  ::Envoy::Http::Utility::Url absolute_url;
+  std::string uri(uri_ ? *uri_
+                       : ::Envoy::Http::Utility::newPath(*redirect_action_, *downstream_headers));
+  if (!absolute_url.initialize(uri, false)) {
+    stats_.custom_response_invalid_uri_.inc();
+    ENVOY_BUG(!static_cast<bool>(uri_),
```
adisuissa, line 167: Not sure if this claim is true. IIUC `uri_` is validated, but the call to newPath may return an invalid uri.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -30,23 +57,33 @@ RedirectPolicy::RedirectPolicy(
...
+    if (!absolute_url.initialize(*uri_, false)) {
+      throw EnvoyException(
+          absl::StrCat("Invalid uri specified for redirection for custom response: ", *uri_));
+    }
+  }
+  if (redirect_action_ && redirect_action_->host_redirect_.empty() &&
+      redirect_action_->path_redirect_.empty()) {
+    throw EnvoyException("At least one of host_redirect and path_redirect needs to be specified");
   }
+  ASSERT((uri_ || redirect_action_) && !(uri_ && redirect_action_));
```
adisuissa, line 86: Can you add a comment explaining the invariant?
Should it be at the beginning of the function (not masked by the previous if statements)?

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
```diff
@@ -126,18 +159,26 @@ ::Envoy::Http::FilterHeadersStatus RedirectPolicy::encodeHeaders(
...
-  auto path_and_query = path_;
+  auto path_and_query = absolute_url.pathAndQueryParams();
   // Strip the #fragment from Location URI if it is present. Envoy treats
   // internal redirect as a new request and will reject it if the URI path
   // contains #fragment.
-  const auto fragment_pos = path_.find('#');
-  path_and_query = path_.substr(0, fragment_pos);
-
-  downstream_headers->setPath(path_);
+  const auto fragment_pos = path_and_query.find('#');
```
adisuissa, line 178: Will it make sense to reject '#' in the config parsing step?

adisuissa (commented):
Overall LGTM.
I guess some of my comments may be relevant to the old (refactored) code rather than yours, but decided to add them for discussion.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: We're checking here that `uri_` is not set. If it is, we should have errored out earlier during config load.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
pradeepcrao: Great point! Yes, it does make sense and we need both, since `#` could be in the original request and allow changing only host and not path for the redirect.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
adisuissa: The `if (!absolute_url.initialize(uri, false))` checks `uri` which can be either the member `uri_` or the returned value from `newPath()`, no?

pradeepcrao: We can't check the value of `newPath()` without the request header map, which is not available at config load.

adisuissa: I think there's a disconnect between the GFE_BUG checks and what the `if` statement checks. IIUC the `custom_response_invalid_uri` counter will increase, but the ENVOY_BUG may or may not trigger. Is this intentional?

source/common/http/utility.h
```diff
@@ -668,6 +669,29 @@ bool isSafeRequest(const Http::RequestHeaderMap& request_headers);
...
+  // Keep small members (bools and enums) at the end of class, to reduce alignment overhead.
+  const bool path_redirect_has_query_;
+  const bool https_redirect_;
+  const bool strip_query_;
+};
+
+/*
+ * Compute new path based on RedirectConfig.
+ */
+std::string newPath(::Envoy::OptRef<const RedirectConfig> redirect_config,
```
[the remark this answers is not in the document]
pradeepcrao: Changed to `newUri`.

source/common/http/utility.cc
```diff
@@ -1287,5 +1333,73 @@ Http::Code Utility::maybeRequestTimeoutCode(bool remote_decode_complete) {
...
+  }
+
+  if (redirect_config.has_value() && !redirect_config->host_redirect_.empty()) {
+    final_host = redirect_config->host_redirect_;
+  } else {
+    ASSERT(headers.Host());
+    final_host = processRequestHost(headers, final_scheme, final_port);
+  }
+
+  std::string final_path_value;
```
adisuissa, line 1366: Should this be declared in the nested "if -> else -> if"?
The other place I see it being used (line 1389) seems to be overriding that value.

source/common/http/utility.cc
```diff
@@ -1287,5 +1333,73 @@ Http::Code Utility::maybeRequestTimeoutCode(bool remote_decode_complete) {
...
+        final_path = redirect_config->path_redirect_;
+      }
+    }
+  } else {
+    final_path = headers.getPathValue();
+  }
+
+  if (!absl::StartsWith(final_path, "/")) {
+    final_path_value = absl::StrCat("/", final_path);
+    final_path = final_path_value;
```
adisuissa, line 1390: can probably be merged to 1 line.

source/extensions/http/custom_response/redirect_policy/redirect_policy.cc
adisuissa: I see, thanks for clarifying. Can you please add a comment here?

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
```diff
@@ -132,6 +143,67 @@ TEST_F(CustomResponseFilterTest, DontChangeStatusCode) {
...
+            ::Envoy::Http::FilterHeadersStatus::Continue);
+  // Verify we get the original response
+  EXPECT_EQ("200", response_headers.getStatusValue());
+  EXPECT_EQ(
+      1U,
+      stats_store_.findCounterByString("stats.custom_response_invalid_uri").value().get().value());
+}
+
+TEST_F(CustomResponseFilterTest, InvalidSchemeRedirect) {
+  // Create config with invalid  field
```
adisuissa, line 177: nit:
```suggestion
  // Create config with invalid scheme field.
```

adisuissa (commented):
Thanks, overall LGTM.
A few minor nits.

source/common/http/utility.cc
pradeepcrao: `final_path_value` should live beyond the lifetime of `final_path` as it's the backing store for that stringview. So it cannot be declared within the scope of the `if`.

source/common/http/utility.cc
pradeepcrao: `final_path` is a string_view, so cannot be assigned to a temporary.

test/extensions/filters/http/custom_response/custom_response_filter_test.cc
pradeepcrao: Done.

adisuissa (approved):
/lgtm api
LGTM, thanks!

yanavlasov (approved):

PR #25443, opened 2023-02-08
mobile: adding a config knob for building without listeners

alyssawilk:
Risk Level: low
Testing: common CI, manual testing
Docs Changes: n/a 
Release Notes:  n/a
part of https://github.com/envoyproxy/envoy-mobile/issues/2711

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -30,6 +30,7 @@ jobs:
             --action_env=LD_LIBRARY_PATH \
             --test_env=ENVOY_IP_TEST_VERSIONS=v4only \
             --test_output=all \
+            --define envoy_mobile_listener=disabled \
             $([ -z $GITHUB_TOKEN ] || echo "--config=remote-ci-linux") \
             --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
             //test/common/...
--- a/bazel/envoy_mobile_copts.bzl
+++ b/bazel/envoy_mobile_copts.bzl
@@ -1,8 +1,9 @@
 # DO NOT LOAD THIS FILE. Load envoy_build_system.bzl instead.
-load(":envoy_select.bzl", "envoy_select_admin_functionality", "envoy_select_enable_http3", "envoy_select_envoy_mobile_request_compression")
+load(":envoy_select.bzl", "envoy_select_admin_functionality", "envoy_select_enable_http3", "envoy_select_envoy_mobile_listener", "envoy_select_envoy_mobile_request_compression")
 
 # Compute the copts needed for Envoy Mobile libraries that don't use Envoy's main library wrappers.
 def envoy_mobile_copts(repository):
     return envoy_select_admin_functionality(["-DENVOY_ADMIN_FUNCTIONALITY"], repository) + \
            envoy_select_enable_http3(["-DENVOY_ENABLE_QUIC"], repository) + \
+           envoy_select_envoy_mobile_listener(["-DENVOY_MOBILE_ENABLE_LISTENER"], repository) + \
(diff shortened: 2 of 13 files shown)


mobile/library/common/config/config.cc
```diff
@@ -575,6 +575,9 @@ stats_sinks: *stats_sinks
           reloadable_features:
             always_use_v6: *force_ipv6
             skip_dns_lookup_for_proxied_requests: *skip_dns_lookup_for_proxied_requests
+#ifndef ENVOY_MOBILE_ENABLE_LISTENER
```
jpsim, line 578: Does this work in a string literal like this?

mobile/library/common/engine_common.cc
```diff
@@ -4,11 +4,35 @@
 #include "source/common/runtime/runtime_impl.h"
 
 namespace Envoy {
+namespace {
+
+bool hasApiBootstrap(const envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
```
jpsim, line 9: Should this be named `useApiListener`?

mobile/library/cc/BUILD
```diff
@@ -12,7 +12,7 @@ envoy_cc_library(
     hdrs = [
         "engine_builder.h",
     ],
-    copts = select({
+    copts = envoy_mobile_copts("@envoy") + select({
```
jpsim, line 15: Why is this needed? We should already be getting these from `envoy_cc_library`, no?

alyssawilk: doesn't need it. ended up scrapping this entire buld file and plumbing a different way to catch tests not using the config template.  That said it'd still a draft so feel free to ignore until I've gotten to clean ups

alyssawilk:
OK, finally all cleaned up.  Whoever gets it gets it :-)

.github/workflows/core.yml
```diff
@@ -30,6 +30,7 @@ jobs:
             --action_env=LD_LIBRARY_PATH \
             --test_env=ENVOY_IP_TEST_VERSIONS=v4only \
             --test_output=all \
+            --define envoy_mobile_listener=disabled \
```
jpsim, line 33: Should we flip this in the compile_time_options cc_build CI job?

bazel/envoy_mobile_copts.bzl
```diff
@@ -1,8 +1,9 @@
 # DO NOT LOAD THIS FILE. Load envoy_build_system.bzl instead.
-load(":envoy_select.bzl", "envoy_select_admin_functionality", "envoy_select_enable_http3", "envoy_select_envoy_mobile_request_compression")
+load(":envoy_select.bzl", "envoy_select_admin_functionality", "envoy_select_enable_http3", "envoy_select_envoy_mobile_listener", "envoy_select_envoy_mobile_request_compression")
 
 # Compute the copts needed for Envoy Mobile libraries that don't use Envoy's main library wrappers.
 def envoy_mobile_copts(repository):
     return envoy_select_admin_functionality(["-DENVOY_ADMIN_FUNCTIONALITY"], repository) + \
            envoy_select_enable_http3(["-DENVOY_ENABLE_QUIC"], repository) + \
+           envoy_select_envoy_mobile_listener(["-DENVOY_MOBILE_ENABLE_LISTENER"], repository) + \
```
jpsim, line 8: This is fine here, but it could be in the `envoy_copts` macro since we don't need this defined for `swift_library`, `objc_library` or the JNI `cc_library`s.

jpsim (approved):

.github/workflows/core.yml
alyssawilk: we could as well, but eventually I want to make it the default so I largely wanted it somewhere I could easily regression test and I was also worried about it not playing well with other options.

PR #25454, opened 2023-02-09
mobile: minor fixes to C++ engine builder

jpsim:
* Quote `dns_preresolve_hostnames_` values
* Don't set `EngineBuilder::setAppId` twice

Commit Message:
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


alyssawilk:
Does DNS preresolve hostnames not work without quotes?  If not I'll definiely make sure we add an e2e test

jpsim:
> Does DNS preresolve hostnames not work without quotes? If not I'll definiely make sure we add an e2e test

Given that it's YAML I think it's fine without quotes, but this seemed slightly more resilient.

PR #25456, opened 2023-02-09
[balsa] Add Http1ProtocolOptions field to override HTTP/1 parser.

bencebeky:
If present, force http-parser (if value is false) or BalsaParser (if value is true).  If not present, parser is selected based on envoy.reloadable_features.http1_use_balsa_parser.

Commit Message: [balsa] Add Http1ProtocolOptions field to override HTTP/1 parser.
Additional Description:
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/api/envoy/config/core/v3/protocol.proto
+++ b/api/envoy/config/core/v3/protocol.proto
@@ -259,7 +259,7 @@ message HttpProtocolOptions {
   google.protobuf.UInt32Value max_requests_per_connection = 6;
 }
 
-// [#next-free-field: 9]
+// [#next-free-field: 10]
 message Http1ProtocolOptions {
   option (udpa.annotations.versioning).previous_message_type =
       "envoy.api.v2.core.Http1ProtocolOptions";
@@ -350,6 +350,11 @@ message Http1ProtocolOptions {
   // (inferred if not present), host (from the host/:authority header) and path
   // (from first line or :path header).
   bool send_fully_qualified_url = 8;
+
+  // If set, force HTTP/1 parser: BalsaParser if true, http-parser if false.
+  // If unset, HTTP/1 parser is selected based on
+  // envoy.reloadable_features.http1_use_balsa_parser.
+  optional bool use_balsa_parser = 9;
 }
 
 message KeepaliveSettings {
--- a/envoy/http/codec.h
+++ b/envoy/http/codec.h
@@ -492,6 +492,10 @@ struct Http1Settings {
 
   // If true, Envoy will send a fully qualified URL in the firstline of the request.
   bool send_fully_qualified_url_{false};
+
+  // If true, BalsaParser is used for HTTP/1 parsing; if false, http-parser is
+  // used.
+  bool use_balsa_parser_{false};
 };
 
 /**
--- a/source/common/http/http1/codec_impl.cc
+++ b/source/common/http/http1/codec_impl.cc
@@ -513,7 +513,7 @@ ConnectionImpl::ConnectionImpl(Network::Connection& connection, CodecStats& stat
       processing_trailers_(false), handling_upgrade_(false), reset_stream_called_(false),
       deferred_end_stream_headers_(false), dispatching_(false), max_headers_kb_(max_headers_kb),
       max_headers_count_(max_headers_count) {
-  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.http1_use_balsa_parser")) {
+  if (codec_settings_.use_balsa_parser_) {
     parser_ = std::make_unique<BalsaParser>(type, this, max_headers_kb_ * 1024, enableTrailers());
   } else {
     parser_ = std::make_unique<LegacyHttpParserImpl>(type, this);
@@ -692,7 +692,7 @@ Envoy::StatusOr<size_t> ConnectionImpl::dispatchSlice(const char* slice, size_t 
   const ParserStatus status = parser_->getStatus();
   if (status != ParserStatus::Ok && status != ParserStatus::Paused) {
     absl::string_view error = Http1ResponseCodeDetails::get().HttpCodecError;
-    if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.http1_use_balsa_parser")) {
+    if (codec_settings_.use_balsa_parser_) {
       if (parser_->errorMessage() == "headers size exceeds limit" ||
           parser_->errorMessage() == "trailers size exceeds limit") {
         error_code_ = Http::Code::RequestHeaderFieldsTooLarge;
--- a/source/common/http/http1/settings.cc
+++ b/source/common/http/http1/settings.cc
@@ -3,6 +3,7 @@
 #include "envoy/http/header_formatter.h"
 
 #include "source/common/config/utility.h"
+#include "source/common/runtime/runtime_features.h"
 
 namespace Envoy {
 namespace Http {
@@ -29,6 +30,13 @@ Http1Settings parseHttp1Settings(const envoy::config::core::v3::Http1ProtocolOpt
         factory);
     ret.header_key_format_ = Http1Settings::HeaderKeyFormat::StatefulFormatter;
     ret.stateful_header_key_formatter_ = factory.createFromProto(*header_formatter_config);
+  }
+
+  if (config.has_use_balsa_parser()) {
+    ret.use_balsa_parser_ = config.use_balsa_parser();
+  } else {
+    ret.use_balsa_parser_ =
+        Runtime::runtimeFeatureEnabled("envoy.reloadable_features.http1_use_balsa_parser");
   }
 
   return ret;


api/envoy/config/core/v3/protocol.proto
```diff
@@ -350,6 +350,11 @@ message Http1ProtocolOptions {
   // (inferred if not present), host (from the host/:authority header) and path
   // (from first line or :path header).
   bool send_fully_qualified_url = 8;
+
+  // If set, force HTTP/1 parser: BalsaParser if true, http-parser if false.
+  // If unset, HTTP/1 parser is selected based on
+  // envoy.reloadable_features.http1_use_balsa_parser.
+  optional bool use_balsa_parser = 9;
```
alyssawilk, line 357: I'd be inclined to have a sub-message for migration that's not-implemented-hide, so we can actually remove it after all the work

source/common/http/http1/codec_impl.cc
```diff
@@ -513,7 +513,7 @@ ConnectionImpl::ConnectionImpl(Network::Connection& connection, CodecStats& stat
       processing_trailers_(false), handling_upgrade_(false), reset_stream_called_(false),
       deferred_end_stream_headers_(false), dispatching_(false), max_headers_kb_(max_headers_kb),
       max_headers_count_(max_headers_count) {
-  if (Runtime::runtimeFeatureEnabled("envoy.reloadable_features.http1_use_balsa_parser")) {
+  if (codec_settings_.use_balsa_parser_) {
```
alyssawilk, line 516: should we stick with a global here so we can enable globally or locally?
actually if we use a Boolean instead of bool in the proto,we can check for presence and 
if not present default to flag
if present, use value

api/envoy/config/core/v3/protocol.proto
bencebeky: Done.

source/common/http/http1/codec_impl.cc
bencebeky: Changing to BoolValue, thanks for the suggestion.

bencebeky:
> Exciting! first pass thoughts

Thank you.  Addressed comments and also fixed :codec_impl_test, PTAL.

api/envoy/config/core/v3/protocol.proto
```diff
@@ -350,6 +350,12 @@ message Http1ProtocolOptions {
   // (inferred if not present), host (from the host/:authority header) and path
   // (from first line or :path header).
   bool send_fully_qualified_url = 8;
+
+  // [#not-implemented-hide:] Hiding so that field can be removed after BalsaParser is rolled out.
+  // If set, force HTTP/1 parser: BalsaParser if true, http-parser if false.
+  // If unset, HTTP/1 parser is selected based on
+  // envoy.reloadable_features.http1_use_balsa_parser.
+  google.protobuf.BoolValue use_balsa_parser = 9;
```
alyssawilk, line 358: Hm, so this is a better place to put it but I had kinda wanted a wrapper proto 
message MigrationOptions {
BoolValue use_baslsa
BoolValue use_oghttp2
BoolValue use_uhv
}
so then we'd have one base MigrationOptions and one thing to removed.  I'm going to say leave this as-is and let @envoyproxy/api-shepherds and @adisuissa decide what makes more sense.

envoy/http/codec.h
```diff
@@ -492,6 +492,10 @@ struct Http1Settings {
 
   // If true, Envoy will send a fully qualified URL in the firstline of the request.
   bool send_fully_qualified_url_{false};
+
+  // If true, BalsaParser is used for HTTP/1 parsing; if false, http-parser is
+  // used.
+  bool use_balsa_parser_{false};
```
alyssawilk, line 498: Do we have an issue tracking migration and deprecation?   It'd be nice to link sometihng both here and in the API comments.

source/common/http/http1/settings.cc
```diff
@@ -31,6 +32,13 @@ Http1Settings parseHttp1Settings(const envoy::config::core::v3::Http1ProtocolOpt
     ret.stateful_header_key_formatter_ = factory.createFromProto(*header_formatter_config);
   }
 
+  if (config.has_use_balsa_parser()) {
```
alyssawilk, line 35: unit test?

alyssawilk (commented):
Sorry for the delay!
/wait

envoy/http/codec.h
bencebeky: Yes, #21245.  Thanks for the reminder.

source/common/http/http1/settings.cc
bencebeky: Done, thanks for the reminder.

alyssawilk (approved):
cool LGTM though it might be worth clarifying in the PR and please definitely clarify in yuor PR description this only affects the parser for upstream connections

alyssawilk:
+Matt for !google pass

bencebeky:
> cool LGTM though it might be worth clarifying in the PR and please definitely clarify in yuor PR description this only affects the parser for upstream connections

Is that the case?  http-parser and Balsa are used for parsing both requests and responses.  If the proto only applies to upstream connections, does that mean the runtime flag only applies to upstream connections too?

alyssawilk:
The API field is in UpstreamHttpProtocolOptions

The idea AFIK was to be able to set it on a per-customer basis.
Downstream, there's a single listener so not the granularity.
So upstream you have proto -> runtime guard.  Downstream it's runtime guard as there's no API to set the proto field.

alyssawilk:
(arguably we could have a network filter which did an override per-vip but I don't think we easily can per config)

bencebeky:
> The API field is in UpstreamHttpProtocolOptions
> 
> The idea AFIK was to be able to set it on a per-customer basis. Downstream, there's a single listener so not the granularity. So upstream you have proto -> runtime guard. Downstream it's runtime guard as there's no API to set the proto field.

Sorry, I'm still confused.  I'm adding the field `use_balsa_parser` to message `Http1ProtocolOptions` (lines 262--353), not `UpstreamHttpProtocolOptions` (lines 99--124).  And after this PR the only non-test place where the runtime flag is queried is `parseHttp1Settings()`, used both for upstream and downstream, and it takes an `envoy::config::core::v3::Http1ProtocolOptions` argument in both cases.  Is it the case that Envoy can only have one `Http1ProtocolOptions` for downstream (because there is only one listener), but separate ones for each upstream backend group?

api/envoy/config/core/v3/protocol.proto
bencebeky: This sounds like a reasonable option.  Note that AFAIK UHV is controlled by a compile-time flag, therefore it will be not possible to override behavior with a proto field.  So we could only group `use_balsa` and `use_oghttp2` together.

bencebeky: The UHV transition document mentions the possibility of multiple UHV runtime flags added in the near future, one for each behavioral change.  If that will be the case, a third option would be to group those into one message, but leave `use_balsa` and `use_oghttp2` at the `Http1ProtocolOptions` level.

alyssawilk:
Oh gah, I got my proto files mixed up, you're totally right.  clearly my brain is still not back from vacation:-/

api/envoy/config/core/v3/protocol.proto
mattklein123: I assume we want a config flag here as you don't want to test this globally? Seems as shame to add config that we ultimately will remove? I guess if we hide it it's OK. Can you also add the "work in progress" annotation to make it very clear that it can be removed later?

adisuissa: I thought that uhv is going to be an [extension](https://github.com/envoyproxy/envoy/blob/b812ed619961c5289802ded5df1df27beb77fbb2/api/envoy/extensions/upstreams/http/v3/http_protocol_options.proto#L176).
The changes to balsa may be similar, as long as we assume that both balsa and http-parser could be used.

If the plan is to migrate and remove the old parser, and this migration needs to be at the protocol-options-level, then I agree that a "work-in-progress" tag just for balsa makes sense here.

bencebeky: Thank you for the suggestion.  Adding `work_in_progress`.

mattklein123 (dismissed):

adisuissa (commented):
Thanks!
/lgtm api

zuercher:
Error seems to be related to this PR:
```
[ RUN      ] Protocols/ProtocolIntegrationTest.LargeRequestMethod/IPv4_Http2Downstream_HttpUpstreamBalsaParserNghttp2NoDeferredProcessing
recv: [IB_EXPECT_CONTINUATION]
[2023-03-03 17:50:25.314][17][critical][assert] [external/envoy/test/integration/http_integration.cc:557] assert failure: result. Details: Timed out waiting for end of stream.
```

alyssawilk:
oh oops
/wait on CI

bencebeky:
> Error seems to be related to this PR:
> 
> ```
> [ RUN      ] Protocols/ProtocolIntegrationTest.LargeRequestMethod/IPv4_Http2Downstream_HttpUpstreamBalsaParserNghttp2NoDeferredProcessing
> recv: [IB_EXPECT_CONTINUATION]
> [2023-03-03 17:50:25.314][17][critical][assert] [external/envoy/test/integration/http_integration.cc:557] assert failure: result. Details: Timed out waiting for end of stream.
> ```

Yes.  It is also related to #25714, which got included in the lastest upstream merge of this PR.  I wonder if the issue is that integration tests override runtime flags, and but that can be overriden by config with this PR.

mattklein123 (approved):

alyssawilk (approved):

PR #25457, opened 2023-02-09
mobile: update ObjC/Swift APIs to take arrays instead of strings

jpsim:
This mirrors what was done to builders for other languages in #25297 and #25259.

Commit Message:
Additional Description:
Risk Level: Low
Testing: Unit tests & Lyft apps
Docs Changes: Updated
Release Notes: Updated
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/test/swift/EngineBuilderTests.swift
+++ b/mobile/test/swift/EngineBuilderTests.swift
@@ -367,13 +367,13 @@ final class EngineBuilderTests: XCTestCase {
   func testAddingVirtualClustersAddsToConfigurationWhenRunningEnvoy() {
     let expectation = self.expectation(description: "Run called with expected data")
     MockEnvoyEngine.onRunWithConfig = { config, _ in
-      XCTAssertEqual("[test]", config.virtualClusters)
+      XCTAssertEqual(["test"], config.virtualClusters)
       expectation.fulfill()
     }
 
     _ = EngineBuilder()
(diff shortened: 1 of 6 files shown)


mobile/test/swift/EngineBuilderTests.swift
```diff
@@ -367,13 +367,13 @@ final class EngineBuilderTests: XCTestCase {
...
     MockEnvoyEngine.onRunWithConfig = { config, _ in
-      XCTAssertEqual("[test]", config.virtualClusters)
+      XCTAssertEqual(["test"], config.virtualClusters)
       expectation.fulfill()
     }
 
     _ = EngineBuilder()
       .addEngineType(MockEnvoyEngine.self)
-      .addVirtualClusters("[test]")
+      .addVirtualClusters(["test"])
```
alyssawilk, line 376: optional:  you may want to update these to valid yaml while you're in here.  Bonus points for adding a test for multiple virtual clusters to make sure the joining is correct now we know that's a thing

alyssawilk (approved):

mobile/test/swift/EngineBuilderTests.swift
jpsim: When I enable yaml validation for tests I'll be forced to do that there, which I'm doing very shortly.

PR #25459, opened 2023-02-09
mobile: add option to compare yaml for Swift / Objective-C builders

jpsim:
_Note: This is a fairly large PR, but I've broken it down into more focused commits to help reviewers. I'm happy to split off any of those commits into a dedicated PR if anyone prefers._

The goal of this PR is to allow consumers to validate the YAML produced by the ObjC/Swift engine builders with the YAML produced by the C++ engine builder in order to more safely allow migrating to it following a short testing period. This mirrors what was done in #25297 & #25259 for the Java/Kotlin APIs.

Unlike the approach taken for the Java/Kotlin APIs, the ObjC/Swift validation only occurs if `setExperimentalValidateYAMLCallback(...)` is configured on the builder. Another difference is that there's no release assertion if the comparison fails, which means we can be more confident in rolling this out to a small pool of testers internally, rather than only validating in unit tests.

In my brief testing with Lyft's iOS apps, the yaml comparison succeeds.

Lastly, this comparison isn't yet enabled for all the unit/integration tests or example apps in this repo yet because some of them aren't producing valid configurations and will need to be updated. I didn't want to further bloat this PR by also doing that here.

Commit Message: mobile: add option to compare yaml for Swift / Objective-C builders
Additional Description:
Risk Level: Moderate for Envoy Mobile, off by default but other changes are included, like moving to Objective-C++
Testing: Unit tests & Lyft's iOS apps
Docs Changes: None, no public stable API changes
Release Notes: None, no public stable API changes
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


mobile/library/cc/engine_builder.h
```diff
@@ -104,6 +104,7 @@ class EngineBuilder {
 
   EngineSharedPtr build();
 
+  bool generateBootstrapAndCompare(std::string yaml) const;
```
RyanTheOptimist, line 107: nit: Can you add a comment for this method? Also instead of std::string we should probably pass in absl::string_view. Ditto for generateBootstrapAndCompareForTests.

mobile/library/cc/engine_builder.cc
```diff
@@ -498,11 +498,14 @@ std::string EngineBuilder::generateConfigStr() const {
   return config_str;
 }
 
+bool EngineBuilder::generateBootstrapAndCompare(std::string yaml) const {
+  return generatedStringMatchesGeneratedBoostrap(yaml, *generateBootstrap());
+}
+
 std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap>
 EngineBuilder::generateBootstrapAndCompareForTests(std::string yaml) const {
   auto bootstrap = generateBootstrap();
```
RyanTheOptimist, line 507: nit: This predates you, but can we use the explicit type here, per the style guide.

"The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type"
https://google.github.io/styleguide/cppguide.html#Type_deduction

mobile/library/objective-c/BUILD
```diff
@@ -1,16 +1,16 @@
...
     "EnvoyEngine.h",
 ])
 
-objc_library(
+envoy_objc_library(
     name = "envoy_engine_objc_lib",
     srcs = [
         "EnvoyConfiguration.h",
-        "EnvoyConfiguration.m",
+        "EnvoyConfiguration.mm",
```
RyanTheOptimist, line 13: Out of curiosity what's the difference between the ".m" and ".mm" extension?

RyanTheOptimist (approved):
This looks great! I'm not an obj-c expert but it all looks correct (ish?) to me.

Thanks for doing this!!

mobile/library/objective-c/BUILD
jpsim: `.m` is Objective-C. `.mm` is Objective-C++.

mobile/library/cc/engine_builder.cc
jpsim: I guess it's a personal choice, but I find

```c++
auto bootstrap
```

clearer than

```c++
std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap> bootrap
```

But I'll take your word for it :)

mobile/library/cc/engine_builder.h
jpsim: It looks like all the other functions in this file use `std::string` instead of `absl::string_view`. Is this one special or should this have been done for all APIs?

mobile/library/objective-c/BUILD
RyanTheOptimist: Oh, of course! Thanks.

mobile/library/cc/engine_builder.h
RyanTheOptimist: It depends a bit. We want to avoid copies where possible. For a setter where we have a member which is an std::string, a copy is inevitable. But the question is where it happens. If we take an std::string by value in the method then we can do `member_ = std::move(arg)`. This avoids a copy in the method. If the caller has a string, the caller can do `set_foo(std:move(data))` to avoid a copy (assuming they have the data in a variable that can be moved). So for setters, std::string is the best choice.

But for an argument where we don't need to copy the data, passing `absl::string_view` (or `const std::string&`, though if the caller has a char* that might require a copy) is better. In this case, I think we don't need to copy `yaml` so passing a copy of it is not necessary. 

What do you think?

mobile/library/cc/engine_builder.cc
RyanTheOptimist: Oh, for some reason I thought this was a string. auto is definitely shorter :)

Happy to leave this up to you, especially since this predates your PR.

jpsim: I updated it.

mobile/library/cc/engine_builder.cc
```diff
@@ -49,7 +49,7 @@ void insertCustomFilter(const std::string& filter_config, std::string& config_te
 // Note that updates to the config.cc bootstrap will require a matching update in
 // generateBootstrap() below
 bool generatedStringMatchesGeneratedBoostrap(
-    const std::string& config_str, const envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
+    const absl::string_view& config_str, const envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
```
RyanTheOptimist, line 52: nit: no reason for & here since string_view is basically a reference already.

RyanTheOptimist (approved):

mobile/library/cc/engine_builder.h
jpsim: Yeah avoiding copies is a good reason to do this, I'm still getting used to C++'s model for implicit copying. Thanks for explaining.

RyanTheOptimist (approved):
Thanks!

mobile/library/cc/engine_builder.h
RyanTheOptimist: You're welcome! C++ is just astonishingly complex and with new versions of C++ the best practices keep changing. So on good days I feel like I almost maybe have a handle on it. Sorta :)

Issue #25458, opened 2023-02-09
[balsa] Handle empty field names


bencebeky:
Title: [balsa] Handle empty field names

Description:

This is a bug in Balsa, which is protected by envoy.reloadable_features.http1_use_balsa_parser.  This flag defaults to false and no deployment is known to enable it, so the impact of this bug is minimal.  When the flag is false, another HTTP/1 parser called http-parser is used instead of Balsa.

`ConnectionImpl::completeCurrentHeader()` normally clears `current_header_value_` at https://github.com/envoyproxy/envoy/blob/48364f3adc34048501fb261aef74fb8fada0b0c9/source/common/http/http1/codec_impl.cc#L549, then it asserts that it is empty at https://github.com/envoyproxy/envoy/blob/48364f3adc34048501fb261aef74fb8fada0b0c9/source/common/http/http1/codec_impl.cc#L564.  However, if `current_header_field_` is empty but `current_header_value_` is not when `completeCurrentHeader()` is called, then it does not clear `current_header_value_`, and the assertion fails.

When parsing a field line (header line) with empty name, for example, ": foo", http-parser signals an error when encountering the colon, as it is an invalid token for the field name.  Balsa, however, parser the line as one with an empty name.  If this is followed by another field line, then `onHeaderFieldImpl() calls `completeCurrentHeader()`, and the assertion fails.

However, if the field line with empty name is the last one, then `completeCurrentHeader()` is never called, so there is no assertion failure.  This is because there is a check for empty names at https://quiche.googlesource.com/quiche/+/afbd9cde746c772422d20744b3d61597e29267ef/quiche/balsa/balsa_frame.cc#604, but unfortunately it happens after the `BalsaVisitorInterface::OnHeader()` call at
https://quiche.googlesource.com/quiche/+/afbd9cde746c772422d20744b3d61597e29267ef/quiche/balsa/balsa_frame.cc#458.

This bug was discovered by the fuzz test `//test/common/http/http1:http1_codec_impl_fuzz_test`, with a field line `:scheme: https`, which was correctly parsed as empty name, and value `scheme: https`.

PR #25475, opened 2023-02-09
[balsa] Use ProcessHeaders() instead of OnHeader().

bencebeky:
BalsaVisitorInterface provides two ways of processing headers: OnHeader() is called for each header line, and ProcessHeaders() and ProcessTrailers() are called with the entire header/trailer block.  All of these methods are only called after the entire header block or trailer block is parsed, so using OnHeader() does not actually allow early processing of headers.

The major difference is that some header validation happens after OnHeader() is called, for example, check for empty header names (both for headers and trailers), multiple Content-Length headers, multiple Transfer-Encoding headers, and the presence of Content-Length and Transfer-Encoding in the same header block (not for trailers).  (Some of these test can be disabling using HttpValidationPolicy.)  If the headers fail, for example, because of an empty field name or because of multiple Transfer-Encoding headers, ProcessHeaders() or ProcessTrailers() are not called.

One problem with using OnHeader() is that an assertion in ConnectionImpl::completeCurrentHeader() can fail if the header name is empty, see issue #25458.  Another problem is that header names and values are copied immediately, using unnecessary memory in case an error is detected later.  This PR migrates BalsaParser from using OnHeader() to ProcessHeaders() and ProcessTrailers().

This PR also adds two tests.  The first one, the case of empty header name, would cause an assertion failure without the change.  The second one would pass either way, because the error from Balsa is propagated properly to ConnectionImpl.  Note that the error message is slightly different with http-parser, because http-parser does not have a check for Transfer-Encoding, but ConnectionImpl::onHeadersCompleteImpl() signals an error, because the combined value of "chunked,chunked" does not match "chunked".

Relevant issues: #25458, #21245.

Commit Message: [balsa] Use ProcessHeaders() instead of OnHeader().
Additional Description:
Risk Level: low, changed code is protected by existing default-false runtime flag.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/common/http/http1/codec_impl_test.cc
+++ b/test/common/http/http1/codec_impl_test.cc
@@ -4306,5 +4306,74 @@ TEST_P(Http1ServerConnectionImplTest, LineFolding) {
   }
 }
 
+// Regression test for https://github.com/envoyproxy/envoy/issues/25458.
+TEST_P(Http1ServerConnectionImplTest, EmptyFieldName) {
+  initialize();
+  InSequence s;
+
+  StrictMock<MockRequestDecoder> decoder;
+  Http::ResponseEncoder* response_encoder = nullptr;
+  EXPECT_CALL(callbacks_, newStream(_, _))
+      .WillOnce(Invoke([&](ResponseEncoder& encoder, bool) -> RequestDecoder& {
+        response_encoder = &encoder;
+        return decoder;
+      }));
+  EXPECT_CALL(decoder,
+              sendLocalReply(Http::Code::BadRequest, "Bad Request", _, _, "http1.codec_error"));
+
+  Buffer::OwnedImpl buffer("GET / HTTP/1.1\r\n"
+                           ": value-for-empty-key\r\n"
+                           "Host: host\r\n\r\n");
+
+  auto status = codec_->dispatch(buffer);
+
+  EXPECT_TRUE(isCodecProtocolError(status));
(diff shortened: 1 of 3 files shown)


test/common/http/http1/codec_impl_test.cc
```diff
@@ -4306,5 +4306,74 @@ TEST_P(Http1ServerConnectionImplTest, LineFolding) {
...
+  StrictMock<MockRequestDecoder> decoder;
+  Http::ResponseEncoder* response_encoder = nullptr;
+  EXPECT_CALL(callbacks_, newStream(_, _))
+      .WillOnce(Invoke([&](ResponseEncoder& encoder, bool) -> RequestDecoder& {
+        response_encoder = &encoder;
+        return decoder;
+      }));
+  if (parser_impl_ == Http1ParserImpl::BalsaParser) {
+    EXPECT_CALL(decoder,
+                sendLocalReply(Http::Code::BadRequest, "Bad Request", _, _, "http1.codec_error"));
```
diannahu, line 4354: No action, just want to say that this error message feels clearer/an improvement over the error message resulting from http-parser.

diannahu (approved):
Looks great and thanks for the clear PR message!

birenroy (approved):
Looks like a nice improvement.

bencebeky:
Thank you both for the review.

/assign @KBaichoo

KBaichoo (approved):
Nice! lgtm

PR #25481, opened 2023-02-09
mobile: Use true not "true" as an arg to set_bool_value()

RyanTheOptimist:
mobile: Use true not "true" as an arg to set_bool_value()


alyssawilk (approved):

RyanTheOptimist:
cc: @yanavlasov

Issue #25487, opened 2023-02-10
Circuit breakers trigger outlier detection


Hexta:
*Title*: *Circuit breakers trigger outlier detection*

*Description*:
Max requests circuit breaker triggers outlier detections to eject unrelated endpoints.

*Repro steps*:
- Enable CB and OD.
- Trigger CB.

*Admin and Stats Output*:
```
cluster.a.outlier_detection.ejections_active: 1
cluster.a.outlier_detection.ejections_consecutive_5xx: 0
cluster.a.outlier_detection.ejections_detected_consecutive_5xx: 0
cluster.a.outlier_detection.ejections_detected_consecutive_gateway_failure: 0
cluster.a.outlier_detection.ejections_detected_consecutive_local_origin_failure: 1
cluster.a.outlier_detection.ejections_detected_failure_percentage: 0
cluster.a.outlier_detection.ejections_detected_local_origin_failure_percentage: 0
cluster.a.outlier_detection.ejections_detected_local_origin_success_rate: 0
cluster.a.outlier_detection.ejections_detected_success_rate: 0
cluster.a.outlier_detection.ejections_enforced_consecutive_5xx: 0
cluster.a.outlier_detection.ejections_enforced_consecutive_gateway_failure: 0
cluster.a.outlier_detection.ejections_enforced_consecutive_local_origin_failure: 1
cluster.a.outlier_detection.ejections_enforced_failure_percentage: 0
cluster.a.outlier_detection.ejections_enforced_local_origin_failure_percentage: 0
cluster.a.outlier_detection.ejections_enforced_local_origin_success_rate: 0
cluster.a.outlier_detection.ejections_enforced_success_rate: 0
cluster.a.outlier_detection.ejections_enforced_total: 1
cluster.a.outlier_detection.ejections_overflow: 0
cluster.a.outlier_detection.ejections_success_rate: 0
cluster.a.outlier_detection.ejections_total: 0
```

*Config*:
```yaml
resources:
  - "@type": type.googleapis.com/envoy.config.cluster.v3.Cluster
    name: a
    circuit_breakers:
      thresholds:
        - priority: DEFAULT
          max_requests: 0

    outlier_detection:
      interval: 1s
      
      consecutive_local_origin_failure: 1
      enforcing_consecutive_local_origin_failure: 100
      split_external_local_origin_errors: true

      base_ejection_time: 200s
      max_ejection_time: 200s
    load_assignment:
      cluster_name: a
          
      endpoints:
[...]

PR #25488, opened 2023-02-10
router: Fix outlier detection false positives

Hexta:
Commit Message: Fix outlier detection false positives caused by max requests circuit breaker.
Risk Level: low
Testing: integration tests
Release Notes: Done.
Fixes: #25487

### DIFF
--- a/source/common/router/router.cc
+++ b/source/common/router/router.cc
@@ -1288,6 +1288,12 @@ void Filter::onUpstreamReset(Http::StreamResetReason reset_reason,
 
   const bool dropped = reset_reason == Http::StreamResetReason::Overflow;
 
+  // Ignore upstream reset caused by a resource overflow.
+  // Currently, circuit breakers can only produce this reset reason.
+  // It means that this reason is cluster-wise, not upstream-related.
+  // Therefore removing an upstream in the case of an overloaded cluster
+  // would make the situation even worse.
+  // https://github.com/envoyproxy/envoy/issues/25487
   if (!dropped) {
     // TODO: The reset may also come from upstream over the wire. In this case it should be
     // treated as external origin error and distinguished from local origin error.
--- a/source/common/router/router.cc
+++ b/source/common/router/router.cc
@@ -1286,18 +1286,21 @@ void Filter::onUpstreamReset(Http::StreamResetReason reset_reason,
                    *callbacks_, Http::Utility::resetReasonToString(reset_reason),
                    transport_failure_reason);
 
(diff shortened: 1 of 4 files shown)


test/integration/circuit_breakers_integration_test.cc
```diff
@@ -0,0 +1,65 @@
...
+class CircuitBreakersIntegrationTest : public HttpProtocolIntegrationTest {
+public:
+  void initialize() override { HttpProtocolIntegrationTest::initialize(); }
+};
+
+INSTANTIATE_TEST_SUITE_P(Protocols, CircuitBreakersIntegrationTest,
+                         testing::ValuesIn(HttpProtocolIntegrationTest::getProtocolTestParams()),
+                         HttpProtocolIntegrationTest::protocolTestParamsToString);
+
+TEST_P(CircuitBreakersIntegrationTest, CircuitBreakersWithOutlierDetection) {
```
mattklein123, line 17: Can you add some comment on what this test is testing as it's a little complicated?

source/common/router/router.cc
```diff
@@ -1286,18 +1286,21 @@ void Filter::onUpstreamReset(Http::StreamResetReason reset_reason,
...
 
-  // TODO: The reset may also come from upstream over the wire. In this case it should be
-  // treated as external origin error and distinguished from local origin error.
-  // This matters only when running OutlierDetection with split_external_local_origin_errors
-  // config param set to true.
-  updateOutlierDetection(Upstream::Outlier::Result::LocalOriginConnectFailed, upstream_request,
-                         absl::nullopt);
+  const bool dropped = reset_reason == Http::StreamResetReason::Overflow;
+
+  if (!dropped) {
```
mattklein123, line 1291: Can you add a small comment as to why we are not doing outlier detection for circuit breakers? I think this makes sense but it's worth it to add some comment to the code so the next person coming along has the history.

mattklein123 (changes requested):
Generally LGTM, thanks. Some small comments. Also, I think this is worth a release note. Thank you.

/wait

mattklein123 (approved):
Thanks!

Hexta:
@mattklein123 Thx for the review. 
Unfortunately, some integration test was failed.
As I see this test is failing periodically for unclear reasons.

```
test/integration/redirect_integration_test.cc:704
Expected equality of these values:
1
test_server_->counter("cluster.cluster_0.upstream_internal_redirect_succeeded_total") ->value()
Which is: 0
```

Issue #25426, opened 2023-02-08
Patch release (7/2) post-mortem


phlax:
There were a few issues pushing through patch releases that made the process less than smooth

- github/azp ubuntu-18 brownout (timed perfectly to coincide with ssl release!)
- flakiness on release branches
- changelog fail on 1.22 branch
- publishing on the tag (and not the release commit)

overall i think to get the benefit of automating release we need to iron out all the things that prevent it from working automatically

#### ubuntu brownout

putting aside the wisdom of gh/azp policy here this is a bit of an issue for us

currently both the 1.22 and 1.23 branches use the ubuntu-18 image

1.23 can be trivially updated i think #25422 

1.22 blocks because CodeQL expects glibc2.32 - this is not so easy to resolve

The EOL date for the image/ci availability is april 1st which is ~conveniently just before 1.22 also goes EOL

**i would propose that we just set EOL date for 1.22 to April 1st as it will be difficult to update after**

#### flakiness in CI on release branches

this is a pretty big issue for anyone working on release branches, in Envoy or mirror repos.

the release branches run ~uncached and so, particularly for the ARM build, the chance of failure is much higher than when we test in postsubmit `main` - rerunning the 1.5 hour ARM run 3 times on a branch, twice over as it then fails on the tag run, can add ~10 hours to the process

there is some things i can do to mitigate and i will trawl `main` for any ci fixes i can find that can be backported

going forward i think we need to do **daily uncached CI runs with the medium-term goal of addressing flakes more quickly and trying to make these actually pass**

#### changelog fail

this was my bad - i pre-tested most things (as i couldnt run it in ci beforehand) - but unfortunately not the changelog test 

this is mostly a peculiarity of this branch, so less likely an issue going forward, but frustrating all the same

#### publishing on the tag

this has been discussed elsewhere #23773 

i think for branches 1.23+ that have the newer publishing code this is high priority

the current way doubles the time it takes and doubles the chances of flake failure

phlax:
cc @envoyproxy/envoy-maintainers 

cc @cancecen

phlax:
closing this off as the issues are being addressed and it is superseeded by #25830

PR #25505, opened 2023-02-13
ci: Reorganize release pipelines

phlax:
This PR

- merges the `docker` and `publish` stages in CI
- splits the `distribution` jobs between `publish` and `verify`

partial fix for issues raised in #25426 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -360,95 +360,6 @@ stages:
     - bash: |
         echo "linux_arm64 released"
 
-- stage: distribution
-  dependsOn: ["env", "linux_arm64", "linux_x64"]
-  jobs:
-  - job: package_x64
-    dependsOn: []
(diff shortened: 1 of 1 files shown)


phlax:
@lizan this is the first of several PRs to move towards publishing from the release branch rather than the tagged commit

phlax:
this will need some changes to the github `required` jobs and may cause some temporary disruption on prs

phlax:
in order for azp to be able to evaluate stages that contain jobs with references to securefiles in presubmit (even if the job is disabled in presubmit) i have had to do some convoluted things:

- add an empty key for presubmit
- add a pipeline variable to presubmit pointing to the empty key
- add a pipeline variable to postsubmit pointing to the real key

phlax:
this is going to be a bit of a challenge backporting - but i think its important that we do for 1.23+ so we can continue to update the publishing code together, and be using/maintaining similar pipelines

lizan (approved):

PR #25506, opened 2023-02-13
deps: Bump `actions/checkout` -> v3 (dependabot + fixes)

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim:
Odd that the `android_build` and child jobs never started?

image

jpsim:
Actually it looks like there's a mistake in the yaml configs:

> [Invalid workflow file: .github/workflows/ios_build.yml#L21](https://github.com/envoyproxy/envoy/actions/runs/4161743606/workflow)
The workflow is not valid. .github/workflows/ios_build.yml (Line: 21, Col: 7): Unexpected value 'fetch-depth' .github/workflows/ios_build.yml (Line: 47, Col: 7): Unexpected value 'fetch-depth'

image

phlax:
could be me force pushing - i can look further if its blocking it somehow

phlax:
> Actually it looks like there's a mistake in the yaml configs:

ah that would explain it - let me update ...

phlax:
hmm - wierd - im not seeing the error - its seems correct https://github.com/actions/checkout#usage and is working everywhere else

phlax:
merged `main` altho dont see any reason it would merge badly - lets see

jpsim:
It seems to only be happening on macOS, the Linux jobs are running fine.

> [Invalid workflow file: .github/workflows/android_build.yml#L21](https://github.com/envoyproxy/envoy/actions/runs/4164747882/workflow)
The workflow is not valid. .github/workflows/android_build.yml (Line: 21, Col: 7): Unexpected value 'fetch-depth' .github/workflows/android_build.yml (Line: 53, Col: 7): Unexpected value 'fetch-depth'

https://github.com/envoyproxy/envoy/actions/runs/4164747882

phlax:
nm - i think it needs a `with` not sure why it does/not work - but ill update with correct syntax ...

jpsim:
I see this is now done automatically:

> Adding repository directory to the temporary git global config as a safe directory
> /usr/local/bin/git config --global --add safe.directory /Users/runner/work/envoy/envoy

So can we remove the explicit steps that do that in our workflow configs?

phlax:
> So can we remove the explicit steps that do that in our workflow configs?

yep - i was intending to follow up with that

phlax:
@jpsim would you prefer if i just added that here ?

jpsim:
Yes can you please do that here? The time to cut that out should be offset by not having the overhead of creating/reviewing/waiting for CI on a new PR.

jpsim:
> can we just revert this and land the action update ?

Yes of course.

phlax:
argh - that still failed - reverted!

jpsim (approved):

PR #25518, opened 2023-02-13
mobile: remove chromium from main Android AAR

jpsim:
Before:

```console
$ ./bazelw build --config release-android android_dist
...
INFO: Build completed successfully, 1 total action
$ rg chromium bazel-bin/library/kotlin/io/envoyproxy/envoymobile | wc -l
5
```

After:

```console
$ ./bazelw build --config release-android android_dist
...
INFO: Build completed successfully, 1 total action
$ rg chromium bazel-bin/library/kotlin/io/envoyproxy/envoymobile | wc -l
0
```

Slack discussion: https://envoyproxy.slack.com/archives/CKQ2LK23G/p1676228502975469

Commit Message:
Additional Description:
Risk Level: Only risk is to Envoy Mobile on Android, either using Cronvoy or the standard Java/Kotlin APIs.
Testing: See above
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/bazel/android_artifacts.bzl
+++ b/mobile/bazel/android_artifacts.bzl
@@ -238,7 +238,7 @@ def _create_classes_jar(name, manifest, android_library):
         classes_dir=$$(mktemp -d)
         echo "Creating classes.jar from $(SRCS)"
         pushd $$classes_dir
-        unzip $$original_directory/$(SRCS) "io/envoyproxy/*" "org/chromium/net/*" "META-INF/" > /dev/null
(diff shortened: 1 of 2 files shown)


jpsim:
> If we remove the cronet code, will that break functionality?

Ah, yes it looks like it. Thanks for pointing that out. It seems to be related to cert validation. Does this need to live in Cronet or can it be shared?

jpsim:
I'll close this PR because we can't just rip it out like this. I wasn't aware we were relying on this.

PR #25526, opened 2023-02-13
mobile: assert on Swift yaml comparison

jpsim:
So that CI will fail if a mismatch is introduced.

Note that this doesn't assert that comparison succeeds for cases that the C++ engine builder doesn't support today, such as when using direct responses or raw yaml configurations.

Commit Message:
Additional Description:
Risk Level: Low, test only
Testing: Ran all Swift tests and sample apps, confirmed that `YAML comparison succeeded` is logged and the assertion isn't triggered. Also confirmed that the assertion crashes the app/test if there's any mismatch.
Docs Changes: None, not a user-facing change
Release Notes: None, not a user-facing change
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/examples/swift/hello_world/ViewController.swift
+++ b/mobile/examples/swift/hello_world/ViewController.swift
@@ -1,3 +1,4 @@
+@_spi(YAMLValidation)
 import Envoy
 import UIKit
 
@@ -17,13 +18,20 @@ final class ViewController: UITableViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    let engine = EngineBuilder()
+    let engine = YAMLValidatingEngineBuilder()
       .addLogLevel(.debug)
       .addPlatformFilter(DemoFilter.init)
       .addPlatformFilter(BufferDemoFilter.init)
       .addPlatformFilter(AsyncDemoFilter.init)
-      // swiftlint:disable:next line_length
-      .addNativeFilter(name: "envoy.filters.http.buffer", typedConfig: "{\"@type\":\"type.googleapis.com/envoy.extensions.filters.http.buffer.v3.Buffer\",\"max_request_bytes\":5242880}")
+      .addNativeFilter(
+        name: "envoy.filters.http.buffer",
+        typedConfig: """
+            {\
+            "@type":"type.googleapis.com/envoy.extensions.filters.http.buffer.v3.Buffer",\
+            "max_request_bytes":5242880\
+            }
+            """
+      )
       .setOnEngineRunning { NSLog("Envoy async internal setup completed") }
       .addStringAccessor(name: "demo-accessor", accessor: { return "PlatformString" })
       .setEventTracker { NSLog("Envoy event emitted: \($0)") }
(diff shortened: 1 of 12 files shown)


mobile/examples/swift/hello_world/ViewController.swift
```diff
@@ -17,13 +18,20 @@ final class ViewController: UITableViewController {
...
 
-    let engine = EngineBuilder()
+    let engine = YAMLValidatingEngineBuilder()
       .addLogLevel(.debug)
       .addPlatformFilter(DemoFilter.init)
       .addPlatformFilter(BufferDemoFilter.init)
       .addPlatformFilter(AsyncDemoFilter.init)
-      // swiftlint:disable:next line_length
-      .addNativeFilter(name: "envoy.filters.http.buffer", typedConfig: "{\"@type\":\"type.googleapis.com/envoy.extensions.filters.http.buffer.v3.Buffer\",\"max_request_bytes\":5242880}")
+      .addNativeFilter(
```
jpsim, line 26: This change wasn't needed but when I touched this file, I saw an opportunity to fix the SwiftLint warning and align with the way this is set up in the experimental app.

mobile/library/objective-c/EnvoyConfiguration.mm
```diff
@@ -310,29 +302,20 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
...
-  builder.addMaxConnectionsPerHost(self.maxConnectionsPerHost);
-  builder.addH2ConnectionKeepaliveIdleIntervalMilliseconds(
-      self.h2ConnectionKeepaliveIdleIntervalMilliseconds);
-  builder.addH2ConnectionKeepaliveTimeoutSeconds(self.h2ConnectionKeepaliveTimeoutSeconds);
-  builder.addStatsFlushSeconds(self.statsFlushSeconds);
-
-  builder.setAppVersion([self.appVersion toCXXString]);
-  builder.setAppId([self.appId toCXXString]);
-  builder.setDeviceOs("iOS");
+  for (EnvoyNativeFilterConfig *nativeFilterConfig in self.nativeFilterChain) {
```
jpsim, line 305: I aligned the order in which things are done here to match what's in `-[EnvoyConfiguration resolveTemplate:]`.

mobile/library/swift/YAMLValidatingEngineBuilder.swift
```diff
@@ -0,0 +1,15 @@
+/// An engine builder that validates the YAML configuration and asserts on failure.
+/// Not part of the stable public API.
+///
+/// - returns: The `EngineBuilder`.
+@_spi(YAMLValidation)
+public func YAMLValidatingEngineBuilder() -> EngineBuilder {
```
jpsim, line 6: Even though this is `public`, it's marked as `@_spi(YAMLValidation)` so that it's not available in the public API for users who `import Envoy`.

alyssawilk (approved):

PR #25265, opened 2023-01-31
Reduce overhead by changing alignment

DiazAlan:
Signed-off-by: AlanDiaz <diazalan@google.com>

Commit Message: Reduce overhead via alignment
Additional Description: Moved smaller values like enums, int32 and bools to the end of the class to save on memory via alignment
Risk Level: Low

PR #25529, opened 2023-02-13
Alignment change to save memory in cluster info

DiazAlan:
Additional Description: Moving around a couple of variables to save a small amount of memory in Cluster Info.
Sizeof before changes: 2632
Sizeof after changes: 2624
Risk Level: Low


yurykats (approved):

yurykats (approved):

wbpcode (approved):
Hmmm, I am a little worry about this may be break the dependencies between fields of cluster info. But after doing a quick check, this should be safe? 🤔 

cc @mattklein123 for final check or merge.

DiazAlan:
I've pushed a change like this in the same file just a few weeks ago. I don't see why it would cause issues.

https://github.com/envoyproxy/envoy/pull/25265

adisuissa (approved):
LGTM, thanks!

PR #25531, opened 2023-02-14
mobile: always set metadata values as strings

jpsim:
Otherwise if we set something that looks like an integer for app version it'll lead to C++ yaml comparison mismatches.

Commit Message:
Additional Description:
Risk Level:
Testing: Existing unit tests were updated, and Lyft's downstream tests are now passing when yaml validation is enabled.
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/library/cc/engine_builder.cc
+++ b/mobile/library/cc/engine_builder.cc
@@ -366,7 +366,7 @@ std::string EngineBuilder::generateConfigStr() const {
          fmt::format("{}s", h2_connection_keepalive_timeout_seconds_)},
         {
             "metadata",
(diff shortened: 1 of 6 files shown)


alyssawilk (approved):
huh, I wonder why it wasn't a problem for Java.  maybe that didn't quote either?

jpsim:
It’s only a problem if you pass something that yaml assumes isn’t a string (int, float, etc). We had a Swift test do that downstream but nothing doing that through Java.

PR #25532, opened 2023-02-14
aws: support fetching credentials from the AWS credentials file

### LINKED ISSUE #25445: AWS Request Signing Filter: support getting credentials from a file
*Title*: Support getting credentials from a file in the AWS Request Signing Filter.

*Description*:
Currently, the filter only supports getting credentials from environment variables, EC2 instance metadata, and ECS task metadata. The latter two won't work for Envoys deployed outside of AWS, and the first one won't work for environments that don't allow secrets to be stored in env vars. I propose adding a new credential provider that reads credentials from a file.

The format of the file could be something simple like this:
```
{
    "AccessKeyId": "ACCESS_KEY_ID",
    "Expiration": "EXPIRATION_DATE",
    "SecretAccessKey": "SECRET_ACCESS_KEY",
    "Token": "SECURITY_TOKEN_STRING"
}
```
This matches the format of credentials retrieved from instance/task metadata, and would be easier to parse than AWS's standard [credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html). Note: the filter does not use the AWS SDK.

Let me know if this feature would be accepted, in which case I'd be happy to contribute it. Thanks!

*Relevant Links*:
Supported credential providers: https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/aws/credentials_provider_impl.h

michaelfinch:
Signed-off-by: Michael Finch <mfinch@squareup.com>

Commit Message: support fetching credentials from the AWS credentials file
Additional Description: add a new credentials provider that attempts to get credentials from the [AWS credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)
Risk Level: Low
Testing: unit tests
Docs Changes: updated docs
Release Notes: added
Fixes #25445


michaelfinch:
cc @suniltheta

changelogs/current.yaml
```diff
@@ -20,6 +20,9 @@ minor_behavior_changes:
 - area: http2
   change: |
     Request authorities are now validated with a library function from QUICHE rather than nghttp2. This behavior change can be reverted by setting ``envoy.reloadable_features.http2_validate_authority_with_quiche`` to ``false``.
+- area: aws
+  change: |
+    Added support for fetching credentials from the AWS credentials file. The order of credential fetching methods is now: 1) environment variables, 2) the credentials file, 3) instance metadata, 4) task metadata.
```
suniltheta, line 25: I think the order of 3 or 4 depends on whether the Envoy is running on Ec2 or ECS. So we can probably just mention that Environment is used before credentials file

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +72,85 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+    ENVOY_LOG(error, "Error opening credentials file {}", credentials_file);
+    return;
+  }
+
+  std::string access_key_id, secret_access_key, session_token;
+  std::string profile_start = fmt::format("[{}]", profile);
+
+  bool found_profile = false;
+  std::string line;
+  while (std::getline(file, line)) {
```
suniltheta, line 119: can we trim the line for whitespace and new line character?
`continue` if line is empty.

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +72,85 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+      }
+    }
+  }
+
+  ENVOY_LOG(debug, "Found following AWS credentials for profile '{}' in {}: {}={}, {}={}, {}={}",
+            profile, credentials_file, AWS_ACCESS_KEY_ID, access_key_id, AWS_SECRET_ACCESS_KEY,
+            secret_access_key.empty() ? "" : "*****", AWS_SESSION_TOKEN,
+            session_token.empty() ? "" : "*****");
+
+  cached_credentials_ = Credentials(access_key_id, secret_access_key, session_token);
```
suniltheta, line 150: should we raise error if AWS_ACCESS_KEY_ID & AWS_SECRET_ACCESS_KEY are not found?

In the docs we are saying `The fields 'aws_access_key_id', 'aws_secret_access_key',
   and 'aws_session_token' must be defined for the profile in the credentials file for this method to succeed`. So we need to fails if minimum requirements are not met. I believe `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` are mandatory based on this line https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/aws/credentials_provider_impl.cc#L241 

```
    if (credentials.accessKeyId() && credentials.secretAccessKey()) {
      return credentials;
    }
```

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -253,6 +335,9 @@ DefaultCredentialsProviderChain::DefaultCredentialsProviderChain(
   ENVOY_LOG(debug, "Using environment credentials provider");
   add(factories.createEnvironmentCredentialsProvider());
 
+  ENVOY_LOG(debug, "Using credentials file credentials provider");
+  add(factories.createCredentialsFileCredentialsProvider(api));
```
suniltheta, line 339: I am contemplating whether we want to include this credentials provider by default.

Because if the file is not present then `CredentialsFileCredentialsProvider::refresh()` is called again and again because last_updated_ is not updated in failure scenario.


Prefer if don't include it by default so that it won't affect other providers. Include only if `~/.aws/credentials` or AWS_SHARED_CREDENTIALS_FILE is present. This means after Envoy has initialized we will not be able to updated the credentials in the background without restarting Envoy.

suniltheta: welcome to any other suggestions so that we don't affect performance of other credentials provider.

suniltheta (commented):
Thanks for making this change. I have done initial round of review. Raised some questions regarding the design choice. I am yet to go through the unit tests.

source/extensions/common/aws/credentials_provider_impl.cc
michaelfinch: That's a good point, thanks for raising it. These are the main options I can think of:
1. Add a field to the AwsRequestSigning proto to enable the credentials file. If it's enabled, we add it to the default chain.
2. Add CredentialsFileCredentialsProvider to the default chain iff the credentials file exists.
3. Add CredentialsFileCredentialsProvider to the default chain, but have refreshIfNeeded return immediately unless AWS_SHARED_CREDENTIALS_FILE and AWS_PROFILE are both defined.

I prefer option 1 because it won't impact any existing users of the filter, and since fetching from the credentials file can be expensive (especially if credentials aren't found) it requires the user to explicitly enable it. We can document that it's expensive and should only be enabled if the user is sure credentials will be found in it. What do you think?

suniltheta: I prefer option 2 because that is the approach with least intrusion. Today for other credentials provider we don't expect any configuration on AWS Lambda or AWS Request signing filter for deciding on the credentials provider. It is inline with current implementation of tasks credentials provider where it expects a particular ENV variable to be available. In credentials file credentials provider we expect a file to be present on default path or on the path specified by `AWS_SHARED_CREDENTIALS_FILE`. This is checked only once at the filter initialization.

This comes with a drawback that the file has to be present before Envoy consumes the configuration.

For host & task credentials provider no impact if the file is not present. But if the file is present (for some other purpose) and users don't expect the Envoy to pick the credentials specified on the file then we can add a runtime config to disable this feature.

michaelfinch: Another drawback is that if the credentials file is ever deleted for some reason then the filter will start trying to read the file on every request, though I imagine that's unlikely.

The change you suggested is reasonable and sounds good to me, but let me float one more idea: if `CredentialsFileCredentialsProvider::extractCredentials` returns early, we cache an empty set of credentials and update `last_updated_`. That way if we fail to fetch credentials for any reason, we will cache empty credentials for an hour and the next provider in the chain will be called. This won't be that impactful if the credentials file doesn't exist since we will only attempt to read the file once an hour. The benefits are that the file doesn't have to be present before Envoy consumes the config, it accommodates the credentials file being deleted, and it accommodates the value of `AWS_SHARED_CREDENTIALS_FILE` changing.

suniltheta: This is a good option as well. Practically it could be the same as option 2 because there is no case where Envoy would be considered operational given it went 1 hour without any credentials. Make sure we add debug logs around credentials file not found so that user won't be confused what's happening under the hood.

Would you still be willing to add a runtime configuration so that we have an option to disable credential file credentials provider if we have to? Especially for those use case where it already has credentials file in the Envoy container path but instead prefer to use Instance metadata or task metadata. After 1 year time period (deprecation of runtime configs) when we decide to remove the runtime configuration we would have registered if Envoy users needed to use the runtime config then we can provide a permanent xDS config option to disable the credentials file credentials provider.

michaelfinch: Yep, I will add the runtime config to both the Request Signing and Lambda filters per your suggestion.

adisuissa:
@lavignes @tonya11en can you PTAL as codeowners?

suniltheta:
@michaelfinch we might also need to add a file (`~/.aws/config`) based region provider https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/aws/region_provider_impl.h

Because both `~/.aws/config` & `~/.aws/credentials` are expected to be provided together.

michaelfinch:
> @michaelfinch we might also need to add a file (`~/.aws/config`) based region provider https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/aws/region_provider_impl.h
> 
> Because both `~/.aws/config` & `~/.aws/credentials` are expected to be provided together.

Are you saying that we should add support for getting the region from the AWS config file in `RegionProvider`? If so, which filters do you want this to be used in? Right now `RegionProvider` is only used in the `grpc_credentials` filter.

I'm picking this work back up this week so should have feedback addressed in the next few days.

suniltheta:
> Right now RegionProvider is only used in the grpc_credentials filter.

I see. Then not required for now.

suniltheta (changes requested):
some minor change

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +73,92 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+  // errors extracting credentials from it (e.g. if the credentials file doesn't exist).
+  last_updated_ = api_.timeSource().systemTime();
+
+  std::ifstream file(credentials_file);
+  if (!file) {
+    ENVOY_LOG(error, "Error opening credentials file {}", credentials_file);
+    return;
+  }
+
+  std::string access_key_id, secret_access_key, session_token;
```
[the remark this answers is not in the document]
michaelfinch: Are you okay with me sticking with `std::string`? I tried switching to `absl::string_view` and the tests passed on my Mac, but it failed on other platforms in CI, and it's slow for me to debug.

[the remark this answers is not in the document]
suniltheta: sure, no problem.

changelogs/current.yaml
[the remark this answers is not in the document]
michaelfinch: Done.

docs/root/configuration/http/http_filters/aws_request_signing_filter.rst
```diff
@@ -48,6 +48,23 @@ Example filter configuration:
     - exact: x-amzn-trace-id
 
 
+Credentials
+-----------
+
+The filter attempts to fetch credentials from a few different places, in the following order:
+
+1. Environment variables. The environment variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN must be set
```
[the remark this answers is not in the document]
michaelfinch: Done.

docs/root/configuration/http/http_filters/aws_request_signing_filter.rst
```diff
@@ -48,6 +48,23 @@ Example filter configuration:
...
+   for this method to succeed.
+
+2. The AWS credentials file. The environment variables AWS_SHARED_CREDENTIALS_FILE and AWS_PROFILE are respected if they are set,
+   else the file '~/.aws/credentials' and profile 'default' will be used. The fields 'aws_access_key_id', 'aws_secret_access_key',
+   and 'aws_session_token' must be defined for the profile in the credentials file for this method to succeed. These credentials
+   are cached for 1 hour.
+
+3. EC2 instance metadata. These credentials are cached for 1 hour.
+
+4. ECS task metadata. These credentials are cached for 1 hour or until they expire.
```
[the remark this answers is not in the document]
michaelfinch: Updated the wording.

test/extensions/common/aws/credentials_provider_impl_test.cc
```diff
@@ -17,6 +17,24 @@ namespace Extensions {
 namespace Common {
 namespace Aws {
 
+const char CREDENTIALS_FILE[] = "test-credentials.json";
```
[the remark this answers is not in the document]
michaelfinch: Yes, this is the file we write credentials to during tests.

source/extensions/common/aws/credentials_provider_impl.cc
michaelfinch: Done.

source/extensions/common/aws/credentials_provider_impl.cc
michaelfinch: Updated the docs to make the behavior more clear.

source/extensions/common/aws/credentials_provider_impl.cc
michaelfinch: This is done.

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +73,92 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+
+    if (found_profile) {
+      // Stop reading once we find the start of the next profile.
+      if (absl::StartsWith(line, "[")) {
+        break;
+      }
+
+      std::vector<std::string> parts = absl::StrSplit(line, absl::MaxSplits('=', 1));
+      if (parts.size() == 2) {
+        auto upper = StringUtil::toUpper(parts[0]);
```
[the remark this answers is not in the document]
michaelfinch: Done.

test/extensions/common/aws/credentials_provider_impl_test.cc
```diff
@@ -62,6 +82,89 @@ TEST_F(EvironmentCredentialsProviderTest, NoSessionToken) {
...
+
+  // Confirm that the empty credentials are cached even after we switch
+  // to a legitimate profile with valid credentials.
+  setUpTest(CREDENTIALS_FILE_CONTENTS, "profile1");
+  credentials = provider_.getCredentials();
+  EXPECT_FALSE(credentials.accessKeyId().has_value());
+  EXPECT_FALSE(credentials.secretAccessKey().has_value());
+  EXPECT_FALSE(credentials.sessionToken().has_value());
+}
+
```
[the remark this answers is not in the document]
michaelfinch: Thanks for the tip, I've added a test.

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -1,13 +1,19 @@
...
 #include "envoy/common/exception.h"
 
+#include "source/common/common/fmt.h"
 #include "source/common/common/lock_guard.h"
 #include "source/common/http/message_impl.h"
 #include "source/common/http/utility.h"
 #include "source/common/json/json_loader.h"
 #include "source/extensions/common/aws/utility.h"
 
+#include "absl/strings/match.h"
```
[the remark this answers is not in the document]
michaelfinch: Nope. Removed.

docs/root/configuration/http/http_filters/_include/aws_credentials.rst
```diff
@@ -0,0 +1,16 @@
+Credentials
+-----------
+
+The filter uses a few different methods to obtain an AWS access key ID, AWS secret access key, and AWS session token. It moves through
+the methods in the order described below, stopping when one of the methods returns an access key ID and a secret access key (the session
```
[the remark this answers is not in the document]
michaelfinch: Fixed.

docs/root/configuration/http/http_filters/_include/aws_credentials.rst
```diff
@@ -0,0 +1,16 @@
...
+token is optional).
+
+1. Environment variables. The environment variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN are used.
+
+2. The AWS credentials file. The environment variables AWS_SHARED_CREDENTIALS_FILE and AWS_PROFILE are respected if they are set, else
+   the file '~/.aws/credentials' and profile 'default' are used. The fields 'aws_access_key_id', 'aws_secret_access_key', and
+   'aws_session_token' defined for the profile in the credentials file are used. These credentials are cached for 1 hour.
+
+3. Either EC2 instance metadata or ECS task metadata. For EC2 instance metadata, the fields 'AccessKeyId', 'SecretAccessKey', and
+   'Token' are used, and credentials are cached for 1 hour. For ECS task meatadata, the fields AccessKeyId', 'SecretAccessKey', and
```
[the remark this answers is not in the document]
michaelfinch: Fixed.

michaelfinch:
@suniltheta I believe I've addressed all of your feedback.

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +72,93 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+      AWS_SHARED_CREDENTIALS_FILE, DEFAULT_AWS_SHARED_CREDENTIALS_FILE);
+  const auto profile = getEnvironmentVariableOrDefault(AWS_PROFILE, DEFAULT_AWS_PROFILE);
+
+  extractCredentials(credentials_file, profile);
+}
+
+void CredentialsFileCredentialsProvider::extractCredentials(const std::string& credentials_file,
+                                                            const std::string& profile) {
+  // Update last_updated_ now so that even if this function returns before successfully
+  // extracting credentials, we'll cache an empty set of credentials. This prevents envoy
```
suniltheta, line 105: technically we are not caching an empty set of credentials. We just return without setting anything on `cached_credentials_`.

Do we want to mention that after `REFRESH_INTERVAL` (1 hour) period an attempt will be made to fetch credentials again from the credentials file?

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -249,10 +339,17 @@ Credentials CredentialsProviderChain::getCredentials() {
 
 DefaultCredentialsProviderChain::DefaultCredentialsProviderChain(
     Api::Api& api, const MetadataCredentialsProviderBase::MetadataFetcher& metadata_fetcher,
-    const CredentialsProviderChainFactories& factories) {
+    const CredentialsProviderChainFactories& factories, const bool enable_credentials_file) {
   ENVOY_LOG(debug, "Using environment credentials provider");
   add(factories.createEnvironmentCredentialsProvider());
 
+  if (enable_credentials_file) {
```
suniltheta, line 346: Is it possible to check `Runtime::runtimeFeatureEnabled("envoy.reloadable_features.enable_aws_credentials_file");` directly here instead of passing the bool flag? Any advantage in passing the flag over?

We can also enable and disable the feature flag while unit testing. The extensions or plugin don't have to know whether it has to enable or disable the feature to read from credentials file. It is handled under the hood.

When we finally remove the feature flag the change will be just in this constructor.

test/extensions/common/aws/credentials_provider_impl_test.cc
```diff
@@ -62,6 +91,114 @@ TEST_F(EvironmentCredentialsProviderTest, NoSessionToken) {
...
+TEST_F(CredentialsFileCredentialsProviderTest, SpacesBetweenParts) {
+  setUpTest(CREDENTIALS_FILE_CONTENTS, "profile4");
+
+  const auto credentials = provider_.getCredentials();
+  EXPECT_EQ("profile4_access_key", credentials.accessKeyId().value());
+  EXPECT_EQ("profile4_secret", credentials.secretAccessKey().value());
+  EXPECT_EQ("profile4_token", credentials.sessionToken().value());
+}
+
+TEST_F(CredentialsFileCredentialsProviderTest, EmptyCredsCached) {
```
suniltheta, line 178: I guess the name of the test can change from `EmptyCredsCached` to something that indicate that the credentials expired?

Or other option is to add a separate unit test that will load a valid credentials first instead of an empty credentials. And then after credentials expired if the credentials file or profile is updated it will pick the new credentials.

suniltheta (changes requested):
Few minor comments

source/extensions/common/aws/credentials_provider_impl.cc
michaelfinch: This has been addressed.

source/extensions/common/aws/credentials_provider_impl.cc
michaelfinch: No advantage, your way is better. Updated.

test/extensions/common/aws/credentials_provider_impl_test.cc
michaelfinch: This has been addressed.

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +73,93 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+        } else if (key == AWS_SECRET_ACCESS_KEY) {
+          secret_access_key = val;
+        } else if (key == AWS_SESSION_TOKEN) {
+          session_token = val;
+        }
+      }
+    }
+  }
+
+  ENVOY_LOG(error, "Found following AWS credentials for profile '{}' in {}: {}={}, {}={}, {}={}",
```
suniltheta, line 154: do you want to make this debug level instead of error?

source/extensions/common/aws/credentials_provider_impl.cc
```diff
@@ -62,13 +73,93 @@ Credentials EnvironmentCredentialsProvider::getCredentials() {
...
+                                                            const std::string& profile) {
+  // Update last_updated_ now so that even if this function returns before successfully
+  // extracting credentials, this function won't be called again until after the REFRESH_INTERVAL.
+  // This prevents envoy from attempting and failing to read the credentials file on every request
+  // if there are errors extracting credentials from it (e.g. if the credentials file doesn't exist).
+  last_updated_ = api_.timeSource().systemTime();
+
+  std::ifstream file(credentials_file);
+  if (!file) {
+    ENVOY_LOG(error, "Error opening credentials file {}", credentials_file);
```
suniltheta, line 113: do you want to make this debug log instead of error?

suniltheta (changes requested):
Thanks for picking this up. only 1 comment on log level.

michaelfinch:
The build appears to be stuck. I'm going to close and reopen this to see if that fixes it.

suniltheta (approved):
LGTM. I guess after some format fix we should pass CI

suniltheta (dismissed):
Thanks, LGTM

michaelfinch:
@suniltheta thanks. Any idea how to get the build unstuck? I tried creating a new branch + new empty commit in a draft PR and it gets stuck at the same place: 37 successful checks, 7 expected (with all 7 expected checks saying `Waiting for status to be reported`)

suniltheta:
Even I am unsure how to get the status from these CI actions. Let's check with @mattklein123

mattklein123:
Can you merge main and resolve the conflict and then I can take a quick look? Thanks.

/wait

suniltheta (approved):

mattklein123 (approved):

michaelfinch:
Thanks for merging.

PR #23071, opened 2022-09-12
http: making the admission control filter a dual filter

alyssawilk:
Risk Level: low
Testing: integration tested
Docs Changes: n/a
Release Notes: inline
Part of https://github.com/envoyproxy/envoy/issues/10455

PR #25535, opened 2023-02-14
ext_authz: make the ext_authz filter a dual filter

pianiststickman:
Used #23071 as a model.

Signed-off-by: Eugene Chan <eugenechan@google.com>

Risk Level: low
Testing: integration tested
Docs Changes: n/a
Release Notes: Makes the ext_authz filter a dual filter.
See also #10455

### DIFF
--- a/test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
+++ b/test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
@@ -1006,6 +1006,86 @@ body_format:
   cleanup();
 }
 
+TEST_P(ExtAuthzLocalReplyIntegrationTest, UpstreamTest) {
+  setUpstreamProtocol(Http::CodecType::HTTP1);
+  config_helper_.addConfigModifier([this](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
+    auto* ext_authz_cluster = bootstrap.mutable_static_resources()->add_clusters();
+    ext_authz_cluster->MergeFrom(bootstrap.static_resources().clusters()[0]);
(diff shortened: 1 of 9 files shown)


pianiststickman:
cc @alyssawilk @paul-r-gall

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
```diff
@@ -1006,6 +1006,86 @@ TEST_P(ExtAuthzLocalReplyIntegrationTest, DeniedHeaderTest) {
   cleanup();
 }
 
+TEST_P(ExtAuthzLocalReplyIntegrationTest, UpstreamTest) {
```
alyssawilk, line 1009: would it be plausible to change the whole test to dual mode?

alyssawilk (commented):
Very cool!

Can you do a scan of
source/docs/upstream_filters.md
And make sure you address any open issues?  For example a quick check of source/extensions/filters/http/ext_authz/ there's a call to downstreamCallbacks
For that we may need to do special handling to not crash, and also document in filter docs that the route will not be selected when changing headers in upstream filter mode.
/wait

test/extensions/filters/http/ext_authz/ext_authz_integration_test.cc
pianiststickman: This test is more or less the same as the existing downstream filter test. Are you suggesting to parameterize it?

alyssawilk: right, I'm saying we could run all the tests in both modes and maybe flush out other bugs

PR #25551, opened 2023-02-14
mobile: silence `-Wshorten-64-to-32` warnings when building ObjC

jpsim:
These started being reported when we switched to building ObjC++.

Commit Message: mobile: silence `-Wshorten-64-to-32` warnings when building ObjC
Additional Description:
Risk Level: Low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/mobile/bazel/apple.bzl
+++ b/mobile/bazel/apple.bzl
@@ -8,7 +8,7 @@ def envoy_objc_library(name, hdrs = [], visibility = [], data = [], deps = [], m
         name = name,
         srcs = srcs,
         hdrs = hdrs,
-        copts = envoy_mobile_copts("@envoy") + ["-ObjC++", "-std=c++17"],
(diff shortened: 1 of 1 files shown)


jpsim:
To be clear, these warnings aren't being reported in the ObjC/ObjC++ code, it's all in the envoy C++ code. For example:

```


envoy/envoy/common/union_string.h:223:34: warning: implicit conversion loses integer precision: 'absl::string_view::size_type' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
--
  | return getStrView(buffer_).size();
  | ~~~~~~ ~~~~~~~~~~~~~~~~~~~~^~~~~~


```

RyanTheOptimist (approved):

jpsim:
> Maybe we should file an issue, though I suspect we're unlikely to do this any time soon.

Filed https://github.com/envoyproxy/envoy/issues/25557, thanks for the suggestion!

PR #25552, opened 2023-02-14
mobile: add support for direct responses to C++ builder

jpsim:
And start validating yaml comparison for Swift direct response tests.

Skip C++ assertions for builder configurations that use direct responses since it's going to be a lot of work to generate that YAML, it's only ever used for tests and it's meant as a temporary transition helper.

This helps migrate the ObjC/Swift APIs off the custom yaml codepaths.

Commit Message:
Additional Description:
Risk Level: Low, test-only
Testing: Yes, existing Swift tests
Docs Changes: N/A, test-only
Release Notes: N/A, test-only
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


alyssawilk (commented):
can you give me a day on this one?
I'd really rather deprecate the API, and instead have a test-only JNI call that spins up a test backend.  If I can get it working in a day or two for kotlin, I can send that out as an alternate, and you can port the swift tests to that?

alyssawilk:
@jpsim have one more PR with test upstreams to move to Envoy's autonomous upstream from E-M's (base class allows more options) then do you want to try moving to that or shall I?
largely not sure if the swift tests need any custom responses beyond the defaultl

alyssawilk (approved):
per discussion LGTM but would love a TODO for "test migration and removing this :-P

jpsim:
Added a TODO in 5bb9972b2715893bfdfd3c5933c8eac64f9b7581.

alyssawilk (approved):

PR #25560, opened 2023-02-14
mobile: Add the stats prefix to the ADS xDS registration

abeyad:
The stats_prefix field in the GoogleGrpc config has a proto validation rule that the field must be populated and have a length of at least 1.

This change adds the stat_prefix field whenever the ADS builder API is used, and adds tests to validate it.

Lastly, this change makes the mobile_compile_time_options CI job turn on proto validation, so we can ensure we never commit invalid proto, even though mobile normally disables proto validation for binary size purposes. The default value for the PVG template is the empty string (https://github.com/bufbuild/protoc-gen-validate/blob/38260ee45796b420276ac925d826ecec8fc3e9a8/bazel/BUILD#L11), so the mobile_compile_time_options sets template-flavor to the empty string instead of the default `nop` for mobile.

### DIFF
--- a/.github/workflows/compile_time_options.yml
+++ b/.github/workflows/compile_time_options.yml
@@ -38,6 +38,7 @@ jobs:
             --define=admin_html=enabled \
             --define=envoy_mobile_request_compression=disabled \
             --@envoy//bazel:http3=False \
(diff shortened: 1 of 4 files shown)


RyanTheOptimist (approved):
Thanks for doing this!

jpsim (approved):

PR #25561, opened 2023-02-14
Create LbPolicyConfig class to store variant for LBConfigs

DiazAlan:
Additional Description: Since LBConfigs are one-ofs, storing them inside a std::variant will allow us to save a bit of space.
Sizeof(ClusterInfoImpl)
Before changes: 2624
After changes: 2592
Risk Level: Low

### DIFF
--- a/source/common/upstream/upstream_impl.cc
+++ b/source/common/upstream/upstream_impl.cc
@@ -961,8 +961,6 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
 }
 
 LBPolicyConfig::LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config) {
-  // TODO: Need to verify that lb type and the config for the policy are forced to match in
-  // a 1-1 relationship. If they are, change this to a switch case using config.lb_policy()
   switch (config.lb_config_case()) {
   case envoy::config::cluster::v3::Cluster::kRoundRobinLbConfig:
     lbPolicy_ = std::make_unique<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
--- a/source/common/upstream/upstream_impl.cc
+++ b/source/common/upstream/upstream_impl.cc
@@ -961,43 +961,33 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
 }
 
 LBPolicyConfig::LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config) {
-
-  // Original DST LB is not in the lb_policy enum so can't be included in the switch
+  // TODO: Need to verify that lb type and the config for the policy are forced to match in
+  // a 1-1 relationship. If they are, change this to a switch case using config.lb_policy()
+  if (config.has_round_robin_lb_config()) {
+    lbPolicy_ = std::make_unique<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
+        config.round_robin_lb_config());
+    return;
+  }
+  if (config.has_least_request_lb_config()) {
+    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>(
+        config.least_request_lb_config());
+    return;
+  }
+  if (config.has_ring_hash_lb_config()) {
+    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::RingHashLbConfig>(
+        config.ring_hash_lb_config());
+    return;
+  }
+  if (config.has_maglev_lb_config()) {
+    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
+        config.maglev_lb_config());
+    return;
+  }
   if (config.has_original_dst_lb_config()) {
     lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::OriginalDstLbConfig>(
         config.original_dst_lb_config());
     return;
   }
-
-  switch (config.lb_policy()) {
-    PANIC_ON_PROTO_ENUM_SENTINEL_VALUES;
-  case envoy::config::cluster::v3::Cluster::ROUND_ROBIN:
-    if (config.has_round_robin_lb_config()) {
-      lbPolicy_ = std::make_unique<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
-          config.round_robin_lb_config());
-    }
-    break;
-  case envoy::config::cluster::v3::Cluster::LEAST_REQUEST:
-    if (config.has_least_request_lb_config()) {
-      lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>(
-          config.least_request_lb_config());
-    }
-    break;
-  case envoy::config::cluster::v3::Cluster::RING_HASH:
-    if (config.has_ring_hash_lb_config()) {
-      lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::RingHashLbConfig>(
-          config.ring_hash_lb_config());
-    }
-    break;
-  case envoy::config::cluster::v3::Cluster::MAGLEV:
-    if (config.has_maglev_lb_config()) {
-      lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
-          config.maglev_lb_config());
-    }
-    break;
-  default:
-    break;
-  };
 }
 
 ClusterInfoImpl::ClusterInfoImpl(
--- a/source/common/upstream/upstream_impl.cc
+++ b/source/common/upstream/upstream_impl.cc
@@ -963,23 +963,23 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
 LBPolicyConfig::LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config) {
   switch (config.lb_config_case()) {
   case envoy::config::cluster::v3::Cluster::kRoundRobinLbConfig:
-    lbPolicy_ = std::make_unique<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
+    lb_policy_ = std::make_unique<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
         config.round_robin_lb_config());
     break;
   case envoy::config::cluster::v3::Cluster::kLeastRequestLbConfig:
-    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>(
+    lb_policy_ = std::make_unique<envoy::config::cluster::v3::Cluster::LeastRequestLbConfig>(
         config.least_request_lb_config());
     break;
   case envoy::config::cluster::v3::Cluster::kRingHashLbConfig:
-    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::RingHashLbConfig>(
+    lb_policy_ = std::make_unique<envoy::config::cluster::v3::Cluster::RingHashLbConfig>(
         config.ring_hash_lb_config());
     break;
   case envoy::config::cluster::v3::Cluster::kMaglevLbConfig:
-    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
+    lb_policy_ = std::make_unique<envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
         config.maglev_lb_config());
     break;
   case envoy::config::cluster::v3::Cluster::kOriginalDstLbConfig:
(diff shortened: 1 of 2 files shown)


source/common/upstream/upstream_impl.cc
```diff
@@ -960,6 +960,44 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
...
+    }
+    break;
+  case envoy::config::cluster::v3::Cluster::MAGLEV:
+    if (config.has_maglev_lb_config()) {
+      lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
+          config.maglev_lb_config());
+    }
+    break;
+  default:
+    break;
```
adisuissa, line 992: I don't think default is needed (should be taken care of the PANIC at the beginning).

source/common/upstream/upstream_impl.cc
```diff
@@ -960,6 +960,44 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
...
+  };
+
+  if (config.has_original_dst_lb_config()) {
+    lbPolicy_ = std::make_unique<envoy::config::cluster::v3::Cluster::OriginalDstLbConfig>(
+        config.original_dst_lb_config());
+  }
```
adisuissa, line 998: Should this be done before the switch (and immediately return)?
Please also add a comment of why it is not part of the switch.

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,80 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
   std::vector<UpstreamLocalAddress> upstream_local_addresses_;
 };
 
+/**
+ * Class for LBPolicies
+ * Uses a absl::variant to store pointers for the LBPolicy
+ */
+class LBPolicyConfig {
```
adisuissa, line 109: Good idea shielding the variant in a special class!

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,80 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
...
+public:
+  LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config);
+
+  OptRef<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig> lbRoundRobinConfig() const {
+    if (auto lb_ptr = absl::get_if<
+            std::unique_ptr<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>>(
+            &lbPolicy_)) {
+      return makeOptRefFromPtr<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig>(
+          (*lb_ptr).get());
```
adisuissa, line 118: OOC can the `makeOptRefFromPtr` be avoided here?
I see that in the original code `return *lb_round_robin_config_;` was used, so the question if returning `*lb_ptr` is possible in this case?

adisuissa (commented):
Thanks!
Overall LGTM, just left a few nits.

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,80 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
...
+/**
+ * Class for LBPolicies
+ * Uses a absl::variant to store pointers for the LBPolicy
+ */
+class LBPolicyConfig {
+public:
+  LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config);
+
+  OptRef<const envoy::config::cluster::v3::Cluster::RoundRobinLbConfig> lbRoundRobinConfig() const {
+    if (auto lb_ptr = absl::get_if<
```
pradeepcrao, line 114: nit: using `std::holds_alternative` might make the code more readable. 

Also, it's probably better to have a templated helper method to avoid the repeated code.

source/common/upstream/upstream_impl.cc
DiazAlan: Correct me if I'm wrong but OriginalDstLB is reserved in that enum so when I try to code it, it says it's not found. But it's still apart of the oneof in the cluster proto. But I'll move the check to before the switch.

source/common/upstream/upstream_impl.cc
DiazAlan: I put in the default because without it, I get a build error saying it also wants a case for Random, Cluster Provided and LB Policy Config. Should I just add these cases and leave them blank?

source/common/upstream/upstream_impl.h
DiazAlan: Done

source/common/upstream/upstream_impl.h
DiazAlan: This makes sense, I applied this change into the new helper function in the class.

source/common/upstream/upstream_impl.cc
```diff
@@ -960,6 +960,44 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
       config.has_http2_protocol_options(), validation_visitor);
 }
 
+LBPolicyConfig::LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config) {
+
+  switch (config.lb_policy()) {
+    PANIC_ON_PROTO_ENUM_SENTINEL_VALUES;
+  case envoy::config::cluster::v3::Cluster::ROUND_ROBIN:
```
pradeepcrao, line 974: Currently, `lb_policy` is not used to set the configs for the lb configs. I don't know why it was done that way (and seems like it could be a source for bugs). 
Could using `lb_policy` to set lb config cause some change in existing behaviour? Do they need to be set independently as is done currently?

DiazAlan: Talked offline about this a bit and we're not sure if it is enforced that they match. For now I changed them to if statements that check for the config to make it similar to how it behaved before. I opened up an issue [here](https://github.com/envoyproxy/envoy/issues/25583) to discuss it further with the community. If they are forced to match, then I'll come back and change it to a switch.

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,56 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
...
+ * Uses a absl::variant to store pointers for the LBPolicy
+ */
+class LBPolicyConfig {
+public:
+  LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config);
+
+  template <typename T> OptRef<const T> getConfig() const {
+    if (absl::holds_alternative<std::unique_ptr<const T>>(lbPolicy_)) {
+      if (absl::get<std::unique_ptr<const T>>(lbPolicy_) != nullptr)
+        return *(absl::get<std::unique_ptr<const T>>(lbPolicy_));
```
adisuissa, line 116: nit:
```suggestion
      if (auto config = absl::get<std::unique_ptr<const T>>(lbPolicy_); config != nullptr)
        return *config;
```

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,56 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
...
+/**
+ * Class for LBPolicies
+ * Uses a absl::variant to store pointers for the LBPolicy
+ */
+class LBPolicyConfig {
+public:
+  LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config);
+
+  template <typename T> OptRef<const T> getConfig() const {
+    if (absl::holds_alternative<std::unique_ptr<const T>>(lbPolicy_)) {
```
adisuissa, line 114: Thanks for refactoring this to a function.
With the use of the templated function, will it make sense to use `get_if` again in this case? cc @pradeepcrao

source/common/upstream/upstream_impl.cc
```diff
@@ -960,6 +960,36 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
       config.has_http2_protocol_options(), validation_visitor);
 }
 
+LBPolicyConfig::LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config) {
+  // TODO: Need to verify that lb type and the config for the policy are forced to match in
+  // a 1-1 relationship. If they are, change this to a switch case using config.lb_policy()
```
adisuissa, line 965: Sorry for the back-and-forth here.
Can we use `config.lb_config_case()` to be the switch differentiator here?
It couples the correct `oneof` that determines the correct type.
I would leave the TODO, explaining that the lb type reflects the right config value.

adisuissa (commented):
Thanks!

source/common/upstream/upstream_impl.h
DiazAlan: I'm not sure if this would work. Unless I misunderstood, I thought you couldn't copy a unique ptr?

source/common/upstream/upstream_impl.cc
DiazAlan: Might be a silly question but what cases are in lb_config_case? Do we have an example of this being used already?

source/common/upstream/upstream_impl.h
adisuissa: I see this pattern:
```
if (func() != nullptr) {
  return *func();
}
```
and I suggested replacing it with:
```
if (auto var = func(); var != nullptr) {
  return *var;
}
```

Let's delay this discussion until we see if `get_if` should be used instead.

source/common/upstream/upstream_impl.cc
DiazAlan: Done. Sorry for delay, was sick the last week

source/common/upstream/upstream_impl.cc
```diff
@@ -960,6 +960,33 @@ createOptions(const envoy::config::cluster::v3::Cluster& config,
...
+  case envoy::config::cluster::v3::Cluster::kMaglevLbConfig:
+    lb_policy_ = std::make_unique<envoy::config::cluster::v3::Cluster::MaglevLbConfig>(
+        config.maglev_lb_config());
+    break;
+  case envoy::config::cluster::v3::Cluster::kOriginalDstLbConfig:
+    lb_policy_ = std::make_unique<envoy::config::cluster::v3::Cluster::OriginalDstLbConfig>(
+        config.original_dst_lb_config());
+    break;
+  case envoy::config::cluster::v3::Cluster::LB_CONFIG_NOT_SET:
+    break;
```
adisuissa, line 986: Add `PANIC_DUE_TO_CORRUPT_ENUM;` so one can catch this cases if things go wrong.

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,57 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
...
+ * Class for LBPolicies
+ * Uses a absl::variant to store pointers for the LBPolicy
+ */
+class LBPolicyConfig {
+public:
+  LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config);
+
+  template <typename T> OptRef<const T> getConfig() const {
+    if (const auto lbPtr = absl::get_if<std::unique_ptr<const T>>(&lb_policy_); lbPtr) {
+      if (*lbPtr != nullptr) {
```
adisuissa, line 115: This condition can also be moved to the `if` statement above.
It will make the code more readable as there will be a single if-else.

adisuissa (commented):
Nice work, thanks!
Left a couple of minor comments.

source/common/upstream/upstream_impl.cc
DiazAlan: Done

source/common/upstream/upstream_impl.h
DiazAlan: Done

source/common/upstream/upstream_impl.cc
adisuissa: please move it before the `break;`

source/common/upstream/upstream_impl.h
```diff
@@ -101,6 +102,53 @@ class UpstreamLocalAddressSelectorImpl : public UpstreamLocalAddressSelector {
...
 
+/**
+ * Class for LBPolicies
+ * Uses a absl::variant to store pointers for the LBPolicy
+ */
+class LBPolicyConfig {
+public:
+  LBPolicyConfig(const envoy::config::cluster::v3::Cluster& config);
+
+  template <typename T> OptRef<const T> getConfig() const {
```
adisuissa, line 113: nit: move to private

adisuissa (commented):
Thanks, left a couple of minor comments. Otherwise LGTM!

source/common/upstream/upstream_impl.h
DiazAlan: Done

alyssawilk:
@zuercher can you take a look today?  This was assigned to Lizan who warned he'd be laggy

zuercher (approved):
Thanks!

PR #16598, opened 2021-05-20
proxy protocol: Add http support.

wez470:
Signed-off-by: Weston Carlson <wez470@gmail.com>

Commit Message: Add http proxy proto support.
Additional Description: The main addition here is setting the proxy protocol filter state information so that the upstream proxy proto transport socket can use that info to properly generate the header.
Risk Level: medium?
Testing: functional, integration
Docs Changes: Added details to IP Transparency page for using proxy proto with HTTP
Release Notes: Added note for http proxy proto support
Fixes https://github.com/envoyproxy/envoy/issues/15463

PR #25572, opened 2023-02-15
[http] Make proxy protocol filter state optional

paul-r-gall:
Commit Message: [http] Make proxy protocol filter state optional
Additional Description: Current behavior is to add ProxyProtocolFilterState to Connection lifetime Filter State from within the HCM. This behavior will be wrong in the case of multi-layer proxy setups as described [here](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades#tunneling-tcp-over-http)
Risk Level: Low
Testing: Unit Tests


api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto
```diff
@@ -815,6 +815,9 @@ message HttpConnectionManager {
   // Append the `x-forwarded-port` header with the port value client used to connect to Envoy. It
   // will be ignored if the `x-forwarded-port` header has been set by any trusted proxy in front of Envoy.
   bool append_x_forwarded_port = 51;
+
+  // Whether the HCM will add ProxyProtocolFilterState to the Connection lifetime filter state. Defaults to `true`.
```
kyessenov, line 819: I don't recall the history here, but I think this behavior was added to support ProxyProto on the upstream without ProxyProto on the downstream? I think it'd help to clarify the purpose of HCM inserting the filter state.

paul-r-gall: This was added in https://github.com/envoyproxy/envoy/pull/16598. The purpose seems to be so that upstreams that have proxy protocol enabled will have the correct state.

adisuissa:
Can you please check if the CI failure is relevant?

phlax:
> ... if the CI failure is relevant?

i think it wasnt - that issue has now been reverted in `main` so i think the fix here also needs to be reverted

/wait

alyssawilk (approved):

alyssawilk (approved):
LGTM moduldo lack of release note

alyssawilk:
cc @ggreenway for !google pass

api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto
ggreenway: Please add some documentation here about cases where it makes sense to set this to false.

ggreenway (changes requested):
/wait

ggreenway (approved):

PR #25573, opened 2023-02-15
router redirect: use port from the request if redirect_port not specified

### LINKED ISSUE #17318: host_redirect cleans up the request port when port_redirect is not specified
*Description*:
 When host_redirect is specified but port_redirect is not, port is not preserved in the response. Envoy cleans up the port in the response(https://github.com/envoyproxy/envoy/blob/main/source/common/router/config_impl.cc#L791). 
 Envoy should preserve the port in the request if port_redirect is not specified or it should be clarified in the doc(https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-redirectaction) that the port will be cleaned up if port_redirect is not specified.

*Repro steps*:
* Configure a route to do host_redirect in a listener that listens to a port other than 80 or 443
* Send a request to the route
```
curl -v --resolve redirect-test:10080:127.0.0.1 http://redirect-test:10080
* Added redirect-test:10080:127.0.0.1 to DNS cache
* Rebuilt URL to: http://redirect-test:10080/
* Hostname redirect-test was found in DNS cache
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to redirect-test (127.0.0.1) port 10080 (#0)
> GET / HTTP/1.1
> Host: redirect-test:10080
> User-Agent: curl/7.52.1
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< location: http://redirect-test.mydomain.com/
< date: Tue, 13 Jul 2021 17:33:32 GMT
< server: envoy
< content-length: 0
< 
* Curl_http_done: called premature == 0
* Connection #0 to host redirect-test left intact
```

*Config*:
```
static_resources:
  listeners:
  - address:
      socket_address:
        address: 0.0.0.0
        port_value: 10080
    filter_chains:
    - filters:
      - name: envoy.http_connection_manager
        config:
          codec_type: auto
          stat_prefix: ingress_http
          route_config:
            name: local_route
            virtual_hosts:
            - name: backend
              domains:
              - "*"
              routes:
              - match:
                  prefix: "/"
                redirect:
                  host_redirect: "redirect-test.mydomain.com"
          http_filters:
          - name: envoy.router
            config: {}
```

cpakulski:
Commit Message:
use port from the request if redirect_port not specified
Additional Description:
If redirect_port config item was empty, the redirect URL did not contain any port. Now the same port is used as in the original request.
Risk Level: Low
Testing: added unit tests and manual testing.
Docs Changes: yes
Release Notes: yes
Fixes #17318


source/common/router/config_impl.cc
```diff
@@ -1117,6 +1109,17 @@ std::string RouteEntryImplBase::newPath(const Http::RequestHeaderMap& headers) c
...
   } else {
     final_port = "";
+    const absl::string_view current_path = headers.getHostValue();
+    const auto port_start = Http::HeaderUtility::getPortStart(current_path);
+    if (port_start != absl::string_view::npos) {
+      const auto deduced_port = current_path.substr(port_start);
+      // use deduced port only when scheme did not change to avoid using port 80 when scheme
+      // changed from http to https and to avoid using port 443 when scheme changed from
+      // https to http.
+      if (!newSchemePortMismatch(headers.getForwardedProtoValue(), deduced_port, final_scheme)) {
```
htuch, line 1119: Is the negative case covered in tests below?

source/common/router/config_impl.cc
```diff
@@ -1117,6 +1109,17 @@ std::string RouteEntryImplBase::newPath(const Http::RequestHeaderMap& headers) c
     final_port = redirect_config_->port_redirect_.c_str();
   } else {
     final_port = "";
+    const absl::string_view current_path = headers.getHostValue();
```
htuch, line 1112: Out of curiosity, this is a direct response but not necessarily a redirect when you get here I think, what's to stop this applying even if it's not a redirect?

htuch (commented):
LGTM modulo comments, @yanavlasov to provide another set of eyes.

api/envoy/config/route/v3/route_components.proto
```diff
@@ -1659,6 +1659,7 @@ message RedirectAction {
       [(validate.rules).string = {well_known_regex: HTTP_HEADER_VALUE strict: false}];
 
   // The port value of the URL will be swapped with this value.
+  // If not specified, the request's port will be used.
```
howardjohn, line 1662: The "request port" is ambiguous as there can be multiple ports associated with a given request. Is it from `:authority`? from the actual L4 port?

changelogs/current.yaml
```diff
@@ -20,6 +20,9 @@ minor_behavior_changes:
 - area: http2
   change: |
     Request authorities are now validated with a library function from QUICHE rather than nghttp2. This behavior change can be reverted by setting ``envoy.reloadable_features.http2_validate_authority_with_quiche`` to ``false``.
+- area: router redirect
+  change: |
+    when ``port_redirect`` is not specified, the request's port is used.
```
howardjohn, line 25: Is this a breaking change? Our current API, which exposes this option, has the following enum option:

`FROM_PROTOCOL_DEFAULT: automatically set to 80 for HTTP and 443 for HTTPS.`

AFAIK this was the previous behavior in envoy with port unset. With this new change, it will use another port

howardjohn (changes requested):
I am worried this is a breaking change?

source/common/router/config_impl.cc
cpakulski: It is covered because without this check previously existing tests failed.

api/envoy/config/route/v3/route_components.proto
cpakulski: It is from headers, not L4.

changelogs/current.yaml
cpakulski: > AFAIK this was the previous behavior in envoy with port unset. With this new change, it will use another port

If port_redirect is set explicitly, there is no change in behavior. Does your API requires this param or can it be left unspecified?

howardjohn: Sorry to clarify, when a user sets that in our API we do *not* set port_redirect and explicitly expect the "automatically set to 80 for HTTP and 443 for HTTPS.". This behavior is changed by this PR, which I think represents a breaking API change for envoy.

cpakulski: Thanks for explaining. To be precise, Envoy did not add port 80 or 443, but simply stripped the incoming port. The result was that port was auto chosen based on the protocol: http or https. 
Solution to that problem would be to set `port_redirect` to zero for reusing the port from the request. The logic would be:
- port_redirect not specified - as before (strip port)
- port_redirect = 0 (reuse from request)
- port_redirect = value (use the value specified)

howardjohn: With a `uint32` I don't think you can distinguish unset and 0?

cpakulski: yeah.. you are right :-(

htuch: API LGTM revoked until we resolve this, thanks for noticing @howardjohn

arkodg: @howardjohn if Istio's API says
```
FROM_PROTOCOL_DEFAULT: automatically set to 80 for HTTP and 443 for HTTPS
``` 
It could now explicitly set these published port values in
https://github.com/istio/istio/blob/96f6a4f16d5a5aa4905ff0c0f94d9dd607b5f5f3/pilot/pkg/networking/core/v1alpha3/route/route.go#L603
rather than implictly assuming envoy will

from the original API PR https://github.com/istio/api/pull/2088 it looks like this is what istio would have wanted envoy to do

howardjohn: It really doesn't matter what Istio does or docs say, this is a breaking change in Envoy which will extend beyond Istio.

cpakulski:
I am planning to add a new config item with 3 options:
- STRIP
- AUTO (take port from request)
- STATIC

It will be backwards compatible, so if not specified, it will behave as before (STRIP). If `port_redirect` is specified, it will take precedence over the new config item. `port_redirect` should be deprecated after (with deprecations guidelines: 12 months of support).

cpakulski:
Closing for now. Will open a new PR with new API.

PR #17521, opened 2021-07-28
[WIP]: count sent and received bytes for http stream

chaoqin-li1123:
Commit Message: Count the number of bytes sent to and received from upstream server per request and store them in stream_info for better observability.
Additional Description:NA
Risk Level:NA
Testing:will add testing for the new metric.
Docs Changes:add doc for the new metric.
Release Notes:NA
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #25575, opened 2023-02-15
accesslog: add fields to data.accesslog.v3.AccessLogCommon

lexh:
Commit Message: accesslog: add fields to data.accesslog.v3.AccessLogCommon
Additional Description: The substitution formatter offers all of these fields, but they were not present in the ALS proto. The patch plumbs in the the same values read in `substitution_formatter.cc` ([e.g.](https://github.com/envoyproxy/envoy/blob/f38a9e13e67b0e2506a0ddb20239eaea86e9b554/source/common/formatter/substitution_formatter.cc#L1056-L1062)) into new fields in the access log proto. The doc strings are copy/pasted from the existing docs ([e.g.](https://github.com/envoyproxy/envoy/blob/main/docs/root/configuration/observability/access_log/usage.rst?plain=1#L313-L318)).
Risk Level: Low
Testing: The patch contains minimal test changes, just what was needed to get existing integration tests to pass.
Docs Changes: Yes, API docs in proto.
Release Notes: Yes.
API Considerations: Added new fields to ALS proto.


api/envoy/data/accesslog/v3/accesslog.proto
```diff
@@ -215,6 +215,38 @@ message AccessLogCommon {
...
+  // For TCP: Total number of bytes received from the downstream by the tcp proxy.
+  uint64 downstream_wire_bytes_received = 29;
+
+  // For HTTP: Number of header bytes sent to the downstream by the http stream.
+  // For TCP: Not implemented (0).
+  uint64 downstream_header_bytes_sent = 30;
+
+  // For HTTP: Number of header bytes received from the downstream by the http stream.
+  // For TCP: Not implemented (0).
+  uint64 downstream_header_bytes_received = 31;
```
wbpcode, line 233: Is there a good position for these fields? Seems that these fields are only used by HTTP protocol and IMO they shouldn't added to `AccessLogCommon`? Or the `header` could also be header of other protocol (e.g. thrift)?

htuch: Yeah, good catch, I think these belong in the respective parent message for HTTP.

lexh: Thanks for taking a look!

IIUC, you're saying that the header specific fields should be moved inside of the HTTP request/response properties fields. If I understood correctly, does the following sketch of where the new fields should be instead look correct?

```
AccessLogCommon
	- downstream_wire_bytes_sent
	- downstream_wire_bytes_received
	- upstream_wire_bytes_sent
	- upstream_wire_bytes_received

HTTPAccessLogEntry.HTTPRequestProperties
	- downstream_header_bytes_received
	- upstream_header_bytes_sent

HTTPAccessLogEntry.HTTPResponseProperties
	- upstream_header_bytes_received
	- downstream_header_bytes_sent
```

source/extensions/access_loggers/grpc/grpc_access_log_utils.cc
```diff
@@ -325,6 +325,19 @@ void Utility::extractCommonAccessLogProperties(
   if (auto provider = stream_info.getStreamIdProvider(); provider.has_value()) {
     common_access_log.set_stream_id(std::string(provider->toStringView().value_or("")));
   }
+
+  if (stream_info.getDownstreamBytesMeter() != nullptr) {
+    common_access_log.set_downstream_wire_bytes_sent(
+        stream_info.getDownstreamBytesMeter()->wireBytesSent());
+    common_access_log.set_downstream_wire_bytes_received(
+        stream_info.getDownstreamBytesMeter()->wireBytesReceived());
+  }
```
wbpcode, line 334:
```
if(const auto& bytes_meter = stream_info.getDownstreamBytesMeter(); bytes_meter != nullptr) {
  common_access_log.set_downstream_wire_bytes_sent(bytes_meter->wireBytesSent())
  common_access_log.set_downstream_wire_bytes_received(bytes_meter->wireBytesReceived());
}
```

source/extensions/access_loggers/grpc/grpc_access_log_utils.cc
```diff
@@ -325,6 +325,19 @@ void Utility::extractCommonAccessLogProperties(
...
+  if (stream_info.getUpstreamBytesMeter() != nullptr) {
+    common_access_log.set_upstream_wire_bytes_sent(
+        stream_info.getUpstreamBytesMeter()->wireBytesSent());
+    common_access_log.set_upstream_wire_bytes_received(
+        stream_info.getUpstreamBytesMeter()->wireBytesReceived());
+  }
```
wbpcode, line 340: ditto.

source/extensions/access_loggers/grpc/http_grpc_access_log_impl.cc
```diff
@@ -135,6 +143,14 @@ void HttpGrpcAccessLog::emitLog(const Http::RequestHeaderMap& request_headers,
...
   response_properties->set_response_headers_bytes(response_headers.byteSize());
   response_properties->set_response_body_bytes(stream_info.bytesSent());
+  if (stream_info.getDownstreamBytesMeter() != nullptr) {
+    response_properties->set_downstream_header_bytes_sent(
+        stream_info.getDownstreamBytesMeter()->headerBytesSent());
+  }
+  if (stream_info.getUpstreamBytesMeter() != nullptr) {
+    response_properties->set_upstream_header_bytes_received(
+        stream_info.getUpstreamBytesMeter()->headerBytesReceived());
+  }
```
wbpcode, line 153: could we merge the logic of request properties setting and response properties setting to avoid unnecessary nullptr check?

wbpcode (approved):

wbpcode (commented):
Thanks for your contribution. And some comments are added.

wbpcode:
friendly ping @htuch for API review.

api/envoy/data/accesslog/v3/accesslog.proto
```diff
@@ -220,6 +220,22 @@ message AccessLogCommon {
   // failure reason from the transport socket. The format of this field depends on the configured downstream
   // transport socket. Common TLS failures are in :ref:`TLS trouble shooting <arch_overview_ssl_trouble_shooting>`.
   string downstream_transport_failure_reason = 28;
+
+  // For HTTP: Total number of bytes sent to the downstream by the http stream.
+  // For TCP: Total number of bytes sent to the downstream by the tcp proxy.
+  uint64 downstream_wire_bytes_sent = 29;
+
+  // For HTTP: Total number of bytes received from the downstream by the http stream. Envoy over counts sizes of received HTTP/1.1 pipelined requests by adding up bytes of requests in the pipeline to the one currently being processed.
```
htuch, line 228: This is a bit weird, what is the history here? Is there a GH issue to reference?

lexh: I agree. This was copied from the [existing docs](https://github.com/envoyproxy/envoy/blame/main/docs/root/configuration/observability/access_log/usage.rst#L323-L328).

There is a bit [more context in the originating PR](https://github.com/envoyproxy/envoy/pull/17521#discussion_r729819825).

htuch: OK. Let's keep for now then, I would suggest a reference to the PR given how weird it is. @yanavlasov should there be an issue to track this?

htuch (commented):
/lgtm api

wbpcode (approved):
LGTM. Thanks.

PR #25579, opened 2023-02-15
Add a missing build dep for mobile xds_integration_test

caschoener:
I think this test works without the dependency because it's included as a sub-dependency. But some build flags will cause the dependency to be missing because it's not directly included

Anyways I tested and this makes things work inside Google without breaking anything.


jpsim (approved):
No objections from me

abeyad (approved):

PR #25582, opened 2023-02-15
mobile: adding a fake upstream for TCP tests

alyssawilk:
As an alternate to https://github.com/envoyproxy/envoy/pull/25552 allow creating a TCP test upstream which will respond to requests.

Risk Level: low
Testing: yes
Docs Changes: n/a
Release Notes: n/a


adisuissa:
Assigning Ryan for now, but feel free to reassign to another mobile maintainer.
/assign @RyanTheOptimist

mobile/test/common/integration/quic_test_server.h
```diff
@@ -24,24 +25,29 @@ class QuicTestServer {
   Thread::SkipAsserts skip_asserts_;
   ProcessWide process_wide;
   Thread::MutexBasicLockable lock;
+  ;
```
RyanTheOptimist, line 28: nit: remove this?

RyanTheOptimist (approved):
Wow, nice! One tiny nit.

mobile/test/common/integration/quic_test_server.h
```diff
@@ -10,8 +10,9 @@
 #include "test/mocks/server/transport_socket_factory_context.h"
 #include "test/config/utility.h"
 
+// TODO(alyssawilk) move this to a non-quic file name.
```
jpsim, line 13: Can we do this before merging please?

jpsim (approved):
+1 for this change as-is, it's useful to have real networking tests for Kotlin/Android.

However as discussed in Slack, I don't think this can serve as an alternative to https://github.com/envoyproxy/envoy/pull/25552 because then external consumers will need to pull in an additional dependency from Envoy Mobile in addition to the main library.

PR #25586, opened 2023-02-16
admin: Add admin format that sorts stats by activity rate while the page is open

jmarantz:
Commit Message: Adds a new admin stats format: "active", which opens a JavaScript assisted UI to show
stats ordered by frequency of change, over the duration the page is open.

This entailed adding a rudimentary JS testing flow. I waned to do this without taking new dependencies, though if the JS functionality grows we may wish to go with a fully featured testing framework and class library.

Additional Description:
Screenshot of new interface (it really needs a movie):

Capture

Risk Level: as long as admin's stats?format=active is open it will continually pull stats from the server. For large configurations this may impact main-thread responsiveness and xDS.
Testing: the unit tests for admin, plus the new semi-automatic web test.
Docs Changes: included
Release Notes: included


source/server/admin/dynamic_stats.js
```diff
@@ -0,0 +1,88 @@
...
+//   * detect when user is not looking at page and stop or slow down pinging the server
+//   * hieararchical display
+//   * json flavor to send hierarchical names to save serialization/deserialization costs
+
+let current_stats = new Map;
+
+const param_id_prefix = 'param-1-stats-';
+
+function loadStats() {
+  val = name => name + '=' + document.getElementById(param_id_prefix + name).value;
```
kyessenov, line 16: nit: maybe use some other name, `val` means `const` in some languages

source/server/admin/dynamic_stats.js
```diff
@@ -0,0 +1,88 @@
...
+    method: 'GET',
+    headers: {},
+  })
+      .then(response => response.json())
+      .then(data => renderStats(data));
+}
+
+function renderStats(data) {
+  const content = document.getElementById('dynamic-stats');
+  while (content.firstChild) {
```
kyessenov, line 29: there should be a way to clear a node without iteration

source/server/admin/dynamic_stats.js
jmarantz: sure; I haven't cleaned this up yet. Actually that's a lint error and I will run this through an online linter before marking ready for review.

source/server/admin/dynamic_stats.js
jmarantz: Maybe -- but that's what I found when I looked up "how to clear all children from a dom node" :) Maybe a recent-ish addition to the dom model? Or part of jsquery?

I wanted to keep this using vanilla js.

But actually this isn't necessary anyway right now; I was thinking I might have made the output display have some HTML structure but I just wound up throwing it all in a div for now.

source/server/admin/stats_html_render.h
```diff
@@ -10,16 +10,20 @@ class StatsHtmlRender : public StatsTextRender {
   StatsHtmlRender(Http::ResponseHeaderMap& response_headers, Buffer::Instance& response,
                   const StatsParams& params);
 
+  void dynamicParams(Buffer::Instance& response);
+
   void noStats(Buffer::Instance&, absl::string_view types) override;
   void generate(Buffer::Instance& response, const std::string& name,
                 const std::string& value) override;
+  /*
```
kyessenov, line 18: stale code?

kyessenov (commented):
Looks reasonable. Maybe attach a screenshot because it would be easier to confirm the UI works?

source/server/admin/stats_html_render.h
jmarantz: actually turns out that code is needed to get gcc to build in CI, otherwise gcc gives warnings about hidden virtual functions. I uncommented and added comments.

test/server/admin/admin_test.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "*** Invoking Envoy..."
+./bazel-bin/source/exe/envoy-static \
+      -c test/server/admin/admin_test.yaml \
+      --admin-address-path "$admin_port_file" &
+
+echo "*** Waiting for the server to go live..."
+sleep 1
+ready=""
+admin_port=$(cat $admin_port_file)
+while [ $(curl "$admin_port/ready") != "LIVE" ]; do
```
phlax, line 32:
```suggestion
while [ "$(curl "$admin_port/ready")" != "LIVE" ]; do
```

test/server/admin/admin_test.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+#
+# This test is semi-automatic. It builds envoy-static with
+# ENVOY_ADMIN_BROWSER_TEST set, then runs the binary in a mode where
+# the binary picks the admin port and writes it to a file. Then
+# the script can poll the admin port for /ready. Once live, it runs
+# Firefox (could be any browser) on the test HTML page, which will
```
phlax, line 7: we might want to use selenium for this (as a follow up)

test/server/admin/admin_test.sh
```diff
@@ -0,0 +1,43 @@
...
+      source/exe:envoy-static
+
+echo "*** Invoking Envoy..."
+./bazel-bin/source/exe/envoy-static \
+      -c test/server/admin/admin_test.yaml \
+      --admin-address-path "$admin_port_file" &
+
+echo "*** Waiting for the server to go live..."
+sleep 1
+ready=""
```
phlax, line 30:
```suggestion
```

jmarantz:
Thanks @phlax -- your comments all make sense. I think actually what I want to do is:
 * leave this PR open for comments on the JS and its test logic (even though it still needs some work)
 * leave Selenium for another day; I'm a leery of taxing the Envoy dev community and CI with a potentially flaky dependency, for a feature that I'm guessing won't be worked on all that often
 * remove admin_test.sh entirely and build a separate binary with the admin test functions, and probably shell out to firefox from that binary rather than using a shell script.

test/server/admin/admin_test.sh
jmarantz: done

test/server/admin/admin_test.sh
jmarantz: done

test/server/admin/admin_test.sh
jmarantz: added comment.

jmarantz (commented):
Thanks @phlax I re-orged things a bit so we don't need to use ifdefs to compile the C++ part of the test -- we can use a post-server hook in a custom main -- that's done.

jmarantz:
@kyessenov are you able to take another pass? I think this is ready to go at this point, though I do have some follow-ups in mind.

kyessenov (commented):
I am rusty with Javascript, but it looks reasonable. I'm not sure how far we really want to go down the JS testing path in this project - at some point the UI should really be done in a separate project, but if we keep it small, I think it's an improvement.

source/server/admin/active_stats.js
```diff
@@ -0,0 +1,192 @@
...
+    currentStats.set(stat.name, statRecord);
+    sortedStats.push(statRecord);
+  }
+  prevStats = null;
+
+  // Sorts all the stats. This is inefficient; we should just pick the top N
+  // based on field "active-max-display-count" and sort those. The best
+  // algorithms for this require a heap or priority queue. JS implementations
+  // of those can be found, but that would bloat this relatively modest amount
+  // of code, and compell us to do a better job writing tests.
```
ravenblackx, line 165: Surprised the spelling demon isn't complaining that it should be `compel`.

source/server/admin/stats_handler.h
```diff
@@ -79,9 +79,10 @@ class StatsHandler : public HandlerContextBase {
    * and for rendering in /stats?format=html. We share the same UrlHandler in
    * both contexts by defining an API for it here.
    *
+   * @prams active_mode excludes 'usedonly' -- that setting is required for active mode.
```
ravenblackx, line 82: @prams?

And this description needs some work. Is it saying that if active_mode is true, 'usedonly' stats are not included? This might be a thing where someone with all the context would understand it, but as you know I've been touching stats, and I have no idea what "that setting is required for active mode" means, nor usedonly. A clearer explanation (maybe even a context link?) would be good.

source/server/admin/active_stats.js
```diff
@@ -0,0 +1,192 @@
...
+  let index = 0;
+  let text = '';
+  for (const statRecord of sortedStats) {
+    if (++index > max) {
+      break;
+    }
+    text += statRecord.name + ': ' + statRecord.value + ' (' +
+        statRecord.change_count + ')' + '\n';
+  }
+  activeStatsPreElement.textContent = text;
```
ravenblackx, line 178: This kind of interface can be frustrating, if e.g. the user wants to copy and paste some content, erasing and re-setting the content every 5 seconds makes UI selection awkward.

This can potentially be worked around by making it a block of spans instead of a single text block, then rearranging the existing spans where the new data is similar to the old data, though that can be a bit complicated (kind of like writing a whole `diff`, but since you're only working with double-digit, maybe triple-digit amounts of content you can brute force it.)

Or by including a pause/restart button that just aborts an existing setTimeout / starts a new one.

Or it's reasonable to just deem this unimportant because it's not an important user-facing interface.

source/server/admin/active_stats.js
```diff
@@ -0,0 +1,192 @@
...
+        statRecord.value = stat.value;
+        ++statRecord.change_count;
+      }
+    } else {
+      statRecord = {name: stat.name, value: stat.value, change_count: 0};
+    }
+    currentStats.set(stat.name, statRecord);
+    sortedStats.push(statRecord);
+  }
+  prevStats = null;
```
ravenblackx, line 159: I don't think setting this to null does anything meaningful. It'll go out of scope, and the garbage collector will eventually deal with it. Explicit nulling won't make the gc get to it any quicker.

source/server/admin/active_stats.js
```diff
@@ -0,0 +1,192 @@
...
+  sortedStats.sort(compareStatRecords);
+
+  const max = loadSettingOrUseDefault('active-max-display-count', 50);
+  let index = 0;
+  let text = '';
+  for (const statRecord of sortedStats) {
+    if (++index > max) {
+      break;
+    }
+    text += statRecord.name + ': ' + statRecord.value + ' (' +
```
ravenblackx, line 175: I believe modern js style would have you do this like
```
text += `${statRecord.name}: ${statRecord.value} (${statRecord.change_count})\n`
```

ravenblackx (changes requested):

source/server/admin/active_stats.js
jmarantz: Thanks! I think it shows you how much I rely on that now. I think it must not be running on .js files.

source/server/admin/active_stats.js
jmarantz: you're probably right -- v8 would likely figure it was unreferenced after that anyway.

WDYT though -- maybe I should  not rebuild the map at all, but should remember stat change-counts for the entire life of the page?

E.g. if I mistype a filter it will erase the use-count map, which might be annoying.

source/server/admin/stats_handler.h
jmarantz: reworded; ptal.

source/server/admin/active_stats.js
jmarantz: nice. done.

source/server/admin/javascript.md
```diff
@@ -0,0 +1,60 @@
+### Overview
+
+This file contains JavaScript functionality to periodically fetch JSON stats from
+the Envoy server, and display the top 50 (by default) stats in order of how often
+they've changed since the page was brought up. This can be useful to find potentially
+problematic listeners or clsuters or other high-cardinality subsystems in Envoy whose
```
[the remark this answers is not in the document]
jmarantz: done

source/server/admin/javascript.md
```diff
@@ -0,0 +1,60 @@
...
+activity can be quickly examined for potential problems. The more active a stat, the
+more likely it is to reflect behavior of note.
+
+### Serving
+
+By default, this file is converveted by the BUILD process and stored as a C++
```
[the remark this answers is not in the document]
jmarantz: done

source/server/admin/active_stats.js
jmarantz: actually I added the .js extension to the list included in spell-checks and ran it manually; it caught more typos.

ravenblackx (approved):

source/server/admin/active_stats.js
ravenblackx: I did wonder about the lifespan of change counts, but I think it probably doesn't matter too much if the data gets expired on refiltering - it'll give you a more *immediate* higher-usage indicator, which sometimes would be better, sometimes would be worse, so it's a shrug for me.

jmarantz: I went ahead and made the above change as I think it is better to retain the counts of previously displayed items that have now gone out-of-scope based on the filter. If the user hits browser 'refresh' it will reset that map and we'll start counting from zero which seems reasonably intuitive for now.

source/server/admin/javascript.md
```diff
@@ -1,18 +1,14 @@
...
+which is used to make the admin-site more useful.
 
 ### Serving
 
-By default, this file is converted by the BUILD process and stored as a C++
-absl::string_view constexpr. It is inlined into an HTML response for the stats
-page when the 'active' format is selected. generate_admin_html.sh is used to
-compile HTML and CSS constants into string_view format as well.
+JavaScript, HTML, and CSS files are converted by the BUILD process and a script,
+generate_admin_html.sh. which generates a .cc file with constexpr
```
ravenblackx, line 9: Worth making generate_admin_html.sh a link? (Or if not, backticks around it will at least make it more readable.)

Also backticks around `constexpr absl::string_view`.

source/server/admin/javascript.md
```diff
@@ -38,22 +35,23 @@ there is significantly more JavaScript needed then migrating to that can be
...
-endpoint under test. This is needed for deep inspection of the document model
-during tests. For more details on testing mechanics, see
-test/integration/admin_web_test.sh. Note that the test is semi-automatic using a
-browser, and is not run in blaze tests or CI.
-
+which is similar to envoy-static, but uses a post-server hook to add a /test
+endpoint to the admin port, enabling the test files to run in the same origin as
+the endpoint under test. This is needed for deep inspection of the document
+model during tests, which is possible only for same-origin iframes. For more
+details on testing mechanics, see test/integration/admin_web_test.sh. Note that
```
ravenblackx, line 42: Here too, script-link or at least backticks?

ravenblackx (approved):

ravenblackx (dismissed):

source/server/admin/stats_params.cc
```diff
@@ -64,9 +64,16 @@ Http::Code StatsParams::parse(absl::string_view url, Buffer::Instance& response)
...
+    } else if (format_value.value() == "active") {
+#ifdef ENVOY_ADMIN_HTML
+      format_ = StatsFormat::Active;
+#else
+      response.add("Active HTML output was disabled by building with --define=admin_html=disabled");
+      return Http::Code::BadRequest;
 #endif
     } else {
-      response.add("usage: /stats?format=(html|json|prometheus|text)\n\n");
+      response.add("usage: /stats?format=(html|active|json|prometheus|text)\n\n");
```
ggreenway, line 76: Is there a more descriptive name than `active`? Without some more context, a user seeing this won't have a good idea of what it means. Maybe `activehtml` or similar, so they at least know it's intended to be used by a browser (not curl/cmdline)?

test/integration/admin_active_stats_test.js
```diff
@@ -0,0 +1,122 @@
+// This file contains helper functions for writing JavaScript unit tests, in the
```
ggreenway, line 1: I think it makes sense to move the non-c++ test files to a subdir here also.

ggreenway (dismissed):
This looks good overall. Approach looks good. Just a few nits.

/wait

jmarantz:
Thanks Greg -- will do both of those. I also need to add some doc.

'activehtml' seems like a good name and it's easy to change now.

ravenblackx (dismissed):

test/integration/admin_active_stats_test.js
jmarantz: done

source/server/admin/stats_params.cc
jmarantz: changed to active-html.

jmarantz:
added doc & release note. Thanks everyone for the feedback on this new flavor of PR.

changelogs/current.yaml
```diff
@@ -5,6 +5,9 @@ behavior_changes:
 - area: oauth2
   change: |
     OAuth filter now URL-encodes URL in query parameters. These query parameters are decoded, leaving intact character sequences that must remain encoded in URLs. This behavioral change can be temporarily reverted by setting runtime guard ``envoy.reloadable_features.oauth_use_url_encoding`` to false.
+- area: admin
+  change: |
+    Adds a new admin stats format option 'html-active' to display a periodically list of the top most frequently changed stats.
```
ravenblackx, line 10: Missing the word 'updated' or 'refreshed' or something.

ravenblackx (approved):

ggreenway (approved):

PR #25595, opened 2023-02-16
mobile: change request compression to be per-request

jpsim:
In practice it's very likely for mobile apps to make HTTP requests to endpoints they don't control and that may not support request decompression at all, or may not support certain compression schemes like Brotli.

So this change reworks Envoy Mobile's request compression configuration to operate on a per-request basis by adding a new `RequestHeadersBuilder.enableRequestCompression(...)` function where either gzip or Brotli can be specified as the compression algorithm. Not calling this function does not compress the request at all.

The `enableGzipCompression` and `enableBrotliCompression` functions on engine builders have been removed.

The `enableRequestCompression` function is only available when building with `--define=envoy_mobile_request_compression=enabled`, which is the default.

Commit Message: mobile: change request compression to be per-request
Additional Description:
Risk Level: Low to Envoy, High to Envoy Mobile
Testing: Local unit tests & example apps. I'd like to add more tests if time permits.
Docs Changes: Yes
Release Notes: Existing entry still applies, will add link to new PR
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


mobile/library/cc/request_headers_builder.h
```diff
@@ -11,6 +11,12 @@
 namespace Envoy {
 namespace Platform {
 
+// Available algorithms to compress requests.
+enum CompressionAlgorithm {
+  gzip,
```
RyanTheOptimist, line 16: Should these follow the Envoy style guide recommendation:

Enum values using PascalCase (e.g., RoundRobin).

mobile/library/cc/request_headers_builder.cc
```diff
@@ -43,6 +43,19 @@ RequestHeadersBuilder::addUpstreamHttpProtocol(UpstreamHttpProtocol upstream_htt
...
+RequestHeadersBuilder&
+RequestHeadersBuilder::enableRequestCompression(CompressionAlgorithm algorithm) {
+  std::string value;
+  switch (algorithm) {
+  case CompressionAlgorithm::gzip:
+    value = "gzip";
+  case CompressionAlgorithm::brotli:
+    value = "brotli";
+  }
+  internalSet("x-envoy-mobile-compression", std::vector<std::string>{value});
```
RyanTheOptimist, line 55: nit: Should we consider having a named constant for this value?

RyanTheOptimist (approved):
This is super clean... modulo the crazy complexity of setting up the composite filter :)

LGTM, modulo two small nits.

mobile/library/cc/request_headers_builder.cc
jpsim: Sure, where would that live ideally? Although I don't think it'd be worth plumbing it through to be accessed from other languages personally.

RyanTheOptimist: Good question. I'm not sure. mostly I was thinking it would be nice to share here and in engine_builder.cc. But I'm happy to keep it as is.

RyanTheOptimist (approved):

PR #25597, opened 2023-02-16
Expose upstream handshake time for logging

silverstar194:
Commit Message:
Expose upstream handshake time for logging

Additional Description:
Expose upstream handshake time for logging. This change is in parallel with how downstream handshake.

Risk Level:
Low. No functional changes.

Testing:
Same as downstream handshake. Validate value is set properly.

Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


RyanTheOptimist (approved):

test/common/stream_info/stream_info_impl_test.cc
```diff
@@ -38,7 +38,7 @@ class StreamInfoImplTest : public testing::Test {
 protected:
   void assertStreamInfoSize(StreamInfoImpl stream_info) {
     ASSERT_TRUE(sizeof(stream_info) == 800 || sizeof(stream_info) == 816 ||
-                sizeof(stream_info) == 840)
+                sizeof(stream_info) == 840 || sizeof(stream_info) == 920)
         << "If adding fields to StreamInfoImpl, please check to see if you "
            "need to add them to setFromForRecreateStream or setFrom! Current size "
```
RyanTheOptimist, line 43: Did you change these?

silverstar194: Yes, there is a new field in stream_info (upstream_timing_) so there is new size to consider.

silverstar194:
Any thoughts on why presubmit validation is stuck? Retest doesn't seem to restart either..

KBaichoo:
@silverstar194 

try `git commit -s --allow-empty -m 'Kick CI' && git push`

to rekick the CI

test/common/stream_info/stream_info_impl_test.cc
RyanTheOptimist: Sorry... what I meant was that this comment here says "please check to see if you need to add them to setFromForRecreateStream or setFrom!". I'm asking if this is something you need to do for this new field?

phlax:
> Any thoughts on why presubmit validation is stuck?

i havent tracked down why this is happening - it started ~recently and happens fairly infrequently but sometimes azp is just ignoring/missing certain PRs - ill investigate further

i have manually kicked ci for this PR for now

ramaraochavali:
Where do you use this for logging? Can you please give an example of how you use this info along with downstream handshake info?

test/common/stream_info/stream_info_impl_test.cc
silverstar194: It should not be. Looking at setFromForRecreateStream and setFrom they only need downstream items

silverstar194:
> 

It will be used in a non-public filter.

test/common/stream_info/stream_info_impl_test.cc
tyxia: `setFromForRecreateStream` is for downstream items, but the `setFrom` is to copy every field in streamInfo which includes upstream items, for example [upstream_info](https://github.com/envoyproxy/envoy/blob/a7fab06b8ec2500a421251781cc61d774e3ace60/source/common/stream_info/stream_info_impl.h#L353).

But i think we don't need to add the new fields in the first place.

source/common/stream_info/utility.cc
```diff
@@ -130,6 +130,14 @@ absl::optional<std::chrono::nanoseconds> TimingUtility::downstreamHandshakeCompl
   return duration(timing.value().get().downstreamHandshakeComplete(), stream_info_);
 }
 
+absl::optional<std::chrono::nanoseconds> TimingUtility::upstreamHandshakeComplete() {
+  OptRef<const UpstreamTiming> timing = stream_info_.upstreamTiming();
```
tyxia, line 134: All the info are already available. You can just get the timing : `OptRef<const UpstreamTiming> timing = getUpstreamTiming(stream_info_);`

tyxia (commented):
Thanks for working on this! 

High level comment: `UpstreamTiming` related interface and variable are already available in `UpstreamInfo`/`UpstreamInfoImpl` that should be available to use. So I think we don't need to add those new stuff in streaminfoImpl here

envoy/stream_info/stream_info.h
```diff
@@ -626,6 +624,12 @@ class StreamInfo {
   virtual DownstreamTiming& downstreamTiming() PURE;
   virtual OptRef<const DownstreamTiming> downstreamTiming() const PURE;
 
+  /**
+   * @return the upstream timing information.
+   */
+  virtual UpstreamTiming& upstreamTiming() PURE;
```
tyxia, line 630: I am wondering why we are moving those two functions from `UpstreamInfo` to `StreamInfo` here? 

UpstreamInfo class is a good place for upstream related stuff like this. Your PR doesn't seem to require this move.

test/common/stream_info/stream_info_impl_test.cc
```diff
@@ -37,8 +37,10 @@ std::chrono::nanoseconds checkDuration(std::chrono::nanoseconds last,
 class StreamInfoImplTest : public testing::Test {
 protected:
   void assertStreamInfoSize(StreamInfoImpl stream_info) {
-    ASSERT_TRUE(sizeof(stream_info) == 824 || sizeof(stream_info) == 840 ||
-                sizeof(stream_info) == 872)
+    ASSERT_TRUE(sizeof(stream_info) == 800 || sizeof(stream_info) == 816 ||
```
tyxia, line 40: Rather than just adding checks here, I would try to see if we can keep this number of checks to 3 as before but fix them if needed.

test/common/stream_info/stream_info_impl_test.cc
silverstar194: Irrelevant now due to changes.

source/common/stream_info/utility.cc
silverstar194: Changed

envoy/stream_info/stream_info.h
silverstar194: Changed.

source/common/stream_info/utility.h
```diff
@@ -98,6 +98,7 @@ class TimingUtility {
   absl::optional<std::chrono::nanoseconds> lastDownstreamTxByteSent();
   absl::optional<std::chrono::nanoseconds> lastDownstreamRxByteReceived();
   absl::optional<std::chrono::nanoseconds> downstreamHandshakeComplete();
+  absl::optional<std::chrono::nanoseconds> upstreamHandshakeComplete();
```
tyxia, line 101: small nit: Can you move this function to the line 96? so that upstream and downstream utilities are grouped together as before?

Same for function definition.

tyxia (dismissed):
LGTM, modulo a small format nit.

Thanks! 

Also, pass it on to @RyanTheOptimist

source/common/stream_info/utility.h
silverstar194: Sure, done.

tyxia (approved):

silverstar194:
@RyanTheOptimist  @tyxia Are you able to merge? I don't have permissions

PR #25626, opened 2023-02-17
docker/ci: Add debugging info

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/docker_ci.sh
+++ b/ci/docker_ci.sh
@@ -4,6 +4,26 @@
 # CI logs.
 set -e
 
+## DEBUGGING (NB: Set these in your env to avoided unwanted changes)
+## Set this to _not_ build/push just print what would be
+# DOCKER_CI_DRYRUN=true
+#
+## Set these to tag/push images to your own repo
+# DOCKER_IMAGE_PREFIX=mydocker/repo
+# DOCKERHUB_USERNAME=me
+# DOCKERHUB_PASSWORD=mypassword
+#
+## Set these to simulate types of CI run
+# AZP_SHA1=MOCKSHA
+# AZP_BRANCH=refs/heads/main
+# AZP_BRANCH=refs/heads/release/v1.43
+# AZP_BRANCH=refs/tags/v1.77
+##
+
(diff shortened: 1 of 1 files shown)


phlax:
@lizan i think we should add this first before making too much changes in the docker ci file as its a bit complex and hard to track

one thing running this reveals straight away is that (at least with the context i was testing) the release images are pushed  (afaict needlessly) a second time - not sure if we can just remove - windows is a factor

ci/docker_ci.sh
```diff
@@ -136,7 +154,9 @@ if [[ "${AZP_BRANCH}" != "${MAIN_BRANCH}" ]] &&
   exit 0
 fi
 
-docker login -u "$DOCKERHUB_USERNAME" -p "$DOCKERHUB_PASSWORD"
+if [[ -z "$DOCKER_CI_DEBUG" || -n "$DOCKER_CI_DEBUG_LOGIN" ]]; then
```
phlax, line 157:
```suggestion
if [[ -z "$DOCKER_CI_DEBUG" ]]; then
```

this is not necessary if you want to debug with a real build to a fake repo with fake context - just dont set the debug var

ci/docker_ci.sh
```diff
@@ -85,9 +113,19 @@ push_images() {
...
+    docker push ${BUILD_TAG}"
+  if [[ -n "$DOCKER_CI_DRYRUN" ]]; then
+      echo ""
+      return
+  fi
+  echo "..."
   # docker buildx doesn't do push with default builder
   docker "${BUILD_COMMAND[@]}" --platform "${PLATFORM}" "${args[@]}" -t "${BUILD_TAG}" . --push ||
     docker push "${BUILD_TAG}"
+  echo ""
```
wbpcode, line 128: is this necessary?

phlax: it makes it easier to parse the output - ie it separates the blocks

wbpcode (approved):
LGTM. Thanks.

PR #25624, opened 2023-02-17
ci: Fix flaky test reporting

phlax:
Fix #25623 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #25596, opened 2023-02-16
bazel/tooling: Add `@envoy_repo//:project`

phlax:
This adds a JSON data target to the repo repository rule.

The data is provided by pytooling from the repo environment, changelogs, etc, and can then be used to create eg Dockerhub pages with "supported version" information.

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Fixes commit #PR or SHA]
[Optional Deprecated:]
[Optional [API Considerations](https://github.com/envoyproxy/envoy/blob/main/api/review_checklist.md):]

PR #25628, opened 2023-02-17
readme: Add maintainer flakes guide.

phlax:
Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/maintainer/FLAKES.md
+++ b/maintainer/FLAKES.md
@@ -1,0 +1,137 @@
+
+# Finding, reporting and fixing flakes in Envoy's CI
+
+Flakes cause Pull Requests, builds, and release CI to fail, wasting a large
+amount of developer time and CI cycles.
+
+When CI is sufficiently flakey, it becomes easy to ignore new ones, and hard
+to debug real fails.
+
+More often than not, the solution will be reasonably obvious to the author of
+the PR that has introduced the flake, so the quicker this can be brought to their
+attention, the more likely it is that it can be fixed easily.
+
+## Types of flake
+
+### Tests marked flaky
+
+Envoy's bazel has `--flaky_test_attempts` set for much of its CI.
+
+These flakes are not so common, but are reported to the #flakey-test channel in Slack.
+
+### Tests that just flake
+
+The more common type of flake is a test that simply does or does not fail CI
+in different test runs.
+
+### Flakes caused by external issues
+
+Another cause of flakes is upstream issues, usually Github or Azure pipelines.
+
+For example `apt` mirrors may fail, dockerhub may fail pulling images, and so on.
+
+While we are always looking for ways to mitigate this, these are generally beyond
+our control, and outside the scope of this document.
+
+## Flake hunting
+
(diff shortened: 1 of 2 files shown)


phlax:
@alyssawilk this is a first draft of the flakes guide i was talking about 

im wondering if we should add an issue template, either with the readme or in gh issues to get the important info

as discussed ive put it a new maintainer folder for now - i was also hoping to merge in (and update) the project release guides from `tools/project` (#25605 ) and had some other things in mind that could go here

maintainer/FLAKES.md
```diff
@@ -0,0 +1,137 @@
...
+
+More often than not, the solution will be reasonably obvious to the author of
+the PR that has introduced the flake, so the quicker this can be brought to their
+attention, the more likely it is that it can be fixed easily.
+
+## Types of flake
+
+### Tests marked flaky
+
+Envoy's bazel has `--flaky_test_attempts` set for much of its CI.
```
phlax, line 18: this is not currently correct - but i think will be with #25624 (fwiw)

maintainer/FLAKES.md
```diff
@@ -0,0 +1,137 @@
...
+
+These flakes are not so common, but are reported to the #flakey-test channel in Slack.
```
phlax, line 20: this is not currently correct - but i think will be with https://github.com/envoyproxy/envoy/pull/25624 (fwiw)

phlax: and the channel is #test-flaky

phlax:
there is some good deflaking info here https://github.com/envoyproxy/envoy/blob/main/test/integration/README.md that should be linked to

maintainer/FLAKES.md
phlax: still not quite working, not sure why

maintainer/README.md
```diff
@@ -0,0 +1,2 @@
+The folder contains an ad-hoc collection of guides, policies and rules of thumb
```
alyssawilk, line 1: As said on slack, I don't think we want to create this here unless we want to move things out of the google maintainers doc we've been using and I'd appreciate a for/against poll on #maintainers before doing that switch.  I've done well over half the edits to that doc and find PRs significantly (and annoyingly) more time consuming hence my preference, but if other maintainers would find this more discover-able by all means kick it off.

phlax: fwiw the `maintainer/` dir was added in #25596 - i did try to flag that there

re using PRs/markdown/public - i think there are quite a lot of benefits - without going into them all here - and im pretty adverse to using word processors - regardless tho, i think the key point is that this info should be public - visible to contributors, prospective maintainers etc

my prefernence would be to find consensus over a poll, and im just wondering whether it needs to be either/or

phlax:
marking this no stale - i will work on it further when the dust has settled from deflaking

PR #25631, opened 2023-02-17
mobile: Improve xDS APIs in EngineBuilder

abeyad:
This commit does a few things:
  - adds an API to EngineBuilder for setting the Bootstrap's node locality.
  - adds the ability to set call and channel credentials in ADS.
  - adds the ability to use xdstp for CDS via the cds_resources_locator field.
  - removes the `api_type` parameter from the ADS API, since the only supported API type is `GRPC`.
  - adds tests for all of these in envoy_config_test.cc.
  - adds an xdstp integration test in CdsIntegrationTest.
  - removes the Delta parameter in CdsIntegrationTest since mobile xDS doesn't support Delta.

### DIFF
--- a/mobile/library/cc/engine_builder.cc
+++ b/mobile/library/cc/engine_builder.cc
@@ -259,26 +259,46 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
   return *this;
 }
 
+EngineBuilder& EngineBuilder::setNodeId(std::string node_id) {
+  node_id_ = std::move(node_id);
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::setNodeLocality(std::string region, std::string zone,
+                                              std::string sub_zone) {
+  set_node_locality_ = true;
+  node_locality_region_ = std::move(region);
+  node_locality_zone_ = std::move(zone);
+  node_locality_sub_zone_ = std::move(sub_zone);
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& address,
+                                                            const int port, std::string jwt_token,
+                                                            const int jwt_token_lifetime_seconds,
+                                                            std::string ssl_root_certs) {
+#ifndef ENVOY_GOOGLE_GRPC
+  throw std::runtime_error("google_grpc must be enabled in bazel to use ADS");
+#endif
+  ads_address_ = address;
+  ads_port_ = port;
+  jwt_token_ = std::move(jwt_token);
+  jwt_token_lifetime_seconds_ = jwt_token_lifetime_seconds;
+  ssl_root_certs_ = std::move(ssl_root_certs);
+  return *this;
+}
+
 EngineBuilder& EngineBuilder::addRtdsLayer(const std::string& layer_name,
                                            const int timeout_seconds) {
   rtds_layer_name_ = layer_name;
   rtds_timeout_seconds_ = timeout_seconds;
   return *this;
 }
-EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& api_type,
-                                                            const std::string& address,
-                                                            const int port) {
-#ifndef ENVOY_GOOGLE_GRPC
-  throw std::runtime_error("google_grpc must be enabled in bazel to use ADS");
-#endif
-  ads_api_type_ = api_type;
-  ads_address_ = address;
-  ads_port_ = port;
-  return *this;
-}
 
-EngineBuilder& EngineBuilder::addCdsLayer(const int timeout_seconds) {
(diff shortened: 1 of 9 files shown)


abeyad:
cc @caschoener FYI

mobile/library/cc/engine_builder.cc
```diff
@@ -259,26 +259,46 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
...
+                                           const int timeout_seconds) {
+  rtds_layer_name_ = layer_name;
+  rtds_timeout_seconds_ = timeout_seconds;
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::addCdsLayer(std::string cds_resources_locator,
+                                          const int timeout_seconds) {
   enable_cds_ = true;
+  cds_resources_locator_ = cds_resources_locator;
```
caschoener, line 301: Seems like this one should be a std::move too?

mobile/library/cc/engine_builder.cc
```diff
@@ -259,26 +259,46 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
...
+                                                            std::string ssl_root_certs) {
 #ifndef ENVOY_GOOGLE_GRPC
   throw std::runtime_error("google_grpc must be enabled in bazel to use ADS");
 #endif
-  ads_api_type_ = api_type;
   ads_address_ = address;
   ads_port_ = port;
+  jwt_token_ = std::move(jwt_token);
+  jwt_token_lifetime_seconds_ = jwt_token_lifetime_seconds;
+  ssl_root_certs_ = std::move(ssl_root_certs);
```
caschoener, line 287: My thinking was to prefix all the ADS config variables with `ads_`, and `rtds_` for the RTDS ones.  Just to avoid naming conflicts as this file grows.

caschoener (approved):
Left 2 comments but nothing critical so I shall approve, LGTM!

mobile/library/cc/engine_builder.cc
abeyad: Done

mobile/library/cc/engine_builder.cc
abeyad: done

caschoener (dismissed):

mobile/library/cc/engine_builder.cc
```diff
@@ -259,26 +259,46 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
...
+  node_id_ = std::move(node_id);
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::setNodeLocality(std::string region, std::string zone,
+                                              std::string sub_zone) {
+  set_node_locality_ = true;
+  node_locality_region_ = std::move(region);
+  node_locality_zone_ = std::move(zone);
+  node_locality_sub_zone_ = std::move(sub_zone);
```
jpsim, line 272: If these are always set together, would it be better to use something like `std::optional<NodeLocality>` ?

jpsim (dismissed):
+1 just an optional suggestion for changing a bool + 3 fields into an optional of a struct with those 3 fields.

mobile/library/cc/engine_builder.cc
abeyad: Added NodeLocality struct

mobile/library/cc/engine_builder.h
```diff
@@ -18,8 +19,16 @@
 namespace Envoy {
 namespace Platform {
 
+constexpr int DefaultJwtTokenLifetimeSeconds = 31536000; // 1 year
 constexpr int DefaultXdsTimeout = 5;
 
+// Represents the locality information in the Bootstrap's node.
```
jpsim, line 25: Nit: Can you include a link to what "locality" means in this context?

mobile/library/cc/engine_builder.h
```diff
@@ -18,8 +19,16 @@
...
+struct NodeLocality {
```
jpsim, line 26: Nit: Consider moving to its own file.

jpsim (approved):
Two optional nits.

abeyad:
Sorry, the PR got merged based on my setting auto-merge, I will address the comments in a follow up PR now.

mobile/library/cc/engine_builder.h
abeyad: I thought about this, but preferred to keep it in engine_builder.h b/c someone would only use it in the context of setNodeLocality, which is an EngineBuilder function, so they would only have to include one file, and it is small enough of a struct.

jpsim: Yeah it's fine

PR #25638, opened 2023-02-17
mobile: Add RTDS and ADS apis for all languages

caschoener:
I added the c++ APIs previously, just filling out support for the other languages and adding appropriate builder tests. 

Added some github comments to the latest revision to explain things.

This should mostly match the layout of [this recent similar PR](https://github.com/envoyproxy/envoy/commit/95a3faf69e8eef6edde867a3055ff4a5a9cd791a#diff-c9d2a3fe6f80c563588924eb894ef5a846bab0946f82615fb5e53177e1d4df56), except without the C++ components which were already done.


mobile/library/cc/engine_builder.cc
```diff
@@ -222,43 +222,50 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
...
+                                                            bool use_ads) {
 #ifndef ENVOY_GOOGLE_GRPC
   throw std::runtime_error("google_grpc must be enabled in bazel to use ADS");
 #endif
   ads_address_ = address;
   ads_port_ = port;
   ads_jwt_token_ = std::move(jwt_token);
-  ads_jwt_token_lifetime_seconds_ = jwt_token_lifetime_seconds;
+  ads_jwt_token_lifetime_seconds_ =
+      jwt_token_lifetime_seconds == 0 ? DefaultJwtTokenLifetimeSeconds : jwt_token_lifetime_seconds;
```
caschoener, line 250: Moved the timeout defaults into the function bodies because we're passing in 0 from other languages to represent null (no easy way to pass nulls around).

mobile/library/common/jni/jni_interface.cc
```diff
@@ -1125,6 +1127,17 @@ void setString(JNIEnv* env, jstring java_string, EngineBuilder* builder,
...
+// Convert jstring to std::string
+std::string getCppString(JNIEnv* env, jstring java_string) {
+  if (!java_string) {
+    return "";
+  }
+  const char* native_java_string = env->GetStringUTFChars(java_string, nullptr);
+  std::string cpp_string = std::string(native_java_string);
+  env->ReleaseStringUTFChars(java_string, native_java_string);
+  return cpp_string;
+}
```
caschoener, line 1139: We currently only have a helper function `setString` which works for builder APIs which set exactly one string.  This approach works for any APIs that need cpp strings

mobile/library/objective-c/EnvoyConfiguration.mm
```diff
@@ -244,4 +285,18 @@ - (instancetype)initWithAdminInterfaceEnabled:(BOOL)adminInterfaceEnabled
...
+- (NSString *)generateYamlString {
+  try {
+    Envoy::Platform::EngineBuilder builder = [self applyToCXXBuilder];
+    std::unique_ptr<envoy::config::bootstrap::v3::Bootstrap> bootstrap =
+        builder.generateBootstrap();
+    std::string bootstrap_str = Envoy::MessageUtil::getYamlStringFromMessage(*bootstrap);
+    return [NSString stringWithUTF8String:bootstrap_str.c_str()];
+  } catch (const std::exception &e) {
+    NSLog(@"[Envoy] error generating yaml: %@", @(e.what()));
+    return nullptr;
```
caschoener, line 298: Added a helper function to the EnvoyConfiguration object to get a yaml string out for tests to parse.  This kinda matches how things work in the jni_interface. Let me know if there's some way to make it test-only that I should do.

mobile/test/swift/EngineBuilderTests.swift
```diff
@@ -408,6 +392,42 @@ final class EngineBuilderTests: XCTestCase {
     self.waitForExpectations(timeout: 0.01)
   }
 
+  func testAddingRtdsAndAdsConfigurationWhenRunningEnvoy() {
+    let expectation = self.expectation(description: "Run called with expected data")
+    MockEnvoyEngine.onRunWithConfig = { config, _ in
+      XCTAssertTrue(config.generateYamlString().contains("rtds_layer_name"))
```
caschoener, line 398: These are now the only tests for swift or objective C which actually assert that the generated config is right - the current tests just assert that the EnvoyConfiguration object has stored the input params correctly which is trivial.  I'd be inclined to do a follow-up PR to move all tests here over to the new style.

mobile/library/cc/engine_builder.cc
```diff
@@ -222,43 +222,50 @@ EngineBuilder& EngineBuilder::enforceTrustChainVerification(bool trust_chain_ver
   return *this;
 }
 
-EngineBuilder& EngineBuilder::setNodeId(std::string node_id) {
+EngineBuilder& EngineBuilder::setNodeId(std::string node_id, bool use_node_id) {
```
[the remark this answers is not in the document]
caschoener: After talking with Ali, I will remove these from the API functions.  Instead there's an if statement in java and objective-c to decide whether to call the builder API or not.

mobile/library/cc/engine_builder.cc
```diff
@@ -232,8 +232,8 @@ EngineBuilder& EngineBuilder::setNodeLocality(const NodeLocality& node_locality)
   return *this;
 }
 
-EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& address,
-                                                            const int port, std::string jwt_token,
+EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(std::string address, const int port,
```
abeyad, line 235: since `address` is no longer a const ref, you'll want to do std::move when assigning it to `ads_addres_`

mobile/library/cc/engine_builder.cc
```diff
@@ -242,23 +242,23 @@ EngineBuilder& EngineBuilder::setAggregatedDiscoveryService(const std::string& a
...
+  ads_jwt_token_lifetime_seconds_ =
+      jwt_token_lifetime_seconds == 0 ? DefaultJwtTokenLifetimeSeconds : jwt_token_lifetime_seconds;
   ads_ssl_root_certs_ = std::move(ssl_root_certs);
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addRtdsLayer(const std::string& layer_name,
-                                           const int timeout_seconds) {
+EngineBuilder& EngineBuilder::addRtdsLayer(std::string layer_name, const int timeout_seconds) {
   rtds_layer_name_ = layer_name;
```
abeyad, line 252: ditto: std::move here

mobile/library/cc/engine_builder.h
```diff
@@ -164,14 +160,14 @@ class EngineBuilder {
   int rtds_timeout_seconds_;
   std::string node_id_;
   absl::optional<NodeLocality> node_locality_ = absl::nullopt;
-  std::string ads_address_ = "";
+  std::string ads_address_ = ""; // make absl:optional?
```
abeyad, line 163: remove the comment

mobile/library/cc/engine_builder.h
```diff
@@ -164,14 +160,14 @@ class EngineBuilder {
...
   int ads_port_;
   std::string ads_jwt_token_;
   int ads_jwt_token_lifetime_seconds_;
   std::string ads_ssl_root_certs_;
-  bool enable_cds_ = false;
   std::string cds_resources_locator_;
   int cds_timeout_seconds_;
+  bool enable_cds_ = false;
```
abeyad, line 170: why was this moved 2 lines down?  best to keep it where it was so it doesn't show up in the diffs since we aren't changing anything

mobile/library/common/jni/jni_interface.cc
```diff
@@ -1125,6 +1127,17 @@ void setString(JNIEnv* env, jstring java_string, EngineBuilder* builder,
   }
 }
 
+// Convert jstring to std::string
+std::string getCppString(JNIEnv* env, jstring java_string) {
+  if (!java_string) {
+    return "";
+  }
+  const char* native_java_string = env->GetStringUTFChars(java_string, nullptr);
+  std::string cpp_string = std::string(native_java_string);
```
abeyad, line 1136: std::string cpp_string(native_java_string)

mobile/library/objective-c/EnvoyConfiguration.mm
abeyad: I assume this function and the Swift one are meant to be analogs for https://github.com/envoyproxy/envoy/blob/3eab4eab695bd7069e7a9a6b2ed83edb3804010f/mobile/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java#L242 ?

If that's the case, let's put TODOs to move them to test only libraries.  Or we could just move these functions here and the Swift one into separate test directory files and use them in the tests only.

mobile/test/swift/EngineBuilderTests.swift
```diff
@@ -5,22 +5,6 @@ import XCTest
...
-  listeners:
-#{custom_listeners}
-  filters:
-#{custom_filters}
-  runtime:
-#{custom_runtime}
-  routes:
-#{custom_routes}
-"""
-
```
abeyad, line 23: thanks for cleaning this up!

mobile/test/swift/EngineBuilderTests.swift
abeyad: +1, this is a good test and we could use more of them!

mobile/library/cc/engine_builder.h
```diff
@@ -77,20 +76,17 @@ class EngineBuilder {
...
   // Adds an ADS layer. Note that only the state-of-the-world gRPC protocol is supported, not Delta
   // gRPC.
-  EngineBuilder&
-  setAggregatedDiscoveryService(const std::string& address, const int port,
-                                std::string jwt_token = "",
-                                int jwt_token_lifetime_seconds = DefaultJwtTokenLifetimeSeconds,
-                                std::string ssl_root_certs = "");
+  EngineBuilder& setAggregatedDiscoveryService(std::string address, const int port,
+                                               std::string jwt_token = "",
+                                               int jwt_token_lifetime_seconds = 0,
```
abeyad, line 81: is the reason that this (and the timeouts below) were changed from Default to 0 so that the other languages can just pass in 0 instead of having to know the defaults?

abeyad (commented):
Thanks!  Left a few comments.

mobile/test/swift/EngineBuilderTests.swift
```diff
@@ -408,6 +392,43 @@ final class EngineBuilderTests: XCTestCase {
...
+      XCTAssertFalse(config.generateYamlString().contains("ads_config:"))
+      expectation.fulfill()
+    }
+
+    _ = EngineBuilder()
+      .addEngineType(MockEnvoyEngine.self)
+      .build()
+    self.waitForExpectations(timeout: 0.01)
+  }
+
```
abeyad, line 431: Also, tests for node.id and node.locality would be good too

caschoener:
Addressed your comments.  Also yes, I think I'll go back and migrate the rest of the swift tests to the new format as a followup PR

mobile/docs/root/api/starting_envoy.rst
```diff
@@ -542,18 +545,21 @@ Adds an RTDS layer to envoy configuration. Requires that ADS be configured via `
 ``setAggregatedDiscoveryService``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Adds ADS to envoy configuration, for instance to be used with RTDS and CDS layers.
+Adds ADS to bootstrap configuration, for instance to be used with RTDS and CDS layers.
+Optional params allow configuring jwt token and ssl. See the following link for details:
```
abeyad, line 549: nit: a JWT token and SSL

abeyad (approved):
Just left a minor nit, otherwise LGTM!  @jpsim if you can kindly review, thanks!

mobile/docs/root/api/starting_envoy.rst
```diff
@@ -527,6 +527,60 @@ Note that Envoy will fail to start up in debug mode if an unknown guard is speci
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adds an RTDS layer to the bootstrap configuration.
+Requires that ADS be configured via `setAggregatedDiscoveryService()`.
+See the following link for details:
+https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/runtime/v3/rtds.proto
+
+**Example**::
+
+  // Kotlin
```
jpsim, line 541: Nit: Should examples be added for Swift & C++ too? When the syntax is valid for more than one language we should also just do something like `// C++ / Kotlin / Swift` and have a single code sample instead of duplicating for each language.

mobile/docs/root/api/starting_envoy.rst
```diff
@@ -527,6 +527,60 @@ Note that Envoy will fail to start up in debug mode if an unknown guard is speci
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adds ADS to bootstrap configuration, for instance to be used with RTDS and CDS layers.
+Optional params allow configuring a JWT token and SSL. See the following link for details:
+https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/xds_api#config-overview-ads
+
+**Example**::
+
+  // Kotlin
+  builder.setAggregatedDiscoveryService("192.168.1.1", 0)
```
jpsim, line 555: It's unclear (to me anyway) what the parameters do here. I'm assuming that the first parameter must be an IPv4 address where the ADS is hosted? I can't really guess what the `0` signifies since there's no parameter label and I can't tell what it could be based on the type. The paragraph above mentions that there are optional params for configuring a JWT and SSL, but I don't know if the `0` is related to that or if those params are omitted from this example.

mobile/library/cc/engine_builder.cc
jpsim: Still remains to be done, right?

mobile/library/objective-c/EnvoyConfiguration.h
```diff
@@ -47,7 +47,17 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong) NSDictionary<NSString *, EnvoyStringAccessor *> *stringAccessors;
 @property (nonatomic, strong) NSDictionary<NSString *, id<EnvoyKeyValueStore>> *keyValueStores;
 @property (nonatomic, strong) NSArray<NSString *> *statsSinks;
-
+@property (nonatomic, strong) NSString *rtdsLayerName;
```
jpsim, line 50: Is this RTDS / ADS stuff always configured together? Ideally we encapsulate these fields into a more structured `RTDSLayer` type so you can build that layer separately from the engine then call `EngineBuilder.addRTDSLayer(...)`. AKA [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

mobile/library/swift/EngineBuilder.swift
```diff
@@ -546,6 +557,74 @@ open class EngineBuilder: NSObject {
     return self
   }
 
+  /// Sets the node.id field in the Bootstrap configuration.
+  ///
+  /// - parameter nodeId: The node ID.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setNodeId(_ nodeId: String) -> Self {
```
jpsim, line 566: Nit: We should follow [Swift's API naming guidelines](https://www.swift.org/documentation/api-design-guidelines/) here

```suggestion
  public func setNodeID(_ nodeID: String) -> Self {
```

> [Acronyms and initialisms](https://en.wikipedia.org/wiki/Acronym) that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions:
> ```swift
> var utf8Bytes: [UTF8.CodeUnit]
> var isRepresentableAsASCII = true
> var userSMTPServer: SecureSMTPServer
> ```

mobile/library/swift/EngineBuilder.swift
```diff
@@ -546,6 +557,74 @@ open class EngineBuilder: NSObject {
...
+  public func setNodeLocality(
+    _ region: String,
+    _ zone: String,
+    _ subZone: String
+    ) -> Self {
+    self.nodeRegion = region
+    self.nodeZone = zone
+    self.nodeSubZone = subZone
+    return self
+  }
```
jpsim, line 588: Similarly, we should follow Swift's API guidelines here:

```suggestion
  public func setNodeLocality(
    region: String,
    zone: String,
    subZone: String
  ) -> Self {
    self.nodeRegion = region
    self.nodeZone = zone
    self.nodeSubZone = subZone
    return self
  }
```

mobile/library/swift/EngineBuilder.swift
```diff
@@ -546,6 +557,74 @@ open class EngineBuilder: NSObject {
...
+
+  /// Adds an RTDS layer to the configuration.
+  ///
+  /// - parameter layerName:      The layer name.
+  /// - parameter timeoutSeconds: The timeout in seconds.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func addRtdsLayer(name layerName: String, timeoutSeconds: UInt32 = 0) -> Self {
```
jpsim, line 622:
```suggestion
  public func addRTDSLayer(name layerName: String, timeoutSeconds: UInt32 = 0) -> Self {
```

mobile/test/swift/EngineBuilderTests.swift
```diff
@@ -421,6 +405,46 @@ final class EngineBuilderTests: XCTestCase {
     self.waitForExpectations(timeout: 0.01)
   }
 
+  func testAddingRtdsAndAdsConfigurationWhenRunningEnvoy() {
+    let bootstrapDebugDescription = EngineBuilder()
+      .addEngineType(MockEnvoyEngine.self)
+      .addRtdsLayer(name: "rtds_layer_name", timeoutSeconds: 14325)
+      .setAggregatedDiscoveryService(address: "FAKE_SWIFT_ADDRESS", port: 0)
+      .bootstrapDebugDescription()
+    print(bootstrapDebugDescription)
```
jpsim, line 414:
```suggestion
```

mobile/test/swift/EngineBuilderTests.swift
```diff
@@ -421,6 +405,46 @@ final class EngineBuilderTests: XCTestCase {
...
+    XCTAssertTrue(bootstrapDebugDescription.contains("rtds_layer_name"))
+    XCTAssertTrue(bootstrapDebugDescription.contains(#"initial_fetch_timeout { seconds: 14325 }"#))
```
jpsim, line 416: Nit: no need for a raw string literal here since nothing needs to be escaped:

```suggestion
    XCTAssertTrue(bootstrapDebugDescription.contains("initial_fetch_timeout { seconds: 14325 }"))
```

jpsim (commented):
Looks good overall, I just have some documentation & API nits. Thanks for doing this!

mobile/library/objective-c/EnvoyConfiguration.h
caschoener: ADS is also used with CDS, and maybe other xds types in the future.  So it seemed simplest to split it out.

mobile/docs/root/api/starting_envoy.rst
caschoener: Tried adding more thorough documentation, had to come up with some new formatting style.  Let me know what you think!

mobile/library/swift/EngineBuilder.swift
```diff
@@ -546,6 +557,74 @@ open class EngineBuilder: NSObject {
...
+  /// - parameter zone:    The zone.
+  /// - parameter subZone: The sub-zone.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setNodeLocality(
+    region: String,
+    zone: String,
+    subZone: String
+    ) -> Self {
```
jpsim, line 583: Nit:
```suggestion
  ) -> Self {
```

mobile/library/swift/EngineBuilder.swift
```diff
@@ -546,6 +557,74 @@ open class EngineBuilder: NSObject {
...
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setAggregatedDiscoveryService(
+    address: String,
+    port: UInt32,
+    jwtToken: String = "",
+    jwtTokenLifetimeSeconds: UInt32 = 0,
+    sslRootCerts: String = ""
+    ) -> Self {
```
jpsim, line 606: Nit:
```suggestion
  ) -> Self {
```

jpsim (approved):

abeyad (approved):

Issue #25614, opened 2023-02-17
tests FileReadToEndNotReadable keeps failing on release/v1.25


noah8713:
**If you are reporting *any* crash or *any* potential security issue, *do not*
open an issue in this repo. Please report the issue via emailing
envoy-security@googlegroups.com where the issue will be triaged appropriately.**

*Title*: tests FileReadToEndNotReadable keeps failing

*Description*:
> bazel.release //test/... --test_env=ENVOY_IP_TEST_VERSIONS=v4only --flaky_test_attempts=9
FAILED: //test/common/filesystem:filesystem_impl_test (Summary)
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test.log
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test_attempts/attempt_1.log
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test_attempts/attempt_2.log
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test_attempts/attempt_3.log
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test_attempts/attempt_4.log
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test_attempts/attempt_5.log
      /build/tmp/_bazel_root/3712b2d6d3ae88430b07e4b942dcf6a8/execroot/envoy/bazel-out/k8-opt/testlogs/test/common/filesystem/filesystem_impl_test/test_attempts/attempt_6.log
[...]

PR #25641, opened 2023-02-17
FileReadToEndNotReadable keeps failing

noah8713:
Fixes:
https://github.com/envoyproxy/envoy/issues/25614

Signed-off-by: Aliasgar Ginwala <aginwala@ebay.com>

### DIFF
--- a/test/common/filesystem/filesystem_impl_test.cc
+++ b/test/common/filesystem/filesystem_impl_test.cc
@@ -108,17 +108,15 @@ TEST_F(FileSystemImplTest, FileReadToEndDoesNotExist) {
                EnvoyException);
 }
 
-#ifndef WIN32
-// In Windows this method of removing the permissions does not make read fail.
-TEST_F(FileSystemImplTest, FileReadToEndNotReadable) {
-  const std::string data = "test string\ntest";
-  const std::string file_path = TestEnvironment::writeStringToFileForTest("test_envoy", data);
+// Issue https://github.com/envoyproxy/envoy/issues/25614
+// TEST_F(FileSystemImplTest, FileReadToEndNotReadable) {
+//  const std::string data = "test string\ntest";
+//  const std::string file_path = TestEnvironment::writeStringToFileForTest("test_envoy", data);
 
-  std::filesystem::permissions(file_path, std::filesystem::perms::owner_all,
-                               std::filesystem::perm_options::remove);
-  EXPECT_THROW(file_system_.fileReadToEnd(file_path), EnvoyException);
-}
-#endif
+//  std::filesystem::permissions(file_path, std::filesystem::perms::owner_all,
+//                               std::filesystem::perm_options::remove);
+//  EXPECT_THROW(file_system_.fileReadToEnd(file_path), EnvoyException);
+// }
 
 TEST_F(FileSystemImplTest, FileReadToEndDenylisted) {
   EXPECT_THROW(file_system_.fileReadToEnd("/dev/urandom"), EnvoyException);
--- a/test/common/filesystem/filesystem_impl_test.cc
+++ b/test/common/filesystem/filesystem_impl_test.cc
@@ -108,15 +108,18 @@ TEST_F(FileSystemImplTest, FileReadToEndDoesNotExist) {
                EnvoyException);
 }
 
-// Issue https://github.com/envoyproxy/envoy/issues/25614
-// TEST_F(FileSystemImplTest, FileReadToEndNotReadable) {
-//  const std::string data = "test string\ntest";
-//  const std::string file_path = TestEnvironment::writeStringToFileForTest("test_envoy", data);
+#ifndef WIN32
+// In Windows this method of removing the permissions does not make read fail.
+// Issue https://github.com/envoyproxy/envoy/issues/25614, disabling this test
+TEST_F(FileSystemImplTest, DISABLED_FileReadToEndNotReadable) {
+  const std::string data = "test string\ntest";
+  const std::string file_path = TestEnvironment::writeStringToFileForTest("test_envoy", data);
 
-//  std::filesystem::permissions(file_path, std::filesystem::perms::owner_all,
-//                               std::filesystem::perm_options::remove);
-//  EXPECT_THROW(file_system_.fileReadToEnd(file_path), EnvoyException);
-// }
+  std::filesystem::permissions(file_path, std::filesystem::perms::owner_all,
+                               std::filesystem::perm_options::remove);
+  EXPECT_THROW(file_system_.fileReadToEnd(file_path), EnvoyException);
+}
+#endif
 
 TEST_F(FileSystemImplTest, FileReadToEndDenylisted) {
   EXPECT_THROW(file_system_.fileReadToEnd("/dev/urandom"), EnvoyException);
--- a/test/per_file_coverage.sh
+++ b/test/per_file_coverage.sh
@@ -10,6 +10,7 @@ declare -a KNOWN_LOW_COVERAGE=(
 "source/common/config:96.4"
 "source/common/crypto:88.1"
 "source/common/event:95.1" # Emulated edge events guards don't report LCOV
+"source/common/filesystem/posix:96.5" # FileReadToEndNotReadable keeps failing
 "source/common/http:96.3"
 "source/common/http/http2:95.0"
 "source/common/json:93.4"


ravenblackx:
Note, this fix just makes the test a duplicate of the test above it (with a misleading name and some unrelated actions suggesting there's some coverage that there is not).

phlax:
@ravenblackx can you advise on the best way forward

(following up on https://github.com/envoyproxy/envoy/issues/25614#issuecomment-1435903136)

if coverage is dropping then i guess that means another test is required for the uncovered lines 

also a bit confused - if it is duplicating above test then i would assume that removing the test should have no affect on coverage

phlax:
> i would assume that removing the test ...

nm, i see - currently the test gives some coverage lines, when "fixed" it becomes a dupe of above and loses that coverage

i guess the question is what is needed to properly cover those lines

ravenblackx:
Yeah, I think this is a situation where those lines can't be covered *and* have the test not break in environments where the file permissions don't behave the desirable way (can't mock it out because the test is explicitly of the non-mock filesystem code). I'm not aware of a more reliable way to cause a 'read' operation to fail, after an 'open' operation succeeded.

It's possible the test could be conditional (check that the permissions preclude reading the file after they've been changed, and if they don't, just log that the test was skipped and don't do the EXPECT), which would provide the coverage in the coverage environment where the test works, and then 'skip' in the environment where the test doesn't work.

A reasonable alternative would be to just delete the problem test, annotate that branch as not reasonable to cover, and explicitly reduce the coverage for that file.

noah8713:
> Yeah, I think this is a situation where those lines can't be covered _and_ have the test not break in environments where the file permissions don't behave the desirable way (can't mock it out because the test is explicitly of the non-mock filesystem code). I'm not aware of a more reliable way to cause a 'read' operation to fail, after an 'open' operation succeeded.
> 
> It's possible the test could be conditional (check that the permissions preclude reading the file after they've been changed, and if they don't, just log that the test was skipped and don't do the EXPECT), which would provide the coverage in the coverage environment where the test works, and then 'skip' in the environment where the test doesn't work.
> 
> A reasonable alternative would be to just delete the problem test, annotate that branch as not reasonable to cover, and explicitly reduce the coverage for that file.

@ravenblackx @phlax thanks. What is the final suggestion? Can we do incremental patch with alternatives post merging this? Please advice.

ravenblackx:
It looks like this won't merge as-is because it fails coverage, so you'd have to update the coverage file anyway if you're making this change.

My suggestion would be, if we're going for a quick fix for now and maybe iterating later:
1. Change the test case back to the version that isn't meaningless, then comment it out entirely, with an explanatory comment about how it doesn't work in some envs. (Maybe link to this PR or the issue for more context.)
2. Add "source/common/filesystem/posix:96.5" in [/test/per_file_coverage.sh](https://github.com/envoyproxy/envoy/blob/main/test/per_file_coverage.sh) (argh so close to not needing an exception!)

noah8713:
Thanks @ravenblackx : So coverage failure is due to minimal changes as per [logs](https://dev.azure.com/cncf/envoy/_build/results?buildId=128584&view=logs&j=bbe4b42d-86e6-5e9c-8a0b-fea01d818a24&t=e00c5a13-c6dc-5e9a-6104-69976170e881&l=4565) 
```
Code coverage for source/common/filesystem/posix is lower than limit of 96.6 (96.5)
##[error]Bash exited with code '1'.
Finishing: Run CI script bazel.coverage

```
So for my knowledge why do we need to touch coverage for minimal changes? 

Follow up on your suggestions if I understand correctly: 
For 1.  I an comment out in this PR `TEST_F(FileSystemImplTest, FileReadToEndNotReadable) {` and add issue link. 
For 2. do you mean for call we need to use this https://github.com/envoyproxy/envoy/blob/main/source/common/filesystem/posix/filesystem_impl.cc#L96 so that for linux it wont be showing not needing exception ? . I was thinking if we comment out we can skip 2 for now too. Just not getting the need for this. Please help clarify. Let me know

phlax:
> Just not getting the need for this. Please help clarify

my understanding re point 1 is yep, that is correct

point 2 is to add an entry for `source/common/filesystem/posix` in here https://github.com/envoyproxy/envoy/blob/main/test/per_file_coverage.sh - to lower the coverage expectation

noah8713:
> > Just not getting the need for this. Please help clarify
> 
> my understanding re point 1 is yep, that is correct
> 
> point 2 is to add an entry for `source/common/filesystem/posix` in here https://github.com/envoyproxy/envoy/blob/main/test/per_file_coverage.sh - to lower the coverage expectation

Thanks a bunch @phlax and @ravenblackx for suggestions. Addressed comments. PTAL

noah8713:
@ravenblackx @phlax as per suggestions commenting the code was more work as format checks kept failing.
```
./test/common/filesystem/filesystem_impl_test.cc:112:// #ifndef WIN32
                                                         ^
Checked 4461 file(s) and 121304 comment(s), found 1 error(s).
ERROR: spell check failed. Run 'tools/spelling/check_spelling_pedantic.py fix and/or add new words to tools/spelling/spelling_dictionary.txt'

```
Anyways adjusted as requested to comment out with original working state and issue link so someone else might not just re change it to failing state . PTAL as most of the checks passed.

test/common/filesystem/filesystem_impl_test.cc
```diff
@@ -108,17 +108,16 @@ TEST_F(FileSystemImplTest, FileReadToEndDoesNotExist) {
...
-}
-#endif
+// Issue https://github.com/envoyproxy/envoy/issues/25614
+// TEST_F(FileSystemImplTest, FileReadToEndNotReadable) {
+//  const std::string data = "test string\ntest";
+//  const std::string file_path = TestEnvironment::writeStringToFileForTest("test_envoy", data);
+
+//  std::filesystem::permissions(file_path, std::filesystem::perms::owner_all,
+//                               std::filesystem::perm_options::remove);
+//  EXPECT_THROW(file_system_.fileReadToEnd(TestEnvironment::temporaryPath("envoy_this_not_exist")),
```
ravenblackx, line 118: Please make this line `// EXPECT_THROW(file_system_.fileReadToEnd(file_path), EnvoyException);` - otherwise the commented out test has negative value as documentation, because it's actively misleading.

test/common/filesystem/filesystem_impl_test.cc
```diff
@@ -108,17 +108,16 @@ TEST_F(FileSystemImplTest, FileReadToEndDoesNotExist) {
...
-TEST_F(FileSystemImplTest, FileReadToEndNotReadable) {
-  const std::string data = "test string\ntest";
-  const std::string file_path = TestEnvironment::writeStringToFileForTest("test_envoy", data);
-
-  std::filesystem::permissions(file_path, std::filesystem::perms::owner_all,
-                               std::filesystem::perm_options::remove);
-  EXPECT_THROW(file_system_.fileReadToEnd(file_path), EnvoyException);
-}
-#endif
+// Issue https://github.com/envoyproxy/envoy/issues/25614
```
RyanTheOptimist, line 111: Instead of commenting out the whole test (which changes all the lines in the test) how about changing `#ifndef WIN32` to `#if 0` or some such?

test/per_file_coverage.sh
```diff
@@ -10,6 +10,7 @@ declare -a KNOWN_LOW_COVERAGE=(
 "source/common/config:96.4"
 "source/common/crypto:88.1"
 "source/common/event:95.1" # Emulated edge events guards don't report LCOV
+"source/common/filesystem/posix:96.5" # FileReadToEndNotReadable keeps failing
```
RyanTheOptimist, line 13: It's a bummer that we need to lower the coverage here :/ I see discussed earlier

> It's possible the test could be conditional (check that the permissions preclude reading the file after they've been changed, and if they don't, just log that the test was skipped and don't do the EXPECT), which would provide the coverage in the coverage environment where the test works, and then 'skip' in the environment where the test doesn't work.

Is that still an option? If so, that sounds preferable. (Every time we lower coverage in 1 directory, we lower the aggregate coverage for the repo which has knock-on effects) But maybe it's not possible?

test/common/filesystem/filesystem_impl_test.cc
noah8713: @RyanTheOptimist : thats the proposed approach by @ravenblackx and @phlax . I would have kept it to previous state with issue link. Just that different suggestions. Can you guys consent on whats needed. I dont think this issue needs to spent too much on the brain power. As long as issue ref is linked we are good. Please let me know before I push any other changes.

test/common/filesystem/filesystem_impl_test.cc
noah8713: Done. PTAL. cc @RyanTheOptimist

test/per_file_coverage.sh
RyanTheOptimist: I'm not sure that I quite understand your comment. Have we tried to make the test conditional? If so, did that work? If not, why did it not work? Maybe you've already discussed this and I didn't see it.

noah8713: yes we did discuss that as the thread has grown long for this because for coverage the default failed threshold is just amended to 96.5 upperbound for smaller fix like this. Per @ravenblackx and @phlax discussion lets keep the current PR as suggested. Any future amendments we do later

noah8713:
The current failure is due to bash 
```
E: The repository 'https://packages.microsoft.com/ubuntu/20.04/prod focal Release' is no longer signed.
##[error]Bash exited with code '100'.
Finishing: Bash

```
no formatting other failures @ravenblackx @phlax PTAL

test/common/filesystem/filesystem_impl_test.cc
ravenblackx: I don't mind which way the test is disabled, so long as it's linked to the issue and the code in the test is the version that had coverage but didn't work in some environments, not the version that is a duplicate of an existing test with a name that says it's testing something it's not. :)

Another way to disable it would be to add one of the skip-the-test-and-log-it macros, which might be even better. But I am happy with this as-is.

ravenblackx (approved):

test/common/filesystem/filesystem_impl_test.cc
ravenblackx: It's `GTEST_SKIP()` in case you want to do it that way. That would also play well with any future attempt to make it conditional, since then it could be simply changed to.
```
if (however we detect the test is not going to work) {
  GTEST_SKIP();
}
```

RyanTheOptimist: If we want to disable the test so it doesn't run, then we should just add `DISABLED_` to the name. So that would be:

```
TEST_F(FileSystemImplTest, DISABLED_FileReadToEndNotReadable) {
```

However, I'm still unclear if we've tried to make the test conditional or not. If we make it conditional, we should use `GTEST_SKIP()` as @ravenblackx suggests. If we can make this test conditional that would be much better. @noah8713 have you tried the suggestion here:

> It's possible the test could be conditional (check that the permissions preclude reading the file after they've been changed, and if they don't, just log that the test was skipped and don't do the EXPECT), which would provide the coverage in the coverage environment where the test works, and then 'skip' in the environment where the test doesn't work.

phlax:
ill hold the 1.25 release ( #25828 ) till this has landed so we can backport it

test/common/filesystem/filesystem_impl_test.cc
RyanTheOptimist: Ping

ravenblackx: I think the idea here is making it *not broken* is important enough that landing this without quibbling about the best way to make the test conditional would be worthwhile, with fixing it better being a thing to do later.

A big part of the problem with making it conditional is that we're not even sure what would work, at this point, and it's hard to test because in most environments the test works already - getting it to run in the environment where it doesn't work is tricky, which makes iterating towards a better solution tricky.

RyanTheOptimist: If we're going to land without making test conditional, then lets use `DISABLED_` to disable the test, as that's the supported pattern for doing so.

noah8713: Thanks @ravenblackx and @RyanTheOptimist : made it DISABLED_ and rebased. PTAL

ravenblackx (approved):

test/common/filesystem/filesystem_impl_test.cc
RyanTheOptimist: Thanks. In the future, please do not use rebase and force push, per the guidance in CONTRIBUTING.md:

https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md

> Once your PR is under review, please do not rebase it. If you rebase, you will need to force push to github, and github's user interface will force your reviewer to review the PR from scratch rather than simply look at your latest changes. It's much easier to review new commits and/or merges. We squash rebase the final merged commit so the number of commits you have in the PR don't matter. Again once your PR is assigned a reviewer, unless you need to fix DCO please do not force push. If you need to pull recent changes you can run
>
> ```
> branch=$(git status|head -1|cut -f3 -d\ )
> git checkout main
> git pull
> git checkout "$branch"
> git pull
> ```

RyanTheOptimist (approved):

test/common/filesystem/filesystem_impl_test.cc
noah8713: Ack! Thanks a bunch. Noted!

PR #25642, opened 2023-02-17
Reduce overhead by fixing alignment in SubsetLb

DiazAlan:
Commit Message: Reduce overhead by fixing alignment in SubsetLb
Risk Level: Low

### DIFF
--- a/source/common/upstream/subset_lb.h
+++ b/source/common/upstream/subset_lb.h
@@ -385,7 +385,6 @@ private:
     return absl::nullopt;
   }
 
-  const LoadBalancerType lb_type_;
   const std::unique_ptr<const envoy::config::cluster::v3::Cluster::RingHashLbConfig>
       lb_ring_hash_config_;
   const std::unique_ptr<const envoy::config::cluster::v3::Cluster::MaglevLbConfig>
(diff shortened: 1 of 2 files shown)


source/common/upstream/subset_lb.h
```diff
@@ -426,12 +425,13 @@ class SubsetLoadBalancer : public LoadBalancer, Logger::Loggable<Logger::Id::ups
 
   Stats::Gauge* single_duplicate_stat_{};
 
-  const bool locality_weight_aware_;
-  const bool scale_locality_weight_;
-  const bool list_as_any_;
-
   TimeSource& time_source_;
 
+  const LoadBalancerType lb_type_;
```
adisuissa, line 430: nit: please add a small note about the coalescing of these data members at the end (something like [this](https://github.com/envoyproxy/envoy/blob/1ac0a5948ca317fd418777dab2ffe27935c0a276/source/common/router/config_impl.h#L342)).

adisuissa (commented):
LGTM, modulo small nit.
Thanks!

source/common/upstream/subset_lb.h
DiazAlan: Done

adisuissa (approved):
LGTM, thanks!

mattklein123 (approved):

PR #25643, opened 2023-02-17
Reduce overhead in ZoneAwareLoadBalancerBase by fixing alignment

DiazAlan:
Commit Message: Reduce overhead in ZoneAwareLoadBalancerBase by fixing alignment
Risk Level: Low

### DIFF
--- a/source/common/upstream/load_balancer_impl.h
+++ b/source/common/upstream/load_balancer_impl.h
@@ -430,6 +430,7 @@ private:
 
   // Config for zone aware routing.
   const uint64_t min_cluster_size_;
+  // Keep small members (bools and enums) at the end of class, to reduce alignment overhead.
   const uint32_t routing_enabled_;
   const bool fail_traffic_on_panic_ : 1;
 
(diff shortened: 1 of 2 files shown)


source/common/upstream/load_balancer_impl.h
```diff
@@ -436,6 +428,14 @@ class ZoneAwareLoadBalancerBase : public LoadBalancerBase {
   Common::CallbackHandlePtr priority_update_cb_;
   Common::CallbackHandlePtr local_priority_set_member_update_cb_handle_;
 
+  // Config for zone aware routing.
+  const uint64_t min_cluster_size_;
+  const uint32_t routing_enabled_;
+  const bool fail_traffic_on_panic_ : 1;
```
adisuissa, line 434: nit: please add a small note about the coalescing of these data members at the end (something like [this](https://github.com/envoyproxy/envoy/blob/1ac0a5948ca317fd418777dab2ffe27935c0a276/source/common/router/config_impl.h#L342)).

adisuissa (commented):
Thanks!
LGTM, modulo small nit.

source/common/upstream/load_balancer_impl.h
DiazAlan: Done

adisuissa (approved):
LGTM, thanks!

alyssawilk (approved):
Nice change!

PR #25644, opened 2023-02-17
Reduce overhead in SubsetSelectorImpl by fixing alignment

DiazAlan:
Commit Message: Reduce overhead in SubsetSelectorImpl by fixing alignment
Risk Level: Low

### DIFF
--- a/source/common/upstream/load_balancer_impl.h
+++ b/source/common/upstream/load_balancer_impl.h
@@ -757,10 +757,10 @@ public:
 
 private:
   const std::set<std::string> selector_keys_;
+  const std::set<std::string> fallback_keys_subset_;
   const envoy::config::cluster::v3::Cluster::LbSubsetConfig::LbSubsetSelector::
       LbSubsetSelectorFallbackPolicy fallback_policy_;
-  const std::set<std::string> fallback_keys_subset_;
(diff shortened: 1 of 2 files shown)


source/common/upstream/load_balancer_impl.h
```diff
@@ -757,10 +757,10 @@ class SubsetSelectorImpl : public SubsetSelector {
 
 private:
   const std::set<std::string> selector_keys_;
+  const std::set<std::string> fallback_keys_subset_;
```
adisuissa, line 760: Why was this field moved to the end?

adisuissa (commented):
Thanks! Left a question.
/wait-any

source/common/upstream/load_balancer_impl.h
DiazAlan: I wasn't moved down, it was moved up above an enum

adisuissa: Ah... misread the change. That makes sense.
Can you please add a relevant comment (before the enum) similar to other PRs?

DiazAlan: Done

adisuissa (approved):
LGTM, thanks!

PR #25645, opened 2023-02-17
Reduce overhead in LoadBalancerSubsetInfoImpl by fixing alignment

DiazAlan:
Commit Message: Reduce overhead in LoadBalancerSubsetInfoImpl by fixing alignment
Risk Level: Low

### DIFF
--- a/source/common/upstream/load_balancer_impl.h
+++ b/source/common/upstream/load_balancer_impl.h
@@ -770,10 +770,10 @@ class LoadBalancerSubsetInfoImpl : public LoadBalancerSubsetInfo {
 public:
   LoadBalancerSubsetInfoImpl(
       const envoy::config::cluster::v3::Cluster::LbSubsetConfig& subset_config)
-      : enabled_(!subset_config.subset_selectors().empty()),
-        fallback_policy_(subset_config.fallback_policy()),
-        metadata_fallback_policy_(subset_config.metadata_fallback_policy()),
+      : fallback_policy_(subset_config.fallback_policy()),
         default_subset_(subset_config.default_subset()),
+        metadata_fallback_policy_(subset_config.metadata_fallback_policy()),
+        enabled_(!subset_config.subset_selectors().empty()),
         locality_weight_aware_(subset_config.locality_weight_aware()),
         scale_locality_weight_(subset_config.scale_locality_weight()),
         panic_mode_any_(subset_config.panic_mode_any()), list_as_any_(subset_config.list_as_any()) {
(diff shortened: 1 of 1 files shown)


source/common/upstream/load_balancer_impl.h
```diff
@@ -806,17 +806,17 @@ class LoadBalancerSubsetInfoImpl : public LoadBalancerSubsetInfo {
...
-      metadata_fallback_policy_;
   const ProtobufWkt::Struct default_subset_;
   std::vector<SubsetSelectorPtr> subset_selectors_;
-  const bool locality_weight_aware_;
-  const bool scale_locality_weight_;
-  const bool panic_mode_any_;
-  const bool list_as_any_;
+  const envoy::config::cluster::v3::Cluster::LbSubsetConfig::LbSubsetMetadataFallbackPolicy
+      metadata_fallback_policy_;
+  const bool enabled_ : 1;
```
adisuissa, line 815: nit: please add a small note about the coalescing of these data members at the end (something like [this](https://github.com/envoyproxy/envoy/blob/1ac0a5948ca317fd418777dab2ffe27935c0a276/source/common/router/config_impl.h#L342)).

source/common/upstream/load_balancer_impl.h
```diff
@@ -806,17 +806,17 @@ class LoadBalancerSubsetInfoImpl : public LoadBalancerSubsetInfo {
...
       fallback_policy_;
-  const envoy::config::cluster::v3::Cluster::LbSubsetConfig::LbSubsetMetadataFallbackPolicy
-      metadata_fallback_policy_;
   const ProtobufWkt::Struct default_subset_;
   std::vector<SubsetSelectorPtr> subset_selectors_;
-  const bool locality_weight_aware_;
-  const bool scale_locality_weight_;
-  const bool panic_mode_any_;
-  const bool list_as_any_;
+  const envoy::config::cluster::v3::Cluster::LbSubsetConfig::LbSubsetMetadataFallbackPolicy
```
adisuissa, line 813: I see this field moved but not `fallback_policy_`, why?
Is there added benefit by moving these enum types?

adisuissa (commented):
Thanks!
Left a nit and a question.
Can you also update the PR comment presenting the diffs in the class size?

source/common/upstream/load_balancer_impl.h
DiazAlan: Good catch, I just forgot to pull the other enum down. Technically, moving it does doesn't save memory but I feel it'll be better for consistency to move them down anyway.

source/common/upstream/load_balancer_impl.h
DiazAlan: Done

adisuissa (approved):
LGTM, thanks!

snowp (approved):
Thanks!

PR #25646, opened 2023-02-17
Reduce overhead in health_checker_impl by fixing alignment

DiazAlan:
Commit Message:  Reduce overhead in health_checker_impl by fixing alignment
Risk Level: Low

### DIFF
--- a/source/common/upstream/health_checker_impl.cc
+++ b/source/common/upstream/health_checker_impl.cc
@@ -306,7 +306,7 @@ HttpHealthCheckerImpl::HttpActiveHealthCheckSession::HttpActiveHealthCheckSessio
       local_connection_info_provider_(std::make_shared<Network::ConnectionInfoSetterImpl>(
           Network::Utility::getCanonicalIpv4LoopbackAddress(),
           Network::Utility::getCanonicalIpv4LoopbackAddress())),
-      protocol_(codecClientTypeToProtocol(parent_.codec_client_type_)), expect_reset_{},
+      protocol_(codecClientTypeToProtocol(parent_.codec_client_type_)), expect_reset_(false),
       reuse_connection_(false), request_in_flight_(false) {}
 
 HttpHealthCheckerImpl::HttpActiveHealthCheckSession::~HttpActiveHealthCheckSession() {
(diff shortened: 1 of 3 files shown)


source/common/upstream/health_checker_impl.cc
```diff
@@ -302,10 +302,12 @@ HttpHealthCheckerImpl::HttpActiveHealthCheckSession::HttpActiveHealthCheckSessio
     : ActiveHealthCheckSession(parent, host), parent_(parent),
       response_body_(std::make_unique<Buffer::OwnedImpl>()),
       hostname_(getHostname(host, parent_.host_value_, parent_.cluster_.info())),
-      protocol_(codecClientTypeToProtocol(parent_.codec_client_type_)),
+
       local_connection_info_provider_(std::make_shared<Network::ConnectionInfoSetterImpl>(
           Network::Utility::getCanonicalIpv4LoopbackAddress(),
-          Network::Utility::getCanonicalIpv4LoopbackAddress())) {}
+          Network::Utility::getCanonicalIpv4LoopbackAddress())),
+      protocol_(codecClientTypeToProtocol(parent_.codec_client_type_)), expect_reset_{},
```
adisuissa, line 309:
```suggestion
      protocol_(codecClientTypeToProtocol(parent_.codec_client_type_)), expect_reset_(false),
```

source/common/upstream/health_checker_impl.h
```diff
@@ -206,11 +206,11 @@ class HttpHealthCheckerImpl : public HealthCheckerImplBase {
     Http::ResponseHeaderMapPtr response_headers_;
     Buffer::InstancePtr response_body_;
     const std::string& hostname_;
-    const Http::Protocol protocol_;
     Network::ConnectionInfoProviderSharedPtr local_connection_info_provider_;
-    bool expect_reset_{};
-    bool reuse_connection_ = false;
-    bool request_in_flight_ = false;
+    const Http::Protocol protocol_;
```
adisuissa, line 210: nit: please add a small note about the coalescing of these data members at the end (something like [this](https://github.com/envoyproxy/envoy/blob/1ac0a5948ca317fd418777dab2ffe27935c0a276/source/common/router/config_impl.h#L342)).

adisuissa (commented):
Thanks! Left a couple of nits.

source/common/upstream/health_checker_impl.h
adisuissa, line 210: Can the type of [Protocol](https://github.com/envoyproxy/envoy/blob/1ac0a5948ca317fd418777dab2ffe27935c0a276/envoy/http/protocol.h#L12) be redefined?

source/common/upstream/health_checker_impl.cc
DiazAlan: Done

source/common/upstream/health_checker_impl.h
DiazAlan: Done

DiazAlan: Talked offline, about setting it to uint8. Done

zuercher (approved):
Thanks!

PR #25647, opened 2023-02-17
Reduce overhead in conn_pool_map by fixing alignment

DiazAlan:
Commit Message: Reduce overhead in conn_pool_map by fixing alignment
Risk Level: Low

### DIFF
--- a/source/common/upstream/conn_pool_map.h
+++ b/source/common/upstream/conn_pool_map.h
@@ -84,9 +84,10 @@ private:
   absl::flat_hash_map<KEY_TYPE, std::unique_ptr<POOL_TYPE>> active_pools_;
   Event::Dispatcher& thread_local_dispatcher_;
   std::vector<IdleCb> cached_callbacks_;
-  Common::DebugRecursionChecker recursion_checker_;
   const HostConstSharedPtr host_;
+  // Keep smaller fields near the end to reduce padding
   const ResourcePriority priority_;
+  Common::DebugRecursionChecker recursion_checker_;
 };
 
 } // namespace Upstream


source/common/upstream/conn_pool_map.h
```diff
@@ -84,9 +84,10 @@ template <typename KEY_TYPE, typename POOL_TYPE> class ConnPoolMap {
   absl::flat_hash_map<KEY_TYPE, std::unique_ptr<POOL_TYPE>> active_pools_;
   Event::Dispatcher& thread_local_dispatcher_;
   std::vector<IdleCb> cached_callbacks_;
-  Common::DebugRecursionChecker recursion_checker_;
   const HostConstSharedPtr host_;
+  // Keep smaller fields near the end to reduce padding
   const ResourcePriority priority_;
+  Common::DebugRecursionChecker recursion_checker_;
```
adisuissa, line 90: Is this because `DebugRecursionChecker` has only a single bool field?

adisuissa (commented):
LGTM, left a question.
/wait-any

adisuissa (approved):
LGTM, thanks!

source/common/upstream/conn_pool_map.h
pradeepcrao: Does moving `recursion_checker_` reduce `sizeof` for the class it's in?

DiazAlan: Pre change: Size of ConnPoolMap: 104
Post change: Size of ConnPoolMap: 96

snowp (approved):
Thanks!

PR #25648, opened 2023-02-17
Reduce overhead by fixing alignment in PerPriority

DiazAlan:
Commit Message: Reduce overhead by fixing alignment in PerPriority
Risk Level: Low

### DIFF
--- a/source/common/upstream/cluster_manager_impl.h
+++ b/source/common/upstream/cluster_manager_impl.h
@@ -371,13 +371,13 @@ protected:
   struct ThreadLocalClusterUpdateParams {
     struct PerPriority {
       PerPriority(uint32_t priority, const HostVector& hosts_added, const HostVector& hosts_removed)
-          : priority_(priority), hosts_added_(hosts_added), hosts_removed_(hosts_removed) {}
+          : hosts_added_(hosts_added), hosts_removed_(hosts_removed), priority_(priority) {}
 
-      const uint32_t priority_;
       const HostVector hosts_added_;
       const HostVector hosts_removed_;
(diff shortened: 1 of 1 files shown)


source/common/upstream/cluster_manager_impl.h
```diff
@@ -371,13 +371,13 @@ class ClusterManagerImpl : public ClusterManager,
...
       PerPriority(uint32_t priority, const HostVector& hosts_added, const HostVector& hosts_removed)
-          : priority_(priority), hosts_added_(hosts_added), hosts_removed_(hosts_removed) {}
+          : hosts_added_(hosts_added), hosts_removed_(hosts_removed), priority_(priority) {}
 
-      const uint32_t priority_;
       const HostVector hosts_added_;
       const HostVector hosts_removed_;
       PrioritySet::UpdateHostsParams update_hosts_params_;
       LocalityWeightsConstSharedPtr locality_weights_;
+      const uint32_t priority_;
```
adisuissa, line 380: nit: please add a small note about the coalescing of these data members at the end (something like [this](https://github.com/envoyproxy/envoy/blob/1ac0a5948ca317fd418777dab2ffe27935c0a276/source/common/router/config_impl.h#L342)).

adisuissa (commented):
Thanks!
LGTM modulo nit.

source/common/upstream/cluster_manager_impl.h
DiazAlan: Done

adisuissa (approved):
LGTM, thanks!

zuercher (approved):
Thanks!

PR #25649, opened 2023-02-17
Reduce overhead by fixing alignment in ClusterData

DiazAlan:
Commit Message: Reduce overhead by fixing alignment in ClusterData
Risk Level: Low


adisuissa (approved):
LGTM, thanks!

htuch (approved):

Issue #24100, opened 2022-11-21
http filter: header mutation filter that could be used as normal downstream filter or upstream filter


wbpcode:
*Title*: *http filter: header mutation filter that could be used as normal downstream filter or upstream filter*

*Description*:
`request_headers_to_add` and `response_headers_to_add` and so on in the route are used to mutate request or response headers. However, all these mutations are applied in fixed position.

It would be more flexible because these mutations could run in any position of http filter chain.

~~Partly relate to #23949. IMO, it's too complex to create new `early filter`, but it still make sense to create a http header mutation filter that could run in any position (it could be used as downstream filter or upstream filter).~~

wbpcode:
Hi, @alyssawilk could you take a look to this issue when you have free time? I had make a simple discussion in the #23972 with @kyessenov .

alyssawilk:
yeah flexible header modification filter seems reasonable to me, esp with upstream filters where applying changes in the router may not be "late enough"
If we think it's too complicated to create an early filter are we going to roll that back?

wbpcode:
> yeah flexible header modification filter seems reasonable to me, esp with upstream filters where applying changes in the router may not be "late enough"

Get it. This can also be used to solve the #23972. I will create a PR in the next days.

> If we think it's too complicated to create an early filter are we going to roll that back?
Ignore the `early filter`. This is a expired idea. At that time, I want create new category http filters (early filter) that can be called before route match and tracing start. But finally, I gave up because of the complexity and choose the simple design in the #23949.

alyssawilk:
Right but if it's an expired idea, can we roll the PR back to remove the complexity we don't think we'll use?

wbpcode:
Hmmm, I think we never had a PR that related to the `early filter`? 🤔

The #23949 wasn't a PR to implement `early filter`. It choosen a more simple design and is easy to use or maintain, and it's enough for most scenes.

So, I think there is nothing need to roll back?

PR #25658, opened 2023-02-20
http: added a new dual header mutation filter that could be used as downstream/upstream filter

wbpcode:
Commit Message: http: added a new dual header mutation filter that could be used as downstream/upstream filter
Additional Description:

Check #24100 for more detailed context.

Risk Level: low. new L7 extension.
Testing: unit.
Docs Changes: added.
Release Notes: added.


wbpcode:
cc @alyssawilk here is the pr for #24100. This PR is consist of three commits. 

The first commit added a new `HeaderEvaluator` implementation to the `common/http`. The second PR refactor previous `early_header_mutation` to use the new implementation in the `common/http`. The final commit create a new L7 filter.

CODEOWNERS
```diff
@@ -294,6 +294,8 @@ extensions/filters/http/oauth2 @derekargueta @snowp
 /*/extensions/load_balancing_policies/maglev @wbpcode @UNOWNED
 # Early header mutation
 /*/extensions/http/early_header_mutation/header_mutation @wbpcode @UNOWNED
+# Header mutation
+/*/extensions/filters/http/header_mutation @wbpcode @UNOWNED
```
htuch, line 298: Should this be starting life without an additional maintainer to provide coverage?

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,29 @@
...
+// This extension allows for early header mutation by the substitution formatter.
+message HeaderMutation {
+  // The decoder mutations are applied before the request is forwarded to the upstream cluster.
+  repeated config.common.mutation_rules.v3.HeaderMutation decoder_mutations = 1;
+
+  // The encoder mutations are applied before the response is sent to the downstream client.
+  repeated config.common.mutation_rules.v3.HeaderMutation encoder_mutations = 2;
+}
+
+// Empty filter configuration.
```
htuch, line 27: What is this for?

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,29 @@
...
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/header_mutation/v3;header_mutationv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Header mutation filter configuration]
+// [#extension: envoy.filters.http.header_mutation]
+
+// This extension allows for early header mutation by the substitution formatter.
+message HeaderMutation {
+  // The decoder mutations are applied before the request is forwarded to the upstream cluster.
+  repeated config.common.mutation_rules.v3.HeaderMutation decoder_mutations = 1;
```
htuch, line 21: Does the decoder/encoder terminology still make sense when this is used as an upstream filter?

docs/root/configuration/http/http_filters/header_mutation_filter.rst
```diff
@@ -0,0 +1,10 @@
+.. _config_http_filters_header_mutation:
+
+Header Mutation
+===============
+
+* This filter should be configured with the type URL ``type.googleapis.com/envoy.extensions.filters.http.header_mutation.v3.HeaderMutation``.
```
htuch, line 6: How should an operator think about this in relation to the early HTTP mutation filter that already exists? Is this a strict superset an basically obviates it?

CODEOWNERS
wbpcode: The code/logic of this filter is very simple. I am not sure if it's necessary to have an additional maintainer to help? 🤔

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
wbpcode: Maybe my naming is too ambiguous.

The `HeaderMutation` here is a per route filter proto API. And `HeaderMutationConfig` is a empty filter proto API.

May be using `HeaderMutationPerRoute` and `HeaderMutation` is better?

docs/root/configuration/http/http_filters/header_mutation_filter.rst
wbpcode: This would be a more flexible alternative to the tranditional `headers_to_add`/`headers_to_remove` in most cases.

However, this still be a normal HTTP filter. In some corner cases, this filter cann't be used as alternative of `headers_to_add`/`headers_to_remove`. For example, if `sendDirectLocalReply` is called, then the encoder filters will be skipped.

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
wbpcode: > Does the decoder/encoder terminology still make sense when this is used as an upstream filter?

May be request/response is better?

The `decodeHeaders` will be used to handle the **request headers** in both downstream http filter and upstream http filter for now. Although I personlly think that `decodeHeaders` should be used to handle the **response headers** in upstream http filter.

CODEOWNERS
htuch: Sure, but bus factor of one isn't great for ownership and https://github.com/envoyproxy/envoy/blob/main/EXTENSION_POLICY.md suggests two.

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
htuch: SG with route in the name and explaining this in the comments.

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
htuch: I think request/response makes more sense TBH.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
htuch: OK, maybe inline this explanation.

CODEOWNERS
htuch: Also we could have this other maintainer review this initial implementation to spin them up.

wbpcode: Get it. I wonder do you have free bandwidth for this? 😸  If you have free time, then we needn't to bother third maintainer. Thanks.

htuch: Yeah, I actually have a need to deep dive in some related parts of Envoy, so I can do this if there isn't a more interested party :)

docs/root/configuration/http/http_filters/header_mutation_filter.rst
```diff
@@ -0,0 +1,20 @@
...
+This is a filter that can be used to add, remove, append, or update HTTP headers. It can be added in any position in the filter chain
+and used as downstream or upstream HTTP filter. The filter can be configured to apply the header mutations to the request, response, or both.
+
+
+In most cases, this filter would be a more flexible alternative to the ``request_headers_to_add``, ``request_headers_to_remove``,
+``response_headers_to_add``, and ``response_headers_to_remove`` fields in the :ref:`route configuration <envoy_v3_api_msg_config.route.v3.RouteConfiguration>`.
+Because the filter provides complete control over the position and order of the header mutations and it can be used as upstream filter and
+mutates the request headers after load balancing and host selection.
+
+
```
htuch, line 18: One of the interesting properties of this design, which is probably worth spelling out, is that if a later filter clears the route cache, the mutated headers will influence the new route pick. That was never the case for `request_headers_to_add` etc. I think this makes the new filter quite a bit more useful in some configurations.

htuch (changes requested):
Looks good at API / design level! Some minor docs nits and a couple of Qs.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
```diff
@@ -0,0 +1,20 @@
...
+But note that this still has some limitations. In some corner cases, the header mutations may not be applied because the filter chain processing
+is skipped.
```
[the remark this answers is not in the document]
wbpcode: it's a corner case. If one encoder filter stops the filter chain and calls the send local reply, then the local response will bypass the encoder filter chain.

But it's too corner. Maybe we can remove this note directly?

docs/root/configuration/http/http_filters/header_mutation_filter.rst
wbpcode: This is oneof the target scenes. (And I think may be we can add a `clear_route_cache` option in the filter config in the future? 🤔

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,29 @@
...
+option java_outer_classname = "HeaderMutationProto";
+option java_multiple_files = true;
+option go_package = "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/header_mutation/v3;header_mutationv3";
+option (udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Header mutation filter configuration]
+// [#extension: envoy.filters.http.header_mutation]
+
+// Per route configuration for the header mutation filter.
+message HeaderMutationPerRoute {
```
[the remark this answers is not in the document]
wbpcode: I think the `route` is the most specific level for filter configuration. And for the upstream filters, it still should be possible to apply diffferent configurations for different routes that refer to the same cluster.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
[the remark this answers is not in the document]
htuch: I'd just say "Please note that as an encoder filter, this filter follows the standard rules of when it will execute in situations such as local replies - response headers will not be unconditionally added in cases where the filter would be bypassed."

docs/root/configuration/http/http_filters/header_mutation_filter.rst
htuch: "Header mutations may be used to influence later route picks if the route cache is cleared by a filter executing after the header mutation filter"

Yeah, I think a clear_route_cache option could make sense when there is a use case. I'd split it to a different PR.

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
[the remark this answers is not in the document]
htuch: Sure, it is most specific. But, what if I just want to set header mutation for the whole cluster?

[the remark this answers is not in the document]
wbpcode: Then, non-empty filter configuraion (`HeaderMutation` in the proto) would be a choice.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
```diff
@@ -0,0 +1,25 @@
...
+===============
+
+* This filter should be configured with the type URL ``type.googleapis.com/envoy.extensions.filters.http.header_mutation.v3.HeaderMutation``.
+* :ref:`v3 API reference <envoy_v3_api_msg_extensions.filters.http.header_mutation.v3.HeaderMutation>`
+
+This is a filter that can be used to add, remove, append, or update HTTP headers. It can be added in any position in the filter chain
+and used as downstream or upstream HTTP filter. The filter can be configured to apply the header mutations to the request, response, or both.
+
+
+In most cases, this filter would be a more flexible alternative to the ``request_headers_to_add``, ``request_headers_to_remove``,
```
htuch, line 13: Is it fair to say this alternative use would sometimes also be done via composite filter?

source/extensions/filters/http/header_mutation/header_mutation.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+
+PerRouteHeaderMutation::PerRouteHeaderMutation(const PerRouteProtoConfig& config)
+    : mutations_(config.mutations()) {}
+
+HeaderMutationConfig::HeaderMutationConfig(const ProtoConfig& config)
+    : mutations_(config.mutations()) {}
+
+Http::FilterHeadersStatus HeaderMutation::decodeHeaders(Http::RequestHeaderMap& headers, bool) {
+  config_->mutations().mutateRequestHeaders(headers, decoder_callbacks_->streamInfo());
```
htuch, line 36: Do you want to do both per-route and filter-level config, or have route-specific override the filter config?

source/extensions/filters/http/header_mutation/header_mutation.cc
```diff
@@ -0,0 +1,77 @@
...
+    route_config_->mutations().mutateRequestHeaders(headers, decoder_callbacks_->streamInfo());
+  }
+
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::FilterHeadersStatus HeaderMutation::encodeHeaders(Http::ResponseHeaderMap& headers, bool) {
+  const Http::RequestHeaderMap& request_headers =
+      encoder_callbacks_->streamInfo().getRequestHeaders() == nullptr
+          ? *Http::StaticEmptyHeaders::get().request_headers
```
htuch, line 54: Why is this empty? Don't we know the request headers from earlier?

test/common/http/header_mutation_test.cc
```diff
@@ -0,0 +1,297 @@
+#include "source/common/http/header_mutation.h"
```
htuch, line 1: Do you have any integration tests, in particular ones that cover the upstream filter use case?

wbpcode: I will add some integration tests for this filter.

source/extensions/filters/http/header_mutation/header_mutation.cc
wbpcode: In fact, logically to say, the request headers will never be nullptr.

I will update these code to use a ASSERT here to avoid the nullptr check.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
wbpcode: What do you mean by the `composite filter` here? The `envoy.filters.http.composite` filter or multiple header mutation filters?

source/extensions/filters/http/header_mutation/header_mutation.cc
htuch: Maybe call this out in docs / comments somewhere.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
htuch: envoy.filters.http.composite

test/extensions/filters/http/header_mutation/header_mutation_integration_test.cc
```diff
@@ -0,0 +1,166 @@
...
+      // Upstream header mutation filter.
+      auto filter_0 = http_protocol_options.mutable_http_filters()->Add();
+      filter_0->set_name("upstream-header-mutation");
+      ProtoConfig header_mutation;
+      auto response_mutation =
+          header_mutation.mutable_mutations()->mutable_response_mutations()->Add();
+      response_mutation->mutable_append()->mutable_header()->set_key("upstream-global-flag-header");
+      response_mutation->mutable_append()->mutable_header()->set_value(
+          "upstream-global-flag-header-value");
+      response_mutation->mutable_append()->set_append_action(
```
htuch, line 47: It's a bit jarring to have both YAML and proto generated methods used depending on downstream/upstream. If config helper is missing support for upstream filters, maybe now is the time to add it.

test/extensions/filters/http/header_mutation/header_mutation_integration_test.cc
```diff
@@ -0,0 +1,166 @@
...
+  HeaderMutationIntegrationTest()
+      : HttpIntegrationTest(Http::CodecClient::Type::HTTP1, GetParam()) {}
+
+  void initializeFilter() {
+    config_helper_.addFilter(R"EOF(
+name: donwstream-header-mutation
+typed_config:
+  "@type": type.googleapis.com/envoy.extensions.filters.http.header_mutation.v3.HeaderMutation
+  mutations:
+    response_mutations:
```
htuch, line 26: Is there value in also covering request mutations?

source/extensions/filters/http/header_mutation/header_mutation.cc
wbpcode: I think I has called this out in the proto API.

docs/root/configuration/http/http_filters/header_mutation_filter.rst
wbpcode: I think the answer is yes. The compositer filter could provides very flexible matching. Although for most users, route level configuration is flexible enough.

test/extensions/filters/http/header_mutation/header_mutation_integration_test.cc
wbpcode: ok, I will add new helper function for the upstream http filters.

test/extensions/filters/http/header_mutation/header_mutation_integration_test.cc
wbpcode: get it. I will add some request mutations here.

test/extensions/filters/http/header_mutation/header_mutation_integration_test.cc
wbpcode: Seems that `prependFilter` is enough for this requirement.

api/envoy/extensions/filters/http/header_mutation/v3/header_mutation.proto
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// Per route configuration for the header mutation filter. If this is configured at multiple levels
+// (route level, virtual host level, and route table level), only the most specific one will be used.
+message HeaderMutationPerRoute {
+  Mutations mutations = 1;
+}
+
+// Configuration for the header mutation filter. The mutation rules in the filter configuration will
+// always be applied first.
```
htuch, line 33:
```suggestion
// always be applied first and then the per-route mutation rules, if both are specified.
```

test/extensions/http/early_header_mutation/header_mutation/header_mutation_test.cc
```diff
@@ -59,94 +48,27 @@ TEST(HeaderMutationTest, Basic) {
   HeaderMutation mutation(proto_mutation);
   NiceMock<Envoy::StreamInfo::MockStreamInfo> stream_info;
 
-  {
```
htuch, line 62: Why get rid of all these tests? I realize we want to have this newer filter subsume many of the cases of the early header mutation filter, but not sure why the selective removal of tests in the current PR.

source/extensions/filters/http/header_mutation/header_mutation.cc
htuch: Yeah, I made a small suggestion to make it clearer but you are right.

htuch:
LGTM modulo comment, thanks, this is really nice.

test/extensions/http/early_header_mutation/header_mutation/header_mutation_test.cc
wbpcode: In fact, they ware not removed but move to the https://github.com/envoyproxy/envoy/pull/25658/files#diff-e87c3085497f9392ed99f92ff4d3189cd16810be1574517625f72a08d75b6d7c.

We extracted some common code to the `common/http/header_mutation`, so all these old tests are moved to the common part.

htuch (approved):
LGTM, thanks!

source/extensions/filters/http/header_mutation/header_mutation.cc
```diff
@@ -0,0 +1,79 @@
...
+
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::FilterHeadersStatus HeaderMutation::encodeHeaders(Http::ResponseHeaderMap& headers, bool) {
+  // The request headers will be set to the downstream stream info before the filter chain is
+  // started. And in the current implementation, the upstream filter chain will reuse the downstream
+  // stream info. So the getRequestHeaders() will never return nullptr no matter the filter is
+  // is used as a downstream or upstream filter.
+  ASSERT(encoder_callbacks_->streamInfo().getRequestHeaders() != nullptr);
```
botengyao, line 56: Can we relax this assert here? Seems it is breaking [filter_fuzz_test](https://github.com/envoyproxy/envoy/blob/30cfa5c4619bae49125fff35aacf2f62e81b4410/test/extensions/filters/http/common/fuzz/BUILD#L66) with this corpus:
```
config {
  name: "envoy.filters.http.header_mutation"
  typed_config {
    type_url: "type.googleapis.com/envoy.extensions.filters.http.header_mutation.v3.HeaderMutation"
  }
}
```
Or we may need to change the fuzz test infra? Can you take a look? Thanks!

wbpcode: I think we need a fix here. I just get some new knowledge about the filter chain. It's possible to get a null request headers in the encoding phase.

botengyao: Thanks!

PR #25661, opened 2023-02-20
ci: Fix clang-tidy diff

### LINKED ISSUE #24485: Clang-tidy can fail when run against existing code

phlax:
Fix #24485 

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/ci/run_clang_tidy.sh
+++ b/ci/run_clang_tidy.sh
@@ -126,6 +126,9 @@ else
             git fetch --unshallow
             DIFF_REF="$(git merge-base HEAD "${DIFF_TARGET_BRANCH}")"
         else
+            # TODO(phlax): this is the path used for local CI. Make this work
+            #    similar to above, allow the `remote` to be configurable, and
+            #    document the workflow for devs
             DIFF_REF=$("${ENVOY_SRCDIR}"/tools/git/last_github_commit.sh)
         fi
     fi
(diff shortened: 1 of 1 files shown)


ci/run_clang_tidy.sh
```diff
@@ -113,15 +115,22 @@ elif [[ "${RUN_FULL_CLANG_TIDY}" == 1 ]]; then
...
+        if [[ "${BUILD_REASON}" == *CI ]]; then
+            DIFF_REF="HEAD^"
+        # Presubmit/PR
+        elif [[ -n "${BUILD_REASON}" ]]; then
+            # Common ancestor commit
+            git fetch origin
+            git fetch --unshallow
+            DIFF_REF="$(git merge-base HEAD "${DIFF_TARGET_BRANCH}")"
+        else
+            DIFF_REF=$("${ENVOY_SRCDIR}"/tools/git/last_github_commit.sh)
```
adisuissa, line 129: I guess this applies to running it locally (no pre/post submit-CI).
Will it make sense to have this the same as presubmit-PR sha (so a dev can reproduce the same behavior as the when submitting a PR)?

phlax: this is only set up for PRs - we probably want it to work similar for local - but then we need to make things like the `upstream` (`origin` in this PR) configurable

adisuissa: Ok, let's add a TODO here, referencing the current PR so this can be tracked later. WDYT?

phlax: yep +1 - ill add a comment now - i think the follow up is to resolve this here _and_ to update/add docs about it somewhere

not sure why it takes so long to run this, and also not entirely clear why its necessary to compile everything to run a static analyzer - but either way, as it stands, devs will want to start running this to avoid CI fails

at least once it is run once, and using just the clang-tidy-diff it can run in a ~reasonable time

adisuissa (approved):
LGTM, thanks!

RyanTheOptimist (approved):

PR #25674, opened 2023-02-21
Make test work with both format_ports_as_numbers feature on and off.

pradeepcrao:

Commit Message:
Additional Description:
Risk Level: Low
Testing: Test change
Docs Changes: N/A
Release Notes: N/A


adisuissa (commented):
Thanks for making the test work when the runtime flag is true and false.
If this was a long-term change, I think your PR would have been the right thing to do.
However, as the runtime-flag should be short-lived (~6 months), and we want to easily remove the changes, I think that a less-modular is the right thing to do here.
Specifically, I think that adding `scoped_runtime.mergeValues({{"envoy.reloadable_features.format_ports_as_numbers", "true"}});` to the respective tests is a better approach.
WDYT?

PR #25695, opened 2023-02-21
Use send()/recv() in IoSocketHandleImpl::writev()/readv() when there is  only one IO buffer to work with.

weisisea:
Commit Message: Use send()/recv() in IoSocketHandleImpl::writev()/readv() when there is only one IO buffer to work with.
Additional Description: Performance testing on Redis filter shows that this change gives 7% performance improvement.
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


source/common/api/posix/os_sys_calls_impl.cc
```diff
@@ -42,12 +42,24 @@ SysCallIntResult OsSysCallsImpl::close(os_fd_t fd) {
 }
 
 SysCallSizeResult OsSysCallsImpl::writev(os_fd_t fd, const iovec* iov, int num_iov) {
-  const ssize_t rc = ::writev(fd, iov, num_iov);
+  ssize_t rc;
+  if (num_iov == 1) {
+    // Avoid paying the VFS overhead when there is only one IO buffer to work with
+    rc = ::write(fd, iov[0].iov_base, iov[0].iov_len);
```
soulxu, line 48: `OsSysCallsImpl` is just simple wrap of syscall, we shouldn't hack it here. It seems like the only place write one slice with `writev` is SSL socket transport socket.

https://github.com/envoyproxy/envoy/blob/0086411dd85231434c28ffb6af0ebeaf49e6b02f/source/extensions/transport_sockets/tls/io_handle_bio.cc#L74

if this is the right place, we should change the io handle bio to use `write`.

Maybe we need some benchmark on generic TLS, which probably needs maintainers input.

soulxu: > if this is the right place, we should change the io handle bio to use `IoHandle::write`.
> 

And that needs an different read/write interface IoHandle for accept plain buffer.

weisisea: Thanks for the input!

Actually after talking with my colleague, it looks like that the perf improvement doesn't come from replacing `writev()` with `write()` (the scatter-gather doesn't introduce much overhead), but from replacing `writev()` with `send()`, which will reduce some kernel VFS handling overhead.

I'm thinking whether to move the logic to:
https://github.com/envoyproxy/envoy/blob/0086411dd85231434c28ffb6af0ebeaf49e6b02f/source/common/network/io_socket_handle_impl.cc#L118

weisisea:
After talking with my colleague, it looks like that the perf improvement doesn't come from replacing writev() with write() (the scatter-gather doesn't introduce much overhead), but from replacing writev() with send(), which will reduce some kernel VFS handling overhead.

Planning to change `write()/read()` back to `send()/recv()`

jmarantz:
@weisisea sounds like this is still a WiP?

I'll flip into draft mode until this is sorted out. Thanks.

weisisea:
Sounds good, thanks Joshua!

source/common/network/io_socket_handle_impl.cc
```diff
@@ -95,6 +95,13 @@ Api::IoCallUint64Result IoSocketHandleImpl::readv(uint64_t max_length, Buffer::R
     num_bytes_to_read += slice_length;
   }
   ASSERT(num_bytes_to_read <= max_length);
+
+  if (num_slices_to_read == 1) {
+    // Avoid paying the VFS overhead when there is only one IO buffer to work with
+    return sysCallResultToIoCallResult(
+      Api::OsSysCallsSingleton::get().recv(fd_, iov[0].iov_base, iov[0].iov_len, 0));
```
soulxu, line 102: emm...if this give good performance, I'm little to prefer to replace writev/readv with send/recv at here
https://github.com/envoyproxy/envoy/blob/0086411dd85231434c28ffb6af0ebeaf49e6b02f/source/extensions/transport_sockets/tls/io_handle_bio.cc#L74

But yea, that means we need to figure out what to do for mac and windows. If we just want to focus on linux platform, then this is probably the right place.

soulxu:
And I'm curious whether this benchmark can help here https://github.com/envoyproxy/envoy/blob/main/test/extensions/transport_sockets/tls/tls_throughput_benchmark.cc

source/common/network/io_socket_handle_impl.cc
weisisea: Agree that we can change https://github.com/envoyproxy/envoy/blob/0086411dd85231434c28ffb6af0ebeaf49e6b02f/source/extensions/transport_sockets/tls/io_handle_bio.cc#L74 to use send() since it explicitly specify the number of slices to be 1.

I wonder whether it makes sense to generalize the optimization (not only TLS but other filters like Redis) - this PR tries to look at the number of slices in run time and calls send()/recv() if the number of slices is 1.

source/common/api/win32/os_sys_calls_impl.cc
```diff
@@ -158,6 +158,10 @@ SysCallSizeResult OsSysCallsImpl::pread(os_fd_t fd, void* buffer, size_t length,
   PANIC("not implemented");
 }
 
+SysCallSizeResult OsSysCallsImpl::send(os_fd_t socket, void* buffer, size_t length, int flags) {
+  PANIC("not implemented");
```
yanavlasov, line 162: Why is it "not implemented"? Windows does have the `send` function. Implementing this method will avoid #ifdefs below.

yanavlasov (commented):
/wait-any

source/common/api/win32/os_sys_calls_impl.cc
weisisea: We haven't tested the impact of the change on Windows so I'm trying to keep the logic unchanged for Windows. Does the change look relatively safe for Windows? I can work on enabling it for Windows as well.

yanavlasov (dismissed):

ravenblackx (approved):
Looks okay to me, one optional cleanup.

test/common/buffer/owned_impl_test.cc
```diff
@@ -1348,25 +1348,25 @@ TYPED_TEST(OwnedImplTypedTest, ReserveZeroCommit) {
...
-  os_fd_t pipe_fds[2] = {0, 0};
+  os_fd_t fds[2] = {0, 0};
   auto& os_sys_calls = Api::OsSysCallsSingleton::get();
 #ifdef WIN32
-  ASSERT_EQ(os_sys_calls.socketpair(AF_INET, SOCK_STREAM, 0, pipe_fds).return_value_, 0);
+  ASSERT_EQ(os_sys_calls.socketpair(AF_INET, SOCK_STREAM, 0, fds).return_value_, 0);
 #else
-  ASSERT_EQ(pipe(pipe_fds), 0);
+  ASSERT_EQ(os_sys_calls.socketpair(AF_UNIX, SOCK_STREAM, 0, fds).return_value_, 0);
 #endif
```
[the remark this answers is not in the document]
weisisea: Good suggestion, will change!

weisisea:
Checking how to fix //test/extensions/filters/listener/http_inspector:http_inspector_test on Windows:

```
ERROR: C:/source/test/extensions/filters/listener/http_inspector/BUILD:15:24: Testing //test/extensions/filters/listener/http_inspector:http_inspector_test failed: (Exit 34): Remote Execution Failure:
Invalid Argument: stdout limit exceeded
  Unrecognized error detail: type_url: "type.googleapis.com/google.devtools.remotebuildbot.CommandStatus"
value: "\b\001"
```

https://dev.azure.com/cncf/envoy/_build/results?buildId=129848&view=logs&j=307ebadc-bea7-5cb1-4513-6195e03b9250&t=001b70b1-8584-5b9e-88d7-53fe6d97a0fc

ravenblackx (approved):

ravenblackx (approved):

weisisea:
@yanavlasov Thanks a lot for merging, Yan!

PR #25696, opened 2023-02-21
mobile: use Docker image with Java & Android NDK installed

jpsim:
This allows us to move the `android_build` CI job from macOS to Linux, freeing up some CI capacity.

It also allows us to remove the Java installation step from some other CI jobs since it's now available directly in the docker image the job runs on.

Related PRs:

* https://github.com/envoyproxy/envoy-build-tools/pull/190
* https://github.com/envoyproxy/envoy-build-tools/pull/197
* https://github.com/envoyproxy/envoy-build-tools/pull/198

Commit Message:
Additional Description:
Risk Level: Low, CI only
Testing: CI
Docs Changes: None
Release Notes: None
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim:
I tried moving the jobs that run example apps in the Android emulator to Linux with this new container, but ran into several issues:

* Several hundreds of MBs of apt packages are needed to get the emulator to run
* Even then, the lack of hardware acceleration prevents the emulators from booting
* According to the docs for the [Android Emulator Runner GitHub Action](https://github.com/marketplace/actions/android-emulator-runner#can-i-use-this-action-on-github-hosted-linux-vms), they highly discourage using this on Docker containers running in GitHub Actions

So unless we move to self-hosted runners with hardware acceleration, I don't think we can reasonably do this right now.

.github/workflows/android_build.yml
```diff
@@ -72,7 +70,7 @@ jobs:
       with:
         timeout_minutes: 6
         max_attempts: 3
-        command: cd mobile && ./ci/mac_start_emulator.sh
+        command: ./mobile/ci/start_android_emulator.sh
```
jpsim, line 73: I renamed this file when I was trying to get these jobs to run on Linux, and even though that didn't work, I felt like this new name for the script is better and worth keeping.

phlax (approved):
lgtm to me - not sure if you want to get a stamp from someone more familiar with mobile build

jpsim:
> not sure if you want to get a stamp from someone more familiar with mobile build

I'll get someone else from Envoy Mobile to review. I'll also inform EngFlow about us using this larger Docker image.

abeyad (approved):
LGTM!

PR #25702, opened 2023-02-22
golang filter extension: post header write operations to the envoy safe thread.

### LINKED ISSUE #25289: golang filter extension: header write operation should be dispatched to the safe envoy worker thread.
Now, we may [setting a header](https://github.com/envoyproxy/envoy/blob/79ff4e73f496b76a5e6a45e9745241e0fdb8c6f8/contrib/golang/filters/http/source/golang_filter.cc#L642) in a Go thread, it may break memory safety, since another C++ http filter may reading the header in the Envoy worker thread.
That means the reading in Envoy worker thread, and writing in Go thread.

So, we should dispatch the write operation to the Envoy worker thread, when the write operation happens in a Go thread.

doujiang24:
To avoid race between reading in the envoy worker thread and writing in the Go thread.

fix #25289

Commit Message:
Additional Description:
Risk Level: low
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]


contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -639,8 +639,29 @@ CAPIStatus Filter::setHeader(absl::string_view key, absl::string_view value) {
     ENVOY_LOG(debug, "invoking cgo api at invalid phase: {}", __func__);
     return CAPIStatus::CAPIInvalidPhase;
   }
-  headers_->setCopy(Http::LowerCaseString(key), value);
-  onHeadersModified();
+  if (state.isThreadSafe()) {
+    // it's safe to write header in the safe thread.
+    headers_->setCopy(Http::LowerCaseString(key), value);
+    onHeadersModified();
+  } else {
```
wangfakang, line 646: It's needs to add comments to explain that the API does not take effect immediately, and the API on the GO side comments also.

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -639,8 +639,29 @@ CAPIStatus Filter::setHeader(absl::string_view key, absl::string_view value) {
     ENVOY_LOG(debug, "invoking cgo api at invalid phase: {}", __func__);
     return CAPIStatus::CAPIInvalidPhase;
   }
-  headers_->setCopy(Http::LowerCaseString(key), value);
-  onHeadersModified();
+  if (state.isThreadSafe()) {
```
wangfakang, line 642: Here is a high level question needs to be confirmed that the [headers_](https://github.com/envoyproxy/envoy/blob/main/contrib/golang/filters/http/source/golang_filter.h#L186
) is already protected by a lock and Is it necessary to use the dispatcher method?

```c++
Http::RequestOrResponseHeaderMap* headers_ ABSL_GUARDED_BY(mutex_){nullptr};
```

doujiang24: Yes, it is.
the read operation may come from another previous filter that cached the header pointer, and reading the header while it's processing data and the golang filter is processing the header.
And, the `mutex_` just protects the `headers_` pointer address, and won't protect the map contents of the `headers_`.

i.e.
```
          http_filters:
          - name: filter_read_header_while_processing_data_buffer
            # headers continue, processing data, get a value from header in envoy worker thread.
          - name: envoy.filters.http.golang
            # writing a value to header in go thread.
```

yep, it's not a very normal usage, but we'd better not assume other filters won't do this.
this improves the robustness of the golang filter when working with other filters together.

contrib/golang/filters/http/source/golang_filter.cc
doujiang24: Done, thanks.
Also, changed prefetch the values in `header.Set` in the Go side, so that users won't get outdated values, in the following case:
```
# async goroutine, in Go thread
header.Set("foo", "bar")
header.Get("foo")
```

wangfakang (approved):
LGTM

doujiang24:
friendly ping @wbpcode 

sync to the latest main branch after resolving merge conflict.

contrib/golang/filters/http/source/go/pkg/http/type.go
```diff
@@ -82,6 +82,12 @@ func (h *httpHeaderMap) Values(key string) []string {
 }
 
 func (h *httpHeaderMap) Set(key, value string) {
+	// Get all header values first before setting a value, since the set operation may not take affects immediately
+	// when it's invoked in a Go thread, instead, it will post a callback to run in the envoy worker thread.
+	// Otherwise, we may get outdated values.
```
wbpcode, line 87: Sorry, but I don't get the logic. Will `cAPI.HttpCopyHeaders` waiting all pending modificattions to be completed and return latest values?

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -639,11 +641,34 @@ CAPIStatus Filter::setHeader(absl::string_view key, absl::string_view value) {
...
+    // it's safe to write header in the safe thread.
+    headers_->setCopy(Http::LowerCaseString(key), value);
+    onHeadersModified();
+  } else {
+    // should deep copy the string_view before post to dipatcher callback.
+    auto keyStr = std::string(key);
+    auto valueStr = std::string(value);
```
wbpcode, line 651: `key_str` and `value_str`

contrib/golang/filters/http/source/golang_filter.cc
```diff
@@ -659,8 +684,28 @@ CAPIStatus Filter::removeHeader(absl::string_view key) {
...
   }
-  headers_->remove(Http::LowerCaseString(key));
-  onHeadersModified();
+  if (state.isThreadSafe()) {
+    // it's safe to write header in the safe thread.
+    headers_->remove(Http::LowerCaseString(key));
+    onHeadersModified();
+  } else {
+    // should deep copy the string_view before post to dipatcher callback.
+    auto keyStr = std::string(key);
```
wbpcode, line 693: `key_str`

wbpcode (commented):
Thanks for you contributions. And minor comments are added. Thanks again.

contrib/golang/filters/http/source/go/pkg/http/type.go
doujiang24: Nope, `cAPI.HttpCopyHeaders` only delivers the pending modification, won't wait for it to be completed.

So, we might get outdated values in a following `Get` call after `Set`, i.e.
```
header.Set("foo", "bar") // the pending modification may haven't take affects yet.
header.Get("foo") // might return the outdated value of foo
```

Also, updated the comment clearly.

doujiang24:
Thanks @wbpcode , all comments are addressed, please help to take another look.

contrib/golang/filters/http/source/go/pkg/http/type.go
wbpcode: > Nope, cAPI.HttpCopyHeaders only delivers the pending modification, won't wait for it to be completed.

Then how can the `cAPI.HttpCopyHeaders` prevent you from getting outdated value?

doujiang24: We copy the headers from Envoy first, then, the values on the Go side are in sync.
After that, we set the values on both the Go side and the Envoy side, keeping the values always synced.

Actually, it's not real-time synced, since changes might delay on the Envoy side, but we assume the changes must happen, which means we guarantee eventual consistency.

wbpcode (approved):
LGTM. Thanks.

PR #25706, opened 2023-02-22
[TESTING/WIP] attempt to improve ci caching somehow

### LINKED ISSUE #24001: Improve non-RBE caching for release branches (inc `main`)
Currently we are having to manually disable local caching on non-`main` release branches

this is a bit of a kludge as it needs to be applied to new release branches as they are created and is easily forgotten 

also it would be good for them to have their own caches - per-`VERSION` perhaps

phlax:
this should allow release branches to cache independently.

Possibly fix #24001

Commit Message:
Additional Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.azure-pipelines/bazel.yml
+++ b/.azure-pipelines/bazel.yml
@@ -21,7 +21,7 @@ parameters:
 steps:
 - task: Cache@2
   inputs:
-    key: '"${{ parameters.ciTarget }}" | ./WORKSPACE | **/*.bzl'
+    key: '"${{ parameters.ciTarget }}" | ./VERSION.txt | ./WORKSPACE | **/*.bzl'
     path: $(Build.StagingDirectory)/repository_cache
   continueOnError: true
 
--- a/.azure-pipelines/pipelines.yml
+++ b/.azure-pipelines/pipelines.yml
@@ -129,7 +129,7 @@ stages:
     steps:
     - task: Cache@2
       inputs:
-        key: "format | ./WORKSPACE | **/*.bzl"
+        key: "format | ./VERSION.txt | ./WORKSPACE | **/*.bzl"
         path: $(Build.StagingDirectory)/repository_cache
       continueOnError: true
 
@@ -160,7 +160,7 @@ stages:
     steps:
     - task: Cache@2
       inputs:
-        key: "proto_format | ./WORKSPACE | **/*.bzl"
+        key: "proto_format | ./VERSION.txt | ./WORKSPACE | **/*.bzl"
         path: $(Build.StagingDirectory)/repository_cache
       continueOnError: true
 
@@ -191,7 +191,7 @@ stages:
     steps:
     - task: Cache@2
       inputs:
-        key: "docs | ./WORKSPACE | **/*.bzl"
+        key: "docs | ./VERSION.txt | ./WORKSPACE | **/*.bzl"
         path: $(Build.StagingDirectory)/repository_cache
       continueOnError: true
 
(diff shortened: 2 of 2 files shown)


phlax:
cc @lizan

phlax:
@lizan iiuc with this (added to all branches) we can revert the changes like here https://github.com/envoyproxy/envoy/commit/29d5341d6e5098dc369ae987e497d8f740e81c53#diff-634d3872fb352a01a0a3780a6c242e0c20f1df44c34efedf90c2f1fa3697bc49L58

phlax:
> ... we can revert the changes like here ...

digging into this further and i see now this is using `$(localBuildCache)` which is unrelated to the azp cache that this affects

this still seems like a good idea to me

phlax:
im also wondering if there is some way to make `$(localBuildCache)` work per-version so we can use it for `main` and other release branches

im looking at `--remote_instance_name=${BAZEL_REMOTE_INSTANCE}"` and im wondering if there is any pollution happening as it is between these branches

one thing i do see quite a bit is bazel failing to die in CI - ive wondered if this is related to release branch CI contamination, as usually when i see this locally its because im switching branches/bazel version/config/etc and bazel borks

phlax:
re this PR im now wondering if this is actually helpful if the repo_cache is just shas and everything can share those

phlax:
> ... and i guess also wondering if we can use a per-version ...

assuming we are using https://github.com/buchgr/bazel-remote my reading is that this should be possible

phlax:
@lizan i think what i was trying to do here initially was wrong, but maybe my reasons not so much so

the 2 things i was trying to do

- improve caching on release branches so arm and coverage jobs dont fail so frequently
- figure out if there is some kinda contamination between release branches and main in bazel caches

lizan:
These cache are for downloaded artifacts of dependencies, and bazel will check it and download if neccessary. So it isn't really a build cache. We already hashes repository_locations.bzl and I don't see a reason to add VERSION there.

phlax:
> I don't see a reason to add VERSION there.

yep, agreed

i have 2 ~related questions

1. Can we set something line `--remote_instance_name` for the release branches local non-RBE caches - ie with `$(localBuildCache)`
2. Is it possible that the RBE caches are in some way clashing because `main` and release branches both use the same `--remote_instance_name`

lizan:
> * Can we set something line `--remote_instance_name` for the release branches local non-RBE caches - ie with `$(localBuildCache)`

We need to enable `--enable_ac_key_instance_mangling` to bazel_remote instance that we set up in ci-infra. Then the instance_name will be part of the key in non-RBE caches.

> * Is it possible that the RBE caches are in some way clashing because `main` and release branches both use the same `--remote_instance_name`

It could happen when main branch and release branches use RBE image ID but it should cause problem for us. If that's not the case it should be a bug of Bazel.

phlax:
> We need to enable --enable_ac_key_instance_mangling to bazel_remote instance that we set up in ci-infra.

if we can do this it would make working on the release branches (and releasing) a lot easier/more reliable

> It could happen when main branch and release branches use RBE image ID 

my main reason for asking is that i sometimes sees more/less instances of bazel refusing to die at the end of a job and then it failing

i cant say with total confidence that there is a particular pattern - but it _seems_ to me at least when there is activity on different branches at the same time

my local experience of bazel not going away (frequent) is switching branches and confusing bazel

phlax:
i have separated these issues:

- #25736 
- #25735

PR #25717, opened 2023-02-22
mobile: Remove YAML-based config in the EngineBuilder

abeyad:
With this commit, the EngineBuilder only does protocol buffer based construction of the Bootstrap config, no more
YAML string parsing. The protocol buffer based construction has been tested in production in the Lyft app, so this is
deemed a safe change.

As part of the cleanup, we also remove some unused methods in BaseClientIntegrationTest that invoke the EngineBuilder.


mobile/library/cc/engine_builder.h
```diff
@@ -41,8 +41,15 @@ struct NodeLocality {
...
-  EngineBuilder(std::string config_template);
+  // THIS CONSTRUCTOR SHOULD ONLY BE CALLED BY TESTS. Further uses of this constructor are highly
+  // discouraged as this API is likely to go away soon.
+  //
+  // Creates an EngineBuilder from the provided YAML config string. If this constructor is invoked,
+  // then only the config provided in the string will be used to generate the Envoy engine; none of
+  // the builder function invocations will take effect.
+  //
+  // TODO(abeyad): remove the need for this constructor.
+  explicit EngineBuilder(std::string config);
```
jpsim, line 52: Should we remove this completely in favor of `setOverrideConfigForTests` to make it more explicit?

mobile/library/common/config/BUILD
```diff
@@ -6,25 +6,15 @@ envoy_package()
 
 envoy_cc_library(
     name = "config_lib",
```
jpsim, line 8: Is this basically now only used to store the `certificates.inc` file? If so, maybe it could be renamed to something like `certificates_lib`?

mobile/test/cc/unit/envoy_config_test.cc
```diff
@@ -50,250 +48,197 @@ TEST(TestConfig, ConfigIsApplied) {
...
 
 TEST(TestConfig, EnableDrainPostDnsRefresh) {
   EngineBuilder engine_builder;
 
-  std::string config_str = engine_builder.generateConfigStr();
-  ASSERT_THAT(config_str, HasSubstr("&enable_drain_post_dns_refresh false"));
-  envoy::config::bootstrap::v3::Bootstrap bootstrap;
-  TestUtility::loadFromYaml(absl::StrCat(config_header, config_str), bootstrap);
+  std::unique_ptr<Bootstrap> bootstrap = engine_builder.generateBootstrap();
+  EXPECT_THAT(bootstrap->ShortDebugString(), Not(HasSubstr("enable_drain_post_dns_refresh")));
```
jpsim, line 199: TIL about `ShortDebugString()`

jpsim (dismissed):
Nice cleanup!

abeyad (commented):
Thanks for the feedback JP!  Addressed the comments, PTAL.

jpsim (approved):
Thanks for the big cleanup!

RyanTheOptimist (approved):
LOVE IT!

PR #25728, opened 2023-02-22
network: share filter state objects always

kyessenov:
Commit Message: Move sharing code to support internal listeners. Because of hashing, the downstream filters can only use the upstream filter state to access the objects shared with the upstream internal listener. This allows easy access from the downstream filters for the back-propagated information.
Additional Description: Example:

HTTP Listener 1 writes a shared object `x_i` per request,
Cluster 1 has internal endpoint for listener 2. One of `x_i` objects from the transport options is written.
Listener 2 writes CONNECT response headers back to `x_i` from cluster 1.

Because the upstream is shared, HTTP filter on listener 1 should consult the connection filter state of listener 2 for the tunnel headers. But because of the user space socket, it's not directly accessible. The work around, is to create shared place holders that would be available in the upstream cluster 1 connection filter state.

Risk Level: low
Testing: regression
Docs Changes: none
Release Notes: none

### DIFF
--- a/source/common/network/connection_impl.cc
+++ b/source/common/network/connection_impl.cc
@@ -888,6 +888,16 @@ ClientConnectionImpl::ClientConnectionImpl(
       stream_info_(dispatcher_.timeSource(), socket_->connectionInfoProviderSharedPtr()) {
 
   stream_info_.setUpstreamInfo(std::make_shared<StreamInfo::UpstreamInfoImpl>());
+
+  if (transport_options) {
+    for (const auto& object : transport_options->downstreamSharedFilterStateObjects()) {
+      // This does not throw as all objects are distinctly named and the stream info is empty.
+      stream_info_.filterState()->setData(object.name_, object.data_, object.state_type_,
+                                          StreamInfo::FilterState::LifeSpan::Connection,
+                                          object.stream_sharing_);
+    }
+  }
+
   // There are no meaningful socket options or source address semantics for
   // non-IP sockets, so skip.
   if (socket_->connectionInfoProviderSharedPtr()->remoteAddress()->ip() == nullptr) {
@@ -922,15 +932,6 @@ ClientConnectionImpl::ClientConnectionImpl(
 
       // Trigger a write event to close this connection out-of-band.
       ioHandle().activateFileEvents(Event::FileReadyType::Write);
-    }
-  }
-
-  if (transport_options) {
-    for (const auto& object : transport_options->downstreamSharedFilterStateObjects()) {
-      // This does not throw as all objects are distinctly named and the stream info is empty.
-      stream_info_.filterState()->setData(object.name_, object.data_, object.state_type_,
-                                          StreamInfo::FilterState::LifeSpan::Connection,
-                                          object.stream_sharing_);
     }
   }
 }
--- a/test/common/network/connection_impl_test.cc
+++ b/test/common/network/connection_impl_test.cc
@@ -137,12 +137,12 @@ public:
   Buffer::Instance& readBuffer() { return *read_buffer_; }
 };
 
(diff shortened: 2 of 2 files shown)


htuch:
I think this is safe, but curious why no new tests given this is a behavioral change. Tagging @alyssawilk as reviewer for loopback domain expertise.

kyessenov:
I think it's worth adding tests - this looks like a basic mistake with early return. The reason is probably we'd need upstream filters and internal listeners filters - two fairly advanced features.

I'm also open to a better way on the back-propagation of the tunnel info back to the inner stream listener on the response side. I made it work with this change but it feels awkward.

source/common/network/connection_impl.cc
```diff
@@ -888,6 +888,16 @@ ClientConnectionImpl::ClientConnectionImpl(
       stream_info_(dispatcher_.timeSource(), socket_->connectionInfoProviderSharedPtr()) {
 
   stream_info_.setUpstreamInfo(std::make_shared<StreamInfo::UpstreamInfoImpl>());
+
```
yanjunxiang-google, line 891: IIUC, we moving this ahead is because the following code will return for internal connection since its remoteAddress type is EnvoyInternal, not IP?

  if (socket_->connectionInfoProviderSharedPtr()->remoteAddress()->ip() == nullptr) {
    return;
  }

https://github.com/envoyproxy/envoy/blob/bd480aa2a62af2885f27983f6d1fa96cf51e8fdf/source/common/network/connection_impl.cc#L893

https://github.com/envoyproxy/envoy/blob/bd480aa2a62af2885f27983f6d1fa96cf51e8fdf/envoy/network/address.h#L198

However, one thing confusing me is that even without this change, we had observed that the filter state stored by the external listener filter are able to shared to upstream and got retrieved correctly by the filters in the internal listener.

kyessenov: Yes, internal listener merge is in a different location here:

https://github.com/envoyproxy/envoy/blob/bd480aa2a62af2885f27983f6d1fa96cf51e8fdf/source/extensions/bootstrap/internal_listener/active_internal_listener.cc#L65

So what happens is that a shared object in the downstream filter state for the internal listener, but not in the upstream connection filter state. From downstream, we can access the upstream filter state but not the next internal downstream filter state.

htuch:
LGTM but i will defer to @alyssawilk for final review / merge.

test/common/network/connection_impl_test.cc
```diff
@@ -3086,8 +3093,23 @@ TEST_F(InternalClientConnectionImplTest,
...
 }
 
-class ClientConnectionWithCustomRawBufferSocketTest : public ConnectionImplTest {
+class ClientConnectionWithCustomRawBufferSocketTest : public ConnectionImplTestBase,
+                                                      public testing::TestWithParam<Address::Type> {
 protected:
+  void setUpBasicConnection() {
+    Address::InstanceConstSharedPtr address;
+    switch (GetParam()) {
+    case Address::Type::Pipe:
```
yanjunxiang-google, line 3102: Thanks for adding the test!
I am a little bit confused why adding the test to cover the address Type::Pipe can help us verify the filter state sharing across downstream filter<->upstream filter<->next downstream filter.

kyessenov: The error was that sharing to the upstream connection was guarded by "IP" type, so both pipes and internal addresses were broken. I can't use internal address here since this is a core, and the internal address factory is an extension.

yanjunxiang-google:
LGTM

alyssawilk:
I think this needs non-googler pass?

htuch:
OK, @mattklein123 can you take a look? I think of the details of this PR more like a bug fix, but maybe the general strategy for filter sharing would benefit from non-Googler input.

Issue #25692, opened 2023-02-21
Envoy not using locality weights for load balancing within the same priority


liamhazell:
*Title*:  Envoy not using locality weights for load balancing within the same priority

*Description*:
I'm messing around with the locality weighted load balancing, and found that if I have 2 endpoints that are within the same priority that use different load balancing weights the weights aren't respected. Using 2 endpoints that have a weight of 2 and 1 should be routed 67% and 33% respectively based on these [docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/locality_weight#:~:text=Assume%20a%20simple%20set%2Dup%20with%202%20localities%20X%20and%20Y%2C%20where%20X%20has%20a%20locality%20weight%20of%201%20and%20Y%20has%20a%20locality%20weight%20of%202%2C%20L%3DY%20100%25%20available%2C%20with%20default%20overprovisioning%20factor%201.4.). Traffic ends up being split 50/50 between the endpoints however. I thought that I was misconfiguring something, however slightly modifying the [example config](https://github.com/envoyproxy/envoy/blob/main/examples/locality-load-balancing/envoy-proxy.yaml) for locality load balancing yields the same result. I'm wondering if I have a misunderstanding of this feature, or if this isn't working as expected.
 
*Repro steps*:
Modify the example config from the [example directory](https://github.com/envoyproxy/envoy/blob/main/examples/locality-load-balancing/envoy-proxy.yaml) to have 2 endpoints with priorities of 0 with a weight of 1 and 2:

```
      endpoints:
      - locality:
          region: local
          zone: zone-1
        load_balancing_weight: 2
        priority: 0  # highest
        lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: backend-local-1
                port_value: 8080
            health_check_config:
              port_value: 8080
            hostname: backend-local-1
      - locality:
          region: local
          zone: zone-2
        load_balancing_weight: 1
        priority: 0
        lb_endpoints:
        - endpoint:
            address:
              socket_address:
                address: backend-local-2
                port_value: 8080
            health_check_config:
              port_value: 8080
            hostname: backend-local-2
```

*Config*:
Full config used:
```
node:
  cluster: test-cluster
  id: test-id
admin:
  address:
    socket_address:
      address: 0.0.0.0
      port_value: 8001
static_resources:
  listeners:
  - name: backend
    address:
      socket_address:
        address: 0.0.0.0
        port_value: 3000
[...]

PR #25744, opened 2023-02-23
examples: Add locality_weighted_lb_config to locality-load-balancing example

liamhazell:
Commit Message: Add locality_weighted_lb_config to locality-load-balancing example
Additional Description: For the load_balancing_weight option to be used, this option must be set. Discussed in https://github.com/envoyproxy/envoy/issues/25692
Risk Level: None
Testing:
Docs Changes: 
Release Notes:
Platform Specific Features:
[Optional Runtime guard:]
[Optional Fixes #Issue]
[Optional Deprecated:]

Wasn't sure if there was a good spot to update the docs. The page on the locality weighted load balancer does describe that you need to set this [field](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/locality_weight#arch-overview-load-balancing-locality-weighted-lb:~:text=Locality%20weighted%20load%20balancing%20is%20configured%20by%20setting%20locality_weighted_lb_config%20in%20the%20cluster), it was just user error in not knowing that you need to have the `{}` present. :)

### DIFF
--- a/examples/locality-load-balancing/envoy-proxy.yaml
+++ b/examples/locality-load-balancing/envoy-proxy.yaml
@@ -39,6 +39,8 @@ static_resources:
   - name: backend
     type: STRICT_DNS
     lb_policy: ROUND_ROBIN
+    common_lb_config:
+      locality_weighted_lb_config: {} # Must be set for load_balancing_weight to be used
     health_checks:
     - interval: 2s
       timeout: 3s
--- a/examples/locality-load-balancing/envoy-proxy.yaml
+++ b/examples/locality-load-balancing/envoy-proxy.yaml
@@ -40,7 +40,7 @@ static_resources:
     type: STRICT_DNS
     lb_policy: ROUND_ROBIN
     common_lb_config:
-      locality_weighted_lb_config: {} # Must be set for load_balancing_weight to be used
+      locality_weighted_lb_config: {}
     health_checks:
     - interval: 2s
       timeout: 3s
--- a/docs/root/start/sandboxes/locality_load_balancing.rst
+++ b/docs/root/start/sandboxes/locality_load_balancing.rst
@@ -42,6 +42,9 @@ To build this sandbox example and start the example services, run the following 
 
 The locality configuration is set in the client container via static Envoy configuration file. Please refer to the ``cluster`` section of the :download:`proxy configuration <_include/locality-load-balancing/envoy-proxy.yaml>` file.
 
+.. note::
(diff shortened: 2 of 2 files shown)


examples/locality-load-balancing/envoy-proxy.yaml
```diff
@@ -39,6 +39,8 @@ static_resources:
   - name: backend
     type: STRICT_DNS
     lb_policy: ROUND_ROBIN
+    common_lb_config:
+      locality_weighted_lb_config: {} # Must be set for load_balancing_weight to be used
```
phlax, line 43: i would avoid adding this here - can we add something in the actual example docs here - https://github.com/envoyproxy/envoy/blob/main/docs/root/start/sandboxes/locality_load_balancing.rst (arggh the github ui is useless here)

we probably want to make this a `.. note::` somewhere to draw attention to it

ie

```rst

... note::
    The ``locality_weighted_lb_config`` must be set in ``common_lb_config``
    for the ``load_balancing_weight`` to be used.

```

or similar

phlax (commented):
brilliant, thanks for following up on this @liamhazell

examples/locality-load-balancing/envoy-proxy.yaml
phlax: currently rendered page https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/locality_load_balancing

probably we could put it at the end of the first section, just after where it mentions the config

liamhazell: Fixed, but I messed up the DCO trying to use my full name let me fix that

liamhazell:
Also @phlax, would you be open to having an example contributed for the built in zone-aware routing? I see there currently isn't one, however I was messing around with that too and think it might be useful to have. Happy to try and contribute that when I have time

phlax:
>  would you be open to having an example contributed for the built in zone-aware routing? 

very much so - scanning the docs and it looks like the kinda complexity that is best described by example

phlax (approved):
lgtm, thanks @liamhazell 

would be good to demonstrate the weights in this sandbox, altho its already quite complicated

liamhazell:
> would be good to demonstrate the weights in this sandbox, altho its already quite complicated

Yeah thought this would be best left as an exercise for the reader to mess around with the settings


> very much so - scanning the docs and it looks like the kinda complexity that is best described by example

cool let me see what I can do!

